Issue #70, opened 2024-04-26
Deposit API Lambda CDK Template


AshtonStephens:
## Deposit API Lambda CDK Template

This is part of #50

Create the lambda CDK template that will be used to spin up the Deposit API Rust lambda.

PR #99, opened 2024-05-01
Add Autogenerated Rust Emily Client

AshtonStephens:
## Description

**This ticket adds the autogeneration of a rust client for Emily from the OpenAPI template that smithy generates.**

The `emily` crate will live in `.generated-sources/emily` and its sources will be autogenerated via the npm build scripts, and that npm build script can be triggered by the `build.rs` script in the `emily` directory.

- Add client autogeneration
- Perform client autogeneration during build of `emily`
- Adds dependencies to GitHub workflow
    - Smithy `1.47.0`
    - Java `21` 

### Testing

- I've tested this locally with my MacOS machine.
- I need to test it on my x86 machine.


.generated-sources/emily/build.rs
```diff
@@ -0,0 +1,47 @@
+use std::process::Command;
+use std::env;
+
+fn main() {
+
+    // Ensure that we rerun if the API changes.
+    println!("cargo:rerun-if-changed=../../emily/api-definition/models");
+    println!("cargo:rerun-if-changed=build.rs");
```
djordon, line 8: Is the goal of this line to prevent this build script from being rerun all the time? I guess to cut down on the overhead of the `npm` commands?

setup.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# `.generated-sources/emily/Cargo.toml` changes when the build script runs.
+git update-index --assume-unchanged .generated-sources/emily/Cargo.toml
+# to undo: `git update-index --no-assume-unchanged .generated-sources/emily/Cargo.toml`
```
djordon, line 5: Oh very interesting... I don't have more to add than that.

Cargo.toml
```diff
@@ -12,3 +13,4 @@ debug = 0
 [profile.release]
 lto = "thin"
 codegen-units = 16
+emily = { version = "0.1.0", path = ".generated-sources/emily" }
```
djordon, line 16: Does this go here?

djordon (commented):
I had some clarifying questions but basically seems fine.

Cargo.toml
AshtonStephens: Ah, no not yet. I'll remove that in the next commit because I have to fix the CI.

.generated-sources/emily/build.rs
AshtonStephens: yep

Cargo.toml
AshtonStephens: Removed.

setup.sh
AshtonStephens: Yeah, it's *blocking* if people don't run this, but they'll want to.

When the openapi generator creates the client it overwrites `.generated-sources/emily/Cargo.toml` with the dependencies of the auto-generated code. Because it's pushed in the repo git won't let you switch to another branch without pushing your changes to it or stashing it, despite it being in the`.gitignore`.

This in itself isn't too bad, but if you're using VS code with the rust analyzer extension it'll automatically run `build.rs` occasionally, meaning you have to race the extension running the build command to switch branches.

The `Cargo.toml` in here is a dummy file, so it's better that it's just ignored, and if we do need to update it the undo command is in the `setup.sh` script.

djordon (approved):
Looks good. I was able to do `cargo build` in the `.generated-sources/emily` directory. At some point, we should add instructions on how to build this crate from scratch into the README (or someplace else). I needed `npm` and `java` installed in order to build emily and I used `asdf` to install `npm` and used `sdkman` to install `java`, maybe we can recommend one or several ways for others.

AshtonStephens:
> Looks good. I was able to do `cargo build` in the `.generated-sources/emily` directory. At some point, we should add instructions on how to build this crate from scratch into the README (or someplace else). I needed `npm` and `java` installed in order to build emily and I used `asdf` to install `npm` and used `sdkman` to install `java`, maybe we can recommend one or several ways for others.

Yes I was thinking the same thing, installation instructions / getting started instructions in the `README.md`. I'll make a ticket for this but I'd like to get #70 done first because that'll need another tool and it would be nice to do it in one go.

djordon (approved):

PR #101, opened 2024-05-01
Emily Rust Lambda Boilerplate

AshtonStephens:
## Description

Creates a Rust boilerplate lambda that responds to the requests from the `emily` api spec with responses of the right `501` error format. 

### Build Annoyances

The autogenerated client makes updates to the API definition trivial as you can see with this update, but I'm not convinced the autobuilding setup with `build.rs` that I included in #99 is the whole solution.

**How can we have the autogenerated client not cause `cargo build` to fail unless `cargo build --package emily` is run first?**

### Testing

- Unit tests


emily/lambda/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    #[test_case(Method::GET, "/deposits", None; "get-deposits")]
+    #[test_case(Method::PUT, "/deposits", None; "update-deposits")]
+    #[test_case(Method::POST, "/withdrawals", None; "create-withdrawal")]
+    #[test_case(Method::GET, "/withdrawals/{id}", None; "get-withdrawal")]
+    #[test_case(Method::GET, "/withdrawals", None; "get-withdrawals")]
+    #[test_case(Method::PUT, "/withdrawals", None; "update-withdrawals")]
+    #[test_case(Method::GET, "/chainstate/{height}", None; "get-chainstate")]
+    #[test_case(Method::PUT, "/chainstate/{height}", None; "set-chainstate")]
+    #[test_case(Method::POST, "/chainstate/{height}", None; "update-chainstate")]
+    #[tokio::test]
```
AshtonStephens, line 166: I'd rather have the `test_case` traits defined after `tokio::test` but unfortunately that's not possible. Not a big problem ~

- https://github.com/frondeus/test-case/pull/48

Cargo.toml
```diff
@@ -13,3 +14,16 @@ debug = 0
...
+tokio = "1.32.0"
+serde = "1.0"
+serde_json = "1.0"
+lambda_runtime = "0.11.1"
+aws_lambda_events = "0.15.0"
+http = "1.1.0"
+aws-sdk-dynamodb = { version = "1.3.0" }
+futures = "0.3.24"
+# This is necessary to compile the AWS Lambda as a lambda.
+openssl = { version = "0.10", features = ["vendored"] }
```
AshtonStephens, line 29: Should I be updating these here, or on a package basis?

emily/lambda/src/main.rs
```diff
@@ -0,0 +1,175 @@
...
+            deposit: "DEPOSIT_UNITTEST".to_string(),
+        }).unwrap();
+        let event = create_event(Method::POST, "/deposits", Some(json_body));
+
+        // Act.
+        let response = handle_event(event).await.expect("Failed to handle event");
+
+        // Assert.
+        assert_eq!(response.status_code, 201);
+        assert_eq!(response.body.unwrap(), Body::Text("{\"bitcoinTxid\":\"BITCOIN_TXID_UNITTEST\",\"bitcoinTxOutputIndex\":0.0,\"recipient\":\"MOCK_RECIPIENT\",\"amount\":11111.0,\"status\":\"PENDING\",\"statusMessage\":\"MOCK_CREATE_DEPOSIT_RESPONSE\",\"parameters\":{\"maxFee\":33333.0,\"lockTime\":22222.0,\"reclaimScript\":\"MOCK_RECLAIM_SCRIPT\"}}".to_string()));
```
AshtonStephens, line 149: I know this is bad, this is just boilerplate.

AshtonStephens: Maybe it's okay

Cargo.toml
djordon: I'm okay with either, but I think workspace root makes the most sense. 

As an aside, are you sure that openssl is necessary for AWS Lambda? As in, why wouldn't rustls work? I think this dependency is fine, since having it vendored makes potential compilation issues go away (I think), I just recall having issues with openssl when building with MUSL (might have been my fault, but still).

emily/lambda/src/main.rs
```diff
@@ -0,0 +1,175 @@
...
+            // the identifiers from the request.
+            let response = emily::models::CreateDepositResponseContent {
+                bitcoin_txid: request.bitcoin_txid.to_string(),
+                bitcoin_tx_output_index: request.bitcoin_tx_output_index,
+                recipient: "MOCK_RECIPIENT".to_string(),
+                amount: 11111.0,
+                status: emily::models::OpStatus::Pending,
+                status_message: "MOCK_CREATE_DEPOSIT_RESPONSE".to_string(),
+                parameters: Box::new(DepositParameters {
+                    lock_time: Some(22222.0),
```
djordon, line 47: Wow, that's a long time.

emily/lambda/src/main.rs
djordon: I think it's fine, especially so early in development.

emily/lambda/src/main.rs
djordon: Wait why is `amount`, `lock_time`, and `max_fee` not an integers?

djordon (approved):
As initial boilerplate, this looks fine. I did have a question about some of the generated models though.

emily/lambda/src/main.rs
```diff
@@ -0,0 +1,175 @@
+use aws_lambda_events::apigw::{ApiGatewayProxyRequest, ApiGatewayProxyResponse};
+use aws_lambda_events::encodings::Body;
+use emily::models::{CreateDepositRequestContent, DepositParameters};
+use http::{header, HeaderValue};
+use lambda_runtime::{service_fn, Error, LambdaEvent};
```
netrome, line 5: Nit: I prefer not importing ubiquitous names such as `Error` without any additional namespacing as I find it reduces readability. I'd rather refer to it as `lambda_runtime::Error` in the code, or type alias it as

```suggestion
use lambda_runtime::{service_fn, Error as LambdaRuntimeError, LambdaEvent};
```

emily/lambda/src/main.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    // Dispatch based on API call.
+    match (resource.as_str(), http_method) {
+        ("/deposits", Method::POST) => {
+
+            // Deserialize the create deposit request.
+            let request: CreateDepositRequestContent = body
+                .map(|serailzed_request| {
+                    serde_json::from_str::<CreateDepositRequestContent>(serailzed_request.as_str())
+                    .unwrap() // TODO: Handle errors.
```
netrome, line 33: Is this to be done in the current PR? If you want to do it as a follow-up I'd like to have the todo comment reference an issue as `// TODO(<issue_number>): Handle errors`. In general, I don't want any unlabeled todos to make it to the main branch.

emily/lambda/src/main.rs
```diff
@@ -0,0 +1,175 @@
...
+/// Main entry point for the AWS Lambda function.
+#[tokio::main]
+async fn main() -> Result<(), Error> {
+    // Run the lambda service.
+    lambda_runtime::run(
+        service_fn(
+            |event: LambdaEvent<ApiGatewayProxyRequest>| handle_event(event)
+        )
+    ).await
+}
```
netrome, line 114: Nit (code organization): I'd like this to be the first defined method. In general I find code to be more readable if functions are ordered after their occurrence with the most significant functions at the top, and utility functions at the bottom.

netrome (commented):
Great stuff. I'd just like the TODOs to be addressed or labeled before seeing this merged.

emily/lambda/src/main.rs
AshtonStephens: It's just what the autogenerator provided, but I am not sure why it does that, the template specifically says `Integer`.

Cargo.toml
AshtonStephens: This bit is predictive of the next PR, I can remove it for this one. It's not great practice to include things needed for the next PR.

This is what it's needed for:
- #103 

As part of that PR we need to run this command to compile the `.zip` file that a is unpacked and run in the lambda. 

```sh
cargo lambda build --release --output-format zip [--arm64]
```
The [Cargo Lambda](https://www.cargo-lambda.info/) command uses shared object files during compilation, and MacOS uses the sources in the directory marked by the environment variable `OPENSSL_DIR`. The problem is mac uses files like `libcrypto.3.dylib` instead of `libcrypto.3.so` which is what the compiler needs. 

After looking through a bunch of solutions this was the only thing that seemed to work across platforms. I also wonder if there was a better solution but this looks to be the standard one.

emily/lambda/src/main.rs
AshtonStephens: I didn't know you could do that!

emily/lambda/src/main.rs
AshtonStephens: Yes, that's generally fair. I'd like to leave it for now because I'm about to wipe the entire file, this is just to get something working and to remind myself that that part was dubious so when I rewrite the file I can handle those errors.

I'm 100% in on linking all TODOs to tickets, but in this case this code is going to have a shelf life of around 2 or so days and the GitHub is going to be come out of date nearly immediately.

AshtonStephens: Actually, I just created one of the implementation tickets #111 and linked the TODOS to it. It's not directly related but it's the ticket I'll address it in.

emily/lambda/src/main.rs
AshtonStephens: Done.

emily/lambda/src/main.rs
AshtonStephens: Done.

emily/lambda/src/main.rs
AshtonStephens: Fixed.

emily/lambda/src/main.rs
netrome: Nice, thanks :pray:

emily/lambda/src/main.rs
netrome: :pray:

netrome (approved):

emily/lambda/src/main.rs
netrome: I don't know how, but I bet this has to do with JavaScript using the same "number" type for anything numeric :sweat_smile:.

PR #103, opened 2024-05-01
Emily Lambda CDK Template

AshtonStephens:
## Description - Emily Lambda CDK Template

- Create CDK template for Emily Lambda
    - Unit tests
    - Build Scripts
- Updates root `README.md` with some tools to install locally (note some may be missing)

### Testing

I tested this by running `cdk synth` in my local workspace and also running `npm run test` within the cdk repository.

Depends on:
- #101


setup.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+# `.generated-sources/emily/Cargo.toml` changes when the build script runs.
+git update-index --assume-unchanged .generated-sources/emily/Cargo.toml
```
AshtonStephens, line 4: This shouldn't be here.

emily/cdk/lib/emily-stack.ts
```diff
@@ -0,0 +1,196 @@
...
+        const tableId: string = 'WithdrawalTable';
+        const table: dynamodb.Table = new dynamodb.Table(this, tableId, {
+            tableName: EmilyStackUtils.getResourceName(tableId, props),
+            partitionKey: {
+                name: 'RequestId',
+                type: dynamodb.AttributeType.STRING,
+            },
+            sortKey: {
+                name: 'StacksBlockHash',
+                type: dynamodb.AttributeType.BINARY,
```
djordon, line 86: The `RequestId` is _not_ globally unique, but just unique per stacks block hash? Also, why is the `StacksBlockHash` not a string? Not saying any of this is wrong it was just surprising so I figured I'd ask.

emily/cdk/lib/emily-stack.ts
```diff
@@ -0,0 +1,196 @@
...
+            ]
+        });
+
+        // TODO: Add an additional GSI for querying by user; not required for MVP.
+        return table;
+    }
+
+    createOrUpdateWithdrawalTable(props: EmilyStackProps): dynamodb.Table {
+        // Create DynamoDB table to store the messages. Encrypted by default.
+        const tableId: string = 'WithdrawalTable';
```
djordon, line 77: You know, my first thought was whether we could merge the deposit and withdrawal tables but I'm not sure that would be simpler. I'm pointing it out to see if you have any thoughts on the matter.

djordon: Yeah simpler to have two tables. Feel free to disregard.

emily/lambda/build-lambda.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# TODO: replace this with a more organic build script that
+# automatically triggers when building the cdk.
+if [[ "$(uname -m)" == "arm64" ]]; then
+  cargo lambda build --release --output-format zip --arm64
+else
+  cargo lambda build --release --output-format zip
```
djordon, line 8: Oh yeah I was just about to ask about why the default toolchain isn't used when building the a cargo lambda release, but then I saw the TODO 😄.

But we should include something in the README about installing `cargo-lambda` and the AWS CDK.

emily/cdk/lib/emily-stack.ts
```diff
@@ -0,0 +1,196 @@
...
+        serverLambda: lambda.Function,
+        props: EmilyStackProps
+    ): apig.SpecRestApi {
+
+        const restApiId: string  = "EmilyAPI";
+        const restApi: apig.SpecRestApi = new apig.SpecRestApi(this, restApiId, {
+            restApiName: EmilyStackUtils.getResourceName(restApiId, props),
+            apiDefinition: EmilyStackUtils.restApiDefinition(EmilyStackUtils.getPathFromProjectRoot(
+                ".generated-sources/openapi/openapi/Emily.openapi.json"
+            )),
```
djordon, line 182: Very cool.

Hmm, is there a way for us to ensure that this is built (or specifically rebuilt if it's outdated) before invoking the CDK?

djordon (commented):
Just did a quick look through, figured I'd submit even though it's a draft.

emily/cdk/lib/emily-stack.ts
AshtonStephens: I originally designed it to be one table but it lead to inefficient storage and queries so I changed it to be two. To query by type you'd either need to make multiple queries on the big table and remove the ones of the type you don't care about.

emily/cdk/lib/emily-stack.ts
AshtonStephens: Good callout! I'll have the npm build command trigger the building of the generated sources.

emily/cdk/lib/emily-stack.ts
AshtonStephens: The `RequestId` should be unique, but there's a brief moment on a fork when it isn't, so this avoids the collision without needing any extra forking related info that's unknown at creation.

AshtonStephens: RequestId should be a number though, oops

emily/lambda/build-lambda.sh
AshtonStephens: > I was just about to ask about why the default toolchain isn't used when building the a cargo lambda release

I actually have no idea how to use rust toolchains. Before I look it up, do you have a two sentence explanation on what I should be doing?

emily/cdk/lib/emily-stack.ts
AshtonStephens: Global secondary indices are a wrapper around **just duplicating the table**, so introducing one for each a deposit and a withdrawal leads to the same amount of storage being used for the GSIs as two separate tables would, plus the whole size of the original table.

The GSIs included here have a restricted set of attributes that they duplicate, so it's not the whole table.

emily/cdk/lib/emily-stack.ts
djordon: All of this is good to know, I'm glad I asked.

emily/cdk/lib/emily-stack.ts
djordon: > I originally designed it to be one table but it lead to inefficient storage and queries so I changed it to be two. To query by type you'd either need to make multiple queries on the big table and remove the ones of the type you don't care about.

I think you forgot the *or* part of the second sentence.

emily/lambda/build-lambda.sh
djordon: I don't think you should have to do anything at all to get _right_ build, so it's possible that I had the wrong model of what's going on here.

I think the disconnect for me was what the cargo-lambda folks were thinking here. Cargo will use the default toolchain when making a build unless you specify otherwise (like in the `rust-toolchan.toml` or in the command). Cargo-lambda is probably defaulting to build for whatever architecture is needed to run your Lambda on AWS. So maybe I'm just missing the memo.

emily/cdk/lib/emily-stack.ts
AshtonStephens: Done.

djordon (approved):
I had one question about the `tsc` cli but this look good. And I think this whole setup is pretty slick.

emily/api-definition/smithy-build.json
```diff
@@ -18,7 +18,7 @@
         "openapi": {
           "service": "stacks.sbtc#Emily",
           "protocol": "aws.protocols#restJson1",
-          "version": "3.1.0",
+          "version": "3.0.2",
```
AshtonStephens, line 21: AWS doesn't support `3.1.0`

emily/cdk/lib/emily-stack-utils.ts
```diff
@@ -0,0 +1,151 @@
...
+                                [ originalSubString, { [lambdaIdentifier] : lambdaFunction.functionArn } ]
+                        }
+                    })
+                }
+            })
+        })
+
+        // Return the modified template as an ApiDefinition.
+        return ApiDefinition.fromInline(apiJsonDefiniton)
+    }
```
AshtonStephens, line 150: I'm disappointed that this is the best option, and it relies on the openapi definition being well formatted, but we get some guarantees based on it being an OpenAPI spec. The errors that occur when the template is mangled are very obvious, which is what  bug here would result in.

Lets test everything... except not this please. If it crashes it's a design decision I swear

emily/cdk/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "emily",
+  "version": "0.1.0",
+  "bin": {
+    "emily": "bin/emily.js"
+  },
+  "scripts": {
+    "build": "(cd ../api-definition && npm run build-openapi) && tsc",
```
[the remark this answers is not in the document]
AshtonStephens: I'm actually not sure, but I think for now I'm going to assume I have all the build tools in there and then do a pass of all of this on a clean x86 linux machine to make sure we covered all the bases. I've been needing to fix my linux laptop for ages.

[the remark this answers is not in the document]
AshtonStephens: I know I said that last time, but in the next PR I'm going to be adding the local docker environment and that will then be the last of the dev tools necessary.

AshtonStephens:
I made a number of changes since the last review:
1. I changed the API specification to explicitly list the API operations that weren't directly tied to resources. I guess without them being listed explicitly they won't show up in the spec.
2. I made `build-lambda.sh` executable
3. I added an completely necessary absolutely rancid function called `restApiDefinitionWithLambdaIntegration` that will inject the information necessary for the API to autofill the lambda ARN.

emily/cdk/lib/emily-stack.ts
AshtonStephens: > you'd either need to make multiple queries on the big table and remove the ones of the type you don't care about **...or make a GSI on the table that lets you do the query in one shot but duplicates all the data, at which point you have two tables anyway.**

xoloki (approved):
Looks good!  Minor questions about dependencies and docker, but it sounds like they will all be resolved in the next PR so no need to block this one.

I'm also not at all familiar with Lambda, and my Typescript is basically nonexistent (though I did have the misfortune of needing to code in Javascript a decade or more ago).  But AFAICT it all looks legit.

Do you want me to actually try to build everything locally from the instructions?

PR #123, opened 2024-05-06
Prepend clippy ignore flags to emily client  file.

### LINKED ISSUE #122: [Feature]: Emily Client Pass Clippy Inspection
# Feature - Emily Client Pass Clippy Inspection

## 1. Description

The Emily client is autogenerated but doesn't pass `clippy` inspection. This ticket is to add some code to the `build.rs` file within the autogenerated code that will bypass only the style related clippy complaints.

### 1.1 Context & Purpose

The Emily client is autogenerated but doen't adhere to some of the subjective `clippy` compliants, making the whole workspace fail inspection. While we want clippy to tell us about any issues in the client, we shouldn't listen to any style related complaints because it's internally consistent and editing a client manually is madness; we should continue to have `clippy` analyze for correctness.

## 2. Technical Details:

Update the file at `.autogenerated-sources/emily/build.rs` to prepend the following flags to the autogeneraed `lib.rs`

```rust
#![allow(clippy::style)]
#![allow(clippy::too_many_arguments)]
```

### 2.1 Acceptance Criteria:
   - [ ] The `build.rs` file is changed to prepend the string only when those flags are missing.
   - [ ] It has been locally tested to work.

AshtonStephens:
## Description

Resolves #122 

Prepends the relevant clippy ignore flags to the `lib.rs` file after autogeneration.
- Checks if the full string to prepend is already there.
- Alters the file to include the prepended content.
- Fails with reasonable errors if something goes wrong.


djordon (commented):
Looks okay but is it possible to use `clippy.toml` here? It just seems like the more natural place to put this kind of stuff. I'm not sure it will work since I don't think there is a one-to-one mapping of all supported directives and configuration entries (also I'm not sure if placing it here will work the way we want).

AshtonStephens:
> Looks okay but is it possible to use `clippy.toml` here? It just seems like the more natural place to put this kind of stuff. I'm not sure it will work since I don't think there is a one-to-one mapping of all supported directives and configuration entries (also I'm not sure if placing it here will work the way we want).

I'll give it a shot, I'll admit that wasn't even on my radar as an option.

AshtonStephens:
Since there's instability in `clippy.toml` it'll be really hard to debug, in this method it'll be very easy to see what's going wrong, where, and how to fix it. If `clippy.toml` weren't unstable I'd go with that without question; these kinds of static configs are very dangerous when unstable because they have a lot of trust and create errors that are difficult to debug.

djordon:
> Since there's instability in `clippy.toml` it'll be really hard to debug, in this method it'll be very easy to see what's going wrong, where, and how to fix it. If `clippy.toml` weren't unstable I'd go with that without question; these kinds of static configs are very dangerous when unstable because they have a lot of trust and create errors that are difficult to debug.

Oh, I did not see that it was unstable. Well okay then.

djordon (approved):
Looks good.

A few notes. The "usual" outlets are less than ideal or unavailable.
1. `clippy.toml` is unstable and might be deprecated.
2. `Cargo.toml` clippy lint directives are unavailable since that file is auto-generated.

Issue #111, opened 2024-05-02
[Feature]: Emily POST API Calls


AshtonStephens:
# Feature - Emily POST API Calls

## 1. Description

Handle the POST API calls that update the API state. Use details from LLD [here](https://docs.google.com/document/d/13AN9GdLfPYKKPkeHNU4niq_CyJCDxFR7DGBLzLc5gxU/edit)

### 1.1 Context & Purpose
Emily has some Create API calls that use the POST http method. This ticket is to add the create functionality.

## 2. Technical Details:
Emily needs to be able to create the different artifacts on a POST call.

### 2.1 Acceptance Criteria:
   - [ ] CreateDeposit is done
   - [ ] CreateWithdrawal is done
   - [ ] CreateChainstate is done.

## 3. Related Issues and Pull Requests (optional):
- #101

AshtonStephens:
Removing because switching to utoipa changed the ticket breakdowns.

PR #124, opened 2024-05-06
72 deposit api mock outputs for api events

### LINKED ISSUE #72: Deposit API Mock outputs for API Events
## Deposit API Mock outputs for API Events

This is part of #50 

Mock event outputs from each deposit API function call. This should make the API useful during local testing.

AshtonStephens:
## Description

This PR adds mock outputs to the API function calls with nonsense outputs.

Resolve #72 

> **Note:** This PR looks huge, but break it up into these logical chunks when reviewing, there's a lot of boilerplate. The commit order is effectively the same ordering as below.

1. Changes to the API definition; you'll see these are minor typos / swapped variables.
2. `common.rs` - The common serialization / deserialization / packaging functions.
3. `errors.rs` - The error definitions that can convert to the API outputs expected.
4. `main.rs` + `utils/test.rs` - Look at the form of the code and how the tests work.
5. Read `operations/withdrawal.rs`, `operations/deposits.rs`, and `operations/chainstate.rs` which have a lot of similar code between them that I don't think should be separated into another function.

### HTTP Error Propagation

Instead of wrapping individual errors whenever there's a caught exception the Emily lambda defines the HTTP errors such that they can be packaged into the expected `ApiGatewayProxyRequest` directly from a member function `to_apigw_response(self) -> ApiGatewayProxyResponse`.

### Testing

I've foregone any testing within the endpoint files because the tests I originally wrote duplicated the coverage of the better more intuitive tests in the `main.rs` file. We can add better tests in those sections when we add the implementation details #111 and the other endpoints that modify state.

The tests are written to check each request that expects a body against 5 variations of a request body:

```rust
pub enum RequestType {
    FULL,
    MINIMAL,
    MALFORMED,
    MISSING,
    EMPTY,
}
```

The main test takes a function factory that creates a json string body based on the kind of body needed for the test and then runs the api handler against that body. The status code is all that's currently tested because the serialization is the only aspect that has the capacity to fail in the current code base.

### Directory Structure

I formatted the directories in the way that made sense to me, but this is my first Rust project; the decisions I've made here come from C / C++, not Rust idioms.


emily/api-definition/rust-config.json
```diff
@@ -3,5 +3,6 @@
   "packageVersion": "0.1.0",
   "library": "reqwest",
   "supportAsync": "true",
-  "withAWSV4Signature": "true"
+  "withAWSV4Signature": "true",
+  "useSingleRequestParameter": "true"
```
AshtonStephens, line 7: This makes the request packaged as a single struct as opposed to a flat function call to make the request. This is what lets us deserialize so easily.

emily/lambda/src/errors.rs
```diff
@@ -0,0 +1,166 @@
...
+
+#[derive(Debug)]
+pub enum EmilyApiError {
+    BadRequest(String),
+    // Forbidden(String), // Currently unused.
+    // NotFound(String), // Currently unused.
+    // Conflict(String), // Currently unused.
+    // NotImplemented(String), // Currently unused.
+    // Throttling(String), // Handled by the gateway, here for completeness.
+    Service(Result<String, Box<dyn std::error::Error>>),
```
djordon, line 15: I think we should a type other than `Result` here (or break up this variant into two).

emily/lambda/src/errors.rs
```diff
@@ -0,0 +1,166 @@
+use std::fmt;
+use aws_lambda_events::apigw::ApiGatewayProxyResponse;
+use aws_lambda_events::encodings::Body;
+use emily::models;
+use crate::common;
+
+#[derive(Debug)]
+pub enum EmilyApiError {
+    BadRequest(String),
```
djordon, line 9: You could also use `thiserror` here to generate an `std::error::Error` implementation. This would make this code look like this:
```rust
pub enum EmilyApiError {
    #[error("Bad Request: {0}")]
    BadRequest(String),
```
And then we could get rid of the manual `std::error::Error` and `fmt::Display` implementations.

emily/lambda/src/errors.rs
```diff
@@ -0,0 +1,166 @@
...
+                body: Some(Body::Text(body)),
+            },
+            // This occurs in the rare case that the API Error itself failed to serialize.
+            Err(err) => common::SimpleApiResponse {
+                status_code: http::StatusCode::INTERNAL_SERVER_ERROR.as_u16(),
+                // This output won't be of the right format for the client, but this is an edge case where
+                // the serialization will have needed to fail on a well defined structure, and this error
+                // can be handled by someone receiving data from the client since they'll need to handle
+                // potential errors anyway.
+                body: Some(Body::Text(format!("Error Deserializing: {}, {}", self, err).to_string())),
```
djordon, line 90: I would generate the body manually with this:
```rust
body: Some(Body::Text(serde_json::json!({"message": format!("SERVER ERROR {err}")}).to_string()))
```
Well, I'd break the above into multiple lines but it would equivalent the above.

djordon: Of course, this shouldn't ever be an issue, but still.

djordon (commented):
I did a quick run through, skipping a lot, but here is what I have so far. I'll circle back to the rest of it soon enough.

emily/lambda/src/errors.rs
AshtonStephens: The usage of `serde_json` is better, I'll change that.

emily/api-definition/models/withdrawals/structures.smithy
```diff
@@ -52,7 +52,7 @@ structure WithdrawalData {
 }
 
 list WithdrawalDataList {
-    member: DepositData
+    member: WithdrawalData
```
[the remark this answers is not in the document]
AshtonStephens: Me too, but I also knew we'd be testing them in these PRs so I wasn't too worried :) 

I'm 90% confident they're right now, and we'll see any others as the functionality makes its way in.

AshtonStephens:
Converting to draft while I restructure from feedback.

AshtonStephens:
Restructured the directories to be more idiomatic.

```
.
├── Cargo.toml
├── build-lambda.sh
└── src
    ├── bin
    │   └── entrypoint.rs
    ├── common.rs
    ├── errors.rs
    ├── eventhandler.rs
    ├── lib.rs
    ├── operations
    │   ├── chainstate.rs
    │   ├── deposits.rs
    │   └── withdrawals.rs
    ├── operations.rs
    ├── utils
    │   └── test.rs
    └── utils.rs
```

emily/lambda/src/errors.rs
```diff
@@ -0,0 +1,152 @@
...
+                body: Some(Body::Text(body)),
+            },
+            // This occurs in the rare case that the API Error itself failed to serialize.
+            Err(err) => common::SimpleApiResponse {
+                status_code: http::StatusCode::INTERNAL_SERVER_ERROR.as_u16(),
+                // This output won't be of the right format for the client, but this is an edge case where
+                // the serialization will have needed to fail on a well defined structure, and this error
+                // can be handled by someone receiving data from the client since they'll need to handle
+                // potential errors anyway.
+                body: Some(Body::Text(serde_json::json!({"message": format!("Server error {err} on {self}")}).to_string()))
```
djordon, line 95: Yeah I think last time I had `{err}` in the string, but I now think that this would be a slight mistake. We shouldn't display our internal error in the response.

emily/lambda/src/operations/deposits.rs
```diff
@@ -0,0 +1,179 @@
...
+            },
+            models::DepositData {
+                bitcoin_txid: "Chicken?".to_string(),
+                bitcoin_tx_output_index: 1.0,
+                recipient: "Churro".to_string(),
+                amount: 500000000.0,
+                last_update_height: Some(3.0),
+                last_update_block_hash: Some("aSmallDog".to_string()),
+                status: models::OpStatus::Accepted,
+                status_message: "Naruto Uzumaki will be Hokage".to_string(),
```
djordon, line 61:
```suggestion
                status_message: "Uzumaki, Naruto will be Hokage".to_string(),
```

emily/api-definition/rust-config.json
djordon: I'm not sure I follow, what is happening now and what was happening before? Just curious.

djordon (commented):
I didn't know you could define modules like this in rust 😄. Good to know.

The main thing that stood out was that our error responses shouldn't contain internal implementation details.

AshtonStephens:
> I'm not sure I follow, what is happening now and what was happening before? Just curious.

The original autogeneration made the function inputs a list of parameters, setting this to true makes the request function bodies appear as a single object which is great for automatic deserialization!

(Not sure why I can't respond to your comment)

emily/lambda/src/common.rs
```diff
@@ -0,0 +1,157 @@
...
+    ///
+    /// # Returns
+    /// - `ApiGatewayProxyResponse`: The API Gateway compatible response.
+    #[allow(clippy::wrong_self_convention)] // Allow `to_apigw_response` to consume `self`.
+    pub fn to_apigw_response(self) -> aws_lambda_events::apigw::ApiGatewayProxyResponse {
+        let mut headers = http::HeaderMap::new();
+        headers.insert(http::header::CONTENT_TYPE, http::HeaderValue::from_static("application/json"));
+        headers.insert(http::header::ACCESS_CONTROL_ALLOW_HEADERS, http::HeaderValue::from_static("Content-Type"));
+        headers.insert(http::header::ACCESS_CONTROL_ALLOW_ORIGIN, http::HeaderValue::from_static("*"));
+        headers.insert(http::header::ACCESS_CONTROL_ALLOW_METHODS, http::HeaderValue::from_static("OPTIONS,POST,GET"));
```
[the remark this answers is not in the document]
AshtonStephens: This is one of those things that made errors go away when I hooked in the last API I made to a website. But yes, API Gateway handles the OPTION method and provides the correct CORS response automatically when you tell it to.

emily/lambda/src/operations/deposits.rs
AshtonStephens: Done.

emily/lambda/src/operations.rs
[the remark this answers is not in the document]
AshtonStephens: I've got **zero** opinions on how to structure this, I'm entirely at your whims. But, I don't expect this lambda to become a huge project, so the structure is slightly less important.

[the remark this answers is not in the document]
djordon: Yeah what you have here is fine, I was just surprised. I originally thought it was wrong, but then realized that it compiled (obviously) and was indeed a possibility for quite some time.

djordon (approved):
Looks good. 🚀

djordon:
> I have another place where I propagate the exception message, I think in the short term I'd like to keep that because we've got around two months from now till testnet and we're the only ones who will see this for a _while_.
> 
> I'll make a GH ticket to remind us to remove this before launching anything.

Okay, no problem.

emily/lambda/src/common.rs
```diff
@@ -0,0 +1,157 @@
...
+pub fn deserialize_request<T: serde::de::DeserializeOwned>(
+    body:  Option<String>
+) -> Result<T, errors::EmilyApiError> {
+    body
+        .ok_or_else(|| errors::EmilyApiError::BadRequest("No body present".to_string())) // should be client
+        // Deserialize body.
+        .and_then(|body_string|
+            serde_json::from_str::<T>(&body_string)
+                .map_err(|e| {
+                    errors::EmilyApiError::BadRequest(format!("{} but instead got \"{}\"", e, body_string).to_string())
```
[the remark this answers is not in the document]
AshtonStephens: I actually think we should consider having this in the final version, that way if someone mis-configured their client they at least have a path to resolving the issue.

AshtonStephens:
Squashing and merging because I didn't do these in order the first time and don't want to spend time reorganizing it.

PR #125, opened 2024-05-06
Use environment variables for config management

### LINKED ISSUE #121: Use Environment Variables for Configuration Management
Update Blocklist Client setup to use environment variables for config management as a preferred way to adhere to the twelve-factor-app philosophy instead of using a config file and passing that instance around.

8marz8:
resolves #121 

Using a static read-only config that gets instantiated via a default config file and environment variables for an improved way to handle config management.

### DIFF
--- a/blocklist-client/src/config.rs
+++ b/blocklist-client/src/config.rs
@@ -24,6 +24,8 @@ pub static SETTINGS: Lazy<Settings> =
     Lazy::new(|| Settings::new().expect("Failed to load configuration"));
 
 impl Settings {
+    // Initializing the global config first with default values and then with provided/overwritten environment variables.
+    // The explicit separator with double underscores is needed to correctly parse the nested config structure.
     pub fn new() -> Result<Self, ConfigError> {
         let mut cfg = Config::new();
         cfg.merge(File::with_name("./src/config/default"))?;
--- a/blocklist-client/src/config.rs
+++ b/blocklist-client/src/config.rs
@@ -1,19 +1,35 @@
+use config::{Config, ConfigError, Environment, File};
+use once_cell::sync::Lazy;
 use serde::Deserialize;
 
-#[derive(Deserialize, Clone)]
+#[derive(Deserialize, Clone, Debug)]
 pub struct Settings {
     pub server: ServerConfig,
     pub risk_analysis: RiskAnalysisConfig,
 }
 
-#[derive(Deserialize, Clone)]
+#[derive(Deserialize, Clone, Debug)]
 pub struct ServerConfig {
     pub host: String,
     pub port: u16,
 }
 
-#[derive(Deserialize, Clone)]
+#[derive(Deserialize, Clone, Debug)]
 pub struct RiskAnalysisConfig {
     pub api_url: String,
     pub api_key: String,
+}
+
+pub static SETTINGS: Lazy<Settings> =
+    Lazy::new(|| Settings::new().expect("Failed to load configuration"));
+
+impl Settings {
+    pub fn new() -> Result<Self, ConfigError> {
+        let mut cfg = Config::new();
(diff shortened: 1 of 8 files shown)


blocklist-client/src/config.rs
```diff
@@ -1,19 +1,35 @@
...
 }
+
+pub static SETTINGS: Lazy<Settings> =
+    Lazy::new(|| Settings::new().expect("Failed to load configuration"));
+
+impl Settings {
+    pub fn new() -> Result<Self, ConfigError> {
+        let mut cfg = Config::new();
+        cfg.merge(File::with_name("./src/config/default"))?;
+        let env = Environment::with_prefix("APP").separator("__");
```
netrome, line 30: Why the double underscore? Also, should we consider using a more specific prefix than "APP"? 
```suggestion
        let env = Environment::with_prefix("BLOCKLIST_CLIENT").separator("_");
```

blocklist-client/src/config/default.toml
```diff
@@ -0,0 +1,7 @@
+[server]
+host = "127.0.0.1"
+port = 3030
+
+[risk_analysis]
+api_url = "http://example.com/api"
+api_key = "api-key-here"
```
netrome, line 7: Should we provide defaults for these? I'd rather have the app fail on startup if these aren't set instead of starting it with misconfigured API parameters.

netrome (approved):
Nice stuff. Approved, but I have some enhancement suggestions that I think would make this even better.

Primarily, I would love to see documentation on the environment variables. A list with the name of the variable and a short description in a `README.md` file would be sufficient.

Otherwise, just a few minor suggestions in the code.

blocklist-client/src/config/default.toml
8marz8: That makes sense, I put them for a reference/easier placeholder for running it till I add the README but I'll add the README and remove those dummy defaults

blocklist-client/src/config.rs
8marz8: Sure, I'll update the prefix.

As for the separator, the nested configs wouldn't get picked up with a single underscore - [Related thread](https://github.com/mehcode/config-rs/issues/104#issuecomment-490971742)
I'll poke around with it a bit more but I run it like this:
` APP_server__host=127.0.0.1 APP_server__port=8080 APP_risk_analysis__api_url=https://chainalysis-api.com/ APP_risk_analysis__api_key=your_api_key cargo run`

netrome: Okay, interesting. Are the env variables case sensitive? I'd expect it to be `APP_SERVER__HOST` and not `APP_server__host`.

8marz8: Specified it correctly in the README.md now.

blocklist-client/src/config.rs
```diff
@@ -1,19 +1,35 @@
...
 }
+
+pub static SETTINGS: Lazy<Settings> =
+    Lazy::new(|| Settings::new().expect("Failed to load configuration"));
+
+impl Settings {
+    pub fn new() -> Result<Self, ConfigError> {
+        let mut cfg = Config::new();
+        cfg.merge(File::with_name("./src/config/default"))?;
+        let env = Environment::with_prefix("BLOCKLIST_CLIENT").separator("__");
```
AshtonStephens, line 30: What does this do?

AshtonStephens (approved):
LGTM

blocklist-client/src/config.rs
8marz8: Initializing the global config first with default values and then the ones provided/overwritten by the environment variables. The explicit separator with double underscore was needed to correctly parse the nested config structure. 
(Will add this as a comment)

Issue #75, opened 2024-04-26
Properly construct the spend script in BTC transactions


djordon:
This is part of https://github.com/Trust-Machines/sbtc-v1/issues/52.

The task of constructing a BTC transaction package from requests is broken up into three groups of tasks: 
1. Constructing optimal approval sets, which will be done in
    * https://github.com/Trust-Machines/sbtc-v1/issues/63 and
    * https://github.com/Trust-Machines/sbtc-v1/issues/65.
2. Converting an approval set into a proper BTC transaction package with a valid spend script which is done partly in this ticket
3. Making sure we modify the BTC transaction when new requests come in or when fees change
    * https://github.com/Trust-Machines/sbtc-v1/issues/66.

Issue #67, opened 2024-04-24
Add integration tests for our transaction package


djordon:
This is part of https://github.com/Trust-Machines/sbtc-v1/issues/52.

Add tests that check the basic properties of the transaction package. This will require a bitcoin node being set up during CI.
- [x] (https://github.com/stacks-network/sbtc/pull/146) That the transaction is accepted by the network.
- [ ] Test that only the private key associated with the public key can spend the withdrawal funds.
- [x] (https://github.com/stacks-network/sbtc/pull/146) Test deposits increase the amount of the signers' UTXO by the appropriate amount.
- [ ] That withdrawals reduce the amount of the signers' UTXO by the appropriate amount. 
- [ ] Test that only the signers can spend the signers' UTXO.

Issue #66, opened 2024-04-24
Add functionality to transaction package construction to support RBF


djordon:
This is part of https://github.com/Trust-Machines/sbtc-v1/issues/52.

We need to update our unconfirmed BTC transactions if the market fee rate changes so that pending deposits and withdrawals do not get stuck. This will be done by submitting RBF transactions. Add in functionality that properly constructs an RBF transaction when constructing the next transaction package and we know that we have already submitted such a transaction into our database.

Issue #30, opened 2024-04-01
[Design]: sBTC Deposit UTXO Binary Format


AshtonStephens:
***Completing the issue description and arriving at a conclusion is the deliverable of this issue.***

# Design - sBTC Deposit UTXO Binary Format
This ticket holds the design of the BTC Deposit UTXO Binary Format and how it fits into sBTC-v1.

## 1. Summary
To deposit funds into sBTC according to the deposit flow defined in #15, a Bitcoin holder creates a deposit request. A deposit request in sBTC is a Bitcoin transaction which contains a deposit UTXO. This transaction should be accompanied by a deposit proof sent to Signers, to allow the signers to consume the UTXO. The exact format of the deposit UTXO and the content of the deposit proof is defined in this ticket.

## 2. Context & Purpose
The purpose of defining a deposit UTXO format is to allow any sBTC client to create valid deposit requests, and any sBTC Signer implementation to properly implement deposit UTXO consumption which is a vital part of the current deposit flow.

To support the deposit flow, the deposit UTXO format must define a type of UTXO which can be:
1. Consumed by Signers.
    * The Signers are collectively capable of producing [BIP-340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki)-compliant Schnorr signatures.
    * When the UTXO consumed, it should reveal the sBTC recipient address.
2. If the deposit is rejected, can be reclaimed by the Bitcoin holder.

***Relevant Research Discussions***
Deposit flow https://github.com/Trust-Machines/sbtc-v1/issues/15

***External Resources***

### 3.1 Proposed Component Design
The deposit UTXO is a SegWit version 1 output ([BIP-341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki)). This means that the `scriptPubKey` is a 33 byte long sequence with a leading version byte and a following witness program as illustrated below.

Deposit UTXO `scriptPubKey`:
```
0              1                 33
| version_byte | witness_program |
```

To fulfill the requirements on the deposit UTXO, the witness program must have two general conditions and no public key spend condition. The general conditions should be

1. Accept condition. A condition allowing Signers to consume the UTXO with their aggregate key and while doing so revealing a `sbtc_payload` from the user on the chain.
2. Reclaim condition. A condition allowing the BTC holder to reclaim the UTXO if the deposit is rejected.

#### Witness program
[...]

PR #128, opened 2024-05-06
feat: construct BTC transactions

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/75.

We need to take a set of deposit and withdrawal requests and turn them into BTC transactions that can be signed by the signers. This PR gets us part of the way there.

## Type of Change

This is new code that takes the current state of the signers UTXO, the current market rate in sats-per-vbyte, along with outstanding deposit and withdrawal requests and creates a set of BTC transactions.

### Notes
There are several things worth noting though:
1. The UTXO output scripts for withdrawals are pay-to-witness-public-key-hash (P2WPKH) outputs. I think this is fine, since it's probably okay to a assume our users have native segwit addresses but I don't know about assuming the same for taproot addresses. In the future, maybe the user can decide?
2. I have not implemented filtering of requests based on their max fee and the current market fee rate.
3. There is nothing here that accommodates things related to replace-by-fee (RBF) transactions. That stuff will get done in another PR addressing https://github.com/stacks-network/sbtc/issues/66.
4. I'm not sure what the final API here will be, the public stuff here is not close to final.

No breaking changes

## Testing information

There are unit tests here that test for most of the asserted functionality and/or invariants. But some further integration testing is required to ensure that properly signed witness data is indeed correct. These tests will happen in another PR (such as one addressing https://github.com/stacks-network/sbtc/issues/67).

## Checklist
- [x] Tag reviewers

### DIFF
--- a/signer/src/utxo.rs
+++ b/signer/src/utxo.rs
@@ -6,6 +6,7 @@ use bitcoin::taproot::NodeInfo;
 use bitcoin::taproot::Signature;
 use bitcoin::taproot::TaprootSpendInfo;
 use bitcoin::transaction::Version;
+use bitcoin::Address;
 use bitcoin::Amount;
 use bitcoin::OutPoint;
 use bitcoin::PublicKey;
@@ -172,8 +173,8 @@ pub struct WithdrawalRequest {
     pub amount: u64,
     /// The max fee amount to use for the sBTC deposit transaction.
     pub max_fee: u64,
-    /// The public key that will be able to spend the output.
-    pub public_key: PublicKey,
+    /// The address to spend the output.
+    pub address: Address,
     /// How each of the signers voted for the transaction.
     pub signer_bitmap: Vec<bool>,
 }
@@ -184,13 +185,11 @@ impl WithdrawalRequest {
         self.signer_bitmap.iter().map(|vote| !vote as u32).sum()
     }
 
-    /// Withdrawal UTXOs are Pay-to-Witness Public Key Hash (P2WPKH)
+    /// Withdrawal UTXOs pay to the given address
     fn as_tx_output(&self) -> TxOut {
-        let compressed = bitcoin::CompressedPublicKey(self.public_key.inner);
-        let pubkey_hash = compressed.wpubkey_hash();
         TxOut {
             value: Amount::from_sat(self.amount),
-            script_pubkey: ScriptBuf::new_p2wpkh(&pubkey_hash),
+            script_pubkey: self.address.script_pubkey(),
         }
     }
 }
@@ -462,6 +461,8 @@ mod tests {
 
     use super::*;
     use bitcoin::blockdata::opcodes;
+    use bitcoin::CompressedPublicKey;
+    use bitcoin::KnownHrp;
     use bitcoin::NetworkKind;
     use bitcoin::PrivateKey;
     use bitcoin::Txid;
@@ -477,6 +478,13 @@ mod tests {
     fn generate_public_key() -> PublicKey {
         let private_key = PrivateKey::generate(NetworkKind::Test);
         private_key.public_key(SECP256K1)
+    }
+
+    fn generate_address() -> Address {
+        let public_key = generate_public_key();
+        let pk = CompressedPublicKey(public_key.inner);
+
+        Address::p2wpkh(&pk, KnownHrp::Regtest)
     }
 
     fn generate_outpoint(amount: u64, vout: u32) -> OutPoint {
@@ -525,13 +533,13 @@ mod tests {
         }
     }
 
-    /// Create a new withdrawal request withdrawing to a random key.
+    /// Create a new withdrawal request withdrawing to a random address.
     fn create_withdrawal(amount: u64, max_fee: u64, votes_against: usize) -> WithdrawalRequest {
         WithdrawalRequest {
             max_fee,
             signer_bitmap: std::iter::repeat(false).take(votes_against).collect(),
             amount,
-            public_key: generate_public_key(),
+            address: generate_address(),
         }
     }
 
@@ -804,10 +812,7 @@ mod tests {
         let mut output_scripts: BTreeSet<String> = requests
             .withdrawals
             .iter()
-            .map(|req| {
-                let compressed = bitcoin::CompressedPublicKey(req.public_key.inner);
-                ScriptBuf::new_p2wpkh(&compressed.wpubkey_hash()).to_hex_string()
-            })
+            .map(|req| req.address.script_pubkey().to_hex_string())
             .collect();
 
         // Now we check that the counts of the withdrawals and deposits
@@ -889,11 +894,7 @@ mod tests {
         let mut withdrawal_amounts: BTreeMap<String, u64> = requests
             .withdrawals
             .iter()
-            .map(|req| {
-                let compressed = bitcoin::CompressedPublicKey(req.public_key.inner);
-                let script = ScriptBuf::new_p2wpkh(&compressed.wpubkey_hash());
-                (script.to_hex_string(), req.amount)
-            })
+            .map(|req| (req.address.script_pubkey().to_hex_string(), req.amount))
             .collect();
 
         let transactions = requests.construct_transactions().unwrap();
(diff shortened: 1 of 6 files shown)


signer/src/utxo.rs
```diff
@@ -0,0 +1,964 @@
...
+    /// How each of the signers voted for the transaction.
+    pub signer_bitmap: Vec<bool>,
+    /// The amount of sats in the deposit UTXO.
+    pub amount: u64,
+    /// The deposit script used so that the signers' can spend funds.
+    pub deposit_script: ScriptBuf,
+    /// The redeem script for the deposit.
+    pub redeem_script: ScriptBuf,
+    /// The public key used for the key-spend path of the taproot script.
+    pub taproot_public_key: PublicKey,
```
netrome, line 95: A deposit request doesn't have a key-spend path (as specified in the [design issue](https://github.com/stacks-network/sbtc/issues/30)). We don't want to allow anyone to spend the deposit request while the signers are processing it. Instead, the key-spend should be the NUMS point from [BIP-341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki) $H = lift_x(0x50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0)$

Therefore, I don't think this should be tracked in the `DepositRequest` struct, but rather be a hard-coded constant.

signer/src/utxo.rs
```diff
@@ -0,0 +1,964 @@
...
+    /// The public key used in the deposit script.
+    pub deposit_public_key: PublicKey,
```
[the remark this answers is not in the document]
netrome: Right, so this should be the aggregate key. Can we rename it?
```suggestion
    /// The public key used in the deposit script.
    pub signer_aggregate_key: PublicKey,
```

signer/src/utxo.rs
netrome: Right, I see you've already documented this below that it should be a dummy point. I assume you then expect the validation of this point to happen before it's used in the `construct_witness_data` function then?

signer/src/utxo.rs
```diff
@@ -0,0 +1,964 @@
...
+        }
+        // We now update the remaining withdrawal amounts to account for fees.
+        tx.output.iter_mut().skip(1).for_each(|tx_out| {
+            tx_out.value = Amount::from_sat(tx_out.value.to_sat().saturating_sub(fee));
+        });
+    }
+
+    /// Helper function for generating dummy Schnorr signatures.
+    fn generate_dummy_signature() -> Signature {
+        let key_pair = Keypair::new_global(&mut secp256k1::rand::thread_rng());
```
netrome, line 434: Nit: We might as well include `rand` as an explicit dependency since I'm sure we'll use it more times than just here.

netrome: Also, we've defaulted to using OsRng in many places in the nakamoto signer already. I don't have strong opinions on either, but it's nice to be consistent.

netrome (changes requested):
Great stuff so far. The only blocker I see is that we shouldn't dictate the output type for withdrawals.

signer/src/utxo.rs
```diff
@@ -0,0 +1,964 @@
...
+}
+
+#[derive(Debug)]
+pub struct WithdrawalRequest {
+    /// The amount of BTC, in sats, to withdraw.
+    pub amount: u64,
+    /// The max fee amount to use for the sBTC deposit transaction.
+    pub max_fee: u64,
+    /// The public key that will be able to spend the output.
+    pub public_key: PublicKey,
```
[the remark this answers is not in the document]
djordon: That all makes sense to me, but I think that would lead us being off the low-level designs in the [Deposit API LLD](https://docs.google.com/document/d/13AN9GdLfPYKKPkeHNU4niq_CyJCDxFR7DGBLzLc5gxU/edit#heading=h.vwfrlp5xn7lz) and the [Clarity LLD](https://docs.google.com/document/d/1Bwi1jM_7p9OqNGT41eWCuoKnfoOrNHycqxby3XSonCA/edit#heading=h.y6v41emwi920). I was following the Deposit API LLD, which uses the public key, while the clarity LLD uses a tuple of a 1 byte `version` field with a 32 byte `hashbytes` field (which I think is a compressed public key).

I'm okay changing the designs and it's all equally easy for me to update here. Just want to call out that this would lead to a mismatch.

signer/src/utxo.rs
```diff
@@ -0,0 +1,964 @@
...
+    /// Accepted and pending deposit requests.
+    pub deposits: Vec<DepositRequest>,
+    /// Accepted and pending withdrawal requests.
+    pub withdrawals: Vec<WithdrawalRequest>,
+    /// Summary of the Signers' UTXO and information necessary for
+    /// constructing their next UTXO.
+    pub signer_state: SignerBtcState,
+    /// The maximum acceptable number of votes against for any given
+    /// request.
+    pub reject_capacity: u32,
```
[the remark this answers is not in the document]
djordon: Yeah, it is more intuitive. I think I had that design originally but with an additional field noting the total number of signers so that we could compute the reject capacity (the accept threshold alone is not enough). I'm okay with switching that version of things with the additional field.

signer/src/utxo.rs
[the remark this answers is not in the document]
djordon: That makes sense. Are you opposed to `signers_public_key`?

signer/src/utxo.rs
djordon: > Right, I see you've already documented this below that it should be a dummy point. I assume you then expect the validation of this point to happen before it's used in the `construct_witness_data` function then?

Sorta. I've been assuming both the signers and the Deposit API would do this validation. But it would all happen using the Transaction Analysis library when constructing the `DepositRequest` struct here.

signer/src/utxo.rs
djordon: Cool beans, will update.

signer/src/utxo.rs
djordon: > Therefore, I don't think this should be tracked in the `DepositRequest` struct, but rather be a hard-coded constant.

Okay, I'll hard code it in the `construct_witness_data` function.

netrome: So after a quick sync we've agreed to do this as a follow-up instead because:

1. It's not trivial
2. It's not obvious if the validation of the NUMS point should be the concern of this particular module.

djordon: I created https://github.com/stacks-network/sbtc/issues/130 for this work.

djordon:
Updated! @netrome would you mind taking another look?

signer/src/utxo.rs
```diff
@@ -0,0 +1,976 @@
...
+            return Ok(Vec::new());
+        }
+
+        let withdrawals = self.withdrawals.iter().map(Request::Withdrawal);
+        let deposits = self.deposits.iter().map(Request::Deposit);
+
+        // Create a list of requests where each request can be approved on its own.
+        let items = deposits.chain(withdrawals);
+
+        compute_optimal_packages(items, self.reject_capacity())
```
hstove, line 70: I'm a bit confused as to where input/output weight come into play here. In general, inputs will be very similar size (though not exact), and outputs will be varying in size (depending on the type of address). So, wouldn't this function need to have some mechanism of knowing a request's "cost"?

hstove (commented):
In general this is pretty great, but I have a question around computing the optimal tx. If what I'm talking about is known and out of scope for this PR, that totally makes sense too!

signer/src/utxo.rs
djordon: Good question. The `compute_optimal_packages` function computes approval sets using only how the signers have voted on each of the requests. The number of rejections is the request's "weight", it is completely separate from input/output weight of a BTC transaction. That function doesn't care about the BTC weight at all.

You know, I should rename that term but I couldn't think of anything better (size? heaviness? pounds? lbs?). Everything I thought of kinda missed the mark a little but maybe weight is even worse.

netrome:
> Updated! @netrome would you mind taking another look?

Thanks, will do :eyes:

signer/src/utxo.rs
netrome: Perfect, thanks!

netrome (approved):
Thanks for updating, looks good to me!

PR #129, opened 2024-05-07
feat: Signer message signing and validation

### LINKED ISSUE #105: [Feature]: Message signing and validation
# Feature - Message signing and validation

## 1. Description
Implement logic to sign and validate signer messages.

### 2.1 Acceptance Criteria:
   - [ ] The SignerMessage type exposes a method to construct and attach a secp256k1 ECDSA signature to the message.
   - [ ] The SignerMessage type exposes a method to validate its signature.

netrome:
closes #105 

## Description
This PR introduces signing and validation logic for signer messages. As part of this change, the `signer_public_key` and `signature` fields of the `SignerMessage` type has been moved to the new `Signed` wrapper type to allow for ergonomic handling of signatures.

### DIFF
--- a/signer/src/ecdsa.rs
+++ b/signer/src/ecdsa.rs
@@ -1,11 +1,74 @@
+//! # ECDSA Signing and Verification
+//!
+//! This module provides functionality to sign and verify data using the ECDSA signature scheme.
+//! The core component is the `Signed<T>` structure, which wraps any signable data along with the signer's public key
+//! and a signature. The module also defines the `SignECDSA` trait to facilitate signing operations.
+//!
+//! ## Features
+//!
+//! - **Signed**: A generic wrapper that binds a piece of data (`inner`) with its ECDSA signature and the public key
+//!   of the signer. It provides a method to verify the integrity and authenticity of the data.
+//! - **SignECDSA**: A trait implemented by data types that can be signed using ECDSA. It is automatically implemented for any type implementing `wsts::net::Signable`.
+//!
+//! ## Examples
+//!
+//! ### Signing and Verifying a String
+//!
+//! The following example demonstrates how to sign a simple string and then verify its signature:
+//!
+//! ```
+//! use sha2::Digest;
+//! use sbtc_signer::ecdsa::SignECDSA;
+//!
+//! struct SignableStr(&'static str);
+//!
+//! // Implementing `wsts::net::Signable` unlocks the signing functionality in this module.
+//! impl wsts::net::Signable for SignableStr {
+//!     fn hash(&self, hasher: &mut sha2::Sha256) {
+//!         hasher.update(self.0)
+//!     }
+//! }
+//!
+//! let msg = SignableStr("Sign me please!");
+//! let private_key = p256k1::scalar::Scalar::from(1337);
+//!
+//! // Sign the message.
+//! let signed_msg = msg
+//!     .sign_ecdsa(&private_key)
+//!     .expect("Failed to sign message");
+//!
+//! // Verify the signed message.
+//! assert!(signed_msg.verify());
+
 use p256k1::ecdsa;
 use p256k1::scalar::Scalar;
 
+/// Wraps an inner type with a public key and a signature,
+/// allowing easy verification of the integrity of the inner data.
+#[derive(Debug, Clone)]
+pub struct Signed<T> {
+    /// The signed structure.
+    pub inner: T,
+    /// The public key of the signer.
+    pub signer_pub_key: ecdsa::PublicKey,
+    /// A signature over the hash of the inner structure.
+    pub signature: Vec<u8>,
+}
+
+impl<T: wsts::net::Signable> Signed<T> {
+    /// Verify the signature over the inner data.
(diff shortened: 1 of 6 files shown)


signer/src/ecdsa.rs
```diff
@@ -0,0 +1,183 @@
+//! # ECDSA Signing and Verification
```
AshtonStephens, line 1: Did you write this manually? How do we feel about using AI for documentation boilerplate?

netrome: I co-authored this with ChatGPT, using my normal AI-collaboration-workflow. 1. Ask the AI to do something for me, 2. Rewrite most of it because the AI doesn't write what I want it to 😅

hstove: What is `//!`? I'm not familiar with it, and couldn't find anything with a quick google search. Is it for "top-level" module comments?

djordon: > What is `//!`? I'm not familiar with it, and couldn't find anything with a quick google search. Is it for "top-level" module comments?

Yeah, here it's a module [doc comment](https://doc.rust-lang.org/reference/comments.html#doc-comments). In general it's a special kind of doc comment that applies to what comes before the comment (rather than what comes after).

signer/src/ecdsa.rs
```diff
@@ -0,0 +1,183 @@
...
+}
+
+impl<T> std::ops::DerefMut for Signed<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.inner
+    }
+}
+
+/// Helper trait to provide the ability to construct a `Signed<T>`.
+pub trait SignECDSA: Sized {
```
djordon, line 80: Nit: I think this should be `SignEcdsa`.

signer/src/ecdsa.rs
```diff
@@ -0,0 +1,183 @@
...
+    fn sign_ecdsa(self, private_key: &Scalar) -> Result<Signed<Self>, Error>;
+}
+
+impl<T: wsts::net::Signable> SignECDSA for T {
+    /// Create a `Signed<T>` instance with a signature and public key constructed from `private_key`.
+    fn sign_ecdsa(self, private_key: &Scalar) -> Result<Signed<Self>, Error> {
+        let signer_pub_key = ecdsa::PublicKey::new(private_key)?;
```
djordon, line 87: Ohhh I see, I think. Whenever we send a message _during the WSTS signing process_ we need to sign that message. Those signatures are standard ECDSA signatures. After all of the messages have been sent during the WSTS process, the coordinator will have enough information to construct a Schnorr signature.

So this trait here helps sign the messages during WSTS processes. Is that right?

djordon (approved):
Had a small style nit and a question, but it looks good.

signer/src/ecdsa.rs
netrome: Yes exactly. This is just to ensure message integrity in the signer communication, and is not to be confused with the aggregate signature that the signers produce.

signer/src/ecdsa.rs
netrome: Hahah I actually went back and forth a bit with this. Always unsure of how to properly handle abbreviations (I think we have both StackerDb and SignerDB or the opposite in the node code 😅). Thanks for expressing a clear opinion, I'll change this.

signer/src/ecdsa.rs
netrome: Thanks for chiming in @djordon :pray:

signer/src/message.rs
```diff
@@ -41,3 +46,67 @@ pub struct BitcoinTransactionSignRequest;
...
+        wsts::net::Message::DkgPublicShares(msg) => msg.hash(hasher),
+        wsts::net::Message::DkgPrivateBegin(msg) => msg.hash(hasher),
+        wsts::net::Message::DkgPrivateShares(msg) => msg.hash(hasher),
+        wsts::net::Message::DkgEndBegin(msg) => msg.hash(hasher),
+        wsts::net::Message::DkgEnd(msg) => msg.hash(hasher),
+        wsts::net::Message::NonceRequest(msg) => msg.hash(hasher),
+        wsts::net::Message::NonceResponse(msg) => msg.hash(hasher),
+        wsts::net::Message::SignatureShareRequest(msg) => msg.hash(hasher),
+        wsts::net::Message::SignatureShareResponse(msg) => msg.hash(hasher),
+    }
```
AshtonStephens, line 82: Is this because the `hash` function is different in all of them from the compiler's perspective?

AshtonStephens (approved):
LGTM!

signer/src/message.rs
AshtonStephens: Because the inner class is different for each of them

netrome: This is to dispatch the `.hash` call to all variants. In the next version `wsts` will implement `Signable` directly on the `Message` type so we don't have to do this here.

I know this looks quite repetitive, and you can get around this type of code by using for example the `enum_dispatch` crate: https://crates.io/crates/enum_dispatch.

PR #132, opened 2024-05-07
Setup cdk unit tests to be independent of lambda compilation.

AshtonStephens:
## Description

The Emily CDK uses a zipped lambda file for the AWS lambda, but if this file isn't present during compile time the cdk template will fail to build. This PR has the template use a dummy zip file when the stack's stage name indicates that it's being compiled for unit tests.

This PR does the following:
1. Adds a dummy lambda zip file `emily/cdk/test/assets/empty-lambda.zip` 
2. Creates a `constants.ts` file that defines the unit test stagename so it's consistent across files.
3. Changes the template to use the dummy lambda when being compiled for unit tests
4. Changes the unit tests to expect the new stagename.
5. Optimizes imports. 
6. Changes the `cdk` command to use the local package

Relates to a reported issue that is too small to file a ticket for.

### DIFF
--- a/emily/cdk/bin/emily.ts
+++ b/emily/cdk/bin/emily.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
-import 'source-map-support/register';
 import * as cdk from 'aws-cdk-lib';
+import 'source-map-support/register';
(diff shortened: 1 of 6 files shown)


hstove (approved):
LGTM!

PR #133, opened 2024-05-08
Makefile & Feature Installation instructions

AshtonStephens:
Use the following template to create your pull request

## Description

- Changes some package names to be more idomatic. `emily-operation-lambda` -> `emily-lambda`, etc.
- Change smithy generated source locations so clean command doesn't wipe `build.rs` file.
- Updates the local tools required to run the repository in README.md
- Adds a `Makefile` with some emily related commands and general build command.

Note that the `Makefile` is not complete with all the tools for all the monorepo packages. We should update the `Makefile` with the commands necessary to develop for the packages as we go.


AshtonStephens:
Converted to draft until I fix the CI.

8marz8 (approved):
LGTM. I was able to build it successfully following the readme. Just needed to update my CommandLineTools and Node.js (Got clear errors)

netrome (approved):
It works on my machine :+1:

.github/workflows/rust-tests.yaml
```diff
@@ -29,8 +33,8 @@ jobs:
...
 
-      # Build Emily first so that the autogenerated `lib.rs` file is
-      # present when building the rest of the crates in the workspace.
-      - run: cargo build --package emily
+      - name: Install package dependencies.
+        run: make install
 
-      - run: cargo test
+      - name: Run tests
+        run: make test
```
hstove, line 40: Should we just update this Github workflow to be something like "Run tests", since this will run both rust and JS tests? I'd be fine with that, and if there are Rust / Contract / etc workflows we want separately we can do that.

netrome:
Though one annoying point: `pnpm-lock.yaml` seems to change whenever I run either `make install` or `make test`.

emily/api-definition/package.json
```diff
@@ -1,10 +1,11 @@
 {
+  "name": "emily-api",
   "scripts": {
     "build-openapi": "smithy build",
     "clean-openapi": "smithy clean",
-    "build-rs": "npx @openapitools/openapi-generator-cli generate -i ../../.generated-sources/openapi/openapi/Emily.openapi.json -g rust -c rust-config.json -o ../../.generated-sources/emily",
+    "build-rs": "npx @openapitools/openapi-generator-cli generate -i ../../.generated-sources/smithy/openapi/openapi/Emily.openapi.json -g rust -c rust-config.json -o ../../.generated-sources/emily",
     "build": "npm run build-openapi && npm run build-rs",
```
hstove, line 7:
```suggestion
    "build": "pnpm build-openapi && pnpm build-rs",
```

emily/api-definition/package.json
```diff
@@ -1,10 +1,11 @@
...
-    "clean": "rm -rf node_modules  && npm run clean-openapi"
+    "clean": "rm -rf node_modules && npm run clean-openapi"
```
hstove, line 8:
```suggestion
    "clean": "rm -rf node_modules && npm clean-openapi"
```

hstove (approved):
LGTM! The makefile commands work as expected and are helpful.

One note is that, with `pnpm`, the `package-lock.json` files shouldn't be being updated (and should be deleted actually). That could indicate you're using both `npm install` and `pnpm install`. The `pnpm-lock.yaml` file at the root of the workspace acts as the lockfile for all workspace packages.

AshtonStephens:
> Though one annoying point: `pnpm-lock.yaml` seems to change whenever I run either `make install` or `make test`.

If you try committing that change locally and then rerunning `make install && make test` does it change further?

emily/api-definition/package.json
AshtonStephens: Why did that work before?

hstove:
@netrome if you run `pnpm -v` what version do you have? If it's not `9.x.x`, that could be causing the lockfile to change, but I'm not sure.

hstove:
Ah, yeah the current `pnpm-lock` file was generated with `pnpm` version 6, which is why the lockfile is changing. Happens for me too when I run `pnpm install` after checking out the repo.

hstove:
I've opened https://github.com/stacks-network/sbtc/pull/135 which should prevent accidental usage of different PNPM versions

.github/workflows/rust-tests.yaml
AshtonStephens: Yeah, I'll change this to just be `run tests` and keep the contracts one because it looks like it does more than just this does. Ideally these wouldn't need to be separate because GitHub workflowss aren't really for testing, they're for preventing regression, and so long as the workflow is blocked when we mess something and we cursorily know why then we're good.

emily/api-definition/package.json
hstove: Oh I have a typo but your code did it right (my suggestion is `npm clean..` but I meant `pnpm clean`.

With NPM you have to do `npm run X` for any script in your package.json, but pnpm automatically pulls all your script names into the "top level". So it works with pnpm but not npm.

PR #134, opened 2024-05-08
feat: GPL V3 License

AshtonStephens:
## Description

- Add GPL License
- Add GPL License badge to `README.md`

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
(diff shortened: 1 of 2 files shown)


netrome (approved):

netrome:
Should we also add copyright headers to all source files like we do in stacks-core?

AshtonStephens:
I've never understood that, isn't the license enough?

netrome:
> I've never understood that, isn't the license enough?

Don't know, but I assume the copyright headers are required because why else are they included everywhere in GPL3-licensed code? Could be worth looking into.

netrome:
Yeah it seems to be a requirement https://www.gnu.org/licenses/gpl-howto.en.html

"Here's a brief summary of what you need to do to release a program under one of our licenses:

    Get a copyright disclaimer from your employer or school.
    Give each file the proper copyright notices. Make sure [to clearly identify which versions of the license users can use](https://www.gnu.org/licenses/identify-licenses-clearly.html).
    Add a COPYING file with a copy of the GNU GPL or GNU AGPL.
    Also add a COPYING.LESSER file with a copy of the GNU LGPL, if you use that.
    Put a license notice in each file.
    (Optionally) make the program display a startup notice.
    (If using the AGPL) make the program offer copies of its source code.

This involves adding two elements to each source file of your program: a copyright notice (such as “Copyright 1999 Terry Jones”), and a statement of copying permission, saying that the program is distributed under the terms of the GNU General Public License (or the Lesser GPL, or the Affero GPL)."

PR #137, opened 2024-05-08
feat: Signer message serialization

### LINKED ISSUE #106: [Feature]: Signer message serialization
# Feature - Signer message serialization

## 1. Description

The signer message should support a well-defined serialization format, to allow these messages to be consistently broadcasted to other signers over the signer network.

### 2.1 Acceptance Criteria:
   - [ ] The SignerMessage type implements serialization logic

netrome:
closes #106

This PR introduces serialization logic for Signer messages by

1. Defining the `Encode` and `Decode` traits, which denote the canonical encoding format for signer messages, with a blanket implementation using Bincode for all types implementing serde::Serialize and serde::Deserialize.
2. Deriving `serde::Serialize` and `serde::Deserialize` for all message types.

### DIFF
--- a/signer/Cargo.toml
+++ b/signer/Cargo.toml
@@ -15,6 +15,7 @@ bench = false
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+bincode.workspace = true
 bitcoin = { workspace = true, features = ["rand-std"] }
 p256k1.workspace = true
 rand.workspace = true
--- a/signer/src/codec.rs
+++ b/signer/src/codec.rs
@@ -1,15 +1,14 @@
 //! # Canonical encoding and decoding for the sBTC signer
 //!
-//! This module defines encoding and decoding methods for messages in the signer.
-//! The primary purpose of these utilities is to facilitate convenient serialization and deserialization
-//! of messages that need to be exchanged between signers over a network.
+//! The purpose of this module is to define how to encode and decode
+//! signer messages as byte sequences.
 //!
-//! The module provides two main traits, `Encode` and `Decode`, which
-//! denote the canonical serialization format for any types implementing these traits.
+//! This is achieved by
 //!
-//! While these traits permit custom serialization implementations, this module also provides
-//! blanket implementations for any type that implements `serde::Serialize` and `serde::de::DeserializeOwned`,
-//! using [Bincode](https://docs.rs/bincode/1.3.3/bincode/).
+//! 1. Providing the `Encode` and `Decode` traits, defining the encode and decode
+//!    methods we intend to use throughout the signer.
+//! 2. Implementing these traits for any type implementing `serde::Serialize` and `serde::de::DeserializeOwned`
+//!    using `bincode` as the encoding format.
 //!
 //! ## Examples
 //!
--- a/signer/src/codec.rs
+++ b/signer/src/codec.rs
@@ -1,0 +1,47 @@
+use std::io;
+
+pub trait Encode: Sized {
+    fn encode<W: io::Write>(self, writer: W) -> Result<(), Error>;
+    fn encode_to_vec(self) -> Result<Vec<u8>, Error> {
+        let mut buff = Vec::new();
+        self.encode(&mut buff)?;
+        Ok(buff)
+    }
+}
+
+pub trait Decode: Sized {
+    fn decode<R: io::Read>(reader: R) -> Result<Self, Error>;
+    fn decode_from_bytes(bytes: &[u8]) -> Result<Self, Error> {
+        Self::decode(bytes)
+    }
+}
+
+impl<T: serde::Serialize> Encode for &T {
+    fn encode<W: io::Write>(self, writer: W) -> Result<(), Error> {
+        bincode::serialize_into(writer, self)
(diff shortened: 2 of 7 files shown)


signer/src/codec.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// from a slice of bytes, facilitating easier handling of raw byte data.
+    ///
+    /// # Arguments
+    /// * `bytes` - A byte slice from which the object should be decoded.
+    ///
+    /// # Returns
+    /// A `Result` which is `Ok` containing the decoded object, or an `Error` if decoding failed.
+    fn decode_from_bytes(bytes: &[u8]) -> Result<Self, Error> {
+        Self::decode(bytes)
+    }
```
djordon, line 91: Just curious about why this is needed given the implementation. It just seems like we can use `decode` all the time.

signer/Cargo.toml
```diff
@@ -15,6 +15,7 @@ bench = false
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
AshtonStephens, line 15: I cannot comment on code you didn't write (probably for the best) but I think we should call this package `signer` and not `sbtc-signer`. To run cargo tests on this package we need to run this:

```bash
cargo test --package sbtc-signer
```

Where every external indicator makes it look like it should be `cargo test --package signer`.

signer/src/codec.rs
AshtonStephens: +1

signer/src/codec.rs
```diff
@@ -0,0 +1,122 @@
+//! # Canonical encoding and decoding for the sBTC signer
+//!
+//! This module defines encoding and decoding methods for messages in the signer.
+//! The primary purpose of these utilities is to facilitate convenient serialization and deserialization
+//! of messages that need to be exchanged between signers over a network.
+//!
+//! The module provides two main traits, `Encode` and `Decode`, which
+//! denote the canonical serialization format for any types implementing these traits.
+//!
+//! While thiese traits permit custom serialization implementations, this module also provides
```
AshtonStephens, line 10:
```suggestion
//! While these traits permit custom serialization implementations, this module also provides
```

signer/src/codec.rs
```diff
@@ -0,0 +1,122 @@
+//! # Canonical encoding and decoding for the sBTC signer
+//!
+//! This module defines encoding and decoding methods for messages in the signer.
+//! The primary purpose of these utilities is to facilitate convenient serialization and deserialization
+//! of messages that need to be exchanged between signers over a network.
+//!
+//! The module provides two main traits, `Encode` and `Decode`, which
+//! denote the canonical serialization format for any types implementing these traits.
```
AshtonStephens, line 8: It might make sense to have this module header also justify why this file needs to exist, because my default expectation was that we could simply use the default serde traits, and this looks like we're wrapping serialization traits in another and I'm not sure why we have to do that.

signer/Cargo.toml
djordon: Yeah, that makes sense to me. I think I just neglected to change this when we settled on calling this crate `signer`.

signer/src/codec.rs
netrome: Yeah, this is 100% redundant. I didn't consider this when defining the trait, because it felt somewhat symmetric with `encode_to_vec`. I'll remove this function.

signer/Cargo.toml
netrome: Yeah I 100% agree. Do we have a ticket to change this? I'm happy to do it, but I prefer to do it as a separate change to keep our PRs focused and commit history clean.

netrome: Issue created, I'll have it done as soon as we've merged this PR. https://github.com/stacks-network/sbtc/issues/142

signer/src/codec.rs
netrome: Thanks for spotting! :pray:

signer/src/codec.rs
netrome: Done.

signer/src/codec.rs
netrome: I hoped that the current docs explained that, but I will try to make it clearer.

Let me first elaborate here; The difference between the serde traits and the `Encode`/`Decode` traits is that serde defines the data model to be serialized, but does not specify **how** to serialize the data. The trait essentially says: "This is how I map the current data structure to the serde data model". On top of that you can use any serializing method that supports the serde data model.

The `Encode`/`Decode` traits on the other hand are very different. They are essentially equivalent to the `StacksMessageCodec` trait, in that they specify exactly how to convert a given type into/from a sequence of bytes. This is what we need to be able to send messages over the signer network.

netrome: I hope this makes things clearer, but I'm happy to discuss this further if there is still some ambiguity in these traits or there purpose.

netrome: I updated the documentation of the module now with a more lean text focusing on the purpose of the module and outlining more clearly what the module provides.

AshtonStephens (approved):
LGTM

djordon (approved):
Looks good.

PR #138, opened 2024-05-08
Description - DRAFT

SundarGowtham:
This PR contains the Docker Compose Setup of the non-sBTC components:

- Bitcoin Regtest Node
- Bitcoin Miner
- Stacks Node
- Nakamoto Signer
- Electrs
- Stacks API
- Stacks Explorer
- Mempool Frontend
- Mempool Backend
- MariaDB
- Postgres


SundarGowtham:
You can run:

1) `sh ./build.sh`
2) `sh ./up.sh`

SundarGowtham:
For the Nakamoto Signer I get this error:

```
thread 'main' panicked at stacks-signer/src/main.rs:86:47: called `Result::unwrap()` on an `Err` value: BadField("node_host", ":")
```

These are the env variables being passed in for the signer:
```
      STACKS_NODE_RPC_HOST: "stacks"
      STACKS_NODE_RPC_PORT: "20443"
```

And this is the corresponding part of the Docker file in `./devenv/LOCAL/[0]-DOCKER-COMPOSE/nakamoto-signer/entrypoint.sh`:

```
sudo bash -c cat <<EOF> ./config.toml
# The IP address and port where your Stacks node can be accessed. 
# The port 20443 is the default RPC endpoint for Stacks nodes. 
# Note that you must use an IP address - DNS hosts are not supported at this time.
node_host = "$STACKS_NODE_RPC_HOST:$STACKS_NODE_RPC_PORT"
```

I'm not sure if Docker Compose allows us to pass in hard-coded IPs instead of container aliases

SundarGowtham:
Something similar is also happening with the Stacks Node.
```
[testnet/stacks-node/src/main.rs:422] [main] Invalid config: Invalid burnchain.peer_host: failed to lookup address information: Name or service not known
```

@hstove Does the recent `2.5.0.0.3` version of Stacks Core also not support DNS names?

hstove:
A few notes on the PR at a high level:

- Can you rebase off of `main`? There shouldn't be any conflicts
- Can you make the folder structure a little bit easier to navigate and "canonical"? Ie use lowercase descriptive folder names (i think "devnet" for local and "testnet" for hosted), and no use the `[0]-` folder prefixes? Those prefixes make it harder to quickly `cd` through the folders

PR #139, opened 2024-05-09
Integrate with risk provider with error handling and unit tests

### LINKED ISSUE #77: Dynamic Blocklist Integration with Risk Analysis Provider
This is part of https://github.com/Trust-Machines/sbtc-v1/issues/54

Integrate with Risk Analysis Provider and update the API handlers accordingly.

8marz8:
Closes #77 

This PR makes the blocklist client functional and ready to integrate with sBTC signers.
Changes include:
- Integrate with a risk provider
- Expand error handling/propagation 
- Add unit tests for the client
- Add comments and improve code organization


blocklist-client/src/client/client.rs
```diff
@@ -1,20 +1,316 @@
...
+        (client, config)
+    }
+
+    // Helper function to setup a mock API response
+    fn setup_mock(method: &str, path: &str, status: u16, body: &str) -> Mock {
+        return mock(method, path)
+            .with_status(status.into())
+            .with_header("content-type", "application/json")
+            .with_body(body)
+            .create();
```
djordon, line 160: Nice. I've used `httpmock` for this kind of functionality, but `mockito` seems to have a bit of a nicer API. Good to know.

blocklist-client/src/client/client.rs
```diff
@@ -1,20 +1,316 @@
...
+        match result {
+            Ok(_) => panic!("Test failed: Expected an Error::InvalidApiResponse, but got Ok"),
+            Err(e) => match e {
+                Error::InvalidApiResponse => {
+                    assert!(true, "Received the expected Error::InvalidApiResponse");
+                }
+                _ => panic!(
+                    "Test failed: Expected Error::InvalidApiResponse, got {:?}",
+                    e
+                ),
```
djordon, line 235: I think you can just do this:
```rust
panic!("Test failed: Expected Error::InvalidApiResponse, got {e}")
```

blocklist-client/src/client/client.rs
```diff
@@ -1,20 +1,316 @@
...
+            r#"{"risky": "Severe"}"#,
+        );
+        let (client, config) = setup_client();
+
+        let result = get_risk_assessment(&client, &config, TEST_ADDRESS).await;
+        match result {
+            Ok(_) => panic!("Test failed: Expected an Error::InvalidApiResponse, but got Ok"),
+            Err(e) => match e {
+                Error::InvalidApiResponse => {
+                    assert!(true, "Received the expected Error::InvalidApiResponse");
```
djordon, line 230: Simpler to just leave these kinds of asserts as comments.

blocklist-client/src/common/mod.rs
```diff
@@ -1,21 +1,35 @@
...
+    // Blocklist client's acceptance decision based on the risk severity of the address
+    pub accept: bool,
 }
 
-#[derive(Serialize)]
-pub struct ErrorResponse {
-    pub(crate) message: String,
+/// Risk severity linked to an address
+#[derive(Debug, PartialEq, Eq, Deserialize)]
+pub enum RiskSeverity {
```
djordon, line 19: The implementations here are perfectly fine as is, but in the future you could also use something like [strum](https://github.com/Peternator7/strum) to automatically implement `Display` and `FromStr` for enums.

djordon (commented):
Looks good, I just had a question about `BlocklistStatus` fields.

blocklist-client/src/client/client.rs
```diff
@@ -1,20 +1,316 @@
...
+        "Received risk assessment response: {} for address: {}",
+        register_response.address, register_response.address
+    );
+
+    let blocklist_status = BlocklistStatus {
+        // `is_blocklisted` is set to true if risk is Severe
+        is_blocklisted: matches!(assessed_response, RiskSeverity::Severe),
+        severity: assessed_response.to_string(),
+        // `accept` is set to false if severity is Severe
+        accept: !matches!(assessed_response, RiskSeverity::Severe),
```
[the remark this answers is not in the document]
8marz8: In the LLD, I initially only had `is_blocklisted` but the `accept` field was suggested in a comment. Are you recommending that only one field is necessary? 
They're inverse of each other. (so basically the same)

@hstove  Can you chime in here on what is the benefit of having a separate `accept` field in addition to `is_blocklisted`?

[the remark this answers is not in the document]
8marz8: Oh I see. So if we maintain a separate blocklist in the future, we could refer to our list as well as the provider assessment. In that case, maybe I can set `is_blocklisted` statically to `false` and the `accept` field would be based on the risk severity fetched from the provider. 

I will add the `reason` field too.

8marz8:
> I'm good to approve this, but can you fix the clippy lint warnings first?

Where do you see those warnings? Do I need to run something to see that?

blocklist-client/src/client/client.rs
8marz8: It would panic if it doesn't match that error...so basically it's a dummy assert. I can just add a comment as Daniel suggested.

AshtonStephens:
```
cargo clippy --package blocklist-client
```

or you can rebase from `main` and run this in project root:

```
make lint
```

AshtonStephens (approved):
LGTM

djordon (approved):
Looks good. I had one non-blocking style suggestion. Curious about your thoughts.

blocklist-client/src/client/risk_client.rs
```diff
@@ -0,0 +1,352 @@
...
+struct RegistrationResponse {
+    address: String,
+}
+
+#[derive(Deserialize, Debug)]
+struct RiskResponse {
+    risk: Option<String>,
+    #[serde(rename = "riskReason")]
+    risk_reason: Option<String>,
+}
```
netrome, line 19: Why is `risk` optional here? The `get_risk_assessment` function returns an `InvalidApiResponse` error if this is not provided.

Also, it seems to me that this type is only an intermediate representation used to construct a `RiskAssessment`. It would be simpler to try to deserialize the response in `get_risk_assessment` directly as a `RiskAssessment`, and skip the intermediate `RiskResponse` struct completely. Am I missing something here or is there any good reason to keep the `RiskResponse` type around?

blocklist-client/src/client/risk_client.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    match resp.risk {
+        Some(risk_str) => {
+            let severity = match risk_str.as_str() {
+                "Low" => RiskSeverity::Low,
+                "Medium" => RiskSeverity::Medium,
+                "High" => RiskSeverity::High,
+                "Severe" => RiskSeverity::Severe,
+                _ => return Err(Error::InvalidRiskValue(risk_str)),
+            };
```
netrome, line 79: Nit: It would be more idiomatic to put this match statement into a `FromStr` implementation for `RiskSeverity`. With that, this code would simplify to:
```suggestion
            let severity = risk_str.parse().map_err(|_| Error::InvalidRiskValue(risk_str))?;
```

The `.map_err` could potentially be skipped if the error type used in the `FromStr` implementation implements `Into<Error>` for the common error type.

netrome: Note that this is not needed if we deserialize the response as a `RiskAssessment` directly as per my previous comment.

blocklist-client/src/client/risk_client.rs
```diff
@@ -0,0 +1,352 @@
...
+    debug!("Address registered: {}", register_response.address);
+
+    // If registration is successful, proceed to check the address
+    let RiskAssessment { severity, reason } = get_risk_assessment(client, config, address).await?;
+    debug!(
+        "Received risk assessment: Severity = {}, Reason = {}",
+        severity, reason
+    );
+
+    let is_severe = matches!(severity, RiskSeverity::Severe);
```
netrome, line 107: Nit: This would be nice as a method on the `RiskSeverity` struct, so we could just do `severity.is_severe()`.

blocklist-client/src/client/risk_client.rs
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    // Helper function to setup a mock API response
+    fn setup_mock(method: &str, path: &str, status: u16, body: &str) -> Mock {
+        return mock(method, path)
+            .with_status(status.into())
+            .with_header("content-type", "application/json")
+            .with_body(body)
+            .create();
+    }
```
netrome, line 176: I'm not super familiar with `mockito` (just looking through their docs now), but I think you'll want to use `.create_async` here.

blocklist-client/src/common/mod.rs
```diff
@@ -1,21 +1,37 @@
...
+pub struct BlocklistStatus {
+    // Whether the address is blocklisted or not
+    pub is_blocklisted: bool,
+    // The risk severity associated with an address
+    pub severity: String,
+    // Blocklist client's acceptance decision based on the risk severity of the address
+    pub accept: bool,
+    // Reason for the acceptance decision
+    pub reason: String,
 }
```
netrome, line 17: I guess that we might want to reuse this type in the signer so it would be nice to implement some common traits for it:

```suggestion
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
pub struct BlocklistStatus {
    // Whether the address is blocklisted or not
    pub is_blocklisted: bool,
    // The risk severity associated with an address
    pub severity: String,
    // Blocklist client's acceptance decision based on the risk severity of the address
    pub accept: bool,
    // Reason for the acceptance decision
    pub reason: String,
}
```

netrome: Also, why is `severity` a string here and not a `RiskSeverity`?

netrome (approved):
A few questions and nits, but the code seems to be doing what it's supposed to do.

blocklist-client/src/client/risk_client.rs
8marz8: Removed optional for risk (but risk_reason needs to be optional)
I wanted to separate the API data mapping from the business logic struct...in case we want to manipulate those differently but since this is not the case here, I removed the RiskResponse type -> this made the error mapping a bit different

blocklist-client/src/client/risk_client.rs
8marz8: `create_async` was part of newer versions. I tried updating the version and the test logic but I couldn't get the mock to work properly...wrestled with it for some time but keeping it the way it is for now.

PR #144, opened 2024-05-13
CI: add workflow for lints

### LINKED ISSUE #143: CI: Add workflow for lints
We should add a github workflow for running `make lint`. In particular, code that produces Clippy warnings should not pass CI.

netrome:
closes #143


8marz8 (approved):
lgtm!

AshtonStephens:
Can we make this a step or part of the unit test workflow? It feels like the CI is duplicating work unnecessarily

netrome:
> Can we make this a step or part of the unit test workflow? It feels like the CI is duplicating work unnecessarily
> 
> I'm wondering about the reasoning behind this choice. I recognize it's better not to muddy different workflows with different kinds of data but I bet there's a way to do some of this processing only once and still keep the errors clear.

Sure thing, I think that makes sense. I'll update.

netrome:
> Can we make this a step or part of the unit test workflow? It feels like the CI is duplicating work unnecessarily
> 
> I'm wondering about the reasoning behind this choice. I recognize it's better not to muddy different workflows with different kinds of data but I bet there's a way to do some of this processing only once and still keep the errors clear.

Updated.

AshtonStephens (approved):

Issue #108, opened 2024-05-02
[Feature]: Signer network in-process implementation


netrome:
# Feature - Signer network in-process implementation

## 1. Description
Implement an in-process implementation of the signer network usable for unit tests.

### 1.1 Context & Purpose
To allow signer operations to be easily unit-testable, we should have an implementation of the signer network interface which operates entirely in-process without any I/O. This implementation should allow unit tests to broadcast messages between thread boundaries, and support interfering with the network process such as delaying the delivery of a message or blocking a message from being delivered. It should also support making assertions over messages being sent and received.

### 2.1 Acceptance Criteria:
   - [ ] There is an implementation of the Signer network which runs entirely in-process.
   - [ ] The implementation supports delaying or blocking messages.
   - [ ] The implementation allows inspection of sent and received messages for different consumers.

Issue #109, opened 2024-05-02
[Feature]: Signer network HTTP relay implementation


netrome:
# Feature - Signer network HTTP relay implementation

## 1. Description
Implement the signer network interface with a client that connects to an external HTTP relay, which forwards messages between connected clients.

### 1.1 Context & Purpose
The HTTP relay implementation is needed to allow separate signer processes running on different hosts to connect to each other and exchange messages.

### 2.1 Acceptance Criteria:
   - [ ] There is a HTTP relay service defined which can forward messages between signers.

PR #145, opened 2024-05-13
feat: Signer network interface & in-memory implementation

### LINKED ISSUE #107: [Feature]: Signer network interface
# Feature - Signer network interface

## 1. Description

Define the interface for interacting with the signer network as per the [signer communication LLD](https://docs.google.com/document/d/1q1Ket3ue_H8J3HlvpLhM5-RLRthmw90BugzPwrVpYuw/).

### 2.1 Acceptance Criteria:
   - [ ] The network interface is defined

netrome:
closes #107 and #108 

This PR defines the client interface for the signer network through the `signer::network::Client` trait. It also provides an in-memory implementation of a signer network by wrapping a `tokio::sync::broadcast` channel for illustrative and testing purposes.

### DIFF
--- a/signer/src/network/in_memory.rs
+++ b/signer/src/network/in_memory.rs
@@ -1,0 +1,73 @@
+//! # In-memory signer network client.
+//!
+//! The client itself is a thin wrapper over a tokio broadcast
+//! channel, with deduplication logic to prevent a single client
+//! from receiving it's own messages.
+
+use std::collections::VecDeque;
+
+use tokio::sync::broadcast;
+
+pub const BROADCAST_CHANNEL_CAPACITY: usize = 10_000;
+
+#[derive(Debug)]
+pub struct Client {
+    sender: broadcast::Sender<super::Msg>,
+    receiver: broadcast::Receiver<super::Msg>,
+    recently_sent: VecDeque<super::Msg>,
+}
+
+/// In-memory communication network
+#[derive(Debug)]
+pub struct Network {
+    sender: broadcast::Sender<super::Msg>,
+}
+
+impl Network {
+    /// Construct a new in-memory communication entwork
+    pub fn new() -> Self {
+        let (sender, _) = broadcast::channel(BROADCAST_CHANNEL_CAPACITY);
+        Self { sender }
+    }
+
+    /// Connect a new signer to this network
+    pub fn connect(&self) -> Client {
+        let sender = self.sender.clone();
+        let receiver = sender.subscribe();
+        let recently_sent = VecDeque::new();
+
+        Client {
+            sender,
+            receiver,
+            recently_sent,
+        }
+    }
+}
+
+impl super::Client for Client {
+    type Error = Error;
+    async fn broadcast(&mut self, msg: super::Msg) -> Result<(), Self::Error> {
+        self.recently_sent.push_back(msg.clone());
+        self.sender.send(msg)?;
+        Ok(())
+    }
+
+    async fn receive(&mut self) -> Result<super::Msg, Self::Error> {
+        let mut msg = self.receiver.recv().await?;
+
+        while Some(&msg) == self.recently_sent.get(0) {
+            self.recently_sent.pop_front();
+            msg = self.receiver.recv().await?;
+        }
+
+        Ok(msg)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("send error")]
+    Send(#[from] broadcast::error::SendError<super::Msg>),
+    #[error("receive error")]
+    Recv(#[from] broadcast::error::RecvError),
+}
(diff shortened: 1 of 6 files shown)


8marz8 (approved):

8marz8:
Just a general question: What will come later as a replacement for the in_memory implementation?

netrome:
> Just a general question: What will come later as a replacement for the in_memory implementation?

Good question, the implementation of the signer network for sBTC-v1 is the HTTP relay https://github.com/stacks-network/sbtc/issues/109 - which is what I'm looking into now.

On a longer horizon I see the possibility of implementing this interface backed by StackerDB. If future requirements change we might take this in another direction and implement a pure peer to peer network based on `libp2p` or a custom implementation. But I don't see that happening any time soon though :)

signer/src/network/in_memory.rs
```diff
@@ -0,0 +1,73 @@
...
+    async fn receive(&mut self) -> Result<super::Msg, Self::Error> {
+        let mut msg = self.receiver.recv().await?;
+
+        while Some(&msg) == self.recently_sent.get(0) {
+            self.recently_sent.pop_front();
+            msg = self.receiver.recv().await?;
+        }
+
+        Ok(msg)
+    }
```
djordon, line 64: Nice, I see. This is a simple network that works. Everyone receives everyone else's messages and I also receive my own messages. 👍🏽 

An alternative approach, is to give each message a unique ID and filter on IDs that you have broadcasted. That way you do not need to clone the message and you get (roughly) the same guarantees. The approach here is fine too.

signer/src/network/in_memory.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        Client {
+            sender,
+            receiver,
+            recently_sent,
+        }
+    }
+}
+
+impl super::Client for Client {
```
djordon, line 47: I think we should call either the struct or the trait something different than `Client` (I kind of think we should rename them both). Maybe `InternalNetworkClient` for the trait and `MpmcBroadcaster` for the struct. These aren't the best suggestions, I'm just trying to help.

djordon (commented):
Looks good, I just have some quibbles with using the names `Client` for the trait and the struct. I'd prefer something a little more informative.

signer/src/network/in_memory.rs
netrome: Sure thing, I'm used to different conventions in different repos and I guess we'll figure out what feels best for us. 

On one hand, I think this works pretty well to reuse simple names such as `Error` or `Client` if the name space is informative. In this scenario, I think `in_memory::Client` looks pretty good in the code. I originally called it `InMemoryClient` which felt a bit repetitive with the namespacing, but works better for people who prefer to import structs by name without any namespacing information. Though I'm personally not a big fan of direct type imports, particularly because it makes it very hard to disambiguate between type imports and bringing traits into scope.

On the other hand, I also think it's ugly to have the same name for the trait and the implementation. Typically traits should be verbs (Clone, Print, Serialize, Encode etc.) - so from this convention it is the trait name that should be updated.

From this reasoning, I would propose renaming the trait to `MessageTransfer`, and keep the in memory network client type name to just `Client`.

Does that sound good to you?

signer/src/network/in_memory.rs
netrome: Yeah, this is a bit of an ugly hack to prevent the signers own messages from echoing back. I like the idea of using IDs instead of matching on the entire message. I guess being able to identify messages will come in handy in the future too. I'll update.

signer/src/network/in_memory.rs
netrome: `MpmcBroadcaster` is more specific than `Client` though. I like that name, so let's use it.

TL; DR
I will rename the trait to `MessageTransfer` and the struct to `MpmcBroadcaster`.

djordon: > On one hand, I think this works pretty well to reuse simple names such as `Error` or `Client` if the name space is informative. In this scenario, I think `in_memory::Client` looks pretty good in the code. I originally called it `InMemoryClient` which felt a bit repetitive with the namespacing, but works better for people who prefer to import structs by name without any namespacing information. Though I'm personally not a big fan of direct type imports, particularly because it makes it very hard to disambiguate between type imports and bringing traits into scope.

Ahhh, I see. Yeah I can relate, even though I'm definitely guilty of this in my code.

> On the other hand, I also think it's ugly to have the same name for the trait and the implementation. Typically traits should be verbs (Clone, Print, Serialize, Encode etc.) - so from this convention it is the trait name that should be updated.
> 
> From this reasoning, I would propose renaming the trait to `MessageTransfer`, and keep the in memory network client type name to just `Client`.
> 
> Does that sound good to you?

🙏🏽 Yeah that sounds great.

netrome: Updated. Let me know what you think.

signer/src/network/in_memory.rs
netrome: Updated.

djordon (approved):
Looks good!

PR #146, opened 2024-05-13
ci: add transaction package integration tests

djordon:
## Description

Address part of #67.

This PR mainly sets up integration tests with bitcoin-core. Specifically it:
1. Makes sure bitcoin-core is running in CI in regtest mode.
2. Add helper functions and structs for running integration tests with bitcoin-core in regtest mode.
3. Adds an integration test that makes sure that the signed version of the `UnsignedTransactions` returned in the `utxo` module successfully moves deposits to the signer's UTXO.
4. Updates the Schnorr public keys in the `utxo` module to use `XOnlyPublicKey` since bitcoin-core expects Schnorr public keys to be 32-bit x-coordinate only keys.
5. Change tests to run on every push instead of only when the PR is open and changes have been made to it. I don't mind changing this back.

## Testing information

This PR mainly adds tests but also changes how testing is done. The biggest difference is that `cargo test` now runs our integration tests and will fail if bitcoin-core is not running locally. To get around that do `cargo test --lib` or `make test` if you don't want to run integration tests.

## Checklist
- [x] `make integration-test` passes


emily/api-definition/package-lock.json
djordon, line 1: Is this supposed to be here? It showed up after the rebase, I'm not sure if I need to gitignore it something.

.github/workflows/unit-tests.yaml
```diff
@@ -1,10 +1,5 @@
 name: Pull Request
```
AshtonStephens, line 1: Should probably rename

AshtonStephens, line 1: Should probably rename to push

.github/workflows/unit-tests.yaml
```diff
@@ -1,10 +1,5 @@
...
-on:
```
AshtonStephens, line 2: Is the intention that this runs every time there's a push to any branch or only pushes to the default branch?

.github/workflows/unit-tests.yaml
```diff
@@ -36,5 +31,10 @@ jobs:
       - name: Install package dependencies.
         run: make install
 
+      - name: Setup service containers
+        run: |
+          docker compose -f docker-compose.test.yml pull
+          docker compose -f docker-compose.test.yml up --detach
```
AshtonStephens, line 37: Yeah, I think it would probably make the most sense to have a make command that runs something similar so we can recreate this easily locally with out needing to read this workflow.

emily/api-definition/package-lock.json
AshtonStephens: This is certianly my fault, and this makes me realize I don't think I understand the idioms around lock files. Do these need to be in each package, or is `pnpm` meant to make `package-lock.json` files within the workspace root and not each sub package.

signer/src/utxo.rs
```diff
@@ -98,9 +97,15 @@ pub struct DepositRequest {
     /// The redeem script for the deposit.
     pub redeem_script: ScriptBuf,
     /// The public key used for the key-spend path of the taproot script.
-    pub taproot_public_key: PublicKey,
-    /// The public key used in the deposit script.
-    pub signers_public_key: PublicKey,
+    /// Note that taproot Schnorr public keys are slightly different from
```
AshtonStephens, line 100:
```suggestion
    ///
    /// Note that taproot Schnorr public keys are slightly different from
```

signer/tests/regtest/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+use bitcoin::PublicKey;
+use bitcoin::ScriptBuf;
+use bitcoin::Sequence;
+use bitcoin::TapLeafHash;
+use bitcoin::TapNodeHash;
+use bitcoin::TapSighashType;
+use bitcoin::Transaction;
+use bitcoin::TxIn;
+use bitcoin::TxOut;
+use bitcoin::Witness;
```
AshtonStephens, line 22: What are your thoughts on this import style?
```rust
use bitcoin::{Address, AddressType, Amount, CompressedPublicKey, ...}
```

djordon: I used to prefer the default style, but now I think this `item` style is better. It makes diffs easier to see/read and makes resolving any git conflicts simpler. I know it's the default style so I'm okay switching to it.

.github/workflows/unit-tests.yaml
djordon: Already have in ae497a873ceaf58b8d2e28e0b09e66b429b3c364.

.github/workflows/unit-tests.yaml
djordon: It's intentional to run on every push to every branch, mainly to mirror what was done in our `contract-tests.yaml` workflow. I don't have a preference here, I'm okay with changing it back.

signer/tests/regtest/mod.rs
AshtonStephens: Oh, no this is a reasonable case for it, I kind of like it

.github/workflows/unit-tests.yaml
AshtonStephens: Oh, yeah we should change it since it does cost $$$ And I don't think it's really necessary

AshtonStephens: As in, I'll push things just so my laptop isn't the single point of failure for my work, I would never check the CI for that. I don't know if that's just me.

.github/workflows/unit-tests.yaml
djordon: Oh I forgot to update this after that commit, will update now.

emily/api-definition/package-lock.json
netrome: I don't know about that particular lock file, but when it comes to committing lock files or not the rule of thumb I'm following is that it should be committed for binaries but not for libraries.

Since this is primarily a binary (or a set of binaries) we're developing and not a generic library to be imported by other people, I believe we should generally commit lock files.

Though this is a Cargo convention, and I don't know what the convention is in JavaScript land.

djordon: So I think the this was generated from my `.generated-sources/emily/build.rs` file, which calls `npm` still. Should I just delete that file, is it still relevant?

signer/tests/regtest/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+            .get_received_by_address(&faucet.address, Some(1))
+            .unwrap();
+
+        if amount < Amount::from_int_btc(1) {
+            faucet.generate_blocks(&rpc, 101);
+        }
+
+        (rpc, faucet)
+    })
+}
```
netrome, line 86: I like the pattern in the first example at https://doc.rust-lang.org/std/sync/struct.OnceLock.html where the static variable is local for the getter function, since it should only be read through that function.

```suggestion
/// Initializes a blockchain and wallet on bitcoin-core. It can be called
/// multiple times (even concurrently) but only generates the client and
/// recipient once.
///
/// This function does the following:
/// * Creates an RPC client to bitcoin-core
/// * Creates or loads a watch-only wallet on bitcoin-core
/// * Loads a "faucet" private-public key pair with a P2WPKH address.
/// * Has the bitcoin-core wallet watch the generated address.
/// * Ensures that the faucet has at least 1 bitcoin spent to its address.
pub fn initialize_blockchain() -> &'static (Client, Recipient) {
    static BTC_CLIENT: OnceLock<(Client, Recipient)> = OnceLock::new();
    BTC_CLIENT.get_or_init(|| {
        let username = BITCOIN_CORE_RPC_USERNAME.to_string();
        let password = BITCOIN_CORE_RPC_PASSWORD.to_string();
        let auth = Auth::UserPass(username, password);
        let rpc = Client::new("http://localhost:18443", auth).unwrap();

        get_or_create_wallet(&rpc, BITCOIN_CORE_WALLET_NAME);
        let faucet = Recipient::from_key(FAUCET_SECRET_KEY, AddressType::P2wpkh);
        faucet.track_address(&rpc, FAUCET_LABEL);

        let amount = rpc
            .get_received_by_address(&faucet.address, Some(1))
            .unwrap();

        if amount < Amount::from_int_btc(1) {
            faucet.generate_blocks(&rpc, 101);
        }

        (rpc, faucet)
    })
}
```

signer/tests/regtest/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+        Err(_) => {
+            // We want a wallet that is watch only, since we manage keys
+            let disable_private_keys = Some(true);
+            rpc.create_wallet(wallet, disable_private_keys, None, None, None)
+                .unwrap();
+        }
+    };
+}
+
+/// Helper struct for representing an address on bitcoin.
```
netrome, line 104: This structs represents more than just an address, since that is fully covered by the `Address` type. I assume the extra information is the fact that we have a key pair associated with this address, so I suggest updating the comment to something like

```suggestion
/// Helper struct for representing an address we control on bitcoin.
```

signer/tests/regtest/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+pub struct Recipient {
+    pub keypair: Keypair,
+    pub address: Address,
+}
+
+fn descriptor_base(public_key: &PublicKey, kind: AddressType) -> String {
```
netrome, line 110: FYI, [bdk](https://docs.rs/bdk/0.29.0/bdk/index.html) provides a lot of helpful types for working with descriptors if you haven't looked into it already.

emily/api-definition/package-lock.json
AshtonStephens: Yeah, we still need it because it adds the clippy tags

signer/tests/regtest/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+}
+
+/// The enum `Either` with variants `Left` and `Right` is a general purpose
+/// sum type with two cases.
+pub enum Either<L, R> {
+    /// A value of type `L`.
+    Left(L),
+    /// A value of type `R`.
+    Right(R),
+}
```
netrome, line 318: Might be overkill here, but have you considered using the [Either](https://docs.rs/either/latest/either/) crate for this?

signer/tests/utxo_construction.rs
```diff
@@ -0,0 +1,255 @@
...
+use signer::utxo::SignerUtxo;
+
+mod regtest;
+use regtest::Either;
+use regtest::Recipient;
+
+const SIGNER_ADDRESS_LABEL: Option<&str> = Some("signers-label");
+const DEPOSITS_LABEL: Option<&str> = Some("deposits");
+
+fn make_deposit(
```
netrome, line 35: Nit: Just to clarify that this creates a deposit request.

```suggestion
fn make_deposit_request(
```

signer/tests/utxo_construction.rs
```diff
@@ -0,0 +1,255 @@
...
+    depositor: &Recipient,
+    amount: u64,
+    utxo: &ListUnspentResultEntry,
+    signers_public_key: XOnlyPublicKey,
+    faucet_public_key: XOnlyPublicKey,
+) -> (Transaction, DepositRequest) {
+    let fee = regtest::BITCOIN_CORE_FALLBACK_FEE.to_sat();
+    let deposit_script = ScriptBuf::builder()
+        .push_slice([1, 2, 3, 4])
```
netrome, line 44:
```suggestion
        .push_slice([1, 2, 3, 4]) // Just some dummy data, since we don't test the parsing of the sBTC request data here.
```

signer/tests/utxo_construction.rs
```diff
@@ -0,0 +1,255 @@
...
+        taproot_public_key: faucet_public_key,
+        signers_public_key,
+    };
+    (tx, req)
+}
+
+/// This test just checks that many of the methods on the Recipient struct
+/// work as advertised.
+#[test]
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
```
netrome, line 108: Nit: Instead of individually feature flagging the tests, we could have a single cfg attribute for the whole module, saving us some bytes in the code and some compile time when building and running non-integration tests.

emily/api-definition/package-lock.json
netrome: Wait I'm confused, the lock file doesn't add any Clippy labels right? I mean, I frequently delete it from my local repo and that seems to be alright.

djordon: > Wait I'm confused, the lock file doesn't add any Clippy labels right? I mean, I frequently delete it from my local repo and that seems to be alright.

I think he was referring to `.generated-sources/emily/build.rs`. Also, should that file be unignored?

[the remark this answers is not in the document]
djordon: Yeah, I'll remove it and add it in a separate PR.

djordon: We should figure this all out in a separate PR. I'll be removing this and adding it back later.

signer/src/utxo.rs
```diff
@@ -528,8 +538,8 @@ mod tests {
                 .push_opcode(opcodes::all::OP_CHECKSIG)
                 .into_script(),
             redeem_script: ScriptBuf::new(),
-            taproot_public_key: PublicKey::from_str(PUBLIC_KEY0).unwrap(),
-            signers_public_key: generate_public_key(),
+            taproot_public_key: to_x_only_public_key(PublicKey::from_str(PUBLIC_KEY0).unwrap()),
```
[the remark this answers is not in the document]
djordon: I just forgot about `XOnlyPublicKey::from_str` 😄

signer/tests/regtest/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+use bitcoincore_rpc::json::ListUnspentResultEntry;
+use bitcoincore_rpc::json::Timestamp;
+use bitcoincore_rpc::jsonrpc::error::Error as JsonRpcError;
+use bitcoincore_rpc::jsonrpc::error::RpcError;
+use bitcoincore_rpc::Auth;
+use bitcoincore_rpc::Client;
+use bitcoincore_rpc::Error as BtcRpcError;
+use bitcoincore_rpc::RpcApi;
+use secp256k1::Keypair;
+use secp256k1::Message;
```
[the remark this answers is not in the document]
djordon: Fair point. I'll sort this out.

signer/tests/regtest/mod.rs
djordon: Good to know, it was completely off of my radar.

signer/tests/regtest/mod.rs
djordon: This is actually copied from that crate. I did consider importing it and thought it would be overkill. If we need it for something else I'm okay using the crate.

signer/tests/utxo_construction.rs
djordon: This turns out to be a little more difficult to do than I would like. You have to use `cfg` on the module, but then rust-analyzer doesn't analyze the module by default since it doesn't check for all features by default. You can change it so that rust-analyzer picks up any or all features, but then I'd like to get more buy-in from the team for such a change. For now, keeping the `cfg_attr` on each test is the simplest solution so I'd like to keep it like this until we are all on board for such a change.

I did change things so that they are now under an `integration` folder. If all tests are in that folder, then this will limit the number of test binaries that cargo builds during `cargo test` to one.

netrome: Right, that makes sense. Thank you for trying it out at least. I'm happy to leave this as-is :)

Makefile
```diff
@@ -9,6 +9,12 @@ test: install | emily-client
 	cargo test \
 		&& pnpm --recursive test
 
+integration-test: install | emily-client
+	docker compose --file docker-compose.test.yml up --detach
+	cargo test --all-features \
+		&& pnpm --recursive test
+	docker compose --file docker-compose.test.yml down
```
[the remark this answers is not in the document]
djordon: > Personally, I'd prefer to have the environment set up be more explicit and not hidden behind a `make integration-test` call. Alternatively have a more verbose name for this: `make integration-test-with-docker-environment` (not the best suggestion but you get the idea).

Yeah, maybe we should all chat about this on Friday?


> Also, should we figure out a way to run these integration tests in the CI? That can definitely be done as a follow-up, since I expect the environment setup should be done differently and not with docker-compose within the actual test runner.

We "could" using [GitHub action services](https://docs.github.com/en/actions/using-containerized-services/about-service-containers), but last I checked they do not expose enough options for us to mount a file on the container like we do with docker for `bitcoin.conf`. Plus I think it's better to have CI do the exact same thing that we do locally.

signer/tests/regtest/mod.rs
netrome: Hahah I thought it looked familiar :)

Makefile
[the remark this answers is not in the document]
netrome: Okay interesting. I'm okay with this for now as well. I can see that it is convenient, and for me it's easy to run the integration tests explicitly as well if I prefer.

PR #149, opened 2024-05-14
Add common crate to workspace to avoid cyclic dependencies

8marz8:
This PR adds a common crate to avoid cyclic dependency between blocklist_client and signer. 
After this, the signer can pull in blocklist_client as a lib so it can use the common data types.


netrome (approved):
Nice!

Issue #74, opened 2024-04-26
Deposit API Local environment for x86


AshtonStephens:
## Deposit API Local environment for x86

This is part of #50 

Setup a local development environment for computers that use x86 chips. After this ticket someone should be able to spin up a local testing environment on an x86 computer to test their API code. 

This ticket is to create a bare-minimum local environment that can be salvaged into code for #58

PR #150, opened 2024-05-14
Emily API Local Environment

AshtonStephens:
# Emily API Local Environment

## Description

This PR adds a local Emily development environment.
1. Add Emily integration test environment - **necessary**
1. Add Docker scripts to modify local environment - **necessary**
1. Change CDK Lambda to use `x86` when running locally on an `x86` machine - **necessary**
1. Optimizes Makefile caching - **really nice**

Note that this may still not work for some machines with weird docker setups.

## Makefile Optimization

I included Makefile caching because running `cargo-lambda` takes a while (around 1 minute) when `cargo` believes that the sources for `emily/lambda` or the dependencies of the sources have been changed. We've been using Makefiles incorrectly so far; each target is loosely meant to be a file, and in cases where the target isn't a file you're meant to make it a dependency of `.PHONEY`. 

For example, this command doesn't make the file `install` so it's departing from the "file as target" paradigm.
```Makefile
install: 
	pnpm install
.PHONEY: install
```
Makefiles define a target to be out of date when the target **corresponds to a file** that has a "last modified" timestamp **earlier** than the targets it's dependent on **OR** if any of the dependencies are out of date. Note that the definition of out of date is:
1. **Recursive** - so for caching to work all dependencies all the way down need to cache properly
1. **Entirely dependent on file timestamps** - so all dependencies should be tied to a file 
An approach to this that takes advantage of the caching for installation needs to associate the installation with a file that can be updated when there's a dependency change. Here's the updated installation task from this PR:
```Makefile
INSTALL_TARGET := pnpm-lock.yaml
$(INSTALL_TARGET): $(wildcard package* */package* */*/package*)
	pnpm install
	touch pnpm-lock.yaml
```
Here we're using the last updated timestamp of the `pnpm-lock.yaml` to indicate if the installation is out of date or needs to be rerun because the `package.json` files have been modified. I've kept the `install` target because when the repository is first pulled in the "last modified" timestamp for ` pnpm-lock.yaml` and the `package.json` files will have the same timestamp so `make` will refuse to reinstall because the target is up to date, the `install` command forgoes that.

## Other changes

Comments added in the files

## Testing

Ran the integration environment from a fresh repository by running the following:

```
make install && make emily-integration-test
```

and then hit it with

```
curl http://127.0.0.1:3000/withdrawals | jq
```


emily/api-definition/package.json
```diff
@@ -3,7 +3,7 @@
   "scripts": {
     "build-openapi": "smithy build",
     "clean-openapi": "smithy clean",
-    "build-rs": "openapi-generator-cli generate -i ../../.generated-sources/smithy/openapi/openapi/Emily.openapi.json -g rust -c rust-config.json -o ../../.generated-sources/emily",
+    "build-rs": "npx @openapitools/openapi-generator-cli generate -i ../../.generated-sources/smithy/openapi/openapi/Emily.openapi.json -g rust -c rust-config.json -o ../../.generated-sources/emily",
```
AshtonStephens, line 6: This is necessary to prevent people from needing to install the `openapi-generator-cli` locally.

devenv/aws-setup/initialize.py
```diff
@@ -0,0 +1,72 @@
+import boto3
```
AshtonStephens, line 1: I intentionally didn't add tests to this file:
1. To add testing there would need to be another python file which would bloat the `devenv` code
2. The biggest pitfall of this will be if `cdk` creates a template that changed formats, at which point we'll need to significantly change this file anyway. We can't be resilient to that, but also it shouldn't change unless there's a full new `aws-cdk` version.

devenv/aws-setup/initialize.py
```diff
@@ -0,0 +1,72 @@
...
+
+# Setup DynamoDB Client.
+client = boto3.client('dynamodb', endpoint_url=dynamodb_endpoint)
+
+# Read CDK Template.
+with open(input_template_file) as fp:
+    template = json.load(fp)
+
+# Create convenience mapping.
+template_resource_ids_for_type = defaultdict(list)
```
AshtonStephens, line 23: The resource ids used as a key for a resource in the template is actually a unique token that is associated with the resource for within the template, but there's a separate AWS resource id, which is why this refers to `template_resource_id`.

`template_resource_token` could be a better name, but it is an identifier.

emily/cdk/lib/emily-stack-utils.ts
```diff
@@ -40,9 +41,11 @@ export class EmilyStackUtils {
      */
     public static getAwsAccount(): string {
         this.awsAccount ??= (process.env.AWS_ACCOUNT ?? process.env.CDK_DEFAULT_ACCOUNT);
-        if (this.awsAccount === undefined) {
+        if (this.awsAccount === undefined && this.getStageName() !== Constants.LOCAL_STAGE_NAME) {
             throw new Error('Must define AWS account on either "AWS_ACCOUNT" or "CDK_DEFAULT_ACCOUNT" env variables.');
         }
+        // If the account is undefined and we're letting it slide set the account to 12 "x"s.
```
AshtonStephens, line 47: The local environment needs this to be set, but it doesn't matter what it is. Someone who hasn't set up AWS locally shouldn't need to to run the integration tests if it's unnecessary.

emily/api-definition/package.json
AshtonStephens: @hstove FYI, feel a bit bad about reversing this change especially since I approved it.

devenv/aws-setup/initialize.py
netrome: Fair enough. However, the file would still be much more readable if it was structured in a more testable way, even though we don't have to necessarily test it.

I'd love to see functions defined for most of the tasks, and having all of the functions being called from a top-level main() function residing in a `if __name__ == '__main__'` block.

That would make the code more easy to follow, and experiment with if I'd like to import the module from a shell and call a single function for example.

Example:
```python
def main():
    client = boto3.client(...)
    template, template_resource_ids_for_type = read_cdk_template()
    setup_tables(...)
    setup_functions(...)
    output_cdk_template(...)
    
if __name__ == "__main__":
    main()
```

What would you think of a structure like that?

docker-compose.emily.yml
```diff
@@ -0,0 +1,73 @@
...
+    depends_on:
+      dynamodb:
+        condition: service_started
+    volumes:
+      - "./devenv/aws-setup/initialize.py:/initialize.py"
+      - "./emily/cdk/cdk.out:/cdk.out"
+    environment:
+      - AWS_DEFAULT_REGION=us-west-2
+      - AWS_ACCESS_KEY_ID=xxxxxxxxxxxx
+      - AWS_SECRET_ACCESS_KEY=xxxxxxxxxxxx
```
netrome, line 26: I assume these placeholders doesn't do anything. Can we omit them completely?

Makefile
```diff
@@ -1,50 +1,87 @@
...
+emily-integration-test: devenv $(EMILY_LAMBDA_BINARY) $(EMILY_CDK_TEMPLATE) $(EMILY_DOCKER_COMPOSE)
+	docker compose --file docker-compose.emily.yml up --remove-orphans
+.PHONY: emily-integration-test
 
-# Overwrite the default architecture with --arm64 configuration
-# on arm machines.
-ifneq (filter arm64 aarch64, $(shell uname -m),)
+# Builds all dockerfiles that need to be built for the dev environment.
+devenv: $(wildcard $(subst dir, devenv, $(TWO_DIRS_DEEP)))
+	sudo docker compose -f docker-compose.emily.yml build
```
netrome, line 68: I would prefer not having our build scripts require superuser privileges.

```suggestion
	docker compose -f docker-compose.emily.yml build
```

Makefile
```diff
@@ -1,50 +1,87 @@
...
-EMILY_PATH=emily
+EMILY_DEVENV_PATH=emily/devenv
+EMILY_LAMBDA_PATH=emily/lambda
 EMILY_API_PROJECT_NAME=emily-api
 EMILY_CDK_PROJECT_NAME=emily-cdk
 
-emily-cdk: emily-client | emily-lambda
-	pnpm --filter $(EMILY_CDK_PROJECT_NAME) run build
+# Launches Emily dev environment.
+emily-integration-test: devenv $(EMILY_LAMBDA_BINARY) $(EMILY_CDK_TEMPLATE) $(EMILY_DOCKER_COMPOSE)
```
netrome, line 62: I tried running this locally, but my dynamodb container seems to get stuck on this error

```
Attaching to apigateway-1, aws-setup-1, dynamodb-1
dynamodb-1    | Initializing DynamoDB Local with the following configuration:
dynamodb-1    | Port:	8000
dynamodb-1    | InMemory:	false
dynamodb-1    | Version:	2.4.0
dynamodb-1    | DbPath:	./data
dynamodb-1    | SharedDb:	true
dynamodb-1    | shouldDelayTransientStatuses:	false
dynamodb-1    | CorsParams:	null
dynamodb-1    |
dynamodb-1    | May 16, 2024 5:30:14 PM com.almworks.sqlite4java.Internal log
dynamodb-1    | WARNING: [sqlite] cannot open DB[1]: com.almworks.sqlite4java.SQLiteException: [14] unable to open database file
dynamodb-1    | May 16, 2024 5:30:14 PM com.almworks.sqlite4java.Internal log
dynamodb-1    | SEVERE: [sqlite] SQLiteQueue[shared-local-instance.db]: error running job queue
dynamodb-1    | com.almworks.sqlite4java.SQLiteException: [14] unable to open database file
dynamodb-1    | 	at com.almworks.sqlite4java.SQLiteConnection.open0(SQLiteConnection.java:1480)
dynamodb-1    | 	at com.almworks.sqlite4java.SQLiteConnection.open(SQLiteConnection.java:282)
dynamodb-1    | 	at com.almworks.sqlite4java.SQLiteConnection.open(SQLiteConnection.java:293)
dynamodb-1    | 	at com.almworks.sqlite4java.SQLiteQueue.openConnection(SQLiteQueue.java:464)
dynamodb-1    | 	at com.almworks.sqlite4java.SQLiteQueue.queueFunction(SQLiteQueue.java:641)
dynamodb-1    | 	at com.almworks.sqlite4java.SQLiteQueue.runQueue(SQLiteQueue.java:623)
dynamodb-1    | 	at com.almworks.sqlite4java.SQLiteQueue.access$000(SQLiteQueue.java:77)
dynamodb-1    | 	at com.almworks.sqlite4java.SQLiteQueue$1.run(SQLiteQueue.java:205)
dynamodb-1    | 	at java.base/java.lang.Thread.run(Thread.java:840)
```

Anything you recognize?

netrome (changes requested):
I don't fully follow everything that is happening here, but generally it seems to look alright. The only showstopper for me is calling sudo from within the makefile. Would it be possible to skip that?

Makefile
AshtonStephens: Yep, that only happens to suckers on Arch linux, sorry man

AshtonStephens: Trying to fix it and planned to put that in a follow on PR for other architectures #74, there's another error you'll uniquely get too once you fix that one. You can remove the shared database file and it'll be okay but then you're likely to get a docker issue from the SAM container.

@djordon I'm curious if you get this, I assume you won't.

AshtonStephens: I should have noted that actually, I intended for this to be enough but it's very much not and not even accurate, my bad:
> Note that this may still not work for some machines with weird docker setups.

Makefile
AshtonStephens: SGTM

docker-compose.emily.yml
AshtonStephens: Nope, they need to be set unfortunately

devenv/aws-setup/initialize.py
AshtonStephens: I disagree on this one, but not enough to fight it if someone else agrees with you.

Python is ultimately a scripting language that ended up getting a lot of users because of how easy it is to use and the libraries that support it.

This here is just a bash script written in python; you wouldn't expect there to be a `main` in a bash script, and I would argue you wouldn't want there to be a `main` here because it's a one off script and should be structured like one. 

If this were the lambda written in python or a test suite I'd be 1000% with you, but I think the for the purpose this script has it's actually better to have this clearly be a script and not a program.

Makefile
netrome: SGTM?

docker-compose.emily.yml
netrome: Nice :+1:

Makefile
netrome: Hahah, my machine is definitely weird but docker is one of the things that I have working properly :sweat_smile:

netrome: > Trying to fix it and planned to put that in a follow on PR for other architectures https://github.com/stacks-network/sbtc/issues/74, there's another error you'll uniquely get too once you fix that one. You can remove the shared database file and it'll be okay but then you're likely to get a docker issue from the SAM container.

Okay yeah that sucks. And you're running this on a non-Arch machine?

Makefile
AshtonStephens: sounds good to me

Makefile
AshtonStephens: Yes but don't worry about it let me fix it, I don't plan on leaving you out to dry

AshtonStephens: I'll add it to this PR, probably better I don't segment it

djordon (commented):
Looks good, I did have one question though.

Makefile
djordon: Yeah, I worked perfectly for me 😄

devenv/aws-setup/initialize.py
djordon: My 2 cents: being able to reason about separate operations is the important part. You have comments that help with that, so it's reasonably clear already, but it would be better to separate those as functions. I'd say the same thing about structuring bash scripts (although for bash it's largely aspirational, I sometimes fail to follow it myself). And I don't care about the `if __name__ == '__main__':` block one way or the other.

netrome: > Python is ultimately a scripting language that ended up getting a lot of users because of how easy it is to use and the libraries that support it.

I'd argue that a lot of the success of Python comes from their philosophy which strongly emphasizes code readability. Yes this is "just a bash script", but when Python gives us the ability to structure the code in a readable and more maintainable way at basically no cost - why not do it?

Given that code is read more times than it's written, optimizing our code for readability will save us time in the long run.

netrome: > If this were the lambda written in python or a test suite I'd be 1000% with you, but I think the for the purpose this script has it's actually better to have this clearly be a script and not a program.

I'm curious about your reasoning here. Even for a script, what are the benefits of writing everything as a flat sequence of instructions as opposed to organize it into functions?

Makefile
netrome: Aaah, obviously :sweat_smile:

Makefile
netrome: Sweet, thanks! :pray:

netrome: Let me know when I should try again.

AshtonStephens: It'll be sometime tomorrow, today is GH project board day

emily/cdk/package.json
```diff
@@ -5,16 +5,15 @@
     "emily": "bin/emily.js"
   },
   "scripts": {
-    "build": "(cd ../api-definition && npm run build-openapi) && tsc",
+    "synth": "tsc && npx aws-cdk synth -q --no-staging",
```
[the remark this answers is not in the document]
AshtonStephens: The original `"build"` did something a little misleading.

It would:
1. make the openapi template from the `api-description` sibling directory
1. Compile the typescript to javascript

That just builds the sources, but it doesn't build any `cdk` artifact. Normally you'd run `npm run build` and then run some command like `cdk deploy` or `cdk synth` that utilize the generated javascript.`cdk deploy` synthesizes the template and then deploys the artifacts to AWS, and the `cdk synth` command just makes the template.

I've changed this so that it'll take the `typescript` sources and turn them into a the cdk template right off the bat and assumes that the openapi template has already been generated. This is a better separation of responsibilities given the new structure of the Makefile. In the build command before there was no usable output (no template) and it modified sources even when it didn't need to be rebuilt.

I also changed the name to `synth` because it better matches the cdk idioms.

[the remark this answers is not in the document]
djordon: I see. Thank you for the explanation, I did not know that.

So going forward, when we do `make build` it would not build the cdk template files anymore. That would get done if we run integration tests, setup a dev env (but with local configuration), or when/if we add some commands to deploy.

djordon (approved):
Looks good!

docker-compose.emily.yml
```diff
@@ -0,0 +1,70 @@
+services:
+
+  # Hosts the DynamoDB Tables.
+  dynamodb:
+    image: "amazon/dynamodb-local:latest"
+    command: "-jar DynamoDBLocal.jar -sharedDb -dbPath ./data"
+    ports:
+      - "8000:8000"
+    volumes:
+      - "./devenv/dynamodb/data:/home/dynamodblocal/data"
```
djordon, line 10: Oh I noticed that a database file shows up here as `devenv/dynamodb/data/shared-local-instance.db`. I think we should `.gitignore` it.

.generated-sources/emily/build.rs
```diff
@@ -61,6 +23,10 @@ fn main() -> io::Result<()> {
         file.write_all(new_contents.as_bytes())?;
     }
 
-    // Return
+    // Ensure that we rerun the lib.rs was altered by the client autogeneration.
```
netrome, line 26:
```suggestion
    // Ensure that we rerun if the lib.rs was altered by the client autogeneration.
```

AshtonStephens: Done with a rebase.

docker-compose.emily.yml
AshtonStephens: Done.

Makefile
netrome: I seem to have better luck with this now. However `service-test` seems to fail for me on both `curl` and `jq` not found.

netrome: Tried again and now it worked.

netrome (approved):
I'd love to do a workshop/lab to go over everything here. I can spin things up locally now, but if I try to curl Emily I only get server errors.

```
curl http://localhost:3000/deposits | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    36  100    36    0     0      2      0  0:00:18  0:00:12  0:00:06     9
{
  "message": "Internal server error"
}
```

with the logs
```
Timed out while attempting to establish a connection to the container. You can increase this timeout by setting the SAM_CLI_CONTAINER_CONNECTION_TIMEOUT environment variable. The current timeout is 10.0 (seconds).
```

I'm probably just missing something here. Looking forward to learning more about this!

AshtonStephens:
> I'd love to do a workshop/lab to go over everything here. I can spin things up locally now, but if I try to curl Emily I only get server errors.
> 
> ```
> curl http://localhost:3000/deposits | jq
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 100    36  100    36    0     0      2      0  0:00:18  0:00:12  0:00:06     9
> {
>   "message": "Internal server error"
> }
> ```
> 
> with the logs
> 
> ```
> Timed out while attempting to establish a connection to the container. You can increase this timeout by setting the SAM_CLI_CONTAINER_CONNECTION_TIMEOUT environment variable. The current timeout is 10.0 (seconds).
> ```
> 
> I'm probably just missing something here. Looking forward to learning more about this!

This is what I talked about on the call earlier; there's an issue on Linux computers that I wasn't able to resolve, and I've had very poor luck debugging it. There aren't any tests in CI and most of us are using MacOS which makes me consider this temporarily okay, but I need to find a resolution. I plan to track that in #74

The SAM docker container which I've named `apigateway` is running the [`sam cli`](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/install-sam-cli.html) which simulates a lambda by spinning the code up in a docker container that emulates an AWS lambda. Normally you'd run `sam` from your local command line but for a full dev environment it's better to have it as a container.

By default the SAM cli will create the lambda instance as a child container of the one running the cli, but since we have the cli running in its own docker container it needs to make a sibling container as opposed to a child container. There are a number of differences between MacOS and Linux when it comes to docker, and I think I've caught and handled most of them, but for whatever reason when the `sam` cli launches the sibling lambda container on MacOS the cli can connect to it, but it cannot connect to it on Linux. 

If the sam cli is running outside of a container on Linux it will work, so I believe it's safe to say it's an issue with the way the network is set up, but I've tried a lot of things and nothing seems to work:
- External docker network for the compose.
- User permissions on the bootstrap binary.
- Variations of host IP config options in the `sam local-api start` command.

netrome:
> > I'd love to do a workshop/lab to go over everything here. I can spin things up locally now, but if I try to curl Emily I only get server errors.
> > ```
> > curl http://localhost:3000/deposits | jq
> >   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
> >                                  Dload  Upload   Total   Spent    Left  Speed
> > 100    36  100    36    0     0      2      0  0:00:18  0:00:12  0:00:06     9
> > {
> >   "message": "Internal server error"
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > with the logs
> > ```
> > Timed out while attempting to establish a connection to the container. You can increase this timeout by setting the SAM_CLI_CONTAINER_CONNECTION_TIMEOUT environment variable. The current timeout is 10.0 (seconds).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm probably just missing something here. Looking forward to learning more about this!
> 
> This is what I talked about on the call earlier; there's an issue on Linux computers that I wasn't able to resolve, and I've had very poor luck debugging it. There aren't any tests in CI and most of us are using MacOS which makes me consider this temporarily okay, but I need to find a resolution. I plan to track that in #74
> 
> The SAM docker container which I've named `apigateway` is running the [`sam cli`](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/install-sam-cli.html) which simulates a lambda by spinning the code up in a docker container that emulates an AWS lambda. Normally you'd run `sam` from your local command line but for a full dev environment it's better to have it as a container.
> 
> By default the SAM cli will create the lambda instance as a child container of the one running the cli, but since we have the cli running in its own docker container it needs to make a sibling container as opposed to a child container. There are a number of differences between MacOS and Linux when it comes to docker, and I think I've caught and handled most of them, but for whatever reason when the `sam` cli launches the sibling lambda container on MacOS the cli can connect to it, but it cannot connect to it on Linux.
> 
> If the sam cli is running outside of a container on Linux it will work, so I believe it's safe to say it's an issue with the way the network is set up, but I've tried a lot of things and nothing seems to work:
> 
>     * External docker network for the compose.
> 
>     * User permissions on the bootstrap binary.
> 
>     * Variations of host IP config options in the `sam local-api start` command.

Ah I see. Thank you for the explanation, and nice that you created a ticket to track this. Then it also sounds like I could work around this for the time being by using the sam CLI on my machine instead of in docker.

radicleart:
> If the sam cli is running outside of a container on Linux it will work, so I believe it's safe to say it's an issue with the way the network is set up, but I've tried a lot of things and nothing seems to work:

Did you try running the container with network host - as workaround for `host.docker.internal` on linux for calling host services ?

```
docker run --network host <your-docker-image>
```

djordon:
> > If the sam cli is running outside of a container on Linux it will work, so I believe it's safe to say it's an issue with the way the network is set up, but I've tried a lot of things and nothing seems to work:
> 
> Is the problem calling host services? If so running the container with network host flag on linux might work?
> 
> ```
> docker run --network host <your-docker-image>
> 
> // or in compose
> network_mode: host
> ```

I was going to suggest trying this:
```yaml
    extra_hosts:
      - "host.docker.internal:host-gateway"
```
But I'm not sure about the differences between the two.

AshtonStephens:
Note: @netrome @djordon I made two changes after merging with the latest change from the blocklist client:

1. `Makefile` change to use `Makefile` caching on the blocklist client.
2. GitHub action change to only run the required `Make` steps.

This PR ended up changing more than I wanted because it introduced a new `make` paradigm and that meant that the new changes that used the old paradigm needed to be altered.

Makefile
```diff
@@ -1,64 +1,111 @@
...
+# Example:
+# $(subst dir, emily/cdk/lib, $(THREE_DIRS_DEEP))
+# becomes
+# emily/cdk/lib/*  emily/cdk/lib/*/*  emily/cdk/lib/*/*/*
+THREE_DIRS_DEEP := dir/* dir/*/* dir/*/*/*
+TWO_DIRS_DEEP := dir/* dir/*/*
+
+AUTOGENERATED_BLOCKLIST_CLIENT_CLIENT := .generated-sources/blocklist-api/src/lib.rs
+BLOCKLIST_OPENAPI_PATH=./.generated-sources/blocklist-openapi-gen
+BLOCKLIST_OPENAPI_SPEC=$(BLOCKLIST_OPENAPI_PATH)/blocklist-client-openapi.json
```
AshtonStephens, line 13: New.

Makefile
```diff
@@ -1,64 +1,111 @@
...
+	pnpm --prefix $(BLOCKLIST_OPENAPI_PATH) run build
 
 # Generate the OpenAPI spec for Blocklist Client
-generate-openapi-spec: install
+$(BLOCKLIST_OPENAPI_SPEC): $(INSTALL_TARGET) $(filter-out $(BLOCKLIST_OPENAPI_SPEC), $(wildcard $(subst dir, $(BLOCKLIST_OPENAPI_PATH), $(THREE_DIRS_DEEP))))
 	cd ./.generated-sources/blocklist-openapi-gen && cargo build
-
-# Generate the client code using the OpenAPI spec
-blocklist-api-source: generate-openapi-spec
-	pnpm --prefix $(BLOCKLIST_OPENAPI_PATH) run build
```
AshtonStephens, line 64: New.

.github/workflows/unit-tests.yaml
```diff
@@ -40,13 +40,10 @@ jobs:
...
-      - name: Generate OpenAPI spec for blocklist client
-        run: make generate-openapi-spec
+      - name: Run Basic tests
+        run: make test
 
-      - name: Generate blocklist API client
-        run: make blocklist-api-source
-
-      - name: Run tests
+      - name: Run Integration tests
```
AshtonStephens, line 46: New.

AshtonStephens (commented):
Highlighed the new things.

Makefile
```diff
@@ -1,64 +1,111 @@
...
+EMILY_DEVENV_PATH=emily/devenv
+EMILY_LAMBDA_PATH=emily/lambda
 EMILY_API_PROJECT_NAME=emily-api
 EMILY_CDK_PROJECT_NAME=emily-cdk
+CONTAINER_HOST=host.docker.internal
+
+ifeq ($(findstring Linux, $(shell uname)), Linux)
+_CONTAINER_HOST := localhost
+else
+_CONTAINER_HOST := host.docker.internal
```
djordon, line 70: Oh interesting, I had thought the `host.docker.internal` path would work on both macOS and linux. Good to know.

.github/workflows/unit-tests.yaml
```diff
@@ -40,13 +40,10 @@ jobs:
       - name: Install package dependencies.
         run: make install
 
-      - name: Generate OpenAPI spec for blocklist client
-        run: make generate-openapi-spec
+      - name: Run Basic tests
+        run: make test
```
djordon, line 44: Right now the `make integration-tests` command runs the basic tests as well. We can either remove this or update the stanza in the yaml or update the command. I'm okay with either.

djordon (approved):
I think our CI is doing `cargo test` twice now but besides that it looks good.

.github/workflows/unit-tests.yaml
AshtonStephens: I'll remove the test command then push it.

AshtonStephens:
I see the suggestions on how to fix the issues on Linux computers and will use them in #74, right now just trying to get this through.

Issue #84, opened 2024-04-30
Clarity Signer Key Rotation


setzeus:
## 1. Summary
This issue is also related to kickoff to #55 & is a continuation of #83 . It covers the implementation of the on-chain key rotation feature for the bootstrap signer set.

## 2. Context & Purpose
The purpose of this issue is to create a way for signer set to rotate keys in case of a key compromise.

## 3. Deliverables
- [x] Implement 'rotate-key' function
- [x] Implement 'update-aggregate-pubkey'

setzeus:
Closed with #151

Issue #83, opened 2024-04-30
Clarity Signer Contract Setup


setzeus:
## 1. Summary
This issue is a kickoff issue related to #55. The goal here is simply to setup the 'sbtc-bootstrap-signers' Clarity contract by creating the file & standing up all of the structs defined in the [design document](https://docs.google.com/document/d/1Bwi1jM_7p9OqNGT41eWCuoKnfoOrNHycqxby3XSonCA/edit#heading=h.tahn8ddco07r).

## 2. Context & Purpose
The purpose of this issue is to standup the signer-required Clarity work so that in the next issues we can implement the appropriate functions.

## 3. Deliverables
- [x] Add 'sbtc-bootstrap-signer' contract to the Clarity protocol
- [x] Implement the constants, variables & maps

PR #151, opened 2024-05-14
Bootstrap Signer Setup

setzeus:
## Description
This PR is a draft PR meant to solve both #83 & #84 by introducing both the 'sbtc-bootstrap-signers' contract along with the most important function within (rotate-keys ...).

This also involves updating the '.registry' contract to properly update the signer set, current multi-sig principal & aggregate signing key.

## Type of Change
Setup & core logic.

No.

## Are documentation updates required?
Not yet, but the last setup PR should update the ReadME with all contract error codes.

   The best way to find these is by searching inside the docs at https://github.com/blockstack/docs
-->
- [ ] Link to documentation updates: 

## Testing information
No tests at the moment, partially the reason this is in draft status.

## Checklist
- [x] `pnpm test` passes
- [ ] Changelog is updated
- [x] Tag 1 of @person1 or @person2


contracts/contracts/sbtc-registry.clar
```diff
@@ -44,10 +48,16 @@
...
 )
 
-;; Read-only functions
+;; Data structure to store aggregate pubkey,
+;; stored to avoid replay
+(define-map aggregate-pubkeys (buff 32) bool)
 
+;; Data structure to store the current signer set,
+;; stored to avoid replay
+(define-map multi-sig-address principal bool)
```
hstove, line 57: I think my recollection was that we'd only want to prevent aggregate key replays? Although I don't really think this is an issue. The only downside is that this says "you can't rotate back to an older set of keys", which seems technically valid, but maybe that's more of a good thing to prevent it.

contracts/contracts/sbtc-bootstrap-signers.clar
```diff
@@ -20,11 +17,43 @@
 ;;
 
 ;; public functions
-;;
+;; Rotate keys
+;; Used to rotate the keys of the signers. This is called whenever
+;; the signer set is updated.
+(define-public (rotate-keys-wrapper (new-keys (list 15 (buff 32))) (multi-sig-address principal) (new-aggregate-pubkey (buff 32)))
```
hstove, line 23: I think in the Clarity WG call we mentioned that it's better to construct the multisig principal prove the public keys, instead of accepting a param. That way, we can be 100% sure that the keys and the multisig match.

Also totally fine to add a TODO here and open a new issue for that related code and tests

contracts/contracts/sbtc-bootstrap-signers.clar
```diff
@@ -1,17 +1,14 @@
...
+;; sBTC Bootstrap Signers contract
 
 ;; constants
-;;
+(define-constant signature-threshold u8)
+(define-constant key-size u32)
+
+;; errors
+;; if err is u200, it's the agg key
+;; if err is u210>, it's the key at index (err - 210)
```
hstove, line 9: This is probably fine, but we'd likely want to refactor if this contract ends up having more than 10 error codes

contracts/contracts/sbtc-deposit.clar
```diff
@@ -14,11 +14,9 @@
...
 
 ;; public functions
-
 ;; Accept a new deposit request
 ;; Note that this function can only be called by the current
 ;; bootstrap signer set address - it cannot be called by users directly.
-;; 
 ;; This function handles the validation & minting of sBTC, it then calls
 ;; into the sbtc-registry contract to update the state of the protocol
 (define-public (complete-deposit-wrapper (txid (buff 32)) (vout-index uint) (amount uint) (recipient principal))
```
hstove, line 22: I think it's actually better to revert this change. When you run `pnpm clarigen docs`, which generates markdown docs, this will make the doc comment for this function be

```md
public functions Accept a new deposit ...
```

Same with the "newline" - when converted to markdown it'll start a new paragraph, which I think is helpful.

contracts/contracts/sbtc-registry.clar
```diff
@@ -44,10 +48,16 @@
...
+
+;; Read-only functions
```
hstove, line 59: Same thing here with doc comments. 

I know we don't have an internal process for generating contract documentation yet, so I'm fine with punting this and I can open a ticket to clean up doc comments and also setup a process for auto-generated documentation.

contracts/contracts/sbtc-registry.clar
```diff
@@ -138,6 +157,29 @@
...
+;; Rotate the signer set, multi-sig principal, & aggregate pubkey
+;; This function can only be called by the bootstrap-signers contract.
+(define-public (rotate-keys (new-keys (list 15 (buff 32))) (new-address principal) (new-aggregate-pubkey (buff 32)))
+  (begin
+    ;; Check that caller is protocol contract
+    (try! (validate-caller))
+    ;; Check that the aggregate pubkey is not already in the map
+    (asserts! (is-none (map-get? aggregate-pubkeys new-aggregate-pubkey)) ERR_AGG_PUBKEY_REPLAY)
+    ;; Check that the new address (multi-sig) is not already in the map
+    (asserts! (is-none (map-get? multi-sig-address new-address)) ERR_MULTI_SIG_REPLAY)
```
hstove, line 169: This can be converted to:

```clarity
(asserts! (map-insert aggregate-pubkeys new-aggregate-pubkey) ERR_AGG_PUBKEY_REPLAY)
```

And then we don't need two lines for both checking replay and setting the value

contracts/tests/sbtc-bootstrap-signers.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  import { txOk, txErr } from "@clarigen/test";
+  
+  describe("sBTC bootstrap signers contract", () => {
+    describe("Rotate keys tests", () => {
+      test("Rotate keys wrapper correctly", () => {
+        const receipt = txOk(
+          signers.rotateKeysWrapper({
+            newKeys: [new Uint8Array(32).fill(0),new Uint8Array(32).fill(0)],
+            multiSigAddress: alice,
+            newAggregatePubkey: new Uint8Array(32).fill(0),
```
hstove, line 15: Yeah, now that I'm seeing that we'd also want JS code helpers for generating the multi-sig address, let's punt on automatically calculating it and open a ticket

contracts/tests/sbtc-bootstrap-signers.test.ts
```diff
@@ -0,0 +1,46 @@
...
+          }),
+          alice
+        );
+        expect(receipt.value).toEqual(true);
```
hstove, line 19: Add tests here for validating that the new keys, aggregate key, and multi-sig is stored correctly

hstove (changes requested):
Looks great, mainly would like to see a few more assertions.

Also, can you manually delete `emily/api-definitions/package-lock.json`? I fixed the issue that kept re-creating it in https://github.com/stacks-network/sbtc/pull/156

contracts/contracts/sbtc-deposit.clar
setzeus: Ehhh, I personally don't like the new-line / extra spacing. Nothing to stand my ground but it looks odd to have separated multiple line comments for a single function.

contracts/contracts/sbtc-registry.clar
setzeus: Ah, shoot, yeah, overlooked that, thanks.

contracts/contracts/sbtc-registry.clar
setzeus: Right. I do remember confirming only agg-key but when writing it out I thought the multi-sig made sense as well.

contracts/contracts/sbtc-bootstrap-signers.clar
setzeus: Copy. Will open a TODO but then I'm confused on whether the keys will ever actually be stored in the contract? Is your thinking that signer keys should NOT be stored here?? Only a multi-sig principal?

contracts/contracts/sbtc-bootstrap-signers.clar
setzeus: Copy, good point out. Will leave as-is but will change to "220"+ if we do have more than 10.

contracts/contracts/sbtc-registry.clar
setzeus: Copy let's punt on documentation formatting.

contracts/tests/sbtc-bootstrap-signers.test.ts
setzeus: Completed.

contracts/contracts/sbtc-registry.clar
setzeus: Completed.

contracts/contracts/sbtc-bootstrap-signers.clar
```diff
@@ -20,11 +17,43 @@
...
+    (match helper-response
+        ok-response
+            (begin 
+                (asserts! (is-eq (len current-key) key-size) (err (+ ERR_KEY_SIZE_PREFIX (+ u10 (get index ok-response)))))
+                (ok {index: (+ (get index ok-response) u1)})
+            )
+        err-response
+            (err err-response)
+    )
+)
```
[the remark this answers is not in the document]
setzeus: Yeap, agreed. I had another field used in a previous version, with only one field the tuple no longer makes sense.

Updated.

contracts/contracts/sbtc-bootstrap-signers.clar
hstove: I think both, as the signers will use the keys as their "source of truth", but just that the multi-sig will be computed instead of taken as an argument

hstove (approved):
Nice work!

Issue #152, opened 2024-05-14
[Feature]: Simplify peg-in transaction signing


djordon:
# Feature - Simplify peg-in transaction signing

## 1. Description

Deposit transactions have taproot UTXOs with no key-spend path. This makes signing them in them less straightforward during the peg-in transaction. Write a function to make constructing the correct digest to be signed.

### 2.1 Acceptance Criteria:
   - [ ] The function should give an digest which, once signed and added to the transaction, will be accepted by the bitcoin network as a valid transaction.

PR #155, opened 2024-05-15
feat: simplify peg-in transaction signing

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/152.

Each transaction input needs to be accompanied by a signed digest before being submitted to the bitcoin network. It is not obvious which steps need to be taken to construct this digest, so this PR adds helper functions that simplifies their creation.

## Testing information

Provide context on how tests should be performed.

1. There is already an integration test that constructs a transaction, signs it, and submits it to the bitcoin network. This PR updates that test to use the new helper functions. If bitcoind accepts our transaction, then things are likely constructed correctly.

### DIFF
--- a/signer/src/utxo.rs
+++ b/signer/src/utxo.rs
@@ -1,6 +1,8 @@
 use bitcoin::absolute::LockTime;
 use bitcoin::key::Secp256k1;
 use bitcoin::secp256k1::SECP256K1;
+use bitcoin::sighash::Prevouts;
+use bitcoin::sighash::SighashCache;
 use bitcoin::taproot::LeafVersion;
 use bitcoin::taproot::NodeInfo;
 use bitcoin::taproot::Signature;
@@ -11,6 +13,9 @@ use bitcoin::Amount;
 use bitcoin::OutPoint;
 use bitcoin::ScriptBuf;
 use bitcoin::Sequence;
+use bitcoin::TapLeafHash;
+use bitcoin::TapSighash;
+use bitcoin::TapSighashType;
 use bitcoin::Transaction;
 use bitcoin::TxIn;
 use bitcoin::TxOut;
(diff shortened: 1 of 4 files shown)


signer/src/utxo.rs
```diff
@@ -370,6 +413,62 @@ impl<'a> UnsignedTransaction<'a> {
...
+    /// the transaction.
+    ///
+    /// # Notes
+    ///
+    /// This function uses the fact certain invariants about this struct are
+    /// upheld. They are
+    /// 1. The first input to the Transaction in the `tx` field is the signers'
+    ///    UTXO.
+    /// 2. The other inputs to the Transaction in the `tx` field are ordered
+    ///    the same order as DepositRequests in the `requests` field.
```
netrome, line 426: Nice that you outline this. I wish there was a simple way to enforce this invariant in the type directly, but it does not seem trivial.

signer/tests/integration/utxo_construction.rs
```diff
@@ -219,33 +210,22 @@ fn deposits_add_to_controlled_amounts() {
...
-    tx.input[0].witness = Witness::p2tr_key_spend(&signature1);
-    tx.input[1].witness = requests.deposits[0].construct_witness_data(signature2);
+    unsigned
+        .tx
+        .input
+        .iter_mut()
+        .zip(witness_data)
+        .for_each(|(tx_in, witness)| {
+            tx_in.witness = witness;
+        });
```
netrome, line 225: Is there a use case for creating witness data that doesn't entail joining it with the unsigned transaction input? Otherwise, it feels to me like this functionality could be baked into the `regtest::p2tr_witness` helper function to make the code a bit cleaner in the test here.

netrome (approved):
Nice stuff. Just one question.

signer/src/utxo.rs
djordon: I thought about it, and yeah, it's not obvious how to do so.

signer/tests/integration/utxo_construction.rs
djordon: My original approach attempted to do it in that function, but I was stymied by the borrow checker. There may be one more approach that can work. I'll try again quickly, to see if it can work.

djordon: Oh, it was easy to do.

netrome: Awesome, thanks for updating! :pray:

Issue #64, opened 2024-04-24
Implement function for retrieving the current BTC market fee rate


djordon:
This is part of https://github.com/Trust-Machines/sbtc-v1/issues/52.

Implement the `get_market_rate` function that retrieves the current market rate. This function was outlined [here](https://docs.google.com/document/d/1iQ-OCYqQL31JVb_fhBoHYj2Ay0drBfmypa6HDZa9lVU/edit#heading=h.1p84sqsw60fa), with further specification in that same document [here](https://docs.google.com/document/d/1iQ-OCYqQL31JVb_fhBoHYj2Ay0drBfmypa6HDZa9lVU/edit#heading=h.7eihy25q7xpn).

Issue #160, opened 2024-05-16
[Feature]: Add electrum as a data source for fee estimates


djordon:
# Feature - Add electrum as a data source for fee estimates

### 1.1 Context & Purpose

This ticket is a continuation of https://github.com/stacks-network/sbtc/issues/64, where we added a function to get the market fee rate. The LLD settled on using Electrum server as one of the sources for this information.

We can also add in bitcoin-core as an optional source as well by calling the [`estimatesmartfee`](https://developer.bitcoin.org/reference/rpc/estimatesmartfee.html) RPC on a bitcoin node.

### 2.1 Acceptance Criteria:
   - [ ] If given an electrum host, be able to fetch the market fee rate using its RPC endpoint.
   - [ ] Have an integration test.

PR #161, opened 2024-05-16
feat: retrieve estimates for the going BTC market fee rate

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/64

We need to appropriately set the fee rate for out BTC transactions so that get promptly confirmed. In order to do that, we need to retrieve estimates from some third party source. This PR adds a function that fetches estimates for an appropriate fee rate and returns the average.

### Notes
1. We settled on using Electrum server for estimates in the LLD, so that needs to be added as a source. I'll do that in a separate PR, see https://github.com/stacks-network/sbtc/issues/160.
2. I got these sources from https://github.com/stacks-network/stacks-core/blob/master/contrib/side-cars/fee-estimate.py, but I did not add `blockchain.info` (now `blockchain.com`?) as a source. I think we should have at least three, but I didn't like how I couldn't find documentation for it. 
 
## Testing information

This PR includes an integration test that hits the third party services that we'd like to hit in prod. Maybe we shouldn't do that?


signer/src/error.rs
```diff
@@ -3,6 +3,15 @@ pub enum Error {
...
 
+    #[error("Got an old fee estimate")]
+    OldFeeEstimate,
+
+    #[error("Failed to get fee estimates from all fee estimate sources")]
+    NoGoodFeeEstimates,
+
+    #[error("{0}")]
+    Reqwest(#[from] reqwest::Error),
+
```
netrome, line 14: I'm generally in favor of having multiple dedicated error types over a single large type encompassing everything that can ever go wrong across all modules. These variants are very unrelated to the `InvalidAmount` and `Taproot` error variants. So I would prefer to have a dedicated type in `fees.rs` with just these variants, which will make it easier for us to write proper error handling logic in the signer event loop once we start calling these functions.

signer/src/fees.rs
```diff
@@ -0,0 +1,355 @@
...
+    let sum_sats_per_vbyte = responses
+        .into_iter()
+        .filter_map(Result::ok)
+        .map(|x| x.sats_per_vbyte)
+        .sum::<f64>();
+
+    Ok(sum_sats_per_vbyte / num_responses as f64)
+}
+
+pub async fn estimate_fee_rate(client: &reqwest::Client) -> Result<f64, Error> {
```
netrome, line 174: Given that this call has side effects, I would like to have it behind a trait to make it easier to write tests depending on the functionality. In particular, I think it would be nice to reuse the `FeeEstimator` trait for this.

signer/src/fees.rs
```diff
@@ -0,0 +1,355 @@
...
+pub struct FeeEstimate {
+    pub sats_per_vbyte: f64,
+}
+
+pub trait FeeEstimator {
+    fn estimate_fee_rate(
+        &self,
+        client: &reqwest::Client,
+    ) -> impl Future<Output = Result<FeeEstimate, Error>> + Send;
+}
```
netrome, line 81: I don't think a fee estimator necessarily should have to use a client. I would rather see that as a field on the actual implementations. E.g. `struct MempoolSpaceFeeEstimator { base_url: ..., client: ... }`. This would allow for more flexibility in code depending on the trait. It would be easy to just implement a `ConstantFeeEstimator` for tests that doesn't need to do more fancy fee estimation.

Therefore I think the trait should look like this:

```suggestion
pub trait FeeEstimator {
    fn estimate_fee_rate(
        &self,
    ) -> impl Future<Output = Result<FeeEstimate, Error>> + Send;
}
```

(The receiver could also be `self` and not `&self` which would give the implementer even more flexibility, but this can also lead to some pain with trait bounds. It's not obvious to me what's best here right now)

signer/src/fees.rs
```diff
@@ -0,0 +1,355 @@
...
+    responses.retain(Result::is_ok);
+    let num_responses = responses.len();
+    let sum_sats_per_vbyte = responses
+        .into_iter()
+        .filter_map(Result::ok)
+        .map(|x| x.sats_per_vbyte)
+        .sum::<f64>();
+
+    Ok(sum_sats_per_vbyte / num_responses as f64)
+}
```
netrome, line 172: This feels like the functionality of some type of aggregate fee rate estimator. May I suggest implementing a `AverageFeeRate` type in the style of:

```
struct AverageFeeRate<T> {
    sources: Vec<T>, // This could be a [T; N] if we want to use const generics and skip the heap allocation
}

impl<T: FeeEstimator> FeeEstimator for AverageFeeRate<T> {
    async fn estimate_fee_rate(...) -> ...
}
```

signer/src/fees.rs
netrome: If we introduce the `AverageFeeRate` type as I suggested, we'd achieve this.

netrome (approved):
This is a great starting point. I have some comments on the external interface (suggestions for the `FeeEstimator` trait, and leveraging it more for the public interface), that I think would simplify our work with the signer when we begin depending on this logic. But nothing that we can't iterate upon in follow-ups, so feel free to pick and choose what you want to do now, what to save for later and what you think we should have further dialogues to iron out.

netrome:
> This PR includes an integration test that hits the third party services that we'd like to hit in prod. Maybe we shouldn't do that?

All tests I see seem to be using `mockito` for http mocking. I assume this is a stale comment?

djordon:
> > This PR includes an integration test that hits the third party services that we'd like to hit in prod. Maybe we shouldn't do that?
> 
> All tests I see seem to be using `mockito` for http mocking. I assume this is a stale comment?

No, the `prod_estimate_fee_rate_works` test does this.

netrome:
> > > This PR includes an integration test that hits the third party services that we'd like to hit in prod. Maybe we shouldn't do that?
> > 
> > 
> > All tests I see seem to be using `mockito` for http mocking. I assume this is a stale comment?
> 
> No, the `prod_estimate_fee_rate_works` test does this.

Oh right, thanks for pointing me to it. Yeah ideally I'd like to have the ability to run that test against locally hosted services so that our CI won't fail because services we don't control are down. That being said, I do see the value in also being able to run that test against the actual services. This CI configuration can definitely be done as a follow-up.

What do you think. Should we create a ticket for that CI task so we don't drop it?

signer/src/fees.rs
```diff
@@ -0,0 +1,355 @@
...
+
+pub async fn estimate_fee_rate(client: &reqwest::Client) -> Result<f64, Error> {
+    let fee_sources: [FeeSource; 2] = [
+        FeeSource::MempoolSpace(MempoolSpace {
+            base_url: "https://mempool.space".to_string(),
+        }),
+        FeeSource::BitcoinerLive(BitcoinerLive {
+            base_url: "https://bitcoiner.live".to_string(),
+        }),
```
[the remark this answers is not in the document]
djordon: Yeah, I agree. I think I even started writing code to do it and just decided to do keep this one simple and do it in a separate PR (the one that addresses https://github.com/stacks-network/sbtc/issues/160, but I don't mention it in that ticket).

signer/src/error.rs
djordon: > I'm generally in favor of having multiple dedicated error types over a single large type encompassing everything that can ever go wrong across all modules. 

I probably have a minority view here but I like dedicated error types that encompass everything that can ever go wrong across all modules[^1]. But the dedicated error approach is good, so I'm okay with not enforcing any particular approach here so long as we have enough information to debug when something bad happens.


[^1]: More specifically, I like to follow the approach laid out in [this blog post](https://kazlauskas.me/entries/errors). Basically, the author talks about his experience with one large error enum with enough variants to preserve context, and I think that's the way to go.

signer/src/fees.rs
netrome: Also, if you don't mind I'd like to follow the trait-as-verb convention and rename this to `EstimateFee`.
 
```suggestion
pub trait EstimateFee {
    fn estimate_fee_rate(
        &self,
        client: &reqwest::Client,
    ) -> impl Future<Output = Result<FeeEstimate, Error>> + Send;
}
```

signer/src/fees.rs
djordon: Yeah I like this approach

djordon: Okay, I briefly tried out the `AverageFeeRate` approach but ended up removing it. It felt like it added unnecessary complexity. Simplifying the trait felt like a win 🙏🏽.

djordon:
> Oh right, thanks for pointing me to it. Yeah ideally I'd like to have the ability to run that test against locally hosted services so that our CI won't fail because services we don't control are down. That being said, I do see the value in also being able to run that test against the actual services. This CI configuration can definitely be done as a follow-up.
>
> What do you think. Should we create a ticket for that CI task so we don't drop it?

Yeah, it will get address when we allow for fully configured hosts.

Issue #95, opened 2024-04-30
[Feature]: Withdrawal request Clarity functionality


hstove:
# Feature - `initiate-withdrawal-request` contract function

## 1. Description

In relation to #57 , this ticket involves the creation and testing of the `withdrawal-request` function.

## 2. Technical Details:

The details of what this function entails is described in the withdrawal contract LLD.

### 2.1 Acceptance Criteria:
   - [ ] `initiate-withdrawal-request` function created
   - [ ] Tests for various scenarios when using this function
   - [ ] Tests to ensure the right events are emitted

PR #168, opened 2024-05-19
feat: initiate-withdrawal-request

hstove:
- (Partially) Closes https://github.com/stacks-network/sbtc/issues/95

This PR is essentially two things:

- Validation for the `recipient` param (using the same logic and structure as PoX)
- A user-facing function for initiating a withdrawal request

At the moment, this doesn't actually convert the user's sBTC into locked sBTC, as we don't have that contract yet.

This also builds on top of #167 because I added a few helpers in there that I wanted to use in here.

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -11,12 +11,17 @@ jobs:
     name: Run Unit Tests
     runs-on: ubuntu-20.04
     steps:
-
       - uses: actions/checkout@v4
 
       - uses: pnpm/action-setup@v2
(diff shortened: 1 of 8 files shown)


setzeus (approved):
All Clarity code LGTM & tests fully hit edge-case. Phenomenal work.

setzeus (changes requested):
All Clarity code LGTM & tests fully hit edge-case. Phenomenal work.

However, it looks like CI is currently failing so let's address that before merging.

setzeus (approved):

Issue #68, opened 2024-04-24
Add integration tests for our RBF transactions


djordon:
This is part of https://github.com/Trust-Machines/sbtc-v1/issues/52 and should follow https://github.com/Trust-Machines/sbtc-v1/issues/67.

Check that the following holds.
- [x] (https://github.com/stacks-network/sbtc/pull/169) RBF transactions where we remove one or more deposits are accepted by the network.
- [x] (https://github.com/stacks-network/sbtc/pull/169) RBF transactions where we remove one or more withdrawals are accepted by the network.
- [ ] RBF transaction *package* is accepted by the network.

PR #171, opened 2024-05-21
feat: fee handling of RBF transaction packages

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/68.

This is a follow-up to https://github.com/stacks-network/sbtc/pull/169, that makes sure that we can Replace-by-Fee (RBF) transaction packages. I thought there was a chance that we won't be able to do it, given the policy described in bitcoin-core [bitcoin/doc/policy/packages.md](https://github.com/bitcoin/bitcoin/blob/ecd23656db174adef61d3bd753d02698c3528192/doc/policy/packages.md#package-mempool-acceptance-rules)

> No transaction in a package can conflict with a mempool transaction. Replace By Fee is currently disabled for packages. (#20833)

But the solution is simple, don't submit transaction packages and instead submit transactions individually. The noteworthy part about this is that now the fees users pay is highly dependent on whether they are serviced in the first transaction or not, since the first one now pays the highest fees during RBF.

Another way to bump the fee is to submit a Child-Pays-For-Parent (CPFP) transaction, but the primary issue with this approach is that the signers end up paying the fee.

## Changes

There are three changes in this PR: 
1. We do not require child transactions in the transaction package to be concerned with any of the transactions that they may be replacing. This saves users money in if their request is serviced in the second or subsequent transaction in the package.
2. Our integration tests test that RBF with a transaction chain will be accepted.
3. Modify the integration tests to run more quickly. Bitcoind would slow down before, probably due to all of the addresses that it was tracking. We now only track one address.

## Testing information

The integration tests here check for both transaction packages and a single transaction where all requests fit into one transaction. I played around with more parameters but this setup is reasonably robust and finishes quickly.

### DIFF
--- a/signer/tests/integration/utxo_construction.rs
+++ b/signer/tests/integration/utxo_construction.rs
@@ -16,7 +16,6 @@ use bitcoin::TxIn;
 use bitcoin::TxOut;
 use bitcoin::Witness;
 use bitcoin::XOnlyPublicKey;
-use bitcoincore_rpc::json::ListUnspentResultEntry;
 use bitcoincore_rpc::RpcApi;
 use secp256k1::SECP256K1;
 use signer::utxo::DepositRequest;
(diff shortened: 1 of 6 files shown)


signer/tests/integration/utxo_construction.rs
```diff
@@ -25,19 +24,19 @@ use signer::utxo::SignerBtcState;
...
-use regtest::DEPOSITS_LABEL;
-use regtest::SIGNER_ADDRESS_LABEL;
-use regtest::WITHDRAWAL_LABEL;
-
-pub fn make_deposit_request(
+pub fn make_deposit_request<U>(
     depositor: &Recipient,
     amount: u64,
-    utxo: &ListUnspentResultEntry,
+    utxo: &U,
```
netrome, line 33: Nit: Since `.clone()` is only called once I'd recommend taking ownership of `U` instead of borrowing it - leaving the decision of whether to clone or move it up to the caller.

```suggestion
    utxo: U,
```

signer/tests/integration/utxo_construction.rs
```diff
@@ -25,19 +24,19 @@ use signer::utxo::SignerBtcState;
...
     signers_public_key: XOnlyPublicKey,
     faucet_public_key: XOnlyPublicKey,
-) -> (Transaction, DepositRequest) {
+) -> (Transaction, DepositRequest)
+where
+    U: AsUtxo + Clone,
```
netrome, line 38:
```suggestion
    U: AsUtxo,
```

netrome (approved):
Nice stuff. I'm relieved that we are able to rbf chained transactions. Just one nit from me.

signer/tests/integration/utxo_construction.rs
netrome: (If we take U by value we can skip this Clone bound)

signer/tests/integration/utxo_construction.rs
djordon: Oh yeah, makes sense. Will update.

Issue #177, opened 2024-05-21
[Feature]: Bitcoin block notifier


netrome:
# Feature - Bitcoin block notifier

### Description
Build a block notifier that implements the interface defined in #119. The notifier should allow the signer block observer to subscribe to bitcoin blocks.

PR #187, opened 2024-05-23
Define BlockNotifier Interface and Implementation with Electrum

### LINKED ISSUE #119: [Feature]: Bitcoin block notifier interface
# Feature - Bitcoin block notifier interface

## 1. Description
Define an interface that allows the block observer to subscribe to new bitcoin blocks.

### 1.1 Context & Purpose
The bitcoin block observer will react to new bitcoin blocks and populate the signer tables whenever a new bitcoin block is mined.

### 2.1 Acceptance Criteria:
   - [ ] We have an interface that allows the bitcoin block observer to subscribe to bitcoin blocks.

8marz8:
Closes #119 and #177

This PR provides functionality to notify about new blocks using the Electrum client. It periodically pings the Electrum server and subscribes to block headers, broadcasting the notifications to multiple consumers using a broadcast channel.


8marz8:
@netrome Since there was a separate issue for the interface, I started with a separate PR. 
I haven't gotten to the impl part, but before I proceed, is this in line with the design? in terms of the interaction/inter-dependency between the Block observers and block notifier.

netrome:
> @netrome Since there was a separate issue for the interface, I started with a separate PR. I haven't gotten to the impl part, but before I proceed, is this in line with the design? in terms of the interaction/inter-dependency between the Block observers and block notifier.

Thank you for putting this together. While this looks like it would work, it's creating a strong coupling between the block observer and block notifier since both entities are required to be aware of each other. I would love it if we could maintain an interface that requires at most one of the entities to be aware of the other.

For example, if you want the `Block Observer` to expose a `notify` method, I don't think the `Block Notifier` will even need a "subscribe" method or an interface at all. You could just design a block notifier to be generic over the Block Observer, and have it call `notify` directly whenever new blocks arrive.

The converse also works. If we have a `Block Notifier` interface that exposes a `subscribe` method, that method could just return a stream of block headers - so the interface would not require any knowledge of a block observer.

Personally I'd lean on the latter (just having a `Block Notifier` trait with a `subscribe` method) but I think both will work.

signer/src/block_notifier.rs
```diff
@@ -0,0 +1,33 @@
...
+    fn notify(&self, block_hash: &bitcoin::BlockHash) -> Result<(), Self::Error>;
+}
+
+/// The `BlockNotifier` trait defines methods for subscribing observers and running the notifier.
+pub trait BlockNotifier {
+    /// Errors occurring during subscription or running the notifier
+    type Error: Error;
+
+    /// Adds the provided observer to the list of observers that will be notified of new block headers
+    fn subscribe(&mut self, observer: Arc<dyn BlockObserver<Error = Self::Error>>) -> impl Future<Output = Result<(), Self::Error>> + Send;
```
netrome, line 27: I think this could be simplified so that the `BlockNotifier` won't have to be aware of the `BlockObserver`. This could just return a `tokio::mpsc::Receiver` or a [`futures_core::stream::Stream`](https://docs.rs/futures-core/0.3.30/futures_core/stream/trait.Stream.html) of blocks (though the latter would probably require an associated type for the actual stream that implements the `Stream` trait).

netrome (changes requested):
Nice stuff, but as commented I think we only need one of the interfaces. My preference is to keep just the `BlockNotifier` trait, but I can go either way.

signer/src/block_notifier.rs
8marz8: @netrome  I got rid of the BlockObserver trait and since the BlockNotifier trait was small enough, I added the impl with Electrum in same PR. Is this closer to what we want?

I made the integ test to work locally by changing Gowtham's electrs docker details to use 0.0.0.0 but unfortunately, I couldn't make it work with the original 121.0.0.1 host so I didn't push the changed devenv files and ignored the test for now.

netrome: Okay, I think it should be using "0.0.0.0" within the container so anything else is probably a bug. Could you include the fix in the PR?

signer/src/block_notifier.rs
```diff
@@ -0,0 +1,150 @@
...
+use tokio_stream::wrappers::ReceiverStream;
+use tracing::{error, warn};
+
+/// The `BlockNotifier` trait defines a method for subscribing to a stream of block headers.
+pub trait BlockNotifier {
+    /// Errors occurring during subscription or running the notifier.
+    type Error: Error;
+
+    /// Returns a stream of block headers.
+    fn subscribe(&self) -> Pin<Box<dyn Stream<Item = Result<BlockHash, Self::Error>> + Send>>;
```
netrome, line 24: While this is alright, I'd prefer to use generics over trait objects so we don't have to do explicit boxing and pinning
```suggestion
    fn subscribe(&self) -> impl Stream<Item = Result<BlockHash, Self::Error>> + Send;
```
this makes the code a bit cleaner (in my opinion) and saves us some redundant heap allocations.

signer/src/block_notifier.rs
```diff
@@ -0,0 +1,150 @@
...
+    subscribe_interval: Duration,
+}
+
+impl ElectrumBlockNotifier {
+    /// Creates a new instance of `ElectrumBlockNotifier` from config settings.
+    ///
+    /// # Returns
+    ///
+    /// A new instance of `ElectrumBlockNotifier`.
+    pub fn from_config() -> Result<Self, ElectrumError> {
```
netrome, line 42: Can we take the config as an argument here? In the binary we can just inject the system config, but for test cases I'd like the test configuration to be explicit and not dependent on the system environment if possible.

signer/tests/integration/block_notifier.rs
```diff
@@ -0,0 +1,29 @@
...
+use tracing_subscriber;
+
+#[ignore]
+#[tokio::test]
+async fn test_electrum_block_notifier_integration() {
+    // Initialize logging
+    tracing_subscriber::fmt().with_max_level(Level::INFO).init();
+
+    // Create the notifier
+    let notifier = ElectrumBlockNotifier::from_config().unwrap();
```
netrome, line 13: I'd like the test configuration to be explicit and not taken from the system config if possible.

signer/src/block_notifier.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    /// Subscribes to a stream of block headers.
+    ///
+    /// # Returns
+    ///
+    /// A pinned boxed stream of block headers.
+    fn subscribe(&self) -> Pin<Box<dyn Stream<Item = Result<BlockHash, Self::Error>> + Send>> {
+        let (sender, receiver) = channel(100); // Bounded channel to control memory usage
+        self.start_notifying(sender);
+        Box::pin(ReceiverStream::new(receiver))
```
netrome, line 148: If you update the trait to use generics, we can return the receiver stream without boxing here
```suggestion
        ReceiverStream::new(receiver)
```

signer/src/block_notifier.rs
```diff
@@ -0,0 +1,150 @@
...
+    type Error = ElectrumError;
+
+    /// Subscribes to a stream of block headers.
+    ///
+    /// # Returns
+    ///
+    /// A pinned boxed stream of block headers.
+    fn subscribe(&self) -> Pin<Box<dyn Stream<Item = Result<BlockHash, Self::Error>> + Send>> {
+        let (sender, receiver) = channel(100); // Bounded channel to control memory usage
+        self.start_notifying(sender);
```
netrome, line 147: I don't think adding a subscriber should spawn a new notification loop. Instead, I think we should separate the block notifier run loop from the receiver - where the receiver would implement the `BlockNotifier` interface. This can be done in several ways, and I'd suggest something along the lines of defining:

`ElectrumBlockNotifier` (could be renamed to `ElectrumWatcher`) - basically what you have now, exposing a `start_notifying` (or I'd just call it `run` - but both work) method which returns a receiver or a struct wrapping a receiver (suggested name `ElectrumBlockReceiver`), and implement `BlockNotifier` for the latter struct, where `subscribe` would just copy the receiver channel and wrap it in a stream.

netrome: I'm happy to jump on a call to further elaborate on this if you have the time.

signer/src/block_notifier.rs
```diff
@@ -0,0 +1,150 @@
...
+                            error!("Max retry attempts reached, giving up.");
+                            continue_loop = false;
+                        }
+                    } else {
+                        retry_attempts = 0; // Reset retry attempts on successful ping
+                    }
+                }
+                // Periodic block header subscription
+                _ = subscribe_interval.tick() => {
+                    match client.block_headers_subscribe() {
```
netrome, line 90: I suspect that this is a blocking call, which means that it should be executed from a separate task using `tokio::spawn_blocking`. I'm also curious how this call works, can we get the same block hash on multiple calls or is it somehow guaranteed to be incremental and how does that work? I'm looking at the `electrum-client` docs (at the [ElectrumAPI](https://docs.rs/electrum-client/0.20.0/electrum_client/trait.ElectrumApi.html#method.block_headers_subscribe) trait) but I find it hard to understand the mechanics of this call.

netrome (changes requested):
Great stuff so far. A few questions and nits from me, but the major things that I'd like to have changed is

1. `client.block_headers_subscribe()` should be called from a task spawned with `tokio::task::spwan_blocking` ([docs](https://docs.rs/tokio/1.37.0/tokio/task/fn.spawn_blocking.html)).
2. Calling `subscribe` should not spawn a new notify loop.

Let me know if you want to jump on a call to discuss these changes some more.

signer/src/block_notifier.rs
```diff
@@ -0,0 +1,177 @@
...
+    subscribe_interval: Duration,
+}
+
+impl ElectrumBlockNotifier {
+    /// Creates a new instance of `ElectrumBlockNotifier` from config.
+    ///
+    /// # Returns
+    ///
+    /// A new instance of `ElectrumBlockNotifier`.
+    pub fn from_config(config: BlockNotifierConfig) -> Result<Self, ElectrumError> {
```
netrome, line 52: Nit: I recommend taking the config as a reference and not by value here.
```suggestion
    pub fn from_config(config: &BlockNotifierConfig) -> Result<Self, ElectrumError> {
```

signer/src/block_notifier.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    /// Creates a new instance of `ElectrumBlockNotifier` from system config.
+    ///
+    /// # Returns
+    ///
+    /// A new instance of `ElectrumBlockNotifier`.
+    pub fn new() -> Result<Self, ElectrumError> {
+        let config = &SETTINGS.block_notifier;
+        Self::from_config(config.clone())
+    }
```
netrome, line 77: If we take the config by reference:
```suggestion
    pub fn new() -> Result<Self, ElectrumError> {
        Self::from_config(&SETTINGS.block_notifier)
    }
```

signer/tests/integration/block_notifier.rs
```diff
@@ -0,0 +1,72 @@
+use futures::StreamExt;
+use signer::block_notifier::{BlockNotifier, ElectrumBlockNotifier};
+use signer::config::BlockNotifierConfig;
+use tracing::{error, info, warn, Level};
+use tracing_subscriber;
+
+// TODO: need to fix docker-compose.test.yml setup so bitcoin and bitcoind
```
netrome, line 7: Isn't the `bitcoin` and `bitcoind` essentially the same container? Otherwise, it would be great if you could create a ticket for this and reference it in the TODO comment.

netrome (approved):
Great stuff, thank you!

signer/src/block_notifier.rs
```diff
@@ -0,0 +1,177 @@
...
+                    let sender_clone = sender.clone();
+
+                    task::spawn_blocking(move || {
+                        match client_clone.blocking_lock().block_headers_subscribe() {
+                            Ok(HeaderNotification { header, .. }) => {
+                                let block_hash = header.block_hash();
+                                let _ = sender_clone.send(Ok(block_hash));
+                            }
+                            Err(e) => {
+                                let _ = sender_clone.send(Err(Error::Electrum(Arc::new(e))));
```
netrome, line 122: Nit: I don't think the electrum errors should be propagated to the subscribers. Instead we should either have some sort of retry logic and only send block headers to the subscribers when the client succeeds.

If you don't have time to implement this, feel free to merge as-is and we can take that as a follow-up.

8marz8:
@netrome  I addressed the small nits. Merging it as-is but I left the comments I didn't address unresolved so it can be referenced more easily. Thanks!

PR #197, opened 2024-05-24
Local Kubernetes setup for BTC Node and Miner

SundarGowtham:
## Description

This PR contains the necessary components in K8s and Docker to get a Bitcoin Regtest node and a BTC miner running locally

This PR contains:
- Docker build files for BTC Node & Miner
- Kubernetes Deployment, Service, and Namespace for BTC Node
- `utils/` folder which contains K8s utility scripts (port-forwarding etc.)
- `up.sh`, `build.sh`, `down.sh`, and `remove-minikube-containers.sh`
- `tests/devnet-liveness.sh` script which tests for BTC liveness and BTC mining ability

## Steps to replicate:

Please follow the instructions at: `devenv/local/k8s/README.md`

After completing the `Installation` section, please run the following:

[1] `minikube start`
[2] `sh build.sh`
[3] `sh up.sh`
[4] `sh ./tests/devnet-liveness.sh`
[5] `sh down.sh`

In order for this PR to pass, the tests in `devnet-liveness.sh` should pass


devenv/local/k8s/custom-k8s-docker-builds/bitcoin-miner-sidecar/docker/Dockerfile
netrome, line 1: This and the other Dockerfile looks identical to the ones under `docker-compose/<service>/docker/Dockerfile`. How about breaking them out to a shared location so that we don't have to duplicate their definitions?

devenv/local/k8s/custom-k8s-docker-builds/bitcoin/docker/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+    ca-certificates \
+    coreutils \
+    nginx \
+    sudo \
+    curl \
+    jq && rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /bitcoin/bin/* /usr/local/bin
+
+COPY default.conf /etc/nginx/conf.d/default.conf
```
netrome, line 59: This doesn't build on my machine unless I add the trailing slashes. I don't know why it worked in the docker-compose version since it's identical.

```suggestion
COPY --from=builder /bitcoin/bin/* /usr/local/bin/

COPY default.conf /etc/nginx/conf.d/default.conf/
```

devenv/local/k8s/yamls/deployments/bitcoin-deployment.yaml
```diff
@@ -0,0 +1,223 @@
...
+              value: "0"
+            - name: BTC_REST_ENABLE
+              value: "0"
+
+
+          volumeMounts:
+            - name: shared-data
+              mountPath: /mnt/shared
+
+      containers:
```
netrome, line 121: What's the reasoning behind having both the regtest container and the miner sidecar in the same pod? I'm pretty used to the one-container-per-pod convention, and it might come in handy to be able to turn off the miner while still running the regtest deployment imo.

netrome (commented):
Some comments, I'm mainly interested in if we can share the dockerfiles between the docker-compose local environment and this one.

devenv/local/k8s/custom-k8s-docker-builds/bitcoin-miner-sidecar/docker/Dockerfile
SundarGowtham: Two considerations.

- Kubernetes does not work with build time configurations like Docker Compose does with the `build:` key. Which means anything which requires build time tuning as you will see in the creation of the `bitcoin.conf` in the Bitcoin container or the `config.toml` file in the Stacks/Nakamoto Signer containers will not be the same between K8s and Docker Compose. In Kubernetes, we will be addressing this by using `initContainers` and creating the files there whereas in Docker, we do not have this limitation. So the `Dockerfile`'s will be different between the two.

- Currently in Docker Compose, we are having to copy a hard-coded `config.toml` into the container which decreases the utility of this devnet since we would like to be able to tune these configs as `env` variables rather than going in and change the raw data. As we start to add more and more of our sBTC components into the devnet, the Dockerfiles between the two will most likely differ.

Once we start to see that certain docker containers are not changing between Docker Compose and K8s, then we could just add a shared folder and reference that in both configs

devenv/local/k8s/yamls/deployments/bitcoin-deployment.yaml
netrome: The same namespace definitely. On this note, I think it would make sense to have the whole devenv in a single `sbtc` namespace. But if you want more fine-grained slicing of namespaces I can work with that too.

devenv/local/k8s/custom-k8s-docker-builds/bitcoin-miner-sidecar/docker/Dockerfile
netrome: Okay, if you expect them to diverge then I'm okay with this.

netrome (commented):
Looks good to me so far. Is this still a draft?

PR #215, opened 2024-05-30
refactor: Gather testing utilities under the top-level `testing` module

### LINKED ISSUE #193: [Refactor]: Gather signer testing utilities under the testing module
Spawned from https://github.com/stacks-network/sbtc/pull/189#discussion_r1611647519

### Description
Currently we have a lot of test utilities as sub modules to the modules that they are testing (e.g. `signer::models::testing`, `signer::network::testing`, `signer::storage::model::testing` etc.). Instead, we'd like to gather these modules under the top-level `testing` module as `signer::testing::network`, `signer::testing::models` etc.

### Definition of done
Everything feature flagged by `#[cfg(feature = "testing")]` exists under the top-level testing module.

netrome:
closes #193


signer/src/testing/message.rs
```diff
@@ -107,9 +108,12 @@ impl fake::Dummy<fake::Faker> for super::WstsMessage {
     }
 }
 
-fn dummy_payload<P: Into<super::Payload> + fake::Dummy<fake::Faker>, R: rand::RngCore + ?Sized>(
+fn dummy_payload<
+    P: Into<message::Payload> + fake::Dummy<fake::Faker>,
+    R: rand::RngCore + ?Sized,
+>(
```
djordon, line 114: style nit: in these kinds of situations, can we go with something like this style instead?
```rust
fn dummy_payload<P, R>(config: &fake::Faker, rng: &mut R) -> message::Payload
where
    P: Into<message::Payload> + fake::Dummy<fake::Faker>,
    R: rand::RngCore + ?Sized,
{
```
I find it much easier to read, but if you don't feel free to keep it as is.

djordon (approved):
Looks good, just had a style nit.

signer/src/testing/message.rs
netrome: Definitely, I'll update.

netrome: Done ✅

PR #226, opened 2024-06-04
Add Dockerfile for arm64 build container

xoloki:
## Description

This change adds a `Dockerfile` that creates a build container for the `arm64` architecture.  There is an absolute nightmare of dependencies for this repository, so having a container will save everyone time (myself included).

## Type of Change
- New feature

No

## Are documentation updates required?
No


Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+RUN apt-get install -y libssl-dev
+RUN cargo install sqlx-cli
+
+RUN apt-get install -y make
+
+RUN mkdir -p /tmp/node-install
+RUN curl -fsSL https://deb.nodesource.com/setup_22.x -o /tmp/node-install/nodesource_setup.sh
+RUN bash /tmp/node-install/nodesource_setup.sh
+RUN apt-get -y install nodejs
```
djordon, line 32: Could this be added to the first `RUN apt-get install` as well.

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+RUN mkdir -p /usr/local/java
+RUN cd /usr/local/java && tar xvzf /tmp/java-install/jdk-21_linux-aarch64_bin.tar.gz
+ENV PATH="${PATH}:/usr/local/java/jdk-21.0.3/bin"
+
+RUN mkdir -p /tmp/rustup-install
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs -o /tmp/rustup-install/rustup-install.sh
+RUN chmod +x /tmp/rustup-install/rustup-install.sh
+RUN /tmp/rustup-install/rustup-install.sh -y
```
djordon, line 41: Is this necessary if cargo is already installed?

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN apt-get install -y protobuf-compiler
+
+RUN apt-get install -y libssl-dev
+RUN cargo install sqlx-cli
+
+RUN apt-get install -y make
+
+RUN mkdir -p /tmp/node-install
+RUN curl -fsSL https://deb.nodesource.com/setup_22.x -o /tmp/node-install/nodesource_setup.sh
+RUN bash /tmp/node-install/nodesource_setup.sh
```
djordon, line 31: What does this do?

AshtonStephens:
Looking

Dockerfile
```diff
@@ -0,0 +1,47 @@
+FROM ubuntu:jammy
+
+MAINTAINER Joey Yandle
```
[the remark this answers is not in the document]
xoloki: Done.

Dockerfile
```diff
@@ -0,0 +1,47 @@
+FROM ubuntu:jammy
```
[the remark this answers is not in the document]
xoloki: It's considered a best practice to only use the new ubuntu LTS after the August point release updates.  Before that the LTS is not exactly stable.

DOCKER.md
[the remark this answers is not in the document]
xoloki: Done.

DOCKER.md
```diff
@@ -0,0 +1,14 @@
...
+```
+docker build -t sbtc-build .
+```
+To run the build container, do
+```
+docker run -it -v `pwd`:/build sbtc-build
+```
+Once inside the build container, do
+```
+cd /build
```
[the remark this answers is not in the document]
xoloki: Done.

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+MAINTAINER Joey Yandle
+
+RUN apt-get update
+RUN apt-get install -y curl wget unzip pkg-config gcc automake autoconf autotools-dev libtool man gdb git screen sudo rsync cargo
```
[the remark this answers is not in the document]
xoloki: I'm really not a fan of multi-line docker commands.  

I do however agree on cleaning up.

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+RUN mkdir -p /tmp/java-install
+RUN curl -L https://download.oracle.com/java/21/latest/jdk-21_linux-aarch64_bin.tar.gz -o /tmp/java-install/jdk-21_linux-aarch64_bin.tar.gz
+
+RUN apt-get install -y protobuf-compiler
+
+RUN apt-get install -y libssl-dev
+RUN cargo install sqlx-cli
+
+RUN apt-get install -y make
```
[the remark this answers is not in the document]
xoloki: Done.

Dockerfile
xoloki: This one has to be after the `nodesource` install because that adds repositories.

Dockerfile
xoloki: It's necessary to be able to get the correct version of `rustc`.

Dockerfile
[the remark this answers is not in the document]
djordon: Oh yeah that makes perfect sense.

Dockerfile
djordon: Oh right, that makes sense. The `https://deb.nodesource.com` url should have been a giveaway.

Dockerfile
djordon: Really? That's surprising, I thought it would auto-update to the correct version after running `cargo build`.

xoloki: It was complaining about having `rustc` `1.75` but needing `1.76` or more recent.

Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM ubuntu:jammy
+
+LABEL Joey Yandle
```
AshtonStephens, line 3:
```suggestion
MAINTAINER Joey Yandle <joey@trustmachines.co>
```

Dockerfile
djordon: Oh right, that's because apt-get doesn't install `rustup`, which I suppose cargo delegates to when it notices that it needs a different version of rust. Okay, never mind.

AshtonStephens: I prefer the all in one `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs/ | sh -s -- -y` so just want to make sure you considered it and are choosing a different path. I prefer this because there's no artifact; we aren't checking any hash here so we don't really need one.

README.md
```diff
@@ -90,3 +90,24 @@ does not support development on Windows or z/OS.
...
+#### Run the build container
+```
+docker run -it -v `pwd`:/build sbtc-build
+```
+
+#### Build using the build container
+```
+cd /sbtc
+make install
+make build
```
AshtonStephens, line 112: For my own education, can you elaborate on why this works? The Makefile hasn't received any edits so I don't understand how this would do what it does.

Dockerfile
djordon: `MAINTAINER` is [deprecated](https://docs.docker.com/reference/dockerfile/#label), but I think this should be changed to
```docker
LABEL org.opencontainers.image.authors="Joey Yandle <joey@trustmachines.co>"
```

Dockerfile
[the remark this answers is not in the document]
djordon: Fair enough. Also we'd have to explicitly add `ca-certificates` for my suggestion to work.

Dockerfile
xoloki: Fixed @djordon @AshtonStephens

xoloki:
@djordon @AshtonStephens when you're good please resolve convos and if necessary approve separately.

AshtonStephens:
Approving, I think this looks good and is better than everyone doing everything locally. I think there might be some rough edges with us having two build setups now but we can fix that down the line. I've opened this ticket for it: https://github.com/stacks-network/sbtc/issues/236

AshtonStephens (approved):

AshtonStephens (changes requested):
Actually sorry, getting this error:

```cli
Packages: +649
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 ERR_PNPM_ENOENT  ENOENT: no such file or directory, copyfile '/build/.pnpm-store/v3/files/7f/65c6403a23d93fb148e8259b012d6552ab3bff178f4a7d6a9d9cec0f60429fc1899e39b4bca8cc08afc75d9a7c7bfdb13fc372ca63c85eb22b0355eb4d6000' -> '/build/node_modules/.pnpm/@nestjs+core@10.3.0_@nestjs+common@10.3.0_reflect-metadata@0.1.13_rxjs@7.8.1__reflect-metadata@0.1.13_rxjs@7.8.1/node_modules/@nestjs/core_tmp_17/inspector/interfaces/extras.interface.js'
```

Any ideas?

xoloki:
> Actually sorry, getting this error:
> 
> ```
> Packages: +649
> ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
>  ERR_PNPM_ENOENT  ENOENT: no such file or directory, copyfile '/build/.pnpm-store/v3/files/7f/65c6403a23d93fb148e8259b012d6552ab3bff178f4a7d6a9d9cec0f60429fc1899e39b4bca8cc08afc75d9a7c7bfdb13fc372ca63c85eb22b0355eb4d6000' -> '/build/node_modules/.pnpm/@nestjs+core@10.3.0_@nestjs+common@10.3.0_reflect-metadata@0.1.13_rxjs@7.8.1__reflect-metadata@0.1.13_rxjs@7.8.1/node_modules/@nestjs/core_tmp_17/inspector/interfaces/extras.interface.js'
> ```
> 
> Any ideas?
> 
> And separately this when I try to run `make build` in the container.
> 
> ```
> > blocklist-api@ build-openapi /build/.generated-sources/blocklist-openapi-gen
> > openapi-generator-cli generate -i ./blocklist-client-openapi.json -g rust -o ../blocklist-api --config ./config.json
> 
> sh: 1: openapi-generator-cli: not found
>  ELIFECYCLE  Command failed.
> ```

Did you run `make install` before `make build` @AshtonStephens ?

xoloki:
You might also need a fresh checkout @AshtonStephens since mounting `pwd` means that the built code in on the host not the container.

AshtonStephens:
> Did you run `make install` before `make build` @AshtonStephens ?

Yes the first error is from the make install. Okay, I'll try a fresh instance of the repository.

xoloki:
> > Did you run `make install` before `make build` @AshtonStephens ?
> 
> Yes the first error is from the make install. Okay, I'll try a fresh instance of the repository.

Just tried fresh checkout of source repo and `make install` def succeeded.  Running `make build` now

AshtonStephens:
I think, though, that there are some rough edges here that need to be rounded out, so I'd keep this method of building the repository out of the `README.md` until we've had a chance to make this a bit more robust. But in the immediate term it makes sense to unblock your development.

xoloki:
> I think, though, that there are some rough edges here that need to be rounded out, so I'd keep this method of building the repository out of the `README.md` until we've had a chance to make this a bit more robust. But in the immediate term it makes sense to unblock your development.

I'll move it back to `DOCKER.md`

xoloki:
Ok back to separate `DOCKER.md` @AshtonStephens @djordon

AshtonStephens:
@xoloki why don't we merge this tomorrow, I might be the weak link here but for whatever reason this isn't working for me and I don't want to rush this in and have a confusing build setup that only works on a subset of computers.

xoloki:
Ok @AshtonStephens happy to merge whenever I get a +1.  Thanks for helping me get this out the door.

.screenrc
djordon, line 1: I think I overlooked this yesterday, but why is this needed in the repo? I would think that it's better to gitignore custom configuration files like this in cases where someone else prefers a different setup. We can still mount a local version during `docker run`. Thoughts?

xoloki: I always `.gitignore` `*~` so emacs backup files don't get added accidentally @djordon , but I'll remove it from this PR since it's orthogonal to the change.

AshtonStephens (approved):
Approving pending @djordon's approval.

.screenrc
xoloki: Done.

djordon: I think you removed the `.gitignore` file, but I want to add the `.screenrc` file to the `.gitignore` file (and remove copying it to the `Dockerfile`). This way you can still have and use it locally by mounting your local copy to the right place during `docker run` and others can use their preferred configuration files as well.

djordon (approved):
Looks good ✅

djordon:
Just to be clear, I'm also getting the same errors as @AshtonStephens but we're going to fix this up later right?

PR #239, opened 2024-06-10
ci: remove the sqlx-cli from ci

djordon:
## Description

I got annoyed with how long it was taking CI to run, so I thought I'd just replace the sqlx-cli with a similar tool that had a docker image that we could use.

Since we still require sqlx-cli for generating prepare stubs, I didn't remove it from our `Dockerfile` or from the README, but it's not needed for building or for running tests. We need to generate prepare stubs if we continue to add new queries that use the `sqlx::query!` and `sqlx::query_as!` macros.

## Changes

1. Remove the sqlx-cli step in CI. Remove the `sqlx migrate` call in the `Makefile`.
2. Add a flyway service for running migrations in CI.
3. Change the `created_at` types to `TIMESTAMP WITH TIME ZONE`. Either type is probably fine,  but I think we'll run into fewer issues by using `TIMESTAMP WITH TIME ZONE`.
4. Update the randomly generated data in our postgres tests to conform to postgres type limits.

## Testing information

If CI passes then we're good.


djordon:
I just realized that you only pay the sqlx-cli install cost for the first CI run and it's cached after that. Oh well.

signer/migrations/0003__create_tables.sql
```diff
@@ -12,14 +12,14 @@ CREATE TABLE sbtc_signer.bitcoin_blocks (
     block_height BIGINT NOT NULL,
     parent_hash BYTEA NOT NULL,
     confirms BYTEA,
-    created_at TIMESTAMP NOT NULL
+    created_at TIMESTAMPTZ NOT NULL
```
netrome, line 15: Do we need the timezone information? I figured let's just store all timestamps in UTC which should save us a few bytes and any timezone-related headaches.

signer/src/testing/storage/model.rs
```diff
@@ -62,6 +63,10 @@ fn idealistic_bitcoin_chain(
 ) -> impl FnMut(&mut Vec<model::BitcoinBlockHash>, usize) -> Option<model::BitcoinBlock> + '_ {
     |block_hashes, _| {
         let mut block: model::BitcoinBlock = fake::Faker.fake_with_rng(rng);
+        let block_hash: [u8; 32] = fake::Faker.fake_with_rng(rng);
+        block.block_hash = block_hash.to_vec();
+        block.created_at = block.created_at.max(OffsetDateTime::now_utc());
```
netrome, line 68: I would like this function to be deterministic. If we need to control the `created_at`, I'd prefer another utility to do so - or simply just modify the created blocks after calling this function.

netrome (changes requested):
Thank you for updating the CI. However, I would like to keep the test utilities for creating fake data deterministic so I'd appreciate if we could avoid calling `OffsetDateTime::now_utc()` in those.

signer/migrations/0003__create_tables.sql
djordon: Both types take up the same amount of space in the database, and Postgres does not store the timezone information. It's just aware of the timezone if we happen to pass it a timestamp with a non-UTC offset (and so stores the correct instant in time). Since everything is probably in UTC, this isn't an issue, but I only see benefits by using `TIMESTAMPTZ`, hence the change.

signer/src/testing/storage/model.rs
djordon: Oh, right whoops. I can make it deterministic. I don't know why I didn't think to use the unix epoch before.

netrome (approved):

PR #248, opened 2024-06-20
feat: Skeleton for transaction signer event loop with request decision handling

### LINKED ISSUE #234: [Feature]: Skeleton implementation of transaction signer event loop
# Feature - Skeleton implementation of transaction signer event loop

### Context
We have already documented what we intend the signer event loop to do in #221. As a next step we should make a skeleton implementation of this loop with placeholder logic for all independent pieces of functionality and create separate tickets for that.

### Definition of done
We have a skeleton implementation of the signer event loop and an exhaustive set of issues to finish it.

netrome:
closes #234 

This PR adds a rough skeleton implementation for the transaction signer event loop with the logic to make and persist decisions for pending requests, including a unit test for the handling of deposit requests.

The most important changes are the additions to `transaction_signer.rs`, whereas most of the remaining changes are database queries and updates to data generation logic for tests.

### DIFF
--- a/contracts/tests/clarigen-types.ts
+++ b/contracts/tests/clarigen-types.ts
@@ -1330,7 +1330,7 @@ export const contracts = {
         value: 4n,
       },
       tokenDecimals: 8n,
-      tokenName: "sBTC Mini",
+      tokenName: "sBTC",
       tokenSymbol: "sBTC",
       tokenUri: null,
     },
--- a/signer/src/storage/model.rs
+++ b/signer/src/storage/model.rs
@@ -1,18 +1,24 @@
 //! Database models for the signer.
 
+use fake::faker::time::en::DateTimeAfter;
+
 /// Bitcoin block.
 #[derive(Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "testing", derive(fake::Dummy))]
 pub struct BitcoinBlock {
     /// Block hash.
+    #[dummy(expr = "fake::vec![u8; 32]")]
     pub block_hash: BitcoinBlockHash,
     /// Block height.
     pub block_height: i64,
     /// Hash of the parent block.
+    #[dummy(expr = "fake::vec![u8; 32]")]
     pub parent_hash: BitcoinBlockHash,
     /// Stacks block confirmed by this block.
+    #[dummy(default)]
     pub confirms: Option<StacksBlockHash>,
     /// The time this block entry was created by the signer.
+    #[dummy(faker = "DateTimeAfter(time::OffsetDateTime::UNIX_EPOCH)")]
     pub created_at: time::OffsetDateTime,
 }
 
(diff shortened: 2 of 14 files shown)


signer/src/storage/model.rs
```diff
@@ -1,18 +1,24 @@
 //! Database models for the signer.
 
+use fake::faker::time::en::DateTimeAfter;
+
 /// Bitcoin block.
 #[derive(Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "testing", derive(fake::Dummy))]
 pub struct BitcoinBlock {
     /// Block hash.
+    #[dummy(expr = "fake::vec![u8; 32]")]
```
djordon, line 10: Ahh, good to know. I was wondering how to constrain the faker to always return 32 bytes without using `[u8; 32]`. I mistakenly assumed it wasn't possible.

Related, why not use `[u8; 32]` for the aliases where we want a fixed number of bytes?

signer/src/storage/model.rs
```diff
@@ -1,18 +1,24 @@
...
     /// Block height.
     pub block_height: i64,
     /// Hash of the parent block.
+    #[dummy(expr = "fake::vec![u8; 32]")]
     pub parent_hash: BitcoinBlockHash,
     /// Stacks block confirmed by this block.
+    #[dummy(default)]
     pub confirms: Option<StacksBlockHash>,
     /// The time this block entry was created by the signer.
+    #[dummy(faker = "DateTimeAfter(time::OffsetDateTime::UNIX_EPOCH)")]
```
djordon, line 21: Oh nice!

contracts/tests/clarigen-types.ts
```diff
@@ -1330,7 +1330,7 @@ export const contracts = {
         value: 4n,
       },
       tokenDecimals: 8n,
-      tokenName: "sBTC Mini",
+      tokenName: "sBTC",
```
djordon, line 1333: Good catch

djordon (commented):
Looks good! Had one question about aliases types.

djordon (approved):
I meant to approve the first time

contracts/tests/clarigen-types.ts
netrome: Yeah, this was autogenerated by "make build" after I had a CI failure caused by this 😅

signer/src/storage/model.rs
netrome: Yeah, I stumbled upon it in the docs and figured it's nice to fix it while I'm at it :blush:

netrome:
> I meant to approve the first time

Thanks, and thank you for the review :pray:

signer/src/storage/model.rs
netrome: Yeah I think writing is fine because it takes slices, but reading doesn't support fixed-size arrays because sqlx can't guarantee the size of the read data.

> We could use postgres [domains](https://www.postgresql.org/docs/16/sql-createdomain.html) to get the fixed length "data type". I'm not trying to suggest that we do so and I do not have a good sense on whether doing so would be a net benefit, I just wanted to point out that it is possible.

Interesting, didn't know of this. Thanks for sharing! :pray:

PR #249, opened 2024-06-20
feat: Modify block observer to use the top-level error::Error type

### LINKED ISSUE #229: [Feature]: Use `crate::error::Error` directly in `block_observer` instead of having a custom error type
### Context
In #228 we had a clash between two error handling philosophies. On one side, we have `crate::error::Error` which is a single big fat error enum spanning a lot of different things that can go wrong in different pieces of the system. On the other side, we have a bunch of specialized `Error` types encapsulating specific failure modes for particular pieces of functionality.

In particular, the block observer currently has a specialized error type for its event loop - which in #228 is modified to wrap the top level `crate::error::Error` for one of its variants. This is quite awkward, and we have decided that for event loops and similar functionality that does a multitude of things we'll use the top level error type - whereas for specialized functions and utilities we'll allow (and encourage) specialized errors.

### Definition of done
The block observer uses the `crate::error::Error` type instead of having its own error variant.

netrome:
Closes #229 

Currently based on PR #248

### DIFF
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -19,6 +19,7 @@
 
 use std::collections::HashMap;
 
+use crate::error;
 use crate::stacks_api::StacksInteract;
 use crate::storage;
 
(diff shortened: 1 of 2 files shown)


djordon (approved):
Looks good

Issue #212, opened 2024-05-30
[Feature]: Block observer writing stacks block


netrome:
### Context
The block observer currently only writes bitcoin blocks, and contains a placeholder function to write stacks blocks which we should implement.

### Definition of done
The `write_stacks_block` method in the block observer is implemented and tested.

netrome:
@djordon hey I realize I missed that #251 doesn't implement the `write_stacks_block` function in the block observer - although it does all the heavy lifting for the postgresql storage, so reopening this so we can do a follow-up PR to actually implement this in the block observer.

djordon:
Oh right, good call. Should be easy to add in.

netrome:
> Oh right, good call. Should be easy to add in.

Yeah, and if it gets too nitty with the in-memory implementation and unit test we could push that as a follow-up.

Issue #250, opened 2024-06-21
[Feature]: Extract sBTC clarity function call transactions


djordon:
# Feature - Extract sBTC clarity function call transactions

### 1.1 Context & Purpose

The signers record Stacks blocks and transactions for sBTC related contract calls to their database. It's not clear which subset of the contract calls are of interest though, so this ticket is to figure that out and update the code so that we extract the relevant transactions from Stacks blocks.

### 2.1 Acceptance Criteria:
   - [x] Have a list of public functions to note
   - [x] Have the corresponding `transaction_type` for each of the public functions that we'll be storing.
   - [x] Break out the `PgStore::write_stacks_sbtc_txs` function into two, one that extracts the Stacks contract call transactions that call the identified public functions, and another that writes these transactions into the database.

## 3. Related Issues and Pull Requests (optional):

This is related to #212.

djordon:
This was actually finished in https://github.com/stacks-network/sbtc/pull/311.

PR #251, opened 2024-06-21
feat: writing stacks block

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/212.

We need to persist stacks block headers and sBTC related Stacks transactions to the database. This PR includes functions for writing such data given a Nakamoto stacks block.

## Changes
1. Rework the `DbRead` and `DbWrite` traits to be a little simpler by only taking a reference to `self` in the functions that need implementing. Although taking `self` allows for more flexibility, we don't actually need or use it, so this change allows for some simplification.
2. Only implement the above traits for the `SharedStore` and the `PgStore`.
3. Add a function for checking for the existence of a stacks block in the database.
4. Add a function for writing stacks blocks -- both the header and the transactions -- to the database.

### Notes

* I'm pretty sure the actual sBTC contracts will have different names than the ones in this PR.
* I've punted on figuring out which contract call functions to track. It's captured by https://github.com/stacks-network/sbtc/issues/250.

## Testing information
There are new two integration tests. Together they check the following
1. That we write the stacks block headers to the database.
2. That we write sBTC related transactions to the database.
3. That storing stacks blocks and stacks transactions are idempotent operations.
4. The function that checks whether we've stored a given Stacks block ID works.

### DIFF
--- a/signer/src/storage.rs
+++ b/signer/src/storage.rs
@@ -82,7 +82,7 @@ pub trait DbWrite {
         deposit_request: &model::DepositRequest,
     ) -> impl Future<Output = Result<(), Self::Error>> + Send;
 
-    /// Write a withdraw request.
+    /// Write a withdrawal request.
     fn write_withdraw_request(
         &self,
         withdraw_request: &model::WithdrawRequest,
@@ -94,7 +94,7 @@ pub trait DbWrite {
         decision: &model::DepositSigner,
     ) -> impl Future<Output = Result<(), Self::Error>> + Send;
 
-    /// Write a signer decision for a withdraw request.
+    /// Write a signer decision for a withdrawal request.
     fn write_withdraw_signer_decision(
         &self,
         decision: &model::WithdrawSigner,
--- a/signer/src/storage.rs
+++ b/signer/src/storage.rs
@@ -12,6 +12,9 @@ pub mod postgres;
 
 use std::future::Future;
 
+use blockstack_lib::chainstate::nakamoto::NakamotoBlock;
+use blockstack_lib::types::chainstate::StacksBlockId;
+
 /// Represents the ability to read data from the signer storage.
 pub trait DbRead {
     /// Read error.
@@ -19,41 +22,47 @@ pub trait DbRead {
 
     /// Get the bitcoin block with the given block hash.
     fn get_bitcoin_block(
-        self,
+        &self,
         block_hash: &model::BitcoinBlockHash,
     ) -> impl Future<Output = Result<Option<model::BitcoinBlock>, Self::Error>> + Send;
 
     /// Get the bitcoin canonical chain tip
     fn get_bitcoin_canonical_chain_tip(
-        self,
+        &self,
     ) -> impl Future<Output = Result<Option<model::BitcoinBlockHash>, Self::Error>> + Send;
 
     /// Get pending deposit requests
     fn get_pending_deposit_requests(
-        self,
+        &self,
         chain_tip: &model::BitcoinBlockHash,
         context_window: usize,
     ) -> impl Future<Output = Result<Vec<model::DepositRequest>, Self::Error>> + Send;
 
     /// Get signer decisions for a deposit request
     fn get_deposit_signers(
-        self,
+        &self,
         txid: &model::BitcoinTxId,
         output_index: usize,
     ) -> impl Future<Output = Result<Vec<model::DepositSigner>, Self::Error>> + Send;
 
     /// Get pending withdraw requests
     fn get_pending_withdraw_requests(
-        self,
+        &self,
         chain_tip: &model::BitcoinBlockHash,
         context_window: usize,
     ) -> impl Future<Output = Result<Vec<model::WithdrawRequest>, Self::Error>> + Send;
 
     /// Get bitcoin blocks that include a particular transaction
     fn get_bitcoin_blocks_with_transaction(
-        self,
+        &self,
         txid: &model::BitcoinTxId,
     ) -> impl Future<Output = Result<Vec<model::BitcoinBlockHash>, Self::Error>> + Send;
+
+    /// Returns whether the given block ID is stored.
+    fn stacks_block_exists(
+        &self,
+        block_id: StacksBlockId,
+    ) -> impl Future<Output = Result<bool, Self::Error>> + Send;
 }
 
 /// Represents the ability to write data to the signer storage.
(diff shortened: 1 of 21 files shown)


netrome:
> Rework the DbRead and DbWrite traits to be a little simpler by only taking a reference to self in the functions that need implementing. Although taking self allows for more flexibility, we don't actually need or use it, so this change allows for some simplification.

Yeah makes sense. One complication down the line is if we want to support transactions the receiver probably need to be mutable, so if we settle for one receiver I'd recommend going with a mutable receiver.

signer/src/storage.rs
```diff
@@ -12,48 +12,57 @@ pub mod postgres;
...
+
 /// Represents the ability to read data from the signer storage.
 pub trait DbRead {
     /// Read error.
     type Error: std::error::Error;
 
     /// Get the bitcoin block with the given block hash.
     fn get_bitcoin_block(
-        self,
+        &self,
```
netrome, line 25: I'm alright with setting the receiver here, however to be cautious I think we should have it be mutable which I think would be necessary if we want to support transactions later down the line.

signer/src/storage/postgres.rs
```diff
@@ -12,6 +52,111 @@ impl PgStore {
...
+
+        if block_summaries.is_empty() {
+            return Ok(());
+        }
+
+        sqlx::query(
+            r#"
+            INSERT INTO sbtc_signer.stacks_blocks (block_hash, block_height, parent_hash, created_at)
+            SELECT
+                decode(block_id, 'hex')
```
netrome, line 75: Why do we both convert the block ID to hex and then later decode it here? Can't we just write the bytes directly?

signer/src/storage/postgres.rs
```diff
@@ -12,6 +52,111 @@ impl PgStore {
...
+              , chain_length
+              , decode(parent_block_id, 'hex')
```
netrome, line 77: Same here

signer/src/storage/postgres.rs
```diff
@@ -12,6 +52,111 @@ impl PgStore {
...
+                chain_length: block.header.chain_length as i64,
+                parent_block_id: block.header.parent_block_id.to_hex(),
+            })
+            .collect();
+
+        if block_summaries.is_empty() {
+            return Ok(());
+        }
+
+        sqlx::query(
```
netrome, line 71: Nit: I prefer using the macros, which makes it quicker and easier for me to verify that the queries are correct.

```suggestion
        sqlx::query!(
```

signer/src/storage/postgres.rs
```diff
@@ -12,6 +52,111 @@ impl PgStore {
...
+                _ => None,
+            })
+            .collect();
+
+        if block_txs.is_empty() {
+            return Ok(());
+        }
+
+        let block_txs_json = serde_json::to_string(&block_txs).map_err(Error::JsonSerialize)?;
+        sqlx::query(
```
netrome, line 121: Nit: Again, I'd like to use the macro here if it's not too much work for you to change it.
```suggestion
        sqlx::query!(
```

netrome (approved):
A few nits and questions, but generally looks good to me. Thank you!

signer/src/storage/postgres.rs
djordon: It might be possible, but I couldn't get it to work by going directly to `BYTEA` from a `[u8; N]` once I serialized to JSONB. Let me try one more way; I'm going to try deserializing to `int[]` and then convert that to a `BYTEA`.

signer/src/storage.rs
djordon: You're right. If we want to implement `DbRead` or `DbWrite` for `sqlx::Transaction` (or something wrapping it) then we'll need to change these traits to require `&mut self`. If we want transaction support I'm okay with changing the API. But we're okay if we just want transaction functionality within the function implementation since we only need a reference to a `PgPool` to create a `sqlx::Transaction`.

---

My personal preferences (loosely held) is to support a (wrapped) `sqlx::Transaction` implementation but only for testing purposes. But this is to avoid `sqlx` macros like `#[sqlx::test]` and `sqlx::query!`. Basically, I'm still conflicted about the utility of the statically checked queries since it doesn't help more than an integration test would (which I think we should always add). Hopefully this helps explain why I didn't use the `sqlx::query` in this PR.

netrome: Makes sense. I'm not sure if we need transactional support, but I'd err on the side of caution and use `&mut` here so we retain the ability to use it and not starting to rely on these references to be immutable.

signer/src/storage/postgres.rs
netrome: Right, though couldn't we let the `block_id` type here just be a `Vec<u8>` to keep the type aligned with what the database expects?

signer/src/storage.rs
netrome: > Basically, I'm still conflicted about the utility of the statically checked queries since it doesn't help more than an integration test would (which I think we should always add). Hopefully this helps explain why I didn't use the sqlx::query in this PR.

That's a fair point. What I like about the queries is that I can trust the query to be correct without having to look if there is an integration test, which makes it much faster to verify the code when reading it. I agree that having an integration test provides strictly more utility, and the macros shouldn't substitute it. Though you will need the database up when running the integration test so at that point it should be fairly easy to also run `cargo sqlx prepare`, so it shouldn't be too much extra effort to use the macros either.

signer/src/storage/postgres.rs
djordon: I'm not sure if the array versus vec difference matters. I'll try though.

djordon: Yeah I couldn't get it to work in the way that I like. We can send `Vec<u8>` but then we need to do something complicated to convert that to a `BYTEA`. Here's an example of a query that will work but don't like:
```sql
INSERT INTO sbtc_signer.transactions (txid, tx, tx_type, created_at)
SELECT
       decode((SELECT string_agg(lpad(to_hex(x), 2, '0'), '') FROM unnest(txid) AS x), 'hex')
     , decode((SELECT string_agg(lpad(to_hex(x), 2, '0'), '') FROM unnest(tx) AS x), 'hex')
     , tx_type::sbtc_signer.transaction_type
     , CURRENT_TIMESTAMP
FROM JSONB_TO_RECORDSET($1::JSONB) AS x(
      txid      int[]
    , tx        int[]
    , tx_type   VARCHAR
    )
ON CONFLICT DO NOTHING
```
This looks unnecessarily complicated to me and I don't think I can do better if we're starting from an `int[]`.

signer/src/storage.rs
djordon: > Though you will need the database up when running the integration test so at that point it should be fairly easy to also run `cargo sqlx prepare`, so it shouldn't be too much extra effort to use the macros either.

True, it's not much extra effort.

---

> What I like about the queries is that I can trust the query to be correct without having to look if there is an integration test, which makes it much faster to verify the code when reading it. 

Yeah for some queries this makes sense and works perfectly, but sometimes not so much. For the queries added here, it's not possible to know if things will work as intended without an integration test.

For example, in this PR one of my original queries looked like this:
```sql
INSERT INTO sbtc_signer.stacks_transactions (txid, block_hash)
SELECT
    txid
  , block_id
FROM JSONB_TO_RECORDSET($1::JSONB) AS x(
    txid        BYTEA
  , block_id    BYTEA
)
ON CONFLICT DO NOTHING
```

Turns out that this will pass any type checks and run successfully but won't do what we want it to do (probably because we go through the `JSONB_TO_RECORDSET` function). Because things don't always work as expected, I'm wary of relying solely on type checks, even for simple queries.

netrome: True. Type checks can tell you that something is wrong but it won't guarantee that it is correct.

signer/src/storage/postgres.rs
netrome: Wait, I'm not following here. I thought we could just dump a `Vec<u8>` into the database as a `BYTEA` type. That's what we already do for bitcoin blocks - which just looks like
```
        sqlx::query!(
            "INSERT INTO sbtc_signer.bitcoin_blocks VALUES ($1, $2, $3, $4, $5)",
            block.block_hash,
            block.block_height,
            block.parent_hash,
            block.confirms,
            block.created_at
        )
```

where `block_hash` is a Vec<u8> which gets written as a BYTEA here. This is tested in the `should_be_able_to_query_bitcoin_blocks` so I'm pretty sure it works.

djordon: It works for when you directly encode the parameter, but not when you go through `JSONB_TO_RECORDSET`. So the function with the above query works (and `stacks_block_exists` works) but the other queries do not.

signer/src/storage.rs
djordon: Yeah, I know you know that. I just wanted to highlight a specific simple example where things don't quite work.

netrome: Yeah it's a good example, thanks for highlighting!

signer/src/storage/postgres.rs
netrome: Right, because you encode the records as a single JSON array before adding it to the query. Which I guess is because you want to write all transactions as a single query as opposed to just writing them one by one? I'd expect the latter to be cleaner but the former to be more efficient - though using the latter with a transaction might be the most efficient option.

PR #253, opened 2024-06-24
feat: Proper coinbase generation

### LINKED ISSUE #213: [Chore]: Proper coinbase generation in dummy block generator
### Context
The `testing::dummy::block` function currently generates blocks with an invalid coinbase transaction. It currently does the bare minimum to allow the `Block::bip34_block_height` method to succeed, which is needed for the block observer tests. However, this can lead to unintentional problems going forward so we should ensure that we create a proper coinbase transaction in this utility function.

### Definition of done
The `testing::dummy::block` function creates a proper coinbase transaction for the block.

netrome:
closes #213

### DIFF
--- a/signer/src/testing/dummy.rs
+++ b/signer/src/testing/dummy.rs
@@ -3,6 +3,7 @@
 use bitcoin::hashes::Hash;
 use blockstack_lib::chainstate::{nakamoto, stacks};
 use fake::Fake;
+use rand::Rng;
 
 /// Dummy block
 pub fn block<R: rand::RngCore + ?Sized>(config: &fake::Faker, rng: &mut R) -> bitcoin::Block {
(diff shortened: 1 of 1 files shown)


signer/src/testing/dummy.rs
```diff
@@ -161,3 +151,25 @@ pub fn signature<R: rand::RngCore + ?Sized>(
...
+/// Coinbase transaction with random block height
+fn coinbase_tx<R: rand::RngCore + ?Sized>(
+    config: &fake::Faker,
+    rng: &mut R,
+) -> bitcoin::Transaction {
+    // Numbers below 17 are encoded differently which messes with the block height decoding
+    let min_block_height = 17;
+    let block_height_span = 10000;
+    let block_height =
+        config.fake_with_rng::<i64, _>(rng).abs() % block_height_span + min_block_height;
```
djordon, line 164: If we import `rand::Rng` then this can be simplified to
```rust
let block_height: i64 = rng.gen_range(17..10000);
```

djordon (approved):
Looks good!

signer/src/testing/dummy.rs
netrome: Thanks for the recommendation. Updated.

PR #259, opened 2024-06-25
Emily API Bulk PR Draft - Will Be Broken Down

AshtonStephens:
> This PR was scrapped in favor of pivoting the EmilyAPI to utilize `utoipa` over `smithy`. https://github.com/stacks-network/sbtc/pull/266 is the first ticket following that decision.

## Description

This is a draft PR that isn't ready to be merged. **It IS ready to be reviewed for structure**. It will not currently pass testing nor is it fully tested.

I'll be breaking this and subsequent prs for Emily into a few PRs:
- AWS CDK template updates
- Config class
- Resource definitions and operation definitions  for POST and GET requests for the chainstate endpoint
- Resource definitions and operation definitions for POST and GET requests for the deposit endpoint
- Resource definitions and operation definitions for POST and GET requests for the withdrawal endpoint
- operation definitions for PUT requests for deposit endpoint
- operation definitions for PUT requests for withdrawal endpoint
- operation definitions for PUT requests for chainstate endpoint


AshtonStephens:
Again note that this PR is very incomplete but does completely show the approach I intend to take for this.

emily/lambda/src/operations/chainstate.rs
```diff
@@ -2,50 +2,148 @@
...
+        // Multiple conflicting blocks found.
+        _ => {
+            let block_hashes = query_output.items()
+                .iter()
+                .map(|item| {
+                    item.get("BlockHash").unwrap().as_s().unwrap().clone()
+                })
+                .collect::<Vec<String>>();
+
+            Err(EmilyApiError::InternalService(
```
djordon, line 103: Why should the existence of multiple conflicting blocks with the same height lead to an error? This kind of thing should be "normal", since forking would lead to this outcome right? If so, maybe we should add a unix timestamp attribute to this table and just fetch the latest one if there are multiples?

emily/lambda/src/operations/chainstate.rs
```diff
@@ -2,50 +2,148 @@
...
+                ).to_string()
+            ))
+        }
+    }
 }
 
 /// Handles the update of an existing chainstate
-pub fn handle_update_chainstate(
-    body:  Option<String>
+pub async fn handle_update_chainstate(
```
djordon, line 115: This function seems identical to the create handler. Why not just remove it and reuse the other one? Also, why have a POST and a PUT endpoint if they do the same thing?

emily/lambda/src/operations/chainstate.rs
```diff
@@ -2,50 +2,148 @@
...
+    };
+
+    let mut item: HashMap<String, AttributeValue> = HashMap::new();
+    item.insert("BlockHeight".to_string(), AttributeValue::N(request.block_height.to_string()));
+    item.insert("BlockHash".to_string(), AttributeValue::S(request.block_hash));
+
+    context.dynamodb_client
+        .put_item()
+        .table_name(context.settings.chainstate_table_name.clone())
+        .set_item(Some(item))
```
djordon, line 32: We might want to add a [condition expression](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html#Expressions.ConditionExpressions.PreventingOverwrites) for whether the block exists by doing something like
```rust
    ...
    .put_item()
    .set_item(Some(item))
    .condition_expression("attribute_not_exists(block_hash)")
    ...
```

Addendum: Well, this only makes sense if this table has a timestamp column, since that is the only thing that is getting updated when we overwrite an entry that already exists. If timestamps are left out here, no need to include it.

emily/lambda/src/operations/chainstate.rs
djordon: Actually, I think that all of these tables should have a timestamp attribute for when they were created. And maybe another for when it was last updated too.

emily/lambda/src/operations/deposits.rs
```diff
@@ -2,184 +2,449 @@
...
+        fulfillment_bitcoin_txid: None,
+        fulfillment_bitcoin_tx_index: None,
+        fulfillment_btc_fee: None,
+        stacks_txid: None,
+    };
+
+    // Put entry into table.
+    context.dynamodb_client
+        .put_item()
+        .table_name(context.settings.deposit_table_name.clone())
```
djordon, line 67: I think you can just use a reference here instead of cloning it. It looks more appealing that way to me.

emily/lambda/src/resources/deposits.rs
```diff
@@ -0,0 +1,487 @@
...
+    pub stacks_block_height: u64,
+    /// Stacks block hash associated with the height of this update.
+    pub stacks_block_hash: String,
+}
+
+// Conversions ------------------------------------------------------
+
+/// Converts from the deposit key to a hashmap that can be used as a key
+/// for searching a DynamoDB table.
+impl From<DepositTableEntryKey> for HashMap<String, AttributeValue> {
```
djordon, line 166: I don't think this is necessary for any of these types. Instead, derive `Serialize` and `Deserialize` for the struct and just use `serde_dynamo::to_item` to transform the type to a `HashMap<String, AttributeValue>`. I think that will significantly simplify the code here.

emily/lambda/src/resources/deposits.rs
```diff
@@ -0,0 +1,487 @@
...
+
+use crate::errors::EmilyApiError;
+
+// ------------------------------------------------------------------
+// Api resources.
+// ------------------------------------------------------------------
+
+/// Deposit resource.
+#[derive(Clone)]
+pub struct DepositResource {
```
djordon, line 14: I'd probably remove all of these and just use the table entry structs. They seem unnecessary.

Also, I'd maybe rename some of the table entry structs. For example, maybe `DepositTableEntry` would be `DepositRequest`? I don't know, you could also keep the names the same.

djordon (commented):
Nice. I have some suggestions.
1. Remove the intermediate structs and just work on the raw structs that eventually get written to the database.
2. Derive `Serialize` and `Deserialize` and then use `serde_dynamo` for the strusts that are written to or read from DynamoDB.
3. Maybe use more expressive error variants to capture the information that you want and format the message using `thiserror` instead of when the error is created. I'm thinking something like https://github.com/stacks-network/sbtc/blob/f837f08fd04e39cfe670a2697e002fb6490c7f1b/signer/src/error.rs#L21-L23
4. Consider removing the update endpoint for the chainstate. Seems redundant.
5. Consider adding unix timestamp attributes to the tables (perhaps in some later PR).

emily/lambda/src/operations/chainstate.rs
AshtonStephens: The PUT one is specifically to update an existing chain link, so there will be more logic to it, the POST is explicitly an attempt to add a link that isn't already there.

djordon: > The PUT one is specifically to update an existing chain link, so there will be more logic to it, the POST is explicitly an attempt to add a link that isn't already there.

Oh right of course, that makes sense.

Thinking about this some more, I think we can remove the PUT call with a slightly different table: (1) add a timestamp for when the row was inserted; (2) Change the sort key to the timestamp.

Then everything can be a PutItem, including data when there is a fork. And for GetItem requests, only take the row with the most recent timestamp at the requested block height.

emily/lambda/src/operations/chainstate.rs
djordon: Yeah nevermind, this isn't needed.

emily/lambda/src/operations/chainstate.rs
```diff
@@ -2,50 +2,148 @@
...
+
+        // A single block found; happy path.
+        1 => {
+            // Extract.
+            let item = query_output.items().first().unwrap();
+
+            // TODO:
+            // extract fields more safely.
+            let block_hash = item.get("BlockHash").unwrap().as_s().unwrap();
+            let block_height = item.get("BlockHeight").unwrap().as_n().unwrap();
```
netrome, line 85: If we expect these conversions to happen often, we could break out some of the boilerplate by implementing `TryFrom<HashMap<String, AttributeValue>>` for `GetChainResponseContent`. Another alternative would be to create a custom trait for this conversion (`trait FromAttributeValueMap`) - which we might need since the `GetChainResponseContent` type seems to be external and the orphan rule would prohibit us from implementing an external trait on an external type.

emily/lambda/src/operations/chainstate.rs
```diff
@@ -2,50 +2,148 @@
...
+        common::deserialize_request::<models::UpdateChainstateRequestContent>(body)?;
+
+    let chainstate_entry: models::UpdateChainstateResponseContent = models::UpdateChainstateResponseContent {
+        block_hash: request.block_hash.clone(),
+        block_height: request.block_height,
+    };
+
+    let mut item: HashMap<String, AttributeValue> = HashMap::new();
+    item.insert("BlockHeight".to_string(), AttributeValue::N(request.block_height.to_string()));
+    item.insert("BlockHash".to_string(), AttributeValue::S(request.block_hash));
```
netrome, line 130: Similarly, we could add a `IntoAttributeValueMap` trait to break out this boilerplate from the business logic.

emily/lambda/src/resources/deposits.rs
```diff
@@ -0,0 +1,487 @@
...
+use emily::models::{self, DepositBasicInfo, OpStatus};
+
+use crate::errors::EmilyApiError;
+
+// ------------------------------------------------------------------
+// Api resources.
+// ------------------------------------------------------------------
+
+/// Deposit resource.
+#[derive(Clone)]
```
netrome, line 13: For these types of data types it's good to be proactive and implement some common traits (in the spirit of https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits)

```suggestion
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```

netrome: The same for the rest of the types here.

emily/lambda/src/resources/deposits.rs
```diff
@@ -0,0 +1,487 @@
...
+    fn from(deposit_table_entry: DepositTableEntry) -> Self {
+        (&deposit_table_entry).into()
+    }
+}
+
+/// Converts from Deposit Table Entry to Deposit Resource.
+impl From<&DepositTableEntry> for DepositResource {
+    fn from(deposit_table_entry: &DepositTableEntry) -> Self {
+        let latest_history_entry = deposit_table_entry.history
+            .last() // There must always be at least one entry.
```
netrome, line 242: How is this invariant enforced? I think we could use something like this crate https://crates.io/crates/nonempty - or define our own non empty Vec type, but I don't want to have a type conversion panic if we forget at some point in the code to maintain this invariant.

emily/lambda/src/operations/chainstate.rs
netrome: I see that you're using `TryFrom` a lot for the types in `resources/deposits.rs`. That's great, and we could define wrapper types here to be able to implement TryFrom for the external types as well.

```
struct GetChainstateResponseContent(models::GetChainResponseContent)

impl TryFrom<...> for GetChainstateResponseContent {...}

etc.
```

netrome (commented):
Nice stuff so far. Sharing some thoughts and comments.

emily/lambda/src/operations/chainstate.rs
netrome: Also if you use `serde_dynamo` as @djordon suggested that might be even better than the custom conversion definitions.

emily/lambda/src/operations/chainstate.rs
djordon: > So setting an item in dynamodb isn't idempotent, but will fail if the item already exists? I didn't know that. Good point!

Well, setting items in dynamodb is indeed idempotent. But if we add a timestamp attribute then the timestamp would likely change if the caller hits the endpoint twice. If we don't want to update that attribute then we'd have to add a condition expression.

netrome: Right, makes sense!

AshtonStephens: Instead of using a timestamp for this I'm using a "version" field that will increment whenever there's an update; I think this is the more standard approach for dynamodb.

emily/lambda/src/config.rs
```diff
@@ -0,0 +1,36 @@
...
+
+/// Lambda Context
+pub struct LambdaContext {
+    /// Lambda settings.
+    pub settings: Settings,
+    /// DynamoDB Client.
+    pub dynamodb_client: Client,
+}
+
+// TODO:
```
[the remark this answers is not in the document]
AshtonStephens: I normally agree, but I'm curious if your opinion changes when its for nits like this. something like this for example is not necessary but is nice to have done; I think that could possibly fall into "untracked work that someone can do if they see it" because there is an overhead risk to too many issues.

emily/lambda/src/config.rs
```diff
@@ -0,0 +1,36 @@
...
+// Create a `new` function similar to the settings in the blocklist client
+// and the Signer that grabs the configuration values from a local default
+// toml and potentially overwrites the fields with environment values.
```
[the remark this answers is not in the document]
AshtonStephens: The main way to configure a lambda is with configuration variables; when it's deployed it needs to know the name of the dynamodb table and that can only be autogenerated during deployment: https://github.com/stacks-network/sbtc/blob/main/emily/cdk/lib/emily-stack.ts#L180-L182

emily/lambda/src/errors.rs
```diff
@@ -63,17 +63,17 @@ impl EmilyApiError {
             EmilyApiError::BadRequest(_) =>
                 serde_json::to_string(&models::BadRequestErrorResponseContent { message: self.to_string() }),
             // EmilyApiError::Forbidden(_) =>
-            //     serde_json::to_string(&ForbiddenErrorResponseContent { message: self.to_string() }),
-            // EmilyApiError::NotFound(_) =>
-            //     serde_json::to_string(&NotFoundErrorResponseContent { message: self.to_string() }),
+            //     serde_json::to_string(&models::ForbiddenErrorResponseContent { message: self.to_string() }),
```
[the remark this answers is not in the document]
AshtonStephens: What's wrong with this that needs cleaning?

emily/lambda/src/main.rs
```diff
@@ -1,14 +1,47 @@
-use emily_lambda::eventhandler::handle_event;
+use std::env;
+
+use aws_config::BehaviorVersion;
+use aws_sdk_dynamodb::Client;
+use emily_lambda::{config::LambdaContext, eventhandler::handle_event};
```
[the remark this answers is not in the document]
AshtonStephens: That's awesome, I'll do that.

emily/lambda/src/operations/chainstate.rs
```diff
@@ -2,50 +2,148 @@
...
+        common::deserialize_request::<models::CreateChainstateRequestContent>(body)?;
+
+    let chainstate_entry = models::CreateChainstateResponseContent {
+        block_hash: request.block_hash.clone(),
+        block_height: request.block_height,
+    };
+
+    let mut item: HashMap<String, AttributeValue> = HashMap::new();
+    item.insert("BlockHeight".to_string(), AttributeValue::N(request.block_height.to_string()));
+    item.insert("BlockHash".to_string(), AttributeValue::S(request.block_hash));
```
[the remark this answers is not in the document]
AshtonStephens: I'll try that, thanks! I had been looking into [`serde_dynamodb`](https://docs.rs/serde_dynamodb/latest/serde_dynamodb/), which does not meet the bar. I'm so glad you suggested this...

emily/lambda/src/operations/chainstate.rs
AshtonStephens: Part of this is a permissions issue; we want people to be able to post a deposit but only the signers can update an existing deposit. We can use the same rust function call but it needs to be a different endpoint.

djordon: I had thought that we didn't have the permission issue for PUT vs POST endpoints for `chainstate` updates since both were private.

emily/lambda/src/config.rs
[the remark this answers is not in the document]
djordon: Gotcha, that makes sense. I thought this comment meant that we wanted to move away from environment variables and use a config file instead. I suppose we could use them both.

emily/lambda/src/operations/chainstate.rs
djordon: > Instead of using a timestamp for this I'm using a "version" field that will increment whenever there's an update; I think this is the more standard approach for dynamodb.

Ahhh I see. And that's why you want a PUT and a POST, you only need to do a get for the `version` with a PUT.

emily/lambda/src/config.rs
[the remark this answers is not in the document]
djordon: > and potentially overwrites the fields with environment values.

I'm just dense. Never mind.

emily/lambda/src/config.rs
```diff
@@ -0,0 +1,36 @@
...
+pub static WITHDRAWAL_TABLE_NAME: &str = "WithdrawalTable-xxxxxxxxxxxx-us-west-2-local";
+
+/// Temporary deposit Table name.
+pub static DEPOSIT_TABLE_NAME: &str = "DepositTable-xxxxxxxxxxxx-us-west-2-local";
+
+/// Emily lambda settings.
+#[derive(Debug)]
+pub struct Settings {
+    /// Whether the Emily lambda is running locally.
+    pub is_local: bool,
```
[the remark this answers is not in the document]
AshtonStephens: This is slightly more complicated than you think it'll be because of some weirdness in the way AWS Lambda spins up locally, so I'm going to put this as a TODO.

emily/lambda/src/utils.rs
```diff
@@ -1,4 +1,81 @@
...
+/// a hashmap encodes a number as a string.
+pub fn deserialize_string_to_number<'de, D, T>(deserializer: D) -> Result<T, D::Error>
+where
+    D: Deserializer<'de>,
+    T: FromStr,
+    T::Err: std::fmt::Display,
+{
+    let s = String::deserialize(deserializer)?;
+    s.parse::<T>().map_err(serde::de::Error::custom)
+}
```
AshtonStephens, line 24: How I've deserialized string representations of numbers.

PR #266, opened 2024-06-28
Pivot Emily Api Implementation to use Utoipa for OpenAPI Specification Generation.

AshtonStephens:
## Description

This is a transitional PR that begins the migration of the Emily API from using Smithy to generate an OpenAPI spec to using `utoipa` within the rust lambda handler to autogenerate the API. The Emily API is now a duplicate of the blocklist client code which implements a `utoipa` openapi spec. I'm doing this in order to verify that the components can interact properly when using utoipa and to get the hooks in place, which has all worked as is shown in this PR.

We made this decision after a few observations within https://github.com/stacks-network/sbtc/pull/259 which was to implement a lot of the Emily API code using the existing setup:
- The smithy API could only generate numbers of the type f64 which does not meet the API precision requirements.
- The API code was more complicated than it should be because of all the type conversions needed to support the peculiar autogeneration of the Rust openapi client.

I recommend ignoring all the BlockList client code *except* the `Cargo.toml` and the api generation `build.rs` and related code.

**Note:** The PR is broken up into concise parts within the individual commits, so reviewing one commit at a time may be easier.


AshtonStephens:
Why is the GitHub workflow `Rust and Clarity Tests` not running?

.generated-sources/emily/openapi/build.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    // Ensure that we rerun if the API changes or the build script changes.
+    println!("cargo:rerun-if-changed=../../../emily/handler/common/mod.rs");
+    println!("cargo:rerun-if-changed=../../../emily/handler/common/handlers.rs");
+    println!("cargo:rerun-if-changed=build.rs");
+
+    let mut api_doc = ApiDoc::openapi();
+    let new_extensions: HashMap<String, serde_json::Value> = new_operation_extensions();
+
+    // Add AWS extension to openapi specification so AWS CDK can attach the appropriate lambda endpoint.
```
djordon, line 29: Ahhh, so the AWS CDK cannot work with any old openapi spec, it needs one with this extension in there. Can you point to some docs that mention this?

Makefile
```diff
@@ -79,23 +91,42 @@ devenv: $(wildcard $(subst dir, devenv, $(TWO_DIRS_DEEP)))
...
 	@touch devenv
 
-ifneq ($(filter arm64 aarch64, $(shell uname -m)),)
-_LAMBDA_FLAGS := --arm64
-endif
+# Emily CDK Template ---------------------------------
+
+EMILY_CDK_SOURCE_FILES :=
+EMILY_CDK_SOURCE_FILES := $(wildcard $(subst dir, emily/cdk/lib, $(FIVE_DIRS_DEEP))) $(EMILY_CDK_SOURCE_FILES)
+EMILY_CDK_SOURCE_FILES := $(wildcard $(subst dir, emily/bin/lib, $(FIVE_DIRS_DEEP))) $(EMILY_CDK_SOURCE_FILES)
```
djordon, line 98: Why not change this to the following:
```makefile
EMILY_CDK_SOURCE_FILES := $(wildcard $(subst dir, emily/cdk/lib, $(FIVE_DIRS_DEEP)))
EMILY_CDK_SOURCE_FILES := $(wildcard $(subst dir, emily/bin/lib, $(FIVE_DIRS_DEEP))) $(EMILY_CDK_SOURCE_FILES)
```

This is more educational for me than anything, I don't know that much about what is or isn't legal in a `Makefile`.

Makefile
```diff
@@ -5,17 +5,24 @@
 # $(subst dir, emily/cdk/lib, $(THREE_DIRS_DEEP))
 # becomes
 # emily/cdk/lib/*  emily/cdk/lib/*/*  emily/cdk/lib/*/*/*
-THREE_DIRS_DEEP := dir/* dir/*/* dir/*/*/*
-TWO_DIRS_DEEP := dir/* dir/*/*
+ONE_DIR_DEEP    := dir/*
+TWO_DIRS_DEEP   := dir/* $(subst dir, dir/*, $(ONE_DIR_DEEP))
+THREE_DIRS_DEEP := dir/* $(subst dir, dir/*, $(TWO_DIRS_DEEP))
+FOUR_DIRS_DEEP  := dir/* $(subst dir, dir/*, $(THREE_DIRS_DEEP))
+FIVE_DIRS_DEEP  := dir/* $(subst dir, dir/*, $(FOUR_DIRS_DEEP))
```
djordon, line 12: Is there no other way to go deeper than something like this? Maybe we can use `find`? I mean, is there a simpler way.

djordon (approved):
Looks good! I have some questions about the `Makefile` changes, but it's because I don't know that stuff as well.

emily/cdk/lib/emily-stack.ts
```diff
@@ -196,42 +199,44 @@ export class EmilyStack extends cdk.Stack {
...
-        operationLambda.addPermission("ApiInvokeLambdaPermission", {
-            principal: new iam.ServicePrincipal("apigateway.amazonaws.com"),
-            action: "lambda:InvokeFunction",
-            sourceArn: restApi.arnForExecuteApi(),
-        });
-
-        // Return api resource.
-        return restApi;
-    }
+    // TODO: Uncomment after adding emily api spec generation.
```
netrome: Will this be done in the current PR or do you have a follow-up ticket for this?

netrome: (Applies to the other TODOs as well)

emily/handler/src/api/handlers.rs
```diff
@@ -0,0 +1,82 @@
...
+    ),
+    responses(
+    (status = 200, description = "Risk assessment retrieved successfully", body = BlocklistStatus),
+    (status = 400, description = "Invalid request body"),
+    (status = 404, description = "Address not found"),
+    (status = 405, description = "Method not allowed"),
+    (status = 500, description = "Internal server error")
+    )
+)]
+pub async fn check_address_handler(
```
netrome, line 30: I feel like there's a lot of unnecessary boilerplate included from the `blocklist_client` when it seems like this attribute is the only thing you seem to need for your purposes. Or do you need to also be able to run this to validate the approach?

netrome (approved):
Looks good to me, though it's not clear if you need all of the blocklist client boilerplate of it it would be enough to just add the `utoipa::path` attribute to an empty placeholder function.

emily/cdk/lib/emily-stack.ts
netrome: nvm, I see you've addressed these in later commits.

emily/handler/src/api/handlers.rs
AshtonStephens: No, but I'd rather merge this in with this extra stuff here so that each change is a single logical step. Here the logical step is "replace emily with the blocklist client and use it's autogeneration approach".

Makefile
AshtonStephens: It just seemed more idiomatic to me to be honest, but I might be inventing my own standard. Originally I had this sequence in front of a number of `Make` commands and it looked like this:

```Make
SOURCE_FILES :=
SOURCE_FILES := $(wildcard $(subst dir, emily/cdk/lib, $(FIVE_DIRS_DEEP))) $(SOURCE_FILES)
SOURCE_FILES := $(wildcard $(subst dir, emily/bin/lib, $(FIVE_DIRS_DEEP))) $(SOURCE_FILES)
```

And then I'd have that in front of every make command so that the longer make commands all followed the same structure. What do you think @djordon? Otherwise I can change these lines to be what you're suggesting.

.generated-sources/emily/openapi/build.rs
AshtonStephens: I cannot, I found this out with much trial and error.

I actually see this here that might make this unnecessary now, but I'd rather get to that later. https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.LambdaIntegration.html

I've added a `TODO`

Makefile
AshtonStephens: It seems like to use `find` I'd need to have the Makefile run a shell command ([stack overflow](https://stackoverflow.com/a/26694693)), and as a rule of thumb I try not to require multiple tools for building unless they're absolutely necessary. Since someone can be running this Makefile and maybe not have `find` I think it's maybe better to just use the tools available in Make. That was my logic at least ~

I realize that might not make sense given there are a bunch of other tools required to build the actual code; what do you think?

Makefile
djordon: Ahh I see. Well, they seem equivalent and equally clear so I have a slight preference for the shorter version, but if one is more idiomatic (even if it's your own standard) then we can go with that.

AshtonStephens: I changed it to your suggestion, no need to make up idioms :)

Makefile
AshtonStephens: I'm going to merge but we can refine this Makefile another day ~

Issue #262, opened 2024-06-27
[Feature]: Simplify stacks transaction signing


djordon:
# Feature - Simplify stacks transaction signing

### 1.1 Context & Purpose

The signers need to collectively sign a stacks transaction before submitting it to the network. To do that, we need to correctly formulate the proper digest. This ticket is to add functionality to make that processes easier (analogous to this: https://github.com/stacks-network/sbtc/issues/152)

### 2.1 Acceptance Criteria:
Stacks does not support Schnorr signatures yet, so the signers need to operate as a multi-sig wallet and authenticate using the appropriate spending conditions for the wallet (details on transaction authorization from [SIP-005](https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md#transaction-authorization)).

   - [ ] Add helper functions to simplify the process of signing a stacks transaction.

## 3. Related Issues and Pull Requests (optional):

Creating a transaction signed by a multi-sig wallet is difficult for stacks transactions in v2.5, since each signature must be generated serially in the proscribed order. But this might change post Nakamoto.

* sBTC Design discussion for multi-sig wallets https://github.com/stacks-network/sbtc/issues/59.
* Stacks SIP draft to support out-of-order signing of transactions for multi-sig addresse/wallets https://github.com/stacksgov/sips/pull/152.
* PR adding support to stacks-core for out-of-order signing of Stacks transactions https://github.com/stacks-network/stacks-core/pull/3710.

I hope order-independent-multisig-spending-conditions will land with Nakamoto, but I need to verify. If it does, it will significantly simplify signing Stacks transactions by the signers.

Issue #267, opened 2024-06-28
[Feature]: Document the process for signing Stacks transactions


djordon:
# Feature - Document the process for signing Stacks transactions

### 1.1 Context & Purpose

Add documentation explaining the flow for signing a transaction.

### 2.1 Acceptance Criteria:
   - [ ] Add documentation describing the transaction signing process for stacks transactions.

## 3. Related Issues and Pull Requests (optional):

Relates to https://github.com/stacks-network/sbtc/issues/262.

djordon:
We will close this for now and open a more targeted ticket about documentation later.

Issue #264, opened 2024-06-27
[Feature]: Add integration test for signing Stacks smart contracts


djordon:
# Feature - Add integration test for signing Stacks smart contracts
### 1.1 Context & Purpose

We should add tests for making a contract call signed by the signer' multi-sig wallet. Because of the complexities of getting Nakamoto running, these integration tests do not need to run in CI.

### 2.1 Acceptance Criteria:

- [ ] Add test(s) for submitting a stacks transaction.

## 4. Related Issues and Pull Requests (optional):

This is follow on work after https://github.com/stacks-network/sbtc/issues/262 and https://github.com/stacks-network/sbtc/issues/263.

Issue #263, opened 2024-06-27
[Feature]: Add integration test scaffolding for deploying smart contracts


djordon:
# Feature - Add integration test scaffolding for deploying smart contracts

### 1.1 Context & Purpose

Signing a stacks transaction is kind-of-complicated. Once we have the mechanism for easily signing stacks transactions, we should add test code for deploying the sBTC smart contracts and making a contract call. This ticket is for deploying the smart contracts during the integration test. Because of the complexities of getting Nakamoto running, these integration tests do not need to run in CI (it's not that hard, it just takes a bit of time).

### 2.1 Acceptance Criteria:
   - [ ] Add code for deploying the smart contract locally.

## 3. Related Issues and Pull Requests (optional):

Follow on work after https://github.com/stacks-network/sbtc/issues/262.

netrome:
I don't think we'll need to be able to deploy smart contracts for v0, just being able to sign contract calls should be enough afaik. Though it should be basically the same complexity though.

djordon:
> I don't think we'll need to be able to deploy smart contracts for v0, just being able to sign contract calls should be enough afaik. Though it should be very similar.

Oh I was referring to deploying the smart contract in order to run the integration test, not for production use.

djordon:
I updated the language to better reflect what I meant.

netrome:
Oh nice, thank you for clarifying!

PR #268, opened 2024-06-28
feat: simplify stacks transaction signing

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/262. What is left is documentation on what the flow is supposed to look like, which is what https://github.com/stacks-network/sbtc/issues/267 is for.

This PR simplifies part of the process of signing stacks transactions. The coordinator can use the code here to generate transactions that other signers can sign and combine the signatures into a final transaction. The final transaction that will (very likely) pass verification on a stacks-core node.

## Changes

1. Add functions for generating the multi-sig stacks address of the all signers participating in a PoX cycle.
2. Add a trait for specifying what is necessary to create a stacks contract call transaction.
3. Add traits and functions for signing a transaction.

## Testing information

There are unit tests that check that the final signed `StacksTransaction` will pass verification. The integration tests that check validation on a stacks node will be done in https://github.com/stacks-network/sbtc/issues/263 and https://github.com/stacks-network/sbtc/issues/264.


signer/src/stacks/wallet.rs
```diff
@@ -0,0 +1,537 @@
...
+use crate::config::NetworkKind;
+use crate::error::Error;
+use crate::stacks::contracts::AsContractCall;
+
+/// Requisite info for the signers' multi-sig wallet on Stacks.
+#[derive(Debug, Clone)]
+pub struct SignerWallet {
+    /// The current set of public keys for all known signers during this
+    /// PoX cycle. These values must be sorted.
+    public_keys: Vec<PublicKey>,
```
netrome, line 40: How about using a `BTreeSet` instead of a `Vec` to enforce that the values are sorted?

signer/src/stacks/wallet.rs
```diff
@@ -0,0 +1,537 @@
...
+    }
+
+    /// Return the stacks address for the signers
+    pub fn address(&self) -> StacksAddress {
+        let public_keys: Vec<Secp256k1PublicKey> = self
+            .public_keys
+            .iter()
+            .map(|pk| Secp256k1PublicKey::from_slice(&pk.serialize()))
+            .collect::<Result<_, _>>()
+            .expect("we know these are all valid public keys");
```
netrome, line 89: It seems to me like this conversion happens whenever we use `public_keys` within this context. Because the `SignerWallet` is a stacks-specific type, perhaps we should consider converting the public keys once in `SignerWallet::new` and internally store the Stacks variant of the `Secp256k1PublicKey` type instead.

netrome (approved):
Really nice stuff! My main consideration is how we represent the public keys in the wallet type, where I think it should be a `BTreeMap` of the stacks variant of the public key type in this case.

AshtonStephens (approved):
Looks good to me!

signer/src/stacks/wallet.rs
```diff
@@ -0,0 +1,537 @@
...
+    /// keys or if the list of public keys is empty.
+    pub fn new(
+        mut public_keys: Vec<PublicKey>,
+        signatures_required: u16,
+        network_kind: NetworkKind,
+    ) -> Result<Self, Error> {
+        public_keys.sort();
+
+        if public_keys.len() < signatures_required as usize || public_keys.is_empty() {
+            return Err(Error::MissingBlock);
```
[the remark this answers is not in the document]
djordon: 🙏🏽 Also, this is a mistake and I need to update the returned error.

signer/src/stacks/wallet.rs
djordon: That makes sense, will update.

signer/src/stacks/wallet.rs
djordon: Hmmm, good point. I'll give it a look. I'll change it if it ends up making things simpler.

Also, my motivation for using these types was that I wanted to rely on only one Secp256k1 library for internal use and do the conversions when necessary. But that decision might have lead to some incidental complexity so let me check.

djordon: So it looks like it is the same complexity with both approaches. When storing the `Secp256k1PublicKey` you get:
1. More complex `SignerWallet::new`.
1. Simpler `StacksAddress` creation.
2. More complex `MultisigTx::push_signature`.
3. Simpler `MultisigTx::finalize_transaction`.

Seems like a wash, at least for now.

PR #274, opened 2024-07-02
Add Emily scaffolding with Utoipa and Warp

AshtonStephens:
## Description

This PR removes most of (if not all of) the `BlocklistClient` specific code from the Emily Handler location and adds a number of endpoints.

The original blocklist routing method that considered errors rejections is not the way that warp intended rejections to be used (rejection means try the next filter, not error) so I've converted errors in the blocklist client handler to come out as a Reply directly.

## Checklist
- [x] `make test` passes
- [x] `make lint1 passes


netrome:
> The original blocklist routing method that considered errors rejections is not the way that warp intended rejections to be used (rejection means try the next filter, not error) so I've converted errors in the blocklist client handler to come out as a Reply directly.

Right, I should have noted this earlier. Great that you fixed it!

.generated-sources/emily/openapi/build.rs
```diff
@@ -7,11 +7,26 @@ use utoipa::OpenApi;
...
+        // Health check endpoints.
+        api::handlers::health::get_health,
+        // Deposit endpoints.
+        api::handlers::deposit::get_deposit,
+        api::handlers::deposit::get_deposits_for_transaction,
+        api::handlers::deposit::get_deposits,
+        api::handlers::deposit::create_deposit,
+        api::handlers::deposit::update_deposits,
+        // TODO: https://github.com/stacks-network/sbtc/issues/272
+        // Add withdrawal and chainstate endpoints.
```
netrome, line 20: Great to see TODOs referencing issues. I typically format these a bit more concisely as

```suggestion
        // TODO(272): Add withdrawal and chainstate endpoints.
```

.generated-sources/emily/openapi/build.rs
```diff
@@ -7,11 +7,26 @@ use utoipa::OpenApi;
...
+    ),
     components(schemas(
-        common::BlocklistStatus,
-        common::RiskSeverity,
-        common::error::ErrorResponse
+        // TODO: https://github.com/stacks-network/sbtc/issues/271
+        // Add request and response schemas.
```
netrome, line 24:
```suggestion
        // TODO(271): Add request and response schemas.
```

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -0,0 +1,129 @@
...
+    operation_id = "getDeposit",
+    path = "/deposit/{txid}/{index}",
+    params(
+        ("txid" = String, Path, description = "txid associated with the Deposit."),
+        ("index" = String, Path, description = "output index associated with the Deposit."),
+    ),
+    tag = "deposit",
+    responses(
+        // TODO: https://github.com/stacks-network/sbtc/issues/271
+        // Add success body.
```
netrome, line 19:
```suggestion
        // TODO(271): Add success body.
```

emily/handler/src/api/handlers/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+use warp::{http::StatusCode, Rejection, Reply};
+
+/// Deposit handlers.
+pub mod deposit;
+/// Health handlers.
+pub mod health;
+
+/// Central error handler for Warp rejections, converting them to appropriate HTTP responses.
+/// TODO: https://github.com/stacks-network/sbtc/issues/131
+/// Alter handler for Emily API.
```
netrome, line 17:
```suggestion
/// TODO(131): Alter handler for Emily API.
```

emily/handler/src/api/handlers/mod.rs
```diff
@@ -0,0 +1,51 @@
+//! Handlers for the blocklist client API
```
netrome, line 1:
```suggestion
//! Handlers for the emily API
```

emily/handler/Cargo.toml
```diff
@@ -21,6 +21,7 @@ tracing-attributes.workspace = true
 tracing-subscriber.workspace = true
 utoipa = "4.2.3"
 warp = "0.3"
+warp_lambda.workspace = true
```
netrome, line 24: TIL about `warp_lambda`. Is this just temporary or is the intention to use warp within the lambda functions?

It's not obvious to me what the benefit of doing this would be, because the pivot we discussed was just about running utoipa and API declarations/routing did not seem to be a pain point with the previous setup. And we shouldn't need warp to use utoipa.

However, if we could build `Emily` as a pure warp service and then isolate `warp_lambda` and all the lambda specific boilerplate to only the production builds - this would make it much simpler to run Emily locally. So that would be amazing.

Would love to hear your thoughts here.

emily/handler/src/api/routes/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+/// Deposit routes.
+pub mod deposit;
+/// Health routes.
+pub mod health;
+
+/// This function sets up the Warp filters for handling all requests.
+pub fn routes() -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
+    // TODO: https://github.com/stacks-network/sbtc/issues/273
+    // Remove the "local" prefix once we figure out why all local testing
+    // calls seem to forcibly start with `local`.
```
netrome, line 15:
```suggestion
    // TODO(273): Remove the "local" prefix once we figure out why all local testing
    // calls seem to forcibly start with `local`.
```

emily/handler/src/api/handlers/health.rs
```diff
@@ -0,0 +1,26 @@
...
+
+/// Get health handler.
+#[utoipa::path(
+    get,
+    operation_id = "checkHealth",
+    path = "/health",
+    tag = "health",
+    responses(
+        // TODO: https://github.com/stacks-network/sbtc/issues/271
+        // Add success body.
```
netrome, line 14:
```suggestion
        // TODO(271): Add success body.
```

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -0,0 +1,129 @@
...
+
+/// Update deposits handler.
+#[utoipa::path(
+    put,
+    operation_id = "updateDeposits",
+    path = "/deposit",
+    tag = "deposit",
+    responses(
+        // TODO: https://github.com/stacks-network/sbtc/issues/271
+        // Add success body.
```
netrome, line 116:
```suggestion
        // TODO(271): Add success body.
```

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -0,0 +1,129 @@
...
+
+/// Create deposit handler.
+#[utoipa::path(
+    post,
+    operation_id = "createDeposit",
+    path = "/deposit",
+    tag = "deposit",
+    responses(
+        // TODO: https://github.com/stacks-network/sbtc/issues/271
+        // Add success body.
```
netrome, line 93:
```suggestion
        // TODO(271): Add success body.
```

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -0,0 +1,129 @@
...
+
+/// Get deposits handler.
+#[utoipa::path(
+    get,
+    operation_id = "getDeposits",
+    path = "/deposit",
+    tag = "deposit",
+    responses(
+        // TODO: https://github.com/stacks-network/sbtc/issues/271
+        // Add success body.
```
netrome, line 70:
```suggestion
        // TODO(271): Add success body.
```

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -0,0 +1,129 @@
...
+    get,
+    operation_id = "getDepositsForTransaction",
+    path = "/deposit/{txid}",
+    params(
+        ("txid" = String, Path, description = "txid associated with the Deposit."),
+    ),
+    tag = "deposit",
+    responses(
+        // TODO: https://github.com/stacks-network/sbtc/issues/271
+        // Add success body.
```
netrome, line 46:
```suggestion
        // TODO(271): Add success body.
```

blocklist-client/src/common/error.rs
```diff
@@ -3,10 +3,11 @@
 use reqwest::StatusCode;
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
-use warp::reject::Reject;
+use warp::{reject::Reject, reply::Reply};
 
 /// Errors occurring from Blocklist client's API calls to risk client and request handling
 #[derive(thiserror::Error, Debug)]
+#[allow(dead_code)]
```
netrome, line 10: Is this still needed? Which code is dead?

netrome (approved):
Good stuff.

My main question is about pivoting to warp in addition to the utoipa pivot.

AshtonStephens:
I switched to warp because it handles all the parameter extraction and error handling as well which is a big improvement over the manual stuff I had it doing before.

blocklist-client/src/common/error.rs
AshtonStephens: I like it for enums but I'll remove it.

emily/handler/Cargo.toml
AshtonStephens: I think running it as a warp service is something for the future, within the suggestion that I pivot emily to `utoipa` was pivoting the whole thing to use standard Rust libraries. So far it's already a big improvement because there's much less bespoke serialization and deserialization.

I do hear what you're saying, I think it this moment my goal is to move quickly and get what we know will work working.

I'll say, using warp is already saving a LOT of time, I should have known when I started.

emily/handler/src/main.rs
```diff
@@ -1,34 +1,23 @@
...
-        Ok(mut addrs) => addrs.next().expect("No addresses found"),
-        Err(e) => {
-            error!("Failed to resolve address: {}", e);
-            return;
-        }
-    };
+    let warp_service = warp::service(routes);
 
-    warp::serve(routes).run(addr).await;
+    warp_lambda::run(warp_service)
```
djordon, line 20: Ahhh I see, this is the only place where we use `warp_lambda`.

I think things are making more sense to me now. I was looking through the official [aws-lambda-rust-runtime](https://github.com/awslabs/aws-lambda-rust-runtime/tree/main) crate to see how things are typically done. There they have some examples, where they build the HTTP router using the framework's mechanism and just call [`lambda_http::run(router)`](https://docs.rs/lambda_http/latest/lambda_http/fn.run.html) in `main`. That works for them since the frameworks they use in their [examples](https://github.com/awslabs/aws-lambda-rust-runtime/blob/c4594f797c9c3a133845298eaa0a2f896ebe4220/examples/http-axum/src/main.rs) (typically with [`axum`](https://github.com/tokio-rs/axum)) are built on `tower`, and the router type in the framework implements the `tower` service traits. I believe `warp::service(routes)` returns a tower service, which means I think we can use `lambda_http::run(warp_service)` here as well (or something close to that). I'd prefer using `lambda_http` over `warp_lambda`, since the former is an official AWS supported crate. Moreover, `warp_lambda` just uses `lambda_http` [under the hood](https://docs.rs/warp_lambda/latest/src/warp_lambda/lib.rs.html#35-41), so it seems unnecessary.

djordon (commented):
Looks good. My only hesitation is that I don't think we should use `warp_lambda`, but instead use the official `lambda_http` crate.

blocklist-client/src/common/error.rs
AshtonStephens: Removed.

blocklist-client/src/api/handlers.rs
```diff
@@ -54,13 +58,6 @@ pub async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible>
         return Ok(warp::reply::with_status(json, StatusCode::BAD_REQUEST));
     }
 
-    if let Some(e) = err.find::<Error>() {
-        // Custom application errors
-        let (code, message) = e.as_http_response();
-        let json = warp::reply::json(&ErrorResponse { message });
-        return Ok(warp::reply::with_status(json, code));
-    }
```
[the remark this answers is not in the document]
AshtonStephens: Yes, this will never happen since the Errors are all replies. This will now only get called when there was no matching filter for the request.

emily/handler/src/common/error.rs
```diff
@@ -3,9 +3,10 @@
 use reqwest::StatusCode;
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
-use warp::reject::Reject;
+use warp::{filters::path::FullPath, reject::Reject, reply::Reply};
 
-/// Errors occurring from Blocklist client's API calls to risk client and request handling
+/// Errors from the internal API logic.
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
AshtonStephens: Yes sort of, I also think it's good to have these primed for when they become useful.

emily/handler/src/main.rs
AshtonStephens: Okay, I'll make a ticket to remove it but I'm eager to just get this whole thing working. 

I looked at the `warp_lambda` crate and was not super impressed, and then realized it didn't matter if it worked which is where I am right now.

djordon: Fair enough. Also, I might be mistaken about how easy it is to remove `warp_lambda`. As in I might be wrong about `warp::service` returning a tower service. I think it might actually return an hyper service or something.

AshtonStephens: Hang on, I see you're saying it's possibly as easy as a swap;  let me try it.

AshtonStephens: Okay, its more complicated than just a swap so I'm going to move it into a different ticket with a TODO since I don't want to spend time on that before the whole system works.

AshtonStephens: Made https://github.com/stacks-network/sbtc/issues/276

PR #275, opened 2024-07-02
feat: Transaction signer storing decisions for pending withdraw requests in Postgres

### LINKED ISSUE #246: [Feature]: Transaction signer storing decisions for pending withdraw requests in postgres
When the transaction signer makes a decision for a pending withdraw request, it should also persist that decision in postgres.

### Definition of done
The transaction signer persists decisions for pending withdraw requests.

netrome:
closes #246


signer/src/storage/postgres.rs
```diff
@@ -361,10 +396,74 @@ impl super::DbRead for PgStore {
...
+                FROM sbtc_signer.bitcoin_blocks parent
+                JOIN extended_context_window last ON parent.block_hash = last.parent_hash
+                WHERE last.depth <= $3 + 1
+            ),
+            last_bitcoin_block AS (
+                SELECT
+                    block_hash
+                  , confirms
+                FROM extended_context_window
+                WHERE depth = $3 + 1
```
djordon, line 426: I would go with an `order by` and `limit 1` clause here, mainly to be resilient to the start-up case where the `context_window` is greater than the number of bitcoin blocks in the database. I mean I'd replace the `where` clause with
```sql
ORDER BY depth DESC
LIMIT 1
```

signer/src/storage/postgres.rs
```diff
@@ -361,10 +396,74 @@ impl super::DbRead for PgStore {
...
+            WITH RECURSIVE extended_context_window AS (
+                SELECT block_hash, parent_hash, confirms, created_at, 1 AS depth
+                FROM sbtc_signer.bitcoin_blocks
+                WHERE block_hash = $1
+                UNION ALL
+                SELECT
+                    parent.block_hash
+                  , parent.parent_hash
+                  , parent.confirms
+                  , parent.created_at
```
djordon, line 415: Looks like these `created_at` columns are unnecessary.

signer/src/storage/postgres.rs
```diff
@@ -361,10 +396,74 @@ impl super::DbRead for PgStore {
 
     async fn get_pending_withdraw_requests(
         &self,
-        _chain_tip: &model::BitcoinBlockHash,
-        _context_window: usize,
+        chain_tip: &model::BitcoinBlockHash,
```
djordon, line 399: Why take this as input? Don't we always want to fetch pending withdrawals referenced from the bitcoin `chain_tip`?

signer/src/storage/postgres.rs
```diff
@@ -361,10 +396,74 @@ impl super::DbRead for PgStore {
...
+                    parent.block_hash
+                  , parent.block_height
+                  , parent.parent_hash
+                  , parent.created_at
+                FROM sbtc_signer.stacks_blocks parent
+                JOIN StacksContext last
+                        ON parent.block_hash = last.parent_hash
+                LEFT JOIN last_bitcoin_block block
+                        ON block.confirms @> ARRAY[parent.block_hash]
+                WHERE block.block_hash IS NULL
```
djordon, line 447: Ohh I see. So here we fetch all stacks blocks that start at the latest stacks block that has been confirmed by a bitcoin block and end at the end of the context window.

I had thought that it would be simpler than that, where we fetch all stacks blocks starting at the chain tip (regardless of whether it was confirmed by a bitcoin block) and end at the end of the context window.

djordon: Also, I think you may have mentioned this recently, but if we want stacks blocks from the latest stacks block confirmed on bitcoin, why not do something like the following, which explicitly joins on the stacks blocks confirmed on bitcoin starting from the bitcoin chain tip.
```sql
WITH RECURSIVE bitcoin_context_window AS (
    SELECT
        block_hash
      , parent_hash
      , UNNEST(confirms) AS stacks_hash
      , 1                AS depth
    FROM sbtc_signer.bitcoin_blocks AS bb
    WHERE block_hash = -- chain tip hash below
          '\xF13F8F587B670909CB8EDF864838D579608A2A8E531841C78595ADA1A3E60196'

    UNION ALL

    SELECT
        parent.block_hash
      , parent.parent_hash
      , UNNEST(parent.confirms) AS stacks_hash
      , last.depth + 1
    FROM sbtc_signer.bitcoin_blocks parent
    JOIN bitcoin_context_window AS last
         ON parent.block_hash = last.parent_hash
    WHERE last.depth <= 8 -- + 1
)
SELECT
    wr.request_id
  , wr.block_hash
  , wr.recipient
  , wr.amount
  , wr.max_fee
  , wr.sender_address
  , wr.created_at
FROM sbtc_signer.withdraw_requests wr
JOIN bitcoin_context_window sc ON wr.block_hash = sc.stacks_hash
```

signer/src/storage/postgres.rs
```diff
@@ -190,6 +190,41 @@ impl PgStore {
...
+                WHERE bitcoin_blocks.block_hash = $1
+            )
+            SELECT
+                block_hash
+              , block_height
+              , parent_hash
+              , created_at
+            FROM confirmed_by_chain_tip
+            ORDER BY block_height DESC, block_hash DESC
+            LIMIT 1;
```
djordon, line 219: Can't we use the order by and limit 1 without the CTE?

djordon (commented):
Very nice. I had a few questions about which stacks blocks we want to use when fetching pending withdrawal requests. I had thought we would use all stacks blocks but maybe not? I feel like adding some comments about that would be useful down the line, since it is not obvious (well, maybe just not obvious to me).

netrome:
> Very nice. I had a few questions about which stacks blocks we want to use when fetching pending withdrawal requests. I had thought we would use all stacks blocks but maybe not? I feel like adding some comments about that would be useful down the line, since it is not obvious (well, maybe just not obvious to me).

Yeah it's definitely not obvious. Right now I chose to define the "canonical stacks chain tip" to be the confirmed stacks block with the highest block height - and only consider ancestors of this block. I would expect in practice that the bitcoin blocks typically confirm a single stacks block, and in rare cases would confirm more than one. However, my hunch can be wrong, so I'm definitely open to having the logic based on fetching all confirmed stacks blocks, but that can be left as a follow-up since it's not obvious to me right now if it would be a significant improvement to the behavior.

signer/src/storage/postgres.rs
netrome: Nice observation. I'll clear it out.

netrome: Done.

signer/src/storage/postgres.rs
```diff
@@ -361,10 +396,74 @@ impl super::DbRead for PgStore {
...
+        let stacks_chain_tip = self.get_stacks_chain_tip(chain_tip).await?;
+        sqlx::query_as!(
+            model::WithdrawRequest,
+            r#"
+            WITH RECURSIVE extended_context_window AS (
+                SELECT block_hash, parent_hash, confirms, created_at, 1 AS depth
+                FROM sbtc_signer.bitcoin_blocks
+                WHERE block_hash = $1
+                UNION ALL
+                SELECT
```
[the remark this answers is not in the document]
netrome: Sure thing, updated.

signer/src/storage/postgres.rs
```diff
@@ -361,10 +396,74 @@ impl super::DbRead for PgStore {
...
+                WHERE last.depth <= $3 + 1
+            ),
+            last_bitcoin_block AS (
+                SELECT
+                    block_hash
+                  , confirms
+                FROM extended_context_window
+                WHERE depth = $3 + 1
+            ),
+            StacksContext AS (
```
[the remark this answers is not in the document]
netrome: Thanks for spotting. Updated.

signer/src/testing/transaction_signer.rs
```diff
@@ -275,7 +275,7 @@ where
             .expect("storage failure")
             .expect("missing block");
 
-        let context_window_end_block = futures::stream::iter(0..=context_window)
+        let context_window_end_block = futures::stream::iter(0..context_window)
```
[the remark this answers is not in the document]
netrome: The advantage of implementing the same logic twice 😅

signer/src/storage/postgres.rs
netrome: I think it makes the behavior easier to reason about since we can control so that the transaction has the same bitcoin chain tip across iterations. Otherwise we could make a query for one chain tip, and then the chain tip changes before the next query. I'm not sure it would cause any problems, but it feels simpler to be able to control it.

signer/src/storage/postgres.rs
netrome: Thank you for suggesting an alternative query. I ran the `should_return_the_same_pending_withdraw_requests_as_in_memory_store` test with this query now, and this suggestion under-fetches withdraw requests.

The way I understand this is that it joins the bitcoin context window with the stacks blocks confirmed by the blocks in the context window. In that case we're missing all the stacks blocks that lies between stacks blocks confirmed by bitcoin, which should explain the under-fetching.

However, if you can think of a more clean/elegant formulation of the query that also joins the stacks blocks between the bitcoin blocks I'm happy to update it.

signer/src/storage/postgres.rs
netrome: Nice, thanks for suggesting. I've updated now.

signer/src/storage/postgres.rs
netrome: Yes we can, and we will!

netrome: And we have!

signer/src/storage/postgres.rs
djordon: Ahh I see, yeah that makes sense.

signer/src/storage/postgres.rs
djordon: > The way I understand this is that the suggested query joins the bitcoin context window with the stacks blocks confirmed by the blocks in the context window. In that case we're missing all the stacks blocks that lies between stacks blocks confirmed by bitcoin, which should explain the under-fetching.

Oh interesting, there could be blocks between the stacks blocks confirmed by bitcoin? Did not know that, so glad I brought it up.

djordon: But I still have the question from above which blocks to fetch https://github.com/stacks-network/sbtc/pull/275#discussion_r1662925032.

djordon: lol you answered it below. never mind.

djordon:
> I would expect in practice that the bitcoin blocks typically confirm a single stacks block, and in rare cases would confirm more than one. However, my hunch can be wrong, so I'm definitely open to having the logic based on fetching all confirmed stacks blocks[.]

Oh, I thought that post-Nakamoto, there will be many stacks blocks confirmed by a single bitcoin blocks, since they are created every 15-45 seconds and miners still get rewards per stacks block (or something like that). So I would expect many stacks blocks between bitcoin blocks. But it's very possible that I got something important about this wrong.

> [B]ut that can be left as a follow-up since it's not obvious to me right now if it would be a significant improvement to the behavior.

I'm okay with leaving it as a follow-up, it's not essential behavior.

djordon (approved):
Looks good ✅

The only thing outstanding is whether we want to fetch additional pending requests on stacks blocks that happen after the blocks confirmed by bitcoin blocks, and we can do that later if need be.

netrome:
> Oh, I thought that post-Nakamoto, there will be many stacks blocks confirmed by a single bitcoin blocks, since they are created every 15-45 seconds and miners still get rewards per stacks block (or something like that). So I would expect many stacks blocks between bitcoin blocks. But it's very possible that I got something important about this wrong.

Right I think I have been a bit inconsistent with the terminology here but you are correct. Let's separate between "directly confirmed" and "indirectly confirmed". What we have in the data model are blocks that are directly confirmed, i.e. the bitcoin block points directly to a stacks block. I.e. if you know the bitcoin block you will know of the existence of a set of stacks blocks. This should typically be one since in the normal scenario, stacks miners should have a fairly consistent view of the chains. However, between the directly confirmed blocks we'll have a lot of stacks blocks that are ancestors of a confirmed block, which are all indirectly confirmed by that bitcoin block. This is not directly parse able from the bitcoin block, but something we can derive by traversing the ancestors of a stacks block until we encounter a block confirmed by a previous bitcoin block.

netrome:
> The only thing outstanding is whether we want to fetch additional pending requests on stacks blocks that happen after the blocks confirmed by bitcoin blocks, and we can do that later if need be.

Yeah I don't think that's going to be necessary, since we'll anyway need to await 6 confirmations on bitcoin before fulfilling a withdraw request.

signer/src/storage/postgres.rs
netrome: Also I realized the coordinator will want to fetch withdraw requests using a context window starting 6 blocks from the chain tip, so there is a use case for not always fetching the latest as well :)

Issue #231, opened 2024-06-05
[Feature]: Add support for broadcasting sBTC burn transactions


djordon:
# Feature - Add support for broadcasting sBTC burn transactions

### 1.1 Context & Purpose

An important part of the sBTC deposit-withdrawal lifecycle done by the signers is calling the stacks contracts to mint and burn sBTC. To do that, we need to support broadcasting stacks contract call transactions. This ticket is for adding support for broadcasting the sBTC burn stacks transaction.

### 2.1 Acceptance Criteria:

   - [ ] Given the sBTC burn contract address and the burn function interface, add support for making a properly formatted request to a stacks node for the contract call. The transaction does not need to contain a valid signature.

This ticket only relates to the "withdrawal-accept" contract call, which relates to https://github.com/stacks-network/sbtc/issues/96.

## 3. Related Issues and Pull Requests (optional):

This is part of #44 and recently came up during this discussion https://github.com/stacks-network/sbtc/pull/228#discussion_r1627170146.

* Withdrawal contract registry https://github.com/stacks-network/sbtc/issues/115.
* Withdrawal contract data structures https://github.com/stacks-network/sbtc/issues/94 https://github.com/stacks-network/sbtc/pull/127.
* Stacks client ticket https://github.com/stacks-network/sbtc/issues/176.

PR #281, opened 2024-07-02
feat: support broadcasting sbtc burn transactions

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/231. This is basically https://github.com/stacks-network/sbtc/pull/270 but for a different contract call.

We need to make it possible to make the contract call that burns sBTC. This PR adds structs that simplify construction of a stacks transaction that calls the `accept-withdrawal-request` function from our `sbtc-withdrawl.clar` contract that we have an open PR for in https://github.com/stacks-network/sbtc/pull/243.

## Changes

1. Add a struct that implements the `AsContractCall` trait that allows us to easily make the `accept-withdrawal-request` contract call.

## Testing information

We add a test that checks that we do not panic when creating the `TransactionContractCall`. If this test fails, then we know that we have a typo in the contract or function names.


netrome (approved):
Beautiful!

Issue #280, opened 2024-07-02
[Feature]: Add support for broadcasting withdrawal reject sBTC transactions


djordon:
# Feature - Add support for broadcasting withdrawal reject sBTC transactions

### 1.1 Context & Purpose

If the signers agree to reject the withdrawal request, the need to make a contract call noting this fact. This ticket is for adding support for broadcasting the sBTC reject-withdrawal stacks transaction.

### 2.1 Acceptance Criteria:

   - [ ] Given the sBTC contract address and the reject-withdrawl function interface, add support for making a properly formatted request to a stacks node for the contract call. The transaction does not need to contain a valid signature.

This ticket only relates to the "reject-withdrawal" contract call, which relates to https://github.com/stacks-network/sbtc/issues/97.

## 3. Related Issues and Pull Requests (optional):

This is a sister issue to https://github.com/stacks-network/sbtc/issues/230 and https://github.com/stacks-network/sbtc/issues/231.

PR #282, opened 2024-07-02
feat: support withdrawal-reject sbtc transactions

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/280. This is basically https://github.com/stacks-network/sbtc/pull/270 and https://github.com/stacks-network/sbtc/pull/281 but for a different contract call and a refactor.

We need to make it possible to make the contract call that burns sBTC. This PR adds structs that simplify construction of a stacks transaction that calls the `reject-withdrawal` function from our `sbtc-withdrawl.clar` contract that we have an open PR for in https://github.com/stacks-network/sbtc/pull/243.

## Changes

1. Add a struct that implements the `AsContractCall` trait that allows us to easily make the `reject-withdrawal` contract call.
2. Refactor the `AsContractCall` trait to make it simpler.

## Testing information

We add a test that checks that we do not panic when creating the `TransactionContractCall`. If this test fails, then we know that we have a typo in the contract or function names.


signer/src/stacks/contracts.rs
```diff
@@ -46,15 +49,40 @@ pub struct StacksTxPostConditions {
...
 
-/// A trait to ease construction of a StacksTransaction making sBTC related contract calls.
+/// A trait to ease construction of a StacksTransaction making sBTC related
+/// contract calls.
 pub trait AsContractCall {
-    /// Converts this struct to a Stacks contract call. The deployer is the
+    /// The name of the clarity smart contract that relates to this struct.
+    fn contract_name() -> &'static str;
+    /// The specific function name that relates to this struct.
+    fn function_name() -> &'static str;
```
netrome, line 58: These could be constants, as you previously had in the `AcceptWithdrawalV1` struct
```suggestion
    /// The name of the clarity smart contract that relates to this struct.
    const CONTRACT_NAME: &'static str;
    /// The specific function name that relates to this struct.
    const FUNCTION_NAME: 'static str;
```

netrome (approved):
Nice stuff, just one suggestion.

signer/src/stacks/contracts.rs
djordon: Oh for some reason I didn't think we could do that

netrome (approved):

Issue #301, opened 2024-07-05
[Feature]: Support fetching the nonce from the stacks node


djordon:
# Feature - Support fetching the nonce from the stacks node

### 1.1 Context & Purpose

Since we have a decentralized multi-sig, we should rely on the stacks node for the nonce value. If it's wrong, our submitted transactions won't make it to the mempool.

The `GET /v2/accounts/<principal>` [endpoint](https://github.com/stacks-network/stacks-core/blob/master/docs/rpc-endpoints.md#get-v2accountsprincipal) can be used to get this information.

### 2.1 Acceptance Criteria:
   - [ ] Implement a function that can make `GET /v2/accounts/<principal>` requests.

PR #302, opened 2024-07-05
feat: fetching nonce from stacks node

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/301. Add a function that gets account information, including the nonce.

## Changes

* Adds a function that fetches the account info from a stacks node.

## Testing information

There is an integration test that does not run in CI. It was tested against a local stacks node.

### DIFF
--- a/signer/src/stacks/api.rs
+++ b/signer/src/stacks/api.rs
@@ -164,24 +164,19 @@ pub struct AccountInfo {
 
 /// Helper function for converting a hexidecimal string into an integer.
 fn parse_hex_u128(hex: &str) -> Result<u128, Error> {
-    let hex_str = if hex.starts_with("0x") {
-        &hex[2..]
-    } else {
-        hex
-    };
-
+    let hex_str = hex.strip_prefix("0x").unwrap_or(hex);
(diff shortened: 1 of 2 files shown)


signer/src/stacks/api.rs
```diff
@@ -150,6 +152,34 @@ pub enum SubmitTxResponse {
...
+    pub balance: u128,
+    /// The amount locked (stacked?) in micro-STX
+    pub locked: u128,
+    /// The last known nonce of the account.
+    pub nonce: u64,
+}
+
+/// Helper function for converting a hexidecimal string into an integer.
+fn parse_hex_u128(hex: &str) -> Result<u128, Error> {
+    let hex_str = hex.strip_prefix("0x").unwrap_or(hex);
```
netrome, line 167: Nit: We can skip the extra unwrap logic by using `trim_start_matches`.
```suggestion
    let hex_str = hex.trim_start_matches("0x");
```

netrome (approved):
Nice! Just one minor suggestion.

signer/src/stacks/api.rs
djordon: Oh, perfect! Thank you.

PR #303, opened 2024-07-05
ci: integration test setup deploying smart contracts

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/263.

This PR builds off of https://github.com/stacks-network/sbtc/pull/302 and https://github.com/stacks-network/sbtc/pull/300 so the changes to signer/src/stacks/api.rs are not in this PR but is in https://github.com/stacks-network/sbtc/pull/302. I'll need to rebase once those are merged.

## Changes

* Adds integration test setup that deploys the smart contracts.

## Testing information

Tests we run against a local stacks node running Nakamoto. This was done by running the `feat/signer` branch of the [stacks-regtest-env](https://github.com/hirosystems/stacks-regtest-env/tree/feat/signer) repo using a stacks node built on commit https://github.com/stacks-network/stacks-core/commit/3d96d53b35409859ca2baa2f0b6ddaa1fbd80265.

After calling the `test_deploy` integration test function, I tested that it was actually deployed by retrieving the contract source with:
```bash
curl "http://localhost:20443/v2/contracts/source/SN326RW8148J2Q1P1MZ5K6JNY6WYDZPPZS45Z5ZBR/sbtc-token?proof=0"
```
For my regtest setup, I manually added the following to my stacks node config file
```toml
[[ustx_balance]]
# secret_key: 99dd7fc1ad584d9b174275ef9de7bda04fc61e38899fdce22fd31a49f3fc47d6
address = "ST1RQHF4VE5CZ6EK3MZPZVQBA0JVSMM9H5PMHMS1Y"
amount = 100000000000000

[[ustx_balance]]
# secret_key: 126242834c575d4ccd51fa7081775c09746305ab0889844fff09c2018a5548bd
address = "ST23T3EMEKV8MD6WMEWRHX4HQG0EC4GYM1W1NEQHQ"
amount = 100000000000000

[[ustx_balance]]
# secret_key: 440adaf1522f26e3d981d114c137090c6bf627ebb163b5cbb449c73f9659a003
address = "ST1SJ3DTE5DN7X54YDH5D64R3BCB6A2AG2ZQ8YPD5"
amount = 100000000000000

[[ustx_balance]]
# This is a 2-3 multi-sig address controlled using the above three addresses
address = "SN326RW8148J2Q1P1MZ5K6JNY6WYDZPPZS45Z5ZBR"
amount = 100000000000000
```


signer/tests/integration/contracts.rs
```diff
@@ -1,3 +1,128 @@
...
+async fn deploy_smart_contract<T>(state: &SignerKeyState, client: &StacksClient, deploy: T)
+where
+    T: AsContractDeploy,
+{
+    let mut unsigned = MultisigTx::new_tx(ContractDeploy(deploy), &state.state, TX_FEE);
+
+    let signatures: Vec<RecoverableSignature> = state
+        .keys
+        .iter()
+        .map(|kp| sign_ecdsa(unsigned.tx(), &kp.secret_key()))
```
netrome, line 73: Nit: I find it easier to read and navigate code when we don't import functions.

```suggestion
        .map(|kp| wallet::sign_ecdsa(unsigned.tx(), &kp.secret_key()))
```

signer/tests/integration/contracts.rs
```diff
@@ -1,3 +1,128 @@
...
+    // This only fails when we are given an invalid signature.
+    for signature in signatures {
+        unsigned.add_signature(signature).unwrap();
+    }
+
+    let tx = unsigned.finalize_transaction();
+
+    match client.submit_tx(&tx).await.unwrap() {
+        SubmitTxResponse::Acceptance(_) => (),
+        SubmitTxResponse::Rejection(err) => println!("{}", serde_json::to_string(&err).unwrap()),
```
netrome, line 85: Since this is for a test, wouldn't it be better to panic here? Right now I believe the test would pass even if the transaction is rejected.

signer/tests/integration/contracts.rs
```diff
@@ -1,3 +1,128 @@
...
+        })
+        .await;
+
+    state
+}
+
+#[ignore]
+#[tokio::test]
+async fn test_deploy() {
+    let _state = deploy_smart_contracts().await;
```
netrome, line 123: How do we know if the deployment has been successful? I don't see any assertion being made anywhere.

netrome (commented):
I looks to me like the integration test may pass even if we fail to deploy the contract. Am I missing something here or do we need to add some assertion or panic somewhere? I've suggested a place to add a panic.

signer/tests/integration/contracts.rs
djordon: Oh right, will update.

signer/tests/integration/contracts.rs
djordon: Oh yeah, I was supposed to remove this test. It's only here so that I can test that I setup smart contract deployment correctly.

djordon: Oh and to silence the warnings. I'm going to keep it in for now and remove it when I get to https://github.com/stacks-network/sbtc/issues/264.

netrome: Alright :+1: so this PR is basically only for the utilities and not the actual tests.

netrome (approved):

signer/tests/integration/contracts.rs
djordon: Yup, just for setup utilities.

Issue #307, opened 2024-07-08
[Feature]: Support key rotation contract calls


djordon:
# Feature - Support key rotation contract calls

### 1.1 Context & Purpose

Whenever we update the signer set, we need to rotate the list of known keys and the associated aggregate key. This ticket is for adding code to the signer for making that contract call.

### 2.1 Acceptance Criteria:
   - [ ] We add the structs necessary for making the contract call.

PR #308, opened 2024-07-09
feat: support key rotation contract calls

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/307.

I believe this is the last contract call that we need to support.

## Changes

* Bump the maximum number of signers to 128 from 15. @hstove @setzeus thoughts (it's this commit d6ac6ec046ef0bb8c1b22fba213dfdd3dc24c4e2)?
* Update the address hash mode to match the hash mode assumed in the clarity contracts.
* Add a constant in the signer codebase that sets the cap on the number of signers.
* Add an implementation of a struct that we can use to easily rotate the Signers' multi-sig wallet.

## Testing information

I added a test that checks that we do not error when creating the struct under common circumstances.

### DIFF
--- a/signer/src/stacks/contracts.rs
+++ b/signer/src/stacks/contracts.rs
@@ -40,7 +40,6 @@ use blockstack_lib::clarity::vm::Value;
 use blockstack_lib::types::chainstate::StacksAddress;
 use secp256k1::PublicKey;
 
-use crate::error::Error;
 use crate::stacks::wallet::SignerWallet;
 
 /// A struct describing any transaction post-execution conditions that we'd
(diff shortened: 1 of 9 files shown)


signer/src/stacks/contracts.rs
```diff
@@ -266,8 +279,116 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+    /// * The number of elements in the provided slice is greater than
+    ///   `i32::MAX`.
+    ///
+    /// But the SignerWallet checks for these two cases and rejects them,
+    /// so they cannot happen here.
+    pub fn new(wallet: &SignerWallet, deployer: StacksAddress) -> Result<Self, Error> {
+        let keys: Vec<&PublicKey> = wallet.public_keys().iter().collect();
+
+        Ok(Self {
+            aggregate_key: PublicKey::combine_keys(&keys).map_err(Error::InvalidAggregateKey)?,
```
djordon, line 316: Oh, I'm going to make this aggregate key in the `SignerWallet` struct.

netrome (approved):
Looks good to me!

djordon:
Merging now. @hstove @setzeus, if you guys have opinions on https://github.com/stacks-network/sbtc/commit/d6ac6ec046ef0bb8c1b22fba213dfdd3dc24c4e2 let me know, we can revert or whatever.

PR #311, opened 2024-07-10
ci: integration test stacks smart contracts

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/264.

We have a bunch of code for creating and signing stacks contract call transactions. This PR adds tests for whether these transactions will be accepted by a stacks node post-Nakamoto.

## Changes

* Update comments/documentation in a sBTC clarity contract. This became outdated after https://github.com/stacks-network/sbtc/pull/308.
* Updated the documentation of fields in some Stacks structs. Specifically, the nonce returned by `GET /v2/account/<principal>` request is the next nonce.
* Combine the `SignerWallet` and `SignerStxState` structs into one.
* Add a tests that checks whether each our contract calls will be accepted by a stacks node.

## Testing information

The tests that are added here are not run in CI, since we do not have Nakamoto running there yet. Tests were run against a local stacks node running Nakamoto. This was done by running the `feat/signer` branch of the [stacks-regtest-env](https://github.com/hirosystems/stacks-regtest-env/tree/feat/signer) repo using a stacks node built on commit https://github.com/stacks-network/stacks-core/commit/3d96d53b35409859ca2baa2f0b6ddaa1fbd80265.

For my regtest setup, I manually added the following to my stacks node config file
```toml
[[ustx_balance]]
# secret_key: 99dd7fc1ad584d9b174275ef9de7bda04fc61e38899fdce22fd31a49f3fc47d6
address = "ST1RQHF4VE5CZ6EK3MZPZVQBA0JVSMM9H5PMHMS1Y"
amount = 100000000000000

[[ustx_balance]]
# secret_key: 126242834c575d4ccd51fa7081775c09746305ab0889844fff09c2018a5548bd
address = "ST23T3EMEKV8MD6WMEWRHX4HQG0EC4GYM1W1NEQHQ"
amount = 100000000000000

[[ustx_balance]]
# secret_key: 440adaf1522f26e3d981d114c137090c6bf627ebb163b5cbb449c73f9659a003
address = "ST1SJ3DTE5DN7X54YDH5D64R3BCB6A2AG2ZQ8YPD5"
amount = 100000000000000

[[ustx_balance]]
# This is a 2-3 multi-sig address controlled using the above three addresses
address = "SN2V7WTJ7BHR03MPHZ1C9A9ZR6NZGR4WM8HT4V67Y"
amount = 100000000000000
```

Some of the logic included in these tests check whether the contract call transaction is included in a block. This implicitly checks a few of our other functions, and it turns out they work as expected. Unfortunately, this logic isn't enabled by default, even locally, since they would take a long time to complete. In my local regtest setup, Nakamoto blocks seem to be created every ~5 seconds, but the contract call transaction stays in the mempool for quite some time (on the order of 10 minutes).

### DIFF
--- a/signer/src/stacks/wallet.rs
+++ b/signer/src/stacks/wallet.rs
@@ -169,7 +169,7 @@ impl SignerWallet {
 #[derive(Debug)]
 pub struct SignerStxState {
     /// The multi-sig wallet for all known signers during this PoX cycle.
-    wallet: SignerWallet,
+    pub wallet: SignerWallet,
     /// The next nonce for the StacksAddress associated with the address of
     /// the wallet.
     nonce: AtomicU64,
--- a/signer/src/stacks/wallet.rs
+++ b/signer/src/stacks/wallet.rs
@@ -5,6 +5,7 @@ use std::collections::BTreeMap;
 use std::collections::BTreeSet;
 use std::sync::atomic::AtomicU64;
 use std::sync::atomic::Ordering;
+use std::sync::Arc;
 
 use blockstack_lib::address::C32_ADDRESS_VERSION_MAINNET_MULTISIG;
 use blockstack_lib::address::C32_ADDRESS_VERSION_TESTNET_MULTISIG;
@@ -60,6 +61,9 @@ pub struct SignerWallet {
     network_kind: NetworkKind,
     /// The multi-sig address associated with the public keys.
     address: StacksAddress,
+    /// The next nonce for the StacksAddress associated with the address of
+    /// the wallet.
+    nonce: Arc<AtomicU64>,
 }
 
 impl SignerWallet {
@@ -98,7 +102,9 @@ impl SignerWallet {
         public_keys: &[PublicKey],
         signatures_required: u16,
         network_kind: NetworkKind,
+        nonce: u64,
     ) -> Result<Self, Error> {
+        // Check most error conditions
         let num_keys = public_keys.len();
         let invalid_threshold = num_keys < signatures_required as usize;
         let invalid_num_keys = num_keys == 0 || num_keys > MAX_KEYS as usize;
@@ -142,6 +148,7 @@ impl SignerWallet {
             network_kind,
             address: StacksAddress::from_public_keys(version, &hash_mode, num_sigs, &pubkeys)
                 .ok_or(Error::StacksMultiSig(signatures_required, num_keys))?,
+            nonce: Arc::new(AtomicU64::new(nonce)),
         })
     }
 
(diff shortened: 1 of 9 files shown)


signer/tests/integration/contracts.rs
```diff
@@ -54,23 +73,26 @@ impl AsContractDeploy for SbtcBootstrapContract {
...
+
+fn make_signatures(tx: &StacksTransaction, keys: &[Keypair]) -> Vec<RecoverableSignature> {
+    keys.iter()
+        .map(|kp| stacks::wallet::sign_ecdsa(tx, &kp.secret_key()))
+        .collect()
 }
 
 /// Deploy an sBTC smart contract to the stacks node
-async fn deploy_smart_contract<T>(state: &SignerKeyState, client: &StacksClient, deploy: T)
+async fn deploy_smart_contract<T>(signer: &SignerStxState, deploy: T)
```
netrome, line 89: Not really related to this PR, but it might be nicer if this is a method on the `SignerStxState` struct rather than an independent function.

signer/tests/integration/contracts.rs
```diff
@@ -79,63 +101,123 @@ where
...
+        unsigned.add_signature(signature).unwrap();
+    }
+    let tx = unsigned.finalize_transaction();
 
-#[ignore]
-#[tokio::test]
-async fn accept_withdrawal_request_tx_accepted() {
-    // TODO(#264): Add integration test for signing Stacks smart contracts
-}
+    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
netrome, line 188: What are we waiting for here? Could you add a comment to describe. I assume this is for the smart contract deployments to be mined right?

signer/src/stacks/wallet.rs
```diff
@@ -60,6 +61,9 @@ pub struct SignerWallet {
     network_kind: NetworkKind,
     /// The multi-sig address associated with the public keys.
     address: StacksAddress,
+    /// The next nonce for the StacksAddress associated with the address of
+    /// the wallet.
+    nonce: Arc<AtomicU64>,
```
netrome, line 66: I don't find where this is shared. I could be missing something ofc. Why do we need to have an `Arc` here?

netrome (approved):
Looks good to me. I have two questions.

signer/tests/integration/contracts.rs
djordon: Yeah, it's to wait for the smart contract to be mined. I'll add a comment.

signer/src/stacks/wallet.rs
djordon: We need it to support cloning the struct. Not sure if we want it to implement `Clone` though. 

Note that this keeps the clones tied together, since once a `nonce` is incremented in one of the clones it's incremented everywhere (I'll add a test for this). While desirable and possibly "correct", I also don't like it since it's probably surprising. Maybe I should remove the atomic altogether? Any opinions here?

djordon: I could also remove the default `Clone` implementation, or manually implement it to make it more "natural".

netrome: Yeah if we can get rid of the `Clone` that seems like the best option.

signer/tests/integration/contracts.rs
netrome: Right, thank you!

PR #314, opened 2024-07-11
feat: adds costs test for completing withdrawals

### LINKED ISSUE #312: Clarity withdrawal completion optimization
When signers complete many withdrawals at once, each individual withdrawal makes a new `contract-call?` to `sbtc-registry` from the withdrawal contract. While this is working just fine, it's worthwhile to do some cost estimation and potential optimization. Primarily, the concern is that it may be too expensive to be able to process many withdrawals in a single tenure, which would create a "backlog" of withdrawals that need to be completed. Similarly, too high of cost could lead to excessive fees being paid by signers.

The work required is as follows:

- Determine the per-withdrawal cost when completing a withdrawal in order to determine the upper limit
- Investigate optimizations (such as a batch call to `sbtc-registry`) to reduce costs.

hstove:
- Closes #312 

This PR adds a unit tests that measures the amount of cost used when "maxing out" the number of individual withdrawals completed in a single transaction.

At the moment, the unit test completes 500 withdrawals in one tx. To view the costs of a transaction, run `pnpm test:report`, then open `contracts/costs-report.json`. Search for `"test_name": "tests/sbtc-withdrawal.test.ts__optimization tests` to see that transaction's costs. The result with 500 withdrawals is:

```json
"cost_result": {
  "total": {
    "write_length": 10500,
    "write_count": 1500,
    "read_length": 17978964,
    "read_count": 12009,
    "runtime": 52492065
  },
  "limit": {
    "write_length": 15000000,
    "write_count": 15000,
    "read_length": 100000000,
    "read_count": 15000,
    "runtime": 5000000000
  },
  "memory": 91500,
  "memory_limit": 100000000
}
 ```
 
As you can see, the `read_count` is pretty close to the full block budget, so we can assume that  ~600 is the upper limit **without optimization**. In general, `read_count` is the metric that most blocks max out, so the fact that it's also our highest relative metric is important.

On the other hand, I'm not so sure that making a batch call to `.sbtc-registry` would help that much, because (I think) each `contract-call?` is only 1 extra `read_count`. Additionally, `sbtc-registry` will need to validate the contract caller each time, so that's another single read per call. So, we'd reduce the read count by 1000, but that's not that much of an optimization (compared to what I thought it could be).


setzeus (changes requested):
Copy copy. Thank you for the great work here! Can confirm the same results. Happy to approve overall just have two small requirements:

1. Can we change the list limit in parameter to 600? (the max amount), no point in leaving it at 1000
2. CI currently failing over a mistype error

hstove:
@setzeus I've made some changes:

- Wrote a similar test for deposits - it has a very similar maximum (650), which I've updated the argument for.
- Fixed the type error
- Updated the max withdrawals to 600 as you mentioned

setzeus (approved):

Issue #210, opened 2024-05-29
[Feature]: Block observer - extract set aggregate key transactions


netrome:
### Context
The block observer should as part of its operation extract any observed set aggregate key transactions from the stacks blocks it's processing and write the extracted requests to the signer database.

### Definition of done
The `extract_set_aggregate_key_transactions` method in the block observer is implemented and tested.

Issue #207, opened 2024-05-29
[Feature]: Block observer - extract withdraw reject transactions


netrome:
### Context
The block observer should as part of its operation extract any observed withdraw reject transactions from the stacks blocks it's processing and write the extracted requests to the signer database.

### Definition of done
The `extract_withdraw_reject_transactions` method in the block observer is implemented and tested.

Issue #206, opened 2024-05-29
[Feature]: Block observer - extract withdraw accept transactions


netrome:
### Context
The block observer should as part of its operation extract any observed withdraw accept transactions from the stacks blocks it's processing and write the extracted requests to the signer database.

### Definition of done
The `extract_withdraw_accept_transactions` method in the block observer is implemented and tested.

PR #315, opened 2024-07-11
feat: extract sbtc contract calls block observer

djordon:
## Description

* Closes https://github.com/stacks-network/sbtc/issues/203
* Closes https://github.com/stacks-network/sbtc/issues/205
* Closes https://github.com/stacks-network/sbtc/issues/206
* Closes https://github.com/stacks-network/sbtc/issues/207
* Closes https://github.com/stacks-network/sbtc/issues/208
* Closes https://github.com/stacks-network/sbtc/issues/210
* Closes https://github.com/stacks-network/sbtc/issues/212

This might also address https://github.com/stacks-network/sbtc/issues/209, but I'm not sure.

## Changes

1. Modify one of the functions in the `DbWrite` trait to use the model types as input. This keeps the interface a bit more logical and makes it easier to test.
2. Rewrote the functions that wrote stacks blocks to send slices of the correct type, instead of sending JSON. This allows us to make use of the actual type that we want to write.
3. Finish implementation of the function that writes stacks blocks in the block observer.

## Testing information

This PR is mostly a refactor. The core functionality was added in 2e30ef3ad8fe78eee88b84d276f3777629c20ddb, but subsequent commits addressed some oddities in the previous code.


netrome (approved):
This is great stuff so far. What we should also do (either here or as follow-ups) is to populate some dedicated tables for the different transaction types. We should:

* Parse any calls to `rotate-keys-wrapper` and write the updated signer set into a table. Right now the storage model has the `aggregate_key_transactions` table that tracks the aggregate keys. I think we could rename this to `key_rotations`, and we also need to track the updated signer set, so we need to add a field for that as well.
* Populate the `deposit_responses` and `withdraw_responses` tables in the state model.

djordon:
> LGTM, though how are we ultimately getting the data from the deleted functions? For example with the function below:
> 
> ```rust
>     fn extract_update_signer_set_transactions(&self, _transactions: &[stacks::StacksTransaction]) {
>         // TODO(#208): Implement
>     }
> ```
> 
> Do we no longer need that data?

We do, it's just done in the `BlockObserver::write_stacks_blocks` function. That function which extracts the relevant transactions and writes them all to the database as two queries.

Issue #44, opened 2024-04-10
[Design]: Bootstrap signer components


netrome:
***Completing the issue description and arriving at a conclusion is the deliverable of this issue.***

# Design - Bootstrap signer components

This ticket defines the main components which constitute the sBTC bootstrap signer and how they interact to fulfill deposits and withdrawals in sBTC v1.

## 1. Summary
We have broken down the sBTC bootstrap signer into the following main components

* Block observer
* Signer set manager
* Decision engine
* Transaction signer
* Transaction coordinator
* Request sequencer

building on top of these supporting modules

* Bitcoin interface
* Stacks interface
* Deposit API interface
* Signer network
* sBTC transactions 

## 2. Context & Purpose
The sBTC bootstrap signer should be able to coordinate the creation of transactions, maintain shared custody of a bitcoin wallet, and make smart contract calls in the `.sbtc` contract. To manage the complexity of this system, we want to break it down into multiple loosely coupled components of separate functionality.

***Relevant Research Discussions***

- #34 
- #33 
- #39 
- #40 
- #37 
- #18 

***External Resources***

### 3.1 Proposed Component Design
The bootstrap signer is responsible for:

* Storing the signer DKG shares and private network key
* Observing and reacting to sBTC requests
* Coordinating DKG and signing rounds when needed

To fulfill these duties, we have identified the following sub-components of functionality in the Signer. 

##### Block observer
The block observer is the component responsible for populating the bitcoin and stacks blocks in the state. In addition, the observer should parse any requests, responses or other transactions it observes and populate the corresponding pieces of the state.

##### Signer set manager
The signer set manager is the component orchestrating signer bootstrapping, key rotation and DKG.

##### Decision engine
The decision engine exposes an extendible interface for deciding whether to accept or reject a sBTC request.

##### Transaction signer
The transaction signer is responsible for participating in WSTS signing rounds for any approved transactions.

##### Transaction coordinator
The transaction coordinator is responsible for orchestrating WSTS signing rounds, and submitting the resulting transactions to the blockchain.

##### Request sequencer
The request sequencer is responsible for creating the next sBTC transaction for the coordinator to handle in a signing round given the signer state.

##### Bitcoin interface
Support module that helps the signer interact with the Bitcoin blockchain.

##### Stacks interface
Support module that helps the signer interact with the Stacks blockchain.

##### Deposit API interface
Support module that allows the signer to communicate with the Deposit API.

##### Signer network
[...]

PR #316, opened 2024-07-11
feat: Skeleton implementation of transaction coordinator

### LINKED ISSUE #235: [Feature]: Skeleton implementation of transaction coordinator event loop
# Feature - Skeleton implementation of transaction coordinator event loop

### Context
We have already documented what we intend the transaction coordinator event loop to do in #222 . As a next step we should make a skeleton implementation of this loop with placeholder logic for all independent pieces of functionality and create separate tickets for that.

### Definition of done
We have a skeleton implementation of the transaction coordinator event loop and an exhaustive set of issues to finish it.

netrome:
closes #235 

This PR adds initial boilerplate and does some refactors to pave the way for implementing a functioning transaction coordinator as designed in #44.

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -209,6 +209,10 @@ where
 
     // Determine if the current coordinator is the coordinator
     //
+    // The coordinator is decided using the hash of the bitcoin
+    // chain tip. We don't use the chain tip directly because
+    // it typically starts with a lot of leading zeros.
+    //
     // Note that this function is technically not fallible,
     // but for now we have chosen to return phantom errors
     // instead of adding expects/unwraps in the code.
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -5,6 +5,18 @@
 //!
 //! For more details, see the [`TxCoordinatorEventLoop`] documentation.
 
+use std::collections::BTreeSet;
+
+use sha2::Digest;
+
+use crate::bitcoin::BitcoinInteract;
+use crate::error;
+use crate::network;
+use crate::storage;
+use crate::storage::model;
+use crate::utxo;
+use crate::wsts_state_machine;
+
 #[cfg_attr(doc, aquamarine::aquamarine)]
 /// # Transaction coordinator event loop
 ///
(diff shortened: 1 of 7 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -73,4 +85,189 @@
...
+    // expressions. However, that is left for future work.
+    fn is_coordinator(
+        &self,
+        bitcoin_chain_tip: &model::BitcoinBlockHash,
+    ) -> Result<bool, error::Error> {
+        let mut hasher = sha2::Sha256::new();
+        hasher.update(bitcoin_chain_tip);
+        let digest = hasher.finalize();
+        let index =
+            usize::from_be_bytes(*digest.first_chunk().ok_or(error::Error::TypeConversion)?);
```
djordon, line 226: Nice.

The `core::slice::first_chunk` function is generic, and returns an option of an array of fixed length where the length is the generic parameter (here the length is inferred to be 8 on a typical machine). `None` is only returned if the generic length parameter is longer than the length of the underlying slice. Since the Sha256 type contains 32 bytes, this should never error (which you noted already).

Also, using the bitcoin block hash directly is no good since it isn't random enough (although it's 32 bytes, it typically starts with lots of leading zeros bits). So we hash it first and use the first 8 bytes of that to make sure that it actually is random. I would add a comment somewhere noting this.

signer/src/transaction_coordinator.rs
```diff
@@ -73,4 +85,189 @@
...
+
+        let pub_key = self.pub_key()?;
```
djordon, line 228: Oh right, this also should never error.

djordon (approved):
Looks good! I would add another comment about why we hash the bitcoin block hash when computing the coordinator index.

signer/src/transaction_coordinator.rs
netrome: Right, actually I think the block hash in itself is probably good enough. Definitely for v1 :sweat_smile: Though there's no harm in the extra hashing of it. I'll update the comment.

signer/src/transaction_coordinator.rs
netrome: Yeah exactly, already commented it.

signer/src/transaction_coordinator.rs
netrome: Actually, I'll skip the hashing. I don't see any reason for doing it.

netrome: Changed my mind (or rather you did in a DM in slack), we'll keep the hashing. I've added an explanatory comment now.

Issue #319, opened 2024-07-11
[Feature]: Transaction coordinator capable of computing and signing a valid bitcoin transaction


netrome:
Depends on #317 and #318 

### Context
Once the transaction coordinator is able to load a functioning wsts coordinator state machine, we should implement the ability for the coordinator to collect all the necessary information to construct a valid transaction package and provide the ability to run a signing round with this transaction package.

### Definition of done
The transaction coordinator can compute and sign valid bitcoin transactions.

PR #325, opened 2024-07-14
feat: Transaction coordinator able to fetch pending requests to sign

### LINKED ISSUE #318: [Feature]: Transaction coordinator able to fetch pending requests to sign
### Context
The transaction coordinator should upon it's tenure fetch any pending deposit requests and withdraw requests that fulfills the following:

* The request has been approved by at least `threshold` signers.
* The request has enough confirmations if it is a withdraw request.
* The request has not been fulfilled on the given bitcoin chain tip.

### Definition of done
The transaction signer is capable of loading the set of pending requests it should coordinate signing rounds for.

netrome:
Closes #318 

This PR introduces the queries needed for the transaction coordinator to fetch pending accepted deposit and withdraw requests and convert them into types usable for the transaction packaging function.

Note that while we have tests for the queries, testing this logic within the transaction coordinator is dependent on #319.


signer/src/storage/postgres.rs
```diff
@@ -247,6 +247,66 @@ impl super::DbRead for PgStore {
...
+              , deposit_requests.amount
+              , deposit_requests.max_fee
+              , deposit_requests.sender_addresses
+              , deposit_requests.created_at
+            FROM transactions_in_window transactions
+            JOIN sbtc_signer.deposit_requests deposit_requests ON
+                deposit_requests.txid = transactions.txid
+            JOIN sbtc_signer.deposit_signers signers ON
+                deposit_requests.txid = signers.txid AND
+                deposit_requests.output_index = signers.output_index
```
djordon, line 295: You might be able to do `USING (txid)` and `USING (txid, output_index)` here.

signer/src/storage/postgres.rs
```diff
@@ -247,6 +247,66 @@ impl super::DbRead for PgStore {
...
+                WHERE block_hash = $1
+                
+                UNION ALL
+                
+                -- Recursive member: Fetch the parent block using the last block's parent_hash
+                SELECT parent.block_hash, parent.block_height, parent.parent_hash,
+                       parent.created_at, last.depth + 1
+                FROM sbtc_signer.bitcoin_blocks parent
+                JOIN context_window last ON parent.block_hash = last.parent_hash
+                WHERE last.depth < $2
```
djordon, line 272: Is this supposed to be `<=`? We might need to change this and thw query in `get_pending_deposit_requests` or the query in `get_pending_withdraw_requests`.

signer/src/transaction_coordinator.rs
```diff
@@ -261,11 +263,50 @@ where
...
+            .map(|request| request.try_into())
+            .collect::<Result<_, _>>()?;
+
+        let withdrawals = pending_withdraw_requests
+            .into_iter()
+            .map(|request| request.try_into())
+            .collect::<Result<_, _>>()?;
+
+        let accept_threshold = self.threshold;
+        let num_signers = self
```
djordon, line 297: Hmmm, maybe I should change `num_signers` to a `usize` in `utxo::SbtcRequests`? Eh, it that change looks annoying.

djordon (commented):
Largely looks good. Might be an off-by-one error in the query, and had a suggested change to the `TryFrom` `impl`s.

signer/src/utxo.rs
```diff
@@ -387,6 +438,34 @@ impl WithdrawalRequest {
...
+            .map_err(|_| Error::TypeConversion)?;
+        let max_fee = request
+            .max_fee
+            .try_into()
+            .map_err(|_| Error::TypeConversion)?;
+        let address: Address<NetworkUnchecked> = request
+            .sender_address
+            .parse()
+            .map_err(Error::ParseAddress)?;
+        let address = address.assume_checked(); // We do not validate the request address here
```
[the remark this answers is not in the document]
AshtonStephens: I don't fully understand this suggestion, but as a blanket statement on all sBTC PRs, the main things we want to prioritize is readability, verifiability, and delivery speed. I say that here because "a little more robust" might not be something to change, and might be more of a "nit".

Given this is the approach we're taking for Deposits with a `try_from` I think the symmetry between our conversion implementation for `Deposit` and `Withdrawal`s from the models using `try_from` is probably the best / most readable approach.

signer/src/utxo.rs
```diff
@@ -359,6 +362,54 @@ impl DepositRequest {
...
+            .map_err(|_| Error::TypeConversion)?;
+
+        let deposit_script = bitcoin::consensus::deserialize(&request.spend_script)
+            .map_err(Error::BitcoinSerialization)?;
+
+        let redeem_script = bitcoin::consensus::deserialize(&request.reclaim_script)
+            .map_err(Error::BitcoinSerialization)?;
+
+        let signers_public_key =
+            XOnlyPublicKey::from_slice(&[0; 32]).map_err(|_| Error::TypeConversion)?;
```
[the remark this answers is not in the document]
netrome: Oh, thanks for spotting. I had a todo note to fix this, but I think when I fixed it I only handled the error handling and forgot about the hard-coded zeroes. I guess as you suggested in another comment, this is also some information we need to add to the conversion, so I'll change the TryFrom to a custom method.

> What would be wrong about this?
This creates a hard-coded zeroed public-key.

signer/src/utxo.rs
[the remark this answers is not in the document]
netrome: Yeah I don't think this is the place to validate the received address, although it doesn't hurt to be conservative here since the risk would be users loosing their funds.

> the main things we want to prioritize is readability, verifiability, and delivery speed

I agree, and I would also add correctness to this list.

Since I'll change the conversion for the `Deposit` I might as well do it here. It's not too much work.

signer/src/utxo.rs
```diff
@@ -359,6 +362,54 @@ impl DepositRequest {
...
+        let amount = request
+            .amount
+            .try_into()
+            .map_err(|_| Error::TypeConversion)?;
+
+        let deposit_script = bitcoin::consensus::deserialize(&request.spend_script)
+            .map_err(Error::BitcoinSerialization)?;
+
+        let redeem_script = bitcoin::consensus::deserialize(&request.reclaim_script)
+            .map_err(Error::BitcoinSerialization)?;
```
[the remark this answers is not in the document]
netrome: Right, yeah that's probably better. For some reason, deserialize felt more handy here. I don't remember why, because for other types I've used the direct byte conversion. I'll update.

signer/src/storage/postgres.rs
netrome: I do think this is correct, since we're joining the parent which has `last.depth + 1`. We also have tests ensuring this is consistent with the in memory implementation, which iterates over `(0..context_window)`. Let me know if you still believe this to be incorrect, and we can investigate this further.

signer/src/utxo.rs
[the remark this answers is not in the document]
netrome: Updated ✅

signer/src/utxo.rs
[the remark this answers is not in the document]
netrome: Fixed it :heavy_check_mark:

signer/src/storage/postgres.rs
netrome: Thanks for suggesting. Updated.

signer/src/utxo.rs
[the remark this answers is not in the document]
netrome: Done.

signer/src/error.rs
```diff
@@ -171,13 +171,21 @@ pub enum Error {
...
     #[error("WSTS error: {0}")]
     Wsts(#[source] wsts::state_machine::signer::Error),
 
     /// No chain tip found.
     #[error("no bitcoin chain tip")]
     NoChainTip,
+
+    /// Bitcoin address parse error
+    #[error("bitcoin address parse error")]
+    BitcoinAddressParseError(#[source] bitcoin::address::ParseError),
```
djordon, line 184: nit: This should be `BitcoinAddressParse`, the added `Error` is redundant. But maybe we should rename it to `BitcoinNetworkValidation`? Or not, the `Parse` part is okay as well since it's conceptually part of string parsing.

signer/src/storage/postgres.rs
djordon: Yeah, you're right, this is fine. It gets `context_window` blocks. And the query in `get_pending_withdraw_requests` is fine as well since we explicitly want to go past the context window. Never mind 😅!

signer/src/storage/postgres.rs
```diff
@@ -247,6 +247,63 @@ impl super::DbRead for PgStore {
...
+                -- Anchor member: Initialize the recursion with the chain tip
+                SELECT block_hash, block_height, parent_hash, created_at, 1 AS depth
+                FROM sbtc_signer.bitcoin_blocks
+                WHERE block_hash = $1
+                
+                UNION ALL
+                
+                -- Recursive member: Fetch the parent block using the last block's parent_hash
+                SELECT parent.block_hash, parent.block_height, parent.parent_hash,
+                       parent.created_at, last.depth + 1
```
djordon, line 269: nit: when the columns wrap like this, I'd prefer we switch to having one column per line.

djordon (approved):
Looks good! And all is well with the query (no off-by-one).

I had a couple of nits but it looks good.

signer/src/storage/postgres.rs
netrome: Right, updated.

signer/src/error.rs
netrome: Ah how could I miss this. Fixed!

PR #331, opened 2024-07-15
289 feature dynamodb data accessor integration tests

AshtonStephens:
## Description
Emily integration tests

This PR adds some Makefile changes and an integration test suite for the Emily API.


emily/handler/Cargo.toml
```diff
@@ -32,6 +34,10 @@ utoipa = "4.2.3"
 warp = "0.3"
 warp_lambda.workspace = true
 
+fake = { version = "2.9.2", features = ["derive", "time"], optional = true }
+
 [dev-dependencies]
 mockito = "0.28"
 mockall = "0.10"
+serial_test = "0.5"
```
djordon, line 42: Is this necessary since you we set `--test-threads==1` when running tests?

emily/handler/Cargo.toml
```diff
@@ -6,9 +6,11 @@ edition = "2021"
 [[bin]]
 name = "emily-handler"
 
+[features]
 default = ["testing"]
 integration-tests = ["testing"]
 testing = ["fake"]
+emily = ["integration-tests"]
```
djordon, line 13: I think this can be removed. It doesn't seem to be used anywhere.

djordon (commented):
Basically looks good. Just a few comments.

emily/handler/Cargo.toml
AshtonStephens: If someone changes that I'd rather they get defined behavior, seems like something one could accidentally turn off not realizing that the integration tests no longer have a guaranteed order.

emily/handler/Cargo.toml
```diff
@@ -32,6 +34,10 @@ utoipa = "4.2.3"
 warp = "0.3"
 warp_lambda.workspace = true
 
+fake = { version = "2.9.2", features = ["derive", "time"], optional = true }
+
 [dev-dependencies]
 mockito = "0.28"
 mockall = "0.10"
+serial_test = "0.5"
+assert_unordered = "0.2"
```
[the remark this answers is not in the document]
AshtonStephens: Why not use a library?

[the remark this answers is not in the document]
AshtonStephens: I'm defaulting to libraries more because they seem to be so common in Rust, I don't want to go around reinventing these things. There's a huge benefit to readability if we're using common libraries.

emily/handler/Cargo.toml
AshtonStephens: Worst case scenario is that someone changes this and then tests happen to go in the right order for a while and at some point there's a cargo update that makes some optimization that changes the order. I'd rather just make that impossible, and this library does that.

emily/handler/Cargo.toml
AshtonStephens: Done.

emily/handler/Cargo.toml
djordon: I didn't think this library guaranteed the order. In the project README they state:
> Multiple tests with the serial attribute are guaranteed to be executed in serial. Ordering of the tests is not guaranteed however.

---

But I was curious if this is something that we should consider using in the signer codebase. It seems like it does what `--test-threads==1` but only for the tests that were tagged with the macro. Is that right?

AshtonStephens: ugh, crud. I suspect so, but now I don't know. I'm not sure how we'd guarantee an order.

AshtonStephens: Looks like you can't, I'll find a way around this in the next PR, but for now this should do it's job ~

emily/handler/Cargo.toml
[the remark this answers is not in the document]
djordon: Yeah, libraries are often used in Rust-land. In particular, there are a few very commonly used libraries. For not-so-commonly used ones, I tend to not use them since I require myself to vet them. There have been a few trojan crates discovered on `crates.io` (they had a blog post about it a few months ago). These dependencies here seem okay (I did a basic vet of them and their dependencies and they looked largely fine, I wasn't super robust about it though). But yeah, I wouldn't go wild with the dependencies.


> I'm defaulting to libraries more because they seem to be so common in Rust, I don't want to go around reinventing these things. There's a huge benefit to readability if we're using common libraries.

Yeah, but I didn't think `assert_unordered` was a common library. I had to look it up (I did a basic vet of it too). Using `.sort_by` is also readable and easy to understand, and is in the standard library.


---

I'm not trying to harp on library use or even this one in particular. I like them, use them, blah blah blah. But I do want to increase the cost of adding them, and this seems like a case where I'd go without adding them (because I wouldn't want to vet it and the alternative is simple enough).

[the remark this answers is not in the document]
AshtonStephens: Works for me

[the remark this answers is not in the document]
djordon: 👍🏽 cool. One last thing, I'd change this to
```toml
assert_unordered = { version = "0.2", default-features = false }
```
It doesn't pull in any transient dependencies if you don't use the default features.

emily/handler/Cargo.toml
djordon: Like the other dependency, I'd change this to
```toml
serial_test = { version = "0.5", default-features = false }
```

djordon (approved):
Looks good. Just had two nits about the dependencies if we're going to keep them.

Issue #329, opened 2024-07-15
[Feature]: Construct bitcoin deposit addresses


djordon:
# Feature - Construct bitcoin deposit addresses

### 1.1 Context & Purpose

This is part of https://github.com/stacks-network/sbtc/issues/322, and has the same context and but with a narrow focus on constructing a bitcoin address from the variable inputs. These inputs are:

1. The public key hash used in the deposit script.
2. The stacks principal address to mint sBTC to in the deposit script.
3. The max fee for when the signers act on the BTC transaction.
4. The lock time component in the beginning of the reclaim script.
5. The rest of the reclaim script.

### 2.1 Acceptance Criteria:
   - [ ] Have a function that takes the five inputs above and generates the expected deposit address.

## 3. Related Issues and Pull Requests (optional):

Relates to the work outlined in the LLD referenced in https://github.com/stacks-network/sbtc/issues/48 and is part of https://github.com/stacks-network/sbtc/issues/322.

PR #333, opened 2024-07-15
feat: construct bitcoin deposit addresses

djordon:
Use the following template to create your pull request

## Description

Closes https://github.com/stacks-network/sbtc/issues/329.

## Changes

1. Add a function that can construct a deposit script from the deposit info (the signers' public key, the recipient address, and the max fee).
2. Add a function that can construct a BTC address given the deposit script variables.
3. Saw that the `utxo::DepositRequest` struct had a typo in one of the field names, corrected it.

## Testing information

The unit tests here are okay

### DIFF
--- a/common/src/deposits.rs
+++ b/common/src/deposits.rs
@@ -1,0 +1,114 @@
+//! This is the transaction analysis module
+//!
+
+use bitcoin::opcodes::all as opcodes;
+use bitcoin::script::PushBytesBuf;
+use bitcoin::taproot::LeafVersion;
+use bitcoin::taproot::NodeInfo;
+use bitcoin::taproot::TaprootSpendInfo;
+use bitcoin::Address;
+use bitcoin::Network;
+use bitcoin::ScriptBuf;
+use bitcoin::XOnlyPublicKey;
+use clarity::codec::StacksMessageCodec;
+use clarity::vm::types::PrincipalData;
+use secp256k1::SECP256K1;
+
+/// This struct contains the key variable inputs when constructing a
+/// deposit script address.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DepositScript {
+    /// The last known public key of the signers.
+    pub signers_public_key: XOnlyPublicKey,
+    /// The stacks address to deposit the sBTC to. This can be either a
+    /// standard address or a contract address.
+    pub stacks_address: PrincipalData,
+    /// The max fee amount to use for the BTC deposit transaction.
+    pub max_fee: u64,
+}
+
+impl DepositScript {
+    /// Construct a bitcoin address for a deposit transaction on the given
+    /// network.
+    pub fn to_address(&self, reclaim_script: ScriptBuf, network: Network) -> Address {
+        let deposit_script = self.deposit_script();
+        let ver = LeafVersion::TapScript;
+
+        // For such a simple tree, we construct it by hand.
+        let leaf1 = NodeInfo::new_leaf_with_ver(deposit_script, ver);
+        let leaf2 = NodeInfo::new_leaf_with_ver(reclaim_script, ver);
+
+        // A Result::Err is returned by NodeInfo::combine if the depth of
+        // our taproot tree exceeds the maximum depth of taproot trees,
+        // which is 128. We have two nodes so the depth is 1 so this will
+        // never panic.
+        let node =
+            NodeInfo::combine(leaf1, leaf2).expect("Tree depth is greater than the max of 128");
+        let internal_key = crate::unspendable_taproot_key();
+
+        let merkle_root =
+            TaprootSpendInfo::from_node_info(SECP256K1, *internal_key, node).merkle_root();
+        Address::p2tr(SECP256K1, *internal_key, merkle_root, network)
+    }
+
+    /// Construct a deposit script from the inputs
+    pub fn deposit_script(&self) -> ScriptBuf {
+        // The format of the OP_DROP data, as shown in
+        // https://github.com/stacks-network/sbtc/issues/30, is 8 bytes for
+        // the max fee followed by up to 151 bytes for the stacks address.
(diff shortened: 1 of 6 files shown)


signer/src/utxo.rs
```diff
@@ -260,8 +260,8 @@ pub struct DepositRequest {
     pub amount: u64,
     /// The deposit script used so that the signers' can spend funds.
     pub deposit_script: ScriptBuf,
-    /// The redeem script for the deposit.
-    pub redeem_script: ScriptBuf,
+    /// The reclaim script for the deposit.
+    pub reclaim_script: ScriptBuf,
```
AshtonStephens, line 264: 👍

common/src/lib.rs
```diff
@@ -3,5 +3,46 @@
...
+pub const NUMS_X_COORDINATE: [u8; 32] = [
+    0x50, 0x92, 0x9b, 0x74, 0xc1, 0xa0, 0x49, 0x54,
+    0xb7, 0x8b, 0x4b, 0x60, 0x35, 0xe9, 0x7a, 0x5e,
+    0x07, 0x8a, 0x5a, 0x0f, 0x28, 0xec, 0x96, 0xd5,
+    0x47, 0xbf, 0xee, 0x9a, 0xce, 0x80, 0x3a, 0xc0,
+];
+
+/// The number of bitcoin blocks that must elapse from when the deposit
+/// transaction is confirmed before it can be reclaimed by the depositor.
+pub const LOCK_TIME: u32 = 150;
```
AshtonStephens, line 34: How are we verifying that in the script? My rust analyzer doesn't seem to think it's being used.

common/src/deposits.rs
```diff
@@ -0,0 +1,114 @@
...
+            NodeInfo::combine(leaf1, leaf2).expect("Tree depth is greater than the max of 128");
+        let internal_key = crate::unspendable_taproot_key();
+
+        let merkle_root =
+            TaprootSpendInfo::from_node_info(SECP256K1, *internal_key, node).merkle_root();
+        Address::p2tr(SECP256K1, *internal_key, merkle_root, network)
+    }
+
+    /// Construct a deposit script from the inputs
+    pub fn deposit_script(&self) -> ScriptBuf {
```
AshtonStephens, line 55: This is so clean.

common/src/deposits.rs
```diff
@@ -0,0 +1,114 @@
...
+use bitcoin::ScriptBuf;
+use bitcoin::XOnlyPublicKey;
+use clarity::codec::StacksMessageCodec;
+use clarity::vm::types::PrincipalData;
+use secp256k1::SECP256K1;
+
+/// This struct contains the key variable inputs when constructing a
+/// deposit script address.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DepositScript {
```
AshtonStephens, line 20: Can you update this to also "extract" the block height at which the reclaim script becomes spendable?

Also, if I understand this correctly, the depositor provides all the parameters including block height at which they can redeem and then here we're making the taproot script that we'd expect given those parameters and will use that to check whether the lock script does what's expected?

AshtonStephens (approved):
Overall this is very clean! I'm unclear on how some of it works; can you explain how the LOCK_TIME is checked in the deposit script?

common/src/lib.rs
djordon: Oh yeah it's not being used. It's for the reclaim script stuff, which I'll start working on soon.

AshtonStephens: Okay, roger, this addresses my comment on https://github.com/stacks-network/sbtc/pull/334 as well.

common/src/deposits.rs
netrome: I think getting the block height might be trickier than it seems. What happens if the deposit request exists on multiple forks with different heights? The block height is not a property of the request, but rather something we can derive from the chain state when we observe a request on the chain.

From that perspective, when it comes to the block height I think it makes sense to query that information from the database when we need it.

common/src/lib.rs
```diff
@@ -3,5 +3,46 @@
...
+/// wrote the X-coordinate that is reproduced below.
+///
+/// [1]: https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs
+#[rustfmt::skip]
+pub const NUMS_X_COORDINATE: [u8; 32] = [
+    0x50, 0x92, 0x9b, 0x74, 0xc1, 0xa0, 0x49, 0x54,
+    0xb7, 0x8b, 0x4b, 0x60, 0x35, 0xe9, 0x7a, 0x5e,
+    0x07, 0x8a, 0x5a, 0x0f, 0x28, 0xec, 0x96, 0xd5,
+    0x47, 0xbf, 0xee, 0x9a, 0xce, 0x80, 0x3a, 0xc0,
+];
```
netrome, line 30: Don't we have this constant already in `utxo.rs`?

netrome (approved):
Good stuff. Just one question.

common/src/deposits.rs
djordon: Oh right forking! I knew it wasn't a property of the request since it's a relative time lock, but yeah, gotta handle forking. In any case, I was interpreting the "block height" question as "when are we going to extract that parameter in the reclaim script?" which is basically all that I have planned for in this library with respect to the reclaim script.

Also, if forking weren't bad enough the parameter that is fed into `OP_CHECKSEQUENCEVERIFY` can be either a relative time or the number of bitcoin blocks (I know you know this already). The bad part is that the choice happens when you spend the UTXO (as opposed to when you create the UTXO). That distinction was lost on me when I was first reading up about it. This means that now we have to care about time as well.

common/src/lib.rs
djordon: We do, I'm going to remove one of them when these libraries get integrated.

Issue #327, opened 2024-07-15
[Feature]: Parse and validate deposit scripts


djordon:
# Feature - Parse and validate deposit scripts

### 1.1 Context & Purpose

This is part of https://github.com/stacks-network/sbtc/issues/322, and has the same context and but with a narrow focus on parsing the deposit script.

### 2.1 Acceptance Criteria:
   - [ ] Have a function that parses the deposit script, validates it, and extracts the relevant components.

## 3. Related Issues and Pull Requests (optional):

Relates to the work outlined in the LLD referenced in https://github.com/stacks-network/sbtc/issues/48 and is part of https://github.com/stacks-network/sbtc/issues/322.

PR #334, opened 2024-07-15
feat: add a deposit script parsing function

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/327.

## Changes

- Add a function that can parse a deposit script and extract the relevant pieces of information.

## Testing information

The tests check that the parser function is an inverse of the `DepositScriptInputs::deposit_script` function.

### DIFF
--- a/common/src/deposits.rs
+++ b/common/src/deposits.rs
@@ -13,6 +13,39 @@ use bitcoin::XOnlyPublicKey;
 use clarity::codec::StacksMessageCodec;
 use clarity::vm::types::PrincipalData;
 use secp256k1::SECP256K1;
+use stacks_common::types::chainstate::STACKS_ADDRESS_ENCODED_SIZE;
+
+/// This is the length of the fixed portion of the deposit script, which
+/// is:
+/// ```text
+///  OP_DROP OP_PUSHBYTES_32 <x-only-public-key> OP_CHECKSIG
+/// ```
+/// Since we are using Schnorr signatures, we only use the x-coordinate of
+/// the public key. The full public key is assumed to be even.
+const DEPOSIT_SCRIPT_FIXED_LENGTH: usize = 35;
+
+/// This is the typical number of bytes of a deposit script. It's 1 byte
+/// for the length of the following 30 bytes of data, which is 8 bytes for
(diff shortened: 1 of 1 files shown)


common/src/deposits.rs
```diff
@@ -96,16 +197,67 @@ mod tests {
...
+        let parsed_deposit = parse_deposit_script(&deposit_script).unwrap();
+
+        assert_eq!(deposit, parsed_deposit);
+    }
+
     /// Basic check that we can create an address without any issues
     #[test_case(PrincipalData::from(StacksAddress::burn_address(false)) ; "standard address")]
     #[test_case(PrincipalData::parse(CONTRACT_ADDRESS).unwrap(); "contract address")]
-    fn btc_address(stacks_address: PrincipalData) {
+    fn btc_address(recipient: PrincipalData) {
```
AshtonStephens, line 254: 👍

AshtonStephens (approved):
This looks good to me. I'm still not clear on how we plan to verify the reclaim script from https://github.com/stacks-network/sbtc/issues/30. Is that going to be a separate issue?

```
<lock_time> CHECKSEQUENCEVERIFY <reclaim script>
```

---
*Edit:* This will be addressed in a later ticket: https://github.com/stacks-network/sbtc/pull/333#discussion_r1678300773

common/src/deposits.rs
```diff
@@ -83,11 +116,79 @@ impl DepositScript {
...
+/// ```
+pub fn parse_deposit_script(deposit_script: &ScriptBuf) -> Result<DepositScriptInputs, Error> {
+    let script = deposit_script.as_bytes();
+
+    // Valid deposit scripts cannot be less than this length.
+    if script.len() < STANDARD_SCRIPT_LENGTH {
+        return Err(Error::InvalidDepositScript);
+    }
+    // This cannot panic because of the above check and the fact that
+    // DEPOSIT_SCRIPT_FIXED_LENGTH < STANDARD_SCRIPT_LENGTH.
```
netrome, line 143: Oh I didn't realize `split_at` was a panicking function. Thanks for the clarification!

netrome (approved):
Looks good to me :+1:

PR #344, opened 2024-07-17
feat: add reclaim script parsing function

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/328.

This PR is analogous to https://github.com/stacks-network/sbtc/pull/334, which parsed the deposit script and extracted the relevant variables for validation. Here we parse the reclaim script for validation.

Although the reclaim script is conceptually simpler, since there is only one variable that we care about, extracting that variable happens to be slightly more annoying due to how inputs to `OP_CSV` (and `OP_CLTV`) handle their inputs in bitcoin-core. I try to add enough documentation about this.

## Changes

* Add a function that can parse a reclaim script and extract the relevant pieces of information (the lock time and the other pieces).
* I basically duplicated some functions from rust-bitcoin here (I needed to modify them slightly). The change mirrors (exactly?) some of the code in bitcoin-core.

## Testing information

The tests check that the parser function is an inverse of the `ReclaimScriptInputs::reclaim_script` function.

### DIFF
--- a/common/src/deposits.rs
+++ b/common/src/deposits.rs
@@ -39,6 +39,15 @@ pub enum Error {
     /// The deposit script was invalid
     #[error("Invalid deposit script")]
     InvalidDepositScript,
+    /// The lock time included in the reclaim script was invalid.
+    #[error("the lock time included in the reclaim script was invalid: {0}")]
+    InvalidReclaimScriptLockTime(i64),
+    /// The reclaim script was invalid.
+    #[error("the reclaim script format was invalid")]
+    InvalidReclaimScript,
+    /// The deposit script was invalid
+    #[error("Invalid deposit script")]
+    ScriptError(#[source] bitcoin::script::Error),
     /// The X-only public key was invalid
     #[error("the x-only public key in the script was invalid: {0}")]
     InvalidXOnlyPublicKey(#[source] secp256k1::Error),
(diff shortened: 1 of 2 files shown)


xoloki (approved):
LGTM, but some of the parsing got a little tricky so def good to have more eyes on it

common/src/deposits.rs
```diff
@@ -183,6 +201,150 @@ pub fn parse_deposit_script(deposit_script: &ScriptBuf) -> Result<DepositScriptI
...
+///
+/// This code was lifted from rust-bitcoin without modification:
+/// https://github.com/rust-bitcoin/rust-bitcoin/blob/bitcoin-0.32.2/bitcoin/src/blockdata/script/mod.rs#L218-L226C2.
+///
+/// The logic there follows the logic in bitcoin-core:
+/// https://github.com/bitcoin/bitcoin/blob/v27.1/src/script/script.h#L382C1-L400C3
+fn scriptint_parse(v: &[u8]) -> i64 {
+    let (mut ret, sh) = v
+        .iter()
+        .fold((0, 0), |(acc, sh), n| (acc + ((*n as i64) << sh), sh + 8));
```
AshtonStephens, line 340: wild

AshtonStephens: but I understand.

AshtonStephens (approved):
LGTM.

Issue #332, opened 2024-07-15
[Feature]: Accommodate contract addresses in complete-deposit contract calls


djordon:
# Feature - Accommodate contract addresses in complete-deposit contract calls

### 1.1 Context & Purpose

The signers currently only handle standard principals in the complete deposit contract calls. This ticket is to handle contract principals as well.

### 2.1 Acceptance Criteria:
   - [ ] We can handle contract principals as recipients in complete deposit contract calls.

PR #345, opened 2024-07-17
feat: support contract principals in complete-deposit contract calls

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/332.

## Changes

* Update the contract-call struct to support any stacks principal as the recipient.
* Add integration test cases that exercise the contract principal case.

## Testing information

I ran the integration tests locally, using the same setup as in https://github.com/stacks-network/sbtc/pull/311, and tests passed.

Note that, similar to standard recipients, you do not need to "materialize" or "deploy" the smart contract before it can accept STX or other fungible tokens.


AshtonStephens (approved):
LGTM.

Issue #330, opened 2024-07-15
[Feature]: Validate BTC transaction


djordon:
# Feature - Validate BTC transaction

### 1.1 Context & Purpose

This is part of https://github.com/stacks-network/sbtc/issues/322, and has the same context and but with a focus on validating the BTC transaction given the deposit script and the reclaim script.

### 2.1 Acceptance Criteria:
   - [ ] Have a function that validates a bitcoin transaction given the reclaim script and the deposit script.

## 3. Related Issues and Pull Requests (optional):

Relates to the work outlined in the LLD referenced in https://github.com/stacks-network/sbtc/issues/48 and is part of https://github.com/stacks-network/sbtc/issues/322, and builds off of https://github.com/stacks-network/sbtc/issues/327, https://github.com/stacks-network/sbtc/issues/328, and https://github.com/stacks-network/sbtc/issues/329.

PR #350, opened 2024-07-18
feat: validate bitcoin deposit transactions

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/330.

The last part (or second-to-last part) of "BTC deposit transaction analysis" is validating that the transaction is "valid". Basically, we check that the deposit and reclaim scripts in the transaction follow our proscribed formats.

## Changes
1. A slight refactor involving the parsing functions https://github.com/stacks-network/sbtc/pull/344 and https://github.com/stacks-network/sbtc/pull/334.
2. Add a function that takes in a BTC transaction encoded as a hex string and makes sure that the relevant UTXO matches what was sent in the `CreateDepositRequest` (the object from the [Emily API LLD](https://docs.google.com/document/d/13AN9GdLfPYKKPkeHNU4niq_CyJCDxFR7DGBLzLc5gxU/edit#heading=h.hbinuw8xakzd)).
3. Update some of the comments in the signer codebase.

## Testing information

The unit tests check that:
* We detect mismatches between what is in the transaction and what is in the user's request.
* We detect deposit and reclaim scripts that do not follow the expected format.

### DIFF
--- a/common/src/deposits.rs
+++ b/common/src/deposits.rs
@@ -8,7 +8,10 @@ use bitcoin::taproot::NodeInfo;
 use bitcoin::taproot::TaprootSpendInfo;
 use bitcoin::Address;
 use bitcoin::Network;
+use bitcoin::OutPoint;
 use bitcoin::ScriptBuf;
+use bitcoin::Transaction;
+use bitcoin::Txid;
 use bitcoin::XOnlyPublicKey;
 use clarity::codec::StacksMessageCodec;
 use clarity::vm::types::PrincipalData;
@@ -57,7 +60,7 @@ const OP_PUSHNUM_NEG1: u8 = opcodes::OP_PUSHNUM_NEG1.to_u8();
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
     /// The deposit script was invalid
-    #[error("Invalid deposit script")]
+    #[error("invalid deposit script")]
     InvalidDepositScript,
     /// The lock time included in the reclaim script was invalid.
     #[error("the lock time included in the reclaim script was invalid: {0}")]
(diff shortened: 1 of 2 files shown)


common/src/deposits.rs
```diff
@@ -123,84 +281,65 @@ impl DepositScriptInputs {
...
+        // need to check all cases since contract addresses can have a size
+        // of up to 151 bytes.
+        let data = match params {
+            // This branch represents a contract address.
+            [OP_PUSHDATA1, n, data @ ..] if data.len() == *n as usize && *n < 160 => data,
+            // This branch can be a standard (non-contract) Stacks
+            // addresses when n == 30 and is a contract address otherwise.
+            [n, data @ ..] if data.len() == *n as usize && *n < 76 => data,
+            _ => return Err(Error::InvalidDepositScript),
+        };
```
hstove, line 324: Can you describe what the rust syntax `data @ ..` is doing? I've also never seen this match syntax, it's interesting. Also, why `n < 160`, if contract addresses can be up to 151 bytes?

hstove: Ah, per [this stacks-core code](https://github.com/stacks-network/stacks-core/blob/202b45cc65b88a15d31a33370e47850cf98d7c21/clarity/src/vm/representations.rs#L81), it seems that the max size of a contract name is 128 bytes, so 128 + 30 = 158, and 1 byte for the prefix, which results in a max of 159. I get the "why n < 160" now.

common/src/deposits.rs
```diff
@@ -236,53 +375,53 @@ impl ReclaimScriptInputs {
...
+    /// then we return an error.
+    ///
+    /// laid out in https://github.com/stacks-network/sbtc/issues/30, where
+    /// the script
+    pub fn parse(reclaim_script: &ScriptBuf) -> Result<Self, Error> {
+        let (lock_time, script) = match reclaim_script.as_bytes() {
+            // These first two branches check for the case when the script
+            // is written with as few bytes as possible (called minimal
+            // CScriptNum format or something like that).
+            [0, OP_CSV, script @ ..] => (0, script),
```
hstove, line 398: Is it valid to do `0 CSV`? I'm honestly not sure, just asking if you know. I also personally think that, assuming it's valid BScript, it's a good thing for us to support.

common/src/deposits.rs
```diff
@@ -236,53 +375,53 @@ impl ReclaimScriptInputs {
...
+            // This catches numbers 1-16 and -1.
+            [n, OP_CSV, script @ ..]
+                if OP_PUSHNUM_NEG1 == *n || (OP_PUSHNUM_1..OP_PUSHNUM_16).contains(n) =>
```
hstove, line 401: Per [BIP112 Summary](https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki#summary), it says the script will terminate if the top item of the stack is < 0. Am I reading that wrong and that -1 is valid?

common/src/deposits.rs
hstove: But now I'm realizing that `data` also includes the max fee (8 bytes), so does that make the max size 167?

hstove (approved):
Looking good! Great comments, it made it much easier to follow what's going on in each part of the code. I just had a few questions but no requested changes.

common/src/deposits.rs
djordon: > But now I'm realizing that `data` also includes the max fee (8 bytes), so does that make the max size 167?

Hmmm, I don't think so. For the principal, I thought it was 1 byte for the type prefix, 21 bytes for the standard address, 1 byte for the contract name, and 128 bytes for the contract name, for a total of 151 bytes. Then there is 8 bytes for the max fee, bringing it to 159 bytes. Oh, I need to update the comment to note that it includes the max fee!

Anyway, does that make sense? And thanks for looking this part over again.

common/src/deposits.rs
djordon: > Is it valid to do `0 CSV`?

I thought so. I also thought that almost any 5 byte signed number was valid, which I think you noted below was incorrect.

> I also personally think that, assuming it's valid BScript, it's a good thing for us to support.

Oh, well I don't reject it here in this library but I was thinking that the signers would likely reject it anyway. Why do you think that we should support it?

common/src/deposits.rs
djordon: No, you read it correctly. I reject it later, but because I thought that it would be non-standard, I forgot that it was invalid too. I'll reject it here as well.

djordon: Oh, I should just note that negative number are invalid. It's probably best to catch this all in one place.

common/src/deposits.rs
djordon: I'm going to add a test for the max length contract parse.

hstove: Yeah that makes sense, when I read your comment here "when n == 30 it's standard" I just assumed the address part was 30 bytes and didn't think about it. Yeah, it's 1 byte version, 20 byte pubkey hash.

Math checks out! For future proofing, can you add a test case where a contract address is maxed out? Maybe just change the `CONTRACT_ADDRESS` var to be max length?

common/src/deposits.rs
hstove: My reasoning is that if someone puts a 0-block locktime, then signers can spend it immediately (assuming thats how it works). Maybe not what a user would want, but it's "good" for signers.

common/src/deposits.rs
hstove: Do you think you could add a comment somewhere in this file outlining the "wire format" for `data`? I find [this kind of format](https://github.com/stacks-network/stacks-core/blob/202b45cc65b88a15d31a33370e47850cf98d7c21/stackslib/src/chainstate/burn/operations/stack_stx.rs#L188-L191) helpful:

```
0      2  3                             19           20                  53                 69                        73
|------|--|-----------------------------|------------|-------------------|-------------------|-------------------------|
magic  op         uSTX to lock (u128)     cycles (u8)     signer key (optional)   max_amount (optional u128)  auth_id (optional u32)
```

djordon: > Do you think you could add a comment somewhere in this file outlining the "wire format" for `data`?

Oh yeah, I find it very useful. Good call, I'll add that here.

> Math checks out! For future proofing, can you add a test case where a contract address is maxed out? Maybe just change the `CONTRACT_ADDRESS` var to be max length?

Not a problem, added one here https://github.com/stacks-network/sbtc/pull/350/commits/63b46c04ca17e1f877a55851e0b9d97b3b9bf970. I think this is what you had in mind, I wanted to add one and just forgot.

common/src/deposits.rs
djordon: Ahh I see. This is supposed to be the relative lock time used at the start of the reclaim script. So low numbers allows the depositor to reclaim their funds more quickly. At least, that is my understanding.

common/src/deposits.rs
djordon: Added a wire format diagram thingy in https://github.com/stacks-network/sbtc/pull/350/commits/a0f146e9647d6776fc6751ac669f4c75d84eeac2.

common/src/deposits.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/350/commits/63b46c04ca17e1f877a55851e0b9d97b3b9bf970

PR #351, opened 2024-07-18
refactor: rename `sbtc-common` library to `sbtc`

djordon:
## Description

Rename the sbtc-common crate to sbtc.

## Changes

1. Rename the `sbtc-common` library to `sbtc`.
2. Allow the logging function in the `sbtc` library to support custom directives. This is needed in order to properly filter log levels for the importing binary (instead of essentially a blanket `info` log filter).
3. Don't have the blocklist client import the `sbtc` crate. No need, just copy over the necessary logging utilities.

Shouldn't

## Testing information

If tests pass then we're good.


sbtc/src/logging.rs
djordon, line 1: This is almost identical from what it was before, the functions now just accept logging directives as an input variable.

AshtonStephens: Names look good, assuming the CI passes I'm approving.

AshtonStephens (approved):
LGTM!

Issue #354, opened 2024-07-22
[Feature]: Add signer bitmap to BTC sweep transactions


djordon:
# Feature - Add signer bitmap to BTC sweep transactions

## 1. Description

The design in https://github.com/stacks-network/sbtc/issues/19 specifies that we encode the signers votes for the transaction in an `OP_RETURN` output. This work hasn't been done yet.

### 2.1 Acceptance Criteria:
   - [ ] Add another output to all BTC transaction with a bitmap for the signers that have signed off for the transaction.

PR #356, opened 2024-07-23
feat: add signer bitmap to BTC sweep transactions

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/354.

Add an `OP_RETURN` output containing the signer bitmap for the transaction.

## Changes

1. Use a `BitArray` for the type that contains how the signers voted for the transaction. This puts it more inline with how we represent it elsewhere in the signer codebase.
2. Update the BTC transactions to include an `OP_RETURN` output as the second output in BTC transactions whenever the signers are servicing a request. Leave it out otherwise.
3. Update the deposit script used in tests to match the actual one that we expect. Update the size constants given the new `OP_RETURN` output and the updated script.

## Testing information

The unit and integration tests are sufficient.


signer/tests/integration/rbf.rs
```diff
@@ -323,7 +328,7 @@ pub fn transaction_with_rbf(
             utx.requests
                 .iter()
                 .filter_map(RequestRef::as_withdrawal)
-                .zip(utx.tx.output.iter().skip(1))
+                .zip(utx.tx.output.iter().skip(2))
```
jferrant, line 331: I assume this skip 2 is to skip the magic bytes? Might be worth adding a comment to that effect.

djordon: Oh, right yeah. The outputs always follow a particular order: the first two outputs are the signers' UTXO and the `OP_RETURN` outputs, and then follows outputs servicing withdrawal requests. This test code only cares about the withdrawal request outputs, I'll make sure to add a comment here to that effect.

xoloki (approved):
I'm not particularly familiar with this code, but it looks okay to me

Issue #233, opened 2024-06-05
[Feature]: Add support for estimating fees for sBTC contract calls


djordon:
# Feature - Add support for estimating fees for sBTC contract calls

### 1.1 Context & Purpose

When broadcasting contract calls, we need to make sure that we submit the transaction with enough fees for the contract call to be confirmed promptly. Add some functions that call a stacks nodes or the stacks/Hiro API for fee estimates.

#### Data sources

1. The stacks mode might have the `/v2/fees/transaction` endpoint (from [docs.stacks.co](https://docs.stacks.co/stacks-101/api#v2-fees-transaction), [docs.hiro.so](https://docs.hiro.so/api/get-approximate-fees-for-a-given-transaction), and [the code](https://github.com/stacks-network/stacks-core/blob/9c4a445c85b8f310fe3fe89a168897a2759081b7/stackslib/src/net/api/postfeerate.rs#L105-L116) referencing the stacks-core node API but missing from the [rpc-endpoints.md](https://github.com/stacks-network/stacks-core/blob/master/docs/rpc-endpoints.md) file in the repo stacks-core repo).
2. The stacks node has the `/v2/fees/transfer` [endpoint](https://github.com/stacks-network/stacks-core/blob/master/docs/rpc-endpoints.md#get-v2feestransfer).

### 2.1 Acceptance Criteria:
   - [ ] We can get an relatively accurate estimate of the contract call fees.

## 3. Related Issues and Pull Requests (optional):

This is part of #44 and recently came up during this discussion https://github.com/stacks-network/sbtc/pull/228#discussion_r1627170146.

PR #361, opened 2024-07-29
feat: support estimating contract calls fees

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/233.

We need the current priority  mempool fees for stacks transactions. This PR adds a function that uses the `POST /v2/fees/transaction` endpoint on stacks-core for this.

All fees returned are the high-priority fee amount (not the fee rate). And the implementation for estimating fees in the trait falls back on the high-priority STX transfer fee in the event that the node cannot estimate fees for the transaction.

## Changes

* Add a function that estimates transaction fees (for contract calls or otherwise).
* Add a function to the `StacksInterface` trait for estimating fees, along with a "robust" implementation that falls back to the STX transfer fee estimate if the node does not know how much the particular transaction will cost.

## Testing

This PR adds an integration test that should work with both a pre- and -post-nakamoto stacks-node. It's not set up to run in CI.

### DIFF
--- a/signer/src/stacks/api.rs
+++ b/signer/src/stacks/api.rs
@@ -135,7 +135,7 @@ pub struct TxRejection {
 ///
 /// The stacks node returns three types of responses, either:
 /// 1. A 200 status hex encoded txid in the response body (on acceptance)
-/// 2. A 400 status with s JSON object body (on rejection),
+/// 2. A 400 status with a JSON object body (on rejection),
 /// 3. A 400/500 status string message about some other error (such as
 ///    using an unsupported address mode).
 ///
@@ -232,6 +232,8 @@ impl StacksClient {
             .map_err(Error::StacksNodeRequest)?;
 
         response
+            .error_for_status()
+            .map_err(Error::StacksNodeResponse)?
             .json::<AccountEntryResponse>()
             .await
             .map_err(Error::UnexpectedStacksResponse)
@@ -296,7 +298,10 @@ impl StacksClient {
             .send()
             .await
             .map_err(Error::StacksNodeRequest)?;
+
         let resp = response
+            .error_for_status()
+            .map_err(Error::StacksNodeResponse)?
             .bytes()
             .await
             .map_err(Error::UnexpectedStacksResponse)?;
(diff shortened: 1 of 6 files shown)


signer/src/stacks/api.rs
```diff
@@ -435,6 +524,46 @@ impl StacksInteract for StacksClient {
...
+            .await?;
+        // As of this writing the RPC response includes exactly 3 estimates
+        // (the low, medium, and high priority estimates). We want to use
+        // the high priority estimate (the max) to make sure that the
+        // transaction gets confirmed as quickly as possible.
+        let fee_estimate = fee_estimates
+            .estimations
+            .iter()
+            .map(|estimate| estimate.fee)
+            .max()
```
AshtonStephens, line 562: Are we sure we want to do max and not median? One wonky API could make the fees huge.

AshtonStephens (changes requested):
Looks good, but I'd like a response to the "max" question I put even if we choose not to do it. Then I'll approve.

signer/src/stacks/api.rs
djordon: > Are we sure we want to do max and not median? One wonky API could make the fees huge.
> 
> If we want to do max lets make there a maximum fee config variable to safeguard against a bad API response.

I'm not certain if we should opt for the high priority fee either. On the Bitcoin side, we select the high priority fee, so I applied the same logic here. However, this analogy might not be right since the signers are the ones paying. Medium priority is more cost-effective for them, while high priority could enhance the user experience. I'm open to either approach.

djordon: > If we want to do max lets make there a maximum fee config variable to safeguard against a bad API response.

Yeah that sounds reasonable to me. But I think that also applies to taking the medium. And we should also do that for fees on the bitcoin side. I'll open a ticket for this, we can always choose to de-prioritize it.

djordon: Oh, we don't quite have have this issue on the bitcoin side, since we take an average of two sources.

djordon: I opened https://github.com/stacks-network/sbtc/issues/366 for this.

AshtonStephens:
To move us along I'm happy to approve this ticket, but we should make a TODO note above the `max` part and then add a ticket. The one you just made could encompass the configuration feature as well.

djordon:
> To move us along I'm happy to approve this ticket, but we should make a TODO note above the `max` part and then add a ticket. The one you just made could encompass the configuration feature as well.

Done in https://github.com/stacks-network/sbtc/pull/361/commits/0092844f52c096f80348ab0d9e1b33115592a8d1

xoloki (approved):
Lgtm.  I agree that using the max fee is suboptimal, but not worth holding up this PR

AshtonStephens (approved):
LGTM.

PR #362, opened 2024-07-29
refactor: bump rust to 1.80

djordon:
## Description

Bump rust to 1.80, update to use the now stabilized `LazyLock` type instead of the `once-cell` equivalents.

## Changes

* Bump the rust version to 1.80 from 1.77.2.
* Switch to using `std::sync::LazyLock` instead of `once_cell::sync:Lazy` thoughout.
* Various doc updates after new clippy lints. These clippy checks were surprising since clippy now checks the docs. Maybe it always did but we always passed those checks?
* I also committed the `cargo fmt` changes for stuff in the `.generated-sources` folder.


AshtonStephens (approved):
LGTM

PR #367, opened 2024-07-30
ci: add deposit validation integration tests

djordon:
## Description

This is a follow-up to https://github.com/stacks-network/sbtc/pull/364 with integration tests and a slight refactor of testing stuff.

## Changes

* Move the integration test helper functions to the `sbtc::testing` module so that both the `signer` and the `sbtc` crates can use them.
* Move other testing functions to the `sbtc::testing` module.
* Add `electrs` to the test services run in CI.
* Add integration tests that exercise the functionality added in https://github.com/stacks-network/sbtc/pull/364.

## Testing information

This PR adds the `electrs` service to CI, but uses a different one from the one in devnet. The devnet one builds off of [mempool/electrs](https://github.com/mempool/electrs), but this one uses the `getumbrel/electrs` container from docker hub, which builds off of [romanz/electrs](https://github.com/romanz/electrs). I'll reconcile them eventually.


sbtc/tests/integration/rpc.rs
```diff
@@ -0,0 +1,88 @@
...
+use sbtc::testing::regtest::Recipient;
+
+use sbtc::rpc::BitcoinRpcClient;
+use sbtc::rpc::BtcClient;
+use sbtc::rpc::ElectrumClient;
+use test_case::test_case;
+
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
+#[test_case(BtcClient::new(
+        "http://localhost:18443",
```
jferrant, line 16: Should we use a hardcoded port? Wonder if it would be better to select one dynamically? Same goes with below.

jferrant (approved):

Issue #366, opened 2024-07-30
[Feature]: Make stacks transaction fee estimation more robust


djordon:
# Feature - Make stacks transaction fee estimation more robust

## 1. Description

We only have one source for expected fees for stacks transactions, which could potentially return weird estimates for the expected fees. This ticket is to have some safeguards in place for fee estimation.

Also, we need to figure out whether we should use the high or medium priority transaction fee estimates for sBTC related stacks transactions.

### 2.1 Acceptance Criteria:
   - [ ] Make the stacks transaction fee estimation code more robust.
   - [ ] Maybe change to using the medium priority stacks transaction fee estimate.

## 3. Related Issues and Pull Requests (optional):

This came up in this discussion https://github.com/stacks-network/sbtc/pull/361#discussion_r1697009463, which addressed https://github.com/stacks-network/sbtc/issues/233.

djordon:
Hmm, we probably want high priority for `rotate-keys` contract calls and medium (or maybe even low) priority for other transactions, like `withdrawal-accept`.

PR #368, opened 2024-07-30
feat: robust stacks transaction fee estimation

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/366. And is a follow-up to https://github.com/stacks-network/sbtc/pull/361.

## Changes

* Cap the maximum fee that signers will pay for a transaction to 10 STX.
* Allow the fee priority to be selected by the caller of the estimation function.

## Testing information

The existing tests we updated to support the new functionality.

### DIFF
--- a/signer/src/stacks/api.rs
+++ b/signer/src/stacks/api.rs
@@ -851,7 +851,7 @@ mod tests {
             .with_status(200)
             .with_header("content-type", "application/json")
             .with_body(raw_json_response)
-            .expect(2)
+            .expect(4)
             .create();
 
         let settings = StacksSettings {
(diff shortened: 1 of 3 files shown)


jferrant (approved):

signer/src/stacks/api.rs
```diff
@@ -552,18 +584,22 @@ impl StacksInteract for StacksClient {
...
-            .iter()
-            .map(|estimate| estimate.fee)
-            // TODO(366): have priority be configurable.
-            .max()
-            .unwrap_or(DEFAULT_TX_FEE);
+        // (the low, medium, and high priority estimates). It's note worthy
+        // if this changes so we log it but the code here is robust to such
+        // a change.
+        let num_estimates = resp.estimations.len();
+        if num_estimates != 3 {
```
jferrant, line 591: Does it EVER return an estimate that is not 3? D:

djordon: We'll find out when it's deployed!

PR #369, opened 2024-07-31
feat: add electrum for fee estimates

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/160.

During the design phase, we aligned on using electrum as a source for fee estimates. This PR adds two implementations, one from electrum and another from bitcoin-core. The bitcoin-core implementation makes testing easier, since some implementations of electrum use bitcoin-core's `estimatesmartfee` RPC under the hood.

## Changes

* Add an implementation that estimates fees using electrum.
* Add an implementation that estimates fees using bitcoin-core.
* Update the `fees` module in the signer create to return a `FeeEstimate` type.
* Updated the error messages in the `sbtc` crate, I forgot to do this in an earlier PR.

## Testing information

This PR adds an integration test for the `electrum` and `bitcoin-core` estimate fee implementations, but I stopped short of making the test "self-contained". In order to test the we get reasonable fee estimates we need to make sure there are enough transactions on the block chain (I'm not sure what counts as enough). I tested these functions locally by running all integration tests and then running a slightly modified version of the `estimate_fee_rate` test to see if the output was reasonable.

### DIFF
--- a/sbtc/src/deposits.rs
+++ b/sbtc/src/deposits.rs
@@ -262,10 +262,12 @@ impl DepositScriptInputs {
     /// the format for a principal from SIP-005. So the expected wire
     /// format is:
     ///
+    /// ```text
     /// 0         8             9         10        30            31             159
     /// |---------|-------------|---------|---------|-------------|---------------|
     ///   max fee   type prefix   version   hash160   name length   contract name
     ///                                               (optional)    (optional)
+    /// ```
     ///
     /// Above, the max fee is expressed as an 8-byte big endian integer and
     /// the contract name is a UTF-8 encoded string and must be accepted by
--- a/sbtc/src/rpc.rs
+++ b/sbtc/src/rpc.rs
@@ -4,11 +4,14 @@ use std::num::NonZeroU8;
 
 use bitcoin::consensus;
 use bitcoin::consensus::Decodable as _;
+use bitcoin::Amount;
 use bitcoin::BlockHash;
+use bitcoin::Denomination;
 use bitcoin::Transaction;
 use bitcoin::Txid;
+use bitcoincore_rpc::json::EstimateMode;
 use bitcoincore_rpc::Auth;
-use bitcoincore_rpc::RpcApi;
+use bitcoincore_rpc::RpcApi as _;
 use electrum_client::ElectrumApi as _;
 use electrum_client::Param;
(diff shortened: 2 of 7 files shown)


sbtc/src/deposits.rs
```diff
@@ -262,10 +262,12 @@ impl DepositScriptInputs {
     /// the format for a principal from SIP-005. So the expected wire
     /// format is:
     ///
+    /// ```text
```
AshtonStephens, line 265: What made you add this? Is there a docs command I should be running?

sbtc/src/rpc.rs
```diff
@@ -4,11 +4,14 @@ use std::num::NonZeroU8;
...
 use bitcoin::consensus::Decodable as _;
+use bitcoin::Amount;
 use bitcoin::BlockHash;
+use bitcoin::Denomination;
 use bitcoin::Transaction;
 use bitcoin::Txid;
+use bitcoincore_rpc::json::EstimateMode;
 use bitcoincore_rpc::Auth;
-use bitcoincore_rpc::RpcApi;
+use bitcoincore_rpc::RpcApi as _;
```
AshtonStephens, line 14: That's this for?

sbtc/src/rpc.rs
```diff
@@ -44,6 +47,14 @@ pub struct GetTxResponse {
     pub in_active_chain: Option<bool>,
 }
 
+/// A struct representing the recommended fee, in sats per vbyte, from a
+/// particular source.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct FeeEstimate {
```
AshtonStephens, line 53: What's the thought process behind making this a struct with a single field?

AshtonStephens (approved):
LGTM. Put some questions.

sbtc/src/deposits.rs
djordon: I use an extension for formatting my comments and it will format the diagram below unless it knows that it shouldn't be. This is one way to tell it that these lines shouldn't be formatted.

sbtc/src/rpc.rs
djordon: This is to communicate that this imported thing is a trait that is necessary but isn't directly used. It's more of a stylistic change than anything.

PR #372, opened 2024-07-31
365 feature emily api default database entries

AshtonStephens:
## Description

Adds both a script for populating the Emily API with some default values and adds some Makefile commands to perform those commands.

To run the emily API locally with the pre-populated database run the following:

```
make emily-integration-env-up-populated-database
```

To repopulate the database you can run the following:

```
make emily-populate-database
```

This change works by introducing two sub directories of the `./devenv/dynamodb/` directory:
- `./devenv/dynamodb/populated`
- `./devenv/dynamodb/dynamic`

The populated directory will contain the populated database and the dynamic one will expect to take changes and be used for dynamic testing.

## Type of Change

- New feature

No


setbern (approved):
LGTM

devenv/dynamodb/populated/shared-local-instance.db
djordon, line 1: Is this supposed to be committed?

djordon (commented):
Looks good, just had one question.

devenv/dynamodb/populated/shared-local-instance.db
AshtonStephens: Yes

djordon (approved):
Looks good. I think all is needed a `cargo fmt`

Issue #225, opened 2024-06-04
[Feature]: Support multiple Stacks node URLs


djordon:
# Feature - Support multiple Stacks node URLs

### 1.1 Context & Purpose

Right now we only support one endpoint for getting nakamoto block data from a stacks node. Add support so that users can specify more than one endpoint for fetching stacks related data. Specifically, modify the code to expect a comma separated list of stacks node URLs in the signer's `config.toml`.

### 2.1 Acceptance Criteria:

It's not clear if we'll want to always try the first endpoint in the list and fallback to the next one if that fails or do some kind of round robin load balancing. I'm leaning toward always trying the first one and falling back to the next one on connection/network errors since it is simpler.

## 3. Related Issues and Pull Requests (optional):

The original ticket for implementing the stacks client is #176.

PR #382, opened 2024-08-01
Update signer to support multiple stacks node urls

xoloki:
## Description
Update code and config to support multiple stacks urls.

Closes: #225 

## Changes
Config and stacks api now store multiple endpoints. Stacks api also keeps an index for the current endpoint. This index is incremented on network failure. 

## Testing Information
Existing tests have been updated to check previous and new functionality. 

## Checklist:

### DIFF
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -814,7 +814,7 @@ mod tests {
                 .map(|(_, block, _)| block.header.chain_length)
                 .unwrap_or_default()
         }
-        async fn estimate_fees<T>(&mut self, _: &T, _: FeePriority) -> Result<u64, error::Error>
+        async fn estimate_fees<T>(&self, _: &T, _: FeePriority) -> Result<u64, error::Error>
         where
             T: crate::stacks::contracts::AsTxPayload,
         {
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -742,10 +742,12 @@ mod tests {
 
     impl StacksInteract for TestHarness {
         async fn get_account(&mut self, _address: &StacksAddress) -> Result<AccountInfo, Error> {
+            // issue #118
             todo!()
         }
 
         async fn submit_tx(&mut self, _tx: &StacksTransaction) -> Result<SubmitTxResponse, Error> {
+            // issue #118
             todo!()
         }
 
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -158,7 +158,7 @@ where
     async fn process_bitcoin_block(&mut self, block: bitcoin::Block) -> Result<(), error::Error> {
         let info = self.stacks_client.get_tenure_info().await?;
         let stacks_blocks = crate::stacks::api::fetch_unknown_ancestors(
-            &self.stacks_client,
+            &mut self.stacks_client,
             &self.storage,
             info.tip_block_id,
         )
@@ -290,13 +290,18 @@ mod tests {
     use blockstack_lib::chainstate::burn::ConsensusHash;
     use blockstack_lib::chainstate::nakamoto::NakamotoBlock;
     use blockstack_lib::chainstate::nakamoto::NakamotoBlockHeader;
+    use blockstack_lib::chainstate::stacks::StacksTransaction;
     use blockstack_lib::net::api::gettenureinfo::RPCGetTenureInfo;
+    use blockstack_lib::types::chainstate::StacksAddress;
     use blockstack_lib::types::chainstate::StacksBlockId;
     use rand::seq::IteratorRandom;
     use rand::SeedableRng;
     use sbtc::rpc::BitcoinClient;
 
+    use crate::error::Error;
+    use crate::stacks::api::AccountInfo;
     use crate::stacks::api::FeePriority;
+    use crate::stacks::api::SubmitTxResponse;
     use crate::storage;
     use crate::testing::dummy;
 
@@ -736,7 +741,18 @@ mod tests {
     }
 
     impl StacksInteract for TestHarness {
-        async fn get_block(&self, block_id: StacksBlockId) -> Result<NakamotoBlock, error::Error> {
+        async fn get_account(&mut self, _address: &StacksAddress) -> Result<AccountInfo, Error> {
+            todo!()
+        }
+
+        async fn submit_tx(&mut self, _tx: &StacksTransaction) -> Result<SubmitTxResponse, Error> {
+            todo!()
+        }
+
+        async fn get_block(
+            &mut self,
+            block_id: StacksBlockId,
+        ) -> Result<NakamotoBlock, error::Error> {
             self.stacks_blocks
                 .iter()
                 .skip_while(|(id, _, _)| &block_id != id)
@@ -746,7 +762,7 @@ mod tests {
                 .ok_or(error::Error::MissingBlock)
         }
         async fn get_tenure(
-            &self,
+            &mut self,
             block_id: StacksBlockId,
         ) -> Result<Vec<NakamotoBlock>, error::Error> {
             let (stx_block_id, stx_block, btc_block_id) = self
@@ -768,7 +784,7 @@ mod tests {
 
             Ok(blocks)
         }
-        async fn get_tenure_info(&self) -> Result<RPCGetTenureInfo, error::Error> {
+        async fn get_tenure_info(&mut self) -> Result<RPCGetTenureInfo, error::Error> {
             let (_, _, btc_block_id) = self.stacks_blocks.last().unwrap();
 
             Ok(RPCGetTenureInfo {
@@ -798,7 +814,7 @@ mod tests {
                 .map(|(_, block, _)| block.header.chain_length)
                 .unwrap_or_default()
         }
-        async fn estimate_fees<T>(&self, _: &T, _: FeePriority) -> Result<u64, error::Error>
+        async fn estimate_fees<T>(&mut self, _: &T, _: FeePriority) -> Result<u64, error::Error>
         where
             T: crate::stacks::contracts::AsTxPayload,
         {
--- a/signer/src/config.rs
+++ b/signer/src/config.rs
@@ -92,13 +92,16 @@ impl Settings {
 }
 
 /// A deserializer for the url::Url type.
-fn url_deserializer<'de, D>(deserializer: D) -> Result<url::Url, D::Error>
+fn url_deserializer<'de, D>(deserializer: D) -> Result<Vec<url::Url>, D::Error>
 where
     D: Deserializer<'de>,
 {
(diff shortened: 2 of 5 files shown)


signer/src/stacks/api.rs
```diff
@@ -250,7 +249,10 @@ impl TryFrom<AccountEntryResponse> for AccountInfo {
 pub struct StacksClient {
     /// The base URL (with the port) that will be used when making requests
     /// for to a Stacks node.
-    pub node_endpoint: url::Url,
+    pub node_endpoints: Vec<url::Url>,
+    /// The current index into the endpoints list
+    /// Increment this on network failure
+    pub endpoint_index: usize,
```
djordon, line 255: I think we should make these fields private, since there are certain invariants that the struct needs to uphold. The `node_endpoints` must be non-empty and the `endpoint_index` value is always less than the `node_endpoints`' length.

---

As an aside, you could have a `node_endpoint` field that is still a `url::Url` and a `endpoints` field that is a `std::iter::Cycle<std::vec::IntoIter<url::Url>>`. Getting the next endpoint is then just 
```rust
// You can unwrap because an invariant is the `Vec` that created
// the `IntoIter` was nonempty.
self.node_endpoint = self.endpoints.next().unwrap();
```
Eh, actually, that might be overkill. Getting the endpoints the way that you have here is fine and more obvious if since you don't need to know about `std::iter::Cycle` type.

djordon (commented):
This looks okay, but I think we inadvertently introduced some behavior that we don't want. I think I see one main issue, and another "stylistic issue":
1. We switch to another stacks node even when the current stacks node that we were using was perfectly fine. For example, Stacks nodes return an error when they don't have enough data to estimate fees, the node could be perfectly fine. With these changes, we'd change node endpoints in that example, but I think it would be better to change node endpoints only when we suspect that the node is down.
2. We switch to a new node within the endpoint functions when we it feels more natural to switch outside of them. To me, it feels better to have a simple endpoint function that wraps the API call, and another that uses the simple endpoint function and rotates the node endpoint if it gets connectivity error (or a similar error) from the simple function. This other function could also retry the call after switching nodes (but we can save that part for another day). So I wouldn't change the endpoint functions here, probably just the trait functions.

Anyway, let me know if you think that these suggestions make sense.

xoloki:
1. Totally agree. Earlier versions of this PR only rotated urls on connection errors or if the response didn’t complete. But once I realized I could put a closure in map_err, I just did them all 😂
2. Disagree on this. What you’ve described sounds much more complicated, and other than saving a few mutables doesn’t seem to gain us anything.  Logically the API should know the URLs, and it knows when connection errors happen, so it should be the one to rotate the URLs.

signer/src/stacks/api.rs
```diff
@@ -394,9 +426,9 @@ impl StacksClient {
     /// If the given block ID does not exist or is an ID for a non-Nakamoto
     /// block then a Result::Err is returned.
     #[tracing::instrument(skip(self))]
-    async fn get_block(&self, block_id: StacksBlockId) -> Result<NakamotoBlock, Error> {
+    async fn get_block(&mut self, block_id: StacksBlockId) -> Result<NakamotoBlock, Error> {
```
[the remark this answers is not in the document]
xoloki: I really don’t like that idea. I know it has awkward consequences in other parts of the code, but it seems to wildly break encapsulation to do anything other than rotate the URLs inside these functions

signer/src/stacks/api.rs
```diff
@@ -394,9 +426,9 @@ impl StacksClient {
...
         let path = format!("/v3/blocks/{}", block_id.to_hex());
-        let base = self.node_endpoint.clone();
+        let base = self.node_endpoints[self.endpoint_index].clone();
```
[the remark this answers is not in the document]
xoloki: Agreed, will do.

signer/src/config.rs
```diff
@@ -117,7 +120,7 @@ pub struct StacksNodeSettings {
     ///
     /// The endpoint to use when making requests to a stacks node.
     #[serde(deserialize_with = "url_deserializer")]
-    pub endpoint: url::Url,
+    pub endpoints: Vec<url::Url>,
```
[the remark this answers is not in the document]
xoloki: Yeah, split should always return at least one string. If it doesn’t convert to a URL then the function fails, otherwise we get a vec with at least one entry.

signer/src/stacks/api.rs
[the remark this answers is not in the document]
xoloki: Done

signer/src/stacks/api.rs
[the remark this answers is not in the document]
xoloki: Moved endpoint rotation to trait as per our convo

signer/src/stacks/api.rs
```diff
@@ -374,16 +397,25 @@ impl StacksClient {
...
+                self.next_endpoint();
+                Error::StacksNodeResponse(e)
+            })?
             .json()
             .await
-            .map_err(Error::UnexpectedStacksResponse)
+            .map_err(|e| {
+                self.next_endpoint();
+                Error::UnexpectedStacksResponse(e)
+            })
```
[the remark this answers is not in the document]
xoloki: All trait functions are now rotating endpoint on network errors

signer/src/stacks/api.rs
```diff
@@ -535,25 +586,34 @@ impl StacksClient {
...
     }
 }
 
 impl StacksInteract for StacksClient {
-    async fn get_block(&self, block_id: StacksBlockId) -> Result<NakamotoBlock, Error> {
+    async fn get_block(&mut self, block_id: StacksBlockId) -> Result<NakamotoBlock, Error> {
```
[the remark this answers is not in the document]
xoloki: Done

signer/src/stacks/api.rs
xoloki: endpoint fields are now private

signer/src/config.rs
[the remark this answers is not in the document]
xoloki: Done

xoloki:
Ok @djordon I've addressed all review comments, and also squashed the bulk of the commits to make the PR less verbose.  lmk what you think.

signer/src/stacks/api.rs
```diff
@@ -547,14 +572,57 @@ impl StacksClient {
 }
 
 impl StacksInteract for StacksClient {
-    async fn get_block(&self, block_id: StacksBlockId) -> Result<NakamotoBlock, Error> {
-        self.get_block(block_id).await
+    async fn get_account(&mut self, address: &StacksAddress) -> Result<AccountInfo, Error> {
+        let result = StacksClient::get_account(self, address).await;
+        match result {
+            Err(Error::StacksNodeRequest(_)) | Err(Error::StacksNodeResponse(_)) => {
```
djordon, line 578: It's not clear that you want to match on all `Error::StacksNodeRequest` and all `Error::StacksNodeResponse` variants here. Instead you probably want to check the type of request error in `Error::StacksNodeRequest` and match on whether it is a timeout error, connect error, and maybe some other kind. Matching on all of it is probably too broad, so for now just match on timeout and connect errors.

signer/src/stacks/api.rs
```diff
@@ -565,24 +633,26 @@ impl StacksInteract for StacksClient {
     /// have enough information to provide an estimate, we then get the
     /// current high priority fee for an STX transfer and use that as an
     /// estimate for the transaction fee.
-    async fn estimate_fees<T>(&self, payload: &T, priority: FeePriority) -> Result<u64, Error>
+    async fn estimate_fees<T>(&mut self, payload: &T, priority: FeePriority) -> Result<u64, Error>
```
djordon, line 636: I'd just revert the changes here

signer/src/block_observer.rs
```diff
@@ -736,7 +741,18 @@ mod tests {
     }
 
     impl StacksInteract for TestHarness {
-        async fn get_block(&self, block_id: StacksBlockId) -> Result<NakamotoBlock, error::Error> {
+        async fn get_account(&mut self, _address: &StacksAddress) -> Result<AccountInfo, Error> {
+            todo!()
+        }
+
+        async fn submit_tx(&mut self, _tx: &StacksTransaction) -> Result<SubmitTxResponse, Error> {
+            todo!()
```
djordon, line 749: Whenever we add TODOs like this we've been adding a comment about the ticket that is associated with this TODO. I think we should continue that practice and add such a comment here.

signer/src/stacks/api.rs
```diff
@@ -547,14 +572,57 @@ impl StacksClient {
...
+    }
+    async fn get_tenure_info(&mut self) -> Result<RPCGetTenureInfo, Error> {
+        let result = StacksClient::get_tenure_info(self).await;
+        match result {
+            Err(Error::StacksNodeRequest(_)) | Err(Error::StacksNodeResponse(_)) => {
+                self.next_endpoint();
+            }
+            _ => (),
+        }
+        result
```
djordon, line 625: I think it would better to just create a function for this logic. I believe most of these functions can lean on the same logic here.

djordon (commented):
Great, I think this feels better. My main comments are around factoring out the identical logic into a function and matching on fewer error variants.

signer/src/stacks/api.rs
```diff
@@ -521,8 +546,8 @@ impl StacksClient {
     /// Uses the GET /v3/tenures/info stacks node endpoint for retrieving
     /// tenure information.
     #[tracing::instrument(skip(self))]
-    pub async fn get_tenure_info(&self) -> Result<RPCGetTenureInfo, Error> {
-        let base = self.node_endpoint.clone();
+    pub async fn get_tenure_info(&mut self) -> Result<RPCGetTenureInfo, Error> {
```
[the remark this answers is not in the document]
xoloki: Done.

signer/src/stacks/api.rs
xoloki: Done.

signer/src/stacks/api.rs
xoloki: Done.

signer/src/stacks/api.rs
xoloki: Done

signer/src/block_observer.rs
xoloki: Are we planning on using these functions in the block observer test harness?

xoloki: I tagged them with issue 118, which is to implement StacksInteract for the block observer

djordon: Oh oops, you're right. I don't think we're going to implement these. No need to modify it further, we can remove the TODO and replace it with a comment about how implementing these is unnecessary later.

djordon (approved):
Looks good ✅

Issue #203, opened 2024-05-29
[Feature]: Block observer - extract deposit requests


netrome:
### Context
The block observer should as part of its operation extract any pending deposit requests from the bitcoin blocks it's processing and write the extracted requests to the signer database.

### Definition of done
The `extract_deposit_requests` method in the block observer is implemented and tested.

djordon:
This one wasn't actually closed by https://github.com/stacks-network/sbtc/pull/315.

PR #385, opened 2024-08-02
feat: extract deposit requests in block observer

djordon:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/203.

## Changes

* Rename structs and to conform more closely to our specs.
* Add a new storage trait function, `write_deposit_requests`, that stores deposit requests.
* Implement the `write_deposit_requests` function for postgres and the in-memory thing.

## Testing Information

There are unit tests for the new block observer logic and an integration test for the query.

## Checklist:

### DIFF
--- a/sbtc/src/deposits.rs
+++ b/sbtc/src/deposits.rs
@@ -102,7 +102,7 @@ pub struct DepositInfo {
 pub struct Deposit {
     /// The transaction that includes the deposit information.
     pub tx: Transaction,
-    /// The information included in one of the outputs of the above
+    /// The deposit information included in one of the outputs of the above
     /// transaction.
     pub info: DepositInfo,
     /// The block hash of the block that includes this transaction. If this
@@ -532,7 +532,7 @@ mod tests {
 
     const CONTRACT_ADDRESS: &str = "ST1RQHF4VE5CZ6EK3MZPZVQBA0JVSMM9H5PMHMS1Y.contract-name";
 
-    struct DummyClient(HashMap<Txid, Transaction>);
+    struct DummyClient(pub HashMap<Txid, Transaction>);
 
     impl DummyClient {
         fn new_from_tx(tx: &Transaction) -> Self {
@@ -545,16 +545,10 @@ mod tests {
     impl BitcoinClient for DummyClient {
         type Error = Error;
         fn get_tx(&self, txid: &Txid) -> Result<GetTxResponse, Error> {
-            let tx = self
-                .0
-                .get(txid)
-                .cloned()
-                .ok_or(Error::GetTransactionElectrum(
-                    electrum_client::Error::CouldntLockReader,
-                    *txid,
-                ))?;
+            let tx = self.0.get(txid).cloned();
+
             Ok(GetTxResponse {
-                tx,
+                tx: tx.ok_or(Error::BitcoinClient(Box::new(Error::InvalidDepositScript)))?,
                 block_hash: None,
                 confirmations: None,
                 block_time: None,
--- a/sbtc/src/deposits.rs
+++ b/sbtc/src/deposits.rs
@@ -100,10 +100,10 @@ pub struct DepositInfo {
 /// A full deposit, with a fully extracted and verified `scriptPubKey`
 #[derive(Debug, Clone)]
 pub struct Deposit {
-    /// The transaction that includes the deposit information.
+    /// The transaction spent to the signers as a deposit for sBTC.
     pub tx: Transaction,
-    /// The deposit information included in one of the outputs of the above
-    /// transaction.
+    /// The deposit information included in one of the output
+    /// `scriptPubKey`s of the above transaction.
     pub info: DepositInfo,
     /// The block hash of the block that includes this transaction. If this
     /// is `None` then this transaction is in the mempool. TODO(384): In
--- a/sbtc/src/deposits.rs
+++ b/sbtc/src/deposits.rs
@@ -97,7 +97,9 @@ pub struct DepositInfo {
     pub lock_time: u64,
 }
 
-/// A full deposit, with a fully extracted and verified `scriptPubKey`
+/// A full "deposit", containing the bitcoin transaction and a fully
+/// extracted and verified `scriptPubKey` from one of the transaction's
+/// UTXOs.
 #[derive(Debug, Clone)]
 pub struct Deposit {
     /// The transaction spent to the signers as a deposit for sBTC.
--- a/sbtc/src/deposits.rs
+++ b/sbtc/src/deposits.rs
@@ -134,6 +134,7 @@ impl CreateDepositRequest {
         let response = client
             .get_tx(&self.outpoint.txid)
             .map_err(|err| Error::BitcoinClient(Box::new(err)))?;
+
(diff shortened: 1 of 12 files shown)


signer/tests/integration/postgres.rs
```diff
@@ -497,3 +498,45 @@ async fn should_return_the_same_pending_accepted_withdraw_requests_as_in_memory_
...
+        .unwrap();
+    let count =
+        sqlx::query_scalar::<_, i64>(r#"SELECT COUNT(*) FROM sbtc_signer.deposit_requests"#)
+            .fetch_one(&pool)
+            .await
+            .unwrap();
+
+    // No new records written right?
+    assert_eq!(num_rows, count as usize);
+}
```
AshtonStephens, line 542: Great comments on this test.

signer/src/storage/postgres.rs
```diff
@@ -687,6 +687,94 @@ impl super::DbWrite for PgStore {
...
+            max_fee.push(req.max_fee);
+            // We need to join the addresses like this (and later split
+            // them), because handling of multi-dimensional arrays in
+            // postgres is tough. The naive approach of doing
+            // UNNEST($1::VARCHAR[][]) doesn't work, since that completely
+            // flattens the array.
+            sender_addresses.push(req.sender_addresses.join(","));
+        }
+
+        sqlx::query(
```
AshtonStephens, line 723: Not for right now, but do we have any documentation on the structure of the SQL database? I find these very hard to follow and am really just trusting the unit tests here.

signer/src/block_observer.rs
```diff
@@ -440,7 +642,7 @@ mod tests {
     }
 
     impl EmilyInteract for () {
-        async fn get_deposits(&mut self) -> Vec<DepositRequest> {
+        async fn get_deposits(&mut self) -> Vec<CreateDepositRequest> {
```
AshtonStephens, line 645: This should likely be swapped for some autogenerated struct one day. Just flagging, for now it's fine and we may ultimately need to ship with this depending on timing.

sbtc/src/rpc.rs
```diff
@@ -117,12 +116,13 @@ impl BtcClient {
     ///
     /// Modified from the bitcoin-core docs[1]:
     ///
-    /// This returns a "conservative" estimate, which is potentially
-    /// returns a higher fee rate and is more likely to be sufficient for
-    /// the desired target, but is not as responsive to short term drops in
-    /// the prevailing fee market. Also, the docs mention the response is
-    /// in BTC/kB, but from the comments in  bitcoin-core[2] this is really
-    /// BTC/kvB (kvB is kilo-vbyte).
+    /// Bitcoin-core has too different modes for fee rate estimation,
```
AshtonStephens, line 119:
```suggestion
    /// Bitcoin-core has two different modes for fee rate estimation,
```

sbtc/src/rpc.rs
```diff
@@ -117,12 +116,13 @@ impl BtcClient {
...
+    /// "conservative" and "economical". We use the "conservative" estimate
+    /// because it is more likely to be sufficient for the desired target,
+    /// but is not as responsive to short term drops in the prevailing fee
+    /// market when compared to the "economical" fee rate. Also, the docs
+    /// mention the response is in BTC/kB, but from the comments in
+    /// bitcoin-core[2] this is really BTC/kvB (kvB is kilo-vbyte).
```
AshtonStephens, line 125: Love the footnotes in the docstring.

sbtc/src/error.rs
```diff
@@ -10,6 +10,9 @@ pub enum Error {
     /// Error when creating an RPC client to bitcoin-core
     #[error("could not create RPC client to {1}; {0}")]
     BitcoinCoreRpcClient(#[source] bitcoincore_rpc::Error, String),
+    /// Error when using a BitcoinClient trait function
```
AshtonStephens, line 13: Don't take any action on this, but just for my own curiosity why have these in the `sbtc` crate instead of the `signer` crate?

AshtonStephens (approved):
Looks good to me. The structure and the naming are sound.

I find SQL hard to assess, especially when I'm not familiar with the initial database structure. Don't consider my reviews to be correct on the SQL analysis front, but I'm very in favor of committing what we have with good unit tests and finding any hiccups in this later.

One day it may make sense to write down the way we structured this database if we haven't already.

signer/src/storage/postgres.rs
djordon: Yeah we do. The migration query that created this table is here: https://github.com/stacks-network/sbtc/blob/dcbdb12183797cfe69067a789120992ab0cb4621/signer/migrations/0003__create_tables.sql#L27-L38

The rest are in that file.

signer/src/block_observer.rs
djordon: Yeah, I figured we'll need to switch types eventually.

sbtc/src/error.rs
djordon: Good question. I figured Emily would want to use this functionality as well when validating deposit requests so I put the bitcoin client trait here. There is probably a better way to organize this all though.

xoloki (approved):
Lgtm. Small question about phantom data, but it’s not blocking anything.

sbtc/src/deposits.rs
```diff
@@ -93,19 +97,51 @@ pub struct ParsedDepositRequest {
...
+    /// the case of a reorg, it's not clear what happens if this was
+    /// confirmed.
+    pub block_hash: Option<BlockHash>,
+    /// The number of confirmations deep from that chain tip of the bitcoin
+    /// block that includes this transaction. If `None` then this is in the
+    /// mempool. TODO(384): In the case of a reorg, it's not clear what
+    /// happens if this was confirmed.
+    pub confirmations: Option<u32>,
+    /// This is to make sure that this struct cannot be created without the
+    /// above invariants being upheld.
```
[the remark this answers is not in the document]
djordon: Oh you're right it doesn't uphold them, it just makes sure that the struct cannot be created (outside of this file) without the invariants being upheld, since you cannot set a private field of a struct outside of the file that creates it.

sbtc/src/rpc.rs
djordon: Oh oops, good find. Will update.

sbtc/src/deposits.rs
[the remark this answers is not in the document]
xoloki: Ah that makes waaay more sense 😂

sbtc/src/rpc.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/385/commits/a2d8c5dec94c161d1b2448d4f82e407d272be2e5

djordon: Ooops, meant to write that in the above one.

sbtc/src/rpc.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/385/commits/a2d8c5dec94c161d1b2448d4f82e407d272be2e5.

Issue #348, opened 2024-07-18
[Feature]: Emily PUT Requests


AshtonStephens:
# Feature - Emily PUT Requests

## 1. Description

This ticket is to handle the PUT requests that update the chainstate on forks and other update calls that are more complex than other endpoints.

Relates to the LLD:
- #50 

### 1.1 Context & Purpose
Details in the LLD.

## 2. Technical Details:
Details in the LLD.

### 2.1 Acceptance Criteria:
   - [ ] A PR exists for PUT requests
   - [ ] The PR includes complicated integration tests
   - [ ] The PR is approved

## 3. Related Issues and Pull Requests (optional):
- #50 
- #289

PR #386, opened 2024-08-05
Feat: Add chain tip related queries and prepare for reorg handling.

AshtonStephens:
## Description

Handles: #399 (created retroactively)
Patially handles: #348 

This PR introduces a good deal of generic test code and adds support for the a chaintip endpoint. 

Also note that this PR adds the ticket number for a ton of `TODO(TBD)` comments, so the number of files changed is misleading.

_That said_, this _is_ a fairly large PR that touches a lot of test code and a good deal less production code. I'm hoping not to break this up unless absolutely necessary.

I'm considering this the "move quickly" time for the Emily API and I'll go back through and wipe out all the tech-debt in  not-so-many PRs because the places where I'm introducting tech debt are, in my mind, obvious and easy to reverse, but hard to get right in the initial PR (or, at least, not worth the time).

## Changes

This PR includes three features:
- Improves the integration test interfaces 
- Adds chaintip related queries with a special entry that's always at a special spot
- Adds some error states for when there's a conflicting chaintip that's not yet handled
- Bonus:
    - Replace all `TODO(TBD):` comments with `TODO([TICKET_NUMBER])` for all todos that have an obvious ticket to create for them.
    - Adds scaffolding to make error messages better for the public when there's an error if the feature "testing" isn't set.

Note the following repo structure from emily handler crate root.

- Tests
    - `tests/integration/complex/*` - Complex tests, newly created
    - `tests/integration/util/mod.rs` - Test utilities
- Source
    - `src/api/handlers/*` - endpoint handler functions
    - `src/api/routes/*` - API route to handler mappings
    - `src/database/entries/*` - Table entry representations
    - `src/database/accessors.rs` - Accessor functions for every single entry type

## Testing Information

To run the tests here we run two commands. 

2. Terminal 2: `make emily-complex-integration-test`

If you'd like to run them all in one terminal you can uncomment the `--detach` flag in the makefile command for `emily-integration-env-up`

## Checklist:


emily/handler/src/database/accessors.rs
```diff
@@ -242,6 +209,174 @@ async fn wipe_chainstate_table(context: &EmilyContext) -> Result<(), Error> {
...
+            return irrecoverable_or_okay;
+        }
+    };
+
+    // TODO(390): Determine whether the order for these operations is correct
+    // given the eventual consistency guarantees of dynamodb.
+    //
+    // TODO(TBD): Handle api status being "Reorg" during this period.
+    let mut api_state = get_api_state(context).await?;
+    let blocks_higher_than_current_tip = entry.key.height - api_state.chaintip.key.height;
```
djordon, line 249: How do we know that the `entry.key.height` is greater than the `api_state.chaintip.key.height`?

djordon: Oh, given the else clause below, I think we want to cast these heights as `i64`s before doing the subtraction.

emily/handler/tests/integration/complex/reorg.rs
```diff
@@ -0,0 +1,157 @@
...
+    // fork_id += 1;
+
+    // // Set a conflicting chainstate for a lower than top depth to initiate an internal reorg.
+    // let reorganized_chainstates = scenario.reorganized_chainstates(fork_id);
+    // let lowest_reorganized_block: Chainstate = reorganized_chainstates.first().unwrap().clone();
+    // client.update_chainstate(&lowest_reorganized_block).await;
+
+    // // Verify that the chain tip is updated to be the new reorg height.
+    // let chain_tip: Chainstate = client.get_chaintip().await;
+    // assert_eq!(chain_tip, lowest_reorganized_block);
```
djordon, line 140: Is all of this supposed to be commented out?

djordon (commented):
This largely looks good. I had a few questions but the one that I think needs addressing is around subtraction of two `u64`s.

emily/handler/src/database/accessors.rs
```diff
@@ -242,6 +209,174 @@ async fn wipe_chainstate_table(context: &EmilyContext) -> Result<(), Error> {
...
+pub async fn add_chainstate_entry(
+    context: &EmilyContext,
+    entry: &ChainstateEntry,
+) -> Result<(), Error> {
+    // Get the existing chainstate entry for height. If there's a conflict
+    // then propagate it back to the caller.
+    let current_chainstate_entry_result = get_chainstate_entry_at_height(context, entry.key.height)
+        .await
+        .and_then(|existing_entry| {
+            if existing_entry != *entry {
```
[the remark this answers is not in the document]
AshtonStephens: Why would this be wrong? In this moment the entry is a reference to a entry struct, and the `existing_entry` is a struct instance. Am I wrong in thinking that `&existing_entry != entry` would compare the pointer values to the structs and give us the wrong comparison?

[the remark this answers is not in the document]
AshtonStephens: If this is like `C`, then `&existing_entry` is the numeric pointer to the stack and `entry` is the numeric pointer to the stack from an earlier period.

emily/handler/tests/integration/complex/reorg.rs
AshtonStephens: Yep, I'm going to comment it back in since this logic isn't implemented yet.

AshtonStephens: Since this is for a future PR do you think I should delete it or keep it?

emily/handler/src/database/accessors.rs
AshtonStephens: Done.

emily/handler/src/database/accessors.rs
```diff
@@ -204,7 +165,7 @@ pub async fn get_withdrawal_entries(
 }
 
 /// Wipes all the tables.
-/// TODO(TBD): Include check for whether the table is running locally.
+/// TODO(395): Include check for whether the table is running locally.
```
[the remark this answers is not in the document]
AshtonStephens: That's a good further guard to prevent this from accidentally falling into any logic outside the `testing` feature! That said, it still leaves the path where we accidentally compile the lambda with the testing flag. The lambda should be able to do a quick check to see if it's in production.

emily/handler/src/api/handlers/chainstate.rs
```diff
@@ -104,29 +132,14 @@ pub async fn set_chainstate(
...
-        // - two chain tips are proposed at the same time for conflicting hashes
-        // - etc.
-        //
-        // `get_chain_tip_or_set_if_absent` is inappropriate in the long run and should be
-        // deleted.
-
-        accessors::add_chainstate_entry(&context, &new_chainstate_entry).await?;
-        let response: Chainstate = new_chainstate_entry.into();
+        // Convert body to the correct type.
+        let chainstate: Chainstate = body as Chainstate;
```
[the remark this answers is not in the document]
AshtonStephens: I figured this setup was the clearest to read, but I do know that nothing actually happens here. All of these are type definitions so for them to follow all the same idioms of requests and responses it's always going to be doing impactless casting.

emily/handler/src/database/accessors.rs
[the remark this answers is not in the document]
djordon: > Why would this be wrong? In this moment the entry is a reference to a entry struct, and the `existing_entry` is a struct instance. 

The reason why this looks wrong to me is because it looks like you're dereferencing the `entry`, but that shouldn't work because the type doesn't implement `Copy` (and/or `Deref`). If you try to break this out as
```rust
let intermediate = *entry;
if existing_entry != intermediate {
...
```
you'll probably get a compiler error. So I was thinking that maybe this is okay because of how `PartialEq` or `Eq` work under the hood, or maybe because the compiler does some magic to make the seemingly equivalent construction work.

> Am I wrong in thinking that `&existing_entry != entry` would compare the pointer values to the structs and give us the wrong comparison?

Yeah, in Rust references are not like pointers, so the reference comparison will actually do what we want.

emily/handler/tests/integration/complex/reorg.rs
djordon: Keep it, I was just wondering if it was an oversight or something.

emily/handler/src/api/handlers/chainstate.rs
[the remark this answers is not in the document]
djordon: Ahh yeah I can see how it's easier to read. I was a little confused at first since I wouldn't think you can do a cast like this without some more fanfare (I thought you needed to use `transmute` or whatever).

I think you should do `Chainstate::from(body)` here. It's cleaner, it's also a no-op, and it paves the way for when/if we differentiate the `SetChainstateRequestBody` type from the `Chainstate` type.

emily/handler/src/database/accessors.rs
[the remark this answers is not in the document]
djordon: > That's a good further guard to prevent this from accidentally falling into any logic outside the `testing` feature! That said, it still leaves the path where we accidentally compile the lambda with the testing flag. The lambda should be able to do a quick check to see if it's in production.

Oh yeah good point.

djordon (approved):
Looks good

emily/handler/src/database/accessors.rs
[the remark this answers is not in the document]
AshtonStephens: Got it, fixed!

emily/handler/src/api/handlers/chainstate.rs
```diff
@@ -104,29 +132,14 @@ pub async fn set_chainstate(
...
-        // - two chain tips are proposed at the same time for conflicting hashes
-        // - etc.
-        //
-        // `get_chain_tip_or_set_if_absent` is inappropriate in the long run and should be
-        // deleted.
-
-        accessors::add_chainstate_entry(&context, &new_chainstate_entry).await?;
-        let response: Chainstate = new_chainstate_entry.into();
+        // Convert body to the correct type.
+        let chainstate: Chainstate = Chainstate::from(body);
```
djordon, line 136: Now `clippy` is complaining about a useless conversion 😞. Okay, well just do what you want here.

djordon (approved):

emily/handler/src/api/handlers/chainstate.rs
AshtonStephens: Reverted to old version :(

Issue #392, opened 2024-08-05
[Tech Debt]: Make Emily API integration tests self-containing


AshtonStephens:
## Description

![low-priority]

PR that introduced tech-debt: #305

The Emily API integration tests rely on a specific execution order of separate tests to complete successfully. That's bad, we should fix that.

Included within this may be ensuring that the the `serial` dev-dependency is no longer being used.

[low-priority]: https://img.shields.io/badge/priority-low-low.svg?style=for-the-badge&labelColor=gray
[medium-priority]: https://img.shields.io/badge/priority-medium-yellow.svg?style=for-the-badge&labelColor=gray
[high-priority]: https://img.shields.io/badge/priority-high-orange.svg?style=for-the-badge&labelColor=gray

AshtonStephens:
Doing this now to make local testing faster. It's a ticket with good ROI.

PR #400, opened 2024-08-05
feat: Make emily integration tests self containing.

AshtonStephens:
## Description

Closes: #392 

Makes the emily integration tests self containing by wiping the database before each integration test.

The tests still use the underlying client to the `TestClient` instead of the methods in the `TestClient` for two reasons:
1. To make the tests more transparent when they fail for these unit - like tests.
2. To reduce the number of changes in this single PR.

## Changes

Adds setup function to each of the endpoint tests.

Note I opted to have a somewhat-useless `client.teardown().await;` call to every integration test to further make sure that we aren't leaving anything behind, and to make the `teardown` function available to more features going forward if there becomes more to teardown.

I'm all for redundancy if it makes what our tests do more obvious.

## Testing Information

Run tests with the following:

2. Terminal 2: `make emily-integration-test`

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1610,7 +1610,6 @@ dependencies = [
  "serde 1.0.203",
  "serde_dynamo",
  "serde_json",
- "serial_test",
  "thiserror",
  "time 0.3.36",
  "tokio",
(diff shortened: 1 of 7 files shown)


djordon (approved):
Looks good. I like this, we may move to a more explicit set up and tear down approach for the signer's integration tests.

Issue #383, opened 2024-08-02
[Feature]: Use stacks-core or bitcoin-core for bitcoin block notifier


djordon:
# Feature - Use stacks-core or bitcoin-core for block notifier

## 1. Description

We currently have a bitcoin block notifier that relies on electrum. Our electrum client use appears to poll electrum for new bitcoin block headers (rather than, say, registering an `epoll` callback or use some condition variable or something like that). What's currently there is not terrible, but not the design that we were going for. Instead of adjusting our electrum block subscription code, we can just use a stacks node (or bitcoin-core) as the source for bitcoin blocks. This approach would also eliminate the need to use electrum as a production dependency for the signers. 

### 2.1 Acceptance Criteria:
   - [ ] Setup the sBTC signer as a server that can receive `POST /new_burn_block` events from stacks-core, or as a zeromq client for bitcoin-core events.

## 3. Related Issues and Pull Requests (optional):

The original issue for this was https://github.com/stacks-network/sbtc/issues/177 which was closed in https://github.com/stacks-network/sbtc/pull/187.

Issue #405, opened 2024-08-06
[Feature]: Test logic during a bitcoin reorg


djordon:
# Feature - Test logic during a bitcoin reorg

## 1. Description

This is a big task, but there are many scenarios that we need to check during a bitcoin reorg, this ticket is to track them all.

### 1.1 Context & Purpose

Bitcoin reorgs require a lot of logic that needs to be added and tested. Having an overview of them all here will be helpful to track what has been done and what needs to be done.

### 2.1 Acceptance Criteria:
Add more as they come up.
   - [ ] The block observer in the signer detects reorgs and fetches missing blocks.
   - [ ] The signers resubmit rotate key transactions.
   - [ ] The signers resubmit withdrawal accept transactions
   - [ ] The signers resubmit withdrawal decline transactions
   - [ ] The signers resubmit deposit accept transactions

djordon:
Yeah, let's go with the new one.

PR #406, opened 2024-08-06
feat: use bitcoin-core as the block notifier

djordon:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/383.

The electrum client that we have for rust requires you to poll the server for bitcoin blocks. While not terrible, using the ZeroMQ interface on bitcoin-core gives us real-time notifications, and allows us to remove electrum server as a requirement for running the signer application.

## Changes

* Add structs subscribe to bitcoin-cores ZeroMQ notification interface for block and block hash notifications.
* Add an integration test that listens to bitcoin-core messages and checks the new code.

## Testing Information

The two integration tests check the new logic that was added here. It does not validate the messages received and does not reproduce the logic laid out in the bitcoin docs for what happens during a reorg. That we should do later in https://github.com/stacks-network/sbtc/issues/405.

## Checklist:

### DIFF
--- a/signer/tests/integration/zmq.rs
+++ b/signer/tests/integration/zmq.rs
@@ -1,0 +1,59 @@
+use bitcoin::Block;
+use futures::StreamExt;
+use sbtc::testing::regtest;
+use signer::zmq::BitcoinCoreMessageStream;
+
+const BITCOIN_CORE_ZMQ_ENDPOINT: &str = "tcp://localhost:28332";
+
+/// This tests that out bitcoin block stream receives new blocks from
+/// bitcoin-core as it receives them. We create the stream, generate
+/// bitcoin blocks, and wait for the blocks to be received from the stream.
+#[tokio::test]
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
+async fn helper_struct_methods_work() {
+    sbtc::logging::setup_logging("info,signer=debug", false);
+    let (_, faucet) = regtest::initialize_blockchain();
+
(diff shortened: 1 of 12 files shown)


signer/tests/integration/zmq.rs
```diff
@@ -0,0 +1,113 @@
...
+            if sx.is_closed() {
+                break;
+            }
+
+            tracing::info!("Sending block {:?}", block.block_hash());
+            sx.send(block).await.unwrap();
+        }
+    });
+
+    // When the faucet generates a blocks it returns the block hash of the
```
xoloki, line 43: generates a block

signer/tests/integration/zmq.rs
```diff
@@ -0,0 +1,113 @@
...
+            if sx.is_closed() {
+                break;
+            }
+
+            tracing::info!("Sending block hash {block_hash}");
+            sx.send(block_hash).await.unwrap();
+        }
+    });
+
+    // When the faucet generates a blocks it returns the block hash of the
```
xoloki, line 97: generates a block

xoloki (approved):
Lgtm, just a few comment corrections. I really like the zmq interface.

signer/tests/integration/zmq.rs
djordon: https://github.com/stacks-network/sbtc/pull/406/commits/4cebe35e4f5d8d3edd90818224ba280402d4c374

signer/tests/integration/zmq.rs
djordon: https://github.com/stacks-network/sbtc/pull/406/commits/4cebe35e4f5d8d3edd90818224ba280402d4c374

Issue #204, opened 2024-05-29
[Feature]: Block observer - extract sBTC transactions


netrome:
### Context
The block observer should as part of its operation extract any sBTC transactions from the bitcoin blocks it's processing and write the extracted requests to the signer database.

### Definition of done
The `extract_signer_transactions` method in the block observer is implemented and tested.

Issue #353, opened 2024-07-22
[Feature]: Accept donations into the signers' wallet


djordon:
# Feature - Accept donations into the signers' wallet

## 1. Description

We want to be able to accept donations into the signers' wallet as described in https://github.com/stacks-network/sbtc/issues/19.

## 2. Technical Details:

I think for this we need to either monitor bitcoin blocks for transactions to the signers' addresses or we need electrum server to do this.

### 2.1 Acceptance Criteria:
   - [ ] If someone sends funds to the signers address, those funds are eventually swept up during normal operations.

PR #410, opened 2024-08-08
feat: extract sBTC transactions in block observer

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/204.

We can probably remove the `tweaked_aggregate_key` column from the `dkg_shares` table now.

## Changes

* Add a `script_pubkey` column to the `dkg_shares` table. This makes it simple for us to find BTC transactions that the signers have sent. The alternative is to assume that witness program version is always 1, and derive the `scriptPubKey` from the tweaked public key. It also makes it easy for us to find donations for https://github.com/stacks-network/sbtc/issues/353.
* Add a trait for converting `p256k1` types and `rust-secp256k1` types to the signers' `scriptPubKey`.
* Implement the `BlockObserver::extract_sbtc_transactions` function.
* Add function for writing bitcoin transactions into the database.

## Testing Information

There are unit tests for the new block observer functionality and integration tests for the query.

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1368,7 +1368,7 @@ dependencies = [
  "hashbrown 0.14.5",
  "lock_api",
  "once_cell",
- "parking_lot_core 0.9.10",
+ "parking_lot_core",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


AshtonStephens (approved):
I did read it all, LGTM!

Issue #416, opened 2024-08-12
[Feature]: Simplify cryptographic key type usage


djordon:
# Feature - Simplify cryptographic key type usage

## 1. Description

There are two problems:
1. We use at least three different secp256k1 public key types in the signer crate (one for `rust-sepc256k1`, `p256k1`, and `stacks-common`). We should simplify all of the conversion code to make things simpler.
2. We incorrectly use the `p256k1::scalar::Scalar` type for the private key. This is probably not going to be a problem, but it's best to fix it anyway.

We should simplify our usage of public and private keys and provide tested conversion functions whenever we need to use a specific type.

### 2.1 Acceptance Criteria:
   - [ ] We have public and private key types with conversion implementations to the other types.

PR #418, opened 2024-08-12
feat: simplify cryptographic key types

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/416.

Add cryptographic types for internal use.

## Changes

* Add new cryptographic types for internal use in the signer crate.
* Implement conversion traits to and from `p256k1`, `stacks-common`, and `rust-secp256k1` types.

## Testing Information

This PR adds unit tests that check the assertions made in this crate (although these tests cannot be exhaustive, it should be fine).

## Checklist:


signer/src/keys.rs
```diff
@@ -0,0 +1,369 @@
...
+        let x_data = value.x().to_bytes();
+
+        let pk = secp256k1::XOnlyPublicKey::from_slice(&x_data).map_err(Error::InvalidPublicKey)?;
+        let parity = if value.has_even_y() {
+            Parity::Even
+        } else {
+            Parity::Odd
+        };
+
+        let public_key = secp256k1::PublicKey::from_x_only_public_key(pk, parity);
```
AshtonStephens, line 115: Remind me, why is the parity necessary?

AshtonStephens (approved):
LGTM

signer/src/keys.rs
xoloki: It’s necessary for the 32 byte representation of an XOnlyPublicKey, because there are two points on the curve with the same x value.

xoloki (approved):
Lgtm

Issue #417, opened 2024-08-12
[Feature]: Use simplified crypto types in signer crate


djordon:
# Feature - Use simplified crypto types in signer crate

## 1. Description

This follows from https://github.com/stacks-network/sbtc/issues/416 and has roughly the same description. This ticket is for using the new crypto types.

### 2.1 Acceptance Criteria:
   - [ ] We use these new types instead of the patch-work types we currently use.

## 3. Related Issues and Pull Requests (optional):

PR #420, opened 2024-08-13
refactor: use simplified crypto types

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/417.

## Changes

* Use the new `PublicKey` and `PrivateKey` types throughout.
* The tweaked public key is generated from the same source throughout.
* Add implementations for sending and receiving these types to and from the database.
* Fixed a bunch of typos in the comments.

## Testing Information

If the existing tests pass then these changes should be fine.

## Checklist:

### DIFF
--- a/signer/src/keys.rs
+++ b/signer/src/keys.rs
@@ -202,11 +202,23 @@ impl<'r> sqlx::Decode<'r, sqlx::Postgres> for PublicKey {
     }
 }
 
+impl<'r> sqlx::Type<sqlx::Postgres> for PublicKey {
+    fn type_info() -> sqlx::postgres::PgTypeInfo {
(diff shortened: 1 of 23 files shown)


signer/src/keys.rs
```diff
@@ -163,6 +196,29 @@ impl PublicKey {
     pub fn serialize(&self) -> [u8; 33] {
         self.0.serialize()
     }
+
+    /// Creates a new public key from a [`Private`] and the global
+    /// [`SECP256K1`] context.
+    pub fn from_private_key(key: &PrivateKey) -> Self {
```
xoloki, line 202: Why is this a separate function and not an `impl From`?

xoloki (approved):
One small question but looks great 👍

Issue #255, opened 2024-06-25
[Feature]: Transaction signer signing stacks transactions


netrome:
When the transaction signer receives a `StacksTransactionSignRequest` message, it should validate the request and if valid return an appropriate signature for the request.

### Definition of done
The transaction signer can respond to `StacksTransactionSignRequest` messages with a valid signature.

djordon:
Signing a stacks transaction requires each signer to validate the stacks transaction, and that is blocked on https://github.com/stacks-network/sbtc/issues/426.

PR #422, opened 2024-08-14
feat: add signature module

djordon:
## Description

This is part of the work for https://github.com/stacks-network/sbtc/issues/255, but does not close it.

Stacks transactions use recoverable ECDSA signatures, so this helps with the transition from standard ECDSA signatures to the recoverable variants.

## Changes

* Add a signatures module with a trait to help with common operations for the `secp256k1` recoverable signature type.
*  Implement `Serialize` or `Deserialize` for `secp256k1` recoverable signature type.

## Testing Information

This PR adds some tests that check that the signature producing functions work as intended. We also have existing tests that pass, so this should be fine without the new tests.

## Checklist:


AshtonStephens (approved):
Looks clean! LGTM!

jferrant (approved):

Issue #360, opened 2024-07-24
[Feature]: Use sbtc lib within Emily API.


AshtonStephens:
# Feature - Use sbtc lib within Emily API.

## 1. Description
Use the sBTC lib within the Emily API; it was removed because of the `sha2-asm` crate used in stacks core.

### 2.1 Acceptance Criteria:
   - [ ] The sbtc common lib is used in the Emily API

## 3. Related Issues and Pull Requests (optional):
- https://github.com/stacks-network/sbtc/pull/359

djordon:
This was done in https://github.com/stacks-sbtc/sbtc/pull/635.

PR #423, opened 2024-08-14
Emily as Local Warp Service

AshtonStephens:
## Description

Closes: #N/A

I added this change because iterating on the update command was taking too long with the 30 seconds of recompile and run time for emily running in a docker container with the apigateway wrapper. This change adds the ability to run emily as a warp service locally.

Note that this is required for https://github.com/stacks-network/sbtc/issues/360 because running the apigateway container locally won't be feasible for `arm64` chips once the `sbtc` library is imported. This isn't going to be an issue in production because it will only use `intel x86` processors for which this library is fine.

## Changes

- Changes the Makefile to have slightly better commands
- Changes the emily docker compose to not always include the apigateway container
- Creates two binary files for the emily handler - one for prod and one for local warp service
- Changes the integration tests to expect to point to the new warp service
- Adds new emily context option to automatically detect locally running tables. 
- Adds the beginnings of a test for just the table accessor

## Testing Information

Tested with the integration tests. Unfortunately this now requires 3 terminals, but this won't last long.

2. Terminal 2: `make emily-server`
3. Terminal 3: `make emily-integration-test`

Variations that require a single or two terminals created odd error scenarios for the database that were not remotely obvious to debug; this inconvenience saves a lot of debug time.

## Checklist:

### DIFF
--- a/emily/handler/Cargo.toml
+++ b/emily/handler/Cargo.toml
@@ -8,7 +8,6 @@ name = "emily-lambda"
 
 [[bin]]
 name = "emily-server"
-features = ["testing"]
 
 [features]
 default = ["testing"]
--- a/emily/handler/Cargo.toml
+++ b/emily/handler/Cargo.toml
@@ -4,7 +4,11 @@ version = "0.1.0"
 edition = "2021"
 
 [[bin]]
-name = "emily-handler"
+name = "emily-lambda"
+
+[[bin]]
+name = "emily-server"
+features = ["testing"]
 
 [features]
 default = ["testing"]
--- a/emily/handler/src/bin/emily-lambda.rs
+++ b/emily/handler/src/bin/emily-lambda.rs
@@ -1,0 +1,37 @@
+//! Emily API entrypoint.
+
+use emily_handler::context::EmilyContext;
+use tracing::info;
+use warp::Filter;
+
+use emily_handler::api;
+use emily_handler::logging;
+
+#[tokio::main]
+async fn main() {
+    crate::logging::setup_logging("info,emily-handler=debug", false);
+
+    // TODO(389 + 358): Handle config pickup in a way that will only fail for the relevant call.
+    let emily_context: EmilyContext = EmilyContext::from_env()
+        .await
+        .unwrap_or_else(|e| panic!("{e}"));
+
+    // Print configuration.
+    info!("Emily context setup for Emily Lambda.");
+    let emily_context_string =
+        serde_json::to_string_pretty(&emily_context).expect("Context must be serializable.");
+    info!(emily_context_string);
+
+    // Make routes.
+    let routes = api::routes::routes(emily_context)
+        .recover(api::handlers::handle_rejection)
+        .with(warp::log("api"));
+
+    // Create warp service.
+    let warp_service = warp::service(routes);
+
+    // TODO(276): Remove warp_lambda in Emily API and use different library.
+    warp_lambda::run(warp_service)
+        .await
+        .expect("An error occurred");
+}
--- a/emily/handler/src/bin/emily-lambda.rs
+++ b/emily/handler/src/bin/emily-lambda.rs
@@ -9,7 +9,7 @@ use emily_handler::logging;
 
 #[tokio::main]
 async fn main() {
-    crate::logging::setup_logging("info,emily-handler=debug", false);
+    logging::setup_logging("info,emily-handler=debug", false);
 
     // TODO(389 + 358): Handle config pickup in a way that will only fail for the relevant call.
     let emily_context: EmilyContext = EmilyContext::from_env()
--- a/emily/handler/src/bin/emily-server.rs
+++ b/emily/handler/src/bin/emily-server.rs
@@ -1,10 +1,10 @@
 //! Emily Warp Service Binary.
 
 use emily_handler::context::EmilyContext;
+use std::net::ToSocketAddrs;
 use tracing::error;
 use tracing::info;
 use warp::Filter;
-use std::net::ToSocketAddrs;
 
 use emily_handler::api;
 use emily_handler::logging;
--- a/emily/handler/src/bin/emily-server.rs
+++ b/emily/handler/src/bin/emily-server.rs
@@ -1,8 +1,6 @@
 //! Emily Warp Service Binary.
 
 use emily_handler::context::EmilyContext;
-use std::net::ToSocketAddrs;
-use tracing::error;
 use tracing::info;
 use warp::Filter;
 
@@ -35,13 +33,7 @@ async fn main() {
     let addr_str = format!("{}:{}", host, port);
 
     info!("Server will run locally on {}", addr_str);
-    let addr = match addr_str.to_socket_addrs() {
-        Ok(mut addrs) => addrs.next().expect("No addresses found"),
-        Err(e) => {
-            error!("Failed to resolve address: {}", e);
-            return;
-        }
-    };
+    let addr: std::net::SocketAddr = addr_str.parse().expect("Failed to parse address");
 
     warp::serve(routes).run(addr).await;
 }
--- a/emily/handler/src/bin/emily-server.rs
+++ b/emily/handler/src/bin/emily-server.rs
@@ -1,0 +1,47 @@
+//! Emily Warp Service Binary.
+
+use emily_handler::context::EmilyContext;
+use tracing::error;
+use tracing::info;
+use warp::Filter;
+use std::net::ToSocketAddrs;
+
+use emily_handler::api;
+use emily_handler::logging;
+
+#[tokio::main]
+async fn main() {
+    logging::setup_logging("info,emily-handler=debug", false);
+
+    // TODO(389 + 358): Handle config pickup in a way that will only fail for the relevant call.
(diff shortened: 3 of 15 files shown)


emily/handler/Cargo.toml
```diff
@@ -4,7 +4,11 @@ version = "0.1.0"
 edition = "2021"
 
 [[bin]]
-name = "emily-handler"
+name = "emily-lambda"
+
+[[bin]]
+name = "emily-server"
+features = ["testing"]
```
djordon, line 11: Cargo complained that the `features` key is unused.

emily/handler/src/bin/emily-lambda.rs
```diff
@@ -0,0 +1,37 @@
+//! Emily API entrypoint.
+
+use emily_handler::context::EmilyContext;
+use tracing::info;
+use warp::Filter;
+
+use emily_handler::api;
+use emily_handler::logging;
```
djordon, line 8: I think this is unused. Weird, I cannot get clippy or vs code to point it out.

emily/handler/tests/integration/accessors.rs
```diff
@@ -0,0 +1,57 @@
+//! Integration tests for the database accessors
+
+use aws_sdk_dynamodb::operation::update_item::builders::UpdateItemFluentBuilder;
```
djordon, line 3: There are a few things in this file that appear to be unused.

emily/handler/tests/integration/accessors.rs
```diff
@@ -0,0 +1,57 @@
...
+use emily_handler::{
+    api::models::deposit::requests::DepositUpdate,
```
djordon, line 5: Unused

emily/handler/tests/integration/accessors.rs
```diff
@@ -0,0 +1,57 @@
...
+        client
+            .create_chainstate(&util::test_chainstate(height, fork_id))
+            .await;
+    }
+    // Make a new deposit.
+    let deposit = client
+        .create_deposit(&util::test_create_deposit_request(5, 0))
+        .await;
+    // Get the corresponding deposit entry.
+    let deposit_entry = accessors::get_deposit_entry(
```
djordon, line 48: Unused, maybe just prefix with an underscore.

emily/handler/src/context.rs
```diff
@@ -73,4 +73,66 @@ impl EmilyContext {
...
+
+        // Get config that always points to the dynamodb table directly
+        // from outside of a docker compose setup.
+        let dynamodb_client = Client::new(
+            &aws_config::load_defaults(BehaviorVersion::latest())
+                .await
+                .into_builder()
+                .endpoint_url("http://localhost:8000")
+                .build(),
+        );
```
djordon, line 89: nit: can we break this out like so? What's here isn't so bad, but I find breakout approach a bit easier to read
```rust
let sdk_config = aws_config::load_defaults(BehaviorVersion::latest())
    .await
    .into_builder()
    .endpoint_url("http://localhost:8000")
    .build();
let dynamodb_client = Client::new(&sdk_config);
```

emily/handler/src/context.rs
```diff
@@ -73,4 +73,66 @@ impl EmilyContext {
...
+            .list_tables()
+            // Get at most 20 table names - there should be 3...
+            .limit(20)
+            .send()
+            .await
+            .map_err(|err| {
+                Error::Debug(format!(
+                    "Failed setting up settings from table names - {err:?}"
+                ))
+            })?
```
djordon, line 102: I believe what's here is equivalent to `.expect("Failed setting up settings from table names")`

emily/handler/tests/integration/accessors.rs
```diff
@@ -0,0 +1,57 @@
...
+        &context,
+        &DepositEntryKey {
+            bitcoin_txid: deposit.bitcoin_txid.clone(),
+            bitcoin_tx_output_index: deposit.bitcoin_tx_output_index,
+        },
+    )
+    .await
+    .expect("Get deposit entry for newly created deposit should work.");
```
djordon, line 56: Is there anything to check here? As in shouldn't we check that the returned `deposit_entry` matches the created deposit?

djordon (commented):
This is basically good, but we should remove the unused stuff in the new integration test file.

emily/handler/src/bin/emily-server.rs
```diff
@@ -0,0 +1,47 @@
...
+    let addr_str = format!("{}:{}", host, port);
+
+    info!("Server will run locally on {}", addr_str);
+    let addr = match addr_str.to_socket_addrs() {
+        Ok(mut addrs) => addrs.next().expect("No addresses found"),
+        Err(e) => {
+            error!("Failed to resolve address: {}", e);
+            return;
+        }
+    };
```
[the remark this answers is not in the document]
AshtonStephens: I just copy pasted this from the blocklist client. I'll change this there too.

emily/handler/tests/integration/accessors.rs
AshtonStephens: It's my continuous transgression of leaving in things I intended to use in the next PR, but you're right its definitely better that I delete them.

emily/handler/tests/integration/accessors.rs
AshtonStephens: Done.

emily/handler/tests/integration/accessors.rs
AshtonStephens: Done.

emily/handler/src/context.rs
AshtonStephens: Do you have any specific "rule" for when to break variables out? Could help me calibrate. My personal rule has typically been "if it can be done without semicolons, it should be done without semicolons" but I agree that it's leading to unclear code so I should change that.

emily/handler/tests/integration/accessors.rs
AshtonStephens: Adding. This is part of a very unfinished PR - it was going to be part of a really long PR but then I figured it's much better to break it out - so that's why this test is really unfinished and not related to the actual code. I shouldn't have done that.

emily/handler/src/bin/emily-server.rs
```diff
@@ -0,0 +1,47 @@
...
+        serde_json::to_string_pretty(&emily_context).expect("Context must be serializable.");
+    info!(emily_context_string);
+
+    let routes = api::routes::routes(emily_context)
+        .recover(api::handlers::handle_rejection)
+        .with(warp::log("api"));
+
+    // Create warp service as a local service.
+    // TODO(TBD): Make these fields configurable.
+    let host: &str = "127.0.0.1";
```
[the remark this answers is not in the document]
AshtonStephens: This is for it to run locally, so I think it should be 127.0.0.1?

emily/handler/Cargo.toml
AshtonStephens: Removed.

emily/handler/src/bin/emily-lambda.rs
AshtonStephens: Changed logging import to use this.

emily/handler/src/context.rs
djordon: Yeah, my rough "rule" is something like "assign 'computation outputs' to a new variable unless 'the whole thing' can fit in one line". So I might do something like this
```rust
// This is simple enough
let dynamodb_client = Client::new(&make_my_config());
```
But it's a rough rule, since I probably wouldn't do the following even if it fits on one line:
```rust
// There is too much going on here. Also, clippy might change
// this anyway, and if it does then I'd break it out
let dynamodb_client = Client::new(&ConfigBuilder::new().endpoint(url).build());
```
I don't expect anyone to follow my "rule" but when I think that other people will find it much easier to read then I'll nit on it.

emily/handler/src/bin/emily-server.rs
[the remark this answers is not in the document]
djordon: Oh yeah, I think you might be right. Is this supposed to be used at all in our devnet? If so then we probably want to use `0.0.0.0` but I forget how we intend for local devnet stuff is supposed to work.

emily/handler/tests/integration/accessors.rs
djordon: 🙏🏽 I very much appreciate you breaking this out. And I'm okay leaving it as is since you planned on changing it soon anyway.

djordon (approved):
Looks good! The only question I have is around devnet usage of the emily-server binary. I don't know what we were planning there.

AshtonStephens:
> Looks good! The only question I have is around devnet usage of the emily-server binary. I don't know what we were planning there.

I don't know what you mean around this.

djordon:
> > Looks good! The only question I have is around devnet usage of the emily-server binary. I don't know what we were planning there.
> 
> I don't know what you mean around this.

If we intend to use the `emily-server` binary in devnet then I think we need to bind the application `0.0.0.0` (instead of the loopback address `127.0.0.1`) in order for other containers to talk to it. But I don't know if the plan is to use the `emily-server` binary on devnet at all so it might be fine to use the loopback address. I could also be wrong about needing to use `0.0.0.0` in docker, but I thought that was the case.

Issue #424, opened 2024-08-14
[Feature]: Modify the signature threshold in rotate-key contract calls


djordon:
# Feature - Modify the signature threshold in rotate-key contract calls

## 1. Description

Right now the signature threshold is a constant in the `sbtc-bootstrap-signers.clar` contract. It's used to specify the `m` in the signers' `m-of-n` multi-sig wallet on the Stacks blockchain. Although, this value can be fixed for now I think we'll want to allow it to change for later releases of the sBTC signer binary.

### 1.1 Context & Purpose

On the bitcoin side, the threshold for the multi-sig wallet is fixed, both in the bootstrap phase and in later phases. On the stacks side we don't have Schnorr signatures so we note the threshold in smart contract until that happens.

## 2. Technical Details:

On the signers side, this is straightforward, we need to:
1. Add a column in the database for the threshold.
2. Set the value in contract call.
3. Add fields for writing and reading the value to and from the database.

On the clarity side I think we need to accommodate the threshold in the rotate-keys public function and store the variable in the `sbtc-registry`.

### 2.1 Acceptance Criteria:
   - [ ] The signers can save the threshold in the database
   - [ ] The signers send the value in the contract call

## 3. Related Issues and Pull Requests (optional):

[Smart contract design doc](https://docs.google.com/document/d/1Bwi1jM_7p9OqNGT41eWCuoKnfoOrNHycqxby3XSonCA/edit) and opening PR https://github.com/stacks-network/sbtc/pull/151.

PR #425, opened 2024-08-14
feat: allow for modifying the signer threshold in rotate key contract call

djordon:
## Description

Addresses part of https://github.com/stacks-network/sbtc/issues/424.

## Changes

* Add a new column to the `rotate_keys_transactions` table for the signature threshold.
* Update the signers' internal struct for representing `rotate-keys-wrapper` contract call to accommodate the signature threshold.

## Testing Information

If the tests pass then we should be in the clear.

## Checklist:


AshtonStephens (approved):
LGTM, simple change to add a field.

Issue #273, opened 2024-07-02
[Feature]: Remove `local` prefix from Emily API path


AshtonStephens:
# Feature - Remove `local` prefix from Emily API path

## 1. Description
While working on a recent PR I found that all local testing with the lambda using `warp` had the path `local` appended to the front. I can't figure out why, this ticket is to figure it out and prevent it.

### 2.1 Acceptance Criteria:
   - [ ] Figure out what's wrong
   - [ ] Fix it
   - [ ] Release and merge PR.

AshtonStephens:
Done.

PR #435, opened 2024-08-17
feat: Emily server cli args + remove "/local" path prefix

AshtonStephens:
## Description

Closes: #273

This ticket closes the local prefix issue by only adding the prefix on the lambda binary configuration with the `testing` feature. The `sam` cli that's used to run the lambda locally forcibly includes the `/local` prefix to prevent any accidental running of the `sam` development environment in the cloud, so the only way to get around this is to make adding the path conditional.

## Changes

The `route` commands needed to change to return tuples of replies because the output of the filters needs to be a tuple in order to continue composing the filter with `and` warp filters - this is because it needs to be possible to convert the output into function arguments for the next filter to receive.

It's a bit wonky, but necessary.

## Testing Information

Tested with integration tests and also on local make commands.

## Checklist:

### DIFF
--- a/emily/handler/src/api/routes/mod.rs
+++ b/emily/handler/src/api/routes/mod.rs
@@ -21,29 +21,25 @@ mod withdrawal;
 #[cfg(feature = "testing")]
 pub fn routes(
     context: EmilyContext,
-) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
-    // TODO(273):  Remove the "local" prefix once we figure out why all local
-    // testing calls seem to forcibly start with `local`.
-    warp::path("local").and(
-        health::routes()
-            .or(chainstate::routes(context.clone()))
-            .or(deposit::routes(context.clone()))
-            .or(withdrawal::routes(context.clone()))
-            .or(testing::routes(context)),
-    )
+) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
+    health::routes()
+        .or(chainstate::routes(context.clone()))
+        .or(deposit::routes(context.clone()))
+        .or(withdrawal::routes(context.clone()))
+        .or(testing::routes(context))
+        // Convert reply to tuple to that more routes can be added to the returned filter.
+        .map(|reply| (reply,))
 }
 
 /// This function sets the Warp filters for handling all requests.
 #[cfg(not(feature = "testing"))]
 pub fn routes(
     context: EmilyContext,
-) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
-    // TODO(273):  Remove the "local" prefix once we figure out why all local
-    // testing calls seem to forcibly start with `local`.
-    warp::path("local").and(
-        health::routes()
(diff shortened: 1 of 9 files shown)


emily/handler/src/api/routes/mod.rs
```diff
@@ -21,29 +21,25 @@ mod withdrawal;
...
-            .or(testing::routes(context)),
-    )
+) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
+    health::routes()
+        .or(chainstate::routes(context.clone()))
+        .or(deposit::routes(context.clone()))
+        .or(withdrawal::routes(context.clone()))
+        .or(testing::routes(context))
+        // Convert reply to tuple to that more routes can be added to the returned filter.
+        .map(|reply| (reply,))
```
djordon, line 31: Not sure that I understand the tuple change. Curious, what is going on here?

emily/handler/tests/integration/util/constants.rs
```diff
@@ -2,13 +2,11 @@
...
-pub const EMILY_ENDPOINT: &'static str = "http://localhost:3031/local";
-pub const EMILY_WITHDRAWAL_ENDPOINT: &'static str = "http://localhost:3031/local/withdrawal";
-pub const EMILY_DEPOSIT_ENDPOINT: &'static str = "http://localhost:3031/local/deposit";
-pub const EMILY_CHAINSTATE_ENDPOINT: &'static str = "http://localhost:3031/local/chainstate";
-pub const EMILY_TESTING_ENDPOINT: &'static str = "http://localhost:3031/local/testing";
+pub const EMILY_ENDPOINT: &'static str = "http://localhost:3031";
+pub const EMILY_WITHDRAWAL_ENDPOINT: &'static str = "http://localhost:3031/withdrawal";
+pub const EMILY_DEPOSIT_ENDPOINT: &'static str = "http://localhost:3031/deposit";
+pub const EMILY_CHAINSTATE_ENDPOINT: &'static str = "http://localhost:3031/chainstate";
+pub const EMILY_TESTING_ENDPOINT: &'static str = "http://localhost:3031/testing";
```
djordon, line 9: I think I understand. We need the local prefix for local lambda with SAM, but we can don't do integration tests with SAM, just regular warp, so can ditch the local prefix. Is that right?

djordon (approved):
Looks good.

emily/handler/tests/integration/util/constants.rs
AshtonStephens: Yes. Sam forces us to have it.

emily/handler/src/api/routes/mod.rs
AshtonStephens: Warp filters all expect to take the output of the previous filter as the input to the next one, but it's implemented where each filter actually takes in the previous arguments as a tuple where the individual elements of the tuple are the arguments to the function.

This logic doesn't work if the output isn't a tuple because tuples can be turned into itemized function variables, but a single output cannot. Replying with a reply as a single element tuple allows it to be provided as a function argument :)

Issue #304, opened 2024-07-06
[Feature]: Automatically set the `created_at` columns


djordon:
# Feature - Automatically set the `created_at` columns

### 1.1 Context & Purpose

It seems like we don't need to return or set the `created_at` columns. Change our insert queries to automatically derive them on by the database. This is really just a simplification.

### 2.1 Acceptance Criteria:
   - [ ] No more inserting or retrieving timestamps unless they are necessary.

PR #436, opened 2024-08-20
refactor: automatically set created_at columns

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/304.

This isn't really necessary. I just decided to working on something that requires less mental energy while in transit and this fit the bill.

## Changes

* Allow `created_at` fields in all tables to be set automatically on insert with defaults
* Don't set the `created_at` fields when inserting data in the database.
* Remove the `created_at` fields from the model types.
* Don't update the `query!` and `query_as!` macro usage. I don't think this is worth the hassle.

## Testing Information

I had to update one of the parameters in one of the tests since the test now fails due to randomness. Signers randomly reject requests and if enough of them reject the request than the test will fail. So I made sure that there were enough signers so that it was very unlikely by chance that the test will fail. Maybe I should make it even more unlikely.

## Checklist:

### DIFF
--- a/signer/migrations/0003__create_tables.sql
+++ b/signer/migrations/0003__create_tables.sql
@@ -13,14 +13,14 @@ CREATE TABLE sbtc_signer.bitcoin_blocks (
     block_height BIGINT NOT NULL,
     parent_hash BYTEA NOT NULL,
     confirms BYTEA[] NOT NULL,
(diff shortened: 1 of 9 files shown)


AshtonStephens (approved):
Lots of SQL - When I'm done with the API I'll join the signer work and I'll go through and vet all this, but for the time being I'm assuming it's correct.

Approved!

PR #440, opened 2024-08-20
Sort key ordered dynamodb access (reorg part 3/4)

AshtonStephens:
## Description

Closes: #N/A

Part of the reorg handling PR but broken up.

## Changes

- Adds sort key ordered queries. 
- Adds specific functions for querying by last modified height for deposit and withdrawal entries.
- Additionally makes the output of the operation by status queries in sorted order by last modified height.

## Testing Information

Integration and unit tests.

## Checklist:

### DIFF
--- a/emily/handler/src/database/accessors.rs
+++ b/emily/handler/src/database/accessors.rs
@@ -2,7 +2,7 @@
 
 use aws_sdk_dynamodb::types::AttributeValue;
 use serde_dynamo::Item;
-use tracing::info;
+use tracing::{info, warn};
 
 use crate::{
     api::models::{common::BlockHeight, withdrawal::WithdrawalId},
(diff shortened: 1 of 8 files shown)


emily/handler/src/database/accessors.rs
```diff
@@ -480,6 +574,60 @@ async fn query_with_partition_key<T: TableIndexTrait>(
...
+    sort_key: &<<<T as TableIndexTrait>::Entry as EntryTrait>::Key as KeyTrait>::SortKey,
+    sort_key_operator: &str,
+    maybe_page_size: Option<i32>,
+) -> Result<Vec<<T as TableIndexTrait>::Entry>, Error> {
+    // item aggregator.
+    let mut items: Vec<<T as TableIndexTrait>::Entry> = Vec::new();
+    // Next token.
+    let mut next_token: Option<String> = None;
+    // Loop over all items.
+    loop {
```
djordon, line 589: I'm usually a little wary of loops. This is definitely the right thing to so, but we might want to have some other break condition if this is taking too long.

Actually, maybe the lambda timeout is that break condition so this isn't that bad. Also, we're reaching out to dynamodb so this very likely isn't terrible either. Never mind.

emily/handler/src/database/entries/mod.rs
```diff
@@ -255,6 +255,52 @@ pub(crate) trait TableIndexTrait {
                 <<Self::Entry as EntryTrait>::Key as KeyTrait>::PARTITION_KEY_NAME,
             )
             .expression_attribute_values(":v", serde_dynamo::to_attribute_value(partition_key)?)
+            .scan_index_forward(false)
```
djordon, line 258: What are the sort keys for these tables? Maybe we should add a comment explaining this.

emily/handler/tests/integration/accessors.rs
```diff
@@ -85,3 +92,62 @@ async fn test_chaintip_update() {
...
+        accessors::get_all_deposit_entries_modified_after_height_with_status(
+            &context,
+            &status,
+            minimum_height,
+            None,
+        )
+        .await
+        .expect("Query succeeds")
+        .into_iter()
+        .map(|entry| -> DepositInfo { entry.into() })
```
djordon, line 144: Oh interesting. I did not know that this would work. I thought I would have to do
```rust
DepositInfo { entry: entry.into() }
```

djordon (approved):
Looks good. I would probably remove the `query_with_partition_and_sort_key` function but I'm fine with whatever you decide.

emily/handler/tests/integration/accessors.rs
AshtonStephens: It's because this isn't doing what you think it is! It's actually euquivalent to this:
```rust
        .map(|entry| -> { 
            let deposit_info: DepositInfo = entry.into();
            deposit_info
        })
```

emily/handler/src/database/accessors.rs
```diff
@@ -480,6 +574,60 @@ async fn query_with_partition_key<T: TableIndexTrait>(
...
+        let mut new_items: Vec<<T as TableIndexTrait>::Entry>;
+        (new_items, next_token) = query_with_partition_and_sort_key::<T>(
+            context,
+            parition_key,
+            sort_key,
+            sort_key_operator,
+            next_token,
+            maybe_page_size,
+        )
+        .await?;
```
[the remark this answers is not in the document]
AshtonStephens: I'll remove it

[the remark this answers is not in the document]
AshtonStephens: Done.

emily/handler/tests/integration/accessors.rs
djordon: Ahhh, that makes much more sense.

djordon: In that case, why not just
```rust
.map(DepositInfo::from)
```
In know it's too late now, maybe next time.

Issue #441, opened 2024-08-21
[Feature]: Define types for smart contract events


djordon:
# Feature - Define types for smart contract events

## 1. Description

Stacks-core sends webhooks to event observers as JSON and the signers need to listen for them and deserialize them. We should define concrete types so that we can easily extract the information that we need.

### 1.1 Context & Purpose

Unfortunately, the specific format for these webhooks are not really specified outside of the source and the JSON sent is not a serialization of some type, they use the `json!` macro extensively. There is the [event-dispatcher doc](https://github.com/stacks-network/stacks-core/blob/cbf0c52fb7a0b8ac55badadcf3773ca0848a25cf/docs/event-dispatcher.md#post-new_block) but it is outdated and incorrect (for example, the `burn_block_hash` is always sent and has been [for at least a year](https://github.com/stacks-network/stacks-core/blob/501352f5d831feb04755c3cfd69bf62d4e20fb97/testnet/stacks-node/src/event_dispatcher.rs#L499-L511) but it is missing in the official docs). So we need to define the types ourselves.

### 2.1 Acceptance Criteria:
   - [ ] We have types that can deserialize smart contract events that are sent in `POST /new_block` webhooks.

## 3. Related Issues and Pull Requests (optional):

This is part of https://github.com/stacks-network/sbtc/issues/426.

PR #442, opened 2024-08-21
feat: add types for smart contract events

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/441.

## Changes

* Add concrete types for `POST /new_block` webhooks with types that match on smart contract events. There are a few other kinds of events by this PR only adds types for smart contract events.
* Set up a basic web server with two endpoints, one for the webhooks and another for the application status.
* Rename `Value` to `ClarityValue` where appropriate.

## Testing Information

This includes a unit test that checks the deserialization of the webhooks emitted from stacks core. This payload was sent from a stacks core node under Nakamoto on commit https://github.com/stacks-network/stacks-core/commit/cbf0c52fb7a0b8ac55badadcf3773ca0848a25cf which was made on 2024-08-20. The testing setup here followed the same testing setup as in https://github.com/stacks-network/sbtc/pull/303.

## Checklist:

### DIFF
--- a/signer/src/api/new_block.rs
+++ b/signer/src/api/new_block.rs
@@ -1,0 +1,21 @@
+//! This module is contains the handler for the POST /new_block endpoint,
+//! which is for processing new block webhooks from a stacks node.
+
+use axum::http::StatusCode;
+
+/// A basic handler that responds with 200 OK
+pub async fn new_block_handler() -> StatusCode {
+    StatusCode::OK
+}
+
+/// The Schema for new block events are defined in the source here:
+/// https://github.com/stacks-network/stacks-core/blob/09c4b066e25104be8b066e8f7530ff0c6df4ccd5/testnet/stacks-node/src/event_dispatcher.rs#L644-L687
+pub struct NewBlockEvent {
+
+}
+
+/// https://github.com/stacks-network/stacks-core/blob/develop/clarity/src/vm/events.rs#L358-L363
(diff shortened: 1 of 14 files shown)


AshtonStephens (approved):
LGTM.

signer/src/api/new_block.rs
```diff
@@ -0,0 +1,70 @@
...
+/// second might succeed, we will return a 200 OK status code.
+///
+/// [^1]: <https://github.com/stacks-network/stacks-core/blob/09c4b066e25104be8b066e8f7530ff0c6df4ccd5/testnet/stacks-node/src/event_dispatcher.rs#L317-L385>
+pub async fn new_block_handler<S>(_state: State<S>, body: StacksNodePayload) -> StatusCode
+where
+    S: DbWrite,
+{
+    tracing::info!("Received a new block event from stacks-core");
+
+    let _event: NewBlockEvent = match body {
```
xoloki, line 41: Why store event?

xoloki (approved):
Lgtm, one tiny question but not a blocker

signer/src/api/new_block.rs
djordon: Basically, the signers need it to verify that the sign-stacks-transaction-request from another signer is legit. They need to be able to validate that the amounts, recipient, max fee, are all correct. So when the request comes in they look it up in their database and check that everything matches.

Issue #411, opened 2024-08-09
[Feature]: Bump the version of our stacks-core dependency


djordon:
# Feature - Bump the version of our stacks-core dependency

## 1. Description

The signer binary pulls in crates from the a fork of the stacks-core repo but from a slightly modified version of the develop branch. This code is likely out of date with the latest Nakamoto changes so we will likely need to update it.

### 1.1 Context & Purpose

We need the latest version of stacks-core crates so that there is no discrepancy of the types when we communicate with stacks core.

We currently cannot rely on the official stacks-core crates because of a dependency conflict with `sqlx`. Our `sqlx` dependency pulls in sqlite as a dependency because the `macros` and/or `migrate` features pull it in, and our libstacks dependency pulls in sqlite as well. There is a conflict because the versions don't match and they need to match exactly (because sqlite is basically a C dependency). There is no way around this conflict without doing one of the following:
1. Maintain our own version of stacks-core with a version of rusqlite that is compatible with the version used by `sqlx`. This is what we currently do.
2. Update stacks-core's to a version of rusqlite that is compatible with the version used by `sqlx`.
3. Refactor stacks-core so that the sqlite dependency is always optional. It's supposed to be optional for many of the crates there, just not for the libstacks crate.
4. Change `sqlx` so that we do not pull in the sqlite dependency ever (either the official crate our by fixing it and using our own version).
5. Remove our sqlite dependency from `sqlx` by not using the features that pull in sqlite.

I don't want to do (1) anymore but it's the lowest amount of work. I'd love to do (3) but that is probably out of the question. (5) is the second lowest lift, but requires us to give up on all `sqlx` macros.

### 2.1 Acceptance Criteria:
   - [ ] Use the latest version of the stacks-core crates.

djordon:
A few of notes:
1. The issue that we have with a sqlite conflict is a well known sqlx issue https://github.com/launchbadge/sqlx/issues/3211 that is actually caused by a cargo bug https://github.com/rust-lang/cargo/issues/10801.
2. stacks-core recently updated their dependencies (on the develop branch) to use the latest rusqlite, which pulls in the latest libsqlite-sys. The latest version of sqlx, version 0.8.0, uses the same libsqlite-sys version.
3. We need to remove the `migrate`, `json`, and `time` sqlx features in order to not make it seem like we don't pull in `libsqlite-sys`. After https://github.com/stacks-network/sbtc/pull/436 was merged, we don't actually use the `json` and `time` features.

PR #444, opened 2024-08-22
refactor: bump our version of stacks core

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/411.

## Changes

* Update the signer sqlx dependency.
* Change our use of sqlx features so that we don't trigger the cargo bug. This is kinda described in https://github.com/stacks-network/sbtc/issues/411#issuecomment-2305152215.
* Update the stacks-core dependency. The sqlite dependency on stacks core happens to match the sqlx quasi-dependency of sqlite.
* Add test helpers so that we can easily create and drop databases when running tests. Although this mirrors what happens using the sqlx macro, we needed the ability to create database when simulated signer behavior when there are many signers. We need the ability to drop databases so that our CI server doesn't run out of space. I ran into some issue and I just assumed it was related to this.
* Update the Nakamoto block test fixtures. This is necessary after we updated our version of stacks core, since they changed the structure of the block.
* Update the tests to manually create databases and drop them at the end. If a test fails the database won't be cleaned up. Also, if there is a name collision of a test database during a test, it is forcibly dropped and then recreated.

## Testing Information

There is no new functionality added here, so if tests pass then we are good.

## Checklist:


AshtonStephens (approved):
Tests pass, lgtm.

djordon:
I should note that the stacks-core revision https://github.com/stacks-network/stacks-core/tree/cbf0c52fb7a0b8ac55badadcf3773ca0848a25cf, is based off of the develop branch.

PR #446, opened 2024-08-23
Emily openapi client generation - (OpenAPI Client 2/3)

AshtonStephens:
## Description

Closes: #?

## Checklist:


.generated-sources/emily/client/rust/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "emily-client"
```
djordon, line 2: Oh interesting. Why are we creating a client? Was this the plan the whole time and I just missed the memo?

djordon (approved):
Looks good, just had a more existential question.

.generated-sources/emily/client/rust/Cargo.toml
AshtonStephens: Yeah, the client is really going to be the most useful for the signer itself so it can make the api update calls, and then there's whatever side car rust code is necessary there.

This same client generation will make it easier for the sbtc web app to interface with Emily.

Issue #394, opened 2024-08-05
[Tech Debt]: Use autogenerated OpenAPI client to interface with Emily API in integration tests.


AshtonStephens:
## Description

PR that introduced tech-debt: #386

Use autogenerated OpenAPI client to interface with Emily API in integration tests instead of bespoke manual setup.

AshtonStephens:
Done in #678

Issue #50, opened 2024-04-15
[Design]: Deposit API LLD


AshtonStephens:
# Design - Deposit API LLD

Ownership breakdown: https://github.com/Trust-Machines/sbtc-v1/discussions/47

- ***Primary:*** Ashton (@AshtonStephens )
- **Secondary:** Joey (@xoloki)
- reviewer: Jesus (@setzeus)

This ticket tracks creating the low level design document for the deposit api. For a low level design to be "complete" it must have the approval of the majority of the sBTC team and meet the criteria of the checklist below.

- [WIP Deposit API Design Doc](https://docs.google.com/document/d/13AN9GdLfPYKKPkeHNU4niq_CyJCDxFR7DGBLzLc5gxU/edit#heading=h.7ji77bejwa1f)

## Related Research
- Deposit API 
    - #18 
    - #41 
- Deposit Flow
    - #15
    - #23 
    - #45 
- Withdrawal Flow
    - #16 
- Clarity Info
    - #24  
    - #56 
    - #17

## Acceptance Criteria Checklist

- [x] There's a written low level design doc with the following
    - [x] Clear descriptions of all component interactions including serialization format
    - [x] Diagram/s on how the component will work
    - [x] Clear descriptions of the programming languages and infrastructure that will be used
    - [x] Justifications for the design decisions
    - [x] A **rough task breakdown** of the component's implementation plan referencing inter-dependencies
    - [x] A general list of the of tests required for the component
- [x] Low level design has been reviewed and approved with the ***secondary***.
- [x] Low level design has been reviewed and approved with the team.
- [x] The task breakdown is refined with GitHub issues for each task

AshtonStephens:
Closing out

PR #447, opened 2024-08-23
Emily openapi client test suite - (OpenAPI Client 3/3)

AshtonStephens:
## Description

Closes: #394
Related: #50 

## Changes

Add a test suite for the Emily API that uses the OpenAPI client instead of a bespoke client like in the original integration tests. This is a better model for testing the API than we previously had because it directly mimics how outside users would use the Emily API client.

## Testing Information

To run this you need to run a number of commands concurrently. #505 will make these unnecessary, this is just to test this current code:

- Terminal 1: `make emily-integration-env-up`
- Terminal 2: `make emily-server-watch`
- Terminal 3: `make emily-integration-test`

## Checklist:


emily/client-tests/tests/integration/chainstate.rs
```diff
@@ -0,0 +1,179 @@
...
+            .expect("Received an error after making a valid set chainstate api call.");
+
+    let gotten_post_reorg_chaintip = apis::chainstate_api::get_chain_tip(&configuration)
+        .await
+        .expect("Received an error after making a valid get chaintip api call.");
+
+    // Assert.
+    // --------
+    assert_eq!(expected_post_reorg_chaintip, created_reorged_chainstate);
+    assert_eq!(expected_post_reorg_chaintip, gotten_post_reorg_chaintip);
```
djordon, line 125: At the end of this test we check that the chain tip is at the `reorg_height`, but isn't that incorrect? Why doesn't the chain tip always point to the chain with the max length? I think this implies that a new block with a height that is lower than the current max height is not really a reorg yet.

emily/client-tests/tests/integration/withdrawal.rs
```diff
@@ -0,0 +1,282 @@
...
+        .collect::<Vec<_>>();
+
+    expected_withdrawal_infos.sort_by(arbitrary_withdrawal_info_partial_cmp);
+    gotten_withdrawal_infos.sort_by(arbitrary_withdrawal_info_partial_cmp);
+    assert_eq!(expected_withdrawal_infos, gotten_withdrawal_infos);
+}
+
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
+#[tokio::test]
+async fn update_deposits() {
```
djordon, line 196: nit: typo, should be `update_withdrawals` right?

emily/client-tests/tests/integration/deposit.rs
```diff
@@ -0,0 +1,397 @@
...
+                "Received an error after making a valid get deposits for transaction api call.",
+            );
+
+    // Assert.
+    // -------
+    // Expect the deposits to be sorted by output index.
+    // TODO(506): Reverse this order of deposits for this specific api call.
+    expected_deposits.sort_by(|a, b| {
+        b.bitcoin_tx_output_index
+            .partial_cmp(&a.bitcoin_tx_output_index)
```
djordon, line 215: This can be just `.cmp`

djordon (commented):
This largely looks good. I had a question about the behavior after the API is notified about a block with lower height that is now part of the longest chain following a reorg. The behavior seems unintuitive to me. These tests just check that the behavior so not really a blocker per se.

Also, these tests can be included in the integration tests in the emily handler, so I think that they should be. Cargo thinks `emily-client` is an independent crate, even before we do `make build` so including these tests in the emily handler is no worse than having these tests in a separate crate from cargo's perspective.

emily/client-tests/tests/integration/chainstate.rs
```diff
@@ -0,0 +1,179 @@
...
+        stacks_block_height: height,
+    }
+}
+
+/// An arbitrary fully ordered partial cmp comparator for Chainstate.
+/// This is useful for sorting vectors of chainstates so that vectors with
+/// the same elements will be considered equal in a test assert.
+fn arbitrary_chainstate_partial_cmp(a: &Chainstate, b: &Chainstate) -> Ordering {
+    let a_str: String = format!("{}-{}", a.stacks_block_hash, a.stacks_block_height);
+    let b_str: String = format!("{}-{}", b.stacks_block_hash, b.stacks_block_height);
```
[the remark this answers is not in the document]
AshtonStephens: Neato!

emily/client-tests/tests/integration/chainstate.rs
AshtonStephens: It actually is a reorg! The API follows the event observer, and the event observer will only ever report a burn block if it's now considered the canonical chain tip. The moment there's a conflicting block that's lower than the existing one it should be considered a reorg.

AshtonStephens:
Moving this into

Issue #434, opened 2024-08-17
[Feature]: Add support for `get_data_var` Stacks RPC function in the Stacks API client


cylewitruk:
# Feature - Add support for `get_data_var` Stacks RPC function in the Stacks API client

## 1. Description

Add support for the [/v2/data_var](https://github.com/stacks-network/stacks-core/blob/47db1d0a8bf70eda1c93cb3e0731bdf5595f7baa/docs/rpc-endpoints.md#get-v2data_varstacks-addresscontract-namevar-name) method in [Stacks API client](https://github.com/stacks-network/sbtc/blob/main/signer/src/stacks/api.rs).

### 1.1 Context & Purpose
Support for this method is required by the signer to retrieve signers' public keys from the `current-signer-set` variable in the `sbtc-registry` contract. This method will however be generic, so it will work for retrieving other data variables from other contracts as the need arises as well.

## 2. Technical Details:
The more time-consuming portion of this ticket will be handling deserialization of the values returned from the API, as the value returned can be any valid Clarity value, depending on how the variable is defined in the contract.

I think for this ticket it can be worth scoping the work to only support deserializing `(list x (buff 33))` i.e. a `Vec<Vec<u8>>` as this is the type of the `current-signer-set` variable, which is the immediate need.

Testing will also likely take a little time as the contract needs to be published to a running node and values added to the variable in order to get some example outputs from the node to use in mocked tests.

### 2.1 Acceptance Criteria:
   - [ ] The `current-signer-set` data variable can be read from the `sbtc-registry` contract.

## 3. Related Issues and Pull Requests (optional):
Required for #409

Issue #409, opened 2024-08-07
[Feature]: Use libp2p for signer communication


djordon:
# Feature - Use libp2p for signer communication

## 1. Description

We should configure the sBTC signers to use libp2p for communication. This would increase our resilience if the relay server fails.

The signers would be structured as follows:

1. Use a clarity contract as a central data source for the public keys of the signers. Each signer can query their stacks node for the `current-signer-set` data variable in the `sbtc-registry` contract. This is just an [API call](https://github.com/stacks-network/stacks-core/blob/47db1d0a8bf70eda1c93cb3e0731bdf5595f7baa/docs/rpc-endpoints.md#get-v2data_varstacks-addresscontract-namevar-name) to the stacks node to retrieve the current-signer-set, not a contract call.
2. We could setup a relay server as before, but its role is to facilitate establishing direct connections with peers that are behind a NAT.
3. Have each sBTC signer use a libp2p GossipSub messaging protocol with authentication, using the private key that corresponds to the known public keys.
4. Use connection gating to ensure that only peers from the address noted in the smart contract are valid. In libp2p, peers are identified by their PeerID, and this identifier is set by their public key.

### 1.1 Context & Purpose

This involves pulling the [libp2p crate](https://crates.io/crates/libp2p) into the signer binary and configuring it for signer to signer communication.

## 2. Technical Details:

There are at least two hard parts here: 

1. Implementing connection gating. I believe this requires an implementation of the `libp2p::swarm::NetworkBehaviour` trait.
2. Managing connection changes when we rotate keys. When we rotate keys, signers will probably start signing messages with the new key. It seems dangerous to keep signing messages over existing connections using an "old" key. This is something that we need to handle using the current design.

### 2.1 Acceptance Criteria:
For this ticket, we do not need to support establishing connections with servers behind a NAT.

   - [ ] Signers can communicate in a p2p, without the help of a relay server.

djordon:
## Use of libp2p

The libp2p seems to be a very mature and actively maintained library.

1. The library is created and primarily maintained by [Protocol Labs](https://protocol.ai/), which are the same folks behind IPFS and Filecoin.
2. Ethereum 2.0 (post The Merge) uses libp2p in its consensus layer. I believe they use the go implementation of libp2p.
3. The polkadot blockchain https://github.com/paritytech/polkadot-sdk uses the Rust implementation of libp2p.
[...]

PR #451, opened 2024-08-26
add support for get-data-var stacks node api call

cylewitruk:
## Description

Closes: #434
Related to: #409 

This took a little longer than I thought as I was trying to get a node with the contracts running where I could see what's actually being output. I really wanted to do this without pulling more dependencies on `stacks-core`, but that turned out to take more time than I think it's worth for right now, so I ended up just using Clarity's own `Value` type for de/serialization so we can get this closed.

## Changes
Adds an additional method to the Stacks API client for calling the `/v2/data_var` endpoint and retrieving the value of a contract `data-var`.

## Testing Information
Added mocked tests. Mocked responses are generated by building up the Clarity type as a `Value`, serializing it and placing it into the API's response JSON format. I added test cases for both empty list and max-size (128).

## Checklist:

### DIFF
--- a/signer/src/stacks/api.rs
+++ b/signer/src/stacks/api.rs
@@ -26,6 +26,7 @@ use serde::{Deserialize, Deserializer};
 
 use crate::config::StacksSettings;
 use crate::error::Error;
+use crate::keys::PublicKey;
 use crate::storage::DbRead;
 
 use super::contracts::AsTxPayload;
@@ -73,13 +74,14 @@ pub enum FeePriority {
 
 /// A trait detailing the interface with the Stacks API and Stacks Nodes.
 pub trait StacksInteract {
-    /// Retrieve a data variable from a contract.
-    fn get_data_var(
+    /// Retrieve the current signer set from the `sbtc-registry` contract.
+    ///
+    /// This is done by making a `GET /v2/data_var/<contract-principal>/sbtc-registry/current-signer-set`
+    /// request.
+    fn get_current_signer_set(
         &mut self,
         contract_principal: &StacksAddress,
-        contract_name: &ContractName,
-        var_name: &ClarityName,
-    ) -> impl Future<Output = Result<Value, Error>>;
+    ) -> impl Future<Output = Result<Vec<PublicKey>, Error>>;
     /// Get the latest account info for the given address.
     fn get_account(
         &mut self,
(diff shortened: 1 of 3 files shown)


signer/src/stacks/api.rs
```diff
@@ -71,6 +73,13 @@ pub enum FeePriority {
 
 /// A trait detailing the interface with the Stacks API and Stacks Nodes.
 pub trait StacksInteract {
+    /// Retrieve a data variable from a contract.
+    fn get_data_var(
+        &mut self,
+        contract_principal: &StacksAddress,
+        contract_name: &ContractName,
+        var_name: &ClarityName,
+    ) -> impl Future<Output = Result<Value, Error>>;
```
djordon, line 82: Hmmm, I think it would be better to leave out the generic `get_data_var` function from the trait and instead have a `get_current_signing_set` function (or whatever name) that does the conversion of a clarity `Value` into a "regular" type for you. This should make testing the behavior a little simpler since you wouldn't need to return a `Value` in the test structs that implement the trait.

djordon (commented):
Looks good. I would just change the `get_data_var` function in the trait to a more specific `get_signing_set` function that returns a nicer type.

signer/src/stacks/api.rs
cylewitruk: @djordon yeah that's a good point -- I pushed some changes now to refactor how that works a little, with some slightly improved error handling as well.

signer/src/stacks/api.rs
```diff
@@ -590,6 +650,42 @@ fn check_result<T>(client: &mut StacksClient, result: Result<T, Error>) -> Resul
 }
 
 impl StacksInteract for StacksClient {
+    async fn get_current_signer_set(
+        &mut self,
+        contract_principal: &StacksAddress,
+    ) -> Result<Vec<PublicKey>, Error> {
+        use clarity::vm::types::{BuffData, ListData, SequenceData, Value};
```
djordon, line 657: Isn't `Value` already imported? Well, I would move these to the top of the file.

signer/src/stacks/api.rs
```diff
@@ -590,6 +650,42 @@ fn check_result<T>(client: &mut StacksClient, result: Result<T, Error>) -> Resul
...
+        let result = check_result(
+            self,
+            StacksClient::get_data_var(
+                self,
+                contract_principal,
+                &ContractName::from("sbtc-registry"),
+                &ClarityName::from("current-signer-set"),
+            )
+            .await,
+        )?;
```
djordon, line 668: I find it a lot easier to read if this is broken out into two different variables, like:
```rust
let res = self.get_data_var(
    contract_principal,
    &ContractName::from("sbtc-registry"),
    &ClarityName::from("current-signer-set"),
)
.await;
let result = check_result(self, res)?;
```

djordon (commented):
Looks good, I had two small style nits.

signer/src/stacks/api.rs
cylewitruk: Yeah it was. For this case I just wanted to avoid polluting the import space with low-level Clarity types, since they'll probably only be used in one or two specific cases.

Anyway, moved them up.

signer/src/stacks/api.rs
cylewitruk: Separated them.

Tbh `check_result` feels a bit weird here, feels more like something that should be refactored up into the `StacksClient` itself.

signer/src/stacks/api.rs
djordon: Oh, good call out. I think what you have here is fine.

djordon (approved):
Looks good

signer/src/stacks/api.rs
djordon: Yeah, I can see that.

Maybe it's just me, but I feel like the `StacksClient` should have simple functions that hit the endpoint and do basic conversions into the return type. For additional functionality like retries and `check_result`, they should be done at a layer on top of these simple functions. Well, that's why I pushed to move them outside of the core functions themselves, with the compromise being that they are now in the traits. But yeah, it looks a bit weird to me too.

cylewitruk: I think that since `check_result` right now is only checking for connection-related errors and rotating the endpoint (which belongs to `StacksClient`) that that feels more "transport-level" i.e. `StacksClient`. Ideally I think that `StacksInteract` should be oblivious to such things - either it succeeded or failed. `StacksClient` could have configurable timeout/retry policies etc. which would give the caller more control over when it should error.

My $0.02 :)

PR #452, opened 2024-08-26
feat: modify signature threshold rotate keys

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/424

## Changes

* Allow the signature threshold be changed during rotate-keys contract calls.
* Add new tests given the new behavior.

## Checklist:

### DIFF
--- a/contracts/contracts/sbtc-bootstrap-signers.clar
+++ b/contracts/contracts/sbtc-bootstrap-signers.clar
@@ -14,7 +14,7 @@
 ;; The function caller is not the current signer principal
 (define-constant ERR_INVALID_CALLER (err u201))
 ;; The given signature threshold must be greater than 50% and less than or
-;; equal to 100% of the total number signer keys.
+;; equal to 100% of the total number of signer keys.
 (define-constant ERR_SIGNATURE_THRESHOLD (err u202))
 
 ;; data vars
@@ -38,7 +38,7 @@
             (current-signer-data (contract-call? .sbtc-registry get-current-signer-data))   
             (new-signer-principal (pubkeys-to-principal new-keys new-signature-threshold))
         )
-        ;; Checkc that the signature threshold is valid
+        ;; Check that the signature threshold is valid
         (asserts! (and (> new-signature-threshold (/ (len new-keys) u2))
                        (<= new-signature-threshold (len new-keys))) ERR_SIGNATURE_THRESHOLD)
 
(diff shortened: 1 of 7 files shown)


contracts/contracts/sbtc-bootstrap-signers.clar
```diff
@@ -27,12 +28,20 @@
...
+    (new-aggregate-pubkey (buff 33))
+    (new-signature-threshold uint)
+  )
     (let 
         (
             (current-signer-data (contract-call? .sbtc-registry get-current-signer-data))   
-            (new-signer-principal (pubkeys-to-principal new-keys signature-threshold))
+            (new-signer-principal (pubkeys-to-principal new-keys new-signature-threshold))
         )
+        ;; Check that the signature threshold is valid
```
setzeus, line 41: Clarity division rounds down so just making sure that it's clear here that u1 / u2 == 0. I don't believe that it materially changes the logic here but just a NIT.

setzeus (commented):
Everything LGTM except for one NIT that we should be aware of but doesn't change anything.

setzeus (approved):

contracts/contracts/sbtc-bootstrap-signers.clar
```diff
@@ -27,12 +28,20 @@
...
+        (asserts! (and (> new-signature-threshold (/ (len new-keys) u2))
```
hstove, line 42: Where does this "rule" come from? If we're going to let the signers determine a new threshold (in this code, between 50%-100%), why not lower?

hstove (commented):
Overall the code seems valid and makes sense, but I'm uncertain how I feel regarding the comment I made about the valid threshold range.

contracts/contracts/sbtc-bootstrap-signers.clar
djordon: Good questions. I think I originally had it at any positive threshold and thought "we'll probably want to ensure that the threshold will never be less than 50%". So, this comes out of thin air. Either greater than 50% or greater than 0 sound defensible to me and I do not have any strong feels on either.

hstove (approved):

contracts/contracts/sbtc-bootstrap-signers.clar
hstove: Yep, I think that's fair!

Issue #448, opened 2024-08-25
[Feature]: Transform smart contract print events into proper types


djordon:
# Feature - Transform smart contract print events into proper types

## 1. Description

When we subscribe to sBTC smart contract print events, the data that we get is a clarity value. Before we write it to our database, we to validate that it is of the right form, and transform it into a type that we can insert into the database. This ticket is about writing to code that transform the clarity value into more "normal" rust type for each of the print events that we are expecting.

### 2.1 Acceptance Criteria:
   - [ ] We have code that transforms clarity values into proper types for each of the print events that we are watching.

## 3. Related Issues and Pull Requests (optional):

Relates to https://github.com/stacks-network/sbtc/issues/426.

PR #461, opened 2024-08-27
feat: transform print events into proper types

djordon:
## Description

Closes part of https://github.com/stacks-network/sbtc/issues/448. There is a small part left out of this PR, where I do not parse the recipient into a proper type. That will be done in a subsequent PR.

## Changes

1. Add functions for deconstructing print event Clarity `Values` into regular types.
2. Be explicit about the endianness of the signer bitmap when converting it into an integer.

## Testing Information

I did some local testing, but ran into an issue with my local stacks-node not confirming my contract call transaction after the first. Not sure what the issue is. When I get that sorted I think this can be considered "well-tested".

## Checklist:

### DIFF
--- a/signer/src/error.rs
+++ b/signer/src/error.rs
@@ -36,6 +36,20 @@ pub enum Error {
     #[error("Could not convert an integer in clarity event into the expected integer {0}")]
     ClarityIntConversion(#[source] std::num::TryFromIntError),
 
+    /// This happens when we attempt to create s String from the raw bytes
+    /// returned in a Clarity [`Value`](clarity::vm::Value).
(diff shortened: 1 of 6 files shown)


signer/src/error.rs
```diff
@@ -31,6 +31,34 @@ pub enum Error {
     #[error("could not serialize bitcoin transaction into bytes.")]
     BitcoinEncodeTransaction(#[source] bitcoin::io::Error),
 
+    /// This error is thrown when converting a u128 some other smaller
```
xoloki, line 34: to u128?

xoloki (approved):
Lgtm, just one janky comment

PR #462, opened 2024-08-28
feat: transform event recipients values

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/448. This is the second part of https://github.com/stacks-network/sbtc/pull/461 that deals with converting recipients into addresses.

## Changes

* Add a function that transforms a clarity recipient into a `bitcoin::Address`.
* Update the documentation in the initiate-withdrawal-request function that details what is expected for the recipient.
* Update the `InitiateWithdrawalRequest` testing struct and add it to the contract to the integration tests to test generating events.

## Testing Information

We add lots of unit tests for the new function. Sadly my integration test setup has a snafu (the transaction is accepted but doesn't appear to be confirmed). I'll try again later.

## Checklist:


signer/src/keys.rs
```diff
@@ -254,7 +254,7 @@ impl sqlx::postgres::PgHasArrayType for PublicKey {
     }
 }
 
-#[cfg(feature = "testing")]
+#[cfg(any(test, feature = "testing"))]
```
netrome, line 257: Do we ever run the tests without the `testing` feature? If that is so, I think we should do this update on all the other dummy implementations.

signer/src/stacks/events.rs
```diff
@@ -215,8 +220,159 @@ fn withdrawal_create(mut map: RawTupleData, network: NetworkKind) -> Result<Regi
...
+/// ## P2TR
+///
+/// Pay-to-taproot addresses are formed by "tweaking" the x-coordinate of a
+/// public key with a merkle tree. The result of the tweak is used as the
+/// witness program for the address.
+///
+/// To specify this address type in the `initiate-withdrawal-request`
+/// contract call, the `version` is 0x06 and the `hashbytes` is the
+/// "tweaked" public key.
+fn recipient_to_address(mut map: RawTupleData, network: NetworkKind) -> Result<Address, Error> {
```
netrome, line 317: Style suggestion: Given that this is essentially a constructor, I'd consider implementing this using an extension trait on the `Address` type. For example

```
trait FromClarityRecipient {
    from_recipient(map: RawTupleData, network: NetworkKind) -> Result<Self, Error>
}

impl FromClarityRecipient for Address {
    ...
}
```

This way, the code would later read as `Address::from_recipient` which to me reads better than an arbitrary function.

netrome: Alternatively this could be a method on `RawTupleData` as
```
fn to_address(network: NetworkKind) -> ...
```

signer/src/stacks/events.rs
```diff
@@ -505,10 +682,103 @@ mod tests {
...
+    // hashes.
+    fn new_p2sh_segwit<T: AsRef<bitcoin::script::PushBytes>>(data: T) -> ScriptHash {
+        ScriptBuf::builder()
+            .push_int(0)
+            .push_slice(data)
+            .into_script()
+            .script_hash()
+    }
+
+    fn to_tuple_data<const N: usize>(version: u8, hash: [u8; N]) -> RawTupleData {
```
netrome, line 714: Since this is a perfect conversion, this could be nice as a `From`- implementation

```
impl<const N: usize> From<(u8, [u8; N])> for RawTupleData
```

Alternatively as a constructor on the `RawTupleData` struct.

```
impl RawTupleData {
    fn new<const N: usize>(version: u8, hash: [u8; N]) -> Self {}
    }
```

netrome (approved):
Looks good to me. I have a few style suggestions, and a question about the testing feature.

signer/src/keys.rs
djordon: > Do we ever run the tests without the `testing` feature? 

No we don't, and testing is one our default features too. 

I think I originally made more substantial changes to this file and rolled them all back and just didn't roll back this one. I don't mind doing the same for our other dummy implementations, I just would prefer to do it in a separate PR. I don't mind removing this from this PR too.

signer/src/stacks/events.rs
djordon: Good ideas. I think I like the second one more, since I don't see much of a future for the trait approach (what else is going to implement it?). So I'll make it a method for the type.

signer/src/stacks/events.rs
djordon: 👍🏽 today I like the constructor approach.

signer/src/keys.rs
netrome: Yeah sure, I think it makes sense to do it as a follow-up :)

PR #473, opened 2024-08-30
feat: libp2p integration

cylewitruk:
## Description

The first skeleton implementation of libp2p and its integration into the main runloop.

Partially resolves: #409

## Changes
- Refactored the app signalling channel to use commands/events (two new enums). The current existing shutdown signal is moved to commands.
- Added the libp2p implementation, which includes
  - mDNS (for local peer discovery, probably not used by most signers unless they're running multiple on the same network, but it's nice for local testing -- might gate this to dev builds only..)
  - DNS resolution for `Multiaddr`s which are given using hostname
  - AutoNAT (automatic detection of whether or not the peer is behind a NAT)
  - Kademlia DHT (peer discovery over WAN). _Note: we don't use Kademlia's DHT directly, but the Kademlia protocol includes efficient peer discovery for which it internally uses its DHT, and is the recommended way to handle peer discovery over WAN without centralization (rendezvouz)._
  - Ping (network maintenance)
  - Relay (for NAT/hole-punching) - **not fully implemented**
- The libp2p swarm listens for publish commands from the application signalling channel and emits `P2PPublishSuccess`/`P2PPublishFailure` events, employing a fully de-coupled and asynchronous pattern. A caller can use the `MsgId` to wait themselves for a success or failure event if they need to react in a more-or-less synchronous manner.
- The libp2p swarm emits `P2PMessageReceived` events when gossipsub messages are received.

## Testing Information
TBD

## To-do (not necessarily in this PR):
- [x] Write a `MessageTransfer` impl which uses app signalling (and thus completely decoupled from libp2p itself)
- [ ] Add support for above impl (`P2PNetwork`) to handle buffering of messages instead of requiring the caller to do it themselves.
- [ ] Investigate if we want to use more than one gossipsub topic
- [ ] Implement peer filtering - _what messages should be filtered to the active signer set? I would assume that some messages still need to be allowed for a peer to be able to participate in signer-set selection?_
- [ ] Implement NAT hole-punching
- [x] Write tests
- [ ] Implement message validation
- [ ] It should be an error on startup if no seed addrs could be resolved to IPs (i.e. we have no IP addresses of seeds to try to bootstrap with)

## Thoughts:
- ~~Maybe we want to use a separate `watch` channel for signalling termination... I'm on the fence on whether or not there will be enough chatter to cause any noticeable performance deviations (vs. added complexity) as each receiver will receive its own cloned copy of the `SignerSignal`, but the cost of cloning and ignoring (dropping) irrelevant messages is probably very small.~~
EDIT: Went ahead and refactored this.

## Checklist:


cylewitruk:
I'll move the signalling changes to a separate PR if they're OK with you guys

signer/src/error.rs
```diff
@@ -3,11 +3,31 @@ use std::borrow::Cow;
...
 /// Top-level signer error
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+    /// Error broadcasting a message to the application channel
+    #[error("Error broadcasting message to application channel: {0}")]
+    BroadcastSendError(#[from] tokio::sync::broadcast::error::SendError<SignerSignal>),
+
+    /// Error receiving a message from the application channel
+    #[error("Error receiving message from application channel: {0}")]
+    BroadcastRecvError(#[from] tokio::sync::broadcast::error::RecvError),
```
netrome, line 17: Both of these error types should be handled by the logic and not propagated imo. For example, the send error should only occur if the receiver side is dropped, which should only happen during shutdown and can be used as a signal to terminate the specific sub-channel instead of having a dedicated termination channel.

signer/src/error.rs
```diff
@@ -3,11 +3,31 @@ use std::borrow::Cow;
...
+
+    /// Error sending a message to the application termination channel
+    #[error("Error sending message to application termination channel: {0}")]
+    TerminationSendError(#[from] tokio::sync::watch::error::SendError<bool>),
+
+    /// Error sending a message to the application termination channel
+    #[error("Error receiving message from application termination channel: {0}")]
+    TerminationRecvError(#[from] tokio::sync::watch::error::RecvError),
```
netrome, line 25: The same goes for these types. I think we can explicitly handle all failure cases in the code and should not re-propagate them.

signer/src/network/mod.rs
```diff
@@ -26,6 +37,80 @@ pub trait MessageTransfer {
     fn receive(&mut self) -> impl Future<Output = Result<Msg, Self::Error>> + Send;
 }
 
+/// MessageTransfer interface for the application signalling channel.
+/// TODO: Better name?
+pub struct SignallingNetwork {
```
netrome, line 42: I guess the fact that this uses the signer signals is an implementation detail, but in general this represents a hook to the p2p network, so perhaps it should simply be named `P2PNetwork`.

signer/src/network/libp2p/swarm.rs
```diff
@@ -0,0 +1,215 @@
...
+
+use super::errors::SignerSwarmError;
+use super::event_loop;
+
+/// Define the behaviors of the [`SignerSwarm`] libp2p network.
+#[derive(NetworkBehaviour)]
+pub struct SignerBehavior {
+    pub gossipsub: gossipsub::Behaviour,
+    mdns: mdns::tokio::Behaviour,
+    kademlia: kad::Behaviour<MemoryStore>,
```
netrome, line 29: What do we need to store in kademlia?

netrome (commented):
Nice stuff so far!. I like the general approach of running the libp2p swarm separately and implementing `MessageTransfer` on a struct that communicates with the swarm over message passing.

At this point, I just have a few minor questions and some comments.

signer/src/network/libp2p/swarm.rs
cylewitruk: We don't specifically store anything in Kademlia, but the Kademlia protocol includes efficient peer discovery and propogation, so it's generally the recommended peer discovery mechanism. mDNS can only discover local nodes, and rendezvouz is "centralized".

signer/src/network/mod.rs
cylewitruk: Changed in [376cace](https://github.com/stacks-network/sbtc/pull/473/commits/376cacef68fb40239e51df5049feabaeed3189f6)

signer/src/error.rs
cylewitruk: Removed these in [376cace](https://github.com/stacks-network/sbtc/pull/473/commits/376cacef68fb40239e51df5049feabaeed3189f6).

I changed these to simply be swallowed because if there's a send error, then we're already going down, and a receive error is impossible (the one place it's being used).

signer/src/network/libp2p/swarm.rs
netrome: Oh interesting. So just adding the kademlia behavior here gives us peer discover that way. Nice, thanks for the info!

signer/src/error.rs
netrome: Great. Note that we use these errors as a shutdown signal in many other components right now, which I think is pretty nice because it means that dropping a channel to one of these actors shuts it down without requiring an explicit shut-down mechanism. This could be an alternative to using a termination handle altogether.

netrome: But I'm okay with having an explicit termination handle as well.

signer/src/error.rs
netrome: My two cents here is that while I'm not against handling shutdown signals explicitly, I think it can be pretty nice to handle it implicitly as well. If you have a handle to a subtask in the form of a send or receiver channel (or both), it makes sense for the subtask to gracefully shut down once either of those are dropped, similar to how normal structures are deallocated when any references to them are dropped.

I do see that there might be scenarios where some extra sophisticated shutdown management is needed, but in my experience you can get very far by this simple rule.

cylewitruk: The issue I see is that it explicitly requires a `send` to occur in order to detect that all receivers have been dropped -- otherwise the loop will continue indefinitely waiting for a command to broadcast a p2p message (thus blocking shutdown). The channel lives as long as the `SignerContext` (as it holds a reference to a `Sender` to be able to give out new subscriptions), so the channel itself won't be dropped until the end of the program.

netrome: Oh right, similar to how cyclic references are never garbage collected. Fair enough, then I see the point in having the explicit termination signal.

signer/src/network/libp2p/swarm.rs
```diff
@@ -0,0 +1,273 @@
...
+            if !self.seed_addrs.contains(addr) {
+                self.seed_addrs.push(addr.clone());
+            }
+        }
+        self
+    }
+
+    /// Build the [`SignerSwarm`], consuming the builder.
+    pub fn build(self) -> Result<SignerSwarm, SignerSwarmError> {
+        let keypair = Keypair::ed25519_from_bytes(self.private_key.to_bytes())?;
```
djordon, line 141: Curious, why the `ed25519` curve instead of `secp256k1`?

signer/src/network/libp2p/utils.rs
```diff
@@ -0,0 +1,94 @@
...
+
+/// A trait for converting a value into a vector of [`Multiaddr`]s.
+pub trait TryIntoMultiAddrs {
+    /// Attempts to convert the value into a vector of [`Multiaddr`]s.
+    #[allow(dead_code)]
+    fn try_into_multiaddrs(&self) -> Result<Vec<Multiaddr>, SignerSwarmError>;
+}
+
+impl TryIntoMultiAddrs for Url {
+    fn try_into_multiaddrs(&self) -> Result<Vec<Multiaddr>, SignerSwarmError> {
```
djordon, line 18: Hmmm, what are your thoughts of writing multi-addresses directly in the config? That's probably what I would do, but I'm curious about why you chose this approach?

djordon (commented):
In general, I really like this approach. I think we should continue down this path.

I mainly tried to stay away from doing a "full" review since I know this is still in progress. But I commented on some things that were worth pointing out and also asked a few questions.

As for the TODOs, I think the scope that we have in this PR is fine, and we can get to the other stuff later (which is what I think you had in mind too).

signer/src/network/libp2p/event_loop.rs
```diff
@@ -0,0 +1,285 @@
...
+                    SwarmEvent::Behaviour(SignerBehaviorEvent::AutonatServer(event)) =>
+                        handle_autonat_server_event(&mut swarm, event),
+                    SwarmEvent::NewListenAddr { address, .. } => {
+                        tracing::info!(%address, "Listener started");
+                    },
+                    SwarmEvent::ExpiredListenAddr { address, .. } => {
+                        tracing::debug!(%address, "Listener expired");
+                    },
+                    SwarmEvent::ListenerClosed { addresses, reason, .. } => {
+                        tracing::info!(addresses = format!("{:?}", addresses), reason = format!("{:?}", reason), "Listener closed");
```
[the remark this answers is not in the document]
cylewitruk: Ah right, think I learned that after I wrote this :sweat_smile:

signer/src/main.rs
```diff
@@ -87,28 +138,52 @@ async fn run_shutdown_signal_watcher(ctx: &impl Context) -> Result<(), Error> {
...
+    tracing::debug!("building the libp2p swarm");
+    let mut swarm = SignerSwarmBuilder::new(&ctx.config().signer.stacks_account.private_key)
+        .add_listen_endpoints(&listen_addrs)
+        .add_seed_addrs(&seed_addrs)
+        .build()?;
+
+    // Start the libp2p swarm. This will run until either the shutdown signal is
+    // received, or an unrecoverable error has occurred.
+    tracing::info!("starting the libp2p swarm");
+    swarm.start(ctx).await.map_err(|error| {
```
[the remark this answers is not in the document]
cylewitruk: Nah that was to squeeze in a log row and send shutdown signal, looks like I forgot to remove this after introducing `checked_run()` as that will take care of it now.

signer/src/network/mod.rs
```diff
@@ -26,6 +37,80 @@ pub trait MessageTransfer {
...
+    /// To avoid ending up in a slow-receiver situation, you should queue
+    /// messages in a local buffer (i.e. [`std::collections::VecDeque`]) and
+    /// process them in your own time. Otherwise, if there are a large number
+    /// of messages being sent, you risk lagging and eventually having the tail
+    /// of the receiver being dropped, thus missing messages.
+    ///
+    /// In other words, you should be calling this method as rapidly as possible.
+    async fn receive(&mut self) -> Result<Msg, Self::Error> {
+        loop {
+            match self.signal_rx.recv().await {
```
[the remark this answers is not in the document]
cylewitruk: That's an exercise for the caller, to match requests to responses (`MessageTransfer` in itself doesn't provide any request/response functionality).

Basically the caller, if expecting specific responses to a message, would need to maintain a map of in-flight requests and try to match any received messages to the in-flight messages.

signer/src/network/mod.rs
```diff
@@ -26,6 +37,80 @@ pub trait MessageTransfer {
...
+                Ok(SignerSignal::Event(SignerEvent::P2PMessageReceived(msg))) => {
+                    return Ok(msg);
+                }
+                // And if we get an error when attempting to read from the
+                // channel.
+                Err(error) => {
+                    return Err(error.into());
+                }
+                // Anything else, we ignore.
+                _ => continue,
```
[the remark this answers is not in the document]
cylewitruk: That was because we're only interested in `P2PMessageReceived` events, i.e. `receive()` doesn't need to know about `P2PPublish` commands coming through.

But you made me realize that this would never shutdown on its own so I've refactored it to a `select!` instead (which is also implicitly ignoring other message types we're not interested in).

signer/src/network/libp2p/event_loop.rs
```diff
@@ -0,0 +1,304 @@
...
+        Event::Discovered(peers) => {
+            for (peer_id, addr) in peers {
+                tracing::info!(%peer_id, %addr, "Discovered peer via mDNS");
+                swarm.behaviour_mut().gossipsub.add_explicit_peer(&peer_id);
+            }
+        }
+        // A multicast-DNS event indicating that a previously discovered peer
+        // has expired. This is raised when the TTL of the autodiscovery has
+        // expired and the peer's address has not been updated.
+        Event::Expired(peers) => {
```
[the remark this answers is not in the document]
cylewitruk: There's a number of events that I think we can make use of for keeping better internal track of peers. Like, we may want to implement some sort of penalization for bad behavior, etc. 

This specific event is only for mDNS peer discovery, which can only discover peers on the local network (basically it's useful for testing), so I don't think we have any real use for this one specifically.

signer/src/network/libp2p/event_loop.rs
```diff
@@ -0,0 +1,304 @@
...
+    use identify::Event;
+
+    match event {
+        Event::Received { peer_id, info, .. } => {
+            tracing::debug!(%peer_id, "Received identify message from peer; adding to confirmed external addresses");
+            swarm.add_external_address(info.observed_addr.clone());
+        }
+        Event::Pushed { connection_id, peer_id, info } => {
+            tracing::trace!(%connection_id, %peer_id, ?info, "Pushed identify message to peer");
+        }
```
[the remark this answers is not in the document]
cylewitruk: This is an event that occurs after the `identify/push` protocol has been used, which is a protocol op for proactively informing peers of a change like a new public listening address, etc. We don't use this now and I'm not sure we will (I think we said signers will likely be behind static addrs?).

signer/src/network/libp2p/swarm.rs
cylewitruk: Not by choice, it's the only method I found to get a `Keypair` from bytes. Seems that libp2p only supports ed25519. Do you think this will be an issue?

signer/src/network/libp2p/utils.rs
cylewitruk: I chose this method because multiaddrs aren't really that popular, users will be more familiar with the standard `scheme://host:port` format in the config.

I.e., a `quic-v1` address looks like `/ip4/127.0.0.1/udp/1234/quick-v1` and tcp is `/ip4/127.0.0.1/tcp/1234` -- just easier for the user imo.

signer/src/context.rs
```diff
@@ -17,39 +17,110 @@ pub trait Context {
...
+    config: Arc<Settings>,
+    // Handle to the app signalling channel. This keeps the channel alive
+    // for the duration of the program and is used both to send messages
+    // and to hand out new receivers.
+    signal_tx: Arc<Sender<SignerSignal>>,
+    /// Handle to the app termination channel. This keeps the channel alive
+    /// for the duration of the program and is used to provide new senders
+    /// and receivers for a [`TerminationHandle`].
+    term_tx: Arc<tokio::sync::watch::Sender<bool>>,
 }
```
netrome, line 40: Instead of having individual `Arc` pointers to all the internal fields, how about just creating a single arc to the `SignerContext` that we pass around instead?

This could also be either a newtype or a type alias:

```rust
pub type SharedSignerContext = Arc<SignerContext>,
```
or
```rust
pub struct SharedSignerContext(Arc<SignerContext>);

impl AsRef<SignerContext>  for SharedSignerContext{...}

impl Deref for SharedSignerContext {type Target=SignerContext; ...}
```

signer/src/network/mod.rs
```diff
@@ -41,4 +133,70 @@ mod tests {
...
+            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+            .try_init();
+
+        let key1 = PrivateKey::new(&mut rand::thread_rng());
+        let key2 = PrivateKey::new(&mut rand::thread_rng());
+
+        // TODO: I'm not sure why we need to set the env vars here, when you run this test
+        // individually or with `cargo nextest run` it works fine without, but with
+        // `cargo test -- --test-threads 1` and `make test` it fails with a config
+        // error that the private key is only 2 bytes long .oO
```
netrome, line 154: Weird. Can we create a follow-up issue to track this? Or do you plan to address it before merging in this PR?

signer/src/network/libp2p/swarm.rs
netrome: I don't think it will be an issue. The messages are signed with ECDSA signatures over `secp256ka` and they are transported over a p2p network which, while using the same keys, uses EdDSA signatures over Curve25519. Those doesn't have to have anything to do with each other and the fact that we're reusing the same keys is more of a convenience than a necessity here.

signer/src/network/mod.rs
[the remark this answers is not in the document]
cylewitruk: Actually just removed that err match as it will never be hit

signer/src/main.rs
```diff
@@ -37,7 +39,12 @@ struct SignerArgs {
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Initialize logging
-    sbtc::logging::setup_logging("info,signer=debug", false);
+    // TODO: The whole logging thing should be revisited. We should support
+    //   enabling different layers, i.e. for pretty console, for opentelem, etc.
```
netrome, line 43: Sure, can you create and link an issue for this?

netrome (commented):
Haven't done an exhaustive review as the PR seems to still be in flux but I have a few comments and thoughts so far. Otherwise, most of the PR looks sensible to me. Great job so far! :rocket:

signer/src/context.rs
cylewitruk: Yeah I've thought about this too, my plan was on using the pattern that `Sender`s and `sqlx` do, with an outer and inner type:
```rust
struct SignerContext {
  inner: Arc<InnerSignerContext>
}
```

signer/src/network/mod.rs
cylewitruk: I'm not sure there's anything to address, I played with this for a while.  Moving to `nextest` is the solution :wink:

signer/src/network/libp2p/swarm.rs
cylewitruk: Referencing @netrome's comment below:
https://github.com/stacks-network/sbtc/pull/473#discussion_r1745996675

signer/src/network/mod.rs
cylewitruk: Fixed it, actually. I changed `SignerContext::init()` to take a `Settings` instead of the path (which I was planning on doing anyway, since we will likely need `Settings` in `main()` to setup logging) -- that seems to have done it :shrug:

signer/src/network/libp2p/utils.rs
cylewitruk: Oh right, I also didn't want any network requirements on loading the config (i.e. dns) or for a dns resolution failure to be a config error (we can handle that gracefully at runtime).

signer/src/main.rs
cylewitruk: Linked #497

signer/src/network/libp2p/event_loop.rs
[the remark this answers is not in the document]
djordon: > (I think we said signers will likely be behind static addrs?)

We haven't settled on the requirements for the signers, but I might have said that one option is that we require signers to be behind static addresses.

> For me right now this falls under the list of one of the events we'll likely remove later.

👍🏽 cool, I was just curious.

signer/src/network/libp2p/swarm.rs
djordon: Yeah, I didn't think this will be an issue either. But I believe `secp256k1` is supported via a feature flag. The thought that I had is that we can easily create a `secp256k1` variant of the `libp2p::Keypair` from our internal types, and I was curious if there was any reason why we should not do that.

cylewitruk: > The thought that I had is that we can easily create a secp256k1 variant of the libp2p::Keypair from our internal types

This sounds nicer, i'll look into it 😊

djordon: We can do that after this PR too. I just wanted to share what I was thinking.

signer/src/network/libp2p/utils.rs
djordon: Hmmm, I think I understand now. There are a couple of things that seem clearer now that weren't obvious to me before:

1. This function does a DNS look-up via the `std::net::ToSocketAddrs::to_socket_addrs` function. So it requires an internet connection in order to work as designed[^1].
2. A single URL can contain a domain name, which may resolve to multiple IPv4 and/or IPv6 addresses. This implementation uses the raw IPs when constructing the multi-addresses. That's why we return a `Vec` here.
3. There is a `libp2p::core::multiaddr::from_url_lossy` function that looks like a good fit here, but it hasn't kept up with all of the supported protocols, so we need our own implementation if we want to support regular URLs.
4. A multi-address does not currently include the username, password, path, and query stuff that we can see in a URL. Multi-addresses only use the scheme, host, and port of a URL.

If we deserialize straight into a `MultiAddr` I don't think that we'd have any network requirements when loading the config. Right now I'm leaning toward expecting multi-addresses in the config as it's what we actually use. Using a URL does not allow us to use the `p2p` protocol part of a multi-address (or any other features that I do not know about). But that might be a non-issue.

Well, I get how people are more familiar with URLs, but something feels off with using them instead of multi-addresses.

[^1]: My first thought was that having the domain resolve into a static IP is an issue. The IPs associated with a domain name can change, so the IP stored locally would be incorrect. I think clients should have some expiry that requires it to re-lookup the IP associated with a domain name but if the signers have the IP fixed then that doesn't happen. This turns out to be a non-issue unless every seed address happens to have their IP changed at the same time or something silly like that.

signer/src/network/libp2p/utils.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        let port = if let Some(p) = self.port() {
+            p
+        } else {
+            DEFAULT_P2P_PORT
+        };
+
+        let mut multiaddrs: Vec<Multiaddr> = Vec::new();
+
+        if let Ok(addrs) = format!("{host}:{port}").to_socket_addrs() {
```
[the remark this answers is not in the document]
djordon: I figured it out, basically, my suggestion doesn't allow us to support domain names.

djordon (commented):
This basically looks good! My only issue is with the `TryIntoMultiAddrs` trait. I'm leaning toward using multi-addresses in the config, but even if we don't go down that route I think that we should change that trait.

signer/src/network/libp2p/utils.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+impl TryIntoMultiAddrs for Url {
+    fn try_into_multiaddrs(&self) -> Result<Vec<Multiaddr>, SignerSwarmError> {
+        eprintln!("url: {:?}", self);
+        let host = self
+            .host_str()
+            .ok_or(SignerSwarmError::Builder("host cannot be empty"))?;
+
+        if !["tcp", "quic-v1"].contains(&self.scheme()) {
```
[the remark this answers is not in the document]
cylewitruk: They're really the only relevant transport protocols (tcp is tcp, quic-v1 is udp but with transport, security and multiplexing built-in, so might as well use it if udp is allowed)

signer/src/network/libp2p/utils.rs
```diff
@@ -0,0 +1,94 @@
...
+            return Err(SignerSwarmError::Builder(
+                "Only `tcp` and `quic-v1` schemes are supported",
+            ));
+        }
+
+        let port = if let Some(p) = self.port() {
+            p
+        } else {
+            DEFAULT_P2P_PORT
+        };
```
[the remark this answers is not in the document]
cylewitruk: Agreed, it's incorrect to assume here that the default should be anything, will change it.

signer/src/network/libp2p/utils.rs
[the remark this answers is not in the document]
cylewitruk: Path is validated and returns an error, username + password should be added to that validation. Will fix.

signer/src/network/libp2p/utils.rs
cylewitruk: > If we deserialize straight into a MultiAddr I don't think that we'd have any network requirements when loading the config.

Yeah, this is right, I checked the `FromStr` code and it doesn't perform DNS lookups at this stage.

>Well, I get how people are more familiar with URLs, but something feels off with using them instead of multi-addresses.

I feel the opposite :sweat_smile: I'm generally against exposing multiaddrs in the config because:
1. They are unconventional
2. They require much more explanation in the config docs (users have to be explicit about ip4/ip6/dnsaddr)
3. We're exposing what is essentially an implementation detail of libp2p in the config (we wouldn't be using multiaddrs if we were using a `TcpListener`, for example).

> My first thought was that having the domain resolve into a static IP is an issue. The IPs associated with a domain name can change, so the IP stored locally would be incorrect. I think clients should have some expiry that requires it to re-lookup the IP associated with a domain name but if the signers have the IP fixed then that doesn't happen. This turns out to be a non-issue unless every seed address happens to have their IP changed at the same time or something silly like that. 

This is what the `identify/push` protocol you were asking about before is for, but in general yeah, unless ALL of the connected peers suddenly change their IP, then this isn't really an issue.

djordon: Okay, I'm cool with this way.

I think we should make a few further chain to how the trait works then. Instead of returning returning multiple multi-addresses, I think we should always return one. For FQDNs in the URI we should just convert that to a FQDN multi-address.

Basically, I think we should do what [`libp2p::core::multiaddr::from_url_lossy`](https://docs.rs/libp2p/0.54.1/libp2p/multiaddr/fn.from_url.html) does, except we accept or reject different protocols. It also makes it clear what's going on, we decided to have the "address" represented as a regular [RFC 8820](https://datatracker.ietf.org/doc/html/rfc8820)[^1] URI and we map a subset of those to multi-addresses internally.

[^1]: I just took the latest RFC here, but yeah it's the regular URI that people are familiar with.

cylewitruk: > I think we should make a few further changes to how the trait works then. Instead of returning returning multiple multi-addresses, I think we should always return one. For FQDNs in the URI we should just convert that to a FQDN multi-address.

Yeah I can agree with this, I didn't actually know about the `dns4` and `dnsaddr` protocols when I originally wrote this -- I enabled DNS for the swarm so it should be able to resolve those itself.

And with that we can just use `Multiaddr` on the config struct and use a custom parser there (string -> url -> multiaddr), and get rid of the trait alltogether :+1:

cylewitruk: Alright, fixed the test issue with a "hack" in [e17719e](https://github.com/stacks-network/sbtc/pull/473/commits/e17719e79709acd2fa80ccd417e75795e67ee538).

signer/src/testing/network.rs
```diff
@@ -45,9 +45,17 @@ fn spawn_client_task(
...
+            assert!(should_receive.contains(&received));
         }
+
+        // !! This is the original implementation, but it assumes exact ordering of
+        // !! messages, which is not guaranteed in the network (this makes tests
+        // !! using libp2p's gossipsub flakey).
+        // for msg in should_receive {
+        //     let received = client.receive().await.expect("Failed to receive message");
+        //     assert_eq!(received, msg);
+        // }
```
djordon, line 59: Very useful comment. I would delete the commented out code here and move the comment about how we cannot assume exact ordering of messages to the modified code above.

signer/src/network/libp2p/utils.rs
djordon, line 1: Isn't this file unnecessary now?

signer/src/network/mod.rs
```diff
@@ -41,4 +138,68 @@ mod tests {
...
         testing::network::assert_clients_can_exchange_messages(client_1, client_2).await;
     }
+
+    #[tokio::test]
+    async fn two_clients_should_be_able_to_exchange_messages_given_a_libp2p_network() {
+        clear_env();
+
+        let _ = tracing_subscriber::fmt()
+            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+            .try_init();
```
djordon, line 148: I don't think we should have this in the tests, since I believe it installs a global subscriber. I mean, I do this locally when developing but remove it in the PR. It's only an issue if we have any other tests with logging turned on though.

signer/src/main.rs
```diff
@@ -68,48 +110,88 @@ async fn run_shutdown_signal_watcher(ctx: &impl Context) -> Result<(), Error> {
...
-    let mut signal = ctx.get_signal_receiver();
+    tracing::info!("initializing the p2p network");
 
-    // TODO(409): Add libp2p swarm initialization here.
+    // Convert the listen `Url`s from the config into `Multiaddr`s.
+    tracing::debug!("parsing listen addresses");
+    let mut listen_addrs: Vec<Multiaddr> = Vec::new();
+    for addr in ctx.config().signer.p2p.listen_on.iter() {
+        listen_addrs.push(addr.clone());
+    }
```
djordon, line 168: I think this was just missed after the recent changes. But all of this can be simplified by just cloning the entire `listen_on` field, and the comment is off. Same thing for the `seed_addrs` variable below.

djordon (commented):
This looks good. I just have a few nits after the latest changes.

signer/src/testing/network.rs
cylewitruk: Yeah the comment was more for you guys (reviewers) to understand why I changed it. Updated in [7173ec1](https://github.com/stacks-network/sbtc/pull/473/commits/7173ec1c34c181344ea16407bb81ccf4e4c260ad)

signer/src/network/libp2p/utils.rs
cylewitruk: Haha yup, miss. Removed in [7173ec1](https://github.com/stacks-network/sbtc/pull/473/commits/7173ec1c34c181344ea16407bb81ccf4e4c260ad)

signer/src/network/mod.rs
cylewitruk: Agreed (though it is nice to have when debugging why something went to crap). The global subscriber is only an issue with `cargo test` btw, `cargo nextest` makes this a non-issue (hint hint :wink: )

Removed in [7173ec1](https://github.com/stacks-network/sbtc/pull/473/commits/7173ec1c34c181344ea16407bb81ccf4e4c260ad)

djordon: lol, I'm sold on `nextest`. I'll ask in the channel, I don't anticipate any objections (I don't know why I haven't already, maybe you already did but didn't get much feedback?).

djordon (approved):
This is really great work, 🙏🏽 thank you for all of this.

signer/src/network/mod.rs
cylewitruk: > maybe you already did but didn't get much feedback?

Yeah I mentioned it in the channel but it was pretty much :cricket:s iirc :sweat_smile:

Issue #426, opened 2024-08-16
[Feature]: Add handler for smart contract events from a stacks node


djordon:
# Feature - Add handler for new smart contract events from a stacks node

## 1. Description

The signers should add an endpoint that their stacks node sends webhooks to for smart contract events. Write sBTC-relevant transactions with their associated events to the database.

### 1.1 Context & Purpose

Each `accept-withdrawal-request` contract call uses a `request-id` as input into the function call. This `request-id` is returned as the contract call output, and is emitted as a print event during the contract call. These events (print events and probably the return events) seem to be the only way to get the `request-id` associated with the contract call. Since these events are emitted as webhooks by the stacks node, we need to add a handler to listen for them.

Note 1: this `request-id` is used to store the input variables used in the `initiate-withdrawal-request` contract call made by the user. It also serves as a way to ensure that both `withdrawal-accept` and `withdrawal-reject` contract calls cannot succeed for the same initial `initiate-withdrawal-request` contract call.

Note 2: I think these events are the only way to know if a transaction failed, since the transaction is still stored on the blockchain, but it's success or failure is not.

### 2.1 Acceptance Criteria:
   - [ ] We have an endpoint that listens to smart contract events from the stack node.
   - [ ] We write the relevant transactions and their events to the database.

cylewitruk:
> Note 2: I think these events are the only way to know if a transaction failed, since the transaction is still stored on the blockchain, but it's success or failure is not.

This is correct

Issue #39, opened 2024-04-05
[Research]: Signer Coordinator Selection and Failover


8marz8:
***Completing the issue description and arriving at a conclusion is the deliverable of this issue.***

# Research - Signer Coordinator Selection and Failover

This ticket holds the research relating to Signer Coordinator Selection and Failover and how it impacts sBTC V1.

## 1. Summary

Stemming from issue https://github.com/Trust-Machines/sbtc-v1/issues/37, we concluded that it's best to have a VRF coordinator selection executed within the signers. It seems that we can mostly leverage the current coordinator code in stacks_signer with a slight difference of using the Bitcoin ConsensusHash that the user's request is tied to instead of the signer's view of the Bitcoin ConsensusHash from burnchain tip. 

## 2. Context & Relevance

The need for a coordinator in signers exists in both the Nakamoto and sBTC releases as the introduction of signers as pivotal entities to process and approve certain operations requires synchronization and collective consensus. 
In context of Nakamoto release, the signers run a DKG round to collectively agree on signing or rejecting blocks (using the DKG aggregate key). A coordinator is needed to trigger and finalize these commands on behalf of all the signers. 
In context of sBTC, signers act as collective actors to govern the peg mechanism. Hence, they need to collectively agree on honoring or rejecting the peg request.

In Nakamoto, there were some iterations on how to calculate the coordinator:
1. Pick the 0th signer as the coordinator. 
2. Introduce a VRF selection to randomly but deterministically choose a signer by calling the `/v2/info` RPC endpoint to fetch the `ConsensusHash` for the latest stacks tip to hash with signer public keys and pick the smallest corresponding ID.
3. The above approach created issues when more end-to-end testing was done as the fast spinning of signers and chain boot-up would cuz different signers to have a different view of the Stacks chain and thus calculate a different coordinator. This would halt the tests (and possibly real production) as DKG would never get calculated. Burnchain ConsensusHash was used as an alternative to pick an infrequent value but this also didn't resolve the issue.
[...]

PR #474, opened 2024-08-30
feat: add handler for smart contract events

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/426. This wraps up the handler for stacks print events.

## Changes

* Add the network to the list of options in the signer's config file.
* Have the private key in the config file to be the generic signer private key.
* Add new tables for our print events from the `sbtc-registry` smart contract.
* Add functions for writing the above print events to the database.
* Add test fixtures for the various print events and unit tests that use them.

## Testing Information

This includes a unit test that checks the deserialization of the webhooks emitted from stacks core. This payload was sent from a stacks core node under Nakamoto on commit https://github.com/stacks-network/stacks-core/commit/cbf0c52fb7a0b8ac55badadcf3773ca0848a25cf which was made on 2024-08-20. The raw body of the webhook was stored as the test fixtures added in this PR. These events were triggered by running test cases in the `complete_deposit_wrapper_tx_accepted` integration test. The stacks node was configured to emit events like so:

```toml
[[events_observer]]
endpoint = "host.docker.internal:8801"
retry_count = 3
include_data_events = false
events_keys = [
    "SN2V7WTJ7BHR03MPHZ1C9A9ZR6NZGR4WM8HT4V67Y.sbtc-registry::print",
]
```

This testing setup followed the same testing setup as in https://github.com/stacks-network/sbtc/pull/303 and https://github.com/stacks-network/sbtc/pull/442. 

## Checklist:

### DIFF
--- a/signer/src/api/new_block.rs
+++ b/signer/src/api/new_block.rs
@@ -72,6 +72,8 @@ where
         };
         // If we got an error writing to the database, this might be an
         // issue that will resolve itself if we try again in a few moments.
+        // So we return a non success status code so that the node retries
+        // in a second.
         if let Err(err) = res {
             tracing::error!("Got an error when writting event to database: {err}");
             return StatusCode::INTERNAL_SERVER_ERROR;
@@ -85,30 +87,33 @@ where
 mod tests {
     use super::*;
 
+    use bitcoin::OutPoint;
     use test_case::test_case;
 
     use crate::config::Settings;
     use crate::storage::in_memory::Store;
 
-    /// This was generated from an actual stacks node after running the
-    /// "complete-deposit standard recipient" variant of the
-    /// `complete_deposit_wrapper_tx_accepted` integration test.
-    const RAW_COMPLETED_DEPOSIT_WEBHOOK: &str =
+    /// These were generated from a stacks node after running the
+    /// "complete-deposit standard recipient", "accept-withdrawal",
+    /// "create-withdrawal", and "reject-withdrawal" variants,
+    /// respectively, of the `complete_deposit_wrapper_tx_accepted`
+    /// integration test.
+    const COMPLETED_DEPOSIT_WEBHOOK: &str =
         include_str!("../../tests/fixtures/complete-deposit-event.json");
 
-    const RAW_WITHDRAWAL_ACCEPT_WEBHOOK: &str =
+    const WITHDRAWAL_ACCEPT_WEBHOOK: &str =
         include_str!("../../tests/fixtures/withdrawal-accept-event.json");
 
-    const RAW_WITHDRAWAL_CREATE_WEBHOOK: &str =
+    const WITHDRAWAL_CREATE_WEBHOOK: &str =
         include_str!("../../tests/fixtures/withdrawal-create-event.json");
 
-    const RAW_WITHDRAWAL_REJECT_WEBHOOK: &str =
+    const WITHDRAWAL_REJECT_WEBHOOK: &str =
         include_str!("../../tests/fixtures/withdrawal-reject-event.json");
 
-    #[test_case(RAW_COMPLETED_DEPOSIT_WEBHOOK, |db| db.completed_deposit_events.get(&bitcoin::OutPoint::null()).is_none(); "completed-deposit")]
-    #[test_case(RAW_WITHDRAWAL_CREATE_WEBHOOK, |db| db.withdrawal_create_events.get(&1).is_none(); "withdrawal-create")]
-    #[test_case(RAW_WITHDRAWAL_ACCEPT_WEBHOOK, |db| db.withdrawal_accept_events.get(&1).is_none(); "withdrawal-accept")]
-    #[test_case(RAW_WITHDRAWAL_REJECT_WEBHOOK, |db| db.withdrawal_reject_events.get(&2).is_none(); "withdrawal-reject")]
+    #[test_case(COMPLETED_DEPOSIT_WEBHOOK, |db| db.completed_deposit_events.get(&OutPoint::null()).is_none(); "completed-deposit")]
+    #[test_case(WITHDRAWAL_CREATE_WEBHOOK, |db| db.withdrawal_create_events.get(&1).is_none(); "withdrawal-create")]
+    #[test_case(WITHDRAWAL_ACCEPT_WEBHOOK, |db| db.withdrawal_accept_events.get(&1).is_none(); "withdrawal-accept")]
+    #[test_case(WITHDRAWAL_REJECT_WEBHOOK, |db| db.withdrawal_reject_events.get(&2).is_none(); "withdrawal-reject")]
     #[tokio::test]
     async fn test_events<F>(body_str: &str, func: F)
     where
--- a/signer/src/api/new_block.rs
+++ b/signer/src/api/new_block.rs
@@ -6,12 +6,12 @@ use axum::extract::rejection::JsonRejection;
 use axum::extract::State;
 use axum::http::StatusCode;
 use axum::Json;
-use blockstack_lib::chainstate::stacks::TransactionPayload;
 
 use crate::stacks::events::RegistryEvent;
 use crate::stacks::webhooks::NewBlockEvent;
-use crate::storage::postgres::PgStore;
-// use crate::storage::DbWrite;
+use crate::storage::DbWrite;
+
+use super::ApiState;
 
 /// We denote the stacks node payload by this type so that our handler
 /// always gets to handle the request, regardless of whether there is a
@@ -35,10 +35,15 @@ pub type StacksNodePayload = Result<Json<serde_json::Value>, JsonRejection>;
 /// second might succeed, we will return a 200 OK status code.
 ///
 /// [^1]: <https://github.com/stacks-network/stacks-core/blob/09c4b066e25104be8b066e8f7530ff0c6df4ccd5/testnet/stacks-node/src/event_dispatcher.rs#L317-L385>
-pub async fn new_block_handler(_state: State<PgStore>, body: String) -> StatusCode {
+pub async fn new_block_handler<S>(state: State<ApiState<S>>, body: String) -> StatusCode
+where
+    S: DbWrite,
+{
     tracing::info!("Received a new block event from stacks-core");
+    let api = state.0;
+    let network = api.settings.signer.network;
 
-    let mut _event: NewBlockEvent = match serde_json::from_str(&body) {
+    let new_block_event: NewBlockEvent = match serde_json::from_str(&body) {
         Ok(value) => value,
         // If we are here, then we failed to deserialize the webhook body
         // into the expected type. It's unlikely that retying this webhook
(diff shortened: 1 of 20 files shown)


signer/src/api/new_block.rs
```diff
@@ -32,39 +26,119 @@ pub type StacksNodePayload = Result<Json<serde_json::Value>, JsonRejection>;
...
-        }
+    let new_block_event: NewBlockEvent = match serde_json::from_str(&body) {
+        Ok(value) => value,
         // If we are here, then we failed to deserialize the webhook body
         // into the expected type. It's unlikely that retying this webhook
         // will lead to success, so we log the error and return `200 OK` so
         // that the node does not retry the webhook.
         Err(error) => {
-            tracing::error!("could not deserialize POST /new_block webhook: {error:?}");
+            tracing::error!("could not deserialize POST /new_block webhook: {body}, error {error}");
```
cylewitruk, line 44: I think this is fine for now since realistically this shouldn't happen (unless Stacks makes a release that completely borks the json schema... totally possible :sweat_smile: ), but in reality this is probably a _fatal_ situation, isn't it? If we're missing blocks then we're potentially missing important sbtc contract calls?

signer/src/api/new_block.rs
```diff
@@ -32,39 +26,119 @@ pub type StacksNodePayload = Result<Json<serde_json::Value>, JsonRejection>;
...
+    #[test_case(COMPLETED_DEPOSIT_WEBHOOK, |db| db.completed_deposit_events.get(&OutPoint::null()).is_none(); "completed-deposit")]
+    #[test_case(WITHDRAWAL_CREATE_WEBHOOK, |db| db.withdrawal_create_events.get(&1).is_none(); "withdrawal-create")]
+    #[test_case(WITHDRAWAL_ACCEPT_WEBHOOK, |db| db.withdrawal_accept_events.get(&1).is_none(); "withdrawal-accept")]
+    #[test_case(WITHDRAWAL_REJECT_WEBHOOK, |db| db.withdrawal_reject_events.get(&2).is_none(); "withdrawal-reject")]
+    #[tokio::test]
+    async fn test_events<F>(body_str: &str, func: F)
+    where
+        F: Fn(tokio::sync::MutexGuard<'_, Store>) -> bool,
+    {
+        let config_path = Some("./src/config/default");
```
cylewitruk, line 122: Maybe we should actually make this a const in `config` (i.e. move the const from `config::tests` that i added to the `config` module with a `#[cfg(test)]` on it), since it seems we'll be using this in multiple places.

signer/src/config.rs
```diff
@@ -187,11 +188,13 @@ impl Validatable for BlockNotifierConfig {
 /// Signer-specific configuration
 #[derive(Deserialize, Clone, Debug)]
 pub struct SignerConfig {
-    /// Stacks account configuration. This is the account that the signer will
-    /// use to identify itself on the network and sign transactions.
-    pub stacks_account: StacksAccountConfig,
+    /// The private key of the signer
```
cylewitruk, line 191: :+1:

signer/src/config/default.toml
```diff
@@ -40,14 +40,22 @@ endpoints = ["http://localhost:20443"]
...
 # Required: true
 private_key = "8183dc385a7a1fc8353b9e781ee0859a71e57abea478a5bca679334094f7adb5"
-
+# Specifies which network to use when constructing and sending transactions
+# on stacks and bitcoin. This cooresponds to the `chain` flag in the
+# bitcoin.conf file of the connected bitcoin-core node, and the
+# `burnchain.mode` flag int he config.toml of the connected stacks-core
+# node.
+#
+# Required: true
```
cylewitruk, line 57: I'd add:
```
# Possible Values: mainnet, testnet, regtest
# Environment: SIGNER_SIGNER__NETWORK
```

signer/tests/integration/contracts.rs
```diff
@@ -114,6 +120,36 @@ impl SignerStxState {
...
     }
+
+    /// Get the source of the a deployed smart contract.
+    ///
+    /// # Notes
+    ///
+    /// This is useful just to know whether a contract has been deployed
+    /// already or not. If the smart contract has not been deployed yet,
+    /// the stacks node returns a 404 Not Found.
+    async fn get_contract_source<T>(&self) -> Result<ContractSrcResponse, reqwest::Error>
```
cylewitruk, line 131: Tbh it wouldn't hurt to have this in the actual `StacksClient`, you're already half-using it

signer/tests/integration/postgres.rs
```diff
@@ -702,3 +706,147 @@ async fn writing_transactions_postgres() {
...
+
+/// Here we test that we can store completed deposit events.
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
+#[tokio::test]
+async fn writing_completed_deposit_requests_postgres() {
+    let db_num = DATABASE_NUM.fetch_add(1, Ordering::SeqCst);
+    let store = signer::testing::storage::new_test_database(db_num).await;
+
+    let mut rng = rand::rngs::StdRng::seed_from_u64(51);
+    let event: CompletedDepositEvent = fake::Faker.fake_with_rng(&mut rng);
```
cylewitruk, line 718: Not something for this PR, but I would suggest moving from `fake` to [proptest](https://crates.io/crates/proptest) [strategies](https://altsysrq.github.io/proptest-book/proptest/tutorial/strategy-basics.html) and writing [proptests](https://altsysrq.github.io/proptest-book/proptest/tutorial/macro-proptest.html) instead. It provides much more robust testing using a multitude of different randomized values w/value shrinking etc. vs. the one random value you get here.

cylewitruk (commented):
Lgtm, just had a few small comments/suggestions

signer/src/api/new_block.rs
djordon: Good idea, will do.

signer/src/api/new_block.rs
djordon: Yeah that makes sense, this shouldn't happen. Although it's more likely that we have some "bug" in our expectation of the schema. If we hit this we'd have to have the stacks node replay the events somehow (no idea how to actually do that).

signer/tests/integration/contracts.rs
djordon: Honestly, I got lazy and didn't want to add the tests that I normally would for a proper function for the `StacksClient` 😅. Plus I wanted to keep this PR focused. But this is a good point, I'll add it in another PR though.

cylewitruk: Haha okay, was just a suggestion anyway, i'll resolve this comment then :)

signer/src/api/new_block.rs
cylewitruk: Something in me feels like there maybe should be some p2p chatter around sbtc events, so a signer (and its peers) can determine if it's lagging/faulty... which should probably result in the signer being ejected from the active set?

signer/tests/integration/postgres.rs
djordon: Yeah, that's a good idea. I'll give it look.

signer/src/api/new_block.rs
djordon: I ended up moving it to the testing module, in case we need it for our integration tests.

signer/src/config/default.toml
djordon: Updated in https://github.com/stacks-network/sbtc/pull/474/commits/57efea14a608dde5e06ab2393a77221ae3b2dde4

signer/src/api/new_block.rs
djordon: > This isn't possible today :( The events are only emitted as a part of block processing. 

That makes sense. Is there a way to have a node forget/delete blocks past a certain height (this is closer to what I had in mind)? If not we'd have to wipe it clean? Can we do that in a "nice" way?

> Something in me feels like there maybe should be some p2p chatter around sbtc events, so a signer (and its peers) can determine if it's lagging/faulty... which should probably result in the signer being ejected from the active set?

The chatter idea might be a good one 🤔. But they probably shouldn't be ejected if they are falling behind, we've just been designing the signers with that possibility in mind.

cylewitruk: > That makes sense. Is there a way to have a node forget/delete blocks after a certain height (this is closer to what I had in mind)? If not we'd have to wipe it clean? Can we do that in a "nice" way?

Unfortunately, no :frowning_face:  The concept of "deleting" data is a bit anti-blockchain, so I can understand why rolling-back a node hasn't been part of the design considerations. I'm sure it'd be possible to implement, but I'm not so sure stacks-core would go along with it.

>The chatter idea might be a good one 🤔. But they probably shouldn't be ejected if they are falling behind, we've just been designing the signers with that possibility in mind.

Are there some design docs I could dig into regarding this design so I can get a better idea how it's handled?

cylewitruk: I'll go ahead and give an approve so you don't get blocked on this discussion, but maybe we should create an issue to track this?

cylewitruk (approved):
Approving so you don't get blocked, but we should probably create an issue to track the "what happens if event-observer deserialization fails" -- if nothing else than that we can land in a documented routine/workaround in the unlikely event of..

signer/src/api/new_block.rs
```diff
@@ -32,39 +26,122 @@ pub type StacksNodePayload = Result<Json<serde_json::Value>, JsonRejection>;
...
+
+    const WITHDRAWAL_REJECT_WEBHOOK: &str =
+        include_str!("../../tests/fixtures/withdrawal-reject-event.json");
+
+    #[test_case(COMPLETED_DEPOSIT_WEBHOOK, |db| db.completed_deposit_events.get(&OutPoint::null()).is_none(); "completed-deposit")]
+    #[test_case(WITHDRAWAL_CREATE_WEBHOOK, |db| db.withdrawal_create_events.get(&1).is_none(); "withdrawal-create")]
+    #[test_case(WITHDRAWAL_ACCEPT_WEBHOOK, |db| db.withdrawal_accept_events.get(&1).is_none(); "withdrawal-accept")]
+    #[test_case(WITHDRAWAL_REJECT_WEBHOOK, |db| db.withdrawal_reject_events.get(&2).is_none(); "withdrawal-reject")]
+    #[tokio::test]
+    async fn test_events<F>(body_str: &str, func: F)
```
netrome, line 122: Nit: `func` is a quite uninformative name. Perhaps we could rename it to `table_is_empty`?

signer/src/testing/dummy.rs
```diff
@@ -225,3 +239,79 @@ fn coinbase_tx<R: rand::RngCore + ?Sized>(
...
+            .take(num_addresses)
+            .map(|kp| {
+                let pk = bitcoin::CompressedPublicKey(kp.public_key());
+                Address::p2wpkh(&pk, Network::Regtest).to_string()
+            })
+            .collect()
+    }
+}
+
+impl<T> fake::Dummy<T> for WithdrawalAcceptEvent {
```
netrome, line 260: I don't think we'd want to do a blanket implementation here. This prevents any other implementations for other fakers. Instead, I'd recommend specifying this to just the `fake::Faker` type:

```suggestion
impl fake::Dummy<fake::Faker> for WithdrawalAcceptEvent {
```

signer/src/testing/dummy.rs
```diff
@@ -225,3 +239,79 @@ fn coinbase_tx<R: rand::RngCore + ?Sized>(
...
+            outpoint: OutPoint {
+                txid: txid(&fake::Faker, rng),
+                vout: rng.next_u32(),
+            },
+            fee: rng.next_u32() as u64,
+        }
+    }
+}
+
+impl<T> fake::Dummy<T> for WithdrawalRejectEvent {
```
netrome, line 276:
```suggestion
impl fake::Dummy<fake::Faker> for WithdrawalRejectEvent {
```

signer/src/testing/dummy.rs
```diff
@@ -225,3 +239,79 @@ fn coinbase_tx<R: rand::RngCore + ?Sized>(
...
+        let bitmap = rng.next_u64() as u128;
+        WithdrawalRejectEvent {
+            txid: StacksTxid(fake::Faker.fake_with_rng(rng)),
+            request_id: rng.next_u32() as u64,
+            signer_bitmap: BitArray::new(bitmap.to_le_bytes()),
+        }
+    }
+}
+
+impl<T> fake::Dummy<T> for WithdrawalCreateEvent {
```
netrome, line 287:
```suggestion
impl fake::Dummy<fake::Faker> for WithdrawalCreateEvent {
```

signer/src/testing/dummy.rs
```diff
@@ -225,3 +239,79 @@ fn coinbase_tx<R: rand::RngCore + ?Sized>(
...
+            amount: rng.next_u32() as u64,
+            sender: StacksAddress::new(version, Hash160(address_hash)).into(),
+            recipient: Address::p2wpkh(&pk, Network::Regtest),
+            max_fee: rng.next_u32() as u64,
+            block_height: rng.next_u32() as u64,
+        }
+    }
+}
+
+impl<T> fake::Dummy<T> for CompletedDepositEvent {
```
netrome, line 306:
```suggestion
impl fake::Dummy<fake::Faker> for CompletedDepositEvent {
```

netrome (approved):
Looks good to me, but I think the Fake implementations can be specific rather than using blanket implementations.

signer/src/api/new_block.rs
djordon: Great idea. Will update.

signer/src/api/new_block.rs
djordon: > I'll go ahead and give an approve so you don't get blocked on this discussion, but maybe we should create an issue to track this?

Yeah, good point, I'll open one for adding metrics and alerts. That's how I would handle this case.

djordon: I opened https://github.com/stacks-network/sbtc/issues/481, but it needs some refining.

signer/src/testing/dummy.rs
djordon: Oh right 🤦🏽 good point.

djordon:
> Approving so you don't get blocked, but we should probably create an issue to track the "what happens if event-observer deserialization fails" -- if for nothing else than that we can land in a documented routine/workaround in the unlikely event of..

Good point, I added https://github.com/stacks-network/sbtc/issues/481 for this. Also, we'll need to break down that ticket further.

signer/src/api/new_block.rs
djordon: > I'll go ahead and give an approve so you don't get blocked on this discussion, but maybe we should create an issue to track this?

There are a few. For this particular question I think the designs are laid out in https://github.com/stacks-network/sbtc/issues/39 and https://github.com/stacks-network/sbtc/issues/44. There are proper google docs too, with their links in different issues. I generally find them using this query https://github.com/stacks-network/sbtc/issues?q=is%3Aissue+design+is%3Aclosed+LLD, their usually the tickets with LLD in the name.

Issue #297, opened 2024-07-05
[Feature]: Validate bitcoin chain tip when handling signer messages


netrome:
### Context
Signer messages include a bitcoin chain tip to represent the state of the bitcoin blockchain when the request was initiated. This ensures that there is a consistent view of the blockchains across the distributed set of signers even if they receive messages out of order.

However, there is currently no validation of the received bitcoin chain tip. The signer should do the following:

1. Ensure that the bitcoin block exists in it's own database.
2. If the message is a coordinator message (i.e. a request to sign a transaction), ensure that it comes from the selected coordinator for that chain tip.
3. Ignore coordinator messages if the chain tip isn't the current canonical chain tip.

### Definition of done
The signers validate the received bitcoin chain tip when handling signer messages.

Issue #465, opened 2024-08-29
[Feature]: [FEATURE_REQESTED]


netrome:
# Feature - Prevent network interference between concurrent coordinators

## 1. Description

In #313 the transaction coordinator functionality for coordinating signing rounds was implemented. However, this implementation is sensitive to multiple coordinators acting concurrently, as the coordinator assumes all WSTS messages on the network relates to it's own signing round.

To make the coordinator more robust, we should ensure the coordinator filters out any messages with another bitcoin chain tip than the one the coordinator is working on. We should also introduce a test case to verify that coordinators can operate concurrently on the same network.

### 2.1 Acceptance Criteria:
   - [ ] Multiple coordinators are able to coordinate signing rounds without interfering with each other.
   - [ ] The above behavior is tested.

## 3. Related Issues and Pull Requests (optional):
#313

netrome:
Part of this is done in https://github.com/stacks-network/sbtc/pull/480, see comment https://github.com/stacks-network/sbtc/pull/480#discussion_r1742467926

aldur:
@djordon reports that the code has been implemented but there is no test yet.

PR #480, opened 2024-09-02
feat: Bitcoin chain tip validation in transaction signer and coordinator

netrome:
Closes: #297 

## Checklist:

### DIFF
--- a/signer/src/testing/transaction_signer.rs
+++ b/signer/src/testing/transaction_signer.rs
@@ -13,6 +13,7 @@ use crate::network;
 use crate::storage;
 use crate::storage::model;
 use crate::testing;
+use crate::transaction_coordinator;
 use crate::transaction_signer;
 
 use crate::ecdsa::SignEcdsa as _;
@@ -331,19 +332,33 @@ where
         let test_data = self.generate_test_data(&mut rng);
         Self::write_test_data(&test_data, &mut handle.storage).await;
 
-        let coordinator_private_key = PrivateKey::new(&mut rng);
-
-        let transaction_sign_request = message::BitcoinTransactionSignRequest {
-            tx: testing::dummy::tx(&fake::Faker, &mut rng),
-            aggregate_key: dummy_aggregate_key,
-        };
-
         let chain_tip = handle
             .storage
             .get_bitcoin_canonical_chain_tip()
             .await
             .expect("storage failure")
             .expect("no chain tip");
+
+        let coordinator_public_key = transaction_coordinator::coordinator_public_key(
+            &chain_tip,
+            &signer_info.first().unwrap().signer_public_keys,
+        )
+        .expect("failed to compute coordinator public key")
+        .unwrap();
+
+        let coordinator_private_key = signer_info
+            .iter()
+            .find(|signer_info| {
+                PublicKey::from_private_key(&signer_info.signer_private_key)
+                    == coordinator_public_key
+            })
+            .unwrap()
+            .signer_private_key;
+
+        let transaction_sign_request = message::BitcoinTransactionSignRequest {
+            tx: testing::dummy::tx(&fake::Faker, &mut rng),
+            aggregate_key: dummy_aggregate_key,
+        };
 
         run_dkg_and_store_results_for_signers(
             &signer_info,
@@ -464,7 +479,6 @@ where
         let mut rng = rand::rngs::StdRng::seed_from_u64(46);
         let network = network::in_memory::Network::new();
         let signer_info = testing::wsts::generate_signer_info(&mut rng, self.num_signers);
-        let coordinator_signer_info = signer_info.first().unwrap().clone();
 
         let mut event_loop_handles: Vec<_> = signer_info
             .clone()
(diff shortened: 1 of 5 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -344,6 +344,11 @@ where
         loop {
             let msg = self.network.receive().await?;
 
+            if msg.bitcoin_chain_tip.as_byte_array() != bitcoin_chain_tip.as_slice() {
+                tracing::warn!(msg=?msg, "concurrent wsts signing round message observed");
+                continue;
+            }
```
djordon, line 350: I'm not sure that this is correct (well, it doesn't seem incorrect either). What happens if there is a bitcoin block that lands in the middle of a signing round? Wait, I think this issue is basically encapsulated by https://github.com/stacks-network/sbtc/issues/465, so nothing obvious to change here.

signer/src/testing/transaction_signer.rs
```diff
@@ -331,20 +332,34 @@ where
...
+        .unwrap();
+
+        let coordinator_private_key = signer_info
+            .iter()
+            .find(|signer_info| {
+                PublicKey::from_private_key(&signer_info.signer_private_key)
+                    == coordinator_public_key
+            })
+            .unwrap()
+            .signer_private_key;
```
djordon, line 356: It's probably worth making a proper type that wraps `Vec<SignerInfo>` so that we can have functions for these common operations. No need to do anything in the PR though.

djordon (approved):
Looks good. The only substantial comment I have is related to https://github.com/stacks-network/sbtc/issues/465.

signer/src/transaction_signer.rs
```diff
@@ -678,6 +747,19 @@ where
...
+    sender_is_coordinator: bool,
+    chain_tip_status: ChainTipStatus,
+}
+
+#[derive(Debug, Clone, Copy)]
+enum ChainTipStatus {
+    Canonical,
+    Known,
+    Unknown,
+}
```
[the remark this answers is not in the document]
netrome: Sure thing, done in https://github.com/stacks-network/sbtc/pull/480/commits/be6998bddb94dcbbb911726177ac2eed0a9e2312

signer/src/transaction_coordinator.rs
```diff
@@ -344,6 +344,11 @@ where
         loop {
             let msg = self.network.receive().await?;
 
+            if msg.bitcoin_chain_tip.as_byte_array() != bitcoin_chain_tip.as_slice() {
+                tracing::warn!(msg=?msg, "concurrent wsts signing round message observed");
```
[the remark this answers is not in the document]
netrome: Right, fixed in https://github.com/stacks-network/sbtc/pull/480/commits/be6998bddb94dcbbb911726177ac2eed0a9e2312

signer/src/transaction_coordinator.rs
netrome: I suspect that this actually solves #465, but we should add tests for it in the context of #465 and ensure there aren't any edge cases we're missing here, and that the system works as intended. The following scenarios should be okay at least:

### Scenario 1
1. Coordinator A: Begins signing round
2. A wild bitcoin block appears
3. Coordinator B != A: Begins signing round

Coordinator A will likely not succeed because the signers would ignore it's messages as stale (although we would tolerate the scenario when it does), but Coordinator B would just ignore any messages relating to coordinator A:s signing round and should be able to continue do it's work without any problems.

Note that even though Coordinator A will hang for a while, the transaction signer task within the same binary will still be responsive to the messages of Coordinator B.

### Scenario 2
1. Coordinator A: Begins signing round
2. A wild bitcoin block appears
3. Coordinator B == A is coordinator for the new block.

Because this coordinator will likely hang and time out, we will have a delay in processing this block. But once the timeout is reached, the coordinator should be able to continue coordinating signing rounds as usual.

netrome: ### Scenario three
1. Two competing forks exist but not all blocks are relayed to all signers, some signers follow one fork other signers follow another fork.
2. Coordinator A and B simultaneously try to coordinate a signing round.

Since they ignore messages from other block hashes, they should not interfere with each other here which I think is the correct behavior in this scenario. This could lead to a temporary stalemate if the partition is somewhat equal between the two sides.

signer/src/testing/transaction_signer.rs
netrome: Makes sense. I can do it as a small follow-up if I find the time.

signer/src/transaction_coordinator.rs
djordon: All of those scenarios make sense to me. 

Okay, let me say the full context:
1. One of the signers is the coordinator and the coordinator is the only signer that can lead signing rounds for Stacks and BTC transactions.
2. Coordinators are the ones that initiate most communication, so if they receive a message it's probably a response from something that the coordinator initiated. Right now, everything else is dropped.
3. If a coordinator ceases to be the proper coordinator mid signing, they will miss signing requests from the new coordinator. This request might be a request that they would sign too (assuming their view of the blockchain has updated as well). The new coordinator now assumes the old one rejected the sign request.
4. The system still functions since only one signer is not fully "fulfilling its duties". Everything else should work just fine.

The part that I have the least clarity on is part (2). What communication does the coordinator drop during a signing round? How consequential is this information? I think the answer is that it they drop sign request messages only. Is that right?

netrome: > If a coordinator ceases to be the proper coordinator mid signing, they will miss signing requests from the new coordinator. This request might be a request that they would sign too (assuming their view of the blockchain has updated as well). The new coordinator now assumes the old one rejected the sign request.

Just to clarify, the coordinator doesn't respond to signing requests - the signer does. So even if the transaction coordinator process hangs for a while, the same binary will have a transaction signer process (with the same keys as the coordinator process) running which will still be responsive to requests from other coordinators.

> The part that I have the least clarity on is part (2). What communication does the coordinator drop during a signing round? How consequential is this information? I think the answer is that it they drop sign request messages only. Is that right?

The coordinator isn't much of a listener and ignores most messages. When a coordinator is coordinating a signing round, it ignores all messages which are not wsts messages from other signers for the given chain tip. That is because these are the only messages providing the information the coordinator needs to produce a valid signature.

djordon: Ahhh right, that makes sense.

PR #482, opened 2024-09-03
feat: add config for event observer

cylewitruk:
## Description

Adds config support for the event observer. 

## Changes

Adds a new configuration section under `signer`: `event_observer`, with two properties: `bind` and `port`.

```toml
# !! ==============================================================================
# !! Stacks Event Observer Configuration
# !! ==============================================================================

# The event observer listens for events on the Stacks blockchain. The listen 
# address must be reachable by your Stacks node, and must be configured in the 
# node's `event_observer` configuration section.
#
# This endpoint _does not_ support TLS and are served over HTTP.
[signer.event_observer]
# The network interface (ip address) to bind the event observer server to.
# 
# Format: "<ip>"
# Required: true
# Environment: SIGNER_SIGNER__EVENT_OBSERVER__BIND
bind = "0.0.0.0"

# The port to bind the event observer server to (on the above bind interface).
#
# Format: <port>
# Required: true
# Environment: SIGNER_SIGNER__EVENT_OBSERVER__PORT
port = 8801
```

## Testing Information

Added relevant asserts to the `load_default_config` test.

## Checklist:

### DIFF
--- a/signer/src/config.rs
+++ b/signer/src/config.rs
@@ -210,10 +210,8 @@ impl Validatable for SignerConfig {
 /// Configuration for the Stacks event observer server (hosted within the signer).
 #[derive(Debug, Clone, Deserialize)]
 pub struct EventObserverConfig {
-    /// The address to bind the server to.
-    pub bind: std::net::IpAddr,
-    /// The port to bind the server to.
-    pub port: u16,
+    /// The address and port to bind the server to.
+    pub bind: std::net::SocketAddr,
 }
 
 impl Settings {
@@ -355,7 +353,7 @@ where
 
 #[cfg(test)]
 mod tests {
-    use std::net::IpAddr;
+    use std::net::SocketAddr;
 
     use super::*;
 
@@ -397,9 +395,8 @@ mod tests {
         assert_eq!(settings.signer.network, NetworkKind::Regtest);
         assert_eq!(
             settings.signer.event_observer.bind,
-            "0.0.0.0".parse::<IpAddr>().unwrap()
+            "0.0.0.0:8801".parse::<SocketAddr>().unwrap()
         );
-        assert_eq!(settings.signer.event_observer.port, 8801);
     }
 
     #[test]
--- a/signer/src/config.rs
+++ b/signer/src/config.rs
@@ -68,17 +68,17 @@ pub struct Settings {
 pub struct P2PNetworkConfig {
     /// List of seeds for the P2P network. If empty then the signer will
     /// only use peers discovered via StackerDB.
-    #[serde(deserialize_with = "url_deserializer")]
+    #[serde(deserialize_with = "url_deserializer_vec")]
     pub seeds: Vec<url::Url>,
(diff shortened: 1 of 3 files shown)


signer/src/main.rs
```diff
@@ -121,9 +122,12 @@ async fn run_stacks_event_observer(ctx: &impl Context) -> Result<(), Error> {
         .route("/new_block", post(api::new_block_handler))
         .with_state(state);
 
+    let config = ctx.config().signer.event_observer.clone();
+
     // run our app with hyper
     // TODO: This should be read from configuration
```
netrome, line 128: I guess this TODO should be removed now.

signer/src/config.rs
```diff
@@ -205,6 +207,15 @@ impl Validatable for SignerConfig {
     }
 }
 
+/// Configuration for the Stacks event observer server (hosted within the signer).
+#[derive(Debug, Clone, Deserialize)]
+pub struct EventObserverConfig {
+    /// The address to bind the server to.
+    pub bind: std::net::IpAddr,
+    /// The port to bind the server to.
+    pub port: u16,
```
djordon, line 216: Why not combine them into one field and deserialize to a `std::net::SocketAddr`?

djordon (commented):
Might as well make the field a `std::net::SocketAddr`, since that is what we need anyway.

signer/src/config.rs
cylewitruk: Good, I thought I had tried that and it wasn't supported, but apparently I tried something else -- changed in [e05f368](https://github.com/stacks-network/sbtc/pull/482/commits/e05f36856afed2f4c25663fc7fe56ff6a3fece7a) :)

signer/src/config/default.toml
```diff
@@ -58,6 +58,31 @@ private_key = "8183dc385a7a1fc8353b9e781ee0859a71e57abea478a5bca679334094f7adb5"
...
+# !! Stacks Event Observer Configuration
+# !! ==============================================================================
+
+# The event observer listens for events on the Stacks blockchain. The listen 
+# address must be reachable by your Stacks node, and must be configured in the 
+# node's `event_observer` configuration section.
+#
+# This endpoint _does not_ support TLS and are served over HTTP.
+[signer.event_observer]
+# The network interface (ip address) to bind the event observer server to.
```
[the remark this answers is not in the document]
cylewitruk: I see a bunch of IPv6 handling in stacks-core, but what that means in practice is an unknown to me :sweat_smile: We'll have to test it -- so I'd prefer to not add any definitive comment on it yet until we know for certain.

signer/src/main.rs
cylewitruk: Haha whooops, done!

signer/src/config/default.toml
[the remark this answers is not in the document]
djordon: Makes sense.

djordon (approved):
Looks good ✅

PR #484, opened 2024-09-04
feat: add support for bitcoin core rpc endpoint configuration

cylewitruk:
## Description

Adds support for configuring bitcoin RPC endpoints, both via config file and environment.

## Changes

- Add support for bitcoin RPC endpoint configuration. This accepts a list, so we should do the same with the `BitcoinCoreClient` as was done with the `StacksClient` and support rotating endpoints on failure. This could be important if the signer wants to avoid any potential downtime during i.e. bitcoin core binary updates, etc.
- Added tests, and added a new convenience method on `Settings` for testing.

## Checklist:


djordon (approved):
Looks good ✅

PR #485, opened 2024-09-04
re-add support for 33-byte stacks private keys

cylewitruk:
## Description

Re-adds support for 33-byte Stacks private keys. 

This is due to the fact that the private key needs to be the same as the one used for stacking so that compensations can be calculated correctly:

> "By securing the sBTC protocol, signers are compensated proportionally to their stacked STX."

This increases the chances that a user will copy-paste a Stacks-generated private key, which contains an extra byte indicating whether or not its public keys should be compressed or not.

## Changes

Add an extra check to the private key deserializer for length and return an appropriate error if not 64 or 66 characters (hex).

## Checklist:

### DIFF
--- a/signer/src/config.rs
+++ b/signer/src/config.rs
@@ -333,13 +333,28 @@ where
     Ok(v)
 }
 
+/// Format an error message for an invalid private key length.
+fn format_invalid_private_key_length_msg(len: usize) -> String {
+    format!(
+        "Invalid Stacks private key length (got {}, expected 64 or 66)",
+        len
+    )
+}
+
 /// A deserializer for the [`PrivateKey`] type. Returns an error if the private
 /// key is not valid hex or is not the correct length.
 fn private_key_deserializer<'de, D>(deserializer: D) -> Result<PrivateKey, D::Error>
 where
     D: Deserializer<'de>,
 {
-    PrivateKey::from_str(&String::deserialize(deserializer)?).map_err(serde::de::Error::custom)
+    let s = String::deserialize(deserializer)?;
+    if [64, 65].contains(&s.len()) {
+        PrivateKey::from_str(&s[..64]).map_err(serde::de::Error::custom)
+    } else {
+        Err(serde::de::Error::custom(
+            format_invalid_private_key_length_msg(s.len()),
+        ))
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


signer/src/config.rs
```diff
@@ -333,13 +333,28 @@ where
...
 where
     D: Deserializer<'de>,
 {
-    PrivateKey::from_str(&String::deserialize(deserializer)?).map_err(serde::de::Error::custom)
+    let s = String::deserialize(deserializer)?;
+    if [64, 65].contains(&s.len()) {
+        PrivateKey::from_str(&s[..64]).map_err(serde::de::Error::custom)
+    } else {
+        Err(serde::de::Error::custom(
+            format_invalid_private_key_length_msg(s.len()),
```
djordon, line 355: Hmmm, I see. You use this function so that you can more easily match on the message in the tests. And you match on the message since this is a user facing error. Also the existing error variant for this, `InvalidPrivateKeyLength`, isn't quite right.

Well, I think it would be better to control error messages from a single place (or at least a single kind of place), which would be the error variant display message. Maybe create new error variants?

djordon (commented):
The main thing is that I think the check for the private key needs to change a little.

signer/src/config.rs
```diff
@@ -333,13 +333,28 @@ where
...
+
 /// A deserializer for the [`PrivateKey`] type. Returns an error if the private
 /// key is not valid hex or is not the correct length.
 fn private_key_deserializer<'de, D>(deserializer: D) -> Result<PrivateKey, D::Error>
 where
     D: Deserializer<'de>,
 {
-    PrivateKey::from_str(&String::deserialize(deserializer)?).map_err(serde::de::Error::custom)
+    let s = String::deserialize(deserializer)?;
+    if [64, 65].contains(&s.len()) {
```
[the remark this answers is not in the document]
cylewitruk: Updated in [d179173](https://github.com/stacks-network/sbtc/pull/485/commits/d1791732d24772a4b7ad07ce2a12e286005b70b8)

I'd be surprised if it ever happens, but now we can inform the user that their weird 33-byte key is wrong :sweat_smile:

signer/src/config.rs
cylewitruk: Fixed in [d179173](https://github.com/stacks-network/sbtc/pull/485/commits/d1791732d24772a4b7ad07ce2a12e286005b70b8)

Tbh I don't think it matters so much when dealing with config error messages since they're generally one-off, i.e. we already use `ConfigError::Message(...)` in this file a fair bit. But now I've added a variant, so we can migrate those messages to the enum as well in a later PR.

djordon (approved):
Looks good

signer/src/config.rs
djordon: 👍🏽 oh right, good point.

Issue #488, opened 2024-09-04
[Feature]: Check the bitcoin block hash in complete-deposit transactions


djordon:
# Feature - Check the bitcoin block hash in `complete-deposit` transactions

## 1. Description

When there is a bitcoin reorg that affects a BTC transaction that moves a user’s deposit, the associated `complete-deposit` that mints sBTC to the user will be orphaned as well. This is all fine, the signers will try again to move the user’s deposited funds and mint sBTC again.

Now, although the `complete-deposit` transaction was orphaned during the reorg, that transaction is still publicly known and contains valid signatures so anyone can resubmit it. The issue is the signers can fail to sweep in the deposited funds (maybe they run out of time, maybe BTC fees spike again).

### 1.1 Context & Purpose

We need to guard against the reorg in the smart contract. One way to do that is to condition the stacks transaction on the existence of a particular block. In clarity, we can do that using the [get-burn-block-info?](https://docs.stacks.co/reference/functions#get-burn-block-info) function. The signers know the block header-hash and height of the block that includes the BTC transaction that sweeps in the users funds, so we can check that the bitcoin block at height X has hash Y, ensuring that a replay of the transaction fails if there is a reorg that affected the BTC transaction.

## 2. Technical Details:
 Modify the smart contracts to expect the bitcoin block hash and block height and check this in the contract call for the `complete-deposit` flow.

### 2.1 Acceptance Criteria:
   - [ ] We guard against reorgs in the smart contract when completing deposits.

PR #493, opened 2024-09-04
feat: check against known burn hash when completing deposit and withdrawals

hstove:
Closes https://github.com/stacks-network/sbtc/issues/488 and closes https://github.com/stacks-network/sbtc/issues/547

This PR updates the `complete-deposits-wrapper` function to take a `burn-height` and `burn-hash`. Then the contract calls `(get-tenure-info? burnchain-header-hash)` to verify that the "known" burn hash at that height matches the argument provided. This allows deposits to be processed right away, while still defending against Bitcoin forks. If Bitcoin forks, these deposit transactions will need to be re-submitted with the new (height, hash) tuple.

I had trouble using `get-burn-block-info?` with Clarinet - I think Clarinet mocks that function and returns pseudo-random information, so I couldn't get a matching hash to work. Using `get-tenure-info?` worked as expected. This required updating our contracts to use Clarity version 3. I also had to push an update to Clarigen to get things working correctly with Clarity v3.


contracts/Clarinet.toml
```diff
@@ -7,28 +7,28 @@ cache_dir = './.cache'
...
-epoch = 2.5
+clarity_version = 3
+epoch = 3.0
 
 [contracts.sbtc-registry]
 path = 'contracts/sbtc-registry.clar'
-clarity_version = 2
-epoch = 2.5
+clarity_version = 3
+epoch = 3.0
```
djordon, line 21: Oh interesting, I didn't realize that clarity v3 was a thing. I didn't see any mention of it in the docs, where can I find out more about this? (Well, I've only looked here https://docs.stacks.co/reference/functions for info in clarity v3)

contracts/contracts/sbtc-deposit.clar
```diff
@@ -54,20 +55,31 @@
     )
 )
 
+(define-read-only (get-burn-header (height uint))
+    ;; (get-burn-block-info? header-hash height)
+    (get-tenure-info? burnchain-header-hash height)
```
djordon, line 60: Oh interesting, I don't know much about `get-tenure-info?`. I was able to track it down in the [source](https://github.com/stacks-network/stacks-core/blob/b09c4ac455363293ad6bd7416fe4bad6a75ed65b/clarity/src/vm/types/mod.rs#L741-L749) but my main question is whether it returns the exact same info as `get-burn-block-info?`. I suppose it probably does, but I couldn't follow the source much to verify.

djordon (commented):
Had a couple of questions, and we also need to update the other one (the `complete-deposit-wrapper` public function).

contracts/Clarinet.toml
cylewitruk: Yeah it's pretty much just an update to include some nakamoto-specific functions like for tenures etc. Don't think the docs are published yet but they will be generated from here:

https://github.com/stacks-network/stacks-core/tree/develop/clarity%2Fsrc%2Fvm%2Fdocs

djordon:
Should we update the `completed-deposit` print event as well?

setzeus:
Reviewing this, just want to make sure I'm making sense here: tests aren't done / might not be possible for the burn hash check since Clarinet is generating pseudo-random info in 'get-burn-block-info?.

contracts/contracts/sbtc-deposit.clar
```diff
@@ -54,20 +55,31 @@
...
 ;; 
 ;; This function handles the validation & minting of sBTC by handling multiple (up to 1000) deposits at a time, 
 ;; it then calls into the sbtc-registry contract to update the state of the protocol. 
-(define-public (complete-deposits-wrapper (deposits (list 650 {txid: (buff 32), vout-index: uint, amount: uint, recipient: principal})))
-    (let 
-        (
-            (current-signer-data (contract-call? .sbtc-registry get-current-signer-data))
-        )
-
+(define-public (complete-deposits-wrapper
```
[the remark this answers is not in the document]
setzeus: Copy

djordon:
> Reviewing this, just want to make sure I'm making sense here: tests aren't done / might not be possible for the burn hash check since Clarinet is generating pseudo-random info in 'get-burn-block-info?.

Yeah I think that is my understanding too. Well, I mean this:
1. Tests aren't done.
2. The original plan was to use `get-burn-block-info?` to do the burn block check, but it's not clear how we could test the smart-contract call because it looks like clarinet generates random values each time that function is called (that's my interpretation, haven't checked myself).
3. If we use `get-tenure-info?` instead of `get-burn-block-info?` then we can test this and get the same functionality. So the changes here use `get-tenure-info?`.

contracts/contracts/sbtc-withdrawal.clar
```diff
@@ -189,12 +194,14 @@
...
-                                     output-index: (optional uint), 
-                                     fee: (optional uint)})))
+                                      {request-id: uint, 
+                                      status: bool, 
+                                      signer-bitmap: uint, 
+                                      bitcoin-txid: (optional (buff 32)), 
+                                      output-index: (optional uint), 
+                                      fee: (optional uint)}))
+                                     (burn-height uint)
+                                     (burn-hash (buff 32)))
```
djordon, line 204: Can we do this for `accept-withdrawal-request` as well?

djordon (commented):
Can we also update the emitted print events to include the burn-block-hash and burn-block-height for both deposits and withdrawals?.

contracts/contracts/sbtc-withdrawal.clar
setzeus: Completed tests & was just looking at this & it requires a bit more refactor that I thought going in. As a result of using a helper to loop through multiple deposits & withdrawals I see two options moving forward, neither of which are great:

1. We update *individual* deposits & withdrawals with a burn-height & burn-hash field (which I don't love since multiple deposits & withdrawals might now have different burn-heights & burn-hashes)
2. We don't include burn-height/burn-hash in individual deposits & withdrawals but instead print them both from the sbtc-deposit & sbtc-withdrawal contracts instead of in the confirmed sbtc-registry contract (which I don't love since we currently have all print statements in sbtc-registry)

djordon: Yeah, I definitely prefer option (1). And there is also option (3): each withdrawal tuple has a `burn-height` and `burn-hash` field. The trade off there is that we need to submit more data when making the contract call, making things more expensive. This is a slight modification on option (1) but I like (1) more.

But I think we can keep the updated signature as is and clarity things in the docs: only make `complete-withdrawals` and `complete-deposits` contract calls where the requests were completed in the same bitcoin block.

hstove:
I just pushed some fixes for the failing CI test. It was two things:

- Some checks used a fixed block height, where you should use `simnet.blockHeight` as a reference for figuring out when a request was made
- There was (I think) a bug in the Clarinet SDK version being used that wasn't allowing `get-tenure-info?` to work. I updated Clarinet SDK and that seems to have fixed the issue.

contracts/contracts/sbtc-withdrawal.clar
setzeus: Updated individual deposits to take in the burn-hash & the burn-height fields, can you confirm that the deposit print statements are what you expect? If so I'll continue with refactoring withdrawal.

djordon: Yup, this looks good.

contracts/contracts/sbtc-deposit.clar
```diff
@@ -46,28 +48,39 @@
...
-(define-public (complete-deposits-wrapper (deposits (list 650 {txid: (buff 32), vout-index: uint, amount: uint, recipient: principal})))
-    (let 
-        (
-            (current-signer-data (contract-call? .sbtc-registry get-current-signer-data))
-        )
-
+(define-public (complete-deposits-wrapper 
+        (deposits (list 650 {txid: (buff 32), vout-index: uint, amount: uint, recipient: principal, burn-hash: (buff 32), burn-height: uint}))
+        (burn-height uint)
+        (burn-hash (buff 32))
```
djordon, line 76: I don't think we need the `burn-height` and `burn-hash` again here on line 75-76.

contracts/contracts/sbtc-registry.clar
```diff
@@ -234,7 +236,9 @@
       topic: "completed-deposit",
       bitcoin-txid: txid,
       output-index: vout-index,
-      amount: amount
+      amount: amount,
+      burn-hash: burn-hash,
+      burn-height: burn-height,
```
djordon, line 241: We should add these same fields for the print event for withdrawals too.

djordon (commented):
Looks good, the only thing we should do is add the `burn-hash` and `burn-height` fields to the withdrawal print event as well.

contracts/contracts/sbtc-registry.clar
setzeus: Done

contracts/contracts/sbtc-withdrawal.clar
```diff
@@ -172,14 +183,14 @@
...
-      (try! (contract-call? .sbtc-registry complete-withdrawal-accept request-id bitcoin-txid output-index signer-bitmap fee))
+      (try! (contract-call? .sbtc-registry complete-withdrawal-accept request-id bitcoin-txid output-index signer-bitmap fee burn-hash burn-height))
 
       (ok true)
   )
 )
 
 ;; Reject a withdrawal request
-(define-public (reject-withdrawal-request (request-id uint) (signer-bitmap uint))
+(define-public (reject-withdrawal-request (request-id uint) (signer-bitmap uint) (burn-hash (buff 32)) (burn-height uint))
```
djordon, line 193: My bad, when I asked for these fields for withdrawals I meant for withdrawal accept only. We don't need to include this for rejects as well.

setzeus: Updated

djordon (approved):
Looks good! ✅ 🚢

Issue #494, opened 2024-09-04
[Feature]: Replace type aliases with newtypes


djordon:
# Feature - Replace type aliases with newtypes

## 1. Description

This ticket is very similar to https://github.com/stacks-network/sbtc/issues/287, but focused on the signer.

There are quite a few aliases that we can just make proper `newtype` types, similar to the internal `PublicKey` type introduced in https://github.com/stacks-network/sbtc/issues/417.

### 1.1 Context & Purpose

This is just a quality of life improvement.

### 2.1 Acceptance Criteria:
   - [ ] We have very few type aliases in the signer codebase.

## 3. Related Issues and Pull Requests (optional):

* https://github.com/stacks-network/sbtc/issues/287

PR #495, opened 2024-09-04
refactor: use newtypes instead of aliases

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/494.

I'm not sure if we should use these newtypes more or less than the types that they wrap.

## Changes

* Add `newtype` types for most of our aliases. The only remaining alias is the `BitcoinAddress` alias, and that one is tricky.
* Moved all sqlx impls to its own module.

## Testing Information

This is just a refactor, so if tests pass then this should be okay.

## Checklist:


signer/src/block_observer.rs
```diff
@@ -521,7 +522,7 @@ mod tests {
             .unwrap();
         let storage = block_observer.storage.lock().await;
         assert_eq!(storage.deposit_requests.len(), 1);
-        let db_outpoint = (tx_setup0.tx.compute_txid().to_byte_array().to_vec(), 0);
+        let db_outpoint = (BitcoinTxId::from(tx_setup0.tx.compute_txid()), 0);
```
cylewitruk, line 525: I personally think that this syntax is more readable:
```rust
let db_outpoint: (BitcoinTxId, u32) = (tx_setup0.tx.compute_txid().into(), 0);
```
Just because it reduces one level of nesting on the right side (and vscode prints out the types on the left anyway); but the way it is is fine, too.

signer/src/storage/in_memory.rs
```diff
@@ -241,7 +241,7 @@ impl super::DbRead for SharedStore {
             .lock()
             .await
             .deposit_request_to_signers
-            .get(&(txid.clone(), output_index))
+            .get(&(*txid, output_index))
```
cylewitruk, line 244: You _could_ implement `Deref` for `BitcoinTxId` and use deref coercion instead (i.e. implicit `*`) -- if you wanted.

cylewitruk (approved):
Lgtm, don't see any real functional changes so this should be just fine -- I just left a couple of optional suggestions :v:

signer/src/storage/model.rs
```diff
@@ -181,25 +180,23 @@ pub struct StacksTransaction {
 #[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
 pub struct TransactionIds {
     /// Transaction IDs.
-    pub tx_ids: Vec<Vec<u8>>,
+    pub tx_ids: Vec<[u8; 32]>,
```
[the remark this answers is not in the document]
djordon: Yeah, the issue here is with the struct itself, which can contain either stacks or bitcoin transaction IDs. The solution would be to create a generic txid alias, a generic TxId type, or a new struct so that we could specialize between the two. I don't like the first two options and the third felt like too much unnecessary work (which this PR kind-of qualifies for already 😅).

signer/src/block_observer.rs
djordon: Good point, will update.

signer/src/storage/in_memory.rs
djordon: Hmmm, yeah I thought so too, but I don't think you can (I think I tried this before and I tried it again just now). Maybe the additional reference of the tuple confuses the compiler. I removed `Deref` since I did not need it but I'm not opposed to adding it back if the want/need arises.

cylewitruk: Ah weird, yeah there's something there that's confusing the compiler anyway

signer/src/storage/model.rs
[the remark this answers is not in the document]
netrome: I guess the overkill alternative is to create a wrapper enum that is either a bitcoin txid or a stacks txid, and can dereference to `[u8; 32]`. But that can definitely be left as a future enhancement.

netrome (commented):
Nice refactor!

signer/src/storage/model.rs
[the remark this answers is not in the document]
djordon: Oh right, I didn't even think of that.

Issue #514, opened 2024-09-12
[Feature]: BTC deposit tx validation returns potentially unvalidated script data


evonide:
# (Low) Bitcoin deposit tx validation stores potentially unvalidated script data

## 1. Description

This concerns [sbtc/sbtc/src/deposits.rs at 9c7b66525c02eeba64f3d6b0aaa4f9bd1a20a62e · stacks-network/sbtc](https://github.com/stacks-network/sbtc/blob/9c7b66525c02eeba64f3d6b0aaa4f9bd1a20a62e/sbtc/src/deposits.rs#L158):
```
    /// Validate this deposit request.
[...]
    /// * That the deposit script and reclaim script in the request match
    ///   the expected formats for deposit transactions.
    /// * That deposit script and the reclaim script are part of the UTXO
    ///   ScriptPubKey.
    pub fn validate_tx(&self, tx: &Transaction) -> Result<DepositInfo, Error> {
[...]
        // Validate that the deposit and reclaim scripts in the request
        // match the expected formats for deposit transactions.
        let deposit = DepositScriptInputs::parse(&self.deposit_script)?;
        let reclaim = ReclaimScriptInputs::parse(&self.reclaim_script)?;
        // Okay, the deposit and reclaim scripts are valid. Now make sure
        // that the ScriptPubKey in the transaction matches the one implied
        // by the given scripts. So now create the expected ScriptPubKey.
        let deposit_script = deposit.deposit_script();
        let reclaim_script = reclaim.reclaim_script();

        debug_assert_eq!(deposit_script, self.deposit_script); // <-- Skipped in production
        debug_assert_eq!(reclaim_script, self.reclaim_script);  // <-- Skipped in production

        let expected_script_pubkey = to_script_pubkey(deposit_script, reclaim_script);
        // Check that the expected scriptPubkey matches the actual public
        // key of our parsed UTXO.
        if expected_script_pubkey != tx_out.script_pubkey {
            return Err(Error::UtxoScriptPubKeyMismatch(self.outpoint));
        }

        Ok(DepositInfo {
            max_fee: deposit.max_fee,
            deposit_script: self.deposit_script.clone(), // <-- Using the initial data instead of `deposit_script`
            reclaim_script: self.reclaim_script.clone(), // <-- Using the initial data instead of `reclaim_script`
            signers_public_key: deposit.signers_public_key,
            recipient: deposit.recipient,
            lock_time: reclaim.lock_time as u64,
[...]

PR #519, opened 2024-09-13
fix: pass on validatated script when validating deposit scripts

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/514

## Changes

* Use the regenerated deposit and reclaim scripts when passing on the validated deposit info.

## Testing Information

There isn't really a new test for this change, since it involves constructing a collision in a merkle tree. So if the existing tests pass then we should be good.

## Checklist:


cylewitruk (approved):
lgtm, not much to say on this one :)

Issue #527, opened 2024-09-13
[Feature]: Store the `ScriptPubKey` as the withdrawal request recipient 


djordon:
# Feature - Store the `ScriptPubKey` as the withdrawal request recipient

## 1. Description

Right now we store the bitcoin address as the recipient in withdrawal requests. This is incorrect, since what we actually get from the user is the `ScriptPubKey`. It also makes things more complex than they need to be in the signer code base.

### 1.1 Context & Purpose

Switch from using the bitcoin address as a the recipient to using the `ScriptPubKey` as the recipient.

### 2.1 Acceptance Criteria:
   - [ ] The database tables are updated to store the `ScriptPubKey`
   - [ ] The internal types use the `ScriptPubKey` instead of the `bitcoin::Address`.

PR #528, opened 2024-09-13
refactor: use the `ScriptPubKey` as recipient

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/527.

## Changes

* Update 'recipient' column names and their type to expect a `bytea` for a `ScriptPubKey`. This makes it more consistent with the code base, as well as reduces overall complexity (using an address is incorrect).
* Use `WithdrawalX` instead of `WithdrawX` for table, function, and struct names.
* Add the transaction ID to the `withdrawal_request` and `withdrawal_signer` tables.
* Add an internal `ScriptPubKey` type for interacting with the database.

## Testing Information

This is a refactor, so if tests pass then we're good.

## Checklist:

### DIFF
--- a/signer/migrations/0003__create_tables.sql
+++ b/signer/migrations/0003__create_tables.sql
@@ -31,7 +31,7 @@ CREATE TABLE sbtc_signer.deposit_requests (
     recipient TEXT NOT NULL,
     amount BIGINT NOT NULL,
     max_fee BIGINT NOT NULL,
-    sender_addresses TEXT[] NOT NULL,
+    sender_script_pub_keys BYTEA[] NOT NULL,
     created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL,
     PRIMARY KEY (txid, output_index)
 );
@@ -46,27 +46,28 @@ CREATE TABLE sbtc_signer.deposit_signers (
     FOREIGN KEY (txid, output_index) REFERENCES sbtc_signer.deposit_requests(txid, output_index) ON DELETE CASCADE
 );
 
-CREATE TABLE sbtc_signer.withdraw_requests (
+CREATE TABLE sbtc_signer.withdrawal_requests (
     request_id BIGINT NOT NULL,
     txid BYTEA NOT NULL,
     block_hash BYTEA NOT NULL,
-    recipient TEXT NOT NULL,
+    recipient BYTEA NOT NULL,
     amount BIGINT NOT NULL,
     max_fee BIGINT NOT NULL,
     sender_address TEXT NOT NULL,
     created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL,
-    PRIMARY KEY (request_id, block_hash),
+    PRIMARY KEY (request_id, txid, block_hash),
     FOREIGN KEY (block_hash) REFERENCES sbtc_signer.stacks_blocks(block_hash) ON DELETE CASCADE
 );
 
-CREATE TABLE sbtc_signer.withdraw_signers (
+CREATE TABLE sbtc_signer.withdrawal_signers (
     request_id BIGINT NOT NULL,
+    txid BYTEA NOT NULL,
     block_hash BYTEA NOT NULL,
     signer_pub_key BYTEA NOT NULL,
     is_accepted BOOL NOT NULL,
     created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL,
-    PRIMARY KEY (request_id, block_hash, signer_pub_key),
-    FOREIGN KEY (request_id, block_hash) REFERENCES sbtc_signer.withdraw_requests(request_id, block_hash) ON DELETE CASCADE
+    PRIMARY KEY (request_id, txid, block_hash, signer_pub_key),
+    FOREIGN KEY (request_id, txid, block_hash) REFERENCES sbtc_signer.withdrawal_requests(request_id, txid, block_hash) ON DELETE CASCADE
 );
 
 CREATE TABLE sbtc_signer.transactions (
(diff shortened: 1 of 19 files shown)


signer/migrations/0003__create_tables.sql
```diff
@@ -46,27 +46,28 @@ CREATE TABLE sbtc_signer.deposit_signers (
...
     txid BYTEA NOT NULL,
     block_hash BYTEA NOT NULL,
-    recipient TEXT NOT NULL,
+    recipient BYTEA NOT NULL,
     amount BIGINT NOT NULL,
     max_fee BIGINT NOT NULL,
     sender_address TEXT NOT NULL,
     created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL,
-    PRIMARY KEY (request_id, block_hash),
+    PRIMARY KEY (request_id, txid, block_hash),
```
cylewitruk, line 58: Just checking -- this is valid because of a reorg situation right? That the same request_id could potentially exist within multiple blocks? But could it really exist within multiple txids within the same block (as a withdrawal in this case)?

signer/src/storage/in_memory.rs
```diff
@@ -34,7 +34,7 @@ pub struct Store {
     pub deposit_requests: HashMap<DepositRequestPk, model::DepositRequest>,
 
     /// Deposit requests
-    pub withdraw_requests: HashMap<WithdrawRequestPk, model::WithdrawRequest>,
+    pub withdraw_requests: HashMap<WithdrawRequestPk, model::WithdrawalRequest>,
```
cylewitruk, line 37: Should we change var names while we're at it? i.e. `withdraw_requests` -> `withdrawal_requests`?

signer/src/storage/mod.rs
```diff
@@ -79,26 +79,26 @@ pub trait DbRead {
...
-    fn get_withdraw_signers(
+    fn get_withdrawal_signers(
         &self,
         request_id: u64,
         block_hash: &model::StacksBlockHash,
-    ) -> impl Future<Output = Result<Vec<model::WithdrawSigner>, Self::Error>> + Send;
+    ) -> impl Future<Output = Result<Vec<model::WithdrawalSigner>, Self::Error>> + Send;
 
     /// Get pending withdraw requests
     fn get_pending_withdraw_requests(
```
cylewitruk, line 89: withdrawal?

signer/src/storage/mod.rs
```diff
@@ -79,26 +79,26 @@ pub trait DbRead {
...
         &self,
         chain_tip: &model::BitcoinBlockHash,
         context_window: u16,
-    ) -> impl Future<Output = Result<Vec<model::WithdrawRequest>, Self::Error>> + Send;
+    ) -> impl Future<Output = Result<Vec<model::WithdrawalRequest>, Self::Error>> + Send;
 
     /// Get pending withdraw requests that have been accepted by at least `threshold` signers and has no responses
     fn get_pending_accepted_withdraw_requests(
```
cylewitruk, line 96: withdrawal?

cylewitruk (commented):
lgtm, just have the one question about the txid in primary key before i hit accept

signer/migrations/0003__create_tables.sql
djordon: Actually, I'm pretty sure that it's unnecessary to include the `txid` here. As in a request should be uniquely identified by the `request_id` and the `block_hash`. So I'm pretty sure that it can be safely reverted back to what it was.

signer/src/storage/in_memory.rs
djordon: 🤦🏽 oh I missed this, and it was right there.

signer/src/storage/mod.rs
djordon: 🤦🏽 missed this one too. will update

signer/migrations/0003__create_tables.sql
djordon: > That the same request_id could potentially exist within multiple blocks?

Yes that is correct, a `request_id` could exist within multiple blocks because of reorgs.

signer/src/storage/mod.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/528/commits/e0b18d6f08b0bf4d53c60e51a463ba670a964c66

signer/src/storage/mod.rs
djordon: https://github.com/stacks-network/sbtc/pull/528/commits/e0b18d6f08b0bf4d53c60e51a463ba670a964c66

signer/src/storage/in_memory.rs
djordon: https://github.com/stacks-network/sbtc/pull/528/commits/e0b18d6f08b0bf4d53c60e51a463ba670a964c66

signer/migrations/0003__create_tables.sql
cylewitruk: Yeah I'd revert it if it's invalid to have multiple in the same block since we lose the unique constraint this way

djordon: I'm not sure I follow, I was saying that `(request_id, block_hash)` is probably unique. If that is unique then `(request_id, txid, block_hash)` is necessarily unique as well.

But yeah I'm kinda okay with reverting, with the reason being that it is probably unnecessary. The main reason why I changed the primary key here was to be consistent with how we identify withdrawal requests elsewhere (like in the receipt that is stored on the bitcoin blockchain). It's easier (mentally) for someone new (or me after a month or so) to understand things if we use the same identifier everywhere for the same concept. Not sure how to balance it all, I'll default to what you think here.

cylewitruk: On phone but I just meant that if
```
request_id, block hash, txid
1                  abc123        1
1                  abc123        2
```
Isn't valid then the PK (or a separate UK) should be on the request id and block, otherwise it opens for multiple txids for the same request and block.

djordon: Ahh gotcha, makes sense.

djordon: https://github.com/stacks-network/sbtc/pull/528/commits/fc05924c824bbbed5c6028c06777c683532b0485

cylewitruk (approved):
Looks good!

Issue #326, opened 2024-07-14
[Feature]: Populate signer bitmap in the `utxo::Deposit/WithdrawRequest` conversions


netrome:
### Context
In #325, two `TryFrom` implementations were added to `utxo.rs` to convert between the storage model of `DepositRequest` and `WithdrawRequest` to the corresponding types in `utxo.rs`. However, the types in `utxo.rs` has a `signer_bitmap` field, which requires additional queries from the database to be able to populate.

### Definition of Done
The `TryFrom` implementations are adjusted so they also popualte the `signer_bitmap` fields.

Note that this means that the input type of the conversion has to be updated. This could for example be a new type with all necessary information, or a tuple of the current model type and a bitvec of signers.

PR #509, opened 2024-09-12
feat: populate signer bitmap in deposit requests

djordon:
## Description

Closes part of https://github.com/stacks-network/sbtc/issues/326.

A left some TODOs without tickets, I apologize. I'll fix that tomorrow.

## Changes

* Add a query that fetches how the signers voted on a deposit.
* Populate the signer bitmap when constructing the `utxo::DepositRequest` struct.

## Testing Information

This PR adds an integration test for the query and a unit test to make sure the bitmap is constructed correctly.

## Checklist:

- [x] I have performed a self-review of my code

PR #529, opened 2024-09-13
feat: populate signer bitmap in withdrawal requests

djordon:
## Description

This closes https://github.com/stacks-network/sbtc/issues/326, the previous one to tackle that issue was https://github.com/stacks-network/sbtc/pull/509.

## Changes

* Populate the signer bitmap in the `utxo::WithdrawalRequest` struct.
* Add queries for fetching how the signers voted for a withdrawal.
* Update the tests so that the generated data ties how the signers voted for a request to a signing set. Previously, the signer's public key for a vote was generated randomly.

## Testing Information

There are new unit tests for the new code, and an integration test for the query.

## Checklist:


signer/src/storage/model.rs
```diff
@@ -156,6 +156,17 @@ pub struct WithdrawalRequest {
     pub sender_address: StacksPrincipal,
 }
 
+impl WithdrawalRequest {
+    /// Return the identifier for the withdrawal request.
+    pub fn request_identifier(&self) -> WithdrawalRequestId {
```
cylewitruk, line 161: i would probably have just called this `request_id()` since that's what it's called everywhere else

signer/src/testing/transaction_coordinator.rs
```diff
@@ -159,11 +165,13 @@ where
             .once()
             .returning(|_| Box::pin(async { Ok(None) }));
 
-        let (broadcasted_tx_sender, mut broadcasted_tx_receiver) = tokio::sync::mpsc::channel(1);
+        // TODO: multiple transactions can be generated and keeping this
```
cylewitruk, line 168: Is this exclusive to test code or is there something bigger that needs to be looked into?

signer/tests/integration/postgres.rs
```diff
@@ -294,7 +308,15 @@ async fn should_return_the_same_pending_deposit_requests_as_in_memory_store() {
         num_withdraw_requests_per_block: 5,
         num_signers_per_request: 0,
     };
-    let test_data = testing::storage::model::TestData::generate(&mut rng, &test_model_params);
+    let signer_info = testing::wsts::generate_signer_info(&mut rng, num_signers);
+    let coordinator_signer_info = signer_info.first().cloned().unwrap();
+    let signer_set: Vec<_> = coordinator_signer_info
```
cylewitruk, line 313: You're doing this quite a few times so I'd probably make some sort of helper

cylewitruk (commented):
Lgtm in general, just a few minor comments

signer/src/testing/transaction_coordinator.rs
djordon: I believe that it is exclusive to the test code. I'm pretty sure that these senders/receivers are only compiled with the testing feature.

djordon: Just double checked, and this channel is local to this particular test, not elsewhere.

signer/tests/integration/postgres.rs
djordon: yeah good point. Will update.

signer/src/storage/model.rs
djordon: Well, I felt that would be confusing, since there is a `request_id` field, and `request_id` here usually refers to the one generated by smart contract. I'm not thrilled with the name here though, so I'm definitely open to suggestions.

djordon: Maybe I should rename the struct too.

cylewitruk: Aha okay I just quickly assumed it was referring to the same request id i see everywhere else, if it's not then yeah i agree this is confusing 😊

cylewitruk: QualifiedRequestId or something along those lines maybe?

djordon: That works! Using that.

signer/tests/integration/postgres.rs
djordon: https://github.com/stacks-network/sbtc/pull/529/commits/a7bbb2f911139beb51bc022a829dd6fce6a6d20c

signer/src/storage/model.rs
djordon: https://github.com/stacks-network/sbtc/pull/529/commits/dafedbe7cd474fb34390e718f5198bffc04ff2ed

cylewitruk: Looks nice!

cylewitruk (approved):
Looks good!

Issue #530, opened 2024-09-14
[Feature]: Store the Stacks block ID for events


djordon:
# Feature - Store the Stacks block ID for events

## 1. Description

Well, we need the Stacks block ID with our events and they are not currently stored.

### 1.1 Context & Purpose

For withdrawals, we will depend on these stacks-core webhook events in order to know about the withdrawal request, and knowing the block ID is key to knowing when the event took place and whether it is still on the canonical stacks blockchain.

### 2.1 Acceptance Criteria:
   - [ ] Add a column to the event tables with the block id
   - [ ] Store the recipient as a `scriptPubKey`, similar to https://github.com/stacks-network/sbtc/issues/527.

## 3. Related Issues and Pull Requests (optional):

https://github.com/stacks-network/sbtc/issues/527

PR #531, opened 2024-09-14
feat: store the stacks block id for events

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/530.

## Changes

* Refactor the events module to use the script pubkey as the recipient instead of the address. This was supposed to be done in https://github.com/stacks-network/sbtc/issues/527.
* Add a block hash column to the event tables that stores the Stacks block ID for sBTC smart contract events.
* Add an error type in the events module to make it more self contained.
* Remove unused error types from the main `Error` type in the signer crate.

## Checklist:


cylewitruk (approved):
lgtm!

side note, i noticed a few places while reviewing where `tracing` is being used with inlined variables which sorta goes against structured logging/tracing -- should we have a "final review" issue to revisit all tracing entries and make sure they're cosher?

djordon:
> side note, i noticed a few places while reviewing where `tracing` is being used with inlined variables which sorta goes against structured logging/tracing -- should we have a "final review" issue to revisit all tracing entries and make sure they're cosher?

You know I thought we had one, but I just searched and it looks like we don't. Yes we should create a ticket for this, I've been assuming that we'd do this.

Issue #532, opened 2024-09-16
[Feature]: Provide database migration mechanism for signer


cylewitruk:
# Feature - Provide database migration mechanism for signer

## 1. Description

We should not make the assumption that a user running a signer is particularly database-inclined.

Users are generally accustomed to most (and especially blockchain) services handling their own database migrations. We should provide a mechanism for the signer to automatically manage its own migrations, but it should not be forced if using an external database.

If the user is using an external database and _chooses_ to run database migrations manually, we should also provide the means for the user to generate the required SQL script (i.e. skipping already-applied migrations).

In addition to user convenience, it also prevents us from needing to separately publish migration scripts and surrounding documentation.

### 1.1 Context & Purpose
See Description

### 2.1 Acceptance Criteria:
   - [ ] Database migrations can OPTIONALLY be applied automatically by the signer binary
   - [ ] Database migrations can be printed by the user (and i.e. dumped to a file for manual execution or piped into the db client command).

PR #533, opened 2024-09-16
feat: db migrations

cylewitruk:
## Description

Introduces db-migration functionality for the signer binary.

**Note 1:** I did first try using `sqlx`'s migration functionality, but it pulls in dependencies which conflict with stacks-core deps. This seems to be documented as a known _Cargo_ issue [here](https://github.com/launchbadge/sqlx/issues/3211). But in any case, handling them manually gives us a few more options, especially re: printing out pending migrations via a command line command.

**Note 2:** I took the db-migrations "as-is" for now. When in a user environment we may want to re-think the two scripts for creating the schema and user as depending on a company's policies these operations may not be allowed.

**Question on Note 2:** Is there any particular reason why we felt we needed to use a separate schema from `public`? I'm going under the assumption that we should require that the sBTC signer has its own dedicated database and in the case of using pgsql the user will probably create their own user and give it appropriate permissions (as designated by us). I otherwise have a hard time envisioning that the signer will grow to a size where schema isolation is desired/required?

Partially Closes: #532 (printing out pending db migrations is not in this PR)

## Changes

- New dependency on `include_dir` (which is similar to `include_bytes` etc. but for entire directories) which compiles the database migrations into the signer binary.
- New **optional** command line flag `--migrate-db` which if set will attempt to apply any pending database migrations on signer startup. Default is **false**.
- `new_test_database()` is updated with a flag for whether or not migrations should be applied. I had this originally for writing integration tests for the migrations themselves, but then decided that the migrations are implicitly tested for any test using this function so it wasn't explicitly needed. I left this flag in as it can be good for adding some negative integration tests (i.e. for when a user isn't using auto-migrations and has forgotten to apply migrations manually, etc.).
- `PgStore` creates a new table `public.__sbtc_migrations ` (I am assuming here that sBTC will have its own, dedicated db and that we can freely use `public`) for keeping track of applied migrations.
- Changed the `flyway` service entry in `docker-compose.test.yml` to not run automatically in favor of db migrations being applied by code. If you use this compose file for local testing then you can still run it, just `up flyway` manually.
- Changed the `integration-test-full` target to `down` the test env first (I ran into some issues here when it was lingering).

## Testing Information

- Integration tests have been updated and all tests pass.
- This also adds a new docker-compose for running the signer locally with a db on a random port (need to fix bitcoin to do the same), with migrations enabled.

## Checklist:


Makefile
```diff
@@ -223,3 +223,14 @@ devenv-up: $(INSTALL_TARGET) $(EMILY_LAMBDA_BINARY) $(EMILY_CDK_TEMPLATE)
...
+run-signer:
+	docker compose --file docker-compose.signer.yml down;
+	docker compose --file docker-compose.signer.yml up postgres bitcoind --detach;
+	POSTGRES_PORT="0"; \
+	while [ "$$POSTGRES_PORT" -le 0 ]; do \
+		sleep 1; \
+		POSTGRES_PORT=$$(docker port sbtc-postgres 5432 | awk -F: '{print $$2}'); \
+	done; \
+	echo $$POSTGRES_PORT; \
+	RUST_LOG=debug SIGNER_SIGNER__DB_ENDPOINT="postgres://postgres:postgres@localhost:$$POSTGRES_PORT/signer" cargo run --bin signer -- -c ./signer/src/config/default.toml --migrate-db
```
djordon, line 236: nit: new line here please.

docker-compose.signer.yml
djordon, line 1: Why add this file instead of modifying the `devenv` one? Funny thing is we used to have this file (it was called `docker-compose.local.yml`, a worse name) but we got rid of it to simplify things.

signer/Cargo.toml
```diff
@@ -73,5 +74,6 @@ more-asserts = "0.3"
 ripemd = "0.1.3"
 test-case = "3.1"
 
+
 [build-dependencies]
-tonic-build.workspace = true
+tonic-build.workspace = true
```
djordon, line 79: nit: new line here too please 🙏🏽

signer/src/storage/postgres.rs
```diff
@@ -92,8 +97,136 @@ impl TryFrom<&NakamotoBlock> for model::StacksBlock {
...
+        ))
+    }
+
+    /// Apply the migrations to the database.
+    // Note 1: This could be generalized and moved up to the `storage` module, but
+    // left that for a future exercise if we need to support other databases.
+    // Note 2: The `sqlx` "migration" feature results in dependency conflicts
+    // with sqlite from the clarity crate.
+    // Note 3: The migration code paths have no explicit integration tests, but are
+    // implicitly tested by all integration tests using `new_test_database()`.
```
djordon, line 114: Maybe we should link to https://github.com/stacks-network/sbtc/issues/411 so that people can read up on the underlying issue.

Also, I think the main dependency conflict happens because of the `stackslib` crate. `rusqlite` is only an optional dependency for the clarity crate (though it is brought in by default). I'm pointing this out because if we want to change this in stacks-core, that's where we should focus.

It would be nice if stacks-core provided a way to _not_ bring in sqlite for almost all of its crates. It does for stacks-common and clarity, but not for stackslib, and we need some key types from stackslib (like `StacksTransaction`, `StacksBlockId`, and `NakamotoBlock`). This would allow us to decouple any sqlite dependency in the sbtc crates from the one in stacks-core. I think stacks-core is close to being able to allow us to not pull in a sqlite dependency by being particular about which feature flags are enabled, but we need to wait for Nakamoto to push that along (I opened https://github.com/stacks-network/stacks-core/pull/5004 for that but it needs to wait).

Also, shouldn't these be `///` instead of `//`?

signer/src/storage/postgres.rs
```diff
@@ -92,8 +97,136 @@ impl TryFrom<&NakamotoBlock> for model::StacksBlock {
...
+            .begin()
+            .await
+            .map_err(Error::SqlxBeginTransaction)?;
+
+        // Collect all migration scripts and sort them by filename. It is important
+        // that the migration scripts are named in a way that they are executed in
+        // the correct order, i.e. the current naming of `0001__`, `0002__`, etc.
+        let mut migrations = PGSQL_MIGRATIONS.files().collect::<Vec<_>>();
+        migrations.sort_by_key(|file| file.path().file_name());
+        for migration in migrations {
```
djordon, line 140: Have you thought about using something like [refinery](https://github.com/rust-db/refinery) for this? Although what is here seems to work, it seems like we can outsource much of this to that crate. And in the event of any migration issues, we'd probably want to add a few more columns to the migrations table to make debugging easier. So maybe it's better to inherit those features from an external crate. Thoughts?

signer/src/storage/postgres.rs
```diff
@@ -92,8 +97,136 @@ impl TryFrom<&NakamotoBlock> for model::StacksBlock {
...
+    // Note 3: The migration code paths have no explicit integration tests, but are
+    // implicitly tested by all integration tests using `new_test_database()`.
+    pub async fn apply_migrations(&self) -> Result<(), Error> {
+        tracing::info!("Preparing to run database migrations");
+
+        sqlx::raw_sql(
+            r#"
+                CREATE TABLE IF NOT EXISTS public.__sbtc_migrations (
+                    key TEXT PRIMARY KEY
+                );
```
djordon, line 122: I was thinking that we'd probably want to add a timestamp and a bunch of other stuff here too (basically the columns added by tools like `flyway` and `sqlx-cli`). But most of those are probably overkill. 🤔

Makefile
```diff
@@ -223,3 +223,14 @@ devenv-up: $(INSTALL_TARGET) $(EMILY_LAMBDA_BINARY) $(EMILY_CDK_TEMPLATE)
...
 	cd devenv/local/docker-compose && \
 	CONTAINER_HOST=$(_CONTAINER_HOST) sh down.sh
+
+run-signer:
+	docker compose --file docker-compose.signer.yml down;
+	docker compose --file docker-compose.signer.yml up postgres bitcoind --detach;
+	POSTGRES_PORT="0"; \
+	while [ "$$POSTGRES_PORT" -le 0 ]; do \
+		sleep 1; \
+		POSTGRES_PORT=$$(docker port sbtc-postgres 5432 | awk -F: '{print $$2}'); \
```
djordon, line 233: Not sure what I did wrong but the following returned an empty string for me:
```bash
docker port postgres 5432 | awk -F: '{print $$2}'
```
maybe try this?
```bash
docker port postgres 5432 | cut -d ":" -f 2
```

djordon (commented):
My main question is about whether we should outsource the db migrations code to the another crate, but in general this seems fine.

signer/src/storage/postgres.rs
cylewitruk: Yeah it's both `stackslib` and `clarity` that cause the big problem. I spent a couple hours trying to find a way around it but ultimately gave up to not waste too much time on it.

`clarity` necessarily pulls in `rusqlite` and uses it directly (see [sqlite.rs](https://github.com/stacks-network/stacks-core/blob/master/clarity/src/vm/database/sqlite.rs)) and changing stackslib would be very difficult and time-consuming as it uses it extensively (just maybe not as obvious since they wrap `rusqlite` connections and transactions in newtypes like `DBConn`, `DBTx`,  etc. just two name a couple).

signer/src/storage/postgres.rs
cylewitruk: Yeah I had some but removed them for simplicity's sake at this stage -- I didn't want to add things "just because", figured we could wait until the use-case presented itself.

Makefile
cylewitruk: Weird, works perfectly for me. If you change it locally does it work for you?

signer/src/storage/postgres.rs
cylewitruk: I wasn't aware of `refinery` and didn't see it when I did a quick search, but it looks nice! Sad it doesn't natively support sqlx and you have to do some workarounds. I'm all-for using an external crate otherwise (I briefly considered `diesel_migrations` but decided against it at this stage to avoid bringing in additional deps since these migrations are for now quite simple).

Idunno, for me "right now" I just felt like we needed some migration logic, and this "works for now" (though it maybe isn't the most advanced). I'd be interested in tabling it for now if this works in general and looking at "upgrading" when we start nearing a public release? Thoughts on that?

docker-compose.signer.yml
cylewitruk: Haha yeah... simply because the `devenv` is a giant soup at the moment :sweat_smile: `docker-compose.signer.yml` only includes the dependencies that are needed to run the signer _right now_ (bitcoind isn't explicitly needed yet but it's coming soon). A lot of the other stuff in `devenv` isn't exactly clear to me yet how they interact with the signer (if at all), so I'm kinda using this file for gathering the minimum necessary deps to run the signer as the integrations become a reality. My thinking was that it could then be used as a template for getting the devenv in better shape.

signer/src/storage/postgres.rs
cylewitruk: > Also, shouldn't these be /// instead of //?

I chose `//` for the notes as they're not really code-docs in the sense that you need them as documentation when using the method, but can be interesting for anyone actually reading the code here.

I can change them to code-docs, but I think the information as rather unnecessary for a caller?

signer/Cargo.toml
cylewitruk: [453237c](https://github.com/stacks-network/sbtc/pull/533/commits/453237c5c59dbadcab0614de699dbef2b296f17d)

Makefile
cylewitruk: [453237c](https://github.com/stacks-network/sbtc/pull/533/commits/453237c5c59dbadcab0614de699dbef2b296f17d)

signer/src/storage/postgres.rs
```diff
@@ -92,8 +97,136 @@ impl TryFrom<&NakamotoBlock> for model::StacksBlock {
...
 impl PgStore {
     /// Connect to the Postgres database at `url`.
-    pub async fn connect(url: &str) -> Result<Self, sqlx::Error> {
-        Ok(Self(sqlx::PgPool::connect(url).await?))
+    pub async fn connect(url: &str) -> Result<Self, Error> {
+        Ok(Self(
+            sqlx::PgPool::connect(url)
+                .await
+                .map_err(Error::SqlxConnect)?,
+        ))
```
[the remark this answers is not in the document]
cylewitruk: [453237c](https://github.com/stacks-network/sbtc/pull/533/commits/453237c5c59dbadcab0614de699dbef2b296f17d)

signer/src/storage/postgres.rs
djordon: That works. Let's do that.

signer/src/storage/postgres.rs
djordon: I see. Well, maybe move it to just after the function definition? For someone reading the official crate documentation it'll be fine but when looking at the source you'll get the wrong impression.

docker-compose.signer.yml
djordon: That makes sense.

Makefile
```diff
@@ -223,3 +223,15 @@ devenv-up: $(INSTALL_TARGET) $(EMILY_LAMBDA_BINARY) $(EMILY_CDK_TEMPLATE)
 devenv-down:
 	cd devenv/local/docker-compose && \
 	CONTAINER_HOST=$(_CONTAINER_HOST) sh down.sh
+
+run-signer:
+	docker compose --file docker-compose.signer.yml down;
+	docker compose --file docker-compose.signer.yml up postgres bitcoind --detach;
```
djordon, line 229: I suppose this is fine for now, but let's add a TODO to clean up the devenv docker compose file so that we can consolidate around that. I think all we need to do is add a profile for the core services that the signer needs, add the signer service, and fix the bitcoin service which is unnecessarily complex.

djordon (approved):
Looks good

signer/src/storage/postgres.rs
cylewitruk: Changed in [e02ec59](https://github.com/stacks-network/sbtc/pull/533/commits/e02ec59a919cd54177bf6ba909cbb9bf92ef80a7)

Makefile
cylewitruk: Added in https://github.com/stacks-network/sbtc/pull/533/commits/b84e42c3a97ad954e6f9ddaf25421b07b9a9a9d1 and linked to #536

Makefile
cylewitruk: Per slack we'll leave it as-is for now. If you end up using this and need to change it then feel free, both work for me.

cylewitruk:
> Looks good

Added #537 to revisit this pre-launch, and added a TODO comment in the `apply_migrations()` fn referencing it as well.

signer/src/storage/postgres.rs
cylewitruk: Added #537 to revisit

Issue #516, opened 2024-09-12
[Feature]: Reclaim script lock time check can be bypassed


evonide:
# (Medium) Reclaim script lock time check can be bypassed

## 1. Description

The code currently parses Bitcoin TX `reclaim_scripts` for deposit requests in [sbtc/sbtc/src/deposits.rs at 033e2cbd2def829a3e30df77bc274b325d68cdbe · stacks-network/sbtc](https://github.com/stacks-network/sbtc/blob/033e2cbd2def829a3e30df77bc274b325d68cdbe/sbtc/src/deposits.rs#L175). The reclaim script’s idea is that a user should be able to reclaim their UTXO after a certain time period. To do so the parsing checks the reclaim format against the following ones:
https://github.com/stacks-network/sbtc/blob/ffbe79e0238825f3546e9429bbf89042703823df/sbtc/src/deposits.rs#L421
```
    pub fn parse(reclaim_script: &ScriptBuf) -> Result<Self, Error> {
        let (lock_time, script) = match reclaim_script.as_bytes() {
            [0, OP_CSV, script @ ..] [...]
            [n, OP_CSV, script @ ..] [...]
            [n, rest @ ..] [...]
            [...]
    }
 ```

There are some constraints on this provided `lock_time` such as:
https://github.com/stacks-network/sbtc/blob/ffbe79e0238825f3546e9429bbf89042703823df/sbtc/src/deposits.rs#L386-L389 

However, this does not enforce a minimum locking time. Accordingly, a format such as `[0, OP_CSV, script @ ..]` should allow a user to currently effectively bypass the lock time check. Additionally, as per [Locktime | Post-dating a Bitcoin Transaction](https://learnmeabitcoin.com/technical/transaction/locktime/) 

> [...] If all of the inputs' sequence values are set to the maximum value of 0xffffffff, the transaction is considered "final" and the locktime feature is disabled.

- There’s doesn't seem to be a check against this edge case which effectively would provide another way to bypass the time lock mechanism.
- Please note hat the Bitcoin code seems to bypass the locktime feature once the MSB at `1 << 31` is set as per  https://github.com/bitcoin/bitcoin/blob/349632e022da22a457a85650360b5be41fa500dc/src/script/interpreter.cpp#L584. This might need some closer observation to implement securely.

As per [[Design]: sBTC Deposit UTXO Binary Format · Issue #30 · stacks-network/sbtc](https://github.com/stacks-network/sbtc/issues/30):
[...]

PR #535, opened 2024-09-16
feat: reject non-standard deposit scripts and problematic lock-times

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/516 and addresses an issue pointed out in https://github.com/stacks-network/sbtc/issues/514#issuecomment-2350054724.

## Changes

* Reject deposits where the reclaim script uses a lock-time that would disable the OP_CSV opcode. I'm still in disbelief at the design of that opcode, so I wouldn't be surprised if my understanding is just plain wrong.
* Fix an issue where `DepositScriptInputs::parse(script).deposit_script() != script`. Although it is very unlikely that this issue would sneak past validation, it could be problematic in other scenarios. Also, the intent is that `DepositScriptInputs::parse(script).deposit_script() == script` so let's make it so.

## Testing Information

I'd like to add an integration test and I may still do so, but what is here is good enough for review.

## Checklist:


xoloki (approved):
Lgtm. The integration test is more readable than i expected.

Issue #539, opened 2024-09-17
[Feature]:  Add `complete-deposit` queries 


djordon:
# Feature - Add `complete-deposit` queries

## 1. Description

Before we can validate `complete-deposit` transactions, we need a few queries to make sure that we can fetch the raw bitcoin transaction from the database, as well as check that a given block is in the canonical bitcoin blockchain.

## 2. Technical Details:

We need to add a few queries

### 2.1 Acceptance Criteria:
   - [ ] We have a query that checks whether a given block is on the canonical bitcoin blockchain.
   - [ ] We have a query that fetches a bitcoin transaction from the database.

## 3. Related Issues and Pull Requests (optional):

This is part of https://github.com/stacks-network/sbtc/issues/479.

PR #540, opened 2024-09-17
feat: add `complete-deposit` queries

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/539.

## Changes

* Add two queries that fetch necessary data from the database.
* Add an internal `BitcoinTx` type for ease of writing to and reading from the database.

## Testing Information

This PR adds integration tests for the queries. The in-memory implementation of the same queries have been added as well, but right now they go unused and so we skip their tests.

## Checklist:

### DIFF
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -329,6 +329,9 @@ impl super::DbRead for PgStore {
         context_window: u16,
         threshold: u16,
     ) -> Result<Vec<model::DepositRequest>, Self::Error> {
+        // TODO: Make sure we get only pending deposits, don't include ones
+        // where we have a completed-deposit event on the canonical stacks
+        // blockchain.
         sqlx::query_as::<_, model::DepositRequest>(
             r#"
             WITH RECURSIVE context_window AS (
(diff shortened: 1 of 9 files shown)


signer/src/storage/postgres.rs
```diff
@@ -329,6 +329,9 @@ impl super::DbRead for PgStore {
         context_window: u16,
         threshold: u16,
     ) -> Result<Vec<model::DepositRequest>, Self::Error> {
+        // TODO: Make sure we get only pending deposits, don't include ones
```
cylewitruk, line 332: Should there be a tracking issue for this?

signer/src/testing/storage/model.rs
```diff
@@ -251,6 +255,14 @@ impl TestData {
 
         stacks_blocks
     }
+
+    /// Fetch the parent block given the hash.
+    pub fn get_parent_bitcoin_block(&self, parent: &BitcoinBlockHash) -> Option<BitcoinBlock> {
```
cylewitruk, line 260: Shouldn't this just be `get_bitcoin_block(block_hash)` since you're not actually doing any parent lookups here?

cylewitruk (commented):
Looks good, just had two questions

signer/src/testing/storage/model.rs
djordon: Oh yeah, that's right, will update.

signer/src/storage/postgres.rs
djordon: Oh, yes. Will update.

signer/src/testing/storage/model.rs
djordon: https://github.com/stacks-network/sbtc/pull/540/commits/16c7b7d5b5c8ecab23c0bb435c0d02933c204de4

signer/src/storage/postgres.rs
djordon: Updated in https://github.com/stacks-network/sbtc/pull/540/commits/16c7b7d5b5c8ecab23c0bb435c0d02933c204de4. I opened https://github.com/stacks-network/sbtc/issues/543 for this.

cylewitruk (approved):
lgtm!

PR #546, opened 2024-09-18
refactor: remove associated error types from internal traits

djordon:
## Description

Implements part of the work done in https://github.com/stacks-network/sbtc/pull/507.

## Changes

* Remove the associated error type from the `DbRead`, `DbWrite`, `BitcoinInteract`, and `MessageTransfer` traits.
* Fixes bug/typo in the tests that may have caused tests in `main` to fail.

## Testing Information

This is just a refactor, so if the tests pass then all should be well.

## Checklist:

### DIFF
--- a/signer/src/network/in_memory.rs
+++ b/signer/src/network/in_memory.rs
@@ -59,19 +59,18 @@ impl Default for Network {
 impl super::MessageTransfer for MpmcBroadcaster {
     async fn broadcast(&mut self, msg: super::Msg) -> Result<(), Error> {
         self.recently_sent.push_back(msg.id());
-        self.sender.send(msg).unwrap();
+        self.sender.send(msg).map_err(|_| Error::SendMessage)?;
         Ok(())
     }
 
     async fn receive(&mut self) -> Result<super::Msg, Error> {
-        let mut msg = self.receiver.recv().await.unwrap();
+        let mut msg = self.receiver.recv().await.map_err(Error::ChannelReceive)?;
 
         while Some(&msg.id()) == self.recently_sent.front() {
             self.recently_sent.pop_front();
-            msg = self.receiver.recv().await.unwrap();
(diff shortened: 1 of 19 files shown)


signer/src/network/in_memory.rs
```diff
@@ -55,32 +57,21 @@ impl Default for Network {
 }
 
 impl super::MessageTransfer for MpmcBroadcaster {
-    type Error = Error;
-    async fn broadcast(&mut self, msg: super::Msg) -> Result<(), Self::Error> {
+    async fn broadcast(&mut self, msg: super::Msg) -> Result<(), Error> {
         self.recently_sent.push_back(msg.id());
-        self.sender.send(msg).map_err(|_| Error::Send)?;
+        self.sender.send(msg).unwrap();
```
cylewitruk, line 62: Are we okay with using unwraps here? Thinking if we ever did enable using the in-memory db at runtime (for regtest/testnet etc.)? I'm not against it, just that we have a decision that unwraps are OK for the in-mem DB.

cylewitruk (commented):
Ahhh, thank you for this! :clinking_glasses: Just causes headaches when we know what error type we're going to be using lol.. Just had one Q about the unwraps in the in-mem db.

signer/src/network/in_memory.rs
djordon: I don't know. I'm generally of the opinion unwraps aren't okay, but this is extremely  close to a testing utility. And unlike our in-memory db, I haven't used this stuff at all, so I had half a mind to remove it. 

But thinking about this some more, I'll switch back to using our normal pattern and later we can decide on what, if anything, to do with this module.

cylewitruk: Sounds good :+1:

cylewitruk (approved):
:heavy_check_mark: lgtm

Issue #548, opened 2024-09-19
[Feature]: Wire up `BlockObserver` in the main entrypoint


cylewitruk:
# Feature - Wire up `BlockObserver` in the main entrypoint

## 1. Description

The `BlockObserver` needs to be wired-up and its runloop run in the main event loop.

### 1.1 Context & Purpose
The `BlockObserver` is a dependency for both the `TransactionSigner` and the `TransactionCoordinator` and needs to be running prior to either of these being started (they are dependent on its messaging channels).

## 2. Technical Details:
The `BlockObserver` itself has several dependencies which also need to be wired up:
- `BitcoinInteract` + `BitcoinClient`
- `EmilyClient` (which I think should be changed to `EmilyInteract`)
- `StacksClient` (which I think should be changed to `StacksInteract`)
- Bitcoin blockhash stream (from zmq)
- ✅ P2P network

Then it also takes a watch channel for informing the application of database updates, which need to be centralized somehow so that the `TransactionSigner` and `TransactionCoordinator` instances can use it. This should probably be updated to use the app signalling channel instead with a proper message type to promote decoupling.

### 2.1 Acceptance Criteria:
   - [ ] The `BlockObserver` can be created with all of its dependencies
   - [ ] All required dependencies for the `TransactionSigner` and `TransactionCoordinator` are available for them to be created.
   - [ ] The `BlockObserver` correctly runs and can process events in the dev environment.

Issue #486, opened 2024-09-04
[Feature]: Add support for multiple bitcoin nodes for `BitcoinCoreClient`


cylewitruk:
# Feature - Add support for multiple bitcoin nodes for `BitcoinCoreClient`

## 1. Description

On par with #484 which allows the configuration of multiple Bitcoin RPC endpoints, we need a mechanism for being able to use multiple endpoints like was done with `StacksClient` in #225.

### 1.1 Context & Purpose
Just like with Stacks nodes, we should support multiple Bitcoin nodes. The primary driver for this is so that the Signer doesn't need to experience potential downtime if a Bitcoin Core node is being updated if the operator runs two Bitcoin nodes.

## 2. Technical Details:
We can either implement this as a part of the `BitcoinCoreClient` like was done with `StacksClient`, or we could implement a pooling wrapper over the client.

### 2.1 Acceptance Criteria:
   - [ ] If multiple Bitcoin RPC endpoints are specified and a connection to one fails, the remaining endpoints should be tried, until the available endpoints have been exhausted, then an error should be returned.

## 3. Related Issues and Pull Requests (optional):
- #484

djordon:
I think this was closed in https://github.com/stacks-network/sbtc/pull/549.

PR #549, opened 2024-09-19
feat: generic api failover client + bitcoin client proof-of-concept with it

cylewitruk:
## Description

Adds a generic fallback/retry type to be used with API clients. Provides the first integration with the `BitcoinCoreClient` and  the `BlockObserver` to ensure that everything plays nicely together and to show its usage.

The goal here is to ensure that we have some measure of resiliency in our external API calls without needing to duplicate  that kind of code everywhere (i.e. Bitcoin, Stacks, Emily & Blocklist clients).

I was a little bit back-and-forth on whether or not to put the API clients on the context, but in the end decided to so that they can be created at application startup and the fallback client's state (i.e. current "live" endpoint) shared with the rest of the app. The context now has two methods for newing up: `init()` is more implicit but requires more trait bounds, where `new()` can be used in tests to provide concrete impls where needed (i.e. the `TestHarness` in `block_observer.rs`).

The fallback client's logic is pretty naive as it is right now, but we can expand upon this later to add exponential backoffs per endpoint, simpler load-balancing across endpoints (round-robin or lowest-latency, for example), etc. if we want.

Closes:
- #486 

Works towards:
- #548 

## Changes
- Adds the `ApiFallbackClient` type which can be used to failover to a new endpoint while also supporting retries (default is 3 retries or the number of clients, whichever is higher). Probably not the best name tbh since it's not necessarily specific to API's, but that's what we use it for right now :shrug: 
- Reworked the `sbtc` library errors a little so that we get rid of the `dyn Error` and problems with `Send`.
- Refactored `load_latest_deposit_requests()` in the `BlockObserver` which otherwise had problems with type signatures (between `signer` and `sbtc` crates) and wouldn't use the failover functionality otherwise.
- Added the bitcoin client to the context and implemented this in the `BlockObserver`.
- Some other small stuff I saw.

## Testing Information

Nothing specific, all existing tests pass and there are new tests testing the `ApiFallbackClient`s failover and retry functionality.

## Checklist:


cylewitruk:
Looks like this PR is hitting the http 403 issue too... :thinking:

signer/src/bitcoin/client.rs
```diff
@@ -0,0 +1,59 @@
...
+        let clients = urls
+            .iter()
+            .map(|url| BitcoinCoreClient::try_from(url.clone()).unwrap())
+            .collect::<Vec<_>>();
+
+        Ok(Self::new(clients))
+    }
+}
+
+impl BitcoinInteract for BitcoinCoreClient {
```
cylewitruk, line 28: Yeah I coulda implemented `BitcoinInteract` for `ApiFallbackClient<BitcoinCoreClient>` and then kept `ApiFallbackClient` out of the `Context` but I didn't... maybe that's the better way to go :thinking:

cylewitruk: I spoke with Matteo and he seemed to agree with the thoughts that my lazy-angel wanted to ignore (i.e. the "yeah i coulda" above, so I'll make that change.

signer/src/block_observer.rs
```diff
@@ -63,21 +61,20 @@ pub struct BlockObserver<BitcoinClient, StacksClient, EmilyClient, BlockHashStre
...
     SC: StacksInteract,
     EC: EmilyInteract,
     S: DbWrite + DbRead + Send + Sync,
     BHS: futures::stream::Stream<Item = bitcoin::BlockHash> + Unpin,
 {
     /// Run the block observer
-    #[tracing::instrument(skip(self))]
-    pub async fn run(mut self) -> Result<(), Error> {
+    #[tracing::instrument(skip(self, ctx))]
+    pub async fn run(mut self, ctx: impl Context) -> Result<(), Error> {
```
djordon, line 73: Ahh, so it looks like this PR moves us to having all (or most) of the clients encapsulated in the context?

signer/src/util.rs
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        Err(Error::AllFailoverClientsFailed)
+    }
+}
+
+impl<T> ApiFallbackClient<T> {
+    /// Create a new fallback client from a list of clients.
+    pub fn new(clients: Vec<T>) -> Self {
+        let retry_count = min(3, clients.len());
```
djordon, line 79: Probably better to make the `retry_count` an explicit input or a global constant.

signer/src/error.rs
```diff
@@ -21,6 +35,11 @@ pub enum Error {
     #[error("tokio i/o error: {0}")]
     TokioIo(#[from] tokio::io::Error),
 
+    /// Error occurred when constructing a BitcoinCoreClient. This takes a String
+    /// as [`sbtc::error::Error`] is not [`Sync`] + [`Send`].
+    #[error("failed to construct a BitcoinCoreClient: {0}")]
+    BitcoinCoreClient(String),
```
djordon, line 41: Is that still the case even after removing the `sbtc::error::Error::BitcoinClient` variant? If not, I wonder if making that error variant take a `Box<dyn std::error::Error + Send + Sync + 'static>` was also a fix. You'd probably have to change the associated error type of the `BitcoinClient`.

sbtc/src/deposits.rs
```diff
@@ -128,14 +128,15 @@ impl CreateDepositRequest {
...
+    pub async fn validate<C>(&self, client: &C) -> Result<Deposit, Error>
     where
         C: BitcoinClient,
     {
         // Fetch the transaction from either a block or from the mempool
         let response = client
             .get_tx(&self.outpoint.txid)
-            .map_err(|err| Error::BitcoinClient(Box::new(err)))?;
+            .await
+            .map_err(|_| Error::GetBitcoinTx(self.outpoint.txid))?;
```
djordon, line 139: Although I'd like to keep the actual error, I'm thinking that this function should be moved anyway since it's not really needed in this crate. But if this stays, we should try to keep the `Error::BitcoinClient` variant, and perhaps enhance it with the `txid` field.

djordon (commented):
I had some comments. I'm almost done, figured I submit what I have already.

sbtc/src/rpc.rs
```diff
@@ -56,7 +58,7 @@ pub trait BitcoinClient {
 
     /// Return the transaction if the transaction is in the mempool or in
     /// any block.
-    fn get_tx(&self, txid: &Txid) -> Result<GetTxResponse, Self::Error>;
+    fn get_tx(&self, txid: &Txid) -> impl Future<Output = Result<GetTxResponse, Self::Error>>;
```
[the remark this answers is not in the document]
cylewitruk: Yeah I had it going the other way, but decided to change the trait here since right now the only user is `signer` and there we're async (and assuming emily etc. will be too).

signer/src/block_observer.rs
cylewitruk: At least for right now -- I've been a bit on the fence, but otoh these clients will only be used in the signer and are created from the config, so it doesn't _not_ make sense to avoid needing to new these up for every dependency :shrug:

signer/src/error.rs
cylewitruk: Now `sbtc::error::Error` is `Sync + Send` -- I don't think I use this specific variant anymore as I added concrete variants, will check. I do think I tried adding `Sync + Send` to the declaration and it still didn't help, as well as using `Arc` instead of `Box`.

cylewitruk: Hm, I must have ended up in some sort of mixed situation because when I went back and tried it again cleanly it worked. So I've updated this now to retain the error variant :+1:

sbtc/src/deposits.rs
cylewitruk: Was able to keep it :+1:

signer/src/util.rs
cylewitruk: I felt like this was a reasonable default -- if you give 5 endpoints you want it to try at least 5 times to touch all endpoints, and if you have one endpoint you probably want it to try at least 2-3 times. It can be updated using `set_retry_count()` otherwise..?

sbtc/Cargo.toml
```diff
@@ -23,6 +23,7 @@ tracing.workspace = true
 tracing-attributes.workspace = true
 tracing-subscriber.workspace = true
 url.workspace = true
+tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
```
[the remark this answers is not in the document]
cylewitruk: Yeah you're right, moved

sbtc/src/rpc.rs
```diff
@@ -63,6 +65,23 @@ pub struct BitcoinCoreClient {
...
 
+/// Implement TryFrom for Url to allow for easy conversion from a URL to a
+/// BitcoinCoreClient.
+impl TryFrom<Url> for BitcoinCoreClient {
+    type Error = Error;
+
+    fn try_from(url: Url) -> Result<Self, Self::Error> {
+        let username = url.username().to_string();
+        let password = url.password().unwrap_or_default().to_string();
+        let host = url.host_str().unwrap();
```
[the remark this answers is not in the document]
cylewitruk: Yup, updated in [1ba447a](https://github.com/stacks-network/sbtc/pull/549/commits/1ba447a881ae945f401ad3645ab89873d756815c)

sbtc/src/rpc.rs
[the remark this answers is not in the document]
djordon: I see. 

Just a heads up, I don't think Emily will actually use this trait or any functionality where we fetch information from bitcoin-core (I originally thought that it would). So we should probably just move this code (and some other stuff) over to the `signer` crate.

signer/src/block_observer.rs
djordon: > [I]t doesn't _not_ make sense to avoid needing to manually new these up for every dependency 🤷 It also lets us share the current "alive" endpoint so we avoid unnecessary retries.
>
> I figure that if it ends up not feeling right, it's not a big refactor to move them out again.

Yeah, that all makes sense me.

signer/src/util.rs
djordon: Oh I was sloppy with my wording. I was trying to say that it's probably better to do something like
```rust
pub fn new(clients: Vec<T>, retry_count: usize) -> Self {
    ...
```
or
```rust
pub fn new(clients: Vec<T>, retry_count: Option<usize>) -> Self {
    let retry_count = retry_count.unwrap_or_else(|| clients.len());
```
or
```rust
/// blah blah blah
const DEFAULT_MINIMUM_RETRY_COUNT: usize = 3;

pub fn new(clients: Vec<T>) -> Self {
    let retry_count = min(DEFAULT_MINIMUM_RETRY_COUNT, clients.len());
```
or maybe some combination of the above.

signer/src/util.rs
```diff
@@ -0,0 +1,258 @@
...
+            return result.map_err(Into::into);
+        }
+
+        Err(Error::AllFailoverClientsFailed)
+    }
+}
+
+impl<T> ApiFallbackClient<T> {
+    /// Create a new fallback client from a list of clients.
+    pub fn new(clients: Vec<T>) -> Self {
```
djordon, line 83: We probably want to check that the vector is non-empty too.

djordon (commented):
Had a few more comments.

signer/src/util.rs
cylewitruk: Yeah I'll add that. In our case it's already guarded by the config validation, but yeah.

djordon: Oh yeah that's right.

cylewitruk: Added in [7574dd2](https://github.com/stacks-network/sbtc/pull/549/commits/7574dd2a9d48a462426cf8874308485add831a29)

signer/src/bitcoin/client.rs
```diff
@@ -0,0 +1,84 @@
...
+};
+
+/// Implement the [`TryFrom`] trait for a slice of [`Url`]s to allow for a
+/// [`ApiFallbackClient`] to be implicitly created from a list of URLs.
+impl TryFrom<&[Url]> for ApiFallbackClient<BitcoinCoreClient> {
+    type Error = Error;
+    fn try_from(urls: &[Url]) -> Result<Self, Self::Error> {
+        let clients = urls
+            .iter()
+            .map(|url| BitcoinCoreClient::try_from(url.clone()).unwrap())
```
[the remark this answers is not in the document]
cylewitruk: Fixed in [7574dd2](https://github.com/stacks-network/sbtc/pull/549/commits/7574dd2a9d48a462426cf8874308485add831a29)

Note: I added the following to catch this kind of thing pre-PR. I added `unwrap_used` and `expect_used` but commented out since there's some left I don't want to address them in this PR, but I think it feels like a pretty good goal to shoot for being able to enable these. Thoughts? They're disabled again in the `testing` mod.

```rust
#![deny(
    clippy::unwrap_in_result,
    // clippy::unwrap_used, // TODO: There's one unwrap left
    // clippy::expect_used, // TODO: There's 14 expects left
)]
```

sbtc/src/rpc.rs
```diff
@@ -63,6 +67,25 @@ pub struct BitcoinCoreClient {
...
+impl TryFrom<Url> for BitcoinCoreClient {
+    type Error = Error;
+
+    fn try_from(url: Url) -> Result<Self, Self::Error> {
+        let username = url.username().to_string();
+        let password = url.password().unwrap_or_default().to_string();
+        let host = url
+            .host_str()
+            .ok_or(Error::InvalidUrl(url::ParseError::EmptyHost))?;
+        let port = url.port().unwrap_or(8332);
```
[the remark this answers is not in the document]
cylewitruk: Addressed in [7574dd2](https://github.com/stacks-network/sbtc/pull/549/commits/7574dd2a9d48a462426cf8874308485add831a29)

signer/src/util.rs
cylewitruk: Yeah no I got it, I just didn't really want to force "users" to calculate their own retry counts when newing up (where they can instead use the `set_retry_count()` fn to set it manually _if they want to_). But I've at least changed it to a constant now so it's not buried in the code :+1:

cylewitruk:
Note that I removed the incorrect impl of bitcoin fee estimation and replaced it with a todo.

signer/src/bitcoin/client.rs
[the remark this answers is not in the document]
djordon: I think `clippy::unwrap_in_result` is a good one to add, no objection from me there. And `clippy::unwrap_used` seems like a good one as well. Outside of tests it is always preferable to use `expect` instead of `unwrap` since you are forced to give context. But with `expect`, there are times where its use is perfectly fine. For example, when "writing" to a `Vec` when using the `std::io::Write` trait, using `expect` is fine. Maybe the computer runs out of memory but the `std::io::Write::write` call won't return a `Result::Err`. So I feel like `.expect` should be viewed with suspicion but not strictly forbidden.

djordon (commented):
Looks good. I have a few quasi suggestions and one concrete question.

signer/src/bitcoin/client.rs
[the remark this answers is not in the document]
cylewitruk: I kinda figured that in the cases where you want to explicitly use any of the above that one would also explicitly disable the clippy warning with a comment on why it's OK to use.

signer/src/bitcoin/client.rs
```diff
@@ -0,0 +1,78 @@
...
+impl BitcoinInteract for ApiFallbackClient<BitcoinCoreClient> {
+    async fn get_block(
+        &self,
+        block_hash: &bitcoin::BlockHash,
+    ) -> Result<Option<bitcoin::Block>, Error> {
+        self.exec(|client| async {
+            match client.inner_client().get_block(block_hash) {
+                Ok(block) => Ok(Some(block)),
+                // TODO: Double check what the error is from bitcoin-core when no block is found
+                Err(bitcoincore_rpc::Error::JsonRpc(_)) => Ok(None),
```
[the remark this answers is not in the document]
cylewitruk: Yeah, at a quick glance at the code it wasn't really clear what the response is for a "404" i.e. block not found. Actually, I can probably write some integration tests around this and get a concrete answer... I'll try doing that.

signer/src/bitcoin/client.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    async fn broadcast_transaction(&self, tx: &bitcoin::Transaction) -> Result<(), Error> {
+        self.exec(|client| async {
+            client
+                .inner_client()
+                .send_raw_transaction(tx)
+                .map_err(Error::BitcoinCoreRpc)
+                .map(|_| ())
+        })
+        .await
```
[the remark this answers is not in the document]
cylewitruk: Yeah that's the implication. Hangs a little together with the todo above re: "block not found". I.e. here I've made the _assumption_ that anything that doesn't result in the transaction in the mempool returns an error.

I'm gonna write some integration tests for the above error to try to figure out what it actually returns -- I'll toss in a couple for this as well so we know.

signer/src/util.rs
```diff
@@ -0,0 +1,290 @@
...
+            let client_index = self.last_client_index.load(Ordering::Relaxed);
+            let result = f(&self.inner_clients[client_index]).await;
+
+            if let Err(error) = result {
+                tracing::warn!(%error, retry_num=i, max_retries=retry_count, "failover client call failed");
+                self.last_client_index.store(
+                    (client_index + 1) % self.inner_clients.len(),
+                    Ordering::Relaxed,
+                );
+                continue;
```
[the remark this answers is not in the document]
cylewitruk: Yeah this is just a naive "something that works for right now", I fully expect that it will be iterated upon :)

You can still achieve the same effect as `retry_if` though even with this, you just bake your evaluation of `f` within `exec()` and return an `Ok` if you don't want to retry.

signer/src/bitcoin/client.rs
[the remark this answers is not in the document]
djordon: Oh don't bother, we have some (or will have some). But we aren't supposed to trigger them. But I can see it happening if someone configures bitcoin-core in a "more strict" way than we expect (like in a new version of bitcoin-core, or purposefully in an existing version). No need to change this here and now though.

signer/src/util.rs
```diff
@@ -0,0 +1,290 @@
...
+    last_client_index: AtomicUsize,
+    retry_count: AtomicU8,
+}
+
+impl<T> InnerApiFallbackClient<T> {
+    /// Set the number of retries to perform before giving up.
+    pub fn set_retry_count(&self, retry_count: u8) -> Result<(), FallbackClientError> {
+        if retry_count == 0 {
+            return Err(FallbackClientError::RetryCountZero);
+        }
```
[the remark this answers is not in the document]
cylewitruk: Yeah but then we're just pushing the problem further away, i.e. with `NonZeroU8::new(4).expect("..")` and then we don't get a nice error stating that the retry count specifically must be > 0 :)

signer/src/bitcoin/client.rs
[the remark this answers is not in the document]
cylewitruk: I put up a skeleton of a test in [1da89a3](https://github.com/stacks-network/sbtc/pull/549/commits/1da89a3268fe12673d07ae0f37456a70c6307ded) -- we should implement it but I didn't want to take the time on it right now with setting up the correct inputs/outputs/etc.

signer/src/bitcoin/client.rs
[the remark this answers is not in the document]
cylewitruk: Added a couple of tests in [1da89a3](https://github.com/stacks-network/sbtc/pull/549/commits/1da89a3268fe12673d07ae0f37456a70c6307ded) which showed a bit of what we get for different scenarios.

Take a look at the comments I had in that test re: the integration test wallet -- do you know any quick workarounds to make it work against the wallet that's created for the integration tests?

djordon (approved):
Looks good ✅

signer/src/util.rs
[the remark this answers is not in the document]
cylewitruk: Yeah good catch, I reverted this to allow a `retry_count` of zero with one implicit try (i.e. changing from `0..retry_count` to `0..=retry_count` (and a test to cover it).

djordon (approved):
Nice, just went through https://github.com/stacks-network/sbtc/pull/549/commits/2d94637f64354aeb49ea3f4dfd81f671dcf66d25, looks good.

signer/src/bitcoin/client.rs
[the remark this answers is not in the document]
djordon: > I kinda figured that in the cases where you want to explicitly use any of the above that one would also explicitly disable the clippy warning with a comment on why it's OK to use there. That makes it clear that it's no accident.

👍🏽 oh yeah, that is better.

PR #551, opened 2024-09-20
ci: more integration tests

djordon:
## Description

Closes nothing.

This is a quick follow-up to https://github.com/stacks-network/sbtc/pull/535.

## Changes

* Adds an integration test that checks the minimal-push-rule (to show that https://github.com/stacks-network/sbtc/pull/535 was necessary).
* Update some of our integration tests helper functions to go through the types in the `sbtc` crate.

## Testing Information

This is a refactor that uses more internal types in the integration tests. The idea is that transactions that conform to the internal types will be accepted by bitcoin-core and make things work smoothly.

## Checklist:

### DIFF
--- a/sbtc/tests/integration/validation.rs
+++ b/sbtc/tests/integration/validation.rs
@@ -178,9 +178,8 @@ fn minimal_push_check() {
         }],
     };
     // We just created a transaction that spends the P2SH UTXO where we
-    // spend a deposit that did not adhere to the non-minimal push rule.
-    // When bicoin-core attempts to validate the transaction it should
-    // fail.
+    // spend a deposit that did not adhere to the minimal push rule. When
+    // bicoin-core attempts to validate the transaction it should fail.
     match rpc.send_raw_transaction(&tx1).unwrap_err() {
         BtcRpcError::JsonRpc(JsonRpcError::Rpc(RpcError { code: -26, message, .. })) => {
             let expected_message =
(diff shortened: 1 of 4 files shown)


sbtc/tests/integration/validation.rs
```diff
@@ -81,6 +81,115 @@ fn tx_validation_from_mempool() {
...
+        }],
+        output: vec![TxOut {
+            value: Amount::from_sat(amount - fee),
+            script_pubkey: depositor.address.script_pubkey(),
+        }],
+    };
+    // We just created a transaction that spends the P2SH UTXO where we
+    // spend a deposit that did not adhere to the minimal push rule. When
+    // bicoin-core attempts to validate the transaction it should fail.
+    match rpc.send_raw_transaction(&tx1).unwrap_err() {
```
cylewitruk, line 183: The following is a nicer way to check this imo (no change required, just a suggestion):

```rust
assert!(matches!(
    rpc.send_raw_transaction(&tx1).unwrap_err(),
    BtcRpcError::JsonRpc(JsonRpcError::Rpc(err))
        if err.code == -26 && err.message == "non-mandatory-script-verify-flag (Data push larger than necessary)"
))
```

cylewitruk (approved):
Had to read the `minimal_push_check()` test a few times to get it but I think it looks good assuming I've understood what it's doing correctly :sweat_smile: (Bitcoin script isn't my forté)

sbtc/tests/integration/validation.rs
djordon: Yeah I like it more aesthetically, but it's harder to debug if bitcoin-core returns a different error response. But I changed it to combine the best of both approaches.

Issue #412, opened 2024-08-09
[Feature]: Use protobuf 3 for signer communication


djordon:
# Feature - Use protobuf 3 for signer communication

## 1. Description

Write down the messages that we intend to send as protobufs, and serialize those types when the signers communicate with one another.

### 1.1 Context & Purpose

Right now the signers use [bincode](https://crates.io/crates/bincode) to serialize their messages when communicating with one another. This is fine, so long as we carefully coordinate updates to the types that they use to communicate, since communication can break down otherwise. So instead of using bincode for serde (serialization and deserialization) we should protobufs 3, since using protobuf 3 (and adhering to best practices) will give us a better upgrade story for how the signers communicate.

## 2. Technical Details:

We'd need to write down the API for each of the types that would be sent as protobufs. We'd need to add conversion functions/impls for converting protobufs into the regular internal types.

### 2.1 Acceptance Criteria:
   - [ ] We have a protobuf file with types that the signers use to communicate. Technically we already have one, but we'd need to flesh it out so that we can remove our use of bincode.
   - [ ] Add conversion functions and impls of protobufs into internal types.
   - [ ] Remove our use of bincode.

## 3. Related Issues and Pull Requests (optional):

- https://github.com/stacks-network/sbtc/issues/846
- https://github.com/stacks-network/sbtc/issues/847

PR #553, opened 2024-09-20
feat: commit generated protobufs

djordon:
## Description

Closes nothing.

We're moving to commit our generated sources to the repo. Although the protobufs aren't the cause of our what is prompting the change, I figured we might was well commit these as well. Plus this change is easy to revert.

## Changes

* Change the build script to output sources to a folder in the signer/src.
* Compile protobufs only when there is a change to the protobuf file.

## Testing Information

Well if everything builds then we're okay, this just affects where the built files are stored.

## Checklist:


cylewitruk:
@djordon what would you think of simply removing this and related code since the relay isn't being used (in favor of LibP2P)? Like, we could stuff it in an archive branch, but I don't think we need it in the active code-base for right now?

djordon:
> @djordon what would you think of simply removing this and related code since the relay isn't being used (in favor of LibP2P)? Like, we could stuff it in an archive branch, but I don't think we need it in the active code-base for right now?

Yeah, that's right, we don't need this stuff right now. And we don't need the relay stuff either. The reason why I think we should keep the protobuf changes is because I think we should still switch to using protobufs as the serialization format for peer-to-peer communication (https://github.com/stacks-network/sbtc/issues/412). So I'm okay with another PR that removes the relay stuff, but we will want this protobuf stuff later when/if we transition to using protobufs for communication.

cylewitruk:
> I think we should still switch to using protobufs as the serialization format for peer-to-peer communication (#412). 

Ah, okay, then it makes sense to keep it yeah. You won't hear any arguments from me re: switching to protobuf encoding :)

cylewitruk (approved):

PR #558, opened 2024-09-24
feat: start protobuf use for signer communication

djordon:
## Description

Starts #412, but does not close it.

This PR is just to get the ball rolling by setting this up for protobuf iteration down the line.

## Changes

* Remove the old unused protobufs.
* Remove the relay example binary and associated code since we now use libp2p for communication.
* Add some protobufs for decision types.
* Add a common `Uint256` type for converting between the `[u8; 32]` representation of things as a protobuf. The conversion uses the little endian representation of a `u64`, but I think using big-endian is more "standard". Edit: The [protobufs are encoded](https://protobuf.dev/programming-guides/encoding/#cheat-sheet) as 8-bytes little-endian, and the host machines are probably little endian so this is fine.
* Add protobuf linting to CI.

## Testing Information

Add unit tests for converting between a `Uint256` and a `[u8; 32]`. But still, all of these types are currently unused.

## Checklist:


cylewitruk (approved):
Nice! Lgtm

Issue #552, opened 2024-09-20
[Feature]: Add bitcoin transaction fee validation


djordon:
# Feature - Add bitcoin transaction fee validation

## 1. Description

Whenever we validate a `complete-deposit` or `withdrawal-accept` Stacks transaction, we need to inspect an associated bitcoin transaction to verify that the "fee" used in the Stacks transaction matches what it should be. This is needed to make sure that every element in the Stacks transaction matches up what the signers think that it should be.

### 1.1 Context & Purpose

The validation code in https://github.com/stacks-network/sbtc/pull/545 only checks that the fee is reasonable, not that it is computed correctly. We should add functionality that computes the fee associated with each input and output of a bitcoin transaction. This is also likely needed for validating for https://github.com/stacks-network/sbtc/issues/286.

### 2.1 Acceptance Criteria:
   - [ ] Add functionality that computes the fee associated with each input and output of a bitcoin transaction.
   - [ ] Use this functionality when validating Stacks `complete-deposit` or `withdrawal-accept` transactions.

## 3. Related Issues and Pull Requests (optional):
* https://github.com/stacks-network/sbtc/pull/545
* https://github.com/stacks-network/sbtc/pull/555
* https://github.com/stacks-network/sbtc/issues/182
* https://github.com/stacks-network/sbtc/issues/286

PR #566, opened 2024-09-24
refactor: move the `sbtc::rpc` module over to the signer

djordon:
## Description

This was on the way to #552.

The `sbtc` crate does not need to implement have functionality for fetching transactions from bitcoin-core so this functionality was moved over to the `signer` crate.

## Changes

* Move the `sbtc::rpc` module to the `signer::bitcoin::rpc` module.
* Remove the `sbtc::rpc::BitcoinClient` trait.
* Move the `CreateDepositRequest::validate` function to the `signer` crate.
* Move the `sbtc::logging` module to the `signer::logging` module.
* Modified the imports a little, essentially using `cargo +nightly fmt`.
* Cleaned up the `sbtc` crate features.

## Testing Information

This is just a refactor, with some tests moving between the crates. So if tests pass then we should be good.

## Checklist:


cylewitruk:
Ahh nice! So happy `BitcoinClient` dies here 😄 Will check this out first thing tomorrow 💯

signer/src/error.rs
```diff
@@ -11,6 +11,32 @@ use crate::stacks::contracts::WithdrawalAcceptValidationError;
 /// Top-level signer error
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
```
cylewitruk, line 13: Just a random comment, nothing to do here:

This enum feels like it just keeps growing and growing; I'm thinking we maybe should start thinking about breaking it up. Like, we could have a `BitcoinCore(BitcoinCoreError)` variant which would break out at least a handful of these errors, for example.

cylewitruk (approved):
Lgtm -- I didn't go in deep on the moved code so I'm trusting that was all copy-pasted with minor tweaks due to its new home (recognized a fair bit of it from the recent PRs) :)

signer/src/error.rs
djordon: Not a bad idea, it's quite common to follow such an approach. I'm on board with doing so. 

The main thing that I want to encourage is to preserve context in our error variants. And we can do that by having dedicated error types that each capture enough context. Most of our error variants fall short of capturing enough context, but still that's the aspiration. Also, implementing `From` on the "main" error type keeps things ergonomic so it's all good.

Note that a little history on error handling in the signer crate can be found in https://github.com/stacks-network/sbtc/pull/161#discussion_r1604451042 and https://github.com/stacks-network/sbtc/issues/229.

Issue #538, opened 2024-09-17
[Feature]: Get signers' UTXO


djordon:
# Feature - Get signers' UTXO

## 1. Description

Before the signers can construct a bitcoin transaction, they need to locate the sole UTXO under their control. This UTXO is used as an input into all transactions that sweep in or out BTC for deposit or withdrawal requests. The goal is to implement functionality for the signers to retrieve the relevant information for this UTXO.

### 1.1 Context & Purpose

We need to implement the `BitcoinInteract::get_signer_utxo` function to get the signers' UTXO. There will always be exactly one such UTXO in our database. This function returns a UTXO where the following holds:

1. The transaction is in a block on the canonical bitcoin blockchain.
2. The output is the first output in the transaction.
3. The output's `scriptPubKey` matches one in the stored values in the `dkg_shares.script_pubkey` column.
4. The output is unspent. It is possible for more than one transaction within the same block to satisfy points 1-3, but if the signers have one or more transactions within a block, exactly one output satisfying points 1-3 will be unspent.
5. The block that includes the transaction that satisfies points 1-4 has the greatest height of all such blocks.

## 2. Technical Details:

The transactions in the `transactions` table with `tx_type` set to `'sbtc_transaction'` are all bitcoin transactions where one of the outputs are spendable by the signers. This makes it possible to find a set of transactions that will satisfy the above criteria.

### 2.1 Acceptance Criteria:

- [ ]  The `BitcoinInteract::get_signer_utxo` returns the signers’ UTXO.

## 3. Related Issues and Pull Requests (optional):

PR #567, opened 2024-09-24
feat: implement `get_signer_utxo` for in memory storage

matteojug:
## Description

Initial work for https://github.com/stacks-network/sbtc/issues/538.

I'll add the postgres impl (and more tests, if needed) in a different PR to keep this small, if there are no objections :)

## Changes

 - Move `get_signer_utxo` to storage trait (as the issue suggest to use the database)
 - Implement `get_signer_utxo` for in memory db
 - Add some tests for it

## Testing Information

Since we no longer mock it, `should_be_able_to_coordinate_signing_rounds` now also need sbtc txns to work.

## Checklist:


signer/src/storage/in_memory.rs
```diff
@@ -414,6 +422,97 @@ impl super::DbRead for SharedStore {
...
+                    })
+            })
+            .next();
+
+        if sbtc_txs.is_none() {
+            return Ok(None);
+        }
+        // `sbtc_txs` contains all the txs in the highest canonical block where the first
+        // output is spendable by script_pubkey
+        let sbtc_txs = sbtc_txs.unwrap();
```
djordon, line 477: You could express the same thing without an unwrap, by using the let-else pattern.

signer/src/storage/mod.rs
```diff
@@ -130,6 +131,12 @@ pub trait DbRead {
         &self,
     ) -> impl Future<Output = Result<Vec<model::Bytes>, Error>> + Send;
 
+    /// Get the outstanding signer UTXO
+    fn get_signer_utxo(
```
djordon, line 135: We should probably add a "full" description of what is fetched here.

signer/src/storage/mod.rs
```diff
@@ -130,6 +131,12 @@ pub trait DbRead {
...
+        &self,
+        aggregate_key: &crate::keys::PublicKey,
```
djordon, line 137: This functions should take the block hash of the bitcoin chain tip as an input.

djordon (commented):
Nice. Gave it a quick look over just now and will have a proper review tomorrow. Here is what I have so far.

cylewitruk (commented):
Tossed in a few comments/suggestions. I didn't look too hard at the tests as I think there will be a bit of refactoring there after [this comment](https://github.com/stacks-network/sbtc/pull/567#discussion_r1774393193). Good work so far!

signer/src/storage/mod.rs
matteojug: In addition to the `aggregate_key`, or only that? The issue description would suggest we "infer" (ie, pick all the valid) aggregate_key from the storage, but given it was present in the original fn signature I kept it.

djordon: Oh yeah, I meant in addition to the `aggregate_key`.

djordon (commented):
Looks good.

I think the logic looks sound. But can we add another test for the case where we have a fork with chains of the same length and UTXOs on both? You can either modify an existing test or add a new one.

djordon (approved):
Looks good ✅

Issue #175, opened 2024-05-21
[Feature]: Bitcoin client


netrome:
# Feature - Bitcoin client

### Description
Build a client that implements the bitcoin interface in #117.

djordon:
This is done

PR #573, opened 2024-09-25
refactor: setup bitcoin client use in validation

djordon:
## Description

Addresses part of https://github.com/stacks-network/sbtc/issues/175 sorta, and paves the way for https://github.com/stacks-network/sbtc/issues/552.

While working on https://github.com/stacks-network/sbtc/issues/552, I realized that I needed to make quite a few auxiliary changes, so I stopped when it felt like a lot and decided to open this PR.

## Changes

* Have the `AsContractCall::validate` function take something that implements the `Context` trait. This is needed for https://github.com/stacks-network/sbtc/issues/552.
* Allow for the inner context type to be exposed in tests.
* Make sure the `BitcoinCoreClient` and `StacksClient` implement Clone. This is preparation for changes to the `SignerContext` type, and allows for a simplification of the `InnerSignerContext` object.
* Implement `TryFrom<&Url>` for `BitcoinCoreClient` instead of `TryFrom<Url>`.
* Have `NoopSignerContext::init` take an owned `Settings` object rather than cloning it internally.
* Move the `get_block` implementation into the `BitcoinCoreClient`.

## Testing Information

This is just a refactor to make subsequent work possible (and hopefully simpler). So if this tests pass then we're good.

## Checklist:

### DIFF
--- a/signer/src/bitcoin/client.rs
+++ b/signer/src/bitcoin/client.rs
@@ -66,7 +66,11 @@ impl BitcoinInteract for ApiFallbackClient<BitcoinCoreClient> {
         self.get_client().get_tx(txid)
     }
 
-    fn get_tx_info(&self, txid: &Txid, block_hash: &BlockHash) -> Result<BitcoinTxInfo, Error> {
+    fn get_tx_info(
+        &self,
+        txid: &Txid,
+        block_hash: &BlockHash,
+    ) -> Result<Option<BitcoinTxInfo>, Error> {
         self.get_client().get_tx_info(txid, block_hash)
     }
 
--- a/signer/src/bitcoin/client.rs
+++ b/signer/src/bitcoin/client.rs
@@ -14,6 +14,8 @@
 //! - Example when trying to get a block that doesn't exist:
 //!   JsonRpc(Rpc(RpcError { code: -5, message: "Block not found", data: None }))
 
+use bitcoin::BlockHash;
+use bitcoin::Txid;
 use bitcoincore_rpc::jsonrpc::error::RpcError;
 use bitcoincore_rpc::RpcApi as _;
 use url::Url;
@@ -26,6 +28,7 @@ use crate::keys::PublicKey;
 use crate::util::ApiFallbackClient;
 
 use super::rpc::BitcoinCoreClient;
+use super::rpc::BitcoinTxInfo;
 use super::rpc::GetTxResponse;
 
 /// Implement the [`TryFrom`] trait for a slice of [`Url`]s to allow for a
@@ -59,8 +62,12 @@ impl BitcoinInteract for ApiFallbackClient<BitcoinCoreClient> {
         .await
     }
 
-    fn get_tx(&self, txid: &bitcoin::Txid) -> Result<GetTxResponse, Error> {
+    fn get_tx(&self, txid: &Txid) -> Result<GetTxResponse, Error> {
         self.get_client().get_tx(txid)
+    }
+
+    fn get_tx_info(&self, txid: &Txid, block_hash: &BlockHash) -> Result<BitcoinTxInfo, Error> {
+        self.get_client().get_tx_info(txid, &block_hash)
     }
 
     async fn estimate_fee_rate(&self) -> Result<f64, Error> {
--- a/signer/src/bitcoin/client.rs
+++ b/signer/src/bitcoin/client.rs
@@ -38,7 +38,7 @@ impl TryFrom<&[Url]> for ApiFallbackClient<BitcoinCoreClient> {
     fn try_from(urls: &[Url]) -> Result<Self, Self::Error> {
         let clients = urls
             .iter()
-            .map(|url| BitcoinCoreClient::try_from(url.clone()))
+            .map(BitcoinCoreClient::try_from)
             .collect::<Result<Vec<_>, _>>()?;
 
         Self::new(clients).map_err(Into::into)
--- a/signer/src/bitcoin/client.rs
+++ b/signer/src/bitcoin/client.rs
@@ -62,7 +62,7 @@ impl BitcoinInteract for ApiFallbackClient<BitcoinCoreClient> {
         .await
     }
 
-    fn get_tx(&self, txid: &Txid) -> Result<GetTxResponse, Error> {
+    fn get_tx(&self, txid: &Txid) -> Result<Option<GetTxResponse>, Error> {
         self.get_client().get_tx(txid)
     }
 
--- a/signer/src/bitcoin/client.rs
+++ b/signer/src/bitcoin/client.rs
@@ -16,7 +16,6 @@
 
 use bitcoin::BlockHash;
 use bitcoin::Txid;
-use bitcoincore_rpc::jsonrpc::error::RpcError;
 use bitcoincore_rpc::RpcApi as _;
 use url::Url;
 
@@ -50,16 +49,8 @@ impl BitcoinInteract for ApiFallbackClient<BitcoinCoreClient> {
         &self,
         block_hash: &bitcoin::BlockHash,
     ) -> Result<Option<bitcoin::Block>, Error> {
-        self.exec(|client| async {
-            match client.inner_client().get_block(block_hash) {
-                Ok(block) => Ok(Some(block)),
-                Err(bitcoincore_rpc::Error::JsonRpc(bitcoincore_rpc::jsonrpc::Error::Rpc(
-                    RpcError { code: -5, .. },
-                ))) => Ok(None),
-                Err(error) => Err(Error::BitcoinCoreRpc(error)),
-            }
-        })
-        .await
+        self.exec(|client| async { client.get_block(block_hash) })
+            .await
     }
 
     fn get_tx(&self, txid: &Txid) -> Result<Option<GetTxResponse>, Error> {
--- a/signer/src/bitcoin/client.rs
+++ b/signer/src/bitcoin/client.rs
@@ -67,7 +67,7 @@ impl BitcoinInteract for ApiFallbackClient<BitcoinCoreClient> {
     }
 
     fn get_tx_info(&self, txid: &Txid, block_hash: &BlockHash) -> Result<BitcoinTxInfo, Error> {
-        self.get_client().get_tx_info(txid, &block_hash)
+        self.get_client().get_tx_info(txid, block_hash)
     }
 
     async fn estimate_fee_rate(&self) -> Result<f64, Error> {
--- a/signer/src/bitcoin/rpc.rs
+++ b/signer/src/bitcoin/rpc.rs
@@ -1,5 +1,7 @@
 //! Contains client wrappers for bitcoin core and electrum.
 
+use std::sync::Arc;
+
 use bitcoin::Amount;
 use bitcoin::BlockHash;
 use bitcoin::Denomination;
@@ -15,7 +17,9 @@ use bitcoincore_rpc_json::GetRawTransactionResultVoutScriptPubKey as BitcoinTxIn
 use serde::Deserialize;
 use url::Url;
 
+use crate::bitcoin::BitcoinInteract;
 use crate::error::Error;
+use crate::keys::PublicKey;
 
 /// A slimmed down type representing a response from bitcoin-core's
 /// getrawtransaction RPC.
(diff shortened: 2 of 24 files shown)


signer/src/bitcoin/rpc.rs
```diff
@@ -151,17 +160,18 @@ pub struct FeeEstimate {
 }
 
 /// A client for interacting with bitcoin-core
+#[derive(Debug, Clone)]
 pub struct BitcoinCoreClient {
     /// The underlying bitcoin-core client
-    inner: bitcoincore_rpc::Client,
+    inner: Arc<bitcoincore_rpc::Client>,
```
cylewitruk, line 166: I'm not sure this is actually needed since it will be used via the FallbackClient which does this internally? My thinking was to keep this complexity out of the clients themselves and aggregate that all up to the fallback client.

signer/src/bitcoin/rpc.rs
```diff
@@ -275,3 +318,40 @@ impl BitcoinCoreClient {
         Ok(FeeEstimate { sats_per_vbyte })
     }
 }
+
+impl BitcoinInteract for BitcoinCoreClient {
```
cylewitruk, line 322: Any particular reason to re-implement this here when it's on `ApiFallbackClient<BitcoinCoreClient>` in `client.rs`? Just asking, I had removed it as I figured that the only real usage will be via the fallback client and so left only the "internal" implementations here. Of course moving the "actual" `BitcoinInteract` implementations here and calling into them via the fallback-client's implementation is another option -- just more code.

signer/src/bitcoin/client.rs
```diff
@@ -47,22 +49,22 @@ impl BitcoinInteract for ApiFallbackClient<BitcoinCoreClient> {
...
+    fn get_tx(&self, txid: &Txid) -> Result<Option<GetTxResponse>, Error> {
         self.get_client().get_tx(txid)
     }
 
+    fn get_tx_info(
+        &self,
+        txid: &Txid,
+        block_hash: &BlockHash,
+    ) -> Result<Option<BitcoinTxInfo>, Error> {
+        self.get_client().get_tx_info(txid, block_hash)
```
cylewitruk, line 65: Is it intentional to skip failover/retry here?

signer/src/stacks/contracts.rs
```diff
@@ -286,14 +287,14 @@ impl AsContractCall for CompleteDepositV1 {
     /// fortunate, because even if we wanted to, the only view into the
     /// stacks-core mempool is through the `POST /new_mempool_tx` webhooks,
     /// which we do not currently ingest.
-    async fn validate<S>(&self, db: &S, ctx: &ReqContext) -> Result<(), Error>
+    async fn validate<S>(&self, db: &S, req_ctx: &ReqContext) -> Result<(), Error>
     where
-        S: DbRead + Send + Sync,
+        S: Context + Send + Sync,
```
cylewitruk, line 292: `S` -> `C` (like you did above)

signer/src/bitcoin/rpc.rs
```diff
@@ -194,16 +205,31 @@ impl BitcoinCoreClient {
...
     /// mempool. If -txindex is enabled on bitcoin-core and no blockhash
     /// argument is passed, it will return the transaction if it is in the
-    /// mempool or any block.
-    pub fn get_tx(&self, txid: &Txid) -> Result<GetTxResponse, Error> {
+    /// mempool or any block. We require -txindex to be enabled (same with
+    /// stacks-core[^1]) so this should work with transactions in either
+    /// the mempool and a bitcoin block.
+    ///
+    /// [^1]: <https://docs.stacks.co/guides-and-tutorials/run-a-miner/mine-mainnet-stacks-tokens>
+    pub fn get_tx(&self, txid: &Txid) -> Result<Option<GetTxResponse>, Error> {
```
cylewitruk, line 232: Do we really need both `get_tx` and `get_tx_info`? Wouldn't it be better to just have `get_tx_info` with verbosity 2 which is always used?

signer/tests/integration/withdrawal_accept.rs
```diff
@@ -243,13 +244,14 @@ async fn accept_withdrawal_validation_happy_path() {
...
+    let (accept_withdrawal_tx, req_ctx) =
         make_withdrawal_accept(&req, sweep_outpoint, aggregate_key, &chain_tip, bitmap);
 
     // This should not return an Err.
-    accept_withdrawal_tx.validate(&db, &ctx).await.unwrap();
+    let ctx = TestSignerContext::from_db(db);
+    accept_withdrawal_tx.validate(&ctx, &req_ctx).await.unwrap();
 
-    testing::storage::drop_db(db).await;
+    testing::storage::drop_db(ctx.into_db()).await;
```
cylewitruk, line 254: `PgStore` is `Clone` so you could probably just have done that instead of `into_db()` (at least right now since both `new_test_db()` and `drop_db()` work explicitly on `PgStore`).

cylewitruk (commented):
Had a few comments -- some of this is stuff I also have in my local branch (moving context into some of these types, this will be fun to resolve :laugh:) and seems like we have some differences of approaches in a couple of areas -- let's hash those out :smile:

signer/src/context/mod.rs
```diff
@@ -106,22 +98,28 @@ where
...
                 term_tx,
                 storage: db,
                 bitcoin_client,
-            }),
-        })
+            },
+        }
+    }
+
+    /// Drop the context and retyur
```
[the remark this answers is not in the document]
djordon: That's my preference too, but what I needed required access to the private field, which isn't available in another module.

signer/src/bitcoin/rpc.rs
djordon: So the `getrawtransaction` RPC's return type differs based a number of conditions:
1. Whether the transaction is in a bitcoin block or the mempool.
2. Whether the block hash that includes the transaction is supplied in the call.
3. The verbosity level supplied in the call.

and probably other stuff too, like whether bitcoin-core is configured with `txindex=1`. 

Later on we need to fetch transactions where we do not know the bitcoin block that includes the transaction (and it might not even be in a bitcoin block), so we cannot always supply the bitcoin block hash. When that happens bitcoin-core won't return some of the fields in the `BitcoinTxInfo` struct. Instead of making those fields optional, I felt that it's better to have separate functions with dedicated types for the expected responses instead of a single type that handles all cases.

So using `get_tx_info` always is not possible, we need to know the bitcoin block hash. And making the bitcoin block hash optional requires us to unify the expected return type, which I felt was less than desirable.

signer/tests/integration/withdrawal_accept.rs
djordon: I think you need to drop all connections to the database before you can drop it. Maybe that can be forced, I forget 🤔, but that is why I went this route. Let me double check on whether or not you can drop it with existing connections.

signer/src/bitcoin/rpc.rs
djordon: Oh I believe I added it because it was necessary for some type bounds. But I might have subsequently changed something that makes it unnecessary now. I'll need to check.

signer/src/context/mod.rs
```diff
@@ -39,16 +37,9 @@ pub trait Context: Clone + Sync + Send {
...
-    inner: Arc<InnerSignerContext<S, BC>>,
-}
-
-/// We implement [`Clone`] manually to avoid the derive macro adding additional
-/// bounds on the generic types.
-impl<S, BC> Clone for SignerContext<S, BC> {
-    fn clone(&self) -> Self {
-        Self { inner: Arc::clone(&self.inner) }
-    }
+    inner: InnerSignerContext<S, BC>,
```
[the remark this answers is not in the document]
djordon: Yeah, I agree. And the ones that we plan on adding will implement `Clone`. The only one that didn't already was the `BitcoinCoreClient`. I think the others already use `Arc` under the hood.

signer/src/bitcoin/rpc.rs
djordon: Oh I added this in so that I can use the plain vanilla `BitcoinCoreClient` in the `SignerContext` in my tests.

> My thinking was to keep this complexity out of the clients themselves and aggregate that all up to the fallback client.

Yeah 🤔. My implicit goal is that we can choose to not use the fallback client in tests, since then we would have fewer layers between the code that we want to test and the actual test.

signer/tests/integration/withdrawal_accept.rs
djordon: Oh I already force drop the connections 😅, so this can be simplified.

signer/src/context/mod.rs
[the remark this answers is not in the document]
djordon: Removed in https://github.com/stacks-network/sbtc/pull/573/commits/76bd8cd97d4d1eebc0937058f7a33c32c24c149c

signer/src/bitcoin/client.rs
djordon: Oh right! Yeah, it was the whole `async` thing; I tried but couldn't because `exec` is `async`. Something felt not right about forcing the trait function to be async when it doesn't need to be, and I just figured that when I revisit it the right path will be obvious.

signer/src/bitcoin/rpc.rs
djordon: Yeah it was necessary in order to use the `SignerContext` with the `BitcoinCoreClient` in the tests.

signer/src/stacks/contracts.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/573/commits/5282a4f33cf0bafb97a5d49624e78db80632c4cb.

signer/src/bitcoin/client.rs
djordon: You know, I'll just make it all `async` and call it a day

signer/src/bitcoin/rpc.rs
cylewitruk: Yeah I get that goal as well, always tradeoffs :sweat_smile:. I think we should agree on a solution then since we've got two kinda right now:

1. That we keep "internal implementations" in the `*Client`s, always implement `*Interact` on the fallback client and use its `get_client()` method to get access to the inner client in tests. Then we say that thread-safety is handled by the fallback client, or
2. We always implement both the "internal implementations" and `*Interact` on the underlying clients. `*Iteract` is also implemented on the fallback client which calls-back into those implementations (within its `exec`), and the underlying client must provide thread-safety.

signer/src/bitcoin/rpc.rs
cylewitruk: Yeah, got it :+1:

signer/src/bitcoin/client.rs
djordon: Oh I must have went down this path and realized that it was more work than it is worth. So yyyyeah, here we are.

cylewitruk: Ah, gotcha. Personally I think that all of the client trait methods should be async because we don't "know" what that underlying implementation is doing and in this case the mix/match of async feels off, like we're explicitly tying the trait to the current underlying impl.

Actually, it wouldn't be stupid to make the internals `async` with a `tokio::spawn` around the API calls since they are I/O.

signer/src/bitcoin/rpc.rs
cylewitruk: Aha okay, I didn't realize it was such a big difference between verbosity levels. Weird design :sweat_smile:

signer/src/context/mod.rs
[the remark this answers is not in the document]
cylewitruk: We probably don't need the whole `SignerContext` + `InnerSignerContext` dance anymore in this case either

signer/src/bitcoin/client.rs
djordon: > Ah, gotcha. Personally I think that all of the client trait methods should be async because we "don't know" what that underlying implementation is doing and in this case the mix/match of async feels off, like we're explicitly tying the trait to the current underlying impl.
> 
> Actually, it wouldn't be stupid to make the internals `async` with a `tokio::spawn` around the API calls since they are indeed I/O.

Yeah, this is the better argument and I agree with it.

I did try to make the change but it was annoying and would be more time-consuming than I felt it was worth at this point.

cylewitruk: We can leave it for now then and attack it in some upcoming PR :)

signer/src/bitcoin/rpc.rs
djordon: I'd vote for (2). With the exception of the `BitcoinCoreClient`, I thought the clients were all thread-safe already and already implemented their `*Interact` traits. `BitcoinCoreClient` was the only one that didn't so this brings things back into alignment.

Wait, that was a little misleading. We don't have an emily client, but that should just be a `reqwest::Client` wrapper so it will be thread safe.

cylewitruk: Cool, let's roll with no. 2 then :) I'll make the changes over in my stacks-client PR as well

cylewitruk (approved):
Looks good -- up to you if you want to refactor away the `InnerSignerContext` now or if we do it later

signer/src/bitcoin/client.rs
cylewitruk: > Oh I must have went down this path and realized that it was more work than it is worth. So yyyyeah, here we are.

Oh for some reason I didn't see this prior to my last two replies 😅 Sorry

PR #577, opened 2024-09-26
feat: add more protobuf types part 2

djordon:
## Description

This PR pushes https://github.com/stacks-network/sbtc/issues/412 further along and a follow-up to https://github.com/stacks-network/sbtc/pull/558.

The documentation was taken from the existing structs.

## Changes

* Add over a some transaction message types.
* Add a recoverable signature message type.
* Split things up into different files.

## Testing Information

There protobufs aren't used, so there are no tests at the moment.

## Checklist:


protobufs/stacks/common.proto
```diff
@@ -2,23 +2,45 @@ syntax = "proto3";
...
+
+// The id for a transaction on the bitcoin blockchain.
+message BitcoinTxid {
+  crypto.Uint256 txid = 1;
+}
+
+// A bitcoin block hash.
+message BitcoinBlockHash {
+  crypto.Uint256 block_hash = 1;
+}
```
cylewitruk, line 15: Feels a _little_ weird that these are part of the `stacks` package

cylewitruk (commented):
Looks good in general, since these aren't used (yet) I'm guessing that these will be iterated upon later. Just had one comment re: the bitcoin types.

protobufs/stacks/common.proto
djordon: Yeah I was thinking the same thing. When I was opening the PR. Figured I'd move them later. I'll just move them now.

djordon: Fixed in https://github.com/stacks-network/sbtc/pull/577/commits/291b72e10d107d675ff7b553618a73410c45cb13

cylewitruk (approved):
looks good :+1:

PR #580, opened 2024-09-27
feat: run block observer

cylewitruk:
## Description

Closes: #548 

## Changes

- Adds the block observer to the main entry-point
- Changes the block observer db-update notification to use the app signalling channel (and updated tests)
- Removes `Storage` from the `BlockObserver` and adds `Context` (storage is retrieved from the context)
- Fixed an issue in block observer tests where the test harness returned a stream of non-result items
- Added configuration for bitcoin zmq block hash stream endpoints
- Updated the run-loop for the block observer

TODO: Add validation for block stream config

## Testing Information

Same test suite as before.

## Checklist:

### DIFF
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -21,6 +21,7 @@ use std::collections::HashMap;
 
 use crate::bitcoin::BitcoinInteract;
 use crate::context::Context;
+use crate::context::SignerEvent;
 use crate::error::Error;
 use crate::stacks::api::StacksInteract;
 use crate::storage;
@@ -41,7 +42,7 @@ use std::collections::HashSet;
 
 /// Block observer
 #[derive(Debug)]
-pub struct BlockObserver<Context, StacksClient, EmilyClient, BlockHashStream, Storage> {
+pub struct BlockObserver<Context, StacksClient, EmilyClient, BlockHashStream> {
     /// Signer context
     pub context: Context,
     /// Stacks client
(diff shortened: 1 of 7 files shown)


signer/src/block_observer.rs
```diff
@@ -369,24 +395,20 @@ mod tests {
             storage.clone(),
             test_harness.clone(),
         );
+        let _signal_rx = ctx.get_signal_receiver();
```
djordon, line 398: Oh, this is needed to make sure that all receivers aren't dropped. Maybe add a comment here?

signer/src/context/messaging.rs
```diff
@@ -22,6 +22,8 @@ pub enum SignerCommand {
 pub enum SignerEvent {
     /// Signals that a P2P event has occurred.
     P2P(P2PEvent),
+    /// Signals that a block observer event has occurred.
+    BlockObserverUpdated,
```
djordon, line 26: I would rename this to `BitcoinBlockObserved`.

djordon (approved):
I would rename the new `SignerEvent` variant but besides that, looks good.

signer/src/block_observer.rs
cylewitruk: Yeah exactly, haha I had considered it teetered towards not 😅 Yeah I'll add one

signer/src/context/messaging.rs
cylewitruk: Ah perfect, was fumbling after a good name 😊

cylewitruk: Updated in [e7064d1](https://github.com/stacks-network/sbtc/pull/580/commits/e7064d1c891fc1e5d30906dc33ba69bed4970f6e)

signer/src/block_observer.rs
cylewitruk: Updated in [e7064d1](https://github.com/stacks-network/sbtc/pull/580/commits/e7064d1c891fc1e5d30906dc33ba69bed4970f6e)

PR #581, opened 2024-09-28
feat: emily integration boilerplate

cylewitruk:
## Description

This provides some of the boilerplate needed to fully implement the Emily API. Right now it's just mocked, so the actual implementation using the generated client lib still needs to be written.

If we're going to Mock emily in the first release, I figure we could at least use a `moquito` server or something so that it's actually using the real client & config etc.  while still letting us control what's returned for different requests.

## Changes

- Adds an `EmilyClient` (which is intended to be a wrapper around the generated stuff -- there's multiple clients generated, so we can collect all of them here).
- Adds the `ApiFallbackClient` implementation and related `From<>`'s.
- `EmilyInteract` is moved into this same file (it was temporary in `block_observer.rs`) and implements it on `EmilyClient` and `ApiFallbackClient<EmilyClient>`. Note that the implementation is currently mocked as we can't build the correct types from the received information (yet?).
- Adds Emily config support.
- Updated some of the testing facilities and reduced the number of mock implementations of `EmilyInteract`.

## Testing Information

Nothing major to note.

## Checklist:


signer/src/config/default.toml
```diff
@@ -7,6 +7,23 @@
...
+# You may specify multiple Emily API servers if you have them. They will be
+# tried round-robin until one succeeds.
+#
+# Format: ["http(s)://<host>:<port>", ..]
+# Default: <none>
+# Required: true
+# Environment: SIGNER_EMILY__ENDPOINTS
+endpoints = [
+    "http://localhost:3000"
+]
```
djordon, line 25: I think Emily will have only one endpoint, but I suppose that fact can change.

djordon (approved):
✅ Looks good!

signer/src/config/default.toml
cylewitruk: Yeah... it just pains me in my soul to code a single-point-of-failure/trust into a distributed "trustless" system :laughing:

Issue #559, opened 2024-09-24
[Feature]: Link stacks blocks with bitcoin blocks


djordon:
# Feature - Link Stacks blocks with Bitcoin blocks

## 1. Description

The signer data model ties Stacks (and bitcoin) transaction data to the bitcoin blockchain. So when there is a reorg of the bitcoin blockchain, we are supposed to know which Stacks transactions have been invalidated. The problem is we haven't yet explicitly linked Stacks transactions to the bitcoin blockchain.

This issue is about completing the final step of linking Stacks blocks to Bitcoin blocks, and thus the canoncial bitcoin blockchain.

### 1.1 Context & Purpose

Knowing which Stacks transactions are invalid is needed for figuring out a number of important things, such as:
1. Whether a withdrawal request has actually occurred.
2. Whether we still need to mint sBTC for an accepted deposit request.
3. Whether the last step of the withdrawal request has been fulfilled.

And probably other stuff too.

## 2. Technical Details:

The signer already listens for new Stacks blocks via the stacks-core event observer. These stacks-core new block webhooks include the chain tip of the bitcoin blockchain in the message, so storing that would get us a long way there for linking Stacks blocks to bitcoin blocks.

The remaining work is in ensuring that we can still link Stacks blocks with bitcoin blocks if a signer misses these webhooks for any reason. This can be done by using one of the newly added endpoints to stacks-core for fetching "sortition" info, which includes the associated bitcoin block for a consensus hash.

### 2.1 Acceptance Criteria:
   - [ ] For each Stacks block, store the associated bitcoin block as part of processing `POST /new_block` webhooks.
   - [ ] We add functionality for fetching the latest sortition information from stacks-core, which returns the bitcoin block that is tied to a consensus hash (and consensus hashes are included in Nakamoto block headers).
   - [ ] We fetch and store the associated bitcoin block when doing any backfilling of the stacks block chain.

## 3. Related Issues and Pull Requests (optional):

The stacks-core event observer was added in https://github.com/stacks-network/sbtc/pull/474.

Solving this ticket would help solve these issues:
* https://github.com/stacks-network/sbtc/issues/524
* https://github.com/stacks-network/sbtc/issues/543
* https://github.com/stacks-network/sbtc/issues/554

PR #582, opened 2024-09-30
feat: add bitcoin transaction fee validation

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/552.

## Changes

* Use from instead of `as _` casts for integers more
* Use the crate `ScriptPubKey` in `EncryptedDkgShares`
* Update the test config bitcoin-core endpoint to use the regtest username and password.
* Add a new function to the `DbRead` trait for checking whether a script pub key is related to the signers
* Update the function signature on `BitcoinTxInfo`. Also update a `BitcoinTxInfo`'s subfield and comment
* Fix lurking bugs in two queries.
* Update stacks validation functions to use bitcoin-core for fetching fee information. Also, check that the sweep transaction was created by the signers.
* Add test helper struct for setting up sweep transactions and making them (hopefully) easier to understand.

## Testing Information

This PR updates the integration tests in validation to be a little easier to understand. The downsides to these tests is that the blockchain in the database does not contain forks by default. Nonetheless, the increase in clarity should more than make up for that fact.

## Checklist:


signer/tests/integration/complete_deposit.rs
```diff
@@ -81,119 +78,54 @@ fn make_complete_deposit(
...
+    // Normal: the signer follows the bitcoin blockchain and event observer
+    // should be getting new block events from bitcoin-core. We haven't
+    // hooked up our block observer so we need to manually update the
+    // database with new bitcoin block headers.
+    backfill_bitcoin_blocks(&db, rpc, &setup.sweep_block_hash).await;
+
+    // Normal: we take the deposit transaction as is from the test setup
+    // and store it in the database. This is necessary for when we fetch
+    // outstanding unfulfilled deposit requests.
+    setup.store_deposit_tx(&db).await;
```
djordon, line 102: At the moment, I do not think that this step is actually necessary since we get the deposit transaction from bitcoin-core. I discovered this along the way and just decided to keep it in, but I don't mind removing it.

signer/tests/integration/complete_deposit.rs
```diff
@@ -81,119 +78,54 @@ fn make_complete_deposit(
...
+
+    // Normal: we take the sweep transaction as is from the test setup and
+    // store it in the database.
+    setup.store_sweep_tx(&db).await;
+
+    // Normel: we need to store a row in the dkg_shares table so that we
```
matteojug, line 108:
```suggestion
    // Normal: we need to store a row in the dkg_shares table so that we
```
also below

signer/tests/integration/setup.rs
```diff
@@ -0,0 +1,351 @@
...
+    pub sweep_tx_info: BitcoinTxInfo,
+    /// The withdrawal request, and a bitmap for how the siggners voted on
+    /// it.
+    pub withdrawal_request: utxo::WithdrawalRequest,
+    /// The address that initiated with withdrawal request.
+    pub withdrawal_sender: PrincipalData,
+}
+
+impl TestSweepSetup {
+    /// Construct a new TestDepositSetup
```
matteojug, line 70:
```suggestion
    /// Construct a new TestSweepSetup
```

signer/tests/integration/setup.rs
```diff
@@ -0,0 +1,351 @@
...
+            db.write_withdrawal_signer_decision(&decision)
+                .await
+                .unwrap();
+        }
+    }
+
+    pub async fn store_withdrawal_request(&self, db: &PgStore) {
+        let block = model::StacksBlock {
+            block_hash: self.withdrawal_request.block_hash,
+            block_height: self.sweep_block_height,
```
matteojug, line 287: nit: I guess is not really important here, but isn't `sweep_block_height` related to the bitcoin block (and not the stacks one)? If so, could it be the case that we end up writing a stacks block that is height-wise disconnected from the rest of the chain (and then maybe break something else)?

djordon: Yeah, it's possible. This part of the test setup is non-intuitive and a little brittle. But the bigger issue is that we need to change how we write stacks blocks to the database (as part of https://github.com/stacks-network/sbtc/issues/559) so doing this "correctly" for these tests felt like wasted effort.

signer/tests/integration/complete_deposit.rs
```diff
@@ -1,75 +1,72 @@
...
         // looking for pending and accepted deposit requests.
         context_window: 10,
         // The value here doesn't matter.
         origin: fake::Faker.fake_with_rng(&mut OsRng),
-        // This value doesn't matter here.
-        aggregate_key: fake::Faker.fake_with_rng(&mut OsRng),
+        // When checking whether the transaction is from the signer, we
+        // check that the first "prevout" has a `scriptPubKey` that the
+        // signers control.
+        aggregate_key: data.signer.keypair.public_key().into(),
```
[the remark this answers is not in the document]
djordon: Yeah that works. Or `controlling_signer`.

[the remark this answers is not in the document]
djordon: yeah `aggregated_signer` is way better

signer/tests/integration/complete_deposit.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/582/commits/8accf065fce0444f491315a3eb672523aecb61a0.

signer/tests/integration/complete_deposit.rs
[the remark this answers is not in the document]
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/582/commits/21ed295d7d378bea8d0eecf8e68031cd070b0e13

signer/tests/integration/complete_deposit.rs
```diff
@@ -1,75 +1,72 @@
...
         // The block hash of the block that includes the above sweep
         // transaction. It must be on the canonical bitcoin blockchain.
-        sweep_block_hash: chain_tip.block_hash,
+        sweep_block_hash: data.sweep_block_hash.into(),
         // This must be the height of the above block.
-        sweep_block_height: chain_tip.block_height,
+        sweep_block_height: data.sweep_block_height,
     };
 
     // This is what the current signer things of the state of things.
```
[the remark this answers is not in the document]
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/582/commits/21ed295d7d378bea8d0eecf8e68031cd070b0e13

signer/tests/integration/setup.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/582/commits/21ed295d7d378bea8d0eecf8e68031cd070b0e13

signer/src/testing/wallet.rs
```diff
@@ -46,6 +49,35 @@ pub fn generate_wallet() -> (SignerWallet, [Keypair; 3], u16) {
...
+    assert_eq!(num_signers % 2, 1);
+
+    let private_key = PrivateKey::from(signer.keypair.secret_key());
+    let aggregate_key = PublicKey::from_private_key(&private_key);
+    // The private keys of half of the other signers
+    let pks = [
+        secp256k1::SecretKey::new(rng),
+        secp256k1::SecretKey::new(rng),
+        secp256k1::SecretKey::new(rng),
+    ];
```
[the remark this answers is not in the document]
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/582/commits/ced05eb75ff7b5a90c58c610b613e2e8ffa9604c

signer/src/stacks/contracts.rs
```diff
@@ -311,12 +319,15 @@ impl CompleteDepositV1 {
     /// 5. That the recipients in the transaction matches that of the
     ///    deposit request.
     /// 6. That the amount to mint does not exceed the deposit amount.
-    /// 7. That the max-fee is less than the desired max-fee.
-    async fn validate_deposit_vars<C>(&self, ctx: &C, req_ctx: &ReqContext) -> Result<(), Error>
+    /// 7. That the fee matches the expected assessed fee for the output.
+    /// 8. That the max-fee is less than the desired max-fee.
```
matteojug, line 323: Those two points numbers are swapped wrt the docs of `CompleteDepositV1::validate`

signer/src/stacks/contracts.rs
```diff
@@ -371,16 +383,20 @@ impl CompleteDepositV1 {
...
-            .get_bitcoin_tx(&self.sweep_txid, &self.sweep_block_hash)
-            .await?
-            .ok_or_else(|| DepositErrorMsg::SweepTransactionMissing.into_error(req_ctx, self))?;
+        let rpc = ctx.get_bitcoin_client();
+        // First we check that bitcoin-core has a record of the transaction
+        // where we think it should be.
+        let txid = &self.sweep_txid;
+        let Some(sweep_tx) = rpc.get_tx_info(txid, &self.sweep_block_hash).await? else {
+            return Err(DepositErrorMsg::SweepTransactionMissing.into_error(req_ctx, self));
+        };
```
matteojug, line 399: qq: here we rely on bitcoin-core, while usually(?) we seem to take our storage as source of truth; mixing the two could cause some not-so-fun-to-track coherence issues, so why are we not trusting our storage here?

signer/src/stacks/contracts.rs
```diff
@@ -277,6 +280,10 @@ impl AsContractCall for CompleteDepositV1 {
     ///    deposit request.
     /// 6. That the amount to mint does not exceed the deposit amount.
     /// 7. That the fee is less than the desired max-fee.
+    /// 8. That the fee matches the expected assessed fee for the output.
+    /// 9. That the first input into the sweep transaction is the signers'
+    ///    UTXO. This checks that the sweep transaction was generated by
+    ///    the signers.
```
matteojug, line 286: Should we also add this check when fetching the signers UTXO? Otherwise, looking at the constraints in https://github.com/stacks-network/sbtc/issues/538, if someone directly pays `scriptPubKey` (is it possible?) then that would become the new UTXO. (not related to this PR, if needed I'll open an issue)

matteojug (approved):

signer/src/stacks/contracts.rs
djordon: Yes (!), we need that check there too. Anyone can spend to a `scriptPubKey` so we need to check one of the inputs. I had forgotten about that constraint when writing that ticket. I'll open a new one to add that in.

signer/src/stacks/contracts.rs
djordon: Yeah, we usually only reach out to the database for this information. Here I used bitcoin-core because I felt that it was much easier than using the database. Specifically, I had thought that I would need to write and test a few more queries before we could get the information that we need. But, I probably miscalculated and maybe we just need one more query.

And I'm not so worried about coherence issues (well, at least not yet). Here the RPC might succeed while the equivalent queries might fail because we haven't processed the bitcoin block yet (but I'm not worried about that either). However, I am worried about bitcoin-core not computing the "UndoData" that is necessary for the RPC to actually work as intended. So when we have time we should fix this up to use the database.

signer/src/storage/mod.rs
```diff
@@ -127,7 +127,9 @@ pub trait DbRead {
         chain_tip: &model::BitcoinBlockHash,
     ) -> impl Future<Output = Result<Option<model::RotateKeysTransaction>, Error>> + Send;
 
-    /// Get the last 365 days worth of the signers' `scriptPubkey`s.
+    /// Get the last 365 days worth of the signers' `scriptPubkey`s. If no
```
cylewitruk, line 130: Just curious, why specifically 1yr?

signer/src/testing/wallet.rs
```diff
@@ -46,6 +49,34 @@ pub fn generate_wallet() -> (SignerWallet, [Keypair; 3], u16) {
...
+    R: rand::Rng,
+{
+    // We only take an odd number of signers so that the math works out.
+    assert_eq!(num_signers % 2, 1);
+
+    let private_key = PrivateKey::from(signer.keypair.secret_key());
+    let aggregate_key = PublicKey::from_private_key(&private_key);
+    // The private keys of half of the other signers
+    let pks: Vec<secp256k1::SecretKey> = std::iter::repeat_with(|| secp256k1::SecretKey::new(rng))
+        .take(num_signers / 2)
```
cylewitruk, line 65: I guess `num_signers` is meant to be _exlusive of_ the current signer?

signer/tests/integration/postgres.rs
```diff
@@ -1268,6 +1271,89 @@ async fn we_can_fetch_bitcoin_txs_from_db() {
...
+
+    let mut rng = rand::rngs::StdRng::seed_from_u64(51);
+
+    let shares: model::EncryptedDkgShares = fake::Faker.fake_with_rng(&mut rng);
+
+    sqlx::query(
+        r#"
+        INSERT INTO sbtc_signer.dkg_shares (
+            aggregate_key
+            , tweaked_aggregate_key
```
cylewitruk, line 1333: Is `tweaked` an official term here? :laughing: (no but more seriously, what does `tweaked` mean here?)

signer/tests/integration/complete_deposit.rs
cylewitruk: I think this hangs a little together with @matteojug's Q regarding using multiple data-sources.

signer/src/testing/wallet.rs
djordon: No, it is supposed to include the current signer. Why do you ask?

signer/tests/integration/complete_deposit.rs
```diff
@@ -442,60 +364,63 @@ async fn complete_deposit_validation_invalid_mint_amount() {
...
-        make_complete_deposit(&deposit_req, &sweep_tx, &chain_tip);
-    // Different: The amount cannot be less than the deposit amount less
-    // the max-fee.
-    complete_deposit_tx.amount = deposit_req.amount - deposit_req.max_fee - 1;
+    let (rpc, faucet) = regtest::initialize_blockchain();
+    let mut setup = TestSweepSetup::new_setup(&rpc, &faucet, 1_000_000, &mut rng);
+
+    // Normal: the signer follows the bitcoin blockchain and event observer
+    // should be getting new block events from bitcoin-core. We haven't
+    // hooked up our block observer, so we need to manually update the
```
cylewitruk, line 383: This should be "hooked-up" now?

signer/src/testing/wallet.rs
cylewitruk: Nah it was just because this isn't technically half of the other signers , but I guess that doesn't really matter here since after looking again you're asserting it's odd above which solves it 👍

signer/tests/integration/postgres.rs
djordon: lol, yeah it is an official term (it's used in the official taproot BIPs, like [BIP-341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki), where they detail the computations). It means that we've modified a number in such a way where we can still recover the original number if we know the modification. So in this case, we have a public key (the aggregate key) which was modified by adding some other number (which here is some hash of something or other).

I also found this simple definition from an [ipython notebook](https://github.com/bitcoinops/taproot-workshop/blob/master/2.2-taptweak.ipynb):
> Tweaking a public key means to alter it with a value (the tweak) such that it remains spendable with knowledge of the original private key and tweak.

signer/tests/integration/withdrawal_accept.rs
```diff
@@ -1,84 +1,79 @@
...
-    aggregate_key: PublicKey,
-    chain_tip: &BitcoinBlock,
-    bitmap: BitArray<[u8; 16]>,
-) -> (AcceptWithdrawalV1, ReqContext) {
-    // Creating `AcceptWithdrawalV1` transactions are tricky. They are a
-    //  mix of data from the bitcoin transaction sweeping out the funds,
-    //  the withdrawal request itself, and how the signers voted.
+fn make_withdrawal_accept(data: &TestSweepSetup) -> (AcceptWithdrawalV1, ReqContext) {
+    // Okay now we get ready to create the transaction using the
+    // `CompleteDepositV1` type.
```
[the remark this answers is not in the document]
djordon: https://github.com/stacks-network/sbtc/pull/582/commits/985b75c78726bf9303d74a33877a75d80ea154fc

signer/tests/integration/withdrawal_accept.rs
```diff
@@ -1,84 +1,79 @@
...
 
-    // This is what the current signer things of the state of things.
+    // This is what the current signer thinks is the state of things.
     let req_ctx = ReqContext {
-        chain_tip: chain_tip.into(),
+        chain_tip: BitcoinBlockRef {
+            block_hash: data.sweep_block_hash.into(),
+            block_height: data.sweep_block_height,
+        },
         // This value means that the signer will go back 10 blocks when
```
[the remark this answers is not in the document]
djordon: https://github.com/stacks-network/sbtc/pull/582/commits/985b75c78726bf9303d74a33877a75d80ea154fc

signer/tests/integration/complete_deposit.rs
```diff
@@ -1,75 +1,72 @@
...
         origin: fake::Faker.fake_with_rng(&mut OsRng),
-        // This value doesn't matter here.
-        aggregate_key: fake::Faker.fake_with_rng(&mut OsRng),
+        // When checking whether the transaction is from the signer, we
+        // check that the first "prevout" has a `scriptPubKey` that the
+        // signers control.
+        aggregate_key: data.aggregated_signer.keypair.public_key().into(),
         // This value affects how many deposit transactions are consider
         // accepted.
         signatures_required: 2,
```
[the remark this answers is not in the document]
djordon: Updated in https://github.com/stacks-network/sbtc/pull/582/commits/985b75c78726bf9303d74a33877a75d80ea154fc

signer/src/stacks/contracts.rs
djordon: I fixed it in the `CompleteDepositV1::validate` docs in https://github.com/stacks-network/sbtc/pull/582/commits/985b75c78726bf9303d74a33877a75d80ea154fc

signer/src/stacks/contracts.rs
djordon: Created https://github.com/stacks-network/sbtc/issues/603 for this.

signer/tests/integration/postgres.rs
cylewitruk: Interesting, I had no idea! Thx

PR #584, opened 2024-09-30
feat: implement `get_signer_utxo` for pg

matteojug:
## Description

Closes https://github.com/stacks-network/sbtc/issues/538.

## Testing Information

Add integration test calls to the tests created in https://github.com/stacks-network/sbtc/pull/567

## Checklist:


signer/src/storage/postgres.rs
```diff
@@ -946,10 +951,80 @@ impl super::DbRead for PgStore {
...
+                FROM sbtc_signer.bitcoin_blocks AS parent
+                JOIN tx_block_chain AS child
+                  ON child.parent_hash = parent.block_hash
+            )
+            SELECT txs.txid, txs.tx, txs.tx_type, tbc.block_hash
+            FROM tx_block_chain AS tbc
+            JOIN sbtc_signer.bitcoin_transactions AS bt ON tbc.block_hash = bt.block_hash
+            JOIN sbtc_signer.transactions AS txs USING (txid)
+            WHERE txs.tx_type = 'sbtc_transaction';
+            "#,
```
matteojug, line 982: Should we add a limit on the recursive CTE here? I'm not sure if pg will do something smart (eg, stopping at the first fetch), or first compute the whole temp table and only then looking at the query.

signer/tests/integration/postgres.rs
```diff
@@ -1265,3 +1265,56 @@ async fn we_can_fetch_bitcoin_txs_from_db() {
...
+        num_signers_per_request: 7,
+    };
+
+    let db_num = testing::storage::DATABASE_NUM.fetch_add(1, Ordering::SeqCst);
+    let store = testing::storage::new_test_database(db_num, true).await;
+    // TODO: something better than this hacky thing?
+    let mut store_vec = vec![store];
+
+    testing::transaction_coordinator::TestEnvironment {
+        storage_constructor: move || store_vec.pop().unwrap(),
```
matteojug, line 1287: Not proud of this, do you have anything smarter to apply here?

djordon: Not that I'm aware. I tried and couldn't come up with anything.

signer/src/storage/postgres.rs
djordon: Oh yeah, we've typically add a context window to explicitly limit the size of these recursive CTEs in case postgres doesn't do the reasonable thing.

signer/tests/integration/postgres.rs
cylewitruk: This goes away in #591 so I wouldn't worry about it

signer/src/storage/postgres.rs
```diff
@@ -946,10 +951,92 @@ impl super::DbRead for PgStore {
...
+        )
+        .bind(chain_tip)
+        .bind(context_window as i32)
+        .fetch(&self.0);
+
+        let mut utxo_block = None;
+        while let Some(tx) = txs.next().await {
+            let tx = tx.map_err(Error::SqlxQuery)?;
+            let bt_tx = bitcoin::Transaction::consensus_decode(&mut tx.tx.as_slice())
+                .map_err(Error::DecodeBitcoinBlock)?;
```
djordon, line 999: We should probably create a separate error variant for this, since an error here will give us a confusing message.

djordon (approved):
Looks good.

I'm adding a new table for https://github.com/stacks-network/sbtc/issues/585. I just realized that this new table can be done in such a way as to make the `get_signer_utxo` function here much simpler.

Issue #586, opened 2024-09-30
[Feature]: Allow for the blocklist client to be unset


djordon:
# Feature - Allow for the blocklist client to be unset

## 1. Description

We don't need a blocklist client, and someone can choose to run their signer without one. In this case, the signer should still run correctly and we should skip the code path that does blocklist checks.

## 2. Technical Details:

Make the blocklist config optional and have the `SignerContext` object optionally return a blocklist client. When there isn't a config for the blocklist client, have an `warn` log and have all `can_accept` calls to the blocklist client return `Ok(true)` (or whatever would be equivalent).

### 2.1 Acceptance Criteria:
   - [ ] Allow for the blocklist client to be unset in the config.

PR #589, opened 2024-10-01
feat: allow unset blocklist clients

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/586.

## Changes

* ~~Move the config file logic out of the `config/mod.rs` file and into a `config/settings.rs` file (I was really hoping that the diff would be nice here).~~ I renamed it back.
* Allow the blocklist client to be unspecified.
* Made the blocklist client endpoint a `url::Url`.
* Add the block observer to the main join.

## Testing Information

Added a unit test for when the blocklist client is set via an environment variable, and manually tested what happens if the stanza in the config is set, but it is currently unset.

## Checklist:


cylewitruk (approved):
Looks good

signer/src/main.rs
```diff
@@ -78,6 +78,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         // Our global termination signal watcher. This does not run using `run_checked`
         // as it sends its own shutdown signal.
         run_shutdown_signal_watcher(context.clone()),
+        run_checked(run_block_observer, &context),
         // The rest of our services which run concurrently, and must all be
         // running for the signer to be operational.
```
matteojug, line 83: nit: is this `run_checked` expected to be above the comment below?

signer/src/transaction_signer.rs
```diff
@@ -657,6 +650,14 @@ where
         Ok(())
     }
 
+    async fn check_blocklist(&self, address: &str) -> bool {
```
matteojug, line 653: nit: the name can be confusing wrt the return value (eg, does true mean the blocklist wanna block this?); what about something like `(check_)?blocklist_accept` or similar?

signer/src/main.rs
djordon: nope, will update.

signer/src/transaction_signer.rs
djordon: Good point, will update.

djordon: Done in https://github.com/stacks-network/sbtc/pull/589/commits/59c41005ab2e6dca751f18aa1ea960129e28acc9.

PR #591, opened 2024-10-01
feat: run transaction coordinator

cylewitruk:
## Description

Runs the transaction coordinator process.

Closes: #593

## Changes
- Made all of the fn's in the `BitcoinInteract` trait async.
- Additional tests for Signer contexts (mostly just proving that things work as expected while showing handling timing issues in tests)
- Refactored away the `Storage` and `BitcoinClient` generic types of the transaction coordinator (in favor of just using the global `Context`).
- Added a new `TestContext` implementation that we can build upon for testing.
- Added the tx coord to the main run-loop join (and block observer but that's also added in another pending PR)
- The testing method `assert_should_be_able_to_coordinate_signing_rounds()` is refactored a fair bit based on the above (and to fix timing issues).

## Testing Information

I think the largest of note is the new `TestContext` implementation and the changes to `assert_should_be_able_to_coordinate_signing_rounds()`.

## Checklist:


signer/src/testing/transaction_coordinator.rs
```diff
@@ -19,87 +26,71 @@ use crate::transaction_coordinator;
...
+    pub async fn start(self) -> RunningEventLoopHandle<C> {
+        let is_started = self.is_started.clone();
+        let join_handle = tokio::spawn(async move {
+            is_started.store(true, Ordering::SeqCst);
+            self.event_loop.run().await
+        });
+
+        while !self.is_started.load(Ordering::SeqCst) {
+            tokio::time::sleep(Duration::from_millis(10)).await;
+        }
```
djordon, line 73: Ahhh, this is to eliminate the race condition?

djordon (approved):
✅ Looks good

signer/src/testing/transaction_coordinator.rs
cylewitruk: Yeah -- could probably have used a Notify channel like I did elsewhere but this is what I ended up with here 🤷🏻‍♂️

signer/src/testing/transaction_coordinator.rs
```diff
@@ -19,87 +26,71 @@ use crate::transaction_coordinator;
...
-    event_loop: EventLoop<S, C>,
-    block_observer_notification_tx: tokio::sync::watch::Sender<()>,
-    storage: S,
+use super::context::TestContext;
+use super::context::WrappedMock;
+
+struct EventLoopHarness<C> {
+    event_loop: EventLoop<C>,
+    context: C,
+    is_started: Arc<AtomicBool>,
```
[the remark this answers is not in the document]
cylewitruk: Yeah that was my original thought too, thought the atomic types were basically "arc+mutex'd" internally, but alas they are only "mutex'd"

Issue #597, opened 2024-10-01
[Feature]: Move Stacks client into signer context


djordon:
# Feature - Move Stacks client into signer context

## 1. Description

The transaction coordinator event loop object needs to be able to broadcast both stacks and bitcoin transactions. Right now it only has the ability to broadcast bitcoin transactions, since the stacks client is missing from the `Context` object.

### 1.1 Context & Purpose

Move the stacks client object into the context so that any of the coordinators that need it can interact with a stacks node.

### 2.1 Acceptance Criteria:
   - [ ] Move the Stacks client into the `SignerContext` object

## 3. Related Issues and Pull Requests (optional):

This is necessary for https://github.com/stacks-network/sbtc/issues/563.

PR #598, opened 2024-10-01
feat: move stacks client into signer context

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/597.

## Changes

* Move the stacks client into the signer context
* Make sure the async trait functions implement `Send`

## Testing Information

Moreover nothing uses the new context for the stacks client so all tests should pass.

## Checklist:


cylewitruk (approved):
Looks good -- note I've got some changes incoming for the `TestContext` (including a builder), so that will be much easier to use (and flexible) moving forward.

Issue #523, opened 2024-09-13
[Feature]: Postgres stacks transactions are not validated correctly


technovision99:
# (High) - Postgres stacks transactions are not validated correctly 

## 1. Description
When fetching withdrawals from the stacks chain and writing them to the db, the `write_stacks_blocks` function is called: https://github.com/stacks-network/sbtc/blob/ffbe79e0238825f3546e9429bbf89042703823df/signer/src/block_observer.rs#L241-L254
This in turn calls `extract_relevant_transactions`:
https://github.com/stacks-network/sbtc/blob/033e2cbd2def829a3e30df77bc274b325d68cdbe/signer/src/storage/postgres.rs#L56-L75
However, notice that the contracts are validated simply by the name and the function call, not by the actual address: 
https://github.com/stacks-network/sbtc/blob/ffbe79e0238825f3546e9429bbf89042703823df/signer/src/storage/postgres.rs#L21-L43
As a result anyone can deploy a clarity contract that has the same name and signature as the associated sBTC actions, and these will be incorrectly written to the database. The contracts should instead filter for events that are from the canonical contracts on the Stacks chain.

PR #599, opened 2024-10-01
feat: db stacks transactions validated incorrectly

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/523.

## Changes

* This change makes sure that we only extract stacks transactions that were deployed by the appropriate entity.

## Testing Information

There was a test for extracting relevant stacks transactions and it was updated to check that contract calls to smart contracts with the incorrect stacks addresses are ignored.

## Checklist:

### DIFF
--- a/signer/tests/integration/postgres.rs
+++ b/signer/tests/integration/postgres.rs
@@ -97,14 +97,16 @@ async fn should_be_able_to_query_bitcoin_blocks() {
     signer::testing::storage::drop_db(store).await;
 }
 
-struct InitiateWithdrawalRequest;
+struct InitiateWithdrawalRequest {
+    deployer: StacksAddress,
+}
 
 impl AsContractCall for InitiateWithdrawalRequest {
     const CONTRACT_NAME: &'static str = "sbtc-withdrawal";
     const FUNCTION_NAME: &'static str = "initiate-withdrawal-request";
     /// The stacks address that deployed the contract.
     fn deployer_address(&self) -> StacksAddress {
-        StacksAddress::burn_address(false)
+        self.deployer
     }
     /// The arguments to the clarity function.
     fn as_contract_args(&self) -> Vec<ClarityValue> {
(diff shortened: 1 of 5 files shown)


signer/src/stacks/wallet.rs
```diff
@@ -495,4 +498,42 @@ mod tests {
...
+        let wallet = SignerWallet::new(&pks1, 5, network_kind, 0).unwrap();
+
+        let rotate_keys = RotateKeysTransaction {
+            txid: fake::Faker.fake_with_rng(&mut rng),
+            aggregate_key: wallet.aggregate_key(),
+            signer_set: public_keys,
+            signatures_required: wallet.signatures_required,
+        };
+
+        // Although it's unlikely, it's possible for the shuffle to not
```
cylewitruk, line 529: Was this something that you had intended to include or is it just here for mental documentation?

signer/tests/integration/postgres.rs
```diff
@@ -97,14 +97,16 @@ async fn should_be_able_to_query_bitcoin_blocks() {
     signer::testing::storage::drop_db(store).await;
 }
 
-struct InitiateWithdrawalRequest;
+struct InitiateWithdrawalRequest {
+    deployer: StacksAddress,
+}
 
 impl AsContractCall for InitiateWithdrawalRequest {
     const CONTRACT_NAME: &'static str = "sbtc-withdrawal";
```
cylewitruk, line 105: Unrelated to this PR specifically, but we should probably create a `constants.rs` or something in the root where we keep the contract names etc.

cylewitruk (approved):
Looks good :)

signer/src/stacks/wallet.rs
djordon: ha, this entire file diff, from https://github.com/stacks-network/sbtc/pull/599/commits/f63ba4fa1b851757c9b37ae01f065325488fe49d) isn't supposed to be here. I was supposed to be in another branch. I'll remove it.

Issue #602, opened 2024-10-02
[Feature]: Update deposit status (via Emily) after tx broadcast


matteojug:
# Feature - Update deposit status (via Emily) after tx broadcast

## 1. Description

Inform the Emily API of the acceptance of deposit request, after enough signatures are gathered and the tx is broadcasted.

Issue #605, opened 2024-10-02
[Feature]: Fetch list of deposits from Emily


matteojug:
# Feature - Fetch list of deposits from Emily

## 1. Description

Reach out to Emily and get the list of deposits if knows about.

## 2. Technical Details:

Sketch of the code is already available (commented out).

### 2.1 Acceptance Criteria:
   - [ ] Criterion 1
   - [ ] Criterion 2
   - [ ] Criterion 3

## 3. Related Issues and Pull Requests (optional):

 - https://github.com/stacks-network/sbtc/issues/595: depends on it to have the required informations available from Emily

matteojug:
Done in https://github.com/stacks-network/sbtc/pull/609

PR #609, opened 2024-10-03
feat: implement emily client to get deposits and update broadcasted ones

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/602, https://github.com/stacks-network/sbtc/issues/605

## Changes

Implement the EmilyClient to get and update deposits status. And add an integration test for the Emily flow.

## Testing Information

Added an integration test, to be run manually, to test the e2e flow wrt Emily: we generate the deposit request there, it's picked up by the block observer, then we accept it and trigger the sweep tx; finally we check that request status on Emily is updated to accepted. The test passes.

## Checklist:


signer/src/emily_client.rs
```diff
@@ -82,6 +101,44 @@ impl EmilyInteract for EmilyClient {
...
+            update_request.push(DepositUpdate {
+                bitcoin_tx_output_index: i32::try_from(deposit.outpoint.vout)
+                    .map_err(|_| Error::TypeConversion)?,
+                bitcoin_txid: deposit.outpoint.txid.encode_hex(),
+                fulfillment: None,
+                last_update_block_hash: bitcoin_chain_tip.block_hash.encode_hex(),
+                last_update_height: i64::try_from(bitcoin_chain_tip.block_height)
+                    .map_err(|_| Error::TypeConversion)?,
+                status: Status::Accepted,
+                status_message: "TODO?".to_string(),
```
matteojug, line 126: Should this `status_message` be something specific for the accepted case?

signer/src/emily_client.rs
```diff
@@ -82,6 +101,44 @@ impl EmilyInteract for EmilyClient {
...
+        let deposits = transaction.requests.iter().filter_map(|e| match e {
+            RequestRef::Deposit(d) => Some(d),
+            _ => None,
+        });
+
+        let mut update_request = Vec::new();
+        for deposit in deposits {
+            update_request.push(DepositUpdate {
+                bitcoin_tx_output_index: i32::try_from(deposit.outpoint.vout)
+                    .map_err(|_| Error::TypeConversion)?,
```
djordon, line 119: Hmm, this doesn't seem right (no ones fault, so definitely not yours). This seems unnecessary and can be fixed by modifying how the source gets generated. 
1. In the `.generated-sources/emily/openapi/package.json` filed, add `--additional-properties=preferUnsignedInt=true` to the end of the `build-rust` "script" command on line 4.
2. Then do `make build`.

These signed integer fields should now be unsigned integers.

djordon (commented):
This looks good so far.

Just had a comment about how we can make the generated Emily sources a little more pleasant to work with.

signer/src/emily_client.rs
djordon: I don't think so.

It's not clear if we should ever put anything here. If we did, I'm not sure why we'd use something other than an enum value here, (I mean, we'd transform it into a string first but still). I'd just put an empty string here always.

signer/src/emily_client.rs
```diff
@@ -57,30 +90,82 @@ impl TryFrom<&Url> for EmilyClient {
...
+                status: Status::Accepted,
+                fulfillment: Some(Some(Box::new(Fulfillment {
+                    bitcoin_txid: transaction.tx.compute_txid().to_string(),
+                    btc_fee: transaction.tx_fee,
+                    // For accepted requests we don't have a block, nor a tx index
+                    bitcoin_block_hash: "".to_string(),
+                    bitcoin_block_height: 0,
+                    bitcoin_tx_index: 0,
+                    stacks_txid: "".to_string(),
+                }))),
```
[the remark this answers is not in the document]
djordon: I don't think so. We have a `Confirmed` state, which should require a non-null fulfillment, but the `Accepted` state should not. @AshtonStephens, is my understanding right here?

signer/tests/integration/emily.rs
```diff
@@ -0,0 +1,575 @@
...
+                output_index: deposit_request.outpoint.vout,
+                signer_pub_key: signer_pub_key.clone(),
+                is_accepted: true,
+            })
+            .await
+            .expect("failed to write deposit decision");
+    }
+
+    // As there was a moment where there was no signer set, the coordinator died when processing the block
+    assert!(tx_coordinator_handle.is_finished());
```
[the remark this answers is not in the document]
djordon: yeah, we'll get to it.

signer/src/emily_client.rs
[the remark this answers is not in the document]
AshtonStephens: You're right, this is design flaw. It should just be confirmed.

signer/src/transaction_coordinator.rs
```diff
@@ -210,14 +218,20 @@ where
...
             self.sign_and_broadcast(
                 bitcoin_chain_tip,
                 aggregate_key,
                 signer_public_keys,
-                transaction,
+                &mut transaction,
             )
             .await?;
+
+            // TODO: need to add a retry (somewhere else?) if this fails?
```
cylewitruk, line 230: This should retry on its own if it's using an `ApiFallbackClient` (which it does in `main.rs`)

signer/src/testing/context.rs
```diff
@@ -327,7 +330,9 @@ impl StacksInteract for WrappedMock<MockStacksInteract> {
     }
 
     fn nakamoto_start_height(&self) -> u64 {
-        tokio::runtime::Handle::current().block_on(async move {
+        let handle = tokio::runtime::Handle::current();
```
cylewitruk, line 333: You should be able to just `tokio::runtime::Handle::current().block_on(..)` I think?

signer/src/transaction_coordinator.rs
matteojug: I was thinking about the case of Emily being offline or unreachable: since we already submitted the tx we would never mark it as accepted. That's why I was thinking of adding the retry in some other place, looking for inconsistencies.

signer/src/testing/context.rs
matteojug: Previously it was like that, but when running the test it would fail with `Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.`. I'm not sure if it's just because of the way the test run, though.

signer/src/emily_client.rs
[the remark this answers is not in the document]
matteojug: https://github.com/stacks-network/sbtc/pull/626

signer/src/transaction_coordinator.rs
cylewitruk: Ahh okay :+1:

signer/src/testing/context.rs
cylewitruk: Hm, weird :thinking: Yeah, must have something to do with the test runtime I would guess, not sure what `tokio::test` does under the covers..

signer/tests/integration/emily.rs
```diff
@@ -0,0 +1,575 @@
...
+        .unwrap()
+        .confirms
+        .push(stacks_tip);
+
+    assert!(context
+        .get_storage()
+        .get_last_key_rotation(&deposit_block_hash.into())
+        .await
+        .unwrap()
+        .is_some());
```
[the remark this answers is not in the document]
matteojug: https://github.com/stacks-network/sbtc/issues/559

signer/tests/integration/emily.rs
```diff
@@ -0,0 +1,575 @@
...
+    deposit_tx.output.push(TxOut {
+        value: bitcoin::Amount::from_sat(1),
+        script_pubkey: aggregate_key.signers_script_pubkey(),
+    });
+    assert!(deposit_tx
+        .output
+        .iter()
+        .any(|txout| signers_pubkey.contains(&txout.script_pubkey)));
+    deposit_request.outpoint.txid = deposit_tx.compute_txid();
+    //
```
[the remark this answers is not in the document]
matteojug: https://github.com/stacks-network/sbtc/issues/625

djordon (commented):
This basically looks good, I had a bunch of "add comment" suggestions. And I think we'll need to change the function signature for `update_broadcasted_deposits` to include enough stacks related information. We can do that either here or in another PR.

signer/src/transaction_coordinator.rs
```diff
@@ -214,14 +222,22 @@ where
...
-                transaction,
+                &mut transaction,
             )
             .await?;
+
+            // TODO: need to add a retry (somewhere else?) if this fails? Otherwise
+            // since we already submitted the bitcoin tx, we would never mark the request
+            // as accepted.
+            self.context
+                .get_emily_client()
```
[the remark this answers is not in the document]
matteojug: > I was thinking about the case of Emily being offline or unreachable: since we already submitted the tx we would never mark it as accepted. That's why I was thinking of adding the retry in some other place, looking for inconsistencies.

I'll remove the comment though, since I guess it's the general _make it robust_ thing.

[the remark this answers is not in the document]
cylewitruk: Realistically what we should probably do here (long-term)  is store all emily "updates" in the db as a persistent queue and push them from there in a separate task. Since it's only the signer coord that sends these, and we don't know how long emily could be unreachable (could be a local network issue), that would help us avoid eating memory storing these in ram and offer better reliability in-case the signer restarts.

signer/src/emily_client.rs
```diff
@@ -57,30 +89,75 @@ impl TryFrom<&Url> for EmilyClient {
...
-        //     )
-        //     .await
-        //     .map_err(EmilyClientError::GetDeposits)?;
+        let resp = deposit_api::get_deposits(&self.config, Status::Pending, None, None)
+            .await
+            .map_err(EmilyClientError::GetDeposits)
+            .map_err(Error::EmilyApi)?;
+
+        // TODO: fetch multiple pages?
+        resp.deposits
```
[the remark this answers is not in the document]
matteojug: The `page_size` value seems to be passed to the emily api endpoint, using `None` should mean _no limit_ since it's used [here](https://github.com/stacks-network/sbtc/blob/main/emily/handler/src/database/entries/mod.rs#L251), but at the same time there's seems to be a "[forced pagination](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html)" once the resulting data reach 1MB.

AshtonStephens:
@matteojug Is this PR good to have re-reviewed?

signer/src/emily_client.rs
```diff
@@ -1,12 +1,24 @@
 //! Emily API client module
```
AshtonStephens, line 1: Can you add a `TODO` to use the autogenerated `Emily` openapi client?

AshtonStephens (approved):
LGTM

signer/src/emily_client.rs
matteojug: Uhm we are using that (eg, the `deposit_api::get_deposits` below). Are you referring to something specific?

signer/src/emily_client.rs
```diff
@@ -57,30 +91,73 @@ impl TryFrom<&Url> for EmilyClient {
...
+                Ok(CreateDepositRequest {
+                    outpoint: OutPoint {
+                        txid: Txid::from_str(&deposit.bitcoin_txid)
+                            .map_err(|_| Error::TypeConversion)?,
+                        vout: deposit.bitcoin_tx_output_index,
+                    },
+                    reclaim_script: ScriptBuf::from_hex(&deposit.reclaim_script)
+                        .map_err(|_| Error::TypeConversion)?,
+                    deposit_script: ScriptBuf::from_hex(&deposit.deposit_script)
+                        .map_err(|_| Error::TypeConversion)?,
```
djordon, line 123: We probably want more specific error variants for these than the ones here.

djordon (approved):
Looks good ✅

signer/src/storage/postgres.rs
```diff
@@ -1433,7 +1433,7 @@ impl super::DbWrite for PgStore {
             , output_index    AS (SELECT ROW_NUMBER() OVER (), output_index FROM UNNEST($2::INTEGER[]) AS output_index)
             , spend_script    AS (SELECT ROW_NUMBER() OVER (), spend_script FROM UNNEST($3::BYTEA[]) AS spend_script)
             , reclaim_script  AS (SELECT ROW_NUMBER() OVER (), reclaim_script FROM UNNEST($4::BYTEA[]) AS reclaim_script)
-            , recipient       AS (SELECT ROW_NUMBER() OVER (), recipient FROM UNNEST($5::BYTEA[]) AS recipient)
+            , recipient       AS (SELECT ROW_NUMBER() OVER (), recipient FROM UNNEST($5::TEXT[]) AS recipient)
```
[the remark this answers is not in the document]
matteojug: This did surface running the emily integration test, so I guess we have one 😅

signer/src/emily_client.rs
matteojug: Done in [7200f63](https://github.com/stacks-network/sbtc/pull/609/commits/7200f633c564223600c5867dfb13a721d278bc77)

PR #610, opened 2024-10-03
feat: run transaction signer

### LINKED ISSUE #594: [Feature]: Run transaction signer
# Feature - Run transaction signer

## 1. Description

The Signer binary needs to run the transaction signer module as part of its main runloop.

### 1.1 Context & Purpose
The transaction signer component runs in its own async task and needs to be run as part of the signer's main runloop.

cylewitruk:
## Description

Closes #594 , i.e. running the transaction signer task in the main runloop.

## Changes

- Integrates the transaction signer with `Context`
- Adds two new `TxSignerEvent`s which are helpful in testing (but can also be useful for logging etc.)
- Increases the internal app signalling channel size from 128 to 1024. Some of the larger tests which execute quickly hit the limit and failed for obscure reasons.
- Some minor refactors/logging stuff for helping debugging tests.
- A lot of small tweaking to get things to work with the context

## Testing Information

All existing tests pass. There are some TODO's with some race-conditions to look into which are temporarily solved with small sleep's etc.

## Checklist:


signer/src/transaction_signer.rs
```diff
@@ -127,71 +131,91 @@ pub struct TxSignerEventLoop<Network, Storage, BlocklistChecker, Rng> {
...
+                }
+
+                // Next, we define a future that polls the network for new messages
+                // which times out after 5ms to ensure we don't block the above
+                // loop. We don't have any methods (atm) on the Network that would
+                // let us `try_recv` or peek. We can get rid of this later on if
+                // we split this run-loop into two separate loops.
+                let future = tokio::time::timeout(Duration::from_millis(5), async {
+                    self.network.receive().await
+                });
```
djordon, line 182: This can be
```rust
let future = tokio::time::timeout(Duration::from_millis(5), self.network.receive());
```

signer/src/transaction_signer.rs
```diff
@@ -127,71 +131,91 @@ pub struct TxSignerEventLoop<Network, Storage, BlocklistChecker, Rng> {
...
+                // First we empty the signal channel subscription, checking for
+                // new Bitcoin block observed events. It doesn't matter how many
+                // of these we get, we only care if it has happened. It's also
+                // important that we empty this channel as quickly as possible
+                // to avoid un-processed messagages being dropped.
+                let mut new_block_observed = false;
+                while let Ok(signal) = signal_rx.try_recv() {
+                    if let SignerSignal::Event(SignerEvent::BitcoinBlockObserved) = signal {
+                        new_block_observed = true;
                     }
```
djordon, line 167: Ahh right, it doesn't care about any other internal messages besides the termination signal message. And everything else comes over the network.

djordon (commented):
Just had one question about the blocklist client.

signer/src/transaction_signer.rs
cylewitruk: Well this one only cares about `BitcoinBlockObserved`, termination channel is a separate `watch` channel, and then yeah just below is the stuff coming over the network. So these two operations run in the same loop, and the `select!` below handles killing it when it dies or termination is signalled.

I did try putting these into two separate futures with their own loops but `&mut` quickly became a problem and I went this route instead to not spend time on that.

djordon: No it makes sense. After reading your comments I couldn't think of a way to do it as separate functions.

signer/src/main.rs
```diff
@@ -261,14 +264,30 @@ async fn run_block_observer(ctx: impl Context) -> Result<(), Error> {
...
+    let config = ctx.config().clone();
+    let network = P2PNetwork::new(&ctx);
+
+    let signer = transaction_signer::TxSignerEventLoop {
+        network,
+        context: ctx,
+        context_window: 10000,
+        threshold: 2,
+        // TODO: Update when we have a blocklist impl
+        blocklist_checker: Some(()),
```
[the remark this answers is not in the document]
cylewitruk: Good point, changed in [1372167](https://github.com/stacks-network/sbtc/pull/610/commits/137216753461fc87ed68541127194b93ba1cb78b)

signer/src/main.rs
```diff
@@ -261,14 +265,30 @@ async fn run_block_observer(ctx: impl Context) -> Result<(), Error> {
...
+    let config = ctx.config().clone();
+    let network = P2PNetwork::new(&ctx);
+
+    let signer = transaction_signer::TxSignerEventLoop {
+        network,
+        context: ctx.clone(),
+        context_window: 10000,
+        threshold: 2,
+        // TODO: Update when we have a blocklist impl
+        blocklist_checker: BlocklistClient::new(&ctx),
```
djordon, line 280: This comment can be removed now.

djordon (approved):
Awesome ✅

signer/src/main.rs
cylewitruk: Sloppy :grimacing: 
Removed together with the merge commit [3958c8e](https://github.com/stacks-network/sbtc/pull/610/commits/3958c8ea34341e94a238136f0fc1dcd325325f92)

Issue #616, opened 2024-10-04
[Feature]: Sign and submit `complete-deposit` transactions


djordon:
# Feature - Sign and submit `complete-deposit` transactions

## 1. Description

This is part of https://github.com/stacks-network/sbtc/issues/320, but focused on `complete-deposit` transactions.

After we have a stacks transactions that needs signing, we need to get it signed and broadcasted.

### 1.1 Context & Purpose

We need to make sure that the coordinator asks the other signers to sign the transaction, add them to the transaction and the broadcast them.

## 2. Technical Details:

We need to hook up the `TxSignerEventLoop` to listen for the right messages and sign them if they are valid. For this ticket, we do not worry too much about making sure that they are valid, just that we sign them.

### 2.1 Acceptance Criteria:
   - [ ] We can finish a full loop where we sign and broadcast complete-deposit stacks transactions that have response transactions on bitcoin.

## 3. Related Issues and Pull Requests (optional):

This is part of https://github.com/stacks-network/sbtc/issues/320, and builds off of https://github.com/stacks-network/sbtc/issues/563.

PR #617, opened 2024-10-04
feat: sign and broadcast Stacks transactions

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/616.

Note that we purposefully skip validation for now, but it should be relatively straightforward to hook up.

## Changes

* Make sure the client types are `&'static`, which they all are. This was needed for borrow checker.
* Send the transaction ID with the `StacksTransactionSignRequest`. We can probably remove the digest from that type, we will do that later.
* Make sure we have a loop for signing stacks transactions.
* Submit the signed transaction to the stacks node

This PR needs some polish and some tests.

## Testing Information

The main test that is needed is that the transaction coordinator broadcasts a signed `complete-deposit` transaction for a deposit request with a sweep transaction on bitcoin. 

## Checklist:


cylewitruk:
> Note that we purposefully skip validation for now, but it should be relatively straightforward to hook up.

Is there a separate ticket for this?

djordon:
> > Note that we purposefully skip validation for now, but it should be relatively straightforward to hook up.
> 
> Is there a separate ticket for this?

No not yet. There are a few TODOs in this PR that I need to turn into outstanding tickets, this was just one of them.

signer/src/message.rs
```diff
@@ -165,6 +165,8 @@ pub struct StacksTransactionSignRequest {
     /// It's essentially a hash of the contract call struct, the nonce, the
     /// tx_fee and a few other things.
     pub digest: [u8; 32],
+    /// The transaction ID of the associated contract call transaction.
+    pub txid: blockstack_lib::burnchains::Txid,
```
matteojug, line 169: Is this docstring/type correct? What associated contract call tx (on bitcoin?) are we referring to?

signer/src/transaction_coordinator.rs
```diff
@@ -288,20 +291,61 @@ where
...
+            // If we fail here, then we need to decrement the nonce.
+            let process_request_fut =
+                self.process_sign_request(sign_request, chain_tip, multi_tx, &wallet);
+
+            match process_request_fut.await {
+                Ok(txid) => {
+                    tracing::info!(%txid, "successfully submitted complete-deposit transaction")
+                }
+                Err(error) => {
+                    tracing::warn!(%error, "could process the sign request for deposit request");
```
matteojug, line 315:
```suggestion
                    tracing::warn!(%error, "could not process the sign request for deposit request");
```

signer/src/transaction_coordinator.rs
```diff
@@ -288,20 +291,61 @@ where
...
         Ok(())
     }
 
+    /// Sign and broadcast the stacks transaction
+    ///
+    /// If we fail to sign the transaction for some reason, we decrement
+    /// the nonce by one, and try the next transaction. This is not a fatal
+    /// error, since we could fail to sign the transaction because someone
+    /// else is now the coordinator, and all of the signers are now
+    /// ignoring us.
```
matteojug, line 330: nit: move this part above? It seems out of place here, since we are not decrementing the nonce (here)

matteojug (commented):
Just some nits, so far LGTM.

signer/src/message.rs
djordon: Yeah I believe this is correct. Maybe we should use a type alias or the internal `StacksTxId` type here to make that clear, but `blockstack_lib::burnchains::Txid` is the type for transaction IDs on stacks. See:
https://github.com/stacks-network/stacks-core/blob/2e67fa900b5a1ef9f7268529b9c2d1fdcc72c5d4/stackslib/src/chainstate/stacks/transaction.rs#L824-L829
and in that same file:
https://github.com/stacks-network/stacks-core/blob/2e67fa900b5a1ef9f7268529b9c2d1fdcc72c5d4/stackslib/src/chainstate/stacks/transaction.rs#L34-L36
.

signer/src/transaction_coordinator.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/617/commits/7cdea8d577b0e2a7d7ed9335e26d38e87252a0e9

signer/src/transaction_coordinator.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/617/commits/7cdea8d577b0e2a7d7ed9335e26d38e87252a0e9

signer/src/transaction_signer.rs
```diff
@@ -435,12 +443,14 @@ where
...
-        ctx: &impl Context,
-        request: &message::StacksTransactionSignRequest,
+    async fn assert_valid_stacks_tx_sign_request(
+        &self,
+        request: &StacksTransactionSignRequest,
         chain_tip: &model::BitcoinBlockHash,
     ) -> Result<(), Error> {
+        if true {
+            return Ok(());
+        }
```
matteojug, line 453: Do we have a TODO/issue for this?

matteojug (approved):

signer/src/testing/context.rs
```diff
@@ -162,10 +162,10 @@ impl<Storage, Bitcoin, Stacks>
 
 impl<Storage, Bitcoin, Stacks, Emily> Context for TestContext<Storage, Bitcoin, Stacks, Emily>
 where
-    Storage: DbRead + DbWrite + Clone + Sync + Send,
-    Bitcoin: BitcoinInteract + Clone + Send + Sync,
-    Stacks: StacksInteract + Clone + Send + Sync,
-    Emily: EmilyInteract + Clone + Send + Sync,
+    Storage: DbRead + DbWrite + Clone + Sync + Send + 'static,
```
cylewitruk, line 165: What was it that made these all need to be 'static?

cylewitruk (approved):
Looks good in general

signer/src/transaction_signer.rs
djordon: Oh we do, in the line below.

aldur:
@djordon can you lay out the test plan so that @matteojug can take it?

signer/src/testing/context.rs
djordon: I was getting some compilation error without the `'static` bound on the Stacks client. I'm not sure if I got the error with all of them or not (certainly not with Emily). I did not revisit whether they could be removed after the code was finished though, so it's possible that we can remove the `'static` bounds now.

Issue #588, opened 2024-09-30
[Feature]: Eliminate race condition between event loops


djordon:
# Feature - Eliminate race condition between event loops

## 1. Description

The block observer event loop notifies both the `transaction-signer` and the `transaction-coordinator` event loops whenever it observes a new bitcoin block. The `transaction-signer` event loop is the one that votes on the requests (as well as signing), while the `transaction-coordinator` is the one that constructs any bitcoin and stacks transactions. They both currently act in parallel, but the `transaction-coordinator` uses the output of the `transaction-signer` when it acts.

### 1.1 Context & Purpose

The problem is that the `transaction-signer` might not be done with gathering votes on a request before the `transaction-coordinator` starts looking for accepted requests to process. The effect is that the signers will wait an additional bitcoin block before acting on deposit requests.

## 2. Technical Details:

The simplest thing to do is to replace the channel between the `block-observer` and the `transaction-coordinator` with a channel between the `transaction-signer` and the `transaction-coordinator`. Then we just notify the `transaction-coordinator` at the end of `TxSignerEventLoop::handle_new_requests`.

### 2.1 Acceptance Criteria:
   - [ ] If there is a new deposit request, we can process it at right after the next bitcoin block.

cylewitruk:
I'll grab this while I'm working on running the tx signer now

PR #618, opened 2024-10-04
feat: fix race condition described in #588

cylewitruk:
## Description

Closes: #588

## Changes

This adds a new `TxSignerEvent` variant `NewRequestsHandled`, which is emitted at the end of the tx-signer's `handle_new_requests()` successful completion and used as the trigger for the transaction coordinator. So the new flow is:

Block Observer: `BitcoinBlockObserved` -> Transaction Signer: `NewRequestsHandled` -> Transaction Coordinator runs

## Testing Information

Coordinator tests have been updated to use this new trigger.

## Checklist:

### DIFF
--- a/signer/src/context/messaging.rs
+++ b/signer/src/context/messaging.rs
@@ -53,6 +53,9 @@ pub enum TxSignerEvent {
     PendingWithdrawalRequestRegistered,
     /// A new pending deposit request has been handled.
     PendingDepositRequestRegistered,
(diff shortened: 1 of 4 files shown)


djordon (approved):
Great.

Would you mind updating the documentation on lines 45, 50, and 91 in `signer/src/transaction_coordinator.rs` too? I don't think there is anywhere else where the documentation needs updating.

Issue #604, opened 2024-10-02
[Feature]: Update deposit status (via Emily) after sweep tx is confirmed


matteojug:
# Feature - Update deposit status (via Emily) after sweep tx is confirmed

## 1. Description

Inform the Emily API that the deposit included in a sweep tx has been confirmed

### 2.1 Acceptance Criteria:
   - [ ] Criterion 1
   - [ ] Criterion 2
   - [ ] Criterion 3

matteojug:
This issue was wrongly phrased:
 - deposit acceptance was implemented in https://github.com/stacks-network/sbtc/issues/602
 - what (in spirit) this wanted to do is actually superseded by https://github.com/stacks-network/sbtc/issues/340

PR #622, opened 2024-10-07
feat: add emily client update confirmed deposit

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/604

## Changes

Add Emily client method to update confirmed deposits.

## Testing Information

Added an integration tests interacting with the client.

## Checklist:

- [ ] Wire actual client call after stacks tx submission and add to integration test


signer/src/emily_client.rs
```diff
@@ -159,6 +171,41 @@ impl EmilyInteract for EmilyClient {
...
+            bitcoin_tx_output_index: deposit.output_index,
+            bitcoin_txid: deposit.txid.to_string(),
+            status: Status::Confirmed,
+            last_update_block_hash: stacks_chain_tip.block_hash.to_string(),
+            last_update_height: stacks_chain_tip.block_height,
+            fulfillment: Some(Some(Box::new(Fulfillment {
+                bitcoin_txid: deposit.sweep_txid.to_string(),
+                // TODO: we don't keep a mapping between sweep tx inputs and
+                // deposits requests fulfilled, should we?
+                bitcoin_tx_index: 0,
```
djordon, line 192: > // TODO: we don't keep a mapping between sweep tx inputs and
> // deposits requests fulfilled, should we?

Well, we implicitly do since we store the raw transaction. We can compute the index here using the `sweep_tx` field of the `SweptDepositRequest` object, which holds the bitcoin transaction. Does this make sense?

matteojug:
As https://github.com/stacks-network/sbtc/issues/604, this is superseded by https://github.com/stacks-network/sbtc/pull/712.

I've pushed the latest changes just in case I'm missing something that need to be saved from here.

PR #626, opened 2024-10-08
Change mandatory fulfillment for confirmed instead of accepted requests

matteojug:
## Description

Make fulfilment required for confirmed requests instead of accepted ones.

## Testing Information

Emily integrations tests are green.

## Checklist:


AshtonStephens (approved):
LGTM! Nice clean PR :) 

There are *integration* tests that will fail from this, like this part for instance: [link](https://github.com/stacks-network/sbtc/blob/main/emily/handler/tests/integration/endpoints/deposit.rs#L395-L404)

To unblock this PR I'm approving. Those integration tests will be easy to fix, but requires some rigamarole to run right now

AshtonStephens:
Merging

Issue #603, opened 2024-10-02
[Feature]: Check the input scriptPubKey when fetching signers' UTXO


djordon:
# Feature - Check the input scriptPubKey when fetching signers' UTXO

## 1. Description

When searching for the signers outstanding UTXO in https://github.com/stacks-network/sbtc/issues/538, we need to only include transactions where one of the inputs points to an output matching one of the signers' `scriptPubKey`s. Anyone can create an output that the signers' control, but we are only interested in transactions that we know the signers have created. This check is necessary because without it, we could fetch a UTXO that the signers have not created. 

### 1.1 Context & Purpose

We need to update the list of criteria in https://github.com/stacks-network/sbtc/issues/538 to specify that we are only interested in transactions where a transactions prevout was signed for by the signers.

### 2.1 Acceptance Criteria:
   - [ ] Make sure that the UTXO was created by the signers.

## 3. Related Issues and Pull Requests (optional):

https://github.com/stacks-network/sbtc/issues/538

PR #627, opened 2024-10-08
Add donations tx type, consider them for signers utxo

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/603

## Changes

When extracting sbtc txs, separate donations from actual sbtc txs (where the first input of the tx is a signers spendable output). When getting signers utxo, as donations are marked differently we will not consider them (kinda, see next part).

This PR also adds a fallback for getting signers utxo: if no sbtc tx is found, we consider the latest donation (`get_utxo_from_donation`, same as `get_signer_utxo` just considering donations instead of sbtc txs); this can help in bootstrapping things initially (as an alternative to manually add the first signers utxo sbtc tx in db).

## Testing Information

Adds a test for the new behaviour.

## Checklist:


djordon:
> this can help in bootstrapping things initially (as an alternative to manually add the first signers utxo sbtc tx in db)

Oh interesting. I was considering us manually spending some coins under the signer's name, but I think this works better long term.

But I don't think this addresses #603

matteojug:
> But I don't think this addresses #603

With this PR we don't insert sBTC txs (with that specific tx type, that is then used to get signers utxo) unless their input is an output spendable by the signers: isn't this doing what 603 asks, just acting on the "input" side (when storing the sbtc tx) instead of when we consume it (in get signers utxo)?

djordon:
> > But I don't think this addresses #603
> 
> With this PR we don't insert sBTC txs (with that specific tx type, that is then used to get signers utxo) unless their input is an output spendable by the signers: isn't this doing what 603 asks, just acting on the "input" side (when storing the sbtc tx) instead of when we consume it (in get signers utxo)?

Oops, my mistake, I just completely missed https://github.com/stacks-network/sbtc/pull/627/files#diff-cdebb5053105186a7fae52ce8a2d49013fcead0f04e0adf7fb4727942fabf709R278-R321. Yeah this checks the input and does fix #603.

signer/src/testing/transaction_coordinator.rs
```diff
@@ -526,6 +540,147 @@ where
...
+
+        let (aggregate_key, bitcoin_chain_tip, mut test_data) = self
+            .prepare_database_and_run_dkg(&mut rng, &mut signer_set)
+            .await;
+
+        let original_test_data = test_data.clone();
+
+        // The scenario is:
+        // [bitcoin_chain_tip] +- [block a1 with signer utxo] - [block a2 with donation]
+        //                     +- [block b1 with donation]
```
djordon, line 562: In this diagram, does block a2 have greater height than block a1?

djordon (approved):
Looks good. I just had one style suggestion and one question.

signer/src/testing/transaction_coordinator.rs
matteojug: Yes -- changed the `bitcoin_chain_tip` to make it clear it refers to the "original" chain tip before adding the test blocks

Issue #625, opened 2024-10-07
[Feature]: Store deposit request transactions


djordon:
# Feature - Store deposit request transactions

## 1. Description

The signers depend on the deposit transaction being stored in their databases when getting pending deposit requests. The issue is that the block observer does not fetch and store the actual transaction, just the parsed deposit request.

### 1.1 Context & Purpose

In order to process a deposit request, we need to have the associated transaction in the signers' databases. This is required in order to know whether a deposit transaction has been confirmed on the canonical bitcoin blockchain.

### 2.1 Acceptance Criteria:
   - [ ] Store the deposit transaction into the database if the deposit transaction has been confirmed in a bitcoin block.

## 3. Related Issues and Pull Requests (optional):

* https://github.com/stacks-network/sbtc/issues/522

PR #628, opened 2024-10-08
Store deposit requests txs when extracted

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/625

## Changes

In `extract_deposit_requests`, we store the bitcoin tx for deposit requests (both the bitcoin tx and the "raw" `TransactionType::DepositRequest` tx).
Also added the `write_transaction` call in `write_bitcoin_transactions` and `write_stacks_transactions` for in memory storage impl, as done by the pg one.

## Testing Information

Added a check on an existing test, and tested on the integration test in https://github.com/stacks-network/sbtc/pull/609.

## Checklist:

### DIFF
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -227,19 +227,43 @@ where
         self.write_stacks_blocks(&stacks_blocks).await?;
         self.write_bitcoin_block(&block).await?;
 
-        self.extract_deposit_requests(&block.txdata).await?;
+        self.extract_deposit_requests(&block.txdata, block.block_hash())
+            .await?;
 
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


signer/src/block_observer.rs
```diff
@@ -227,19 +227,43 @@ where
...
+                    tx_type: model::TransactionType::DepositRequest,
+                    block_hash: block_hash.to_byte_array(),
+                };
 
+                Ok((model::DepositRequest::from(deposit), tx))
+            })
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(Error::BitcoinEncodeTransaction)?
+            .into_iter()
+            .unzip();
```
djordon, line 261: Nice, I did not know about `unzip`. And it is used well here.

djordon (approved):
Looks good ✅!

Issue #590, opened 2024-10-01
[Feature]: Bootstrapping distributed key generation


djordon:
# Feature - Bootstrapping distributed key generation

## 1. Description

When the signer binary starts up, they need to do a few things before they can do their main task of bridging BTC into sBTC (and vice-verse):
1. Discover other sBTC signers and connect to them.
2. Decide when to run distributed key generation (DKG), and actually run it.
3. Construct and broadcast a `rotate-keys-wrapper` contract call to the Stacks network.

For (1) the discovery can be done by hard-coding a list of signers addresses. For (2), we can have a simple rule for first running DKG: have the coordinator start running DKG when connected to all seed nodes and there has never been DKG run before.

The only tricky part is (3). I see two options:
1. We manually construct and broadcast the `rotate-keys-wrapper` contract call transaction ourselves. This would mean that we also manually deploy the smart contracts as well, which is easy. The "difficult" part is making sure the signers *know about the contract call*.
2. The only other way is to have the signers construct and broadcast all `rotate-keys-wrapper` contract calls, including the first one. The "difficult" part here is that it implies that the signers deploy the smart contracts too. Making sure the signers know about the contract call is easy this route.

My best guess is that (1) is easier than (2), but I'm not sure. We have a lot of the code for (2) already, including for deploying the smart contracts.

### 2.1 Acceptance Criteria:
   - [ ] ~We've aligned on a plan for parts 1-3 of bootstrapping DKG.~ Aligning on the plan shouldn't happen in this ticket. One approach is described in https://github.com/stacks-network/sbtc/issues/665, and we'll align on the approach there.
   - [ ] We have code in the signer to bootstrap DKG and start processing requests.
   - [x] https://github.com/stacks-network/sbtc/issues/651
   - [x] https://github.com/stacks-network/sbtc/issues/652

## 3. Related Issues and Pull Requests (optional):

* There is also https://github.com/stacks-network/sbtc/issues/653, which kind of touches on (3).

PR #632, opened 2024-10-08
feat: bootstrapping distributed key generation (stale)

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/590

## Changes

* Run DKG is we need to
* Expect the public keys of the signers in the config. Not sure if this one will last.

There are still some things that need to be done, 
1. Submitting the rotate-keys transaction. Though I'm not entirely sure that we want to do that.
2. Change the code to fallback on using the `dkg_shares` table for the current aggregate key if there is no `rotate_keys_transaction` in the database.

## Testing Information

This also needs tests.

## Checklist:


signer/src/transaction_coordinator.rs
```diff
@@ -169,6 +177,12 @@ where
             .await?
             .ok_or(Error::NoChainTip)?;
 
+        if self.needs_dkg(&bitcoin_chain_tip).await? == DkgState::NeedsDkg {
+            // This function returns the new DKG aggregate key. That
+            // aggregate key is different from aggregate key of the signers.
```
matteojug, line 182: qq: just to be clear on the notation, the `aggregate key of the signers` is the stacks multisig principal?

signer/src/transaction_coordinator.rs
```diff
@@ -392,15 +440,33 @@ where
...
                 None => continue,
-                Some(_) => return Err(Error::UnexpectedOperationResult),
             }
         }
     }
 
+    /// Check whether or not we need to run DKG
+    /// 
+    /// This function checks for the existence of a
+    /// [`RotateKeysTransaction`] in the database, and one does not exist
```
cylewitruk, line 452:
```suggestion
    /// Check whether or not we need to run DKG
    /// 
    /// This function first checks for the existence of a
    /// [`RotateKeysTransaction`] row in the database. If one exists, then DKG
    /// has already been completed.
    ///
    /// If not, it then proceeds to check for DKG shares in the database for
    /// the current signer aggregate key. If DKG shares exist but there is no
    /// [`RotateKeysTransaction`] then one needs to be created, otherwise if
    /// no shares exist then DKG needs to be run.
```

signer/src/transaction_coordinator.rs
```diff
@@ -169,6 +177,12 @@ where
             .await?
             .ok_or(Error::NoChainTip)?;
 
+        if self.needs_dkg(&bitcoin_chain_tip).await? == DkgState::NeedsDkg {
```
matteojug, line 180: Since this is not guarded by some `self.is_coordinator` condition, do we risk to have concurrent DKGs?

signer/src/transaction_coordinator.rs
```diff
@@ -553,6 +619,15 @@ where
         PublicKey::from_private_key(&self.private_key)
     }
 
+    /// This function provides a deterministic 32-byte identifier for the
+    /// signer. This should probably deterministically too.
```
cylewitruk, line 623: "This should probably deterministically too" ... ? :laughing:

cylewitruk (commented):
Looks good, assuming that the actual logic in `wsts_state_machine` and `wsts` is correct :)

signer/src/transaction_coordinator.rs
```diff
@@ -326,45 +340,79 @@ where
...
     }
 
-    #[tracing::instrument(skip(self))]
-    async fn relay_messages_to_wsts_state_machine_until_signature_created(
+    #[tracing::instrument(skip_all)]
+    async fn coordinate_dkg(
+        &mut self,
+        chain_tip: &model::BitcoinBlockHash,
+    ) -> Result<PublicKey, Error> {
+        let mut state_machine = CoordinatorStateMachine::new([], self.threshold, self.private_key);
```
matteojug, line 377: sanity check: is the `[ ]` just a wip thing?

signer/src/transaction_coordinator.rs
matteojug, line 623: nit: `This should probably deterministically too` I guess a word is missing somewhere, or is a repetition of the `deterministic` above

signer/src/config/default.toml
```diff
@@ -132,6 +132,17 @@ deployer = "SN2V7WTJ7BHR03MPHZ1C9A9ZR6NZGR4WM8HT4V67Y"
 # Environment: SIGNER_SIGNER__DB_ENDPOINT
 db_endpoint = "postgresql://postgres:postgres@localhost:5432/signer"
 
+# The public keys of known signers who are approved to be in the signer
+# set.
+# The signer database endpoint (pgsql connection string)
```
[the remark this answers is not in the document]
djordon: lol, 🙏🏽 thank you

signer/src/transaction_coordinator.rs
djordon: perfect, yeah I forgot to update this.

signer/src/transaction_coordinator.rs
djordon: lol will update.

signer/src/transaction_coordinator.rs
djordon: Well not exactly. The stacks multi-sig principal is formed differently. It does not use the aggregate key of the public keys embedded in the address. I added this line here because I wanted to be clear that this will be different from the aggregate key of the signers, even though we associate it with the signers' public keys.

signer/src/transaction_coordinator.rs
djordon: Oh, I just forgot to add that in.

signer/src/transaction_coordinator.rs
djordon: yeah, forgot to change that.

signer/src/transaction_coordinator.rs
djordon: lol yeah typo.

signer/src/transaction_coordinator.rs
```diff
@@ -326,45 +340,79 @@ where
...
 
-        tokio::time::timeout(max_duration, run_signing_round)
+        let operation_result = tokio::time::timeout(max_duration, run_signing_round)
             .await
-            .map_err(|_| Error::CoordinatorTimeout(self.signing_round_max_duration.as_secs()))?
+            .map_err(|_| Error::CoordinatorTimeout(max_duration.as_secs()))??;
+
+        match operation_result {
+            WstsOperationResult::SignTaproot(signature) => Ok(signature),
+            _ => Err(Error::UnexpectedOperationResult),
```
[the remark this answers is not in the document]
djordon: I think so. But this was what was here before.

djordon:
This has been superseded by https://github.com/stacks-network/sbtc/pull/662 and https://github.com/stacks-network/sbtc/pull/655.

PR #633, opened 2024-10-09
feat: fix signer crashes pre-nakamoto

cylewitruk:
## Description

Prevents two crash situations:
- Stacks node(s) unavailable or not ready (i.e. the RPC API hasn't started listening yet). Common when starting things up together as the Stacks node needs some time, especially if performing a first sync.
- Nakamoto block height hasn't been reached: when a new bitcoin block is observed, the signer queries the Stacks node for the related Nakamoto block, but if we're pre-naka then this returns 404 and crashes the signer.

## Changes

This PR implements two new `StacksInteract` methods `get_pox_info()` and `get_node_info()`. I thought I needed both but it turned out I only needed `get_pox_info()`, but `get_node_info()` was already implemented so I left it (though it's currently unused).

`get_pox_info()` is then used in a new wait-loop in `main.rs` which waits for the Stacks RPC to become available and a valid response from `v2/pox` which returns the nakamoto start height, and then it polls waiting for the nakamoto start height to be reached in the bitcoin chain.

It also changes all logic to use `get_pox_info()` for retrieving the stacks-network-configured nakamoto start height (i.e. epoch 3.0 start height), so it doesn't need to be configured in the file (`nakamoto_start_height()` on `StacksInteract` is now removed).

I haven't done all of the cleanup with removing all instances of the `nakamoto_start_height` variable/config yet, figure I'd do that in a separate PR after demo.

## Testing Information

All existing tests pass. Had to refactor the block observer `TestEnvironment` a little to mock some data for the `StacksInteract` impl.

## Checklist:


signer/src/stacks/api.rs
```diff
@@ -640,13 +718,17 @@ where
...
+        .ok_or(Error::MissingNakamotoStartHeight)?;
 
     while let Some(block) = blocks.last() {
         // We won't get anymore Nakamoto blocks before this point, so
         // time to stop.
-        if block.header.chain_length <= stacks.nakamoto_start_height() {
+        if block.header.chain_length <= nakamoto_start_height {
             tracing::info!(
-                nakamoto_start_height = %stacks.nakamoto_start_height(),
+                nakamoto_start_height = %nakamoto_start_height,
```
matteojug, line 731:
```suggestion
                %nakamoto_start_height,
```

signer/src/stacks/api.rs
```diff
@@ -1296,6 +1386,298 @@ mod tests {
         first_mock.assert();
     }
 
+    #[tokio::test]
+    async fn get_pox_info_and_get_nakamoto_start_height_works() {
+        let raw_json_response = r#"
```
matteojug, line 1391: nit: move to `tests/fixtures` and `include_str!` it?

signer/src/main.rs
```diff
@@ -100,6 +107,76 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                    tokio::time::sleep(Duration::from_secs(3)).await;
+                }
+            }
+        }
+
+        // Now we have the Nakamoto start height, wait for the burnchain to catch up to it.
+        while current_burnchain_height <= nakamoto_height {
+            match client.get_pox_info().await {
+                Ok(info) => {
+                    current_burnchain_height = Some(info.current_burnchain_block_height);
```
matteojug, line 151: sanity check: below we compare `nakamoto_start_height` with (stacks)`block.header.chain_length`, while here we use the burnchain height. Are the two the same in pre nakamoto?

signer/src/stacks/api.rs
```diff
@@ -1296,6 +1386,298 @@ mod tests {
...
+        let mock = stacks_node_server
+            .mock("GET", "/v2/pox")
+            .with_status(200)
+            .with_header("content-type", "application/json")
+            .with_body(raw_json_response)
+            .expect(1)
+            .create();
+
+        // Setup our Stacks client. We use a regular client here because we're
+        // testing the `get_fee_estimate` method.
```
matteojug, line 1578:
```suggestion
        // testing the `get_pox_info` method.
```

signer/src/stacks/api.rs
```diff
@@ -1296,6 +1386,298 @@ mod tests {
...
+        let mock = stacks_node_server
+            .mock("GET", "/v2/info")
+            .with_status(200)
+            .with_header("content-type", "application/json")
+            .with_body(raw_json_response)
+            .expect(1)
+            .create();
+
+        // Setup our Stacks client. We use a regular client here because we're
+        // testing the `get_fee_estimate` method.
```
matteojug, line 1668:
```suggestion
        // testing the `get_node_info` method.
```

cylewitruk: [71f2d88](https://github.com/stacks-network/sbtc/pull/633/commits/71f2d88210ede4a8c6f93827deaff6352735c258)

signer/src/stacks/api.rs
cylewitruk: [71f2d88](https://github.com/stacks-network/sbtc/pull/633/commits/71f2d88210ede4a8c6f93827deaff6352735c258)

signer/src/main.rs
cylewitruk: Stacks epoch heights are based on the burnchain (bitcoin) block height and not the stacks height (otherwise it would be unpredictable). So we pull the nakamoto (epoch 3.0) start height from `/v2/pox`, and follow the `current_burnchain_block_height` (current bitcoin height) until it's > the `nakamoto_height` (we give it 1 extra block since there can be weird stuff right around the transition).

Make sense?

signer/src/stacks/api.rs
cylewitruk: Good idea, [2bceba5](https://github.com/stacks-network/sbtc/pull/633/commits/2bceba5882b719400fc94fd0d9c1f950f812815c)

signer/src/stacks/api.rs
cylewitruk: [2bceba5](https://github.com/stacks-network/sbtc/pull/633/commits/2bceba5882b719400fc94fd0d9c1f950f812815c)

signer/src/main.rs
matteojug: Yes, thanks! But I guess now the question is why we have [this check](https://github.com/stacks-network/sbtc/blob/main/signer/src/stacks/api.rs#L647) using the stacks height.

cylewitruk: Yeah that's a good Q, the assumptions in that check are wrong. I added a TODO for now, we'll need to revisit this. We need to find a way to get the burnchain block height from the information we have here.

cylewitruk: [5de6003](https://github.com/stacks-network/sbtc/pull/633/commits/5de6003b021613621f8f4218745cdcb7eaa9c6d0)

matteojug (approved):

signer/src/main.rs
cylewitruk: Another idea is to keep going until we get a 404 from the stacks node. It's not as "exact", but it's what the stacks node will return when trying to get nakamoto data for pre-naka (might include 2.5, but i don't think it should be an issue if we get 2.5 data stored?)

cylewitruk: Created #637

PR #634, opened 2024-10-09
feat: make block observer + signer + coordinator more robust

cylewitruk:
## Description

Removes error propagation in the block observer, transaction signer & coordinator event loops, logging the error and waiting for the next run (as these errors can be temporary) instead of causing the signer to shutdown.

This is primarily a problem when bootstrapping because DKG hasn't yet been run, and then we hit an error regarding this and the signer immediately shuts down (which in turn makes it impossible to run DKG).

## Changes

As described in description; error propagation for these event loops are largely removed (we may want to match on some specific truly-fatal errors later, but now it will just keep retrying).

## Testing Information

Nothing special, has been run against devenv.

## Checklist:

### DIFF
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -121,18 +121,38 @@ where
         // early.
         let run = async {
             while let Some(new_block_hash) = self.bitcoin_blocks.next().await {
-                self.load_latest_deposit_requests().await?;
+                if let Err(error) = self.load_latest_deposit_requests().await {
+                    tracing::warn!(%error, "could not load latest deposit requests from Emily");
+                }
 
                 // TODO: What to do when `new_block_hash?` errors? Perhaps we can
                 // handle this within a failover-stream if this indicates a problem
                 // with the stream, and then we change this back to a plain `BlockHash`
                 // instead of a `Result<>`.
-                for block in self.next_blocks_to_process(new_block_hash?).await? {
-                    self.process_bitcoin_block(block).await?;
-                }
+                match new_block_hash {
+                    Ok(new_block_hash) => {
+                        tracing::info!(%new_block_hash, "observed new bitcoin block from stream");
 
-                self.context
-                    .signal(SignerEvent::BitcoinBlockObserved.into())?;
+                        match self.next_blocks_to_process(new_block_hash).await {
+                            Ok(next_blocks_to_process) => {
+                                for block in next_blocks_to_process {
+                                    if let Err(error) = self.process_bitcoin_block(block).await {
+                                        tracing::warn!(%error, "could not process bitcoin block");
+                                    }
+                                }
+
+                                self.context
+                                    .signal(SignerEvent::BitcoinBlockObserved.into())?;
+                            }
+                            Err(error) => {
+                                tracing::warn!(%error, "could not get next blocks to process");
+                            }
+                        }
+                    }
+                    Err(error) => {
+                        tracing::warn!(%error, "error decoding new bitcoin block hash from stream");
+                    }
+                }
             }
 
             Ok::<_, Error>(())
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -116,43 +116,39 @@ where
     pub async fn run(mut self) -> Result<(), Error> {
         let mut term = self.context.get_termination_handle();
 
-        // TODO: We need to revisit all of the `?`'s in this function to ensure
-        // that we don't accidentally kill the signer by exiting this function
-        // early.
         let run = async {
             while let Some(new_block_hash) = self.bitcoin_blocks.next().await {
                 if let Err(error) = self.load_latest_deposit_requests().await {
                     tracing::warn!(%error, "could not load latest deposit requests from Emily");
                 }
 
-                // TODO: What to do when `new_block_hash?` errors? Perhaps we can
-                // handle this within a failover-stream if this indicates a problem
-                // with the stream, and then we change this back to a plain `BlockHash`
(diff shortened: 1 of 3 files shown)


signer/src/block_observer.rs
```diff
@@ -121,18 +121,38 @@ where
...
+                    tracing::warn!(%error, "could not load latest deposit requests from Emily");
+                }
 
                 // TODO: What to do when `new_block_hash?` errors? Perhaps we can
                 // handle this within a failover-stream if this indicates a problem
                 // with the stream, and then we change this back to a plain `BlockHash`
                 // instead of a `Result<>`.
-                for block in self.next_blocks_to_process(new_block_hash?).await? {
-                    self.process_bitcoin_block(block).await?;
+                match new_block_hash {
```
matteojug, line 132: nit: we could save a couple of indentations to improve readability here with a couple of early returns (continues)

signer/src/block_observer.rs
```diff
@@ -121,18 +121,38 @@ where
...
+                    Ok(new_block_hash) => {
+                        tracing::info!(%new_block_hash, "observed new bitcoin block from stream");
+
+                        match self.next_blocks_to_process(new_block_hash).await {
+                            Ok(next_blocks_to_process) => {
+                                for block in next_blocks_to_process {
+                                    if let Err(error) = self.process_bitcoin_block(block).await {
+                                        tracing::warn!(%error, "could not process bitcoin block");
```
matteojug, line 140: nit: add block id to the `warn!`

signer/src/transaction_signer.rs
```diff
@@ -169,7 +169,9 @@ where
 
                 // If we've observed a new block, we need to handle any new requests.
                 if new_block_observed {
-                    self.handle_new_requests().await?;
+                    if let Err(error) = self.handle_new_requests().await {
+                        tracing::warn!(%error, "error handling new requests; skipping this round");
+                    }
```
matteojug, line 174: we are not skipping the round here (not sure if we should though)

matteojug (approved):

signer/src/block_observer.rs
AshtonStephens: +1 on the early returns, but for the sake of time I'd settle for this in the comment above:

```
// TODO: Return early to save indentations in the following code.
```

signer/src/transaction_signer.rs
cylewitruk: i guess i meant more that we're skipping "handling new requests" for this round

signer/src/block_observer.rs
AshtonStephens: I took a quick crack at it, thoughts on this:

```rust
        // TODO: We need to revisit all of the `?`'s in this function to ensure
        // that we don't accidentally kill the signer by exiting this function
        // early.
        let run = async {
            while let Some(maybe_new_block_hash) = self.bitcoin_blocks.next().await {
                // TODO: What to do when `new_block_hash?` errors? Perhaps we can
                // handle this within a failover-stream if this indicates a problem
                // with the stream, and then we change this back to a plain `BlockHash`
                // instead of a `Result<>`.
                self.load_latest_deposit_requests().await
                    .inspect_err(|error| tracing::warn!(%error, "could not load latest deposit requests from Emily"))?;
                let new_block_hash = maybe_new_block_hash
                    .inspect_err(|error| tracing::warn!(%error, "could not get next blocks to process"))?;
                tracing::info!(%new_block_hash, "observed new bitcoin block from stream");
                let next_blocks_to_process = self.next_blocks_to_process(new_block_hash).await
                    .inspect_err(|error| tracing::warn!(%error, "could not get next blocks to process"))?;
                for block in next_blocks_to_process {
                    if let Err(error) = self.process_bitcoin_block(block).await {
                        tracing::warn!(%error, "could not process bitcoin block");
                    }
                }
                self.context.signal(SignerEvent::BitcoinBlockObserved.into())?;
            }

            Ok::<_, Error>(())
        };
```

AshtonStephens (approved):
If we don't have time to fix up the formatting I'm happy to have this merged in.

signer/src/transaction_coordinator.rs
```diff
@@ -138,7 +138,12 @@ where
                     // signer indicating that it has handled new requests.
                     Ok(SignerSignal::Event(SignerEvent::TxSigner(TxSignerEvent::NewRequestsHandled))) => {
                         tracing::debug!("received block observer notification");
-                        self.process_new_blocks().await?;
+                        match self.process_new_blocks().await {
+                            Ok(_) => continue,
+                            Err(error) => {
+                                tracing::error!(?error, "error processing new blocks; skipping this round");
+                            },
+                        };
```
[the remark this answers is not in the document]
cylewitruk: Good one :+1: [048b329](https://github.com/stacks-network/sbtc/pull/634/commits/048b32963283725d4555891d4310a1c44291d348)

signer/src/block_observer.rs
cylewitruk: [048b329](https://github.com/stacks-network/sbtc/pull/634/commits/048b32963283725d4555891d4310a1c44291d348)

signer/src/block_observer.rs
cylewitruk: Haha, had zero focus on formatting for this PR. Refactored in [048b329](https://github.com/stacks-network/sbtc/pull/634/commits/048b32963283725d4555891d4310a1c44291d348)

AshtonStephens: Lol, sorry for the nits!

Issue #396, opened 2024-08-05
[Feature]: Use actual deposit and withdrawal fields within the deposits and withdrawals.


AshtonStephens:
# Feature - Use actual deposit and withdrawal fields within the deposits and withdrawals.

## 1. Description

Use actual deposit and withdrawal fields within the deposits and withdrawals. The current implementation uses dummy fields to populate some of the table; this PR is to populate the fields correctly.

### 1.1 Context & Purpose

Use actual deposit and withdrawal fields within the deposits and withdrawals. The current implementation uses dummy fields to populate some of the table; this PR is to populate the fields correctly.

## 2. Technical Details:

The fields will need to be compiled with the `sbtc` common crate to extract data from the transactions, and will also need to populate the fields correctly. This will also require updates to the unit tests.

To use the `sbtc` library we many need to run the API as a simple `warp` service instead of spinning up a whole lambda simulation locally. This requires more investigation.

### 2.1 Acceptance Criteria:
   - [ ] The API can run locally using the common `sbtc` crate.
   - [ ] The API can properly use the fields from the transactions.
   - [ ] The PR is added and merged

AshtonStephens:
This ticket is necessary for the MVP deposit flow.

AshtonStephens:
Blocked by https://github.com/stacks-network/sbtc/pull/629, which is necessary for pulling in the `sbtc` crate to not make the `devenv` environment uncompilable.

AshtonStephens:
#629 is merged, working on this

PR #635, opened 2024-10-09
396 feature use actual deposit and withdrawal fields within the deposits and withdrawals

AshtonStephens:
## Description

Closes: #396 

## Changes

This PR lets the Emily API take parameters from the deposit and reclaim scripts.

## Testing Information

Tested with a new unit test.

Note that this change breaks a lot of integration tests - I'm adding that in a follow on PR so as not to clutter this one.

## Checklist:


emily/handler/src/api/handlers/deposit.rs
```diff
@@ -242,6 +252,42 @@ pub async fn create_deposit(
...
+/// Convert scripts to resource parameters.
+///
+/// This function is used to convert the deposit and reclaim scripts into the
+/// parameters that are stored in the database.
+fn scripts_to_resource_parameters(
+    deposit_script: &String,
+    reclaim_script: &String,
+) -> Result<ScriptParameters, Error> {
+    let bytes = hex::decode(deposit_script)?;
+    let script = ScriptBuf::from(bytes);
```
matteojug, line 271: nit: `ScriptBuf::from_hex` (also below)

sbtc/src/deposits.rs
```diff
@@ -352,9 +352,9 @@ pub struct ReclaimScriptInputs {
     /// script. It is not allowed to exceed the bounds expected for a
     /// 5-byte lock-time in bitcoin-core. It is also not allowed to have
     /// the [`SEQUENCE_LOCKTIME_DISABLE_FLAG`] bit set.
-    lock_time: i64,
+    pub lock_time: i64,
     /// The reclaim script after the <locked-time> OP_CSV part of the script
-    script: ScriptBuf,
+    pub script: ScriptBuf,
```
matteojug, line 357: Given we have some validation in the `ReclaimScriptInputs::try_new` fn below, maybe it's better to add some getters and keep the fields private to avoid people from (incorrectly) instantiate this

emily/handler/src/api/models/deposit/mod.rs
```diff
@@ -79,7 +79,7 @@ pub struct DepositParameters {
     /// the transaction.
     pub max_fee: u64,
     /// Bitcoin block height at which the reclaim script becomes executable.
-    pub lock_time: u64,
+    pub lock_time: i64,
```
cylewitruk, line 82: Why `i64` now? Can this be negative?

emily/handler/src/common/error.rs
```diff
@@ -247,6 +248,16 @@ impl From<serde_dynamo::Error> for Error {
         Error::Debug(format!("serde_dynamo::Error - {err:?}"))
     }
 }
+impl From<FromHexError> for Error {
```
cylewitruk, line 251: Not sure if this becomes irrelevant if you make the `ScriptBuf::from_hex(..)` change @matteojug suggested

sbtc/src/deposits.rs
djordon: Yeah, I agree.

emily/handler/src/api/models/deposit/mod.rs
djordon: We can actually make this a `u64` if we add a getter to the `ReclaimScriptInputs` struct for the `lock_time` that returns a `u64`. When you create that struct we check that the `lock_time` is positive, so the `TryFrom` conversion will always succeed. Hmmm, there are a few small fixes that I'm going to be pushing, I'll add this in there.

djordon: I added a getter in the reclaim script type in https://github.com/stacks-network/sbtc/pull/642 (along with some other slight changes)

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -242,6 +252,42 @@ pub async fn create_deposit(
...
+    lock_time: i64,
+}
+
+/// Convert scripts to resource parameters.
+///
+/// This function is used to convert the deposit and reclaim scripts into the
+/// parameters that are stored in the database.
+fn scripts_to_resource_parameters(
+    deposit_script: &String,
+    reclaim_script: &String,
```
djordon, line 268: Typically, you would use a `&str` here instead of a `&String`. I'm actually surprised that clippy didn't through a warning here.

AshtonStephens: Done!

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -300,3 +339,30 @@ pub async fn update_deposits(
...
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+    use sbtc::testing::{self, deposits::TxSetup};
+    use test_case::test_case;
+
+    #[test_case(15000, 500_000, 150; "All parameters are normal numbers")]
+    #[test_case(0, 0, 0; "All parameters are zeros")]
+    fn test_scripts_to_resource_parameters(max_fee: u64, amount_sats: u64, lock_time: u32) {
```
matteojug, line 353: nit: why `lock_time: u32`, given that below is used as i64 and u64?

matteojug (approved):

emily/handler/src/api/handlers/deposit.rs
AshtonStephens: Because `tx_setup` still takes an `i64`, and only an unsigned 32 or fewer bit number can always be cast to both a `u64` and a `i64`

PR #643, opened 2024-10-10
chore: move block observer testing stuff around

cylewitruk:
## Description

Writing some block observer tests and the file was getting a bit long, so I refactored out the test harness etc. to a separate file in `testing` like we do for the signer & coordinator.

Pulling this into a separate PR to make review easy.

## Changes

95% organizational, the rest is just cargo fmt + clippy + some getters on the test harness since now the fields aren't directly accessible from tests in the "real" block observer file.

## Testing Information

Nothing special, everything is as before.

## Checklist:


djordon (approved):
Looks good ✅

Issue #630, opened 2024-10-08
[Tech Debt]: Fix lack of integration tests for Signing and broadcast Stacks transactions


AshtonStephens:
## Description

PR that introduced tech-debt: #617

We didn't include integration tests for the feature in this ticket in order to sprint toward deposit flow demo.

djordon:
This was done for deposits in https://github.com/stacks-sbtc/sbtc/pull/874 and for withdrawals in https://github.com/stacks-sbtc/sbtc/pull/1403.

PR #650, opened 2024-10-10
Add integration test for complete deposit stacks tx

matteojug:
## Description

Partially address https://github.com/stacks-network/sbtc/issues/630
Add an integration test for the create-sign-send flow for complete deposit stacks tx.

## Changes

Add the integration test, also:
 - Add some `pub` around in test code to reuse some logic
 - Add dummy impl for `estimate_fee_rate` and `get_last_fee` since in the test we are using a real bitcoin client 

## Testing Information

New test is green.

## Checklist:


signer/tests/integration/transaction_coordinator.rs
```diff
@@ -0,0 +1,331 @@
...
+            });
+        })
+        .await;
+
+    // Get the private key of the coordinator of the signer set.
+    let private_key = select_coordinator(&setup.sweep_block_hash.into(), &signer_info);
+
+    // TODO: should the tx coordinator evloop use its private key instead of `config().signer.private_key`?
+    let mut coordinator_ctx = context.clone();
+    coordinator_ctx.inner.config_mut().signer.private_key = private_key;
```
matteojug, line 259: Code ref: https://github.com/stacks-network/sbtc/pull/650/files#diff-35276625f928b2224c6df10468d8afee7dcf4168f01de937a30650e521b90bc2R439

With the updated impl (where the coordinator does not sign the tx) this is relevant only if num_signers is 1. Still, as we pass the `private_key` to `TxCoordinatorEventLoop`, should we use that one instead of loading it from the config?

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    // This task logs all transactions broadcasted by the coordinator.
+    let mut wait_for_transaction_rx = broadcasted_transaction_tx.subscribe();
+    let wait_for_transaction_task =
+        tokio::spawn(async move { wait_for_transaction_rx.recv().await });
+
+    // Setup the stacks client mock to broadcast the transaction to our channel.
+    context
+        .with_stacks_client(|client| {
+            client.expect_submit_tx().times(1..).returning(move |tx| {
```
djordon, line 240: There should be exactly one `submit_tx` call here.

signer/tests/integration/transaction_coordinator.rs
djordon: They should be the same. We should remove the field from both `TxCoordinatorEventLoop` and `TxSignerEventLoop`, and always use the one in the config (just not in this PR). 

I would remove this line and the corresponding `config_mut` method that was added to the context.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -0,0 +1,331 @@
...
+        network: network.connect(),
+        private_key,
+        context_window,
+        threshold: signing_threshold as u16,
+        signing_round_max_duration: Duration::from_secs(10),
+    };
+    let tx_coordinator_handle = tokio::spawn(async move { tx_coordinator.run().await });
+
+    // TODO: here signers use all the same storage, should we use separate ones?
+    let event_loop_handles: Vec<_> = signer_info
```
djordon, line 273: Kinda. Ideally, they should all use separate `PgStore` objects pointing to separate postgres databases because it makes for a more convincing test. But this is pretty good as is, I'm convinced that the code that is being tested "works".

signer/src/context/mod.rs
```diff
@@ -121,6 +121,12 @@ where
             emily_client,
         }
     }
+
+    /// Testing method to alter the internal config
+    #[cfg(feature = "testing")]
+    pub fn config_mut(&mut self) -> &mut Settings {
+        &mut self.config
+    }
```
djordon, line 129: Hmmm, I think we can do without this.

djordon (commented):
Looks good! The new `config_mut` is unnecessary so we should remove it. Also, we should probably using separate databases in the test too.

signer/tests/integration/transaction_coordinator.rs
matteojug: Yeah makes sense, once all the pieces land on main we can add a "complete" integration test running all the signers separately.

djordon (approved):
Looks good 🚢

Issue #651, opened 2024-10-11
[Feature]: Store bootstrap signing set info in the config


djordon:
# Feature - Store bootstrap signing set info in the config

## 1. Description

When the signers first boot up they do not know:
1. Who they are allowed to talk to, and relatedly
2. Who is supposed to be in the signing set.

This ticket proposes, at least for right now, putting the public keys of the accepted signers in their config. This can solve both of the issues above: for (1), since each peer ID is based off of the public key of the peer, a signer can filter incoming connections from the peer ID; for (2) this serves as the inputs for when we can begin distributed key generation (DKG), and who the initial signing set is supposed to be.

### 1.1 Context & Purpose

During the bootstrap phase when we start running the signers, they will not have a rotate keys transaction to look to for figuring out the signing set. With this change they should still be able to identify who is in the signing set.

### 2.1 Acceptance Criteria:
   - [ ] Add new fields into the config for the bootstrap signing wallet. 
   - [ ] Add a function for loading the `SignerWallet` from the config if a `rotate-keys` transaction is not in the database.

## 3. Related Issues and Pull Requests (optional):

* https://github.com/stacks-network/sbtc/issues/590. Having this in the config will allow us to know when we can start DKG.

PR #654, opened 2024-10-11
feat: store bootstrap signing set info in the config

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/651.

## Changes

* Add two new fields to the config for the bootstrap signing set and the number of signatures required for signing transactions.
* Add a function to the `SignerWallet` type for loading the bootstrap wallet from the config.
* Update the signing set public keys in the stacks config and our test fixtures to align with the keys from https://github.com/stacks-network/sbtc/blob/5073358ca47ef4f8188af7a69e393c571fab5b0b/devenv/local/docker-compose/sbtc-signer/README.md#L26-L72

## Testing Information

I added tests for the config fields and a test for loading the wallet from the config.

## Checklist:


AshtonStephens (approved):
LGTM

signer/src/config/default.toml
aldur, line 1: Is the plan to leave this as the default config or to fail if the caller does not override this in prod? I am worried this might eventually lead to confusion.

djordon: We can't leave this as the default config in prod or testnet. Several values need to change based on the environment and who is running the signer, such as the private key, the network kind, database endpoints, and so on. Also, when running the signer, we require the user to specify the location of the config.

But we should consider renaming the file to something like `default-regtest.toml` or something like that so that no one gets the wrong idea.

Issue #652, opened 2024-10-11
[Feature]: Use bootstrap keys when missing `rotate-keys` transactions


djordon:
# Feature - Use bootstrap keys when missing `rotate-keys` transactions

## 1. Description

The signer currently fails whenever we are missing a `rotate-keys` transaction. Instead we should fallback to the a set of bootstrap public keys if the `rotate-keys` transaction is missing for signing Stacks transactions and determining the signing set. For Bitcoin transactions we should fallback to the last known distributed key generation (DKG) keys.

### 1.1 Context & Purpose

This is necessary during the bootstrap phase since we need some keys when submitting the first `rotate-keys` contract call transaction and signing Bitcoin transactions.

## 2. Technical Details:

Modify the functions that get the signing set to fall back on the bootstrap keys or DKG keys if there are no key-rotation transactions successfully confirmed in the database.

### 2.1 Acceptance Criteria:

   - [ ] For Stacks transactions, fallback to the bootstrap signing set if there are no confirmed and successful `rotate-key` transactions.
   - [ ] For Bitcoin transactions, fallback to the DKG keys and associated signing set when there are no confirmed and successful `rotate-key` transactions.

## 3. Related Issues and Pull Requests (optional):

* This depends on https://github.com/stacks-network/sbtc/issues/651.

PR #655, opened 2024-10-11
feat: use bootstrap signing set when missing rotate keys tx

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/652.

## Changes

* Add a column into the `dkg_shares` with the signing set information.
* Fallback to the last known DKG aggregate key if there are no key rotation transactions. **Edit**: I'm actually a little nervous about using the DKG keys by timestamp, since this isn't guaranteed to be consistent across signers. Might be best to tie it to the bitcoin block that it was done over. Created https://github.com/stacks-network/sbtc/issues/661 for this.
* Fallback to the bootstrap signing set if there are no key rotation transactions.

## Testing Information

~This still needs an integration test for the query and another integration test for the fallback behaviors.~

## Checklist:


signer/src/transaction_signer.rs
```diff
@@ -514,7 +504,7 @@ where
                 ..
             }) => {
                 tracing::info!("DKG ended in failure: {fail:?}");
```
cylewitruk, line 518: Nothing you did here but just noticed that this should probably be an error/warn.

signer/src/transaction_signer.rs
```diff
@@ -762,21 +752,27 @@ where
...
-        Ok(signer_set)
+        let db = self.context.get_storage();
+
+        // Get the last rotate-keys transaction from the database on the
+        // canonical Stacks blockchain (which we identify using the
+        // canonical bitcoin blockchain). If we don't have such a
+        // transaction then get the bootstrap keys from our config.
+        match db.get_last_key_rotation(chain_tip).await? {
+            Some(last_key) => Ok(last_key.signer_set.into_iter().collect()),
+            None => Ok(self.context.config().signer.bootstrap_signing_set()),
```
matteojug, line 774: Should we have an intermediate fallback here to the DKG shares, like we do in tx coordinator?

matteojug (commented):
Do we need to change the pg `get_deposit_request_signer_votes` impl to take into account the fallback strategy? Otherwise, IIUC, since we are not writing key rotations (and instead rely on the fallback, especially for the demo) we will not get deposit votes.

signer/tests/integration/postgres.rs
```diff
@@ -1364,6 +1367,79 @@ async fn get_signers_script_pubkeys_returns_non_empty_vec_old_rows() {
...
+    // Let's create some random DKG shares and store them in the database.
+    // When we fetch the last one, there is only one to get, so nothing
+    // surprising yet.
+    let shares: model::EncryptedDkgShares = fake::Faker.fake_with_rng(&mut rng);
+    db.write_encrypted_dkg_shares(&shares).await.unwrap();
+
+    let stored_shares = db.get_last_encrypted_dkg_shares().await.unwrap();
+    assert_eq!(stored_shares.as_ref(), Some(&shares));
+
+    // Now let's pretend that we somehow insert into the database some
```
cylewitruk, line 1393: This works but wouldn't it have been a little easier to just `db.write_encrypted_dkg_shares()` a few times with a few ms inbetween and tested that the newest row is consistently returned? :see_no_evil:

signer/src/transaction_signer.rs
djordon: Good catch, yes we need to update that query to use the `dkg_share` table if the `rotate_keys_transactions` table is empty.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -325,3 +331,111 @@ async fn process_complete_deposit() {
...
+    let test_data = TestData::generate(&mut rng, &[], &test_params);
+    test_data.write_to(&db).await;
+
+    // We always need the chain tip.
+    let chain_tip = db.get_bitcoin_canonical_chain_tip().await.unwrap().unwrap();
+
+    // We have no rows in the DKG shares table and no rotate-keys
+    // transactions, so this should error
+    let ans = coord.get_signer_set_and_aggregate_key(&chain_tip).await;
+    assert!(ans.is_err());
```
cylewitruk, line 385: I wouldn't be sad if this tested for the correct error (`matches!`)

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -325,3 +331,111 @@ async fn process_complete_deposit() {
...
+        .await
+        .unwrap();
+
+    let shares_signer_set: BTreeSet<PublicKey> =
+        shares.signer_set_public_keys.iter().copied().collect();
+
+    assert_eq!(shares.aggregate_key, aggregate_key);
+    assert_eq!(shares_signer_set, signer_set);
+
+    // Okay not we write a rotate-keys transaction into the database. To do
```
cylewitruk, line 404:
```suggestion
    // Okay now we write a rotate-keys transaction into the database. To do
```

signer/tests/integration/transaction_signer.rs
```diff
@@ -170,3 +183,95 @@ async fn should_be_able_to_participate_in_signing_round() {
         .collect()
         .await;
 }
+
+/// Test that [`TxSignerEventLoop::get_signer_public_keys`] falls back to
```
cylewitruk, line 187: Should this be falling back to DKG as well if no rotate-keys tx?

cylewitruk (commented):
Looks good in general, just a few minor comments/suggestions/questions

signer/src/storage/postgres.rs
```diff
@@ -981,6 +982,28 @@ impl super::DbRead for PgStore {
         .map_err(Error::SqlxQuery)
     }
 
+    async fn get_last_encrypted_dkg_shares(
```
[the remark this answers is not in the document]
djordon: 👍🏽 will update.

signer/tests/integration/postgres.rs
djordon: yeah that's better. I'll do that.

signer/src/transaction_signer.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/655/commits/7d8ae77e7f816e9e949ea8e95c611c68af19afbb.

signer/tests/integration/transaction_signer.rs
djordon: It doesn't need to. That function wants to return the currently known signing set on the stacks network. If there are no confirmed rotate-keys transaction on the canonical stacks blockchain then the Stacks network will think the signing set is the ones from our config.

cylewitruk: Ah okay, thx

signer/tests/integration/transaction_coordinator.rs
djordon: Updated in https://github.com/stacks-network/sbtc/pull/655/commits/0362a3eb90a0ea06d227866197ddec65e6e092d6

signer/tests/integration/transaction_coordinator.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/655/commits/0362a3eb90a0ea06d227866197ddec65e6e092d6.

signer/tests/integration/postgres.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/655/commits/0362a3eb90a0ea06d227866197ddec65e6e092d6

cylewitruk (approved):
Lgtm

PR #662, opened 2024-10-12
feat: bootstrapping distributed key generation

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/590.

## Changes

* Trigger distributed key generation (DKG) if we detect that it hasn't been run yet.
* Update the `get_signer_set_and_aggregate_key` function to fall back on the bootstrap signer set if DKG needs to be run, and leave out the aggregate key in that case.
* A few minor clean-up edits and renames.
* Change the `TxCoordinatorEventLoop::get_pending_requests` function to return an optional if there are no requests to respond to with.
* Change the `coordinator_public_key` function to return an `Option<PublicKey>` instead of wrapping it in a `Result`.

## Testing Information

This PR adds an integration test that checks that the DKG is actually run and the requisite values are stored in the database.

## Checklist:

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -649,15 +649,36 @@ where
         })
     }
 
+    /// Return the signing set that can make sBTC related contract calls
+    /// along with the current aggregate key to use for locking UTXOs on
+    /// bitcoin.
+    ///
+    /// The aggregate key fetched here is the one confirmed on the
+    /// canonical Stacks blockchain as part of a `rotate-keys` contract
+    /// call. It will be the public key that is the result of a DKG run. If
+    /// there are no rotate-keys transactions on the canonical stacks
+    /// blockchain, then we fall back on the last known DKG shares row in
+    /// our database, and return an error if is not found, implying that
+    /// DKG has never been run.
     #[tracing::instrument(skip(self))]
-    async fn get_signer_public_keys_and_aggregate_key(
-        &mut self,
+    pub async fn get_signer_public_keys_and_aggregate_key(
+        &self,
         bitcoin_chain_tip: &model::BitcoinBlockHash,
     ) -> Result<(PublicKey, BTreeSet<PublicKey>), Error> {
         let db = self.context.get_storage();
-        let last_key_rotation = db.get_last_key_rotation(bitcoin_chain_tip).await?;
 
-        match last_key_rotation {
+        // We are supposed to submit a rotate-keys transaction after
+        // running DKG, but that transaction may not have been submitted
+        // yet (if we have just run DKG) or it may not have been confirmed
+        // on the canonical Stacks blockchain.
+        //
+        // If the signers have already run DKG, then we know that all
+        // participating signers have completed it successfully (well, some
+        // may have failed suddenly at the end, and some may have dropped
+        // off during DKG, but yeah enough should have their DKG shares).
+        // So we can fall back on the stored DKG shares for getting the
+        // current aggregate key and associated signing set.
+        match db.get_last_key_rotation(bitcoin_chain_tip).await? {
             Some(last_key) => {
                 let aggregate_key = last_key.aggregate_key;
                 let signer_set = last_key.signer_set.into_iter().collect();
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -176,7 +176,7 @@ where
         Ok(())
     }
 
-    #[tracing::instrument(skip_all)]
+    #[tracing::instrument(skip(self))]
     async fn process_new_blocks(&mut self) -> Result<(), Error> {
         let bitcoin_chain_tip = self
             .context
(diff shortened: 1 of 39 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -673,32 +763,39 @@ where
...
-                let signer_set = shares.signer_set_public_keys.into_iter().collect();
-                Ok((shares.aggregate_key, signer_set))
+                Ok((Some(aggregate_key), signer_set))
             }
+            None => match db.get_last_encrypted_dkg_shares().await? {
+                Some(shares) => {
+                    let signer_set = shares.signer_set_public_keys.into_iter().collect();
+                    Ok((Some(shares.aggregate_key), signer_set))
+                }
+                None => Ok((None, self.context.config().signer.bootstrap_signing_set())),
```
cylewitruk, line 780: So if a signer joins the set after it's been bootstrapped, it will technically need to have the config updated to the current signer set pubkeys, right? Since it won't have any dkg shares/key rotations stored?

signer/src/transaction_coordinator.rs
```diff
@@ -721,28 +818,30 @@ pub fn given_key_is_coordinator(
...
+) -> bool {
+    coordinator_public_key(bitcoin_chain_tip, signer_public_keys) == Some(pub_key)
 }
 
 /// Find the coordinator public key
 pub fn coordinator_public_key(
     bitcoin_chain_tip: &model::BitcoinBlockHash,
     signer_public_keys: &BTreeSet<PublicKey>,
-) -> Result<Option<PublicKey>, Error> {
+) -> Option<PublicKey> {
```
cylewitruk, line 829: So what's the intended outcome if this would actually return `None`? Can the signer even function if that were the case? I'm thinking like, if other signers see this signer as the coordinator, but itself doesn't, then signing for the current block will be dead.

I guess I'm asking, should this be something-or-fatal?

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -123,8 +139,169 @@ async fn get_signer_public_keys_and_aggregate_key_falls_back() {
...
+/// 3. Check that there are no DKG shares in the database.
+/// 4. Start the [`TxCoordinatorEventLoop`] and [`TxSignerEventLoop`]
+///    processes for each signer.
+/// 5. Once they are all running, signal that DKG should be run. We signal
+///    them all because we do not know which one is the coordinator.
+/// 6. Check that we have exactly one row in the `dkg_shares` table.
+/// 7. Check that they all have the same aggregate key in the `dkg_shares`
+///    table.
+///
+/// Some of the preconditions for this test to run successfully includes
```
cylewitruk, line 166: You could circumvent this by manually grabbing the `Settings::new_from_default_config()` as `mut` and setting the bootstrap set to known values, then passing clones into the `TestContext`'s builder's `with_settings()`.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -123,8 +139,169 @@ async fn get_signer_public_keys_and_aggregate_key_falls_back() {
...
+        tokio::time::sleep(Duration::from_millis(10)).await;
+    }
+
+    // 5. Once they are all running, signal that DKG should be run. We
+    //    signal them all because we do not know which one is the
+    //    coordinator.
+    let event = SignerEvent::TxSigner(TxSignerEvent::NewRequestsHandled);
+    signers.iter().for_each(|(ctx, _, _)| {
+        ctx.get_signal_sender()
+            .send(SignerSignal::Event(event.clone()))
```
cylewitruk, line 273: nit: I think you could just `.send(TxSignerEvent::NewRequestsHandled.into())` here

cylewitruk (commented):
Just a couple Q's but looks good, nice work!

cylewitruk (commented):
Just a couple Q's but looks good, nice work!

signer/src/transaction_coordinator.rs
djordon: Yup, that is correct.

A "new" signer needs to get the current signer set public keys. This seems like a definite problem for v2, but we have lots of problems like this for v2. Oh and it might be a problem for v1 as well, but that depends on what we need to deliver.

cylewitruk: Yeah okay I just wanted to make sure I was reading it right :) We have several options we've discussed so once we get the room/prio we can attack that problem

signer/src/transaction_coordinator.rs
djordon: I believe this is something that is fatal. This should only return `None` if the `signer_public_keys` is an empty set, and that should never happen. If this does end up happening then that means there is some logic error in our code (right now even an empty bootstrap signing set in the config doesn't trigger this).

signer/tests/integration/transaction_coordinator.rs
djordon: Ahh, good to know. 🙏🏽

signer/src/transaction_coordinator.rs
cylewitruk: Nothing to do now here and now imo, but if it is truly fatal & unrecoverable then it might deserve an `expect()` so the signer operator knows something is truly wrong (if it ever manages to happen).

cylewitruk (approved):
Lgtm, I don't think I saw any dragons here anyway.

Issue #667, opened 2024-10-14
[Feature]: Propagate broadcasted messages internally


djordon:
# Feature - Propagate broadcasted messages internally

## 1. Description

Each signer often needs to be a receiver of the very same messages that it broadcasts to peer signers in the peer-to-peer network. In particular, the `TxSignerEventLoop` and `TxCoordinatorEventLoop` tasks need to receive each others’ messages, even when they originate from the same signer process.

### 1.1 Context & Purpose

The two main examples where a message needs to be sent between two tasks within the same signer process are:

1. During distributed key generation (DKG), all `TxSignerEventLoop` tasks on the network need to communicate with the signer coordinator `TxCoordinatorEventLoop` task in order for DKG to succeed.
2. During signing of both Stacks and Bitcoin transactions, the `TxCoordinatorEventLoop` task does not add its own signature but instead relies on a majority of `TxSignerEventLoop` tasks sending in their signature shares. Since these tasks don’t talk within the same process, the signer coordinator never contributes it’s own signature share.

## 2. Technical Details:

The `TxCoordinatorEventLoop` and `TxSignerEventLoop` tasks need to receive all of each others’ broadcasted messages. In libp2p, the default behavior is that you do not receive your own messages, implying that the `TxCoordinatorEventLoop` and `TxSignerEventLoop` tasks don’t communicate in the way that they need to. During tests, we set things up so that all tasks communicate with one another, so this problem isn’t captured in our tests.

There does not appear to be a simple way in libp2p to change this behavior. The `Gossipsub` libp2p  type has an [`allow_self_origin`](https://docs.rs/libp2p/latest/libp2p/gossipsub/struct.ConfigBuilder.html#method.allow_self_origin) flag, but it is not quite what we want (it allows us to receive our own messages, but doesn’t guarantee that we do).

Lastly, we do not want the `TxCoordinatorEventLoop` or the `TxSignerEventLoop` tasks to receive their own messages.

### 2.1 Acceptance Criteria:

- [ ]  `TxCoordinatorEventLoop` and `TxSignerEventLoop` tasks receive each others’ messages in the when the `network` is the `P2PNetwork`.

## **Related Issues and Pull Requests (optional):**

cylewitruk:
@djordon as it is right now the transaction coordinator doesn't process any received network messages, but it does receive the `TxSignerEvent::NewRequestsHandled` event and react on that -- is that what you were referring to regarding the coordinator?

Issue #669, opened 2024-10-14
[Feature]: Revisit `tokio::select!`s which run non-cancel-safe tasks


cylewitruk:
# Feature - Revisit `tokio::select!`s which run non-cancel-safe tasks

## 1. Description

`tokio::select!` was quick and easy for getting things started, but we run a lot of tasks using it which are not cancel-safe, i.e. if terminated in the middle of its run may leave things in a half-processed/invalid state if for example a `SIGINT` or `SIGKILL` is received.

### 1.1 Context & Purpose

As the event loops do not run within database transactions, being terminated half-way may result in invalid database state (for example, a block being stored but not its transactions).

`tokio::select!` aborts the remaining running tasks immediately upon completion/exit of any of the other adjacent tasks it's running, increasing the risk of ending up in invalid state.

## 2. Technical Details:

We should either refactor away `tokio::select!` for these cases and instead do polling ourselves for the termination signal, or refactor the affected event loops to be cancel-safe (i.e. by wrapping each critical path in a database transaction so that it rolls-back automatically if not committed).

PR #670, opened 2024-10-14
feat: propogate tx coordinator's p2p messages internally to the tx signer

cylewitruk:
## Description

Closes: #667

## Changes

- `MessageTransfer` is now `: Clone`
- Refactored the signer event loop to align with #669 (needed to make some changes to how this is run anyway due to the borrow checker)
- Two new app events: `TxSignerEvent::MessageGenerated(msg)` and `TxCoordinatorEvent::MessageGenerated(msg)` which are triggered in their respective `send_message()` functions (after sending to network succeeds).
- Signer event loop now processes all `SignerMessage`'s from both the network and the _tx coordinator_.
- Coordinator's `relay_messages_to_wsts_state_machine_until_signature_created()` loop now processes all `SignerMessage`'s from both the network and the _tx signer_.

I opted to use the signalling channel here due to that in order for the signer/coordinator to ignore their _own_ messages via the network impl we also need some sort of message source (i.e. `network`, `local::signer`, `local::coordinator`, `local::blockobserver`, etc.) and that meant a fair bit of refactoring wherever the network is used to filter/not filter on locally received messages. With that kind of solution the in-memory bus also needs a bit of refactoring so that it can retain the message source for locally re-propagated messages but sets `network` for those received from "the actual network", which I think involves splitting it up into multiple channels which talk to each other. This solution ended up a little easier and more explicit, but it seems to work, though we may want to tackle such a refactor as described above later on.

## Testing Information

There is currently an issue with the following two tests:
- transaction_signer::should_store_decisions_for_pending_deposit_requests
- transaction_signer::should_store_decisions_for_pending_withdraw_requests

These tests work fine when run individually, but consistently fail when run via `make integration-test-full`, which I'm looking into next (some race condition, I had some comments in those tests from before as well re: similar).

## Checklist:


signer/src/transaction_coordinator.rs
```diff
@@ -485,43 +487,69 @@ where
...
-            if &msg.bitcoin_chain_tip != bitcoin_chain_tip {
-                tracing::warn!(?msg, "concurrent wsts signing round message observed");
-                continue;
+            // Check the network for new messages. We don't have a `try_receive()`
+            // equivilent for the network, so we use a timeout to avoid blocking.
+            if let Ok(msg) =
+                tokio::time::timeout(Duration::from_millis(10), self.network.receive()).await
+            {
+                signer_messages.push(msg?);
             }
```
cylewitruk, line 514: This could have been done more concisely with a `tokio::select!` on `signal_rx.recv()` and `network.receive()` and skipped the buffer/for-loop, but I opted not to due to the number of messages flying around and the context switching it would entail.

signer/src/network/in_memory.rs
```diff
@@ -84,8 +95,9 @@ impl super::MessageTransfer for MpmcBroadcaster {
         let mut msg = self.receiver.recv().await.map_err(Error::ChannelReceive)?;
         tracing::trace!("[network{:0>2}] received: {}", self.id, msg);
 
-        while Some(&msg.id()) == self.recently_sent.front() {
-            self.recently_sent.pop_front();
+        //let mut recently_sent = ;
```
djordon, line 98: Forgot this

signer/src/transaction_coordinator.rs
djordon: Actually, I need to revisit this part of the codebase anyway so I'll give this a quick look over to see if it is indeed better.

djordon (approved):
Looks good. We'll give another look over some of our use/non-use of `select!` later.

signer/src/network/in_memory.rs
cylewitruk: :see_no_evil: removed in [c6110e9](https://github.com/stacks-network/sbtc/pull/670/commits/c6110e91f1ba53184468bc91c6e5e3e30edbe599)

signer/src/transaction_coordinator.rs
cylewitruk: It's a lot cleaner looking, I just did it. But then the `process_complete_deposit()` integration test started failing.

What I think might be happening is that `network.receive()` and `signal_rx.recv()` start processing concurrently in the test and one of them gets aborted (by `select!`), meaning the coordinator then misses the message from the receive that got aborted.

I didn't want to spend time on it pre-demo, so I'll go ahead and merge this solution for now and we can revisit :+1:

PR #672, opened 2024-10-14
fix: add a consolidated receiver

djordon:
## Description

Fixes an issue that we were seeing on main.

## Changes

* Add a method to retrieve the number of signatures received by a transaction locked with a multi-sig "script" on Stacks.
* Add a consolidated receive function to the transaction coordinator object.

## Checklist:


cylewitruk (approved):
Nice refactor! lgtm

PR #676, opened 2024-10-17
refactor: minor logging fixes

djordon:
## Description

This PR contains some of the simple changes that we've had from the demo branch.

## Changes

* A few more logs, better span names and such.
* Use the right emily port by default.
* Implement more `BitcoinInteract` methods for the fallback client.
* Use the existing bitcoin block hash at the time of that `get_btc_state` is called.

## Testing Information

This is a refactor, so if tests pass then everything should be fine.

## Checklist:

### DIFF
--- a/signer/src/bitcoin/client.rs
+++ b/signer/src/bitcoin/client.rs
@@ -63,11 +63,17 @@ impl BitcoinInteract for ApiFallbackClient<BitcoinCoreClient> {
     }
 
     async fn estimate_fee_rate(&self) -> Result<f64, Error> {
-        todo!() // TODO(542)
+        // TODO(542): This function is supposed to incorporate other fee
+        // estimation methods, in particular the ones in the
+        // src/bitcoin/fees.rs module.
+        self.exec(|client, _| async { client.estimate_fee_rate(1) })
+            .await
+            .map(|feerate| feerate.sats_per_vbyte)
     }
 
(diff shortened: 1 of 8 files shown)


signer/src/bitcoin/client.rs
```diff
@@ -63,11 +63,17 @@ impl BitcoinInteract for ApiFallbackClient<BitcoinCoreClient> {
...
 
     async fn estimate_fee_rate(&self) -> Result<f64, Error> {
-        todo!() // TODO(542)
+        // TODO(542): This function is supposed to incorporate other fee
+        // estimation methods, in particular the ones in the
+        // src/bitcoin/fees.rs module.
+        self.exec(|client, _| async { client.estimate_fee_rate(1) })
+            .await
+            .map(|feerate| feerate.sats_per_vbyte)
     }
```
matteojug, line 72: qq: (also for `broadcast_transaction`) most of the times the impl for `ApiFallbackClient<C>` is just a passthrough to `C`, why here (and in other places) we are implementing "business logic" instead?

matteojug (approved):

signer/src/bitcoin/client.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/676/commits/9d6f7c14bab26983cc5db4d3192b6dc6653e8b91

PR #677, opened 2024-10-17
Feat/Emily can populate chainstate table with the height and hash inferred from deposit and withdrawal updates.

AshtonStephens:
## Description

Changes Emily to accept deposit / withdrawal updates that have unknown chainstate entries, and updates the chainstate
entries to match the chainstates that can be inferred from the deposit / withdrawal updates.

Closes: #?

## Changes

- Deposit update flow will update chainstate
- Withdrawal update flow will update chainstate

## Testing Information

Tested with integration tests.

## Checklist:

### DIFF
--- a/emily/handler/src/database/entries/deposit.rs
+++ b/emily/handler/src/database/entries/deposit.rs
@@ -282,7 +282,7 @@ impl DepositEvent {
         // Determine if event is valid.
         if self.stacks_block_height > next_event.stacks_block_height {
             let err_msg = format!(
-                "Attempting to update a deposit with a block height earlier than it should be..\n
+                "Attempting to update a deposit with a block height earlier than it should be.\n
                 latest_existing_event:\n{self:?}\n
                 newest_event:\n{next_event:?}"
             );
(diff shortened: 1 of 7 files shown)


emily/handler/src/database/entries/deposit.rs
```diff
@@ -411,16 +418,48 @@ impl TryFrom<UpdateDepositsRequestBody> for ValidatedUpdateDepositsRequest {
...
-        let deposits = update_request
+        let mut deposits: Vec<_> = update_request
             .deposits
             .into_iter()
             .map(|i| i.try_into())
             .collect::<Result<_, Error>>()?;
+
+        // Order the updates by order of when they occur so that it's
+        // as though we got them in chronological order.
+        deposits.sort_by_key(|update: &ValidatedDepositUpdate| update.event.stacks_block_height);
```
matteojug, line 429: Will this mess up the order of the returned deposits in `UpdateDepositsResponse`? wrt the order in the client request.

matteojug (commented):
> Changes Emily to accept deposit / withdrawal updates that have unknown chainstate entries

Was this change implemented? IIUC the request body is still expecting `last_update_height`/`last_update_block_hash`, and so far we were already able to send updates to emily (without sending chainstate updates), so I'm not exactly sure I got this point.

emily/handler/src/database/entries/deposit.rs
AshtonStephens: I actually assumed it wouldn't matter, and that it wouldn't need to be in the same order, but if we care about that I can add that change

AshtonStephens: Fixed ~

emily/handler/src/database/entries/deposit.rs
```diff
@@ -279,17 +281,21 @@ impl DepositEvent {
     pub fn ensure_following_event_is_valid(&self, next_event: &DepositEvent) -> Result<(), Error> {
         // Determine if event is valid.
         if self.stacks_block_height > next_event.stacks_block_height {
-            return Err(Error::InconsistentState(Inconsistency::ItemUpdate(
-                "Attempting to update a deposit with a block height earlier than it should be."
-                    .into(),
-            )));
+            let err_msg = format!(
+                "Attempting to update a deposit with a block height earlier than it should be..\n
```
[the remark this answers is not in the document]
AshtonStephens: Done.

AshtonStephens:
> Was this change implemented? IIUC the request body is still expecting last_update_height/last_update_block_hash, and so far we were already able to send updates to emily (without sending chainstate updates), so I'm not exactly sure I got this point.

Oof, thought I responded to this one - the Emily API still expects these fields. This PR title misleading. In the past the Emily API required prior knowledge of the associated block by height and hash, but with this change it will infer the chainstates it should know about by taking them from the deposits and withdrawals. This means it no longer requires prior knowledge of the blocks, but it still requires the block data.

emily/handler/tests/integration/deposit.rs
```diff
@@ -493,13 +496,23 @@ async fn update_deposits_updates_chainstate() {
...
             last_update_block_hash: format!("hash_{}", update_block_height),
-            last_update_height: update_block_height,
+            last_update_height: update_block_height as u64,
             status: update_status.clone(),
             status_message: update_status_message.into(),
         };
         deposit_updates.push(deposit_update);
     }
 
+    // Order the, pecularily so that they are not in order.
```
matteojug, line 506: nit: missing word, also in the withdrawal test.

matteojug (approved):

Issue #522, opened 2024-09-13
[Feature]: Block observer is skipping certain deposit requests


technovision99:
# (High) Block observer is skipping certain deposit requests

## 1. Description
The block observer iterates over new bitcoin blocks:
https://github.com/stacks-network/sbtc/blob/033e2cbd2def829a3e30df77bc274b325d68cdbe/signer/src/block_observer.rs#L78-L97

It then calls into `process_bitcoin_block` which in turn calls `extract_deposit_requests`:
https://github.com/stacks-network/sbtc/blob/033e2cbd2def829a3e30df77bc274b325d68cdbe/signer/src/block_observer.rs#L157-L185

Note that this means any deposits fetched from the API that reference an old block will effectively be skipped. We should ensure that all deposits can still be processed even if they reference blocks before the new bitcoin blocks fetched from the bitcoin client. 

As discussed, please deduplicate if there's already a tracking bug for this.

djordon:
Yeah this slipped through the cracks. Thanks for opening the issue.

cylewitruk:
Pausing work on this in favor of #667 for demo

djordon:
This isn't actually closed. We need to make sure we get the bitcoin block associated with each deposit transaction first.

djordon:
Actually, this may indeed be fixed if we do https://github.com/stacks-network/sbtc/issues/685.

djordon:
Quick update here, I don't think https://github.com/stacks-network/sbtc/issues/685 solves the outstanding issue. It might be worth doing for other reasons but this ticket probably isn't one of them. Some notes
1. The block observer no longer skips certain deposits. A request is only written to the database if it has been confirmed.
2. The signer does not check to see if the block confirming the deposit is in the database before attempting to write the deposit transaction to the database. That is the outstanding problem with the current code. We have a foreign key constraint on the `bitcoin_transactions` table that will be violated if we do not know about the block.
3. I don't think [#685](https://github.com/stacks-network/sbtc/issues/685) solves this. A transaction may be confirmed on a block that is not on the best chain, but I believe that bitcoin-core only sends ZeroMQ notifications when a block has been added to the best chain (from the [docs here](https://github.com/bitcoin/bitcoin/blob/2a52718d734cf65aaeeb18f627547e5bca3483f4/doc/zmq.md)). This means that it may be possible for a deposit request to pass validation but where the signer do not have the block in our database.

[...]

PR #675, opened 2024-10-17
fix: fetch `scriptPubKeys` for deposit transaction inputs

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/674

## Changes

* Fetch the transaction info for deposits, and store the prevout `scriptPubKey`s.
* Do not take note of deposit requests that have not confirmed. Before we would keep them around in memory if they were in the mempool and passed validation.
* Added more helpful error messages to our error variants.

## Checklist:

- [x] I have performed a self-review of my code

PR #679, opened 2024-10-18
feat: block observer processes confirmed deposit requests

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/522.

This was largely fixed in https://github.com/stacks-network/sbtc/pull/675.

## Changes

* Have the `load_latest_deposit_requests` function fetch and store all validated deposits that have been confirmed.
* Remove unnecessary fields from the `BlockObserver`.
* Load data from Emily after processing the block.

## Testing Information

The tests were modified after two functions were essentially merged.

## Checklist:

### DIFF
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -93,6 +93,10 @@ impl DepositRequestValidator for CreateDepositRequest {
             return Ok(None);
         };
 
+        // TODO(515): After the transaction passes validation, we need to
+        // check whether we know about the public key in the deposit
+        // script.
+
         Ok(Some(Deposit {
             info: self.validate_tx(&tx_info.tx)?,
             tx_info,
@@ -209,8 +213,9 @@ where
                 },
                 // This happens when we cannot find the associated
                 // transaction confirmed on a bitcoin block, or when we
-                // encounted some unexpected error when reaching out the
-                // bitcoin-core.
+                // encounted some unexpected error when reaching out to
+                // bitcoin-core or our database. We've already logged the
+                // error above, so there is nothing more to do.
                 Err(_) | Ok(None) => {}
             }
         }
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -185,7 +185,7 @@ where
     }
 
     /// Fetch deposit requests from Emily and store the validated ones into
-    /// the database. Update Emily about the ones that fail validation.
+    /// the database.
     #[tracing::instrument(skip(self))]
     async fn load_latest_deposit_requests(&mut self) -> Result<(), Error> {
         let mut deposit_requests = Vec::new();
@@ -216,8 +216,6 @@ where
                 Err(_) | Ok(None) => {}
             }
         }
-
-        // TODO: Update emily about the requests that have failed validation.
 
         self.store_deposit_requests(deposit_requests).await?;
 
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -202,15 +202,15 @@ where
                 // request.
                 Ok(Some(deposit)) => deposit_requests.push(deposit),
                 // This happens when the transaction has failed the first
-                // step of validation or has passed the first step but we
-                // don't recongnize the x-only public key in the deposit
+                // step of validation or has passed the first step, but we
+                // don't recognize the x-only public key in the deposit
                 // script.
                 Err(Error::SbtcLib(_)) | Err(Error::UnknownAggregateKey(_, _)) => {
                     failed_requests.push(request.outpoint);
                 }
                 // This happens when we cannot find the associated
                 // transaction confirmed on a bitcoin block, or when we
-                // encounted some unexpected error when reaching out to
+                // encountered some unexpected error when reaching out to
                 // bitcoin-core or our database. We've already logged the
                 // error above, so there is nothing more to do.
                 Err(_) | Ok(None) => {}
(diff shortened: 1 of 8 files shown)


signer/src/block_observer.rs
```diff
@@ -187,24 +184,41 @@ where
...
+            match deposit {
+                // Happy path, we've successfully validated the deposit
+                // request.
+                Ok(Some(deposit)) => deposit_requests.push(deposit),
+                // This happens when the transaction has failed the first
+                // step of validation or has passed the first step but we
+                // don't recongnize the x-only public key in the deposit
+                // script.
+                Err(Error::SbtcLib(_)) | Err(Error::UnknownAggregateKey(_, _)) => {
+                    failed_requests.push(request);
```
djordon, line 209: I'm not entirely sure what, if anything, we should do with these failed requests. Should we tell Emily, persist them to the database, store them in memory but with a capped size, or do nothing?

signer/src/block_observer.rs
```diff
@@ -187,24 +184,41 @@ where
...
+            match deposit {
+                // Happy path, we've successfully validated the deposit
+                // request.
+                Ok(Some(deposit)) => deposit_requests.push(deposit),
+                // This happens when the transaction has failed the first
+                // step of validation or has passed the first step but we
+                // don't recongnize the x-only public key in the deposit
+                // script.
+                Err(Error::SbtcLib(_)) | Err(Error::UnknownAggregateKey(_, _)) => {
+                    failed_requests.push(request.outpoint);
```
djordon, line 209: I'm not entirely sure what, if anything, we should do with these failed requests. Should we tell Emily, persist them to the database, store them in memory but with a capped size, or do nothing?

signer/src/block_observer.rs
```diff
@@ -187,24 +184,41 @@ where
...
+                }
+                // This happens when we cannot find the associated
+                // transaction confirmed on a bitcoin block, or when we
+                // encounted some unexpected error when reaching out to
+                // bitcoin-core or our database. We've already logged the
+                // error above, so there is nothing more to do.
+                Err(_) | Ok(None) => {}
```
cylewitruk, line 216: I'd get rid of the `inspect_err` above and log here instead of splitting it up

signer/src/block_observer.rs
cylewitruk: I think we should report them to Emily so that Emily can stop giving it to us each round.

cylewitruk (commented):
Only got half way, gotta get kit to school etc, will continue in a little bit ^^

signer/src/block_observer.rs
```diff
@@ -266,37 +280,31 @@ where
...
-        block_hash: BlockHash,
-    ) -> Result<(), Error> {
-        let (deposit_request, deposit_request_txs) = txs
-            .iter()
-            .filter_map(|tx| self.deposit_requests.remove(&tx.compute_txid()))
-            .flatten()
+    /// For each of the deposit requests, persist the corresponding
+    /// transaction and the parsed deposit info into the database.
+    ///
+    /// Since this functions writes to the `bitcoin_transactions` table, we
```
cylewitruk, line 290:
```suggestion
    /// Since this function writes to the `bitcoin_transactions` table, we
```

cylewitruk (commented):
Looks good in general, though I can't see that we're explicitly testing #522's:

> any deposits fetched from the API that reference an old block will effectively be skipped

Or am I missing it?

signer/src/block_observer.rs
```diff
@@ -598,21 +601,25 @@ mod tests {
...
+        {
+            let db = storage.lock().await;
+            assert_eq!(db.deposit_requests.len(), 0);
+        }
+
         block_observer.load_latest_deposit_requests().await.unwrap();
-        // Only the transaction from tx_setup0 was valid.
-        assert_eq!(block_observer.deposit_requests.len(), 1);
+        // Only the transaction from tx_setup0 was valid. Note that, since
+        // we are not using a real block hash stored in the database. Out
```
matteojug, line 613:
```suggestion
        // we are not using a real block hash stored in the database. Our
```

matteojug (approved):

signer/src/block_observer.rs
djordon: > I think we should report them to Emily so that Emily can stop giving it to us each round. Otherwise we spam the logs every block with the same validation failures.
> 
> Two other plus's to having it to emily are 1) user feedback, 2) analytics.
> 
> But for short-term we could just have an [indexset](https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html) or something and start popping after like 1000 entries.

The report-to-emily change makes a lot of sense, but I was hesitant to do that because one potentially mistaken signer could then effectively block a deposit. I think we need to open a design issue for this in order for us to figure out what to do long term.

I like the short-term change solution, but I think it should be done in a separate PR with it's own ticket describing why it's necessary and so on.

cylewitruk: Yeah I thought about that multiple signers too... I was thinking that it would actually be best for Emily to store updates from signers in an array, if nothing else than for centralized debugging, but we could also use majority rules for when a deposit is actually rejected.

signer/src/block_observer.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/679/commits/ef23cb0dc04ee25c8f76fc06cd7d6ac7acc3faa9

djordon:
I added an integration test in https://github.com/stacks-network/sbtc/pull/679/commits/5808ed67b947d082d17cbab1c16b01106ada1a25.

signer/src/block_observer.rs
djordon: I created https://github.com/stacks-network/sbtc/issues/683 for the signers side, still need to create one for any potential emily work.

djordon: Hmmm, I'm not sure that we want to do this on the Emily side either. I'll leave this as is for now and we can settle on a long term plan sometime later.

Issue #686, opened 2024-10-20
[Feature]: Enable validation of sign requests for contract calls


djordon:
# Feature - Enable validation of sign requests for contract calls

## 1. Description

We have the meat of the code for validating sign requests for `complete-deposit` and `accept-withdrawal-request` contract calls, but it hasn't been enabled (we exit early in the function that does it). This issue is to finish off the function is supposed to do the validation of a sign request for stacks transactions.

### 1.1 Context & Purpose

Each signer needs to validate the contents of a sign request before signing it. The code that does this is needs to fed the correct sign request context and not exit early.

## 2. Technical Details:

Correctly create the request context in the `TxSignerEventLoop::assert_valid_stacks_tx_sign_request` function.

### 2.1 Acceptance Criteria:
   - [ ] We execute the validation code for all Stacks sign requests.

## 3. Related Issues and Pull Requests (optional):

This relates to:
* Validate all sign requests https://github.com/stacks-network/sbtc/issues/255 and https://github.com/stacks-network/sbtc/issues/476

Issue #517, opened 2024-09-12
[Feature]: Missing integrity protection for Stacks transactions issued by signer


evonide:
# (High) Missing integrity protection for Stacks transactions issued by signer

## 1. Description

In order to issue Stacks transactions, the signer has to sign the hash over members in structs such as `StacksTransactionSignRequest`:
https://github.com/stacks-network/sbtc/blob/033e2cbd2def829a3e30df77bc274b325d68cdbe/signer/src/message.rs#L116-L129

However, the current hash seems to exclude a struct member such as `contract_call` and instead relies on a custom `digest` member:

https://github.com/stacks-network/sbtc/blob/033e2cbd2def829a3e30df77bc274b325d68cdbe/signer/src/message.rs#L221-L231

Currently, the verification of this digest seems to be missing. We assume this is work in progress but still file an explicit tracking bug as **this is a critical check that needs to take place**. Without it signer messages are not fully integrity protected and could be partially spoofed.

From a cursory look we were unable to understand the necessity of having a custom `digest` instead of adding the `contract_call` to be included in the hash here as well.

djordon:
> Currently, the verification of this digest seems to be missing. We assume this is work in progress but still file an explicit tracking bug as **this is a critical check that needs to take place**. Without it signer messages are not fully integrity protected and could be partially spoofed.

Yeah, this just work in progress. All of that will be checked, good call out though.

> From a cursory look we were unable to understand the necessity of having a custom `digest` instead of adding the `contract_call` to be included in the hash here as well.

I vied using the digest as sufficient. If the `contract_call` transaction generates the same digest then using the digest in place of the actual transaction fine. I guess I thought that it was analogous to many cryptographic signature schemes where the signature is over a digest and the check is that the transaction generates the digest and that the signature is valid, but maybe that analogy is inappropriate here? Anyway, is that reasoning incorrect/problematic?

evonide:
The custom `digest` seems to introduces redundancy. Most of the struct members are already being hashed and signed, ensuring the integrity of `StacksTransactionSignRequest`.

In the current approach, the sender computes a digest of the `contract_call`, adds it to the transaction, and signs the hash of the remaining fields. The receiver then verifies both the signature on the partial transaction and checks that the digest matches the `contract_call`.

A simpler approach would be to sign the full transaction, including the `contract_call`, in one step. This eliminates the need for a separate digest and verification, reducing complexity and potential redundancy.

[...]

PR #694, opened 2024-10-22
feat: enable validation for contract calls

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/686.

## Changes

* Add a column to the `dkg_shares` for the number of signature shares that are necessary for a signature.
* Clarify some of the docstrings.
* Enable validation for stacks transactions by finishing up the function that does validation.
* Add a new validation check where each signer checks whether they are in the signer set.

## Testing Information

I added an integration test for the new validation check. I need to update one more integration test though.

## Checklist:


protobufs/stacks/signer/v1/requests.proto
```diff
@@ -16,7 +16,8 @@ message BitcoinTransactionSignRequest {
 
 // Represents a request to sign a Stacks transaction.
 message StacksTransactionSignRequest {
-  // The aggregate public key that will sign the transaction.
+  // This is the bitcoin aggregate key that was output from DKG. It is used
+  // to identifyy the signing set for the transaction.
```
AshtonStephens, line 20:
```suggestion
  // to identify the signing set for the transaction.
```

signer/src/proto/generated/stacks.signer.v1.rs
```diff
@@ -69,7 +69,8 @@ pub struct BitcoinTransactionSignRequest {
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct StacksTransactionSignRequest {
-    /// The aggregate public key that will sign the transaction.
+    /// This is the bitcoin aggregate key that was output from DKG. It is used
+    /// to identifyy the signing set for the transaction.
```
AshtonStephens, line 73:
```suggestion
    /// to identify the signing set for the transaction.
```

signer/src/storage/model.rs
```diff
@@ -339,6 +339,10 @@ pub struct EncryptedDkgShares {
     pub public_shares: Bytes,
     /// The set of public keys that were a party to the DKG.
     pub signer_set_public_keys: Vec<PublicKey>,
+    /// The threshold number of signature shares required to generate a
+    /// Schnorr signature.
```
AshtonStephens, line 343: Can you add a little more detail for simpletons like me? Something like:

```
    /// The threshold number of signature shares required to generate a
    /// Schnorr signature, where a signature share is defined as  ...
```

signer/src/stacks/contracts.rs
```diff
@@ -73,8 +73,8 @@ pub struct ReqContext {
     /// [`AsContractCall::validate`] function, but is here for logging and
     /// tracking purposes.
     pub origin: PublicKey,
-    /// The combined public key of all the signers in the current
-    /// signing set.
+    /// This is the bitcoin aggregate key that was output from DKG. It is
```
AshtonStephens, line 76: I think we should still specify that this is a public key in the description despite the fact it's clear from the type of the field as well.

```suggestion
    /// This is the bitcoin aggregate public key that was output from DKG. It is
```

signer/src/transaction_signer.rs
```diff
@@ -463,7 +465,7 @@ where
         let multi_sig = MultisigTx::new_tx(&request.contract_call, &wallet, request.tx_fee);
         let txid = multi_sig.tx().txid();
 
-        // TODO: Make this more robust. The signer that recieves this won't
+        // TODO: Make this more robust. The signer that receives this won't
```
AshtonStephens, line 468: If there's a ticket for this can you link it like `TODO(#12345):`, and if there isn't a ticket can you very quickly make a `tech debt` ticket? It should be fast, you can just put minimal details into it but we don't want to lose track.

signer/tests/integration/transaction_signer.rs
```diff
@@ -359,3 +363,91 @@ async fn handle_pending_deposit_request_address_script_pub_key() {
...
+    // Let's get the blockchain data into the database.
+    let chain_tip: BitcoinBlockHash = setup.sweep_block_hash.into();
+    backfill_bitcoin_blocks(&db, rpc, &chain_tip).await;
+
+    // This is all normal happy path things that need to happen in order to
+    // pass validation.
+    setup.store_deposit_tx(&db).await;
+    setup.store_sweep_tx(&db).await;
+    setup.store_dkg_shares(&db).await;
+    setup.store_deposit_request(&db).await;
```
AshtonStephens, line 398: Naive question; is this going to be necessary for a class of tests, or is this unique to this one? The comment makes me think it'll be necessary for a class of tests, at which point I think this is a good contender to be in something like a test setup function.

signer/tests/integration/transaction_signer.rs
```diff
@@ -359,3 +363,91 @@ async fn handle_pending_deposit_request_address_script_pub_key() {
 
     testing::storage::drop_db(db).await;
 }
+
+/// Test that [`TxSignerEventLoop::assert_valid_stacks_tx_sign_request`]
+/// errors when the signer is not in the signer set.
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
+#[tokio::test]
+async fn signing_set_validation_check_for_stacks_transactions() {
```
AshtonStephens, line 371: Love the comments on this tests steps :)

signer/tests/integration/transaction_signer.rs
```diff
@@ -359,3 +363,91 @@ async fn handle_pending_deposit_request_address_script_pub_key() {
...
+    // Let's create a proper sign request.
+    let request = StacksTransactionSignRequest {
+        aggregate_key: setup.aggregated_signer.keypair.public_key().into(),
+        contract_call: signer::stacks::contracts::ContractCall::CompleteDepositV1(req),
+        // The nonce and tx_fee aren't really validated against anything at
+        // the moment.
+        nonce: 1,
+        tx_fee: 100_000,
+        // This can probably be removed, but it's not important to remove
+        // now, we should switch to protobuf messages and remove it then.
```
AshtonStephens, line 425: This comment might be fine, but if you haven't already consider if this is worth making a tech-debt type issue for to track, and if so it would be good to link it here with a `TODO`

signer/src/wsts_state_machine.rs
```diff
@@ -148,13 +148,19 @@ impl SignerStateMachine {
             wsts::util::encrypt(&self.0.network_private_key.to_bytes(), &encoded, rng)
                 .map_err(|_| error::Error::Encryption)?;
 
+        let signature_share_threshold: u16 = self
+            .threshold
+            .try_into()
+            .map_err(|_| Error::TypeConversion)?;
```
AshtonStephens, line 154: Not a nit, just a question: 

I see you mostly do the type casts explicitly at the moment that something could fail instead of having a `From` definition for the error that the function returns. Why?

protobufs/stacks/signer/v1/requests.proto
djordon: Oh man, for some reason I thought I got all of these.

signer/src/transaction_signer.rs
djordon: We don't. And I should remove or reword this TODO.

I'm leaning toward removing the digest field since it will be unnecessary if we just switch to signing the entire message bytes. This work is basically https://github.com/stacks-network/sbtc/issues/517, and is super easy to do. In fact, I will do it now.

signer/tests/integration/transaction_signer.rs
djordon: I'm going to fix #517 and remove this.

signer/src/transaction_signer.rs
djordon: Actually, I won't do this now, and just update the todo.

signer/tests/integration/transaction_signer.rs
djordon: Actually, I'm just going to add a TODO https://github.com/stacks-network/sbtc/pull/694/commits/c3ceea7f8a43bcdd8ef975170e6c5d9aaf965461.

signer/src/transaction_signer.rs
djordon: Added a TODO and reworded this in https://github.com/stacks-network/sbtc/pull/694/commits/c3ceea7f8a43bcdd8ef975170e6c5d9aaf965461.

signer/src/stacks/contracts.rs
djordon: https://github.com/stacks-network/sbtc/pull/694/commits/c3ceea7f8a43bcdd8ef975170e6c5d9aaf965461

signer/src/storage/model.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/694/commits/c3ceea7f8a43bcdd8ef975170e6c5d9aaf965461.

signer/src/proto/generated/stacks.signer.v1.rs
djordon: https://github.com/stacks-network/sbtc/pull/694/commits/c3ceea7f8a43bcdd8ef975170e6c5d9aaf965461

signer/src/wsts_state_machine.rs
djordon: I tend to do this for a few reasons:
1. The error returned from the `try_into` is not particularly informative, so having the dedicated variant is not super helpful.
2. I think implementing `From` for an error variant is not the right pattern for us to have. Instead the goal is to construct error variants with enough context data for us to debug when we encounter the error. And `From` impls, particularly `From` impls for foreign types, encourages the wrong behavior.
3. This particular error is not actually "real". The threshold is a `u32` type, but it is created from a `u16`, so the conversion back should always happen. So I don't think it's necessary to add context and all of that, I'd rather get rid of the error.

Because of (3) I could do an `as u16` cast, but I'm hesitant because it's kind of bad practice (as in we should be cautious of using them). Perhaps it's warranted in this case with a comment explaining why it is fine.

signer/tests/integration/transaction_signer.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/694/commits/4870b46563051b1fcf6805c1780cfd37f0bfa7a2.

Issue #697, opened 2024-10-23
[Feature]: Include all deposit variables in deposit request table


djordon:
# Feature - Include all deposit variables in deposit request table

## 1. Description

This is a subtask of https://github.com/stacks-network/sbtc/issues/286.

We do not include the x-only public key and the lock time in the `deposit_requests` table and it needs to be there.

### 1.1 Context & Purpose

We will need to filter requests based off of lock-time, and we need to use the x-only-public key when constructing a signature.

## 2. Technical Details:
 This is mainly adding and populating two more columns in the database.

### 2.1 Acceptance Criteria:
   - [ ] Make sure we record the x-only public key and the lock time embedded in the deposit request `scriptPubKey`.

## 3. Related Issues and Pull Requests (optional):

Well, https://github.com/stacks-network/sbtc/issues/286.

Issue #380, opened 2024-08-01
[Feature]: Handle lock times in deposit requests


djordon:
# Feature - Handle lock times in deposit requests

## 1. Description

The signers need to properly handle the lock times encoded in their deposit scripts in order to properly filter out requests that can be reclaimed "soon".

### 1.1 Context & Purpose

Right now the signers fetch all deposit requests that have been confirmed for a certain number bitcoin blocks from the chain tip. This is has two problems:
1. We haven't fixed the relative lock time that depositors must use in their reclaim script. One depositor can lock their reclaim for 150 blocks, while another could for 240 blocks. But even if we fixed the reclaim blocks to, say, 150 blocks (rejecting all deposits that use another value), then we'd still have the following issue.
2. Relative lock times can be interpreted as either a number of bitcoin blocks or as multiple of 512 seconds measured from the median time past (MTP, the median time of the previous 11 blocks) of the block that confirmed the transaction. The spender of the coins gets to choose, so we need to track the MTP and react on both MTP and the number of confirmed blocks.

## 2. Technical Details:

We need to track each blocks `time` and filter deposit requests based both their relative height and the actual time by tracking by the MTP.

### 2.1 Acceptance Criteria:
   - [ ] Our database bitcoin block tables track the blocks time so that we can calculate the MTP.
   - [ ] The database tables for deposit requests store the lock time embedded in the transaction. It's implicitly in the reclaim script but we should break it out.
   - [ ] When fetching pending deposit requests we filter based on the MTP as well.

AshtonStephens:
I'm planning to do this with a few atomic changes:

1. Add a field to the signer config [here](https://github.com/stacks-network/sbtc/blob/32df425a341121c3bcdd45f12cd6adc71c08d7da/signer/src/config/mod.rs#L215) that specifies how far from the reclaim script's lock time the deposit needs to be for the signer to accept it. This will be in seconds and we'll interpret blocks to each be 512 seconds
2. Add a `u32` `time` field to the [`BitcoinBlock`](https://github.com/stacks-network/sbtc/blob/32df425a341121c3bcdd45f12cd6adc71c08d7da/signer/src/storage/model.rs#L21) representation in our database that we'll populate [here](https://github.com/stacks-network/sbtc/blob/32df425a341121c3bcdd45f12cd6adc71c08d7da/signer/src/block_observer.rs#L427) with `block.header.time` from the Bitcoin library.
[...]

PR #699, opened 2024-10-23
feat: include all variables in deposit request table

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/697.

## Changes

* Add `lock_time` and `signer_public_key` columns to the `deposit_requests` table.
* Add a new `PublicKeyXOnly` type. Maybe we should rename it.

## Testing Information

This is a minor refactor. If tests pass then we should be okay.

## Checklist:

### DIFF
--- a/signer/src/storage/model.rs
+++ b/signer/src/storage/model.rs
@@ -13,6 +13,7 @@ use stacks_common::types::chainstate::StacksBlockId;
 use crate::block_observer::Deposit;
 use crate::error::Error;
 use crate::keys::PublicKey;
+use crate::keys::PublicKeyXOnly;
 
 /// Bitcoin block.
 #[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord, sqlx::FromRow)]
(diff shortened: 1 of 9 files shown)


signer/src/storage/postgres.rs
```diff
@@ -583,6 +585,8 @@ impl super::DbRead for PgStore {
               , deposit_requests.recipient
               , deposit_requests.amount
               , deposit_requests.max_fee
+              , deposit_requests.lock_time
```
AshtonStephens, line 588: Hey, I was going to do that! lol

signer/src/storage/model.rs
```diff
@@ -71,6 +72,13 @@ pub struct DepositRequest {
     #[sqlx(try_from = "i64")]
     #[cfg_attr(feature = "testing", dummy(faker = "100..100_000"))]
     pub max_fee: u64,
+    /// The relative lock time in the reclaim script.
+    #[sqlx(try_from = "i64")]
+    #[cfg_attr(feature = "testing", dummy(faker = "3..u32::MAX as u64"))]
+    pub lock_time: u64,
```
AshtonStephens, line 78: So I understand, this has the same dual meaning as you specified in #380 around the `512` second per block interpretation yeah?

AshtonStephens (approved):
LGTM.

The additional fields being added is tested inherently by the existing tests because the structs would fail to be created if the database didn't add and store these fields properly right?

signer/src/storage/model.rs
djordon: Yup

djordon:
> The additional fields being added is tested inherently by the existing tests because the structs would fail to be created if the database didn't add and store these fields properly right?

Yes that's right.

Issue #693, opened 2024-10-22
[Chore]: Docker Container Publishing


AshtonStephens:
# Chore - Docker Container Publishing

## 1. Description

Publish docker containers from sBTC repository for the following image types:
- sBTC Signer
- Blocklist Client
- Emily Server (optional)

## 2. Why It's Necessary

Needed to onboard sBTC signers.

## 3. Execution Checklist

- [ ] sBTC Signer
- [ ] Blocklist Client
- [ ] Emily Server (optional)

PR #703, opened 2024-10-23
Add Docker Container Release Workflow

AshtonStephens:
## Description

Closes: #693 

This PR adds a docker container publishing CI action that will build release a docker container for the signer. It uses a similar workflow to the one used in `stacks-core`, but it appears that the one in stacks-core has never been run. This workflow step can only be run manually with a release tag that corresponds to a release branch. This is to avoid accidentally releasing the wrong thing.

This means to release docker images for version `a.b.c` you'll need to do the following:
1. Push to the branch `release/a.b.c`
1. Go to workflows and manually trigger this workflow with the following info:
    1.  `tag = a.b.c`
    1. `push-image = true`
    1. and then if you want the latest tag to be this version, you'd do `is-latest = true`

## Changes

- Creates a GitHub workflow to release the sbtc signer.
- Creates a `Dockerfile.signer.debian` dockerfile to build the container

Releases will be published under [`blockstack/sbtc`](https://hub.docker.com/r/blockstack/sbtc/tags).

The signer will be released under the following:
- `signer-x.x.x`
- `signer-x.x.x-debian`
- `signer-latest` (if latest)
- `signer-latest-debian` (if latest)

Only debian is currently supported but `stacks-core` currently supports alpine as well. This format leaves that open for a future improvement only if we get enough requests for it.

## Testing Information

- Image tested locally
- Deployment tested in CI with the following action: [link](https://github.com/stacks-network/sbtc/actions/runs/11503910306/job/32022240494)

## Checklist:

### DIFF
--- a/.github/actions/dockerfiles/Dockerfile.signer.debian
+++ b/.github/actions/dockerfiles/Dockerfile.signer.debian
@@ -2,7 +2,7 @@ FROM rust:1.81.0-slim-bookworm as build
 
 # Install dependencies.
 RUN apt-get update
-RUN apt-get install -y \
+RUN apt-get install -y --no-install-recommends \
     libclang-dev \
     git \
     pkg-config \
--- a/.github/actions/dockerfiles/Dockerfile.signer.debian
+++ b/.github/actions/dockerfiles/Dockerfile.signer.debian
@@ -21,4 +21,4 @@ RUN make install && make build
 # Create Docker image to run the signer.
 FROM debian:bookworm-slim AS signer
 COPY --from=build /code/sbtc/target/debug/signer /usr/local/bin/signer
-ENTRYPOINT ["/usr/local/bin/signer --config /signer-config.toml --migrate-db"]
+CMD ["/usr/local/bin/signer --config /signer-config.toml --migrate-db"]
--- a/.github/actions/dockerfiles/Dockerfile.signer.debian
+++ b/.github/actions/dockerfiles/Dockerfile.signer.debian
@@ -1,0 +1,24 @@
+FROM rust:1.81.0-slim-bookworm as build
+
+# Install dependencies.
+RUN apt-get update
+RUN apt-get install -y \
+    libclang-dev \
+    git \
+    pkg-config \
+    libssl-dev \
+    make \
+    protobuf-compiler \
+    npm \
+    default-jre
+RUN npm install -g pnpm@9
+RUN npm install -g @openapitools/openapi-generator-cli
+
+WORKDIR /code/sbtc
+COPY . .
+RUN make install && make build
+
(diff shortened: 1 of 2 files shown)


.github/workflows/image-build.yaml
```diff
@@ -0,0 +1,100 @@
...
+          - emily-server
+          - blocklist-client
+    steps:
+      ## Setup Docker for the builds
+      - name: Docker setup
+        id: docker_setup
+        uses: stacks-network/actions/docker@main
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
```
AshtonStephens, line 54: Login step.

.github/actions/dockerfiles/Dockerfile.signer.debian
```diff
@@ -0,0 +1,24 @@
...
+RUN npm install -g @openapitools/openapi-generator-cli
+
+WORKDIR /code/sbtc
+COPY . .
+RUN make install && make build
+
+# Create Docker image to run the signer.
+FROM debian:bookworm-slim AS signer
+COPY --from=build /code/sbtc/target/debug/signer /usr/local/bin/signer
+ENTRYPOINT ["/usr/local/bin/signer --config /signer-config.toml --migrate-db"]
```
wileyj, line 24: For this image, i would leave this as CMD for troubleshooting purposes. 
for a production image, entrypoint is fine though (although i personally still prefer CMD since it allows some customization)

AshtonStephens: I'm happy to move this to CMD, but for context we are only expecting these to be used for production

AshtonStephens: Looking at the differences between these I agree, `CMD` makes sense.

.github/actions/dockerfiles/Dockerfile.signer.debian
```diff
@@ -0,0 +1,24 @@
+FROM rust:1.81.0-slim-bookworm as build
+
+# Install dependencies.
+RUN apt-get update
+RUN apt-get install -y \
```
aldur, line 5: nit: `--no-install-recommends`

aldur (approved):

.github/actions/dockerfiles/Dockerfile.signer.debian
AshtonStephens: Done.

.github/actions/dockerfiles/Dockerfile.signer.debian
AshtonStephens: Done.

Issue #689, opened 2024-10-22
[Feature]: Emit print events after we successfully rotate-keys


djordon:
# Feature - Emit print events after we successfully rotate-keys

## 1. Description

We emit events after we've successfully executed almost all sBTC contract calls, the one exception is the `rotate-keys` contract call. We should emit such an event for the signers to consume. This has two benefits:
1. It allows other developers to watch for these events and use the new aggregate public key for depositing BTC into sBTC.
2. It simplifies the logic required in the signers for handling key rotation transactions.

## 2. Technical Details:

Add a print event to the end of the `rotate-keys` public function with a tuple fields for all inputs into the public function.

### 2.1 Acceptance Criteria:
   - [ ] We emit a print event at the end of the `rotate-keys` function in the registry.
   - [ ] The signers note pick up this event and store it in the database.

setzeus:
Was able to get most of this done including a test with the exception of a single field: "new-address." This is a Clarity-specific multi-sig that's calculated at run-time. Working on getting this last element fixed then should be good for review.

PR #706, opened 2024-10-24
Clarity/Key Rotation Print

setzeus:
## Description
This PR adds a print statement to the 'key-rotation' function as required in #689. 

Closes: #689 

## Changes
Single print function added.

## Testing Information
Successful key-rotation test added to check printed events against inputs. There is a caveat here, however, as we've left out the "newAddress" field in this check. We've left it out because the multi-sig stacks address is created at run-time & random, making it impossible to check against a predetermined value. 

## Checklist:

### DIFF
--- a/.generated-sources/blocklist-api/Cargo.toml
+++ b/.generated-sources/blocklist-api/Cargo.toml
@@ -10,6 +10,7 @@ edition = "2021"
 serde = { version = "^1.0", features = ["derive"] }
 serde_with = { version = "^3.8", default-features = false, features = ["base64", "std", "macros"] }
 serde_json = "^1.0"
+serde_repr = "^0.1"
 url = "^2.5"
 uuid = { version = "^1.8", features = ["serde", "v4"] }
 reqwest = { version = "^0.12", features = ["json", "multipart"] }
--- a/contracts/contracts/sbtc-registry.clar
+++ b/contracts/contracts/sbtc-registry.clar
@@ -267,7 +267,7 @@
     (print {
       topic: "key-rotation",
       new-keys: new-keys,
-      ;;new-address: new-address,
+      new-address: new-address,
       new-aggregate-pubkey: new-aggregate-pubkey,
       new-signature-threshold: new-signature-threshold
     })
--- a/contracts/contracts/sbtc-registry.clar
+++ b/contracts/contracts/sbtc-registry.clar
@@ -263,7 +263,15 @@
     ;; Update the current signature threshold
     (var-set current-signature-threshold new-signature-threshold)
     ;; Update the current aggregate pubkey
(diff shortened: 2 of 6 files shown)


.generated-sources/blocklist-api/Cargo.toml
```diff
@@ -10,6 +10,7 @@ edition = "2021"
 serde = { version = "^1.0", features = ["derive"] }
 serde_with = { version = "^3.8", default-features = false, features = ["base64", "std", "macros"] }
 serde_json = "^1.0"
+serde_repr = "^0.1"
```
djordon, line 13: Hmmm, `make install && make build` must have done this.

pnpm-lock.yaml
djordon, line 1: I'm guessing `make install && make build` did this too? I wonder why 🤔.

djordon (approved):
Looks good ✅.

I'm curious about what happened with the changes to `pnpm-lock.yaml`.

contracts/contracts/sbtc-registry.clar
```diff
@@ -263,7 +263,15 @@
...
-    (ok (var-set current-aggregate-pubkey new-aggregate-pubkey))
+    (var-set current-aggregate-pubkey new-aggregate-pubkey)
+    (print {
+      topic: "key-rotation",
+      new-keys: new-keys,
+      new-address: new-address,
+      new-aggregate-pubkey: new-aggregate-pubkey,
+      new-signature-threshold: new-signature-threshold
+    })
+    (ok true)
```
aldur, line 274: nit: I know that `var-set` currently always returns true, but to minimize unintended semantic changes can the `var-set` remain in the return?

.generated-sources/blocklist-api/Cargo.toml
aldur: Locally I have the same, I just never bothered to commit it.

contracts/contracts/sbtc-registry.clar
setzeus: Eh, would you mind if we kept it as-is? I'd prefer to keep it as (ok true) so that it's the same return message as any other function completed in the 'sbtc-registry' contract (they all end with (ok true)).

pnpm-lock.yaml
setzeus: Can confirm this was a result of 'make install' & 'make build.' I still don't understand the 'git is dirty' issue but this seemed to clear it up.

.generated-sources/blocklist-api/Cargo.toml
setzeus: Marking this as "resolved" for now so it doesn't block this PR but happy to open an issue on general 'git is dirty' / 'make install' 'make build,' I imagine it'll happen again.

Issue #515, opened 2024-09-12
[Feature]: Bitcoin deposit script validation is missing signer pub key check


evonide:
# (Medium) Bitcoin deposit script validation is missing signer pub key check

## 1. Description

The `block_observer.rs` iterates over new Bitcoin blocks and fetches new deposit requests in [sbtc/signer/src/block_observer.rs at 033e2cbd2def829a3e30df77bc274b325d68cdbe · stacks-network/sbtc](https://github.com/stacks-network/sbtc/blob/033e2cbd2def829a3e30df77bc274b325d68cdbe/signer/src/block_observer.rs#L82) from the Emily API.

These deposit requests contain information such as the Bitcoin `txid`. It then uses a `bitcoin_client` to fetch the expected tx from Bitcoin and validates it via `validate_tx` in [sbtc/sbtc/src/deposits.rs at 033e2cbd2def829a3e30df77bc274b325d68cdbe · stacks-network/sbtc](https://github.com/stacks-network/sbtc/blob/033e2cbd2def829a3e30df77bc274b325d68cdbe/sbtc/src/deposits.rs#L148-L203).

Note that while the code parses the passed `deposit_script` and parses it in [sbtc/sbtc/src/deposits.rs at 033e2cbd2def829a3e30df77bc274b325d68cdbe · stacks-network/sbtc](https://github.com/stacks-network/sbtc/blob/033e2cbd2def829a3e30df77bc274b325d68cdbe/sbtc/src/deposits.rs#L316) it currently just returns

        Ok(DepositScriptInputs {
            signers_public_key: XOnlyPublicKey::from_slice(public_key)
                .map_err(Error::InvalidXOnlyPublicKey)?,
            max_fee: u64::from_be_bytes(*max_fee_bytes),
            recipient: stacks_address,
        }
without actually verifying that the `signers_public_key` is matching the expected signers public key.

Currently, it might be possible to create a deposit script and to specify your own pub key making the deposit non-spendable by the Signers.

djordon:
I'm going to take this, I need some of these changes for https://github.com/stacks-network/sbtc/issues/286.

djordon:
lol, changed my mind. sorry.

djordon:
I took it back

PR #707, opened 2024-10-24
feat: validate the public key embedded in the deposit script

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/515.

## Changes

* Add a `can_sign` column to the `deposit_signers` table.
* Rename the `signer_public_key` column to `signers_public_key`.
* Add a `can_sign_deposit_tx` query that checks for whether the signer is in the signing set associated with the aggregate key in the deposit script.
* Do the signing set check making a decision for whether the deposit is accepted.

## Testing Information

This PR adds an integration test for the query, another for whether we correctly say that we cannot sign for the transaction if we aren't part of the signing set, and modifies a few others that implicitly do the check for when we are in the signing set.

## Checklist:


signer/src/transaction_signer.rs
```diff
@@ -639,42 +640,52 @@ where
...
             output_index: request.output_index,
             accepted: is_accepted,
+            can_sign,
         };
 
         let signer_decision = model::DepositSigner {
             txid: request.txid,
             output_index: request.output_index,
-            signer_pub_key: self.signer_pub_key(),
+            signer_pub_key: signer_public_key,
```
cylewitruk, line 686: Suggestion: Not related specifically to this line, but after reading through a number of files I'm starting to dislike `signers_public_keys` as we also have `signer_public_key` like here. I'd almost prefer that we called it `signer_set_public_key`. Verbose, but very clear... thoughts?

signer/src/transaction_signer.rs
```diff
@@ -738,6 +749,36 @@ where
...
+        // return early.
+        let Some(client) = self.blocklist_checker.as_ref() else {
+            return Ok(true);
+        };
+
+        // We turn all the input scriptPubKeys into addresses and check
+        // those with the blocklist client.
+        let bitcoin_network = bitcoin::Network::from(self.context.config().signer.network);
+        let params = bitcoin_network.params();
+        let addresses = req
```
cylewitruk, line 763: Question: This is the stuff that crashed before, right? When working towards the demo and trying to convert to an address the blocklist client would accept? Do we know if this code fixes that?

cylewitruk (commented):
Got mostly through but need to afk for a little bit, posting my comments/questions so far

signer/src/transaction_signer.rs
djordon: > This is the stuff that crashed before, right? When working towards the demo and trying to convert to an address the blocklist client would accept?

Yeah, the code would crashed here.

> Do we know if this code fixes that if so?

That bug was fixed by https://github.com/stacks-network/sbtc/pull/675. This change is just a reorganization.

signer/src/transaction_signer.rs
cylewitruk: Just this comment alone left me overly confused 😂

But yeah, naming for the different concepts which are explicit and clear is needed regardless, imo

signer/src/transaction_signer.rs
cylewitruk: Awesome, just wanted to double check👌

signer/src/transaction_signer.rs
djordon: We spoke about this yesterday. Basically, we should stick to a common name for the aggregate key in the code base. I'll open a ticket for the refactor for the name change stuff.

signer/src/storage/postgres.rs
```diff
@@ -735,6 +735,38 @@ impl super::DbRead for PgStore {
         .map_err(Error::SqlxQuery)
     }
 
+    async fn can_sign_deposit_tx(
```
cylewitruk, line 738: I'm not suggesting we change this here and now, but just wanted to note that this is an example of where I probably would have split this up into 1) getting the valid keys, 2) getting the deposit request, and 3) performing the validation in Rust.

I think that would have let you perform more of the testing as simpler unit tests.

signer/src/transaction_signer.rs
djordon: I opened https://github.com/stacks-network/sbtc/issues/729 for the name changes.

cylewitruk (approved):
Just had one comment, but otherwise I think this looks good, approving since that comment is more of a suggestion than anything.

signer/src/storage/postgres.rs
djordon: Very good point

djordon: 🙏🏽

PR #708, opened 2024-10-25
Add consensus hash to link blocks

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/559.

## Changes

Store consensus hash and anchor bitcoin block for stacks, replacing bitcoin block `confirms`. The mapping is done via `/v3/sortitions/consensus` stacks endpoint.

## Testing Information

Add a simple integration test (requires devenv) to check that we link something. I'll add better tests after fixing the broken integration tests.

## Checklist:


signer/src/block_observer.rs
```diff
@@ -267,19 +268,26 @@ where
...
 
         self.write_stacks_blocks(&stacks_blocks).await?;
         self.write_bitcoin_block(&block).await?;
 
+        // NOTE: unless we add a delay, `block` may still be unknown to the stack node,
+        // so we set the block parent consensus hash.
+        // This also applies to `get_tenure_info`: adding a ~3 sec delay in devenv
+        // returns a highest tenure.
+        self.write_bitcoin_consensus_hash(&block.header.prev_blockhash.into())
+            .await?;
```
matteojug, line 289: Do we want to add a delay (in the main loop) to give the stacks node enough time to process the bitcoin block?
Otherwise we will "lag" one bitcoin block behind, and we will probably need to change what stacks tip we fetch (as if we pass the current bitcoin tip, it will have no CH and thus we will get no stacks tip).

djordon (commented):
Awesome. Most of this looks good but I think we could simplify things with a slight change. How about we add the associated bitcoin block right there in the `stacks_blocks` table, instead of adding a `consensus_hash` column to the `bitcoin_blocks` table?

Alternatively, we could add an "bridge table" for connecting the stacks blocks to the bitcoin blocks through the consensus hash, which might be simpler. But with either approach, we don't need to really worry about adding a delay for updating our blockchain data. Thoughts?

matteojug:
> But with either approach, we don't need to really worry about adding a delay for updating our blockchain data. Thoughts?

So, I think I made some confusion about this, let me rephrase:
 - The issue I was mentioning was related to the fact that, as we process a new bitcoin block as soon as it arrives, when querying the stacks node it may still processing it:
```rust
// right after get_tenure_info in process_bitcoin_block; picking the last of those logs
dbg!(&block, &tenure_info, self.stacks_client.get_sortition_info(&block.header.prev_blockhash.into()).await);

// The new block is 0f41, with parent 62d3
[signer/src/block_observer.rs:273:9] &block = Block {
    header: Header {
        block_hash: 0f41b47d883e3f3aa4357b82126a003bcd09375a5c166423148d6bd92b54adab,
        prev_blockhash: 62d3b6bd9faf0d28e3fa3e4d969bdb48a55d5ed276a2e6b92aa33842d13455a5,
        ...
    }
}

// The "current" tenure returned has as CH 602c
[signer/src/block_observer.rs:273:9] &tenure_info = RPCGetTenureInfo {
    consensus_hash: 602cd8c97a9dbf18f5d2723ce5959494c2f39afd,
    parent_consensus_hash: c3463d1a986c59e34356faa27a8b3275cf6b3358,
    ...
}

// But that CH (602c) is actually the one of the parent (62d3) of the current block
[signer/src/block_observer.rs:273:9] self.stacks_client.get_sortition_info(&block.header.prev_blockhash.into()).await = Ok(
    SortitionInfo {
        burn_block_hash: 62d3b6bd9faf0d28e3fa3e4d969bdb48a55d5ed276a2e6b92aa33842d13455a5,
        burn_block_height: 246,
        sortition_id: 802001de2a32c526f1e20d2f99ea5e85c389ae8fe870e496d04de67bac1a7a06,
        parent_sortition_id: 5d6f5f4dfadea2f9051a3d1a7fda8cfb78684501496ba27b167bf540a491bf3f,
        consensus_hash: 602cd8c97a9dbf18f5d2723ce5959494c2f39afd,
        stacks_parent_ch: Some(
            c3463d1a986c59e34356faa27a8b3275cf6b3358,
        ),
        last_sortition_ch: Some(
            c3463d1a986c59e34356faa27a8b3275cf6b3358,
        ),
        committed_block_hash: Some(
            245d91f1f1fa5945e03615e4b37189fd320d8beadf9dd43909fd7ad435c4d0b3,
        ),
    },
)
```
 - If we add a 3s delay, `get_tenure_info` returns the tenure related to the new block (0f41)
 
While I said this was an issue, it's probably fine instead if we consider that the new block (0f41) actually confirms the stacks blocks of the previous bitcoin block (62d3) CH. I'm inclined to think this is correct because of the completely scientific evidence that looking at the stacks explorer when running the above, when we get notified by the new bitcoin block it only shows one stacks block in that bitcoin block, and then it starts adding others during the next 30s.

If the above sounds right, then there's no issue with delays: when a new bitcoin block arrives, it confirms its parent CH (and all the stacks blocks with the same CH -- that we fetch in that same iteration via `get_tenure_info`).

> How about we add the associated bitcoin block right there in the `stacks_blocks` table, instead of adding a `consensus_hash` column to the `bitcoin_blocks` table?

Assuming the child confirms its parent above, and that we want to keep using the bitcoin chain tip as anchor when fetching things, this may get tricky if we have bitcoin forks -- eg, if we add a "confirmed_by_bitcoin_block" column in the stacks blocks table, then both forks may write on that. While if we keep the data on the bitcoin blocks table (eg, via a "confirm_CH" replacing the current "consensus_hash" in this PR) then such forks would not race to write.

> Alternatively, we could add an "bridge table" for connecting the stacks blocks to the bitcoin blocks through the consensus hash, which might be simpler.

Yeah this may be the best way to avoid cluttering the bitcoin block table with _not so related_ info, at the cost of one extra join for some queries. But I would still keep the CH on the stacks block side in the stacks blocks table, as it does belong to the stacks block itself; unless we want to keep a stacks block representation slim for some reason.
Therefore, I would go with a new table storing the bitcoin block hash and "confirmed" CH (ie, the one of its parent).
WDYT?

djordon:
Ahhh thanks for the detailed post, I think I see what your saying. I have a detailed response below, but the summary is that I don't think we need to worry about forks if we write the anchor bitcoin block that each stacks block is building on. And in my suggestions above, I was thinking that we transition to an anchored by approach. Why didn't you read my mind?

Here are the more detailed responses.

> Assuming the child confirms its parent above, and that we want to keep using the bitcoin chain tip as anchor when fetching things, this may get tricky if we have bitcoin forks -- eg, if we add a "confirmed_by_bitcoin_block" column in the stacks blocks table, then both forks may write on that.

Ahh, I wasn't thinking that we go with the "confirmed-by" approach, but with the "anchored-by" approach. In the "anchored-by" approach we note which bitcoin block the stacks block is building on. This approach means we never update anything in our tables when anything forks and it aligns with how stacks describes things. 

> But I would still keep the CH on the stacks block side in the stacks blocks table, as it does belong to the stacks block itself; unless we want to keep a stacks block representation slim for some reason.

Yeah, in either approach I would include the consensus hash in the `stacks_blocks` table.

> Therefore, I would go with a new table storing the bitcoin block hash and "confirmed" CH (ie, the one of its parent).
WDYT?

Even with the bridge table, I think we should go with an "anchored-by" approach.

diwakergupta:
> I second an intermediate table. It's really the only way to guarantee that we can handle forks from both ends, i.e. many-to-many (think worst-case, Bitcoin and Stacks forking simultaneously).

Just to be clear, post Nakamoto, Stacks will not fork independent of Bitcoin. In case it helps simplify some assumptions you're making.

cylewitruk:
> > I second an intermediate table. It's really the only way to guarantee that we can handle forks from both ends, i.e. many-to-many (think worst-case, Bitcoin and Stacks forking simultaneously).
> 
> Just to be clear, post Nakamoto, Stacks will not fork independent of Bitcoin. In case it helps simplify some assumptions you're making.

@diwakergupta hm, but the docs mention that 70% of voting power can approve a fork, or I learned yesterday that the last block in a tenure can fork (but this was possibly old info)?

signer/src/stacks/api.rs
```diff
@@ -663,6 +670,44 @@ impl StacksClient {
...
+            .map_err(Error::StacksNodeResponse)?
+            .json::<Vec<SortitionInfo>>()
+            .await
+            .map_err(Error::UnexpectedStacksResponse)
+            .and_then(|result| {
+                result
+                    .into_iter()
+                    .next()
+                    .ok_or(Error::InvalidStacksResponse("missing sortition info"))
+            })
```
djordon, line 708: Ahhh, so the stacks-core response is a list with one element in our case. If we asked for the "latest-and-last" then we will receive a list with two elements. Well, at least that's the behavior [on main](https://github.com/stacks-network/stacks-core/blob/40059a57cd27e740c5e9d91a833fb2c975b0bf0b/stackslib/src/net/api/getsortition.rs#L333-L369). I'd add a comment with a link to the source somewhere around here stating that the vector should have exactly one element since we supply the consensus hash.

djordon (commented):
Looks good! Curious, what else is needed here?

matteojug:
> Curious, what else is needed here?

I have two failing tests, fixing them and then it should be ready for review.

signer/src/transaction_signer.rs
```diff
@@ -903,7 +903,7 @@ mod tests {
 
         testing::transaction_signer::TestEnvironment {
             context,
-            context_window: 3,
+            context_window: 6,
```
matteojug, line 906: This is a magic number: higher we may fail the `num_expected_decisions` checks in `assert_should_store_decisions_received_from_other_signers` (because of context larger than chain length, so we expect more decision that we can get), lower we may get no withdrawal requests in the intersection between the stacks canonical chain and the context window.

matteojug (commented):
The changes in some tests are required since we originally were doing the context window check wrong for stacks chain (in some places): we would get the bitcoin block at the end of the context window, and then iterate the stack chain as long as we didn't encounter that end block; but since the stack chain itself could be not "linearized" in the blocks, walking it may skip the end block. While this may not happen in prod, the test data we generate may create such scenarios.

signer/migrations/0003__create_tables.sql
```diff
@@ -13,14 +13,15 @@ CREATE TABLE sbtc_signer.bitcoin_blocks (
...
     parent_hash BYTEA NOT NULL,
-    confirms BYTEA[] NOT NULL,
     created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL
 );
 
 CREATE TABLE sbtc_signer.stacks_blocks (
     block_hash BYTEA PRIMARY KEY,
     block_height BIGINT NOT NULL,
     parent_hash BYTEA NOT NULL,
+    consensus_hash BYTEA NOT NULL,
```
matteojug, line 23: In the current implementation we don't actually need this stored -- it's enough to have it when building the stacks block model. Should we remove it?

signer/src/transaction_signer.rs
djordon: Ahhh I think I understand.

signer/migrations/0003__create_tables.sql
djordon: I'm inclined to remove it. If it turns out we actually need it down the line we can easily resurrect it. But I do not feel strongly one way or the other.

djordon (approved):
Looks good ✅.

signer/src/transaction_signer.rs
djordon: Well, we have a known frailty with the generation of data for some of these tests. We can choose to fix them in some other PR.

signer/migrations/0003__create_tables.sql
matteojug: Remove in [6802fa2](https://github.com/stacks-network/sbtc/pull/708/commits/6802fa27d2b2734a1b764930a380a64562dd6d4e)

djordon (approved):

Issue #585, opened 2024-09-30
[Feature]: Store the requests tied to signer bitcoin transactions


djordon:
# Feature - Store the requests tied to signer bitcoin transactions

## 1. Description

After a bitcoin sweep transaction has been confirmed, the signers construct and send stacks transactions related to deposit or withdrawal requests that were serviced by the confirmed transaction. In order to do that, the signers need to figure out which requests were tied to the bitcoin transaction. For deposits, this is relatively easy since an outpoint in the prevouts will point to the deposit request, but for withdrawals it's much more difficult, even after https://github.com/stacks-network/sbtc/issues/483.

### 1.1 Context & Purpose

The signers should store enough data to explicitly tie withdrawal and deposit requests to the bitcoin transactions that they broadcast, so that they can easily figure out which stacks transactions to make.

For outputs, that would mean the request ID, stacks transaction ID, and stacks block ID, associated with the withdrawal for each output index. We do not need to store anything for the inputs though.

## 2. Technical Details:

Add a new table, or repurpose the `coordinator_broadcasts` table, to hold information about requests that are serviced by a bitcoin transaction.

### 2.1 Acceptance Criteria:
   - [ ] We store the request data after successfully broadcasting a bitcoin transaction.

## 3. Related Issues and Pull Requests (optional):

This is part of https://github.com/stacks-network/sbtc/issues/563 and #320.

Issue #709, opened 2024-10-25
[Feature]: Inform peers of newly broadcasted sweep packages/transactions (followup to #585)


cylewitruk:
# Feature - Inform peers of newly broadcasted sweep packages/transactions (followup to #585)

## 1. Description

#585 adds functionality for storing broadcast sweep transaction packages and related tx/deposit/withdrawal information, but the Bitcoin transaction itself which is broadcast doesn't contain the same (mapping) information and so the signer peers won't have it without needing to do extra work. This is an issue when the next coordinator takes over at the next block and may need to construct Stacks transactions or RBF outstanding sweep transactions.

## 2. Technical Details:

The signer coordinator needs to broadcast information about new sweep transactions to its signer peers so that they have the information needed to RBF the last transaction package, if needed.

Possible gotchas:
- The signers receiving these notifications also need to have the deposit and withdrawal requests locally before processing these messages. Can we assume that they do?

### 2.1 Acceptance Criteria:
   - [ ] The transaction coordinator broadcasts information about new Bitcoin sweep transactions
   - [ ] The receiving signers store this information locally.

## 3. Related Issues and Pull Requests (optional):
- #585 
Related discussion from 585's PR: https://github.com/stacks-network/sbtc/pull/710#discussion_r1817081648

Issue #541, opened 2024-09-17
[Feature]: Get last fee for bitcoin transactions


djordon:
# Feature - Get Last Fee for bitcoin transactions

## 1. Description

Before we can replace-by-fee (RBF) a bitcoin transaction, we need to fetch the fee that was paid for the last transaction *package* that is currently in the mempool.

### 1.1 Context & Purpose

In order to RBF a transaction you need to know how much you paid both in total and per vbyte in order to know whether the new fee would be accepted. Since we could be replacing an entire package, we need to know the fees paid for all signer transactions in the mempool. Each of these transactions has their first input and output with a `scriptPubKey` that is in the `dkg_shares` table.

## 2. Technical Details:

Might be easiest to use the [`gettxspendingprevout`](https://bitcoincore.org/en/doc/27.0.0/rpc/blockchain/gettxspendingprevout/) and [`getmempooldescendants`](https://bitcoincore.org/en/doc/27.0.0/rpc/blockchain/getmempooldescendants/) RPC calls on bitcoin-core since we know the previous signer UTXO once https://github.com/stacks-network/sbtc/issues/538 is done.

### 2.1 Acceptance Criteria:

- [ ]  The `BitcoinInteract::get_last_fee` returns the fee paid for the current outstanding transaction package in the mempool.

## 3. Related Issues and Pull Requests (optional):

djordon:
Yes.

cylewitruk:
Replaced by #809

Issue #483, opened 2024-09-03
[Feature]: Record merkle root of withdrawal identifiers on chain


djordon:
# Feature - Record merkle root of withdrawal identifiers on the bitcoin blockchain

## 1. Description

For withdrawals, we do not have an airtight way of tying the withdrawal UTXO to the stacks transaction that initiated the request. When a signer tries to create a `withdrawal-accept` transaction it needs a knowing which UTXOs are associated with which `initiate-withdrawal-request` transactions. The original design for the withdrawal flow is in https://github.com/stacks-network/sbtc/issues/16.

There are two approaches that “work” to varying degrees without further changes to our design:

1. Have the signer who submitted the BTC transaction for the withdrawal create and submit the `withdrawal-accept` stacks transaction that includes this information.
2. Use the recipient, the amount, and maybe time to make an educated guess of the `initiate-withdrawal-request` transaction that is associated with a UTXO.

For (1), we cannot guarantee that signer who created the BTC transaction also creates the `withdrawal-accept` stacks transaction; and for (2) it would be better to have a way where we can “definitively” know which stacks transaction was associated with a UTXO. For deposits, we have a simple way of tying sweep transactions to deposits: the UTXO input in the sweep transaction always points to the deposit transaction.

### 1.1 Context & Purpose

The proposal here is three fold:

1. Change ordering of the UTXOs from the design in https://github.com/stacks-network/sbtc/issues/19: The first UTXO is the signers UTXO containing all deposits, the second UTXO will be the `OP_RETURN` UTXO, and the remaining UTXOs are for withdrawals. This is what is currently implemented in the code.
2. Change the OP byte in the `OP_RETURN` UTXO to a version byte. This version byte would allow us to change the layout of the data in the future by incrementing the version.
3. Include the root of a merkle tree at the end of the `OP_RETURN` UTXO described in https://github.com/stacks-network/sbtc/issues/19. 

The nodes of the merkle tree would be associated with the transaction’s UTXOs. Each node of the tree would be a hash160 of the withdrawal request’s identifier. This identifier is a concatenation of the stacks transaction ID that generated the withdrawal request, the request-id of the withdrawal request, and the stacks block ID that includes the transaction. We need the block Id since a transaction can be included in more than one block and we need the request-id because each transaction can generate more than one request. So the pre-image data for each node of the tree has the following layout:

```
0                 32                             64           72
|-----------------|------------------------------|------------|
[...]

PR #710, opened 2024-10-25
Feature: Store sweep transaction information with associated deposit/withdrawal requests

cylewitruk:
## Description

Closes: #585 
Sets the stage for: #541 (I hope)
Creates: #709

Got a little bigger than I expected so I stashed what I've started for broadcasting the info to peers and moved it to a new ticket #709. 

I could also have moved some of the stuff I did for `get_swept_deposit_requests()` to a separate PR, but I liked having it here for verifying that the data model meets the requirements of that function.

A lot of the line count is comments, though ;)

## Changes

**Directly related changes**
- Adds four new tables for storing sweep transaction package information:
  - `sweep_packages`: Represents a container for an entire transaction package. Not really necessary atm, but offers some future-proofing in-case we ever want to tighten the loops and (re)submit multiple transaction packages within the same block (like if we monitor the mempool and start thinking the current package might not make it into the next block with the current fees and want to RBF).
  - `sweep_transactions`: Represents an individual sweep transaction within a package.
  - `swept_withdrawals`: Links withdrawal requests to the sweep transaction.
  - `swept_deposits`: Links deposit requests to the sweep transaction.
- Adds `DbWrite` functionality for writing an entire transaction package to the db and marking an individual transaction as successfully broadcast.
- Adds `DbRead` functionality for retrieving the latest (newest) transaction package for the given chain tip (hoping this helps with #541, but it's mostly used for verifying that what gets written is correct).
- `TransactionCoordinator` is updated to store all sweep transaction packages and mark each transaction as broadcast after successfully broadcast to Bitcoin.

**Peripheral changes**

- `cargo fmt` snuck a little unrelated formatting in here in the Emily code.
- Removes the unused `coordinator_broadcasts` table.
- Adds a new table-valued function (TVF) `bitcoin_blockchain_of(chain_tip, max_depth)` to help reduce the recursive query boilerplate in `postgres.rs`.
- `get_swept_deposit_requests()` is refactored to use the new data model. It can't be completed until #559 is done, but it allowed for activating the previously-ignored `get_swept_deposit_requests_does_not_return_unswept_deposit_requests` test.
- Boyscout-added a few db indexes which I noticed were missing.
- Added a `SET search_path TO sbtc_schema,public` to the pgsql `after_connect` hook so that we can stop namespacing everything in `postgres.rs` (makes it easier to read).

## Testing Information

Updated the `TestSweepSetup` to support storing transaction packages.

Updated integration tests:
- process_complete_deposit
- get_swept_deposit_requests_returns_swept_deposit_requests
- get_swept_deposit_requests_does_not_return_unswept_deposit_requests (and now un-ignored)

New integration tests:
- can_store_and_get_latest_sweep_transaction_package

## Checklist:


signer/src/storage/model.rs
```diff
@@ -257,8 +418,6 @@ pub struct SweptDepositRequest {
     /// The transaction ID of the bitcoin transaction that swept in the
     /// funds into the signers' UTXO.
     pub sweep_txid: BitcoinTxId,
-    /// The transaction sweeping in the deposit request UTXO.
-    pub sweep_tx: BitcoinTx,
```
cylewitruk, line 261: I couldn't see that this needed to be used anywhere so I removed it to avoid the extra I/O. But if it's needed later for validation etc. then we can easily add it back.

signer/src/storage/model.rs
```diff
@@ -297,8 +456,6 @@ pub struct SweptWithdrawalRequest {
     /// The transaction ID of the bitcoin transaction that swept out the
     /// funds to the intended recipient.
     pub sweep_txid: BitcoinTxId,
-    /// The bitcoin transaction fulfilling the withdrawal request.
-    pub sweep_tx: BitcoinTx,
```
cylewitruk, line 301: I couldn't see that this needed to be used anywhere so I removed it to avoid the extra I/O. But if it's needed later for validation etc. then we can easily add it back.

signer/src/storage/postgres.rs
```diff
@@ -111,7 +114,13 @@ impl TryFrom<&NakamotoBlock> for model::StacksBlock {
 impl PgStore {
     /// Connect to the Postgres database at `url`.
     pub async fn connect(url: &str) -> Result<Self, Error> {
-        let pool = sqlx::PgPool::connect(url)
+        let pool = PgPoolOptions::new()
```
cylewitruk, line 117: Doing this lets us remove schema qualifiers from the queries in this file, which makes them easier to read.

signer/src/storage/model.rs
```diff
@@ -10,11 +10,172 @@ use serde::Deserialize;
...
+}
+
+/// Represents a single deposit which has been swept-in by a sweep transaction
+/// package.
+#[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord, sqlx::FromRow)]
+#[cfg_attr(feature = "testing", derive(fake::Dummy))]
+pub struct SweptDeposit {
+    /// The index of the deposit output in the sBTC sweep transaction.
+    #[sqlx(try_from = "i32")]
+    pub output_index: u32,
```
djordon, line 153: I don't think that we need to add this. It's slower but if we need to know this we should just scan the inputs by the outpoint (the transaction ID and the output index).

signer/migrations/0003__create_tables.sql
```diff
@@ -184,3 +191,111 @@ CREATE TABLE sbtc_signer.withdrawal_reject_events (
...
 );
 
+-- Represents a combined transaction package which is broadcasted to the Bitcoin
+-- network. The transaction package is built up of multiple transactions which
+-- are tracked separately in the `sweep_transactions` table. A transaction
+-- package may contain the sweeping transactions for both deposit and
+-- withdrawal requests.
+CREATE TABLE sbtc_signer.sweep_packages (
+    -- Internal ID of the package
+    id SERIAL PRIMARY KEY,
```
djordon, line 201: I don't think we should note things at the package level in our database. Although we construct and broadcast transactions as a package, they are not guaranteed to all be included or excluded for any given block. A bitcoin miner could include the first, say, 3 transactions from our "package" but not the following ones. For this reason tying them together might not be so helpful.

signer/migrations/0003__create_tables.sql
```diff
@@ -184,3 +191,111 @@ CREATE TABLE sbtc_signer.withdrawal_reject_events (
...
+    txid BYTEA NOT NULL,
+    -- The signer UTXO being spent in this transaction.
+    utxo_txid BYTEA NOT NULL,
+    utxo_output_index INTEGER NOT NULL,
+    -- The total amount of the transaction.
+    amount BIGINT NOT NULL,
+    -- The fee paid for the transaction.
+    fee BIGINT NOT NULL,
+    -- The fee rate in satoshis per vByte used for this transaction.
+    fee_rate DOUBLE PRECISION NOT NULL,
```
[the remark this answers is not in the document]
djordon: But we could note the market fee rate. That information is good for diagnostic purposes.

signer/migrations/0003__create_tables.sql
```diff
@@ -184,3 +191,111 @@ CREATE TABLE sbtc_signer.withdrawal_reject_events (
...
+        REFERENCES sbtc_signer.sweep_transactions(id),
+
+    FOREIGN KEY (deposit_request_txid, deposit_request_output_index) 
+        REFERENCES sbtc_signer.deposit_requests(txid, output_index)
+);
+-- Our main index which will cover searches by 'deposit_request_txid' and
+-- 'deposit_request_output_index', while also restricting the combination to be
+-- unique per 'sweep_transaction_id'.
+CREATE UNIQUE INDEX uix_swept_deposits_req_txid_req_output_index_pkgd_txid
+    ON sbtc_signer.swept_deposits(deposit_request_txid, deposit_request_output_index, sweep_transaction_id);
```
djordon, line 297: Hmmm, although I like all of my tables to have an `id` column, I can also see the case for removing them when we have a natural composite primary. In this case, it seems like we have a natural primary key using the columns in this index.

signer/migrations/0003__create_tables.sql
```diff
@@ -184,3 +191,111 @@ CREATE TABLE sbtc_signer.withdrawal_reject_events (
...
+-- package. Individual deposit and withdrawal requests reference back to these
+-- transactions to keep track of both the overall transaction package as well as
+-- the individual Bitcoin transactions they are related to.
+CREATE TABLE sbtc_signer.sweep_transactions (
+    -- Internal ID of the transaction.
+    id BIGSERIAL PRIMARY KEY,
+    -- The ID of the grouping transaction package.
+    sweep_package_id INTEGER NOT NULL,
+    -- The Bitcoin transaction ID of the transaction.
+    txid BYTEA NOT NULL,
```
djordon, line 220: Shouldn't this be the primary key?

signer/migrations/0003__create_tables.sql
```diff
@@ -184,3 +191,111 @@ CREATE TABLE sbtc_signer.withdrawal_reject_events (
...
+-- Represents a single withdrawal request which has been included in a sweep
+-- transaction package. Withdrawal requests have a unique ID so we use that here
+-- to reference the withdrawal request together with its Stacks block hash,
+-- which can be retrieved from the `withdrawal_requests` table.
+CREATE TABLE sbtc_signer.swept_withdrawals (
+    -- Internal ID of the swept withdrawal.
+    id BIGSERIAL PRIMARY KEY,
+    -- References the `packaged_transaction` in which this withdrawal was
+    -- included.
+    sweep_transaction_id INTEGER NOT NULL,
```
djordon, line 247: Why not use the transaction ID here?

signer/migrations/0003__create_tables.sql
```diff
@@ -184,3 +191,111 @@ CREATE TABLE sbtc_signer.withdrawal_reject_events (
...
+    fee BIGINT NOT NULL,
+    -- The fee rate in satoshis per vByte used for this transaction.
+    fee_rate DOUBLE PRECISION NOT NULL,
+    -- The timestamp that the transaction was broadcast at. This should be
+    -- set after we know that the transaction was successfully broadcast.
+    is_broadcast BOOLEAN NOT NULL DEFAULT FALSE,
+
+    FOREIGN KEY (sweep_package_id)
+        REFERENCES sbtc_signer.sweep_packages(id)
+);
```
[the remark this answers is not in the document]
djordon: I'm still trying to think through whether we absolutely need to do https://github.com/stacks-network/sbtc/issues/709. If we don't then the `updated_at` and `is_broadcast` columns would be unnecessary.

signer/migrations/0004__create_functions.sql
```diff
@@ -0,0 +1,50 @@
+-- TVF for fetching a Bitcoin blockchain from a given block hash, only looking
```
djordon, line 1: Can we write out TVF as table-valued function. It's not obvious what this means (I had to look it up).

signer/migrations/0004__create_functions.sql
```diff
@@ -0,0 +1,50 @@
...
+-- back the `max_depth` number of blocks. The most common use case is to fetch
+-- the blockchain from the canonical chain tip.
+--
+-- - chain_tip: The block hash to start the blockchain from ("chain tip").
+-- - max_depth: The maximum depth of the blockchain to fetch.
+CREATE FUNCTION sbtc_signer.bitcoin_blockchain_of (
+    chain_tip BYTEA,
+    max_depth INT
+)
```
djordon, line 10: I really like this function.

djordon (commented):
I didn't finish the review yet, just thought that I'd pause and give you my thoughts so far.

signer/migrations/0003__create_tables.sql
```diff
@@ -16,13 +16,18 @@ CREATE TABLE sbtc_signer.bitcoin_blocks (
     confirms BYTEA[] NOT NULL,
     created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL
 );
+-- Index to serve queries filtering on `parent_hash`. This is commonly used when
+-- "walking" the chain in recursive CTE's.
+CREATE INDEX ix_bitcoin_blocks_parent_hash ON sbtc_signer.bitcoin_blocks(parent_hash);
```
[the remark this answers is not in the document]
cylewitruk: So in general, unless you're dealing with huge amounts of data and a high volume of inserts, low-cardinality columns or very small tables, an index will never hurt. I didn't add an index anywhere where I didn't think those criteria were met (or would be within a reasonable timespan in prod).

signer/migrations/0003__create_tables.sql
```diff
@@ -184,3 +191,111 @@ CREATE TABLE sbtc_signer.withdrawal_reject_events (
...
+    utxo_output_index INTEGER NOT NULL,
+    -- The total amount of the transaction.
+    amount BIGINT NOT NULL,
+    -- The fee paid for the transaction.
+    fee BIGINT NOT NULL,
+    -- The fee rate in satoshis per vByte used for this transaction.
+    fee_rate DOUBLE PRECISION NOT NULL,
+    -- The timestamp that the transaction was broadcast at. This should be
+    -- set after we know that the transaction was successfully broadcast.
+    is_broadcast BOOLEAN NOT NULL DEFAULT FALSE,
```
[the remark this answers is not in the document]
cylewitruk: Why not? The default for this table imo is not-broadcast-until-explicitly-marked-broadcast? We only have one code-path for inserts here atm.
P.S. realized I missed to update the comment :laughing:

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
cylewitruk: I was originally down a similar path but felt it cost more in code than it was worth. This is provided as an `f64` from `estimate_fees()` (or similar) so I think there's greater risk in losing "data" in the otherwise conversion.

The market fee rate is on the package-level (covering all transactions).

signer/migrations/0004__create_functions.sql
cylewitruk: Yeah I can update that

signer/migrations/0003__create_tables.sql
cylewitruk: It should definitely be a unique index if nothing else, good catch. I'm not a huge fan of unnecessarily using large columns as PK's in growing tables which have a lot of FK's (like in this case with swept_withdrawals and swept_deposits, 8 vs 32 bytes for each reference, and really this doesn't need to be `bigserial` either) -- unless the entire convention of the db is to use i.e. uuid surrogate keys.

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
djordon: I'm not suggesting that we use the fee rate to compute the transaction's size, but get and store the size from the actual transaction.

signer/migrations/0003__create_tables.sql
cylewitruk: I generally avoid natural keys, even composite, because they have almost always bitten me in the a** when requirements change down the line. A surrogate key I can **guarantee** will always uniquely identify the record, even if the natural key were to change (i.e. gain or lose a column), and also _allows_ it.

I've got years of experience of natural keys biting me in the a**, but blockchain is a new domain for me and I'll admit that, while my defaults kick in to not trust them, it may not be the right choice in this domain. Let's huddle it (with others who maybe want to discuss) when we get the chance?

signer/migrations/0003__create_tables.sql
cylewitruk: I think this is part of both the preceeding and succeeding comments - let's discuss further there :)

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
cylewitruk: Yeah we could do that, it's not a problem, but what I meant is that this is the literal value that comes from the `SignerBtcState` associated with the transaction itself.

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
cylewitruk: I have a hard time envisioning how we should keep nodes in sync without informing each other out-of-band about what's happening. It's possible to do it purely based on the mempool of course, but that in itself also requires a bit of work and is much less flexible.

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
djordon: Oh I see. Let me start by acknowledging that my original reason for why I thought that we shouldn't store the fee rate was not compelling. I have a different suggestion, and it allows us to leave this column out if we want:

* we should store the `vsize` for each transaction
* we should store the `vsize` for each input and output linked to a request in the transaction.

This allows us to compute the assessed fee for all inputs and outputs. It also allows us to compute the fee rate for the entire transaction if we want to.

---

About using the `SignerBtcState` struct, that only has the market fee rate. I don't think any of our structs contain the fee rate for the bitcoin transaction.

Unrelated, just wanted to call out that while each transaction targets the market rate, the actual fee rate for any transaction can be quite a bit more than the market rate if we are RBFing a transaction.

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
djordon: Yeah I think we can do it. Here are my thoughts

---

One of the purposes of these tables to easily identify which requests have been serviced by a bitcoin transaction that has been confirmed. We can accomplish that if:
1. each signer stores all confirmed transactions where one of the outputs is a signer UTXO, and
2. each signer notes the requests tied to each bitcoin transaction that may have been broadcast by another signer.

We already do (1) in the block observer. If we added (2) then we know which requests have been serviced on chain by just doing a join on the "confirmed transactions" table with the "maybe broadcasted" table; we don't need an additional filter for `is_broadcast`.

The other purpose of these tables is to easily identify the "signers' UTXO". We have a way now but it is complex. I think we can accomplish the same task by noting the `scriptPubKey` for the signers' UTXO along with all `prevouts` in the "maybe broadcast" transaction row.

---

It's very possible that my thinking misfired here, or that I'm missing a scenario where we need to know what was actually broadcast. What are your thoughts?

signer/migrations/0003__create_tables.sql
djordon: Yeah let's do that. I'll set something up for Monday.

signer/migrations/0003__create_tables.sql
```diff
@@ -184,3 +191,111 @@ CREATE TABLE sbtc_signer.withdrawal_reject_events (
...
+CREATE TABLE sbtc_signer.sweep_transactions (
+    -- Internal ID of the transaction.
+    id BIGSERIAL PRIMARY KEY,
+    -- The ID of the grouping transaction package.
+    sweep_package_id INTEGER NOT NULL,
+    -- The Bitcoin transaction ID of the transaction.
+    txid BYTEA NOT NULL,
+    -- The signer UTXO being spent in this transaction.
+    utxo_txid BYTEA NOT NULL,
+    utxo_output_index INTEGER NOT NULL,
```
djordon, line 223: Hmmm, I think we want `scriptPubKey` and `amount` columns here as well.

djordon: nit: we might want to rename this to `signer_prevout_txid` and `signer_prevout_output_index` or something similar. The `utxo_` prefix is ambiguous on it's own, you need to know that you are referring to the signers' UTXO.

signer/migrations/0003__create_tables.sql
```diff
@@ -128,12 +133,14 @@ CREATE TABLE sbtc_signer.rotate_keys_transactions (
...
 
+-- TODO: Unused?
 CREATE TABLE sbtc_signer.deposit_responses (
     response_txid BYTEA NOT NULL,
     deposit_txid BYTEA NOT NULL,
     deposit_output_index INTEGER NOT NULL
 );
 
+-- TODO: Unused?
 CREATE TABLE sbtc_signer.withdrawal_responses (
```
djordon, line 144: We've basically renamed these `swept_deposits` and `swept_withdrawals`, so we can remove them.

signer/migrations/0003__create_tables.sql
djordon: Oh I just misread this. Yeah let's discuss further there.

signer/migrations/0003__create_tables.sql
djordon: Yeah let's do that. I'll set something up for Monday.

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
cylewitruk: Updated in [7f2d0fb](https://github.com/stacks-network/sbtc/pull/710/commits/7f2d0fb3dd86364a98e95bb277d903ebcb3c5f06)

signer/migrations/0003__create_tables.sql
cylewitruk: Alright, gotcha, then I agree that this isn't needed from a functional perspective.

I'm still in favor of having this from a normalization perspective, though. The transactions in the package are all created at the same block hash, time and using the same market rate, which are three columns which then don't need to be stored separately on each transaction.

Let's discuss this one further on the call.

signer/src/storage/model.rs
```diff
@@ -10,11 +10,172 @@ use serde::Deserialize;
...
+    #[cfg_attr(feature = "testing", dummy(default))]
+    pub fee_rate: f64,
+    /// Whether or not this transaction has been successfully broadcast to the
+    /// Bitcoin network.
+    #[cfg_attr(feature = "testing", dummy(default))]
+    pub is_broadcast: bool,
+
+    /// List of deposits which were swept-in by this transaction.
+    #[sqlx(skip)]
+    pub swept_deposits: Vec<SweptDeposit>,
```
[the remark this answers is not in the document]
cylewitruk: Updated in [81933a7](https://github.com/stacks-network/sbtc/pull/710/commits/81933a7404312197b4e1d4795e498a7d81925929).

We lost the `Dummy` derive due to the `bitcoin::OutPoint`, but I'm not using that right now anyway.

signer/src/storage/model.rs
cylewitruk: Updated in [81933a7](https://github.com/stacks-network/sbtc/pull/710/commits/81933a7404312197b4e1d4795e498a7d81925929)

signer/migrations/0003__create_tables.sql
cylewitruk: Added a unique index for `txid` in the meantime, pending the discussion on monday. [2ee116d](https://github.com/stacks-network/sbtc/pull/710/commits/2ee116d31948a4e5f2f5e95512b468aa5ecf8cb2)

signer/migrations/0003__create_tables.sql
cylewitruk: Renamed and added `signer_prevout_amount` and `signer_prevout_script_pubkey` in [2ee116d](https://github.com/stacks-network/sbtc/pull/710/commits/2ee116d31948a4e5f2f5e95512b468aa5ecf8cb2). I'm not sure I'm putting the actual expected value in the script pubkey column, is this correct?
```rust
transaction
  .signer_utxo
  .utxo
  .public_key
  .signers_script_pubkey()
  .into(),
```

signer/migrations/0004__create_functions.sql
cylewitruk: Updated in [e7b7b8f](https://github.com/stacks-network/sbtc/pull/710/commits/e7b7b8f6464797271ddbed772b81354247214f08)

signer/migrations/0003__create_tables.sql
cylewitruk: Yeah I didn't even notice these until later :sweat_smile: removed in [9340bab](https://github.com/stacks-network/sbtc/pull/710/commits/9340bab1fb8245358c026b5dd3d1e68f91258a20)

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
cylewitruk: Do we really want to store the `vsize` of every input & output? Isn't it enough to have it on the transaction? I'm thinking that if we know our sats/vbyte and the previn fee then we should have everything we need to ensure that both the fee and rate are high enough in an eventual RBF.. but maybe I'm thinking wrong?

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
cylewitruk: I guess what I'm getting hung-up on is that my understanding of #585 is that the information described there is basically needed for withdrawals (and deposits are just nice-to-have) for the construction of the follow-up Stacks transactions. Since the coordinator broadcasting the sweeps (and storing data per #585) only has a small chance of also being the coordinator at the time these Stacks transactions will be constructed (we need to see the sweep confirmed first, right?), I don't see how #585 helps anything if we can't ensure that the Stack-transaction-creating signer doesn't also have the same information.

But if I understand you correctly here you're saying that #585 is basically irrelevant now if we start tying "potentially serviced" requests that we know about at each given Bitcoin block in the block observer, and that we do the `for withdrawals it's much more difficult, even after #483` instead? Or maybe I'm reading this wrong?

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
djordon: > Do we really want to store the `vsize` of every input & output? Isn't it enough to have it on the transaction?

Oh, you're right, we can compute what is necessary when we grab the raw transaction.


> I'm thinking that if we calculate our sats/vbyte on the tx-level from `vsize` and `fee` then we should have everything we need to ensure that both the fee and rate are high enough in an eventual RBF.. but maybe I'm thinking wrong?

I was thinking about using this information for validating the stacks `complete-deposit` contract call. There we ensure that the assessed fee is under the request's max-fee, and for that we use the `vsize` of all request inputs and outputs.

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
djordon: > I guess what I'm getting hung-up on is that my understanding of #585 is that the information described there is basically needed for withdrawals for the construction of the follow-up Stacks transactions (and deposit info is nice-to-have).

Yeah I think a lot of this confusion is due to me when chatting with you about the uses of #585. Here are the ones that come to mind now (there may be more):
1. When constructing a bitcoin transaction, the signers need to know which deposit requests haven't been included as an input on another bitcoin transaction that has been confirmed. To figure this out they'd need to query their database for raw bitcoin transactions, deserialize them and check their inputs. That's very inefficient, but they could at least do it. With [#585](https://github.com/stacks-network/sbtc/issues/585) this becomes super easy.
2. Similarly to (1), when constructing the stacks transactions for deposits, a signer needs to know which requests have been included in a bitcoin transaction and confirmed. [#585](https://github.com/stacks-network/sbtc/issues/585) makes that super easy as well.
3. For withdrawal requests, the signers have analogous requirements to (1-2) above, except that, without [#585](https://github.com/stacks-network/sbtc/issues/585), a signer has to deserailize transactions and do the difficult merkle-tree matching thing in order to know for sure. With [#585](https://github.com/stacks-network/sbtc/issues/585) this becomes super easy.
4. When fetching the last fee info for RBF, the signers could use this information to get the fee rate for the transactions in the mempool. Without this they'd have to query their bitcoin node.

So, in some sense, the signers need this information for creating both bitcoin and stacks transactions, and it's needed for both deposits and withdrawals.


> Since the coordinator broadcasting the sweeps (and storing the request-mapping data) only has a small chance of also being the coordinator at the time these Stacks transactions will be constructed (we need to see the sweep confirmed first, right?), I don't see how #585 helps anything if we can't ensure that the signer creating the Stacks transactions also has the same information -- we must necessarily use another matching method as well as/instead of this solution, don't we?

This is what I meant by:

> * each signer stores all confirmed transactions where one of the outputs is a signer UTXO, and
> 
> * each signer notes the requests tied to each bitcoin transaction that may have been broadcast by another signer.

So, I think we're on the same page here. The one thing that the coordinator may know that the others don't is whether it was successfully broadcast, and I'm saying that this information isn't critical.

> If I'm reading what you wrote above correctly, I think you're saying that #585 is basically irrelevant if we start tying new Bitcoin blocks with essentially all known outstanding requests which "may be broadcast this block", and we do the "much more difficult" merkle-matching solution instead?

No, I'm saying that for (1-3) a signer only cares about whether the transaction has been confirmed in addition to the requests tied to the transaction.

For (4) knowing whether a transaction was broadcast is useful but not necessary. If they make a mistake the most likely outcome is that they end up paying more than necessary for the transaction fee.

signer/src/storage/model.rs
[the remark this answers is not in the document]
djordon: :( oh no. Well, maybe later we can wrap the `OutPoint` type.

signer/migrations/0003__create_tables.sql
djordon: Yeah, that is correct.

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
cylewitruk: Okay, so what I still don't get is how:

> each signer notes the requests tied to each bitcoin transaction that may have been broadcast by another signer.

Can be accomplished without #709? `BitcoinTransactionSignRequest` only contains the raw Bitcoin transaction and the aggregate key right now, no information about which requests are tied to it.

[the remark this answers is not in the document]
djordon: Yeah we'd need to update `BitcoinTransactionSignRequest`. The current version is not right for bitcoin validation already. A minimal updated version would look like [this](https://github.com/stacks-network/sbtc/blob/286-validate-bitcoin-transaction-sign-requests/signer/src/message.rs#L197-L204).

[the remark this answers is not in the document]
cylewitruk: Cool, right, so my only gripe would be that I'd like whatever data #709 sends to be enough that it can be used to populate the same tables.

signer/migrations/0003__create_tables.sql
cylewitruk: From our DB meeting, I will revert to natural keys.

signer/migrations/0003__create_tables.sql
cylewitruk: From our DB meeting, I'll revert this to natural keys.

signer/migrations/0003__create_tables.sql
cylewitruk: Awesome :pray:

cylewitruk: @djordon can this comment be closed?

signer/migrations/0003__create_tables.sql
cylewitruk: This will revert to using natural keys, i.e. txid will be the PK.

signer/migrations/0003__create_tables.sql
cylewitruk: Natural keys implemented

signer/migrations/0003__create_tables.sql
cylewitruk: Natural keys in [25c1b2f](https://github.com/stacks-network/sbtc/pull/710/commits/25c1b2ffd730e593228cc57e12574671aa9530cd)

signer/migrations/0003__create_tables.sql
cylewitruk: Natural keys in [25c1b2f](https://github.com/stacks-network/sbtc/pull/710/commits/25c1b2ffd730e593228cc57e12574671aa9530cd)

signer/migrations/0003__create_tables.sql
cylewitruk: Removed package-level storage in [25c1b2f](https://github.com/stacks-network/sbtc/pull/710/commits/25c1b2ffd730e593228cc57e12574671aa9530cd)

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
cylewitruk: Addressing the actual initial comment -- `created_at` is added and `broadcasted_at` removed in [25c1b2f](https://github.com/stacks-network/sbtc/pull/710/commits/25c1b2ffd730e593228cc57e12574671aa9530cd).

signer/migrations/0003__create_tables.sql
```diff
@@ -184,3 +177,81 @@ CREATE TABLE sbtc_signer.withdrawal_reject_events (
...
+CREATE TABLE sbtc_signer.sweep_transactions (
+    -- The Bitcoin transaction ID of the transaction.
+    txid BYTEA PRIMARY KEY NOT NULL,
+    -- The signer UTXO being spent in this transaction.
+    signer_prevout_txid BYTEA NOT NULL,
+    signer_prevout_output_index INTEGER NOT NULL,
+    signer_prevout_amount BIGINT NOT NULL,
+    signer_prevout_script_pubkey BYTEA NOT NULL,
+    -- The total amount of the transaction.
+    amount BIGINT NOT NULL,
```
djordon, line 193: Is this the total input amount or the total output amount. Either is fine, just thought that we should specify which.

signer/migrations/0003__create_tables.sql
```diff
@@ -184,3 +177,81 @@ CREATE TABLE sbtc_signer.withdrawal_reject_events (
...
+);
+
+-- Represents a single withdrawal request which has been included in a sweep
+-- transaction package. Withdrawal requests have a unique ID so we use that here
+-- to reference the withdrawal request together with its Stacks block hash,
+-- which can be retrieved from the `withdrawal_requests` table.
+CREATE TABLE sbtc_signer.swept_withdrawals (
+    sweep_transaction_txid BYTEA PRIMARY KEY NOT NULL,
+    -- The index of the sweep output in the sweep transaction.
+    output_index INTEGER NOT NULL,
```
djordon, line 211: I think this needs to be a composite primary key with these two fields.

signer/migrations/0003__create_tables.sql
```diff
@@ -184,3 +177,81 @@ CREATE TABLE sbtc_signer.withdrawal_reject_events (
...
+
+-- Represents a single deposit request which has been included in a
+-- transaction package. Deposit requests do not have a unique ID in the same way
+-- as withdrawal requests, so we reference the Bitcoin transaction ID and output
+-- index instead, which can be retrieved from the `deposit_requests` table.
+CREATE TABLE sbtc_signer.swept_deposits (
+    -- References the `packaged_transaction` in which this deposit was included.
+    sweep_transaction_txid BYTEA NOT NULL,
+    -- The index of the sweep input in the sweep transaction.
+    input_index INTEGER NOT NULL,
```
djordon, line 239: These two could be the primary key here. Or did you not want to have an explicit primary key for this table?

signer/src/storage/postgres.rs
```diff
@@ -1951,6 +2062,88 @@ impl super::DbWrite for PgStore {
...
+            ) 
+            VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)",
+        )
+        .bind(transaction.txid)
+        .bind(transaction.signer_prevout_txid)
+        .bind(transaction.signer_prevout_output_index as i32)
+        .bind(transaction.signer_prevout_amount as i64)
+        .bind(transaction.signer_prevout_script_pubkey.clone())
+        .bind(transaction.amount as i64)
+        .bind(transaction.fee as i64)
```
djordon, line 2094: For integer `as _` casts, we should either use the `from`/`try_from` implementations or if the `as _` cast is always fine, just explain why in a comment.

signer/src/storage/postgres.rs
```diff
@@ -401,6 +409,87 @@ impl PgStore {
...
+                signer_prevout_script_pubkey,
+                amount,
+                fee,
+                created_at_block_hash,
+                market_fee_rate
+            FROM
+                sweep_transactions
+            WHERE
+                txid = $1;
+        ",
```
djordon, line 439: nit: we should use a consistent style for our queries.

signer/src/transaction_coordinator.rs
```diff
@@ -296,9 +297,15 @@ where
...
         };
 
+        // Construct the transaction package and store it in the database.
         let transaction_package = pending_requests.construct_transactions()?;
 
         for mut transaction in transaction_package {
+            // Store the transaction in the database before we broadcast.
+            transaction
+                .store_as_sweep_transaction(&self.context.get_storage_mut(), bitcoin_chain_tip)
+                .await?;
```
djordon, line 307: Doesn't this mean that the coordinator stores it at a different time compared to the other signers?

signer/src/storage/postgres.rs
```diff
@@ -1951,6 +2062,88 @@ impl super::DbWrite for PgStore {
...
+        for deposit in transaction.swept_deposits.iter() {
+            sqlx::query(
+                "
+                INSERT INTO swept_deposits (
+                    sweep_transaction_txid
+                  , input_index
+                  , deposit_request_txid
+                  , deposit_request_output_index
+                )
+                VALUES ($1, $2, $3, $4)",
```
djordon, line 2111: We might want to add an "ON CONFLICT DO NOTHING" to the end of these queries.

signer/tests/integration/postgres.rs
```diff
@@ -1424,14 +1424,23 @@ async fn get_swept_deposit_requests_returns_swept_deposit_requests() {
...
 
+    // The request needs to be added to the database. This stores
+    // `setup.deposit_request` into the database.
+    setup.store_deposit_request(&db).await;
+
+    // TODO: Create the initial transaction sweep package without any
+    // withdrawals and have a separate method for creating that sweep (since
+    // it's not realistic to have the withdrawal in the same sweep as the
+    // deposit). Then we wouldn't have to do this.
+    setup.store_withdrawal_request(&db).await;
```
djordon, line 1435: Not sure if this is relevant to this comment but withdrawal request stored here is just random. It's not tied to the deposit.

signer/tests/integration/postgres.rs
```diff
@@ -1446,27 +1455,23 @@ async fn get_swept_deposit_requests_returns_swept_deposit_requests() {
...
+
     assert_eq!(req.amount, setup.deposit_request.amount);
     assert_eq!(req.txid, setup.deposit_request.outpoint.txid.into());
     assert_eq!(req.output_index, setup.deposit_request.outpoint.vout);
     assert_eq!(req.recipient, setup.deposit_recipient.into());
     assert_eq!(req.sweep_block_hash, setup.sweep_block_hash.into());
     assert_eq!(req.sweep_block_height, setup.sweep_block_height);
     assert_eq!(req.sweep_txid, setup.sweep_tx_info.txid.into());
-    assert_eq!(req.sweep_tx, setup.sweep_tx_info.tx.into());
+    //assert_eq!(swept_deposit.sweep_tx, setup.sweep_tx_info.tx.into());
```
djordon, line 1466: We can remove this line right? Or at least uncomment it.

djordon (commented):
Almost done going through the tests, figured I'd share my thoughts.

signer/tests/integration/postgres.rs
cylewitruk: Yeah the problem is that the setup helper creates a sweep tx with the withdrawal in it (which is what i meant by the TODO, that we should leave the withdrawal out of the first sweep tx), so the FK breaks if we haven't stored the withdrawal request.

signer/migrations/0003__create_tables.sql
djordon: yup!

signer/migrations/0003__create_tables.sql
cylewitruk: Good point, it's the output amount, will specify in comments.

signer/migrations/0003__create_tables.sql
cylewitruk: Good catch, just a miss

signer/migrations/0003__create_tables.sql
cylewitruk: Yup, good catch

signer/tests/integration/postgres.rs
cylewitruk: The comment is more related that we should, in the sweep setup helper, not include the withdrawal in the first sweep transaction and instead make a new sweep transaction when we're actually processing the withdrawal. Right now we have to insert the withdrawal request because the sweep tx will include it and otherwise break the FK.

signer/migrations/0003__create_tables.sql
cylewitruk: [a0bac87](https://github.com/stacks-network/sbtc/pull/710/commits/a0bac8711f4e83727a72f11d7a147633d8df4bd6)

signer/migrations/0003__create_tables.sql
cylewitruk: [a0bac87](https://github.com/stacks-network/sbtc/pull/710/commits/a0bac8711f4e83727a72f11d7a147633d8df4bd6)

signer/tests/integration/postgres.rs
djordon: Ahh I see. Makes sense.

signer/src/transaction_coordinator.rs
cylewitruk: Database time i.e. `created_at`? Yeah, but the time is largely useless.

djordon: Well I guess I'm thinking that something can fail between now and when the coordinator informs the other signers that they are going to (or have) broadcast the transaction.

cylewitruk: For #709 I think I'm planning on the broadcast to other signers happening directly after this line, and if the broadcast itself actually fails for some reason then the signers will have an extra "sweep transaction" in their db, but it'll never match any transactions in blocks.

Or do you think this should be closer to the end of `sign_and_broadcast()` right before the actual broadcast?

signer/src/storage/postgres.rs
cylewitruk: Added to all three in [d181170](https://github.com/stacks-network/sbtc/pull/710/commits/d1811705b06b1167b0e178f41a92f81e67242b14)

signer/src/storage/postgres.rs
cylewitruk: Adjusted in [d181170](https://github.com/stacks-network/sbtc/pull/710/commits/d1811705b06b1167b0e178f41a92f81e67242b14)

signer/src/storage/postgres.rs
cylewitruk: Think I got them all in [d181170](https://github.com/stacks-network/sbtc/pull/710/commits/d1811705b06b1167b0e178f41a92f81e67242b14)

signer/src/storage/postgres.rs
cylewitruk: Just a side note, I'm generally not overly fond of using "ON CONFLICT DO NOTHING" because it can mask potential issues; like, it could be that we've somehow ended up with two entries with the same PK but different data, and this will blindly ignore what maybe should be an update or indicative of a problem elsewhere.

signer/tests/integration/postgres.rs
cylewitruk: Yeah, removed in [21fab00](https://github.com/stacks-network/sbtc/pull/710/commits/21fab009a0ddf33f79b4b9928e06d67073656e5d). It was commented out because the raw tx is no longer a field on the swept deposit.

signer/src/storage/postgres.rs
djordon: Good point and we should aim to remove them later. This is a bandaid for the fact that we bail at the first sign of trouble.

djordon: Actually, maybe that's not the best reason to add them. I originally wanted to add them because I was more worried about duplicate messages than messages that have different data but matching primary keys. I still feel that way. The bail early thing is a different beast. I barely considered the different data stuff issue.

signer/src/transaction_coordinator.rs
djordon: Oh won't the tx signer receive this message then? I'm think the signer always reads the tx coordinators messages and does the writing. I'm thinking that it'll be simpler that way. Thoughts?

cylewitruk: I honestly hadn't thought through that yet, but yeah if we go that route then this line will just be moved to the signer and become the msg send instead.

djordon (approved):
Looks good! ✅ 🚢

signer/tests/integration/setup.rs
```diff
@@ -198,6 +210,20 @@ impl TestSweepSetup {
         }
     }
 
+    /// Store all pending transaction packages to the database. Note that this
+    /// is kept separate from [`Self::store_sweep_tx`] because this method
+    /// represents the transaction package which was broadcast to the mempool
+    /// (but not necessarily confirmed), while [`Self::store_sweep_tx`]
+    /// represents the sweep transaction which has been observed on-chain.
+    pub async fn store_sweep_transactions(&mut self, db: &PgStore) -> Vec<model::SweepTransaction> {
```
matteojug, line 218: Can we change the name of this or the other fn? Otherwise is kinda confusing which is doing what

matteojug (commented):
Late review, sorry 🙏

signer/src/storage/postgres.rs
```diff
@@ -1349,8 +1434,36 @@ impl super::DbRead for PgStore {
...
+        chain_tip: &model::BitcoinBlockHash,
+        context_window: u16,
+    ) -> Result<Option<model::SweepTransaction>, Error> {
+        let Some((txid,)): Option<(model::BitcoinTxId,)> = sqlx::query_as(
+            "
+            SELECT
+                tx.txid
+            FROM bitcoin_blockchain_of($1, $2) AS bitcoin
+            JOIN sweep_transactions AS tx ON bitcoin.block_hash = tx.created_at_block_hash
+            ORDER BY tx.created_at DESC",
```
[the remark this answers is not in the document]
cylewitruk: Yeah that's a good idea :+1:

signer/migrations/0004__create_functions.sql
```diff
@@ -0,0 +1,46 @@
...
+
+        SELECT
+            parent.block_hash,
+            parent.parent_hash,
+            parent.block_height,
+            last.depth + 1
+        FROM sbtc_signer.bitcoin_blocks AS parent
+        JOIN blockchain AS last
+            ON parent.block_hash = last.parent_hash
+        WHERE last.depth <= max_depth
```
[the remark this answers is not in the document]
cylewitruk: Sounds reasonable, this was a copy-paste masterpiece so nice catch :+1:

Issue #340, opened 2024-07-16
[Feature]: Emily Event Observer Sidecar


AshtonStephens:
# Feature - Emily Event Observer Sidecar

## 1. Description

This ticket is to implement the Emily event observer which will update the API with events that are considered trusted. 

How to handle the events is linked below:

### 1.1 Context & Purpose
The API sidecar will ingest the information required to update operation status based on the event data printed to chain and sent to the event observer. The sBTC operation clarity contracts will need to publish specific information about a transaction’s status when processing an sBTC operation.

This Sidecar will be updating the API database to be consistent.

## 2. Technical Details:

Technical details are in the link above, reposting here:
- #50 -> [Google Doc](https://docs.google.com/document/d/13AN9GdLfPYKKPkeHNU4niq_CyJCDxFR7DGBLzLc5gxU/edit#heading=h.jkxgl5anwobj)

### 2.1 Acceptance Criteria:
   - [ ] The sidecar can take in clarity events as described in the LLD
   - [ ] The sidecar is tested to the best of current ability in integration tests.
   - [ ] PR is created and merged 

## 3. Related Issues and Pull Requests (optional):
- Low level design: #50
    - [Event observer sidecar specific design](https://docs.google.com/document/d/13AN9GdLfPYKKPkeHNU4niq_CyJCDxFR7DGBLzLc5gxU/edit#heading=h.jkxgl5anwobj)

djordon:
Hmm, might be simpler than we think to do this. I was just looking and it looks like we might be able to specify that we only want [smart-contract-events](https://github.com/stacks-network/stacks-core/blob/47db1d0a8bf70eda1c93cb3e0731bdf5595f7baa/testnet/stacks-node/src/config.rs#L2822-L2823). I think these [lines dictate the format](https://github.com/stacks-network/stacks-core/blob/47db1d0a8bf70eda1c93cb3e0731bdf5595f7baa/testnet/stacks-node/src/config.rs#L2866-L2897).

AshtonStephens:
This totally makes sense, @jovanvuleta are you familiar with how to execute the smart contracts on a local environment?

jovanvuleta:
Until now I have been deploying our contracts via clarinet integrate, will I be able to do the same here or is the flow different? @AshtonStephens

AshtonStephens:
Look here for some existing sidecar code: signer/src/stacks/webhooks.rs

AshtonStephens:
[...]

PR #712, opened 2024-10-25
feat: emily event observer sidecar

Jiloc:
## Description

Closes: #340 

## Changes
- Extends EmilyClient to expose `create_withdrawals`, `update_withdrawals` and `update_deposits`.
- Extends `new_block_handler` to parse the `NewBlockEvent` for sBTC contracts events, construct the corresponding requests, and concurrently send the updates to Emily. 
    - The data required for creating the Emily requests, not contained in the event itself, is read from the database.
    - In case the required data is not in the database or communication with Emily fails, we just log a warning. We rely on the redundancy of all sBTC signers to guarantee that messages are received from Emily.

### Notes: 
- Confirmed `DepositUpdate` and `WithdrawalUpdate` requests to Emily require to fill out the `btc_fee` spent for the tx. The print event for `WithdrawalAcceptEvent` already contains this information, while the `CompletedDepositEvent` doesn't. At the moment I hardcoded it to a value of 1.

## Checklist:

### DIFF
--- a/signer/src/api/mod.rs
+++ b/signer/src/api/mod.rs
@@ -4,8 +4,14 @@
 pub mod new_block;
 pub mod status;
 
+use emily_client::models::Chainstate;
+use emily_client::models::UpdateDepositsResponse;
+use emily_client::models::UpdateWithdrawalsResponse;
+use emily_client::models::Withdrawal;
 pub use new_block::new_block_handler;
 pub use status::status_handler;
+
+use crate::error::Error;
 
 /// A struct with state data necessary for runtime operation.
 #[derive(Debug, Clone)]
@@ -16,3 +22,10 @@ pub struct ApiState<C> {
 
 /// The name of the sbtc registry smart contract.
 const SBTC_REGISTRY_CONTRACT_NAME: &str = "sbtc-registry";
+
+enum UpdateResult {
+    Deposit(Result<UpdateDepositsResponse, Error>),
+    Withdrawal(Result<UpdateWithdrawalsResponse, Error>),
+    CreatedWithdrawal(Vec<Result<Withdrawal, Error>>),
+    Chainstate(Result<Chainstate, Error>),
+}
--- a/signer/src/api/new_block.rs
+++ b/signer/src/api/new_block.rs
@@ -2,21 +2,38 @@
 //! which is for processing new block webhooks from a stacks node.
 //!
 
-use std::sync::OnceLock;
-
 use axum::extract::State;
 use axum::http::StatusCode;
-use emily_client::models::Chainstate;
-
+use bitcoin::Txid;
 use clarity::vm::representations::ContractName;
 use clarity::vm::types::QualifiedContractIdentifier;
 use clarity::vm::types::StandardPrincipalData;
+use emily_client::models::Chainstate;
+use emily_client::models::CreateWithdrawalRequestBody;
+use emily_client::models::DepositUpdate;
+use emily_client::models::Fulfillment;
+use emily_client::models::Status;
+use emily_client::models::WithdrawalParameters;
+use emily_client::models::WithdrawalUpdate;
+use futures::FutureExt;
+use std::sync::OnceLock;
 
+use crate::api::UpdateResult;
 use crate::context::Context;
 use crate::emily_client::EmilyInteract;
+use crate::error::Error;
+use crate::stacks::events::CompletedDepositEvent;
 use crate::stacks::events::RegistryEvent;
 use crate::stacks::events::TxInfo;
+use crate::stacks::events::WithdrawalAcceptEvent;
+use crate::stacks::events::WithdrawalCreateEvent;
+use crate::stacks::events::WithdrawalRejectEvent;
 use crate::stacks::webhooks::NewBlockEvent;
+use crate::storage::model::BitcoinBlock;
+use crate::storage::model::BitcoinTxId;
+use crate::storage::model::StacksBlock;
+use crate::storage::model::StacksBlockHash;
+use crate::storage::DbRead;
 use crate::storage::DbWrite;
 
 use super::ApiState;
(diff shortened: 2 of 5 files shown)


signer/src/api/new_block.rs
```diff
@@ -112,34 +143,266 @@ pub async fn new_block_handler(state: State<ApiState<impl Context>>, body: Strin
...
+        .await?
+        .ok_or(Error::MissingBlock)?;
+
+    Ok(CreateWithdrawalRequestBody {
+        amount: event.amount,
+        parameters: Box::new(WithdrawalParameters { max_fee: event.max_fee }),
+        recipient: event.recipient.to_string(),
+        request_id: event.request_id,
+        stacks_block_hash: event.block_id.to_hex(),
+        stacks_block_height: stacks_block.block_height,
```
djordon, line 352: The height of the stacks block is given in the webhook body as the `block_height` field. We should use that instead.

signer/src/api/new_block.rs
```diff
@@ -112,34 +143,266 @@ pub async fn new_block_handler(state: State<ApiState<impl Context>>, body: Strin
...
+///
+/// # Notes
+/// - If the transaction is missing from any blocks, it returns an `Error::BitcoinTxMissing`.
+/// - In case of database access errors, the caller should handle the error appropriately.
+async fn fetch_bitcoin_block_from_txid(
+    db: &impl DbRead,
+    txid: Txid,
+) -> Result<BitcoinBlock, Error> {
+    let btc_txid = BitcoinTxId::from(txid);
+    let blocks_hashes = db.get_bitcoin_blocks_with_transaction(&btc_txid).await?;
```
djordon, line 227: The necessary bitcoin block info is given in the webhook body. Better to use the webhook body exclusively because our database might not be up to date by the time this function runs.

signer/src/api/new_block.rs
```diff
@@ -112,34 +143,266 @@ pub async fn new_block_handler(state: State<ApiState<impl Context>>, body: Strin
...
+            tracing::warn!(%error, "Failed to update deposits in emily");
+        } else if let UpdateResult::Withdrawal(Err(error)) = result {
+            tracing::warn!(%error, "Failed to update withdrawals in emily");
+        } else if let UpdateResult::CreatedWithdrawal(results) = result {
+            for result in results {
+                if let Err(error) = result {
+                    tracing::warn!(%error, "Failed to create withdrawals in emily");
+                }
+            }
+        }
```
djordon, line 202: nit: why not a match statement here?

signer/src/api/mod.rs
```diff
@@ -16,3 +22,10 @@ pub struct ApiState<C> {
 
 /// The name of the sbtc registry smart contract.
 const SBTC_REGISTRY_CONTRACT_NAME: &str = "sbtc-registry";
+
+enum UpdateResult {
+    Deposit(Result<UpdateDepositsResponse, Error>),
+    Withdrawal(Result<UpdateWithdrawalsResponse, Error>),
+    CreatedWithdrawal(Vec<Result<Withdrawal, Error>>),
+    Chainstate(Result<Chainstate, Error>),
+}
```
djordon, line 31: nit: We should probably just move this to where it is used. We've been trying to keep our `mod.rs` files pretty sparse.

djordon (commented):
I haven't finished the review yet, but the one thing that I've seen so far is that we probably shouldn't be reading from the database in our functions that construct the Emily request object.

signer/src/api/new_block.rs
Jiloc: You are definitely right, unfortunately though for most of the events (3 out of 4) we are missing some values. For this case in particular, I saw the `block_height` field, but it's described to be block height of the bitcoin chain, not for stacks. But maybe the comments are just wrong?

https://github.com/stacks-network/sbtc/blob/main/signer/src/stacks/events.rs#L164

```
    /// The block height of the bitcoin blockchain when the stacks
    /// transaction that emitted this event was executed.
    pub block_height: u64,
    ```

Comment (djordon) [reply signer/src/api/new_block.rs:352]: Ahh, there are two such fields. I was thinking that you use this one https://github.com/stacks-network/sbtc/blob/7fd644b081f0ece1c241ea5e2b2977f011cfbb64/signer/src/stacks/webhooks.rs#L44-L50

File: signer/src/api/new_block.rs (hunk 2 above)
Comment (Jiloc) [reply signer/src/api/new_block.rs:227]: I have a question regarding this. From my understanding, the body of the webhook contains information about the Bitcoin tip (and its parent). The `Fulfillment` struct requires `The Bitcoin block hash of the block that included the fulfilling Bitcoin transaction` (from the docs). Can we be certain that the Bitcoin block information in the webhook corresponds to the block fulfilling the transaction, or could it potentially be more recent?

File: signer/src/api/mod.rs (hunk 4 above)
Comment (Jiloc) [reply signer/src/api/mod.rs:31]: Got it!

File: signer/src/api/new_block.rs (hunk 3 above)
Comment (Jiloc) [reply signer/src/api/new_block.rs:202]: ah, you caught me! I went for the greedy route to avoid adding the extra branch for successful cases :D but I’ll happily switch to the match statement. Thanks for the nudge!

File: signer/src/api/new_block.rs (hunk 1 above)
Comment (Jiloc) [reply signer/src/api/new_block.rs:368]: Oh got it! About that field, I didn't use it on purpose for a similar reason I explained in the other comment. 
In  this case the `CreateWithdrawalRequestBody` wants the `block_height` and `block_hash` of the `The stacks block hash in which this request id was initiated.`

The one we have in the `NewBlockEvent` is the Stacks chaintip. It should represent the block that confirmed the transaction, not the one that initiated the request, right?

If we can guarantee that the block in the `NewBlockEvent` is the same as the one that originated the request, then I can switch to that!

Comment (djordon) [reply signer/src/api/new_block.rs:368]: > The one we have in the `NewBlockEvent` is the Stacks chaintip. It should represent the block that confirmed the transaction, not the one that initiated the request, right?

Great question. In this case, the block where the request was initiated is the same as the block where the transaction gets confirmed.

File: signer/src/api/new_block.rs (hunk 2 above)
Comment (djordon) [reply signer/src/api/new_block.rs:243]: > The `Fulfillment` struct requires `The Bitcoin block hash of the block that included the fulfilling Bitcoin transaction` (from the docs). Can we be certain that the Bitcoin block information in the webhook corresponds to the block fulfilling the transaction, or could it potentially be more recent?

Oh right. To do this properly we need to wrap up https://github.com/stacks-network/sbtc/pull/493. For now I'd add a todo comment there about the dependency and put some dummy value (like the current bitcoin chain tip or something).

File: signer/src/api/new_block.rs (hunk 1 above)
Comment (Jiloc) [reply signer/src/api/new_block.rs:368]: Perfect!

File: signer/src/api/new_block.rs [hunk 5]
```diff
@@ -112,34 +162,251 @@ pub async fn new_block_handler(state: State<ApiState<impl Context>>, body: Strin
...
+        .write_withdrawal_create_event(&event)
+        .await?;
+
+    Ok(CreateWithdrawalRequestBody {
+        amount: event.amount,
+        parameters: Box::new(WithdrawalParameters { max_fee: event.max_fee }),
+        recipient: event.recipient.to_string(),
+        request_id: event.request_id,
+        stacks_block_hash: event.block_id.to_hex(),
+        stacks_block_height,
```
Jiloc, line 356: @djordon once the PRs you mentioned are implemented, all sBTC contract events will include the data needed to create updates in Emily. Currently, WithdrawalCreateEvent is the only outlier, as it has the hash but not stacks_block_height. Do you think it’d be worthwhile to request this field in the sbtc-registry contract’s print event for consistency? It’s not a blocker, but it would align with the others. If so, could you point me to the relevant issue, or should I open a new one?

AshtonStephens:
@Jiloc FYI, the version conflict issue occurs because there are two threads trying to update Emily simultaneously, BUT, that's supposed to be fine. That version conflict isn't meant to throw the error; it's meant to initiate a retry within Emily, which is simple logic to add (if the error is Version Conflict, retry). 

So effectively, this is an Emily bug, and we should fix it, but it's a simple fix.

AshtonStephens:
> @Jiloc FYI, the version conflict issue occurs because there are two threads trying to update Emily simultaneously, BUT, that's supposed to be fine. That version conflict isn't meant to throw the error; it's meant to initiate a retry within Emily, which is simple logic to add (if the error is Version Conflict, retry).
> 
> So effectively, this is an Emily bug, and we should fix it, but it's a simple fix.

You can put nearly this exact snippet in the `update_deposits` function:

```rust
        // Loop through all updates and execute.
        for (index, update) in validated_request.deposits {
            let updated_deposit: Deposit;
            // Attempt update as many times as there are signers maintaining the state.
            for _ in 0..15 {
                // Get original deposit entry.
                let deposit_entry = accessors::get_deposit_entry(&context, &update.key).await?;
                let update_package: DepositUpdatePackage = DepositUpdatePackage::try_from(&deposit_entry, update)?;
                let maybe_update_deposit =
                    accessors::update_deposit(&context, &update_package)
                        .await
                        .map(|entry| Some(entry))
                        .map_err(|e| match e {
                            // Do nothing if there's a version conflict error.
                            Error::VersionConflict => Ok(None),
                            e => Err(e),
                        })?;
                if let Some(updated_deposit_entry) = maybe_update_deposit {
                    updated_deposits.push((index, updated_deposit_entry.try_into()?));
                    break;
                }
            }
        }
```

signer/src/api/new_block.rs
djordon: Ohh, I don't think we need that in the print event. I think for `WithdrawalCreateEvent` you should use the block height included in the webhook, just like you're doing now.

PR #713, opened 2024-10-26
feat: properly parse the reclaim script lock time

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/516, with proper lock-time handling being done when https://github.com/stacks-network/sbtc/issues/380 gets closed.

After this change, locktimes used as `OP_CSV` inputs in the reclaim script only use the 16 least significant bits for the value of the locktime. All other bits in the 32-bit locktime must be zero or else the deposit transaction will fail validation. When we support time-based locktimes, a user may set bit (1 << 22) to indicate that the locktime value is time based, as described in BIP-68.

## Changes

* Use rust-bitcoin's relative `LockTime` type for representing the lock-time under the hood.
* Reject time-based `locktime`s. We've set ourselves up to support them, but the real work lies in the signer.
* Change the underlying type for a `locktime` to be a `u32`.

## Checklist:

### DIFF
--- a/sbtc/src/deposits.rs
+++ b/sbtc/src/deposits.rs
@@ -401,9 +401,6 @@ impl ReclaimScriptInputs {
 
     /// Get the lock time in the reclaim script.
     pub fn lock_time(&self) -> u32 {
-        // We know this number is positive because that is one of the
-        // invariants upheld by this struct, and we check for it in
-        // `Self::try_new`, so this will never be a lossy conversion.
         self.lock_time.to_consensus_u32()
     }
 
@@ -468,7 +465,7 @@ impl ReclaimScriptInputs {
             // works in bitcoin-core:
             // https://github.com/bitcoin/bitcoin/blob/v27.1/src/script/interpreter.cpp#L531-L573
             // That said, we only accepts 4-byte unsigned integers, and we
-            // that check below.
+            // check that below.
             [n, rest @ ..] if *n <= 5 && rest.get(*n as usize) == Some(&OP_CSV) => {
                 // We know the error and panic paths cannot happen because
                 // of the above `if` check.
--- a/sbtc/src/deposits.rs
+++ b/sbtc/src/deposits.rs
@@ -7,6 +7,7 @@ use bitcoin::taproot::LeafVersion;
 use bitcoin::taproot::NodeInfo;
 use bitcoin::taproot::TaprootSpendInfo;
 use bitcoin::Address;
+use bitcoin::locktime::relative::LockTime;
 use bitcoin::Network;
 use bitcoin::OutPoint;
 use bitcoin::Script;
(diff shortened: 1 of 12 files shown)


evonide:
Please note that the timelock type is specified in the 23rd bit (1<< 22) not in the 22nd bit from what  I've seen.

djordon:
> Please note that the timelock type is specified in the 23rd bit (1<< 22) not in the 22nd bit from what I've seen.
> 
> For example, https://learn.saylor.org/mod/book/view.php?id=36369&chapterid=18998 has a useful description for this.

Oh I was thinking that the "first bit" was called the "zeroth bit". To avoid this kind of confusion I'll refer to it as the bit `1 << 22` like they do in BIP 68.

sbtc/src/deposits.rs
```diff
@@ -349,49 +350,57 @@ impl DepositScriptInputs {
...
-            return Err(Error::InvalidReclaimScriptLockTime(lock_time));
+        let lock_time = LockTime::from_consensus(lock_time).map_err(Error::DisabledLockTime)?;
+
+        // For now, we only accept locktimes denominated in bitcoin block
+        // units.
+        if matches!(lock_time, LockTime::Time(_)) {
+            return Err(Error::UnsupportedLockTimeUnits(
+                lock_time.to_consensus_u32(),
+            ));
         }
```
AshtonStephens, line 396: The LockTime bitcoin type is an enum that can support both block height and the time interpretation:

```
pub enum LockTime {
    /// A block height lock time value.
    Blocks(Height),
    /// A 512 second time interval value.
    Time(Time),
}
```

What is the rationale for not supporting both? It's not hard to support in #380

AshtonStephens: Requesting changes until I understand why we're choosing to only support one method. I think this lack of support is something that will catch developers off guard when they don't read the fine print of what we do and don't support.

AshtonStephens (changes requested):

sbtc/src/deposits.rs
djordon: > What is the rationale for not supporting both? It's not hard to support in #380

Well, the lack of support in this PR isn't meant to be the end-all. We can merge this PR and then #380 can remove the restriction if it is straightforward to do so. At the same time, I could see a judgement call being made where we forgo the extra work to support it for now and ship without supporting it.



> Requesting changes until I understand why we're choosing to only support one method. I think this lack of support is something that will catch developers off guard when they don't read the fine print of what we do and don't support.

Well, yeah they'll be caught off guard with this if they don't read the docs. But this is not the only place where this is true, much of our deposit protocol is fine print and we are strict about it's adherence. And I'm not against being mindful of the potential pitfalls a developer may face, I'm just weighing that against the time spent on other work that needs to get done as we sprint towards the finish.

sbtc/src/deposits.rs
```diff
@@ -349,49 +350,57 @@ impl DepositScriptInputs {
...
 ///
 /// This struct upholds the invariant that the `lock_time` is a valid and
-/// standard `locktime` in bitcoin-core. So it is positive and less than or
-/// equal to the maximum acceptable value of `2**39 - 1`. We do not check
-/// whether the user supplied script is correct and standard.
+/// standard locktime in bitcoin-core. We do not verify whether the
+/// user-supplied script is correct and standard.
+///
+/// Currently, we only accept locktimes denominated in Bitcoin blocks. This
+/// implies that bit (1 << 22) in the locktime, must be zero.
```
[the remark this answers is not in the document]
djordon: Yup.

AshtonStephens (approved):
LGTM

Issue #701, opened 2024-10-23
[Feature]: Add delay to processing of new blocks


cylewitruk:
# Feature - Add delay to processing of new blocks

## 1. Description

The signer is written in such a way at the moment that the signers/coordinator will react _immediately_ upon receiving a new block update from the Bitcoin Core ZMQ stream.

The potential issue with this is that, since we don't currently have any consensus mechanism around new blocks, that the coordinator starts sending out signing messages immediately, where other signers may not yet have received the information from their Bitcoin Core (i.e. if they have high latency or are on the other end of the network). This will cause signers which don't have the block to fail signing for related requests.

Since we know that Bitcoin blocks are roughly 10 minutes apart, and our processing should only take some seconds, we could avoid this in the easiest way by adding a delay in the **transaction coordinator** loop which waits _X_-seconds (15, maybe?) before beginning processing of new blocks, making the assumption that most-if-not-all signers should have received the block by this time.

### 1.1 Context & Purpose

Described above.

## 2. Technical Details:

Add the delay time as a parameter to `run()` so that we can set this to the desired value in `main()` but still keep this short in tests; this will also require adding the value (`0`) to all the tests which currently instantiate and run a coordinator.

### 2.1 Acceptance Criteria:
   - [ ] The transaction coordinator waits X seconds prior to beginning block processing (and thus sending signing requests to peers)

PR #714, opened 2024-10-28
feat: add delay to processing of new Bitcoin blocks

aldur:
## Description

Closes: #701 

## Changes

- As suggested in #701, the coordinator now waits 15 seconds before processing new Bitcoin blocks.
- In tests, it doesn't wait (unless explicitly asked to).

## Testing Information

Run: `cargo test --lib --package signer should_wait_before_processing_bitcoin_blocks`.

## Checklist:

Note that the following tests fail locally and on current `main` as well:

```txt
test network::libp2p::network::tests::swarm_rejects_connections_from_unknown_peers ... FAILED
test network::libp2p::network::tests::two_clients_should_be_able_to_exchange_messages_given_a_libp2p_network ... FAILED
```

Failure does not seem related to this PR.

### DIFF
--- a/signer/src/config/default.toml
+++ b/signer/src/config/default.toml
@@ -1,3 +1,4 @@
+# TODO(715): Provide sane/safe configuration defaults. Re-review all of them!
 # TODO(429): Add documentation for all configuration parameters.
 
 # !! ==============================================================================
@@ -91,6 +92,7 @@ endpoints = ["http://localhost:20443"]
 
 # This is the start height of the first EPOCH 3.0 block on the stacks
 # blockchain.
+# TODO(715): Update to 867867
 nakamoto_start_height = 31
 
 # !! ==============================================================================
@@ -108,6 +110,7 @@ nakamoto_start_height = 31
 # Format: "<hex-encoded-private-key>" (64 or 66 hex-characters)
 # Required: true
 # Environment: SIGNER_SIGNER__PRIVATE_KEY
+# TODO(715): Remove default value.
 private_key = "41634762d89dfa09133a4a8e9c1378d0161d29cd0a9433b51f1e3d32947a73dc"
 
 # Specifies which network to use when constructing and sending transactions
@@ -119,11 +122,13 @@ private_key = "41634762d89dfa09133a4a8e9c1378d0161d29cd0a9433b51f1e3d32947a73dc"
 # Required: true
 # Possible values: mainnet, testnet, regtest
 # Environment: SIGNER_SIGNER__NETWORK
+# TODO(715): Change to mainnet.
 network = "regtest"
 
 # The address that deployed the sbtc smart contracts.
 #
 # Required: true
+# TODO(715): Change after SCs have been deployed.
 deployer = "SN3R84XZYA63QS28932XQF3G1J8R9PC3W76P9CSQS"
 
 # The signer database endpoint (pgsql connection string)
@@ -136,6 +141,7 @@ db_endpoint = "postgresql://postgres:postgres@localhost:5432/signer"
 # peers who are approved to be in the sBTC signer set.
 #
 # Required: true Environment: SIGNER_SIGNER__BOOTSTRAP_SIGNING_SET
+# TODO(715): Change after initial signing set has been determined.
 bootstrap_signing_set = [
     "035249137286c077ccee65ecc43e724b9b9e5a588e3d7f51e3b62f9624c2a49e46",
     "031a4d9f4903da97498945a4e01a5023a1d53bc96ad670bfe03adf8a06c52e6380",
@@ -147,10 +153,12 @@ bootstrap_signing_set = [
 # `bootstrap_signing_set`. Must be strictly positive.
 # 
 # Required: true Environment: SIGNER_SIGNER__BOOTSTRAP_SIGNATURES_REQUIRED
+# TODO(715): Change as specified by sBTC SIP.
 bootstrap_signatures_required = 2
 
 # Seconds to wait before processing a new Bitcoin block.
 # Required: true Environment: SIGNER_SIGNER__BITCOIN_PROCESSING_DELAY
+# TODO(715): Add default delay (15 seconds? see #701)
 bitcoin_processing_delay = 0
 
 # !! ==============================================================================
@@ -188,6 +196,7 @@ bind = "0.0.0.0:8801"
 # Environment: SIGNER_SIGNER__P2P__SEEDS
 # Environment Example: tcp://seed-1:4122,tcp://seed-2:4122
 # TODO(429): Add well-known seed nodes
+# TODO(715): Add well-known seed nodes
 seeds = []
 
 # The local network interface(s) and port(s) to listen on. 
--- a/signer/src/config/default.toml
+++ b/signer/src/config/default.toml
@@ -148,6 +148,11 @@ bootstrap_signing_set = [
 # 
 # Required: true Environment: SIGNER_SIGNER__BOOTSTRAP_SIGNATURES_REQUIRED
 bootstrap_signatures_required = 2
+
+# Seconds to wait before processing a new Bitcoin block.
+# Required: true Environment: SIGNER_SIGNER__BITCOIN_PROCESSING_DELAY
+bitcoin_processing_delay = 0
+
 # !! ==============================================================================
 # !! Stacks Event Observer Configuration
 # !!
--- a/signer/src/config/mod.rs
+++ b/signer/src/config/mod.rs
@@ -11,6 +11,7 @@ use std::path::Path;
 use url::Url;
 
 use crate::config::error::SignerConfigError;
+use crate::config::serialization::duration_seconds_deserializer;
 use crate::config::serialization::p2p_multiaddr_deserializer_vec;
 use crate::config::serialization::parse_stacks_address;
 use crate::config::serialization::private_key_deserializer;
@@ -234,6 +235,11 @@ pub struct SignerConfig {
     /// The number of signatures required for the signers' bootstrapped
     /// multi-sig wallet on Stacks.
     pub bootstrap_signatures_required: u16,
+    /// The number of seconds the coordinator will wait
+    /// before processing a new Bitcoin block
+    /// (allowing it to propagate to the others signers)
+    #[serde(deserialize_with = "duration_seconds_deserializer")]
+    pub bitcoin_processing_delay: std::time::Duration,
 }
 
 impl Validatable for SignerConfig {
@@ -595,6 +601,15 @@ mod tests {
             Some(url::Host::Ipv6(ip))
         );
         assert_eq!(settings.stacks.endpoints[0].port(), Some(9101));
+
+        let delay = 42;
+        std::env::set_var("SIGNER_SIGNER__BITCOIN_PROCESSING_DELAY", delay.to_string());
+
+        let settings = Settings::new_from_default_config().unwrap();
+        assert_eq!(
+            settings.signer.bitcoin_processing_delay,
+            std::time::Duration::from_secs(delay),
+        );
     }
 
     #[test]
--- a/signer/src/config/mod.rs
+++ b/signer/src/config/mod.rs
@@ -11,7 +11,7 @@ use std::path::Path;
 use url::Url;
 
 use crate::config::error::SignerConfigError;
-use crate::config::serialization::duration_seconds_deserializer;
+use crate::config::serialization::bitcoin_processing_delay_deserializer;
 use crate::config::serialization::p2p_multiaddr_deserializer_vec;
 use crate::config::serialization::parse_stacks_address;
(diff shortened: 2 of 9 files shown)


signer/src/main.rs
```diff
@@ -382,5 +382,5 @@ async fn run_transaction_coordinator(ctx: impl Context) -> Result<(), Error> {
         dkg_max_duration: Duration::from_secs(10),
     };
 
-    coord.run().await
```
AshtonStephens, line 385: I think we should make this configurable, because if we find that we have significant issues somehow when there's a Bitcoin block storm it would be good to be able to tweak this. 

I don't think it'll be a hard change to make; the config classes are pretty clear, and you can put it under the bitcoin config settings [here](https://github.com/stacks-network/sbtc/blob/main/signer/src/config/mod.rs#L101) named something like `bitcoin_processing_delay`, and you'll have access to it via the `ctx` variable.

signer/src/transaction_coordinator.rs
```diff
@@ -213,7 +217,10 @@ where
     }
 
     #[tracing::instrument(skip(self))]
-    async fn process_new_blocks(&mut self) -> Result<(), Error> {
+    async fn process_new_blocks(
+        &mut self,
+        with_delay: tokio::time::Duration,
+    ) -> Result<(), Error> {
```
AshtonStephens, line 223: If you add the delay to the config you can delete this change and access the configuration via the context within the `self` reference.

signer/src/transaction_coordinator.rs
```diff
@@ -1028,8 +1040,29 @@ mod tests {
...
         test_environment()
-            .assert_should_be_able_to_coordinate_signing_rounds()
+            .assert_should_be_able_to_coordinate_signing_rounds(tokio::time::Duration::ZERO)
+            .await;
+    }
+
+    #[tokio::test]
+    async fn should_wait_before_processing_bitcoin_blocks() {
+        // NOTE: Above test `should_be_able_to_coordinate_signing_rounds`
+        // could be removed as redundant now.
```
AshtonStephens, line 1050: I'm curious why you added this comment instead of just deleting `should_be_able_to_coordinate_signing_rounds`. Are you asking for confirmation that you can delete it?

AshtonStephens (changes requested):
We should change the approach to have the the configuration file take in a `bitcoin_block_processing_delay_s` variable for two reasons:
1. We may later want to change it across signers for some reason we don't yet know
2. It will make the code cleaner because the areas that need this value already have access to the configurations, making it not necessary to add to any function contracts.

signer/src/transaction_coordinator.rs
```diff
@@ -164,7 +168,7 @@ where
                     // signer indicating that it has handled new requests.
                     Ok(SignerSignal::Event(SignerEvent::TxSigner(TxSignerEvent::NewRequestsHandled))) => {
                         tracing::debug!("received block observer notification");
-                        let _ = self.process_new_blocks().await
+                        let _ = self.process_new_blocks(delay_to_process_new_blocks).await
```
cylewitruk, line 171: At first I was going to write that it would be easier to have the delay here, but I actually like that putting it in `process_blocks()` _after_ the "are we the coordinator" check lets us skip the delay if we determine that we're not the coordinator.

signer/src/transaction_coordinator.rs
```diff
@@ -148,7 +149,10 @@ where
 {
     /// Run the coordinator event loop
     #[tracing::instrument(skip(self), name = "tx-coordinator")]
-    pub async fn run(mut self) -> Result<(), Error> {
+    pub async fn run(
+        mut self,
+        delay_to_process_new_blocks: tokio::time::Duration,
```
cylewitruk, line 154: nit: you should be able to use `std::time::Duration`

signer/tests/integration/emily.rs
```diff
@@ -402,7 +402,8 @@ async fn deposit_e2e() {
         signing_round_max_duration: Duration::from_secs(10),
         dkg_max_duration: Duration::from_secs(10),
     };
-    let tx_coordinator_handle = tokio::spawn(async move { tx_coordinator.run().await });
+    let tx_coordinator_handle =
+        tokio::spawn(async move { tx_coordinator.run(tokio::time::Duration::ZERO).await });
```
cylewitruk, line 406: nit: should be able to use `std::time::Duration`

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -278,7 +278,8 @@ async fn process_complete_deposit() {
         signing_round_max_duration: Duration::from_secs(10),
         dkg_max_duration: Duration::from_secs(10),
     };
-    let tx_coordinator_handle = tokio::spawn(async move { tx_coordinator.run().await });
+    let tx_coordinator_handle =
+        tokio::spawn(async move { tx_coordinator.run(tokio::time::Duration::ZERO).await });
```
cylewitruk, line 282: nit: should be able to use `std::time::Duration` (already `use`'d in this file)

signer/src/main.rs
aldur: I like that. The reason I didn't do it already is that there seem to be a bunch of hard-coded values already, eg. here: https://github.com/stacks-network/sbtc/blob/c8c133d0d2e5cd8c572bb964c03ba0b25a1febef/signer/src/main.rs#L380.

Let me add it.

signer/src/transaction_coordinator.rs
aldur: Yep, gathering feedback but also, if in doubt, with tests more is better :)

signer/src/transaction_coordinator.rs
```diff
@@ -248,6 +255,11 @@ where
             None => self.coordinate_dkg(&bitcoin_chain_tip).await?,
         };
 
+        if with_delay > tokio::time::Duration::ZERO {
```
cylewitruk, line 258: bug? I think this needs to be moved before the above block coordinating DKG since it depends on the Bitcoin chain tip (which is what we're worried that the other signers may not have updated yet hence the delay).

signer/src/main.rs
cylewitruk: Sounds good, but I don't think this is a value that we want signers freely messing around with, so we should be careful about how we handle it in the `default.toml` and config examples.

signer/src/config/default.toml
```diff
@@ -148,6 +148,11 @@ bootstrap_signing_set = [
 # 
 # Required: true Environment: SIGNER_SIGNER__BOOTSTRAP_SIGNATURES_REQUIRED
 bootstrap_signatures_required = 2
+
+# Seconds to wait before processing a new Bitcoin block.
+# Required: true Environment: SIGNER_SIGNER__BITCOIN_PROCESSING_DELAY
+bitcoin_processing_delay = 0
```
aldur, line 154: Setting this to 0, but we should have a non-zero default when shipping to prod.

signer/src/transaction_coordinator.rs
```diff
@@ -1028,8 +1035,35 @@ mod tests {
...
+        // Locally this takes a couple seconds to execute.
+        // This truncates the decimals.
+        let baseline_elapsed = tokio::time::Duration::from_secs(baseline_start.elapsed().as_secs());
+
+        let delay_i = 3;
+        let delay = tokio::time::Duration::from_secs(delay_i);
+        std::env::set_var(
+            "SIGNER_SIGNER__BITCOIN_PROCESSING_DELAY",
+            delay_i.to_string(),
+        );
```
aldur, line 1061: Not sure if there's a better way to do this.

cylewitruk: I think I'd make the delay config explicitly in seconds as a uint.

signer/src/transaction_coordinator.rs
AshtonStephens: Resolved.

signer/src/transaction_coordinator.rs
AshtonStephens: Makes sense, resolved.

signer/src/main.rs
AshtonStephens: Resolved.

signer/src/config/mod.rs
```diff
@@ -234,6 +235,11 @@ pub struct SignerConfig {
     /// The number of signatures required for the signers' bootstrapped
     /// multi-sig wallet on Stacks.
     pub bootstrap_signatures_required: u16,
+    /// The number of seconds the coordinator will wait
+    /// before processing a new Bitcoin block
+    /// (allowing it to propagate to the others signers)
+    #[serde(deserialize_with = "duration_seconds_deserializer")]
+    pub bitcoin_processing_delay: std::time::Duration,
```
cylewitruk, line 242: I think i would make the config explicitly in seconds, with a validation ceiling of 300 or so (half the expected time of a bitcoin block); that'll keep things simpler.

AshtonStephens (approved):
Approving, but please wait for any necessary resolutions on @cylewitruk's feedback before merging :)

signer/src/config/mod.rs
cylewitruk: Nevermind, I read too quickly, you're doing this already 😅

aldur: Do you mean renaming the variable to `... _seconds`? Can add the ceiling, no problem.

signer/src/transaction_coordinator.rs
aldur: Uh, I think you are right. Done.

signer/src/config/mod.rs
cylewitruk: No I actually meant what you are doing, I missed your custom deserializer at a glance and briefly thought that this was expected to be provided as a textual duration 🙈

Yeah I think ensuring that we have a reasonable max that's within our expected bitcoin block time would be wise

aldur: So I implanted it because I read this too late (4e85213). I am not 100% sure we want to have that cap though.

aldur:
Thanks both for the comments! @cylewitruk PTAL

signer/src/config/serialization.rs
```diff
@@ -33,6 +35,39 @@ where
...
+pub fn bitcoin_processing_delay_deserializer<'de, D>(
+    deserializer: D,
+) -> Result<std::time::Duration, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let delay = duration_seconds_deserializer(deserializer)?;
+    let delay_secs = delay.as_secs();
+    match delay_secs > MAX_BITCOIN_PROCESSING_DELAY_S {
+        true => Err(serde::de::Error::custom(
```
cylewitruk, line 61: So we're generally not considering "valid" invalid values as errors during deserialization, so I think you could get rid of this custom serializer and just use the `duration_seconds_deserializer`.

There's `impl Validatable for SignerConfig {` on line 245 where the validation should be plugged in. I think that might make this all a little cleaner :)

signer/src/config/default.toml
cylewitruk: Could you just create a new ticket for this in the milestone as a "don't forget" and link it in a `# TODO: xxx` here?

signer/src/transaction_coordinator.rs
cylewitruk: That's what chatgpt is for :rofl:

signer/src/config/default.toml
aldur: Done in https://github.com/stacks-network/sbtc/issues/715.

signer/src/config/mod.rs
cylewitruk: Yeah I'm not sure on the number, but I can't think of a situation where we would ever want to exceed 15-30 secs, so this is probably okay as a failsafe just to make sure that the signer can never wait _longer_ than a typical bitcoin block (i.e. waiting into the next block).

signer/src/config/mod.rs
```diff
@@ -23,6 +24,8 @@ use crate::stacks::wallet::SignerWallet;
 mod error;
 mod serialization;
 
+pub const MAX_BITCOIN_PROCESSING_DELAY_S: u64 = 300;
```
djordon, line 27: It looks like the build is failing because this constant is missing docs (we have a lint to make sure we have documentation for everything. Would you mind addressing this in another PR? Also, would you mind renaming this to `MAX_BITCOIN_PROCESSING_DELAY_SECS` or something similar?

djordon (commented):
In general, we do not merge a PR unless all tests and checks have successfully passed. In the future, please try to make sure CI is green before merging. This might be one of those things that is worth encoding at the repo level, just like resolving all discussions.

signer/src/config/mod.rs
aldur: Ops, sorry! I have seen failing tests locally and attributed the CI failure to that. Let me fix that right away.

PR #718, opened 2024-10-28
feat: Fail fast, run lints before integration tests, stop previous jobs

aldur:
## Description

Small optimization to the CI: 
- Lints are much faster than integration tests. Run them before.
- As per popular demand, I have added `cargo fmt --check` to the `lint` Makefile action.
- Limit concurrency to 1 workflow per branch (new pushes will stop previous jobs).


cylewitruk (approved):
This looks like it addresses the raised issues and if it happens to break something then it's a quick fix to revert. Approve from me.

PR #720, opened 2024-10-28
feat: add optional git hooks

aldur:
## Description

Add _optional_ Git hooks for those who don't like to push code only to find it rejected by the CI linter.

## Testing Information

Read the new `README.md` section. After installing the hooks, try making a modification that `cargo fmt` won't like. Try to commit it, or try running `.git/hooks/pre-commit-make-lint.sh`.

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,3 +181,13 @@ run-signer:
 	RUST_LOG=info SIGNER_SIGNER__DB_ENDPOINT="postgres://devenv:devenv@localhost:$$POSTGRES_PORT/signer" cargo run --bin signer -- -c ./signer/src/config/default.toml --migrate-db
 
 .PHONY: run-signer
+
+# Git hooks
+# ----------------------------------------------------
+
+install-git-hooks:
+	mkdir -p .git/hooks
+	cp -r ./devenv/hooks/pre-commit-make-lint.sh .git/hooks/
+
+.PHONY: install-git-hooks
+
--- a/Makefile
+++ b/Makefile
@@ -187,7 +187,7 @@ run-signer:
 
 install-git-hooks:
 	mkdir -p .git/hooks
-	cp -r ./devenv/hooks/pre-commit-make-lint.sh .git/hooks/
+	ln -s ../../devenv/hooks/pre-commit-make-lint.sh .git/hooks/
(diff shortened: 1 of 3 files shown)


cylewitruk:
I personally like this as a starting point. I did a deep-dive on this topic a while back for [clarity-wasm](https://github.com/stacks-network/clarity-wasm/pull/101), but I never followed through on it. There might be some inspiration there using `commitlint` (not necessarily `commitizen`, that's much more on the convenience-tooling side).

Makefile
```diff
@@ -181,3 +181,13 @@ run-signer:
 	RUST_LOG=info SIGNER_SIGNER__DB_ENDPOINT="postgres://devenv:devenv@localhost:$$POSTGRES_PORT/signer" cargo run --bin signer -- -c ./signer/src/config/default.toml --migrate-db
 
 .PHONY: run-signer
+
+# Git hooks
+# ----------------------------------------------------
+
+install-git-hooks:
+	mkdir -p .git/hooks
+	ln -s ../../devenv/hooks/pre-commit-make-lint.sh .git/hooks/
```
cylewitruk, line 190: Why is this `../../devenv`, shouldn't it just be `./devenv/etc` if the make commands are run from the repo root?

aldur: The link needs to be relative to the target (this has bit me before :))

aldur:
> I personally like this as a starting point. I did a deep-dive on this topic a while back for [clarity-wasm](https://github.com/stacks-network/clarity-wasm/pull/101), but I never followed through on it. There might be some inspiration there using `commitlint` (not necessarily `commitizen`, that's much more on the convenience-tooling side).

Thanks! Unless there's a strong feeling, I'd rather keep things simple and re-use our Makefile rather than introduce additional tools -- I have also used `precommit` in the past, but our Makefile is already there doing the heavy lifting.

cylewitruk:
@aldur yeah it was just "fyi" -- i just have the one question before i approve this :100:

Makefile
cylewitruk: Gotcha :+1:

cylewitruk (approved):

PR #721, opened 2024-10-28
feat: Add latest canonical bitcoin blocks db call

AshtonStephens:
## Description

Part 1/3: #380 

This adds two database access functions that let you get the the ancestors of the given bitcoin block hash, and separately one that gives you the latest canonical bitcoin blocks.

## Changes

1. Adds database access function
2. Adds relevant integration tests

## Testing Information

Integration tests and unit tests.

## Checklist:

### DIFF
--- a/signer/src/storage/mod.rs
+++ b/signer/src/storage/mod.rs
@@ -43,6 +43,19 @@ pub trait DbRead {
         &self,
     ) -> impl Future<Output = Result<Option<model::BitcoinBlockHash>, Error>> + Send;
 
+    /// Get the latest canonical bitcoin blocks.
+    fn get_latest_canonical_bitcoin_blocks(
+        &self,
+        limit: u16,
+    ) -> impl Future<Output = Result<Vec<model::BitcoinBlock>, Error>> + Send;
+
+    /// Get the ancestral bitcoin blocks of the given block hash.
+    fn get_ancestral_bitcoin_blocks(
+        &self,
+        block_hash: &model::BitcoinBlockHash,
+        limit: u16,
+    ) -> impl Future<Output = Result<Vec<model::BitcoinBlock>, Error>> + Send;
+
     /// Get the stacks chain tip, defined as the highest stacks block
     /// confirmed by the bitcoin chain tip.
     fn get_stacks_chain_tip(
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -490,6 +490,56 @@ impl super::DbRead for PgStore {
         .map_err(Error::SqlxQuery)
     }
 
+    /// Get the ancestral bitcoin blocks for the given block hash.
+    async fn get_ancestral_bitcoin_blocks(
+        &self,
+        block_hash: &model::BitcoinBlockHash,
+        limit: u16,
+    ) -> Result<Vec<model::BitcoinBlock>, Error> {
+        sqlx::query_as::<_, model::BitcoinBlock>(
+            r#"
+            WITH RECURSIVE context_window AS (
+                -- Anchor member: Initialize the recursion with the chain tip
+                SELECT block_hash, block_height, parent_hash, confirms, 1 AS depth
+                FROM sbtc_signer.bitcoin_blocks
+                WHERE block_hash = $1
+
+                UNION ALL
+
+                -- Recursive member: Fetch the parent block using the last block's parent_hash
+                SELECT parent.block_hash, parent.block_height, parent.parent_hash,
+                        parent.confirms, last.depth + 1
+                FROM sbtc_signer.bitcoin_blocks parent
+                JOIN context_window last ON parent.block_hash = last.parent_hash
+                WHERE last.depth < $2
+            )
+            SELECT
+                block_hash
+                , block_height
+                , parent_hash
+                , confirms
+            FROM context_window
+            "#,
+        )
+        .bind(block_hash)
+        .bind(i32::from(limit))
+        .fetch_all(&self.0)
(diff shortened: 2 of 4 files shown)


signer/src/storage/mod.rs
```diff
@@ -43,6 +43,19 @@ pub trait DbRead {
         &self,
     ) -> impl Future<Output = Result<Option<model::BitcoinBlockHash>, Error>> + Send;
 
+    /// Get the latest canonical bitcoin blocks.
+    fn get_latest_canonical_bitcoin_blocks(
+        &self,
+        limit: u16,
```
djordon, line 49: You probably want this function to take the bitcoin chain tip like most of the functions here.

signer/src/storage/postgres.rs
```diff
@@ -490,6 +490,56 @@ impl super::DbRead for PgStore {
...
+                -- Anchor member: Initialize the recursion with the chain tip
+                SELECT block_hash, block_height, parent_hash, confirms, 1 AS depth
+                FROM sbtc_signer.bitcoin_blocks
+                WHERE block_hash = $1
+
+                UNION ALL
+
+                -- Recursive member: Fetch the parent block using the last block's parent_hash
+                SELECT parent.block_hash, parent.block_height, parent.parent_hash,
+                        parent.confirms, last.depth + 1
```
djordon, line 511: nit: this should follow the same style like we do for our other queries here, like:
```sql
SELECT 
    block_hash
  , parent_hash
  , confirms
  , 1 AS depth
FROM ...
```

signer/tests/integration/postgres.rs
```diff
@@ -1200,6 +1200,91 @@ async fn block_in_canonical_bitcoin_blockchain_in_other_block_chain() {
...
+    // Ensure that the latest canonical blocks are in descending order.
+    latest_canonical_bitcoin_blocks
+        .windows(2)
+        .for_each(|window| {
+            if let [child, parent] = window {
+                assert_eq!(child.block_height, parent.block_height + 1);
+                assert_eq!(child.parent_hash, parent.block_hash);
+            } else {
+                panic!("Expected a window of size 2");
+            }
```
djordon, line 1282: You might be able to simplify this with something like

```rust
let [child, parent] = window.first_chunk().unwrap();
assert_eq!(child.block_height, parent.block_height + 1);
assert_eq!(child.parent_hash, parent.block_hash);
```

signer/src/storage/mod.rs
```diff
@@ -43,6 +43,19 @@ pub trait DbRead {
         &self,
     ) -> impl Future<Output = Result<Option<model::BitcoinBlockHash>, Error>> + Send;
 
+    /// Get the latest canonical bitcoin blocks.
+    fn get_latest_canonical_bitcoin_blocks(
```
djordon, line 47: Are you sure that you need these functions? I was under the impression that much of the #380 filtering could in a query.

djordon (commented):
Nice, very simple PR. My main question is whether we really need these queries.

signer/src/storage/postgres.rs
AshtonStephens: Will do

signer/src/storage/mod.rs
AshtonStephens: This is the query that will let us get the last 11 blocks and then we can get the average time from them, how would we not need this?

AshtonStephens: What query / filtering are you talking about?

djordon: Instead of fetching the the blocks to compute the MTP associated with a deposit request, we can can write a query that returns the MTP. No need to fetch the blocks. One such postgres function is `percentile_disc`.

> This is the query that will let us get the last 11 blocks and then we can get the average time from them, how would we not need this?

So if we want a query that filters requests where the current timestamp is greater than the `MTP + 512 * X`, where `X` is associated with the locktime, we could do that filtering in the same query using `percentile_disc`.

AshtonStephens: I'd rather have the query have a larger logical unit than be simply responsible for a single thing.

Here's what the final function should look like in the coordinator:

```rust
    async fn get_pending_requests(
        &mut self,
        aggregate_key: &PublicKey,
    ) -> Result<Option<utxo::SbtcRequests>, Error> {
        
        // ...
        
        let maybe_median_block_time = self.context
            .get_storage()
            .get_latest_canonical_bitcoin_blocks(6)
            .await?
            .last()
            .map(|block| block.time_mined);

        let pending_deposit_requests = self
            .context
            .get_storage()
            .get_pending_accepted_deposit_requests(bitcoin_chain_tip, context_window, threshold)
            .await?
            .iter()
            .filter(|&request| {
                match request.lock_time {
                    LockTime::BlockHeight(height) => {
                        // ... Assess based on block height
                    }
                    LockTime::Time(time) => {
                        // ... Use median
                    }
                }
            });
            
        // ...
            
    }
```

djordon: I'm think this approach is going to be way more complex than the SQL based approach. If you don't compute the MTP in the query, you'll need to do so with in code for each deposit request. Doing it all in SQL might be too complex as well, so maybe somewhere in between would be best.

signer/src/storage/postgres.rs
cylewitruk: Just FYI incoming in #710 is a table-valued function that will let you remove the explicit recursive query.

AshtonStephens:
Closing because we're focusing on SQL query based backend implementations

Issue #578, opened 2024-09-27
[Feature]: signatures on WSTS packets need to be verified before feeding them into the state machines


xoloki:
# Feature - Verify WSTS packet signatures 

## 1. Description
Applications which use wsts state machines must verify the signatures on the packets before processing them. 

### 1.1 Context & Purpose
All wsts network packets are signed, to guarantee that they come from the purported source, and have not been tampered with. So they must be verified before processing. 

But wsts applications typically run both coordinator and signer state machines, so it’s better to verify them outside the state machines themselves. Also, coordinator selection is external to the state machines. 

## 2. Technical Details:
Call ` Packet::verify` with the current signer and coordinator public keys after receiving packets, before feeding them into the machines. Bad packets should be dropped. 

### 2.1 Acceptance Criteria:
   - [ ] all packets are verified 
   - [ ] tests added

djordon:
I think that this is a duplicate of https://github.com/stacks-network/sbtc/issues/296.

xoloki:
> I think that this is a duplicate of #296.

I don’t think so. 296 is about signing bitcoin transactions AFAICT.

djordon:
Oh yeah you are right. It is not a duplicate.

djordon:
Oh man, we do here https://github.com/stacks-network/sbtc/blob/53d43bb9f1bd7797626779b16127e09d3d453d87/signer/src/transaction_signer.rs#L219-L223 but do not here https://github.com/stacks-network/sbtc/blob/c8fa6c52341725820c005d6e7e72486c14066a87/signer/src/transaction_coordinator.rs#L332-L377

Okay, that fell through the cracks. I think the best place to do this is here:
https://github.com/stacks-network/sbtc/blob/53d43bb9f1bd7797626779b16127e09d3d453d87/signer/src/network/mod.rs#L95-L113
That way we won't need to track everything down all the time. Hmmm, this might be straitforward.

cylewitruk:
> I wonder if we can skip this entirely, since the networking layer also does this.

It does not at present -- that's where we had the discussion about _which_ messages should be guarded and not (since a pending/joining node may still need to be able to communicate with the network in some capacity).

xoloki:
> We could also just implement it for wsts messages in the networking layer (for now).

I've been looking over this, and while we can verify messages at the network layer, it won't solve the problem this issue was created to solve.  

[...]

PR #723, opened 2024-10-28
check that wsts messages are signed by the specified signer_id

xoloki:
## Description
`sBTC` messages contain a signature and a public key, in addition to the payload.  Verifying the message shows that it has not been tampered with.

`WSTS` message payloads either contain a `signer_id`, or are coordinator messages.  So in addition to verifying the signature, we also need to show that the public key used to sign was the correct public key for that signer (or the coordinator).

Closes: #578 

## Changes
Call `verify` on the `sBTC` message, then check that the message was signed by the correct public key.

## Testing Information
All unit and integration tests still pass.

## Checklist:

- [ x] I have performed a self-review of my code
- [ x] My changes generate no new warnings
- [ x] New and existing unit tests pass locally with my changes
- [ x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/signer/src/testing/transaction_signer.rs
+++ b/signer/src/testing/transaction_signer.rs
@@ -562,6 +562,18 @@ where
             .expect("storage error")
             .expect("no chain tip");
 
+        // now that we have a chain tip, get the real coordinator
+        let coordinator_public_key =
+            crate::transaction_coordinator::coordinator_public_key(&bitcoin_chain_tip, &signer_set)
+                .unwrap();
+        let coordinator_signer_info = signer_info
+            .iter()
+            .find(|signer| {
+                PublicKey::from_private_key(&signer.signer_private_key) == coordinator_public_key
+            })
+            .unwrap()
+            .clone();
+
         run_dkg_and_store_results_for_signers(
             &signer_info,
             &bitcoin_chain_tip,
--- a/signer/src/testing/transaction_signer.rs
+++ b/signer/src/testing/transaction_signer.rs
@@ -564,7 +564,7 @@ where
 
         // now that we have a chain tip, get the real coordinator
         let coordinator_public_key =
-            crate::transaction_coordinator::coordinator_public_key(&bitcoin_chain_tip, &signer_set)
+            crate::transaction_coordinator::coordinator_public_key(&bitcoin_chain_tip, signer_set)
                 .unwrap();
         let coordinator_signer_info = signer_info
             .iter()
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -876,11 +876,11 @@ where
             }
             None => {
                 tracing::warn!(
-                            ?packet.msg,
-                            reason = "no public key for signer",
-                signer_id = %signer_id,
-                            "ignoring packet"
-                        );
+                    ?packet.msg,
+                    reason = "no public key for signer",
+                    %signer_id,
+                    "ignoring packet"
+                );
                 false
             }
             _ => true,
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -825,6 +825,32 @@ where
 
             let public_keys = &coordinator_state_machine.get_config().signer_public_keys;
             let public_key_point = p256k1::point::Point::from(msg_public_key);
+
+            macro_rules! check_signer_public_key {
+                ($message:ident) => {
+                    match &public_keys.get(&$message.signer_id) {
+                        Some(signer_public_key) if &public_key_point != *signer_public_key => {
+                            tracing::warn!(
+                                ?packet,
+                                reason = "message was signed by the wrong signer",
+                                "ignoring packet"
+                            );
+                            continue;
+                        }
+                        None => {
+                            tracing::warn!(
+                                ?packet,
+                                reason =
+                                    format!("no public key for signer {}", &$message.signer_id),
+                                "ignoring packet"
+                            );
+                            continue;
+                        }
+                        _ => (),
+                    }
+                };
+            }
+
             // check that messages were signed by correct key
             match &packet.msg {
                 wsts::net::Message::DkgBegin(_)
@@ -843,54 +869,19 @@ where
                 }
 
                 wsts::net::Message::DkgPublicShares(dkg_public_shares) => {
-                    if &public_key_point != &public_keys[&dkg_public_shares.signer_id] {
-                        tracing::warn!(
-                            ?packet,
-                            reason = "message was signed by the wrong signer",
-                            "ignoring packet"
-                        );
-                        continue;
-                    }
+                    check_signer_public_key!(dkg_public_shares)
                 }
                 wsts::net::Message::DkgPrivateShares(dkg_private_shares) => {
-                    if &public_key_point != &public_keys[&dkg_private_shares.signer_id] {
-                        tracing::warn!(
-                            ?packet,
-                            reason = "message was signed by the wrong signer",
-                            "ignoring packet"
-                        );
-                        continue;
-                    }
+                    check_signer_public_key!(dkg_private_shares)
                 }
                 wsts::net::Message::DkgEnd(dkg_end) => {
-                    if &public_key_point != &public_keys[&dkg_end.signer_id] {
-                        tracing::warn!(
-                            ?packet,
-                            reason = "message was signed by the wrong signer",
-                            "ignoring packet"
-                        );
-                        continue;
-                    }
+                    check_signer_public_key!(dkg_end)
                 }
                 wsts::net::Message::NonceResponse(nonce_response) => {
-                    if &public_key_point != &public_keys[&nonce_response.signer_id] {
-                        tracing::warn!(
-                            ?packet,
-                            reason = "message was signed by the wrong signer",
-                            "ignoring packet"
-                        );
-                        continue;
-                    }
+                    check_signer_public_key!(nonce_response)
                 }
                 wsts::net::Message::SignatureShareResponse(sig_share_response) => {
-                    if &public_key_point != &public_keys[&sig_share_response.signer_id] {
-                        tracing::warn!(
-                            ?packet,
-                            reason = "message was signed by the wrong signer",
-                            "ignoring packet"
-                        );
-                        continue;
-                    }
+                    check_signer_public_key!(sig_share_response)
                 }
             }
 
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -787,6 +787,13 @@ where
         coordinator_state_machine: &mut CoordinatorStateMachine,
         txid: bitcoin::Txid,
     ) -> Result<WstsOperationResult, Error> {
+        // this assumes that the signer set doesn't change for the duration of this call,
+        // but we're already assuming that the bitcoin chain tip doesn't change
+        // alternately we could hit the DB every time we get a new message
+        let (_, signer_set) = self
+            .get_signer_set_and_aggregate_key(bitcoin_chain_tip)
+            .await?;
+
         loop {
             // Let's get the next message from the network or the
             // TxSignerEventLoop.
@@ -794,6 +801,11 @@ where
 
             if &msg.bitcoin_chain_tip != bitcoin_chain_tip {
                 tracing::warn!(?msg, "concurrent WSTS activity observed");
+                continue;
+            }
+
+            if !msg.verify() {
+                tracing::warn!(?msg, "invalid signature");
                 continue;
             }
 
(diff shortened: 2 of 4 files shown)


xoloki:
@djordon the problem with the `tap_tweak_equivalence` test` is that `secp256k1::PublicKey::add_exp_tweak` does not actually produce a `BIP-341` tweaked public key.  It simply adds `tweak * G` to the underlying curve point.  This test used to  work because that's what `WSTS` used to do as well.  

I can't see any way to get a proper tweaked public key using the `secp256k1` crate alone, since it never defines the `lift_x` function from `BIP-340`.  The rust `bitcoin` crate will do it if you call `Address::p2tr`, but of course that gives you an address not a pub key, or a curve point.  You can however get a `ScriptBuf` from it.

djordon:
> @djordon the problem with the `tap_tweak_equivalence` test is that `secp256k1::PublicKey::add_exp_tweak` does not actually produce a `BIP-341` tweaked public key. It simply adds `tweak * G` to the underlying curve point. This test used to work because that's what `WSTS` used to do as well.
> 
> I can't see any way to get a proper tweaked public key using the `secp256k1` crate alone, since it never defines the `lift_x` function from `BIP-340`. The rust `bitcoin` crate will do it if you call `Address::p2tr`, but of course that gives you an address not a pub key, or a curve point. You can however get a `ScriptBuf` from it.

Oh this test has been fixed on main already https://github.com/stacks-network/sbtc/blob/fdfec1c646f3e30bbf1676ef08f41f1ada6731ae/signer/src/keys.rs#L605-L622

You should rebase these commits off of `main` and target this branch to be merged there. Also, I think you can use `secp256k1::PublicKey::from_x_only_public_key` to do a "proper" `lift_x`.

signer/src/transaction_signer.rs
```diff
@@ -517,10 +517,19 @@ where
         &mut self,
         msg: &message::WstsMessage,
         bitcoin_chain_tip: &model::BitcoinBlockHash,
+        msg_public_key: PublicKey,
     ) -> Result<(), Error> {
         tracing::info!("handling message");
+        let chain_tip_report = self
+            .inspect_msg_chain_tip(msg_public_key, bitcoin_chain_tip)
+            .await?;
```
djordon, line 525: Why not just pass in the report from `handle_signer_message`?

djordon (commented):
Nice, had some comments and suggestions. Before digging in we should probably rebase off of main.

signer/src/testing/transaction_signer.rs
```diff
@@ -666,25 +667,47 @@ where
         let mut hasher = sha2::Sha256::new();
         hasher.update("sign here please");
         let msg: [u8; 32] = hasher.finalize().into(); // TODO(296): Compute proper sighash from transaction
+        let signature_type = SignatureType::Schnorr;
+        // XXX Taproot not working here
+        //let signature_type = SignatureType::Taproot(Some(msg));
```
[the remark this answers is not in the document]
xoloki: That code doesn't have anything to do with the focus of this change, so I didn't port it over to the new branch.

signer/src/testing/transaction_signer.rs
```diff
@@ -666,25 +667,47 @@ where
...
 
         coordinator
             .request_sign_transaction(bitcoin_chain_tip, tx, aggregate_key)
             .await;
 
         let signature = coordinator
-            .run_signing_round(bitcoin_chain_tip, txid, &msg, SignatureType::Schnorr)
+            .run_signing_round(bitcoin_chain_tip, txid, &msg, signature_type.clone())
```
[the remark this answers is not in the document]
xoloki: Done, in the current PR branch (`simple-signature-type`).

xoloki:
> Interestingly, `transaction_signer::tests::should_be_able_to_participate_in_dkg` fails on that branch because the chain tip report says that `DkgBegin` isn't coming from the coordinator.

It appears that `TestEnvironment::assert_should_be_able_to_participate_in_dkg` chooses the first signer in the generated set as the coordinator, rather than what the chain tip would imply @djordon:
```
        let signer_info = testing::wsts::generate_signer_info(&mut rng, self.num_signers);
	let coordinator_signer_info = signer_info.first().unwrap().clone();
```
The code is identical on `fix-bitcoin-tx-signing` and `main`, and the error is on both `578-verify` branches.  

I'm trying now to run the tail of that fn with the correct coordinator.

signer/src/testing/transaction_signer.rs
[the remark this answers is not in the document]
xoloki: I force pushed the new branch to this one, so we didn't have to switch to a different PR.  Now we're based on main with just a couple of commits.

signer/src/transaction_coordinator.rs
```diff
@@ -747,6 +758,82 @@ where
...
+                            ?packet,
+                            reason = "got coordinator message from sender who is not coordinator",
+                            "ignoring packet"
+                        );
+                        continue;
+                    }
+                }
+
+                wsts::net::Message::DkgPublicShares(dkg_public_shares) => {
+                    if &public_key_point != &public_keys[&dkg_public_shares.signer_id] {
```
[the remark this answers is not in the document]
xoloki: Done.  It got a little verbose so I wrote my first macro.

signer/src/transaction_coordinator.rs
```diff
@@ -747,6 +758,82 @@ where
...
 
+            let msg_public_key = msg.signer_pub_key.clone();
+
+            let sender_is_coordinator =
+                given_key_is_coordinator(msg_public_key, bitcoin_chain_tip, &signer_set);
+
+            let public_keys = &coordinator_state_machine.get_config().signer_public_keys;
+            let public_key_point = p256k1::point::Point::from(msg_public_key);
+            // check that messages were signed by correct key
+            match &packet.msg {
```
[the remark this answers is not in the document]
xoloki: Moved match into new `fn authenticate_message`

signer/src/transaction_coordinator.rs
```diff
@@ -747,6 +758,82 @@ where
...
+                wsts::net::Message::DkgBegin(_)
+                | wsts::net::Message::DkgPrivateBegin(_)
+                | wsts::net::Message::DkgEndBegin(_)
+                | wsts::net::Message::NonceRequest(_)
+                | wsts::net::Message::SignatureShareRequest(_) => {
+                    if !sender_is_coordinator {
```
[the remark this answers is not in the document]
xoloki: I'd argue that for completeness we should go ahead and do the check.  We might want to use the `fn` elsewhere.

signer/src/transaction_coordinator.rs
```diff
@@ -728,6 +727,13 @@ where
         coordinator_state_machine: &mut CoordinatorStateMachine,
         txid: bitcoin::Txid,
     ) -> Result<WstsOperationResult, Error> {
+        // this assumes that the signer set doesn't change for the duration of this call,
+        // but we're already assuming that the bitcoin chain tip doesn't change
+        // alternately we could hit the DB every time we get a new message
+        let (_, signer_set) = self
+            .get_signer_set_and_aggregate_key(bitcoin_chain_tip)
+            .await?;
```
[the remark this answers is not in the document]
xoloki: Yeah it's a reasonable assumption in the context, and if it's false it shouldn't cause any problems.

We only call this `fn` twice (`coordinate_signing_round` and `coordinate_dkg`); one of those already creates the `signer_set` and the other doesn't.  I don't see much point to passing it down, but I can if you'd like.

signer/src/transaction_signer.rs
xoloki: Done

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
djordon: Na, this is fine.

signer/src/transaction_coordinator.rs
```diff
@@ -827,6 +857,71 @@ where
...
+                            ?packet,
+                            reason = "message was signed by the wrong signer",
+                            "ignoring packet"
+                        );
+                        false
+                    }
+                    None => {
+                        tracing::warn!(
+                            ?packet,
+                            reason = format!("no public key for signer {}", &$message.signer_id),
```
djordon, line 880: I think we should put the `signer_id` as a separate field in the log, so it would more like:
```rust
tracing::warn!(
    ?packet,
    reason = "no public key for signer",
    signer_id = %$message.signer_id,
    ...
```

signer/src/transaction_coordinator.rs
```diff
@@ -827,6 +857,71 @@ where
...
 
+    fn authenticate_message(
+        packet: &wsts::net::Packet,
+        public_keys: &hashbrown::HashMap<u32, p256k1::point::Point>,
+        public_key_point: p256k1::point::Point,
+        sender_is_coordinator: bool,
+    ) -> bool {
+        macro_rules! check_signer_public_key {
+            ($message:ident) => {
+                match &public_keys.get(&$message.signer_id) {
```
djordon, line 868: Congrats on the macro 🎉. It's pretty straightforward but I don't think we should add it. In general, I think we should try hard not to add them, since they are typically harder for people to understand, debug, and so on. Plus VS code barfs when looking at them. So we've tended to not use them, and I'd like to continue that trend unless the payoff clearly exceeds the cost.

Now, this macro is a pretty simple one, and I think we can get by with making it a closure (or proper) function that takes in the `signer_id` (without the message) and goes about it's day.

signer/src/transaction_coordinator.rs
```diff
@@ -806,6 +818,24 @@ where
...
+            let public_keys = &coordinator_state_machine.get_config().signer_public_keys;
+            let public_key_point = p256k1::point::Point::from(msg_public_key);
+
+            // check that messages were signed by correct key
+            if !Self::authenticate_message(
+                &packet,
+                public_keys,
+                public_key_point,
+                sender_is_coordinator,
+            ) {
```
djordon, line 835: I would break this into multiple lines like so:
```rust
let is_authenticated = Self::authenticate_message(
    &packet,
    public_keys,
    public_key_point,
    sender_is_coordinator,
);

if !is_authenticated {
    ...
```
I find it a bit easier to read

signer/src/transaction_signer.rs
```diff
@@ -594,13 +606,38 @@ where
...
+            wsts::net::Message::DkgPrivateBegin(_) => {
+                if !chain_tip_report.sender_is_coordinator {
+                    return Err(Error::NotChainTipCoordinator);
+                }
+
+                self.relay_message(msg.txid, &msg.inner, bitcoin_chain_tip)
+                    .await?;
+            }
+            wsts::net::Message::DkgPublicShares(dkg_public_shares) => {
+                let public_keys = &self.wsts_state_machines[&msg.txid].public_keys;
```
djordon, line 618: I think that we should move away from using the index-style access, since we can panic. In this case we panic if the sender uses a transaction ID that we don't know about (purposefully or otherwise). But more generally we shy away from it and use the non-panic APIs instead.

signer/src/transaction_coordinator.rs
```diff
@@ -827,6 +857,71 @@ where
...
+                        tracing::warn!(
+                            ?packet,
+                            reason = "message was signed by the wrong signer",
+                            "ignoring packet"
+                        );
+                        false
+                    }
+                    None => {
+                        tracing::warn!(
+                            ?packet,
```
djordon, line 879: Any thoughts on just having the `packet.msg`, and skip logging the signature (since presumably we have already checked that once we've gotten here).

djordon (commented):
Left a few more comments. Looks good overall.

signer/src/transaction_coordinator.rs
```diff
@@ -827,6 +857,71 @@ where
         }
     }
 
+    fn authenticate_message(
```
matteojug, line 860: We could refactor this to be used in both the coordinator and the signer, so that we have a single shared fn doing the authentication logic. WDYT?

signer/src/transaction_coordinator.rs
xoloki: Done

signer/src/transaction_coordinator.rs
xoloki: Closure seemed best, to avoid passing in packet/etc.  Fixed.

signer/src/transaction_coordinator.rs
xoloki: Done

signer/src/transaction_signer.rs
xoloki: Done

signer/src/transaction_coordinator.rs
```diff
@@ -827,6 +859,66 @@ where
...
+        packet: &wsts::net::Packet,
+        public_keys: &hashbrown::HashMap<u32, p256k1::point::Point>,
+        public_key_point: p256k1::point::Point,
+        sender_is_coordinator: bool,
+    ) -> bool {
+        let check_signer_public_key = |signer_id| match &public_keys.get(&signer_id) {
```
djordon, line 868: You don't need the reference before `public_keys` here.

signer/src/transaction_coordinator.rs
```diff
@@ -827,6 +859,66 @@ where
...
+                );
+                false
+            }
+            None => {
+                tracing::warn!(
+                            ?packet.msg,
+                            reason = "no public key for signer",
+                signer_id = %signer_id,
+                            "ignoring packet"
+                        );
```
djordon, line 883: nit: rustfmt does not catch the weird spacing here. Also, this could be just
```rust
tracing::warn!(
    ?packet.msg,
    reason = "no public key for signer",
    %signer_id,
    "ignoring packet"
);
```

signer/src/transaction_signer.rs
```diff
@@ -579,10 +584,17 @@ where
...
+        msg_public_key: PublicKey,
+        chain_tip_report: &MsgChainTipReport,
     ) -> Result<(), Error> {
         tracing::info!("handling message");
+
         match &msg.inner {
             wsts::net::Message::DkgBegin(_) => {
+                if !chain_tip_report.sender_is_coordinator {
+                    return Err(Error::NotChainTipCoordinator);
+                }
```
djordon, line 596: For all of these `chain_tip_report.sender_is_coordinator` checks, I imagine that this will happen from time to time, with the consequence being that we sometimes receive an error log message that we have to ignore. Instead I think we should log a warning and return `Ok(())`.

djordon (commented):
This basically looks good!

signer/src/transaction_coordinator.rs
xoloki: Agreed @djordon , it's a nice idea but it might be more effort than it's worth.  Also we do things in the various match branches, so we'd have to have another match statement if we moved this code into a shared fn.

signer/src/transaction_coordinator.rs
xoloki: Done

signer/src/transaction_coordinator.rs
xoloki: Done

signer/src/transaction_signer.rs
xoloki: Done

djordon (approved):
Looks good ✅!

PR #724, opened 2024-10-28
feat: Add time_mined to bitcoin blocks in database

AshtonStephens:
## Description

Part 2/3 of addressing: #380

## Changes

1. Adds the `time_mined` field to the signer database.
2. Fixes an integration test that relies on the order of an unordered set.

## Testing Information

Tested with existing integration tests

## Checklist:

### DIFF
--- a/signer/migrations/0003__create_tables.sql
+++ b/signer/migrations/0003__create_tables.sql
@@ -14,6 +14,7 @@ CREATE TABLE sbtc_signer.bitcoin_blocks (
     block_height BIGINT NOT NULL,
     parent_hash BYTEA NOT NULL,
     confirms BYTEA[] NOT NULL,
+    time_mined INTEGER NOT NULL,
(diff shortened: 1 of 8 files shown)


signer/migrations/0003__create_tables.sql
```diff
@@ -14,6 +14,7 @@ CREATE TABLE sbtc_signer.bitcoin_blocks (
     block_height BIGINT NOT NULL,
     parent_hash BYTEA NOT NULL,
     confirms BYTEA[] NOT NULL,
+    time_mined INTEGER NOT NULL,
```
djordon, line 17: We want this to be a BIGINT

AshtonStephens:
Closing because we're focusing on SQL query based backend implementations, and this PR was based on the approach that was started in #721 

Additionally, we've chosen not to support relative time lock times for this release, so this timestamp is further unnecessary.

I just opened #742 to track this.

PR #727, opened 2024-10-29
feat: add wsts protobufs

djordon:
## Description

This is work done on the path to https://github.com/stacks-network/sbtc/issues/412.

## Changes

* Add protobuf types for WSTS structs.

## Testing Information

This stuff is unused at the moment

## Checklist:


cylewitruk (approved):
Only gave it a quick glance since this isn't used atm but looks good in general.

For anyone else looking at this, we decided on a call that it's better to get this stuff into main instead of having it rot in a branch somewhere, as the general design goal over time is to switch to using protobuf encoding for network messaging.

PR #730, opened 2024-10-29
chore: update emily integration test

matteojug:
## Description

Small leftover from https://github.com/stacks-network/sbtc/issues/559

## Changes

Fix Emily integration test and remove `ignore` guard now that Emily run in the integration test compose. Also sneaked in a small typo fix.

## Testing Information

Test runs and passes.

## Checklist:

### DIFF
--- a/signer/tests/integration/emily.rs
+++ b/signer/tests/integration/emily.rs
@@ -17,13 +17,18 @@ use bitcoincore_rpc_json::Utxo;
 use blockstack_lib::chainstate::nakamoto::NakamotoBlock;
 use blockstack_lib::chainstate::nakamoto::NakamotoBlockHeader;
 use blockstack_lib::net::api::getpoxinfo::RPCPoxInfoData;
+use blockstack_lib::net::api::getsortition::SortitionInfo;
 use blockstack_lib::net::api::gettenureinfo::RPCGetTenureInfo;
+use clarity::types::chainstate::BurnchainHeaderHash;
 use clarity::types::chainstate::ConsensusHash;
+use clarity::types::chainstate::SortitionId;
 use clarity::types::chainstate::StacksBlockId;
 use emily_client::apis::deposit_api;
+use emily_client::apis::testing_api::wipe_databases;
 use emily_client::models::CreateDepositRequestBody;
 use sbtc::testing::regtest::Recipient;
 use sha2::Digest as _;
+use signer::bitcoin::rpc::BitcoinTxInfo;
 use signer::bitcoin::rpc::GetTxResponse;
 use signer::block_observer;
 use signer::context::Context;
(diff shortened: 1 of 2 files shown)


signer/tests/integration/emily.rs
```diff
@@ -372,6 +406,26 @@ async fn deposit_e2e() {
                 .expect_get_account()
                 .once()
                 .returning(|_| Box::pin(async move { Err(Error::InvalidStacksResponse("mock")) }));
+
+            client.expect_get_sortition_info().returning(move |_| {
+                let response = Ok(SortitionInfo {
```
cylewitruk, line 411: suggestion: not sure if this might be a little cleaner (smaller) if you faked the `SortitionInfo` as a `mut` and set `burn_block_hash` (and `was_sortition`?) separately? I feel like the newing-up/defaults detract a little bit from the readability of what's important.

signer/tests/integration/emily.rs
```diff
@@ -297,10 +301,11 @@ async fn deposit_e2e() {
                 });
 
             // Return the deposit tx
-            client.expect_get_tx().returning(move |txid| {
-                let res = if *txid == deposit_tx.compute_txid() {
+            let deposit_tx_ = deposit_tx.clone();
+            client.expect_get_tx().once().returning(move |txid| {
```
cylewitruk, line 305: nit: spread call-chains over newlines (like you did below)

signer/tests/integration/emily.rs
```diff
@@ -443,42 +497,12 @@ async fn deposit_e2e() {
...
-        .unwrap()
-        .is_none());
-
     let stacks_tip = context
         .get_storage()
         .get_stacks_chain_tip(&bitcoin_chain_tip.block_hash)
         .await
         .unwrap()
         .unwrap();
-    sqlx::query(
```
cylewitruk, line 462: Correct to assume that these hang together with the above comment about #559?

signer/tests/integration/emily.rs
matteojug: rustfmt reverts it 😅

signer/tests/integration/emily.rs
matteojug: Yes -- I tried to "enclose" the comment relevant block with the lone `//` below; we still use `stacks_tip` to check that the emily update has the correct information at the end of the test.

signer/tests/integration/emily.rs
matteojug: We don't have a faker for it (`SortitionInfo` comes from stackslib), but point taken: moved the dummy value definition outside to declutter this in [6e03268](https://github.com/stacks-network/sbtc/pull/730/commits/6e032680bcc3410e25de55698464945fdc7af486)

signer/tests/integration/emily.rs
cylewitruk: Whaat :rofl: bad `rustfmt`....!

signer/tests/integration/emily.rs
cylewitruk: Ah gotcha, well your solution solves it anyway and it's much easier to read! thx :)

cylewitruk (approved):
✅

PR #732, opened 2024-10-30
feat: inform peers of new sweep transactions (#709)

cylewitruk:
## Description

Closes: #709
Related to #585

## Changes

- Transaction coordinators now broadcast information about new sweep transactions to the network.
- Transaction signers now receive this message (including the coordinator's own signer) and store the sweep transactions and related deposits/withdrawals in the database.

## Testing Information

There is a new test that sends a 'SweepTransactionInfo` message to a number of signers and verifies that they all correctly store the sweep transactions, and that retrieving these transactions results in the same result as the input.

## Checklist:


signer/tests/integration/transaction_signer.rs
```diff
@@ -161,6 +201,150 @@ async fn should_be_able_to_participate_in_signing_round() {
...
+    // Ensure that the sweep transaction info has been stored in each signer's
+    // database and is the "latest" sweep transaction.
+    for db in signer_dbs.iter() {
+        let retrieved_tx = db
+            .get_latest_sweep_transaction(&bitcoin_chain_tip, 10)
+            .await
+            .expect("failed to get sweep transaction")
+            .expect("no sweep transaction found");
+
+        assert_eq!(*retrieved_tx.txid, sweep_tx_info.txid);
```
cylewitruk, line 334: Note - think i found a bug in the storage; you can start your review but I'll probably be pushing some small fix

cylewitruk: Was the test, `ON CONFLICT DO NOTHING` bit me ;) Fixed in [46ac7a1](https://github.com/stacks-network/sbtc/pull/732/commits/46ac7a10c7c820f3be5a56f8c72dc8476b715b4a)

signer/tests/integration/transaction_signer.rs
```diff
@@ -161,6 +203,151 @@ async fn should_be_able_to_participate_in_signing_round() {
...
+    // deposit and withdraw requests in the sweep transaction info we broadcast.
+    let test_params = testing::storage::model::Params {
+        num_bitcoin_blocks: 2,
+        num_stacks_blocks_per_bitcoin_block: 1,
+        num_deposit_requests_per_block: 2,
+        num_withdraw_requests_per_block: 2,
+        num_signers_per_request: 0,
+    };
+    let test_data = TestData::generate(&mut rng, &[], &test_params);
+    let bitcoin_chain_tip = test_data.bitcoin_blocks.last().unwrap().block_hash;
```
matteojug, line 273: nit: this is (always) the correct chain tip only for `num_bitcoin_blocks` <= 2, to avoid surprises should someone change that number what about using one of the storages to get the chain tip (after we populate them)?

signer/tests/integration/transaction_signer.rs
```diff
@@ -161,6 +203,151 @@ async fn should_be_able_to_participate_in_signing_round() {
...
+
+    // Start the event loops for each signer
+    let handles = signers
+        .into_iter()
+        .map(|signer| tokio::spawn(signer.run()))
+        .collect::<Vec<_>>();
+
+    // Give the event loops some time to start up
+    // NOTE: We could remove this kind of sleep if we had a startup-event or something.
+    tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
```
matteojug, line 301: nit: in other tests we used a [counter](https://github.com/stacks-network/sbtc/blob/main/signer/tests/integration/transaction_coordinator.rs#L548) to wait for startup

matteojug (approved):

signer/src/message.rs
```diff
@@ -227,10 +345,46 @@ impl wsts::net::Signable for Payload {
             Self::BitcoinTransactionSignAck(msg) => msg.hash(hasher),
             Self::StacksTransactionSignRequest(msg) => msg.hash(hasher),
             Self::StacksTransactionSignature(msg) => msg.hash(hasher),
+            Self::SweepTransactionInfo(msg) => msg.hash(hasher),
+        }
+    }
+}
+
+impl wsts::net::Signable for SweepTransactionInfo {
+    fn hash(&self, hasher: &mut sha2::Sha256) {
```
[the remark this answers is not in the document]
cylewitruk: nope, was a miss, good catch! fixed in [f66e1fa](https://github.com/stacks-network/sbtc/pull/732/commits/f66e1fa097f684580fc0c3cf42bb56cf5e806bb8)

signer/tests/integration/transaction_signer.rs
cylewitruk: yeah that's a good point, updated in [f66e1fa](https://github.com/stacks-network/sbtc/pull/732/commits/f66e1fa097f684580fc0c3cf42bb56cf5e806bb8)

signer/tests/integration/transaction_signer.rs
cylewitruk: i'm not really pleased with any of the current solutions :sweat_smile: but i changed it to a counter in [f66e1fa](https://github.com/stacks-network/sbtc/pull/732/commits/f66e1fa097f684580fc0c3cf42bb56cf5e806bb8), that's at least a little better.

signer/src/message.rs
```diff
@@ -124,6 +128,120 @@ impl From<WstsMessage> for Payload {
...
 
+impl From<SweepTransactionInfo> for Payload {
+    fn from(value: SweepTransactionInfo) -> Self {
+        Self::SweepTransactionInfo(value)
+    }
+}
+
+/// Represents information about a new sweep transaction.
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct SweepTransactionInfo {
```
djordon, line 139: We might want to include the actual transaction here as well.

djordon (commented):
I'm thinking we might want to change the structure of these messages to be a bit different. Instead of being nicely laid out, I'm thinking that we structure it in a way where each signer figures out and verifies all of this information themselves. The messages here are structured in a way where you essentially trust everything to be correct.

djordon:
Regarding https://github.com/stacks-network/sbtc/pull/732#pullrequestreview-2409998042, if we all agree, we can merge this but open a follow-up ticket for the change.

signer/tests/integration/transaction_signer.rs
cylewitruk: lol and i would have seen that if i wouldn't have missed to actually remove the 500ms wait below :see_no_evil: i gave this a little refactor to the solution i described in the comment, which gives a definitive way of knowing that the actual signer event loop has started. take a look.

cylewitruk:
> I'm thinking we might want to change the structure of these messages to be a bit different. Instead of being nicely laid out, I'm thinking that we structure it in a way where each signer figures out and verifies all of this information themselves. The messages here are structured in a way where you essentially trust everything to be correct.

@djordon I'm not sure I necessarily agree that the message layout itself needs to change, I think it's rather that signers should be validating the information in the messages (which I purposely left out atm as it's not very clear what we need to validate). If the signers are left to "figure things out themselves" then there's a risk that they "figure things out" differently, and that should be a potential red flag of a misbehaving coordinator which they can't otherwise conclude if they don't have the information to compare to.

signer/src/message.rs
cylewitruk: I'm not a huge fan of throwing around external types in a proprietary messaging protocol tbh. The signer can retrieve the transaction from their db/node if they want it?

djordon: Yeah, I think that you're right, including the transaction is not necessary here.

djordon:
> @djordon I'm not sure I necessarily agree that the message layout itself needs to change, I think it's rather that signers should be validating the information in the messages (which I purposely left out atm as it's not very clear what we need to validate).

Right, we spoke about this offline. These messages are kinda special at the moment. There consensus critical, and we don't have a great story about them. There were two things there we suggested toward the end that might be worth exploring further: (well not in to this PR, just generally):
1. Write the function that matches transaction outputs to withdrawal requests. It might be ugly but it'll be worth it.
2. Have each signer just depend on bitcoin core to figure out broadcast messages. So maybe fetch the transaction relevant transactions from the mempool as part of the regular flow.

With these changes each signer does not need to the coordinator sending this information, and they can figure everything out by looking at the mempool or blockchain.

signer/tests/integration/transaction_signer.rs
matteojug: Yeah better than the custom things we have around.

djordon:
I opened https://github.com/stacks-network/sbtc/issues/758 for the work described in my above comment https://github.com/stacks-network/sbtc/pull/732#issuecomment-2452305919. It's two or three tickets in one.

cylewitruk:
@djordon then I'll go ahead and merge this so it doesn't go stale and we can tackle next steps in #758 :+1:

Issue #682, opened 2024-10-18
[Feature]: Compute a fallback fee for stacks transactions


djordon:
# Feature - Compute a fallback fee for stacks transactions

## 1. Description

When we are first starting out, the fee rate that we use needs to be such that it will be accepted by a Stacks node. Right now our "robust" fee is not robust at all; stacks-nodes reject our contract calls with a `FEE_TOO_LOW` rejection reason.

### 1.1 Context & Purpose

We need to get the fee for our contract call transactions even when the stacks node does not have enough information in order to provide an estimate.

## 2. Technical Details:

A transaction will be rejected with fee too low if the fee is less than the size of the transaction in bytes. The code for this in stacks-core is [here](https://github.com/stacks-network/stacks-core/blob/8b9874d1eb78516397ca07edc4cf61421b7f40ed/stackslib/src/chainstate/stacks/db/blocks.rs#L6693-L6698). So our fallback fee should compute the full fee size and then just multiply that amount by 1 or 2.

### 2.1 Acceptance Criteria:
   - [ ] The transaction fee for stacks transactions is never lower than the transaction size.

## 3. Related Issues and Pull Requests (optional):

* https://github.com/stacks-network/sbtc/pull/368

PR #733, opened 2024-10-30
feat: fallback to calculating stacks tx fee based on payload size

cylewitruk:
## Description

Closes: #682

I don't know if this was the intended change, but it was small so I decided to take a stab at it. I'm not really sure if the numbers are correct (if 1 byte is 1 mSTX or what, it wasn't clear in the stacks code, if we need an additional multiplier as well).

## Changes

If the Stacks node can't provide a fee estimate for either the transaction or the dummy STX transfer, fall-back to using a transaction-size-based calculation.

## Testing Information

Added a new test to enure that the fallback value is correctly used if the Stacks node fails to provide an estimate.

## Checklist:


signer/src/stacks/api.rs
```diff
@@ -902,11 +896,23 @@ impl StacksInteract for StacksClient {
...
+        // method that the stacks node uses to determine transaction size when
+        // verifying admittance into the mempool.
+        let mut payload_bytes = Vec::new();
+        payload
+            .tx_payload()
+            .consensus_serialize(&mut payload_bytes)
+            .map_err(Error::StacksCodec)?;
+        // In Stacks core, the minimum fee is 1 mSTX per byte, so we double
+        // that here as a precaution and cap it at our maximum fee.
+        let default_min_fee = (payload_bytes.len() as u64 * 2).min(MAX_TX_FEE);
```
cylewitruk, line 909: I wasn't quite sure if we need another multiplier here, you had 1_000_000 as the default fee and this will definitely not be 1 million. The stacks code only has "1" listed, and doesn't say if that's mstx or stx.

signer/src/stacks/api.rs
```diff
@@ -902,11 +896,23 @@ impl StacksInteract for StacksClient {
     where
         T: AsTxPayload + Send + Sync,
     {
+        // Consensus serialize the transaction payload to bytes. This is the
+        // method that the stacks node uses to determine transaction size when
+        // verifying admittance into the mempool.
+        let mut payload_bytes = Vec::new();
+        payload
+            .tx_payload()
+            .consensus_serialize(&mut payload_bytes)
```
cylewitruk, line 905: I _think_ that this is the same [`consensus_serialize`](https://github.com/stacks-network/stacks-core/blob/40059a57cd27e740c5e9d91a833fb2c975b0bf0b/stackslib/src/core/mempool.rs#L2356) used to calculate the transaction size for admission into the mempool...

djordon, line 908: Oh, I think we'll need to modify this slightly. Since `tx_payload()` returns a `TransactionPayload` object, it is missing things that we need in the transaction, like version, chain id, signatures, and a few other things. These things increase the size of the transaction and distort the fee.

I think we should consider adding a function that creates a "dummy" transaction from the payload and the multi-sig wallet parameters that will have the exact size as the "proper" transaction but with invalid signatures.

djordon: You could also use `serialize_to_vec()`.

signer/src/stacks/api.rs
```diff
@@ -933,7 +948,7 @@ impl StacksInteract for StacksClient {
             FeePriority::High => resp.estimations.last().map(|est| est.fee),
         };
 
-        Ok(fee_estimate.unwrap_or(DEFAULT_TX_FEE).min(MAX_TX_FEE))
+        Ok(fee_estimate.unwrap_or(default_min_fee))
```
djordon, line 951: Hmm, well I'm not sure this will solve the issue. I think the issue is that fee_estimate is `Some(X)` but `X < default_min_fee`.

Also, this no longer constrains the maximum fee that a signer will pay. That was added in case someone's stacks node returns some wonky estimate that is probably incorrect. No idea if it is necessary though.

djordon (commented):
I think we'll need to modify the approach slightly to get us the right minimum estimate.

signer/src/stacks/api.rs
cylewitruk: Yeah I wasn't sure if it's like Bitcoin or not where witness data is treated differently etc. It was problematic I noticed to do this both pre-signing and post-signing since pre-signing we don't have all the signatures, and post-signing the fee may change.

I'll check out the dummy transaction approach.

signer/src/stacks/api.rs
cylewitruk: > Hmm, well I'm not sure this will solve the issue. I think the issue is that fee_estimate is Some(X) but X < default_min_fee.

Ah okay, changed it now so it solves both stacks-node errors (i.e. if it returns that it can't provide a fee) and if the fee is too small (pending the change for the "transaction" above). We should still "officially" decide on a multiplier though.

> Also, this no longer constrains the maximum fee that a signer will pay. That was added in case someone's stacks node returns some wonky estimate that is probably incorrect. No idea if it is necessary though.

Ahh right whoops, my brain short-circuited there, wasn't intended to remove that :sweat_smile:

signer/src/stacks/api.rs
cylewitruk: Changed to use that

signer/src/stacks/api.rs
```diff
@@ -902,11 +899,21 @@ impl StacksInteract for StacksClient {
...
     {
+        // Consensus serialize the transaction payload to bytes. This is the
+        // method that the stacks node uses to determine transaction size when
+        // verifying admittance into the mempool.
+        let payload_bytes = payload.tx_payload().serialize_to_vec();
+
+        // In Stacks core, the minimum fee is 1 mSTX per byte, so we double
+        // that here as a precaution and cap it at our maximum fee.
+        let default_min_fee =
+            (payload_bytes.len() as u64 * TX_FEE_PAYLOAD_SIZE_MULTIPLIER).min(MAX_TX_FEE);
```
djordon, line 910: Oh we still need to update this to use the new `get_full_tx_size`

signer/src/stacks/wallet.rs
```diff
@@ -397,14 +404,33 @@ pub fn get_full_tx_size<T>(payload: &T, wallet: &SignerWallet) -> Result<u64, Er
...
     let num_signers = wallet.public_keys().len();
-    let private_keys = std::iter::repeat_with(|| PrivateKey::new(&mut rng))
+
+    // We need some dummy private keys to sign the transaction. We check if we've
+    // already generated enough dummy keys, and if not, we generate more. This
+    // is to ensure that always have enough dummy keys even if the signing set
+    // grows.
+    let mut dummy_keys = DUMMY_PRIVATE_KEYS
+        .lock()
+        .expect("BUG! DUMMY_PRIVATE_KEYS lock poisoned");
```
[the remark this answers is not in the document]
cylewitruk: Ah I actually had a LazyCell first but wasn't sure if we should have a cap for key count, but sounds like we have it capped already. Otherwise this impl was just due to my experience of generating private keys to generally be computationally expensive, but it sounds like ecsda256 keys are cheap.

I like the LazyCell solution the best so I'll toss that in :+1:

signer/src/stacks/api.rs
cylewitruk: Yeah I had just pushed changes + test for `get_full_tx_size`, will fix up the above now

djordon (approved):
Made a few changes, looks good to me!

PR #743, opened 2024-10-30
380 feature handle lock times in deposit requests sql only

AshtonStephens:
## Description

Closes: #380 

Prevents the transaction coordinator from trying to sweep in pending deposits that can be reclaimed too close to the current canonical chain tip.

## Changes

- Changes the `get_pending_accepted_deposit_requests` postgres call to not return pending deposits that can be reclaimed by the depositor too close to the current chain tip.
- Adds a constant `MINIMUM_RECLAIM_PROXIMITY_TO_CHAIN_TIP` that encodes the proximity to chain tip.
- Adds a clunky implementation of the same change to `get_pending_accepted_deposit_requests` for the in memory version of the data store.

## Testing Information

Tested with one beefy integration test.

## Checklist:

### DIFF
--- a/signer/src/config/mod.rs
+++ b/signer/src/config/mod.rs
@@ -24,7 +24,7 @@ use crate::stacks::wallet::SignerWallet;
 mod error;
 mod serialization;
 
-/// The minimum number of blocks between the hight at which the reclaim script path can
+/// The minimum number of blocks between the height at which the reclaim script path can
 /// be spent and the current chain tip for the deposit to be considered acceptable.
 pub const MINIMUM_RECLAIM_PROXIMITY_TO_CHAIN_TIP: u16 = 10;
 
--- a/signer/src/config/mod.rs
+++ b/signer/src/config/mod.rs
@@ -24,6 +24,10 @@ use crate::stacks::wallet::SignerWallet;
 mod error;
 mod serialization;
 
+/// The minimum number of blocks between the hight at which the reclaim script path can
+/// be spent and the current chain tip for the deposit to be considered acceptable.
+pub const MINIMUM_RECLAIM_PROXIMITY_TO_CHAIN_TIP: u16 = 10;
+
 /// Maximum configurable delay (in seconds) before processing new Bitcoin blocks.
 pub const MAX_BITCOIN_PROCESSING_DELAY_SECONDS: u64 = 300;
 
--- a/signer/src/config/mod.rs
+++ b/signer/src/config/mod.rs
@@ -24,10 +24,6 @@ use crate::stacks::wallet::SignerWallet;
 mod error;
 mod serialization;
 
-/// The minimum number of blocks between the height at which the reclaim script path can
-/// be spent and the current chain tip for the deposit to be considered acceptable.
-pub const MINIMUM_RECLAIM_PROXIMITY_TO_CHAIN_TIP: u16 = 10;
-
 /// Maximum configurable delay (in seconds) before processing new Bitcoin blocks.
 pub const MAX_BITCOIN_PROCESSING_DELAY_SECONDS: u64 = 300;
 
--- a/signer/src/lib.rs
+++ b/signer/src/lib.rs
@@ -55,4 +55,10 @@ const MAX_KEYS: u16 = 128;
 /// depositer can reclaim the deposit transaction. Signers will not attempt
 /// to sweep in the deposited funds if the number of blocks left is less
 /// than or equal to this value.
+///
+/// If the current chain tip is at height 1000, the reclaim script on a
+/// deposit can be spent on or after block 1001, and this constant value is
+/// set to 1, then the signers WOULD NOT attempt to sweep the deposit. If
+/// it were spendable on block 1002, then the signers WOULD attempt to sweep
+/// the deposit.
 pub const DEPOSIT_LOCKTIME_BLOCK_BUFFER: u16 = 3;
--- a/signer/src/lib.rs
+++ b/signer/src/lib.rs
@@ -55,4 +55,4 @@ const MAX_KEYS: u16 = 128;
 /// depositer can reclaim the deposit transaction. Signers will not attempt
 /// to sweep in the deposited funds if the number of blocks left is less
 /// than or equal to this value.
-const DEPOSIT_LOCKTIME_BLOCK_BUFFER: u16 = 3;
+pub const DEPOSIT_LOCKTIME_BLOCK_BUFFER: u16 = 3;
--- a/signer/src/storage/in_memory.rs
+++ b/signer/src/storage/in_memory.rs
@@ -12,7 +12,6 @@ use time::OffsetDateTime;
 use tokio::sync::Mutex;
 
 use crate::bitcoin::utxo::SignerUtxo;
-use crate::config::MINIMUM_RECLAIM_PROXIMITY_TO_CHAIN_TIP;
 use crate::error::Error;
 use crate::keys::PublicKey;
 use crate::keys::PublicKeyXOnly;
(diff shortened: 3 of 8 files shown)


signer/src/storage/in_memory.rs
```diff
@@ -259,11 +260,66 @@ impl super::DbRead for SharedStore {
...
+                if self
+                    .in_canonical_bitcoin_blockchain(&chain_tip_ref, &block_ref)
+                    .await?
+                {
+                    maybe_canonical_bitcoin_block_with_transaction = Some(block_hash);
+                    break;
+                }
+            }
+            let canonical_bitcoin_block_with_transaction =
+                maybe_canonical_bitcoin_block_with_transaction.ok_or(Error::MissingBlock)?;
```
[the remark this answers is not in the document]
AshtonStephens: since the unlock time is relative we need to know the height of the block it was contained in, but if this call fails we can't figure it out.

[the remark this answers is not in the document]
AshtonStephens: In reality this can't happen because earlier parts of the function would almost certainly not function either, so this is meant to be a flag that something is wrong more than it is meant to be a error one catches. This is in the context of this being a test struct as well

signer/tests/integration/postgres.rs
```diff
@@ -573,6 +575,184 @@ async fn should_return_the_same_pending_accepted_withdraw_requests_as_in_memory_
...
+    };
+    let threshold = 4;
+
+    let signer_set = testing::wsts::generate_signer_set_public_keys(&mut rng, num_signers);
+    let mut test_data = TestData::generate(&mut rng, &signer_set, &test_model_params);
+
+    // Modify the lock times of the deposit requests to be definitely okay to accept because
+    // it's the largest possible lock time.
+    for deposit in test_data.deposit_requests.iter_mut() {
+        deposit.lock_time = i32::MAX as u32;
```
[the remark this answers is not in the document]
AshtonStephens: Done.

AshtonStephens:
> The in-memory implementation was a bit hard to follow. I had one question about that. Otherwise it looks good.

The functions to do it nicely don't really exist because we rely on getting specifically the block height of the transaction's inclusion in the canonical chain, and I chose to forgo adding other helper functions because it's a test struct. Writing this function requires that we get the chain tip, get the pending transactions following from that tip, get the blocks that contain those transactions, then determine which block it's included in that's in the canonical chain, then do another query to get the block info, and then filter out the pending deposits that can be reclaimed outside the window we want.

It might be cleaner to have a function that just returns the canonical chain window of Bitcoin blocks and do the rest in rust with all the data immediately there :P

signer/src/storage/postgres.rs
```diff
@@ -670,7 +680,9 @@ impl super::DbRead for PgStore {
                 WHERE last.depth < $2
             ),
             transactions_in_window AS (
-                SELECT transactions.txid
+                SELECT
+                      transactions.txid
+                    , blocks_in_window.block_height
```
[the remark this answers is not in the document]
AshtonStephens: Done.

signer/src/transaction_signer.rs
```diff
@@ -614,7 +614,7 @@ where
         Ok(())
     }
 
-    /// TODO(#380): This function needs to filter deposit requests based on
+    /// TODO(#742): This function needs to filter deposit requests based on
```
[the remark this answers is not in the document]
AshtonStephens: Done.

djordon (approved):
Looks good

signer/src/config/mod.rs
```diff
@@ -24,6 +24,10 @@ use crate::stacks::wallet::SignerWallet;
 mod error;
 mod serialization;
 
+/// The minimum number of blocks between the hight at which the reclaim script path can
```
matteojug, line 27:
```suggestion
/// The minimum number of blocks between the height at which the reclaim script path can
```

signer/src/storage/in_memory.rs
```diff
@@ -259,11 +260,66 @@ impl super::DbRead for SharedStore {
...
+            .ok_or(Error::MissingBitcoinBlock(*chain_tip))?;
+
+        let chain_tip_ref = self
+            .get_bitcoin_block(chain_tip)
+            .await?
+            .map(|block| model::BitcoinBlockRef {
+                block_hash: *chain_tip,
+                block_height: block.block_height,
+            })
+            .unwrap();
```
matteojug, line 277: nit: call `self.get_bitcoin_block(chain_tip)` once and use the result to compute the two things?

signer/src/storage/in_memory.rs
```diff
@@ -259,11 +260,66 @@ impl super::DbRead for SharedStore {
...
+                    .map(|block| model::BitcoinBlockRef {
+                        block_hash,
+                        block_height: block.block_height,
+                    })
+                    .unwrap();
+                if self
+                    .in_canonical_bitcoin_blockchain(&chain_tip_ref, &block_ref)
+                    .await?
+                {
+                    maybe_canonical_bitcoin_block_with_transaction = Some(block_hash);
```
matteojug, line 302: If we store `block_ref` here we can avoid the call to `.get_bitcoin_block` below

signer/src/storage/postgres.rs
```diff
@@ -691,13 +703,15 @@ impl super::DbRead for PgStore {
             JOIN sbtc_signer.deposit_signers signers USING(txid, output_index)
             WHERE
                 signers.is_accepted
+                AND (transactions.block_height + deposit_requests.lock_time) > $4
```
matteojug, line 706: this check and the in memory one seems to differ for the equal case

signer/tests/integration/postgres.rs
```diff
@@ -573,6 +575,184 @@ async fn should_return_the_same_pending_accepted_withdraw_requests_as_in_memory_
...
+        .get_pending_accepted_deposit_requests(&chain_tip, context_window, threshold)
+        .await
+        .expect("failed to get pending deposit requests from in memory store.");
+
+    pending_accepted_deposit_requests.sort();
+    in_memory_pending_accepted_deposit_requests.sort();
+    assert_eq!(
+        pending_accepted_deposit_requests, in_memory_pending_accepted_deposit_requests,
+        "Basic pending accepted deposit requests functionality is broken."
+    );
```
matteojug, line 644: can we add below the same checks comparing the in memory and pg results?

signer/src/config/mod.rs
```diff
@@ -24,6 +24,10 @@ use crate::stacks::wallet::SignerWallet;
...
+/// be spent and the current chain tip for the deposit to be considered acceptable.
+pub const MINIMUM_RECLAIM_PROXIMITY_TO_CHAIN_TIP: u16 = 10;
```
matteojug, line 29: With #744 merged, do we still need this or can we reuse https://github.com/stacks-network/sbtc/blob/main/signer/src/lib.rs#L58 ?

AshtonStephens: We can reuse that

AshtonStephens: Done.

signer/tests/integration/postgres.rs
AshtonStephens: Done.

signer/src/storage/postgres.rs
AshtonStephens: Fixed.

signer/src/storage/in_memory.rs
AshtonStephens: Changed the approach - effectively done.

signer/src/storage/in_memory.rs
AshtonStephens: Changed the approach - effectively done.

signer/src/config/mod.rs
AshtonStephens: Fixed.

AshtonStephens:
Had a rough time with an off by one iteration error in the `in_memory` version, but I ultimately fixed it.

AshtonStephens:
Note, I chose to leave in the white space changes on the postgres file because I think they make sense.

signer/tests/integration/postgres.rs
```diff
@@ -573,6 +575,194 @@ async fn should_return_the_same_pending_accepted_withdraw_requests_as_in_memory_
...
+            minimum_acceptable_unlock_height - height_included as u32;
+
+        let unique_deposit_id: (BitcoinTxId, u32) =
+            (deposit_request.txid, deposit_request.output_index);
+
+        if out_of_bounds_requests.contains(&unique_deposit_id) {
+            // Make the block the request can be reclaimed at one lower than the minimum.
+            deposit_request.lock_time = minimum_acceptable_unlock_time_for_this_deposit - 1;
+        } else if in_bounds_requests.contains(&unique_deposit_id) {
+            // Make the block the request can be reclaimed at one lower at the minimum and
```
matteojug, line 721: nit?
```suggestion
            // Make the block the request can be reclaimed at one lower than the minimum and
```

matteojug (approved):

signer/src/storage/in_memory.rs
```diff
@@ -259,12 +261,49 @@ impl super::DbRead for SharedStore {
...
+                    .map(|block_included: &model::BitcoinBlock| {
+                        let unlock_height =
+                            block_included.block_height as u32 + deposit_request.lock_time;
+                        unlock_height >= minimum_acceptable_unlock_height
+                    })
+                    .take(1)
+                    .collect::<Vec<_>>()
+                    .first()
+                    .cloned()
+                    .unwrap_or(true)
```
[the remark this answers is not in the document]
AshtonStephens: Changing it to false - at one point the tests were failing when it was false but it must have been another error.

signer/src/storage/in_memory.rs
```diff
@@ -259,12 +261,49 @@ impl super::DbRead for SharedStore {
...
+                    .filter_map(|block_hash| store.bitcoin_blocks.get(block_hash))
+                    .map(|block_included: &model::BitcoinBlock| {
+                        let unlock_height =
+                            block_included.block_height as u32 + deposit_request.lock_time;
+                        unlock_height >= minimum_acceptable_unlock_height
+                    })
+                    .take(1)
+                    .collect::<Vec<_>>()
+                    .first()
+                    .cloned()
```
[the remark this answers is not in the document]
AshtonStephens: Nice!

signer/tests/integration/postgres.rs
AshtonStephens: Done.

signer/src/lib.rs
```diff
@@ -55,4 +55,10 @@ const MAX_KEYS: u16 = 128;
...
 /// to sweep in the deposited funds if the number of blocks left is less
 /// than or equal to this value.
-const DEPOSIT_LOCKTIME_BLOCK_BUFFER: u16 = 3;
+///
+/// If the current chain tip is at height 1000, the reclaim script on a
+/// deposit can be spent on or after block 1001, and this constant value is
+/// set to 1, then the signers WOULD NOT attempt to sweep the deposit. If
+/// it were spendable on block 1002, then the signers WOULD attempt to sweep
+/// the deposit.
+pub const DEPOSIT_LOCKTIME_BLOCK_BUFFER: u16 = 3;
```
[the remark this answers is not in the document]
AshtonStephens: Opened https://github.com/stacks-network/sbtc/issues/751

signer/src/storage/in_memory.rs
```diff
@@ -259,12 +261,49 @@ impl super::DbRead for SharedStore {
...
+        // Add one to the acceptable unlock height because the chain tip is at height one less
+        // than the height of the next block, which is the block for which we are assessing
+        // the threshold.
+        let minimum_acceptable_unlock_height =
+            store.bitcoin_blocks.get(chain_tip).unwrap().block_height as u32
+                + DEPOSIT_LOCKTIME_BLOCK_BUFFER as u32
+                + 1;
+
+        let canonical_bitcoin_blocks = (0..context_window)
+            // Find all tracked transaction IDs in the context window
```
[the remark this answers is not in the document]
AshtonStephens: Done.

Issue #543, opened 2024-09-18
[Feature]: Update function that fetches accepted deposit requests


djordon:
# Feature - Update function that fetches accepted deposit requests

## 1. Description

There are a number of functions on the `DbRead` trait that are supposed to fetch pending deposit or pending withdrawal requests. At this time the queries in these functions do not attempt to distinguish between pending and non-pending requests. They also do not distinguish between requests that were accepted and requests that are currently accepted under the current signing set. This is expected, because when these queries were written we did not settle on a way to distinguish these facts.

This issue focuses on pending deposit requests.

### 1.1 Context & Purpose

We need this to ensuring that we do not attempt to mint to a user more than once. Although the smart contract will stop a double mint for the same deposit, the signers will waste gas attempting to do so.

## 2. Technical Details:

A deposit request is pending if there is no corresponding `completed-deposit` event from a transaction that is on the canonical stacks blockchain. Doing this correctly requires us to implement the part of our data model that ties bitcoin blocks to stacks blocks, and that is difficult to do.

### 2.1 Acceptance Criteria:
   - [ ] Make sure we fetch only pending deposit requests in our queries.
   - [ ] Make sure we fetch only deposit requests that are currently accepted. 

## 3. Related Issues and Pull Requests (optional):

* Much of the remaining work here is dependent on #559.

PR #747, opened 2024-10-31
feat: consider stacks anchors in get_swept_deposit_requests

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/543

## Changes

Small change to the query, added tests.

## Checklist:


signer/src/storage/postgres.rs
```diff
@@ -1411,15 +1409,26 @@ impl super::DbRead for PgStore {
...
+            GROUP BY
+                bc_trx.txid
+              , bc_trx.block_hash
+              , bc_blocks.block_height
+              , deposit_req.txid
+              , deposit_req.output_index
+              , deposit_req.recipient
+              , deposit_req.amount
+            HAVING
+                COUNT(cde_bc.block_hash) = 0
```
matteojug, line 1431: We need the grouping to handle cases where we have two complete deposits events on two forks: just checking `IS NULL` would return the deposit because on a fork it is not confirmed (even though we have a different confirmation on the canonical chain).

signer/tests/integration/postgres.rs
```diff
@@ -1565,39 +1562,93 @@ async fn get_swept_deposit_requests_does_not_return_deposit_requests_with_respon
...
+        block_id: *setup_event_block.block_hash,
         amount: setup.deposit_request.amount,
         outpoint: setup.deposit_request.outpoint,
     };
-
     db.write_completed_deposit_event(&event).await.unwrap();
 
-    let chain_tip = setup.sweep_block_hash.into();
-    let context_window = 20;
+    let event = CompletedDepositEvent {
```
cylewitruk, line 1645: suggestion: maybe add a comment to these two events to make it clear which is on the canonical chain and which isn't and the expected effect.

signer/tests/integration/postgres.rs
```diff
@@ -1676,11 +1755,103 @@ async fn can_sign_deposit_tx_rejects_not_in_signer_set() {
...
+        .get_swept_deposit_requests(&chain_tip.into(), context_window)
+        .await
+        .unwrap();
+
+    assert!(requests.is_empty());
+
+    // Now assume we have a reorg: the new bitcoin chain is `sweep_block_hash`
+    // and the complete deposit event is no longer in the canonical chain.
+    // The deposit should no longer be confirmed.
+    let requests = db
```
cylewitruk, line 1846: Hm, I'm trying to reason whether or not this tests what the test description says... Doesn't this test that the Bitcoin chain has re-org'd? If i'm reading the description correctly, it reads to me that the Bitcoin chain is still canonical but that the `complete-deposit` transaction has been orphaned in Stacks (a Stacks re-org without Bitcoin re-org).

But maybe I'm reading things completely wrong?

signer/migrations/0004__create_functions.sql
```diff
@@ -44,3 +44,57 @@ BEGIN
...
+BEGIN
+    RETURN QUERY
+    WITH RECURSIVE bc_blocks AS (
+        SELECT * FROM sbtc_signer.bitcoin_blockchain_of(bitcoin_chain_tip, max_depth)
+    ),
+    blockchain AS (
+        SELECT
+            stacks_blocks.block_hash,
+            stacks_blocks.block_height,
+            stacks_blocks.parent_hash
```
djordon, line 76: style nit: can we use leading commas please.

signer/src/storage/postgres.rs
```diff
@@ -1376,18 +1376,21 @@ impl super::DbRead for PgStore {
...
+
+        let Some(stacks_chain_tip) = self.get_stacks_chain_tip(chain_tip).await? else {
+            return Ok(Vec::new());
+        };
 
         sqlx::query_as::<_, model::SweptDepositRequest>(
             "
+            WITH s_blocks AS (
+                SELECT * FROM stacks_blockchain_of($3, $1, $2)
+            )
```
djordon, line 1393: Can we include this in the left join below? Like so?
```sql
LEFT JOIN stacks_blockchain_of($3, $1, $2) AS sb
  ON ...
```

signer/src/storage/postgres.rs
```diff
@@ -1411,19 +1414,28 @@ impl super::DbRead for PgStore {
...
+            GROUP BY
+                bc_trx.txid
+              , bc_trx.block_hash
+              , bc_blocks.block_height
+              , deposit_req.txid
+              , deposit_req.output_index
+              , deposit_req.recipient
+              , deposit_req.amount
+            HAVING
+                COUNT(s_blocks.block_hash) = 0
```
djordon, line 1433: I think we can simplify this to
```sql
WHERE s_blocks.block_hash IS NULL
```

signer/tests/integration/postgres.rs
djordon: Hmm, I think the test description is only supposed to check that it the transaction has been affected by a reorg, it doesn't specify how. We may want more tests that check for Stacks reorgs. This one does seem bitcoin specific, but that is fine.

djordon (commented):
Looks good, I have some simplifying comments

signer/src/storage/postgres.rs
djordon: Oh I noticed this too late.

djordon: Oh wait, we were referring to different tables maybe?

signer/tests/integration/postgres.rs
cylewitruk: The description says that the sweep tx has not been reorg'd but the `complete-deposit` has, which reads to me like a stacks reorg without bitcoin, but if this is fine with you then I'm fine :)

djordon: Oh I see, yeah that interpretation makes sense. I think we just need to make sure that the bitcoin reorg doesn't affect the sweep transaction, which it shouldn't because we use the `sweep_block_hash` as the chain tip below.

signer/src/storage/postgres.rs
matteojug: it's the same table (the comment was for a former version of the query): since we do left joins, checking with `s_blocks.block_hash IS NULL` returns requests having a complete-deposit event in a non canonical chain, even if there's another event on the canonical one. This condition is tested in `get_swept_deposit_requests_does_not_return_deposit_requests_with_responses`: if you try removing the grouping in favor of `s_blocks.block_hash IS NULL` the test fails when asserting that we do not get requests at the end.

signer/migrations/0004__create_functions.sql
```diff
@@ -44,3 +44,57 @@ BEGIN
...
 $$ LANGUAGE plpgsql;
+
+-- Table-Valued Function (TVF) for fetching a Stacks blockchain from a given
+-- stacks block hash and bitcoin block hash, only looking back the `max_depth`
+-- number of bitcoin blocks. The bitcoin blocks are used to consider only
+-- stacks blocks anchored to the canonical bitcoin chain.
+--
+-- - stacks_chain_tip: The stacks block hash to start the blockchain from ("chain tip").
+-- - bitcoin_chain_tip: The bitcoin block hash to start the canonical blockchain from ("chain tip").
+-- - max_depth: The maximum depth of the blockchain to fetch.
```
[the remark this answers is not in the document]
matteojug: [c3e0400](https://github.com/stacks-network/sbtc/pull/747/commits/c3e04003d7fb9578c919756d7ed254180c3df45a)

signer/tests/integration/postgres.rs
matteojug: Added comments and renamed things to make it clear which is which in [c3e0400](https://github.com/stacks-network/sbtc/pull/747/commits/c3e04003d7fb9578c919756d7ed254180c3df45a)

signer/migrations/0004__create_functions.sql
matteojug: Done in [c3e0400](https://github.com/stacks-network/sbtc/pull/747/commits/c3e04003d7fb9578c919756d7ed254180c3df45a), but we should settle on some tool+config to consistent format SQL (that is not only available in a specific IDE, ideally 😅 )

signer/src/storage/postgres.rs
matteojug: [c3e0400](https://github.com/stacks-network/sbtc/pull/747/commits/c3e04003d7fb9578c919756d7ed254180c3df45a)

signer/src/storage/postgres.rs
djordon: This would actually be wrong, and an explanation why can be found in https://github.com/stacks-network/sbtc/pull/747#discussion_r1826660102.

djordon (approved):
Looks good! ✅

Issue #752, opened 2024-11-01
[Feature]: Emily Version Conflict Retry


AshtonStephens:
# Feature - Emily Version Conflict Retry

## 1. Description

The Emily Request handlers are meant to retry when there's a version conflict that occurs in the database, but right now it errors out. This issue is to ensure that version conflicts result in a retry in Emily.

### 1.1 Context & Purpose
The Emily Request handlers are meant to retry when there's a version conflict that occurs in the database, but right now it errors out. This issue is to ensure that version conflicts result in a retry in Emily.

## 2. Technical Details:

1. When there's an update request Emily will re-attempt an update if it fails.

### 2.1 Acceptance Criteria:
   - [x] Emily retries on updates
   - [x] There is a PR for the feature
   - [ ] The feature is merged

## 3. Related Issues and Pull Requests (optional):
- #712

PR #753, opened 2024-11-01
feat: Retry updates when a version conflict fails update

AshtonStephens:
## Description

Closes: #752

## Changes

Adds retry on version conflict when updating an entry in a  dynamodb table to Emily. Also adds a function that prevents attempting an update when the entry doesn't need to consume the update.

Additionally, I had to change the order of the concurrent API calls updating Emily; it cannot receive updates to withdrawals that it doesn't know about yet, so the chainstate and create withdrawal calls need to occur first.

## Testing Information

Tested with unit tests and integration tests.

## Checklist:

### DIFF
--- a/emily/handler/src/database/accessors.rs
+++ b/emily/handler/src/database/accessors.rs
@@ -142,6 +142,8 @@ pub async fn get_deposit_entries_for_transaction(
 
 /// Pulls in a deposit entry and then updates it, retrying the specified number
 /// of times when there's a version conflict.
+///
+/// TODO(792): Combine this with the withdrawal version.
 pub async fn pull_and_update_deposit_with_retry(
     context: &EmilyContext,
     update: ValidatedDepositUpdate,
@@ -339,6 +341,8 @@ pub async fn get_all_withdrawal_entries_modified_after_height_with_status(
 
 /// Pulls in a withdrawal entry and then updates it, retrying the specified number
 /// of times when there's a version conflict.
+///
+/// TODO(792): Combine this with the deposit version.
 pub async fn pull_and_update_withdrawal_with_retry(
     context: &EmilyContext,
     update: ValidatedWithdrawalUpdate,
--- a/emily/handler/src/database/accessors.rs
+++ b/emily/handler/src/database/accessors.rs
@@ -4,7 +4,6 @@ use aws_sdk_dynamodb::types::AttributeValue;
 use serde_dynamo::Item;
 use tracing::{info, warn};
 
-use crate::api::models::deposit::Deposit;
 use crate::common::error::{Error, Inconsistency};
 
 use crate::{api::models::common::Status, context::EmilyContext};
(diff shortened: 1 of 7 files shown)


emily/handler/src/database/entries/deposit.rs
```diff
@@ -540,3 +552,79 @@ impl DepositUpdatePackage {
...
+            amount: 0,
+            parameters: Default::default(),
+            status: Status::Pending,
+            reclaim_script: "".to_string(),
+            deposit_script: "".to_string(),
+            last_update_height: 0,
+            last_update_block_hash: "".to_string(),
+            fulfillment: None,
+            history: vec![
+                DepositEvent {
```
cylewitruk, line 582: For readability I would probably create this in a variable above next to the accepted event and have this be `vec![pending, accepted]`. Took me a minute to realize that the `event.clone()` was there :see_no_evil:

emily/handler/src/database/entries/withdrawal.rs
```diff
@@ -528,3 +540,105 @@ impl WithdrawalUpdatePackage {
...
+            recipient: "recipient".to_string(),
+            amount: 1,
+            parameters: WithdrawalParametersEntry { max_fee: 1 },
+            status: Status::Pending,
+            last_update_height: 1,
+            last_update_block_hash: "hash".to_string(),
+            history: vec![
+                WithdrawalEvent {
+                    status: StatusEntry::Pending,
+                    message: "message".to_string(),
```
cylewitruk, line 581: Same suggestion as in `deposit.rs`

emily/handler/src/database/accessors.rs
```diff
@@ -138,6 +140,38 @@ pub async fn get_deposit_entries_for_transaction(
     .await
 }
 
+/// Pulls in a deposit entry and then updates it, retrying the specified number
+/// of times when there's a version conflict.
+pub async fn pull_and_update_deposit_with_retry(
```
cylewitruk, line 145: was there no way to make a generic `pull_and_update_with_retry`? not worth changing right now, but if there is we should aim for that in future changes to this area imo.

cylewitruk (approved):
Looks fine to me, a few suggestions but nothing critical so approving :+1:

Jiloc (approved):
I see some valid comments from @cylewitruk, but none of those are blockers. PR looks good to me!

emily/handler/src/database/accessors.rs
AshtonStephens: I agree, I was just trying to do the fix quickly, and there are other places where I needed to do these as one offs. I'll open a tech debt issue for it.

AshtonStephens: https://github.com/stacks-network/sbtc/issues/792

emily/handler/src/database/entries/withdrawal.rs
AshtonStephens: Done.

emily/handler/src/database/accessors.rs
cylewitruk: :pray:

Issue #731, opened 2024-10-30
[Feature]: Ensure the bitcoin block confirming each deposit is in the database


djordon:
# Feature - Ensure the bitcoin block confirming a each deposit is in the database

## 1. Description

Before a signer attempts to record a deposit request and the transaction associated with it to the database, they should ensure that the block confirming the deposit transaction is in the database.

### 1.1 Context & Purpose

Signers update their blockchain records upon receiving bitcoin-core notifications, which occur when a block is added to the best chain ([docs](https://github.com/bitcoin/bitcoin/blob/2a52718d734cf65aaeeb18f627547e5bca3483f4/doc/zmq.md)). Since deposits might be confirmed by blocks not on the best chain, signers may be unaware of these blocks when recording deposit transactions. Currently, they don't verify if they know about the block before recording the deposit, leading to a foreign key constraint violations preventing the deposit request from being written.

## 2. Technical Details:

Each signer should check to see if it knows about the block that confirmed a deposit before writing it to the database. If it doesn't know it fetch it and it's ancestors.

### 2.1 Acceptance Criteria:
   - [ ] Signers only write deposit requests when the deposit transaction is confirmed on a block that the signer knows about.

PR #757, opened 2024-11-02
feat: ensure bitcoin block confirming deposit is in database

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/731.

This work came about while adding a test for https://github.com/stacks-network/sbtc/pull/756.

## Changes

* Clean up some logging in the block observer.
* Use shared references where possible in block observer functions (that I looked at).
* Make sure bitcoin blocks containing deposits are in the database before writing the deposit transaction.

## Testing Information

I created a new test case for an existing integration test that kinda tests this code. If you comment out the new function that writes the blocks to the database, the test fails with the expected error (a foreign key violation) for the "eight blocks ago" test. But it's not a great case, and I'm not sure how best to go about it using bitcoin-core.

I also needed to update two existing tests that execute this code path.

## Checklist:

### DIFF
--- a/docker/docker-compose.ci.yml
+++ b/docker/docker-compose.ci.yml
@@ -1,7 +1,6 @@
 services:
 
   bitcoind:
-    container_name: bitcoind
     image: lncm/bitcoind:v25.0
     volumes:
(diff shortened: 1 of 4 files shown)


cylewitruk (approved):
Lgtm

Issue #587, opened 2024-09-30
[Feature]: Only handle requests that do not have a response


djordon:
# Feature - Only handle requests that do not have a response

## 1. Description

When we reach out to the Emily API for deposit requests, we end up adding it to our list of deposits-to-watch-for each time. And every time there is a new bitcoin block we attempt to process "pending" deposits but our queries do not distinguish between deposits that do or do not have a response. This causes us to continually include deposit UTXOs in our bitcoin transactions.

### 1.1 Context & Purpose

Update our queries to exclude deposit requests that have a response transaction that has been confirmed ~~or submitted~~.

### 2.1 Acceptance Criteria:
   - [x] The signers' deposits-to-watch-for list never contains duplicates (this isn't critical, but nice to have). This was fixed in https://github.com/stacks-network/sbtc/pull/679.
   - [ ] The signers never construct a response transaction for a deposit request if they have responded already and it is confirmed on the canonical bitcoin blockchain.

## 3. Related Issues and Pull Requests (optional):

* This depends on ~~https://github.com/stacks-network/sbtc/issues/559~~ https://github.com/stacks-network/sbtc/issues/585.

djordon:
This might be nothing to really worry about. We may end up doing more work than is necessary, but we cannot write duplicate records into the database so we're probably good.

cylewitruk:
> confirmed or submitted

@djordon what is the definition of these two for this case?

djordon:
> > confirmed or submitted
> 
> @djordon what is the definition of these two for this case?

You actually only care about it being confirmed, and confirmed means that the transaction has been included in a bitcoin block that is on the canonical bitcoin blockchain.

cylewitruk:
@djordon so no confirmation wait time here? 1 block is enough? (I think we talked about this before, that the sBTC mints will be orphaned along with a bitcoin block and thus be void)

djordon:
> @djordon so no confirmation wait time here? 1 block is enough? (I think we talked about this before, that the sBTC mints will be orphaned along with a bitcoin block and thus be void)

[...]

PR #764, opened 2024-11-04
fix: ignore swept deposits from get_pending_accepted_deposit_requests

matteojug:
## Description

Closes: #587

## Changes

Add the missing logic to ignore swept deposits (and related integration test). Also fix an issue in `CompleteDepositV1` [validation](https://github.com/stacks-network/sbtc/blob/main/signer/src/stacks/contracts.rs#L347): it was expecting the deposit to be a pending accepted one instead of a swept one.

## Checklist:


signer/tests/integration/complete_deposit.rs
```diff
@@ -366,6 +390,18 @@ async fn complete_deposit_validation_invalid_mint_amount() {
...
+    // Store outstanding sweep transaction packages in the database, which
+    // includes the above deposit request. But remember that this represents a
+    // sweep transaction that has been broadcast to the mempool, but not yet
+    // observed in a block (we would need to also call `.store_sweep_tx()` for
+    // that).
+    //
+    // Note: we need to store the withdrawal request to satisfy FK's since
+    // `TestSweepSetup` includes a withdrawal in the sweep transaction by
+    // default, but we don't use it.
+    setup.store_withdrawal_request(&db).await;
```
djordon, line 402: We shouldn't need the `setup.store_withdrawal_request(&db).await;` calls in these tests.

djordon (commented):
Gave it a quick look, so far looks great. Will be circling back in a bit.

signer/src/stacks/contracts.rs
```diff
@@ -339,21 +339,19 @@ impl CompleteDepositV1 {
...
             .await?;
 
         let deposit_request = deposit_requests
             .into_iter()
-            .find(|req| req.outpoint() == self.outpoint)
+            .find(|req| req.deposit_outpoint() == self.outpoint)
+            .ok_or_else(|| DepositErrorMsg::RequestMissing.into_error(req_ctx, self))?;
+
+        let deposit_request = db
+            .get_deposit_request(&deposit_request.txid, deposit_request.output_index)
```
djordon, line 353: Having a `get_deposit_request` seems pretty useful, but in this case I think it would be better to add a `max_fee` field to the `SweptDepositRequest` object. This way we don't need to do the extra query at all.

signer/tests/integration/postgres.rs
```diff
@@ -575,6 +575,85 @@ async fn should_return_the_same_pending_accepted_withdraw_requests_as_in_memory_
...
+
+    // The request needs to be added to the database. This stores
+    // `setup.deposit_request` into the database.
+    setup.store_deposit_request(&db).await;
+
+    // TODO: Create the initial transaction sweep package without any
+    // withdrawals and have a separate method for creating that sweep (since
+    // it's not realistic to have the withdrawal in the same sweep as the
+    // deposit). Then we wouldn't have to do this.
+    setup.store_withdrawal_request(&db).await;
```
djordon, line 615: I'd remove this from here as well, since it isn't necessary for this query.

djordon (commented):
This looks good. The only real change is that we should just add the `max_fee` field to the swept deposit struct so that we don't need to do a `get_deposit_request` during validation.

As for `get_deposit_request`, I find myself reaching for this method sometimes but it usually turns out that I don't really need it (or adding something more targeted would be better). I'm indifferent if you want to keep it though.

signer/tests/integration/complete_deposit.rs
matteojug: We shouldn't, but unfortunately we do: when we `store_sweep_transactions`, we write into `sweep_transactions` as well as `swept_deposits` and `swept_withdrawals` for each request swept, and those two tables have FK to `deposit_requests` and `withdrawal_requests`; since `TestSweepSetup` generates both a deposit and a withdrawal, if we don't store the withdrawal request when `store_sweep_transactions` we get
```
insert or update on table "swept_withdrawals" violates foreign key constraint "swept_withdrawals_withdrawal_request_id_withdrawal_request_fkey"
[...]
Key (withdrawal_request_id, withdrawal_request_block_hash)=(0, x86443fba4481842a9a295ad6caea944116087d37305da2adbb9bd6dc48920574) is not present in table "withdrawal_requests"
```

djordon: Oh whoops, I forgot about that. Okay cool.

signer/tests/integration/postgres.rs
djordon: This is actually probably needed, the details were described in https://github.com/stacks-network/sbtc/pull/764#discussion_r1828242872.

signer/src/stacks/contracts.rs
matteojug: Makes sense: since `SweptDepositRequest` was missing some info wrt `DepositRequest` I initially thought we may need more of them during validation; but it seems the missing fields are more "sweeping" related, so at this point we don't really need them. Removed in favor of adding `max_fee` in [9af3a56](https://github.com/stacks-network/sbtc/pull/764/commits/9af3a56a1b05d9ca7208a3660db447eff825b74d).

djordon (approved):
Looks good ✅ 🚢

PR #769, opened 2024-11-04
fix: byte reverse bug in contract call

djordon:
## Description

This is basically due to an endian issue with our bitcoin block hash type. This fixes the endianness of what is sent in the clarity contract call.

## Changes

* Change the endianness of the bitcoin block hash sent in the clarity contract calls.

## Testing Information

We don't run stacks-core integration tests in CI, so we need to run these tests in devenv, which is running now.

## Checklist:


matteojug (approved):
As chatted offline we may need to revisit this later as the underlying issue may be something else, but as temporary fix this should be enough

Issue #688, opened 2024-10-22
[Feature]: Signers deploy sBTC smart contracts


djordon:
# Feature - Signers deploy sBTC smart contracts

## 1. Description

When the signers first boot, they should coordinate with one another to deploy the sBTC smart contracts. This should be an internal signer code that attempts to deploy the sBTC smart contracts by the signer coordinator.

### 1.1 Context & Purpose

The account that deploys the sBTC smart contracts have initial control over minting and burning of sBTC. The account that deploys them will need to be some trusted entity that then hands over control of the contracts to the bootstrap signers ([nominated here](https://github.com/stacks-network/sbtc/discussions/624#discussioncomment-10918898)). The simplest and most secure thing to do is to have the bootstrap signers deploy the contracts and retain collective control over minting and burning sBTC.

## 2. Technical Details:

We have code in our integration tests for deploying the smart contracts in https://github.com/stacks-network/sbtc/blob/39c7a6f4d9d2f08b8eff245a4e7b67f49cb37999/signer/tests/integration/contracts.rs#L41-L198. Some of it is salvageable, but we need to change the core logic for signing them and doing validation (which is very simple) and integrate this into the `TxCoordinatorEventLoop` and `TxSignerEventLoop`.

### 2.1 Acceptance Criteria:
   - [ ] Add the `get_contract_source` function to the `StacksInteract` trait.
   - [ ] Move the contract deployment setup structs to someplace outside of the integration tests.
   - [ ] Add code to the transaction coordinator that attempts to sign and deploy the smart contracts if they haven't been already.

PR #770, opened 2024-11-05
fix: reuse wallet when deploying contracts

matteojug:
## Description

Followup for https://github.com/stacks-network/sbtc/issues/688

## Changes

Reuse the same wallet object when deploying the different contracts to avoid conflicting nonces in mempool.
Also bump the block observer horizon to a reasonable value.

## Testing Information

Tested on devenv: all the contracts were deployed in the same stacks block.

## Checklist:


djordon (approved):
Looks good ✅

PR #772, opened 2024-11-05
fix: devenv fixes and nakamoto waiting

djordon:
## Description

These are local fixes that affect how the application behaves in devnet

## Changes

1. Move the nakamoto wait code to the transaction coordinator event loop. This addresses an issue where the stacks node is waiting for the sbtc signer before proceeding, and the signers are waiting for nakamoto before starting their event observer. So they kinda deadlock.
2. Add more profiles to the docker compose for starting more subsets of services.
3. Update the stacks config to add the sbtc signers as event observers.
4. Update the `signer-*.env` files to bind to wildcard ports. This might be necessary with docker services connecting to locally running services.

## Testing Information

This is all devenv related. To test that things "work" do
```bash
make devenv-no-sbtc-up
```
in one terminal and 
```bash
./signers.sh run 3
```
in another. If you have the `demo-cli` binary then you can kick start things with.
```bash
psql postgresql://postgres:postgres@localhost:5432/signer -c "SELECT aggregate_key FROM sbtc_signer.dkg_shares LIMIT 1" --no-align --quiet --tuples-only
target/debug/demo-cli donation --amount 2000000 --signer-key <public-key>
target/debug/demo-cli deposit --amount 4200000 --max-fee 80000 --lock-time 50 --stacks-addr ST2SBXRBJJTH7GV5J93HJ62W2NRRQ46XYBK92Y039 --signer-key <public-key>
```
where `<public-key>` is taken from the first comment (and cleaned to just be hex).

## Checklist:


signer/src/transaction_coordinator.rs
```diff
@@ -221,8 +226,31 @@ where
             .await
     }
 
+    async fn is_epoch3(&mut self) -> Result<bool, Error> {
+        if self.is_epoch3 {
+            return Ok(true);
+        }
+        tracing::info!("Checked for whether we are in Epoch 3 or later");
```
cylewitruk, line 233: these should prob be debug
but otherwise nice function :+1: much cleaner than what was there

signer/tests/integration/emily.rs
```diff
@@ -461,6 +461,7 @@ async fn deposit_flow() {
         signing_round_max_duration: Duration::from_secs(10),
         dkg_max_duration: Duration::from_secs(10),
         sbtc_contracts_deployed: true,
+        is_epoch3: true,
```
cylewitruk, line 464: we really should have a `new()` method so that we don't have to refactor all instantiations when a new field is added... but yeah, another day

cylewitruk (approved):
lgtm, just a couple suggestions, and it might be good to toss what you wrote in this PR description in a/the readme

signer/src/transaction_coordinator.rs
djordon: done in https://github.com/stacks-network/sbtc/pull/772/commits/31f87cab56525a908c925becaf8f709abf9e4b60.

AshtonStephens:
We went back to using `./signers.sh`?

djordon:
> We went back to using `./signers.sh`?

I used them both and both work, but I was having some issues and tried `./signers.sh` partly for faster iteration. I think it's a bit faster than the docker path still, and I'm not sure why.

PR #777, opened 2024-11-05
feat: refactor block observer runloop

cylewitruk:
## Description

Partially resolves: #669 

## Changes

Removes the `tokio::select!` from the block observer runloop which will ensure that if the signer receives a shutdown signal (other than `-9`) that it will complete its current run, writing any data, before shutting down.

## Testing Information

- Added a new method on the `TestContext` for waiting for a specific `SignerSignal` and use it in the one test that needs to wait for `BitcoinBlockObserved`.
- Refactored the other tests there to use `TestContext`... just because (consistency?).

## Checklist:

### DIFF
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -19,6 +19,7 @@
 
 use std::collections::HashMap;
 use std::future::Future;
+use std::time::Duration;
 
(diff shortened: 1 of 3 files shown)


djordon (approved):
Looks good ✅

PR #787, opened 2024-11-07
check that wsts messages are signed by the specified signer_id

xoloki:
## Description
`sBTC` messages contain a signature and a public key, in addition to the payload.  Verifying the message shows that it has not been tampered with.

`WSTS` message payloads either contain a `signer_id`, or are coordinator messages.  So in addition to verifying the signature, we also need to show that the public key used to sign was the correct public key for that signer (or the coordinator).

Closes: #578 

## Changes
Call `verify` on the `sBTC` message, then check that the message was signed by the correct public key.

## Testing Information
All unit and integration tests still pass.

## Checklist:

- [ x] I have performed a self-review of my code
- [ x] My changes generate no new warnings
- [ x] New and existing unit tests pass locally with my changes
- [ x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/signer/src/error.rs
+++ b/signer/src/error.rs
@@ -460,6 +460,10 @@ pub enum Error {
     /// The smart contract has already been deployed
     #[error("smart contract already deployed, contract name: {0}")]
     ContractAlreadyDeployed(&'static str),
+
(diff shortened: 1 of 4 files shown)


djordon:
Since the other one is up to date, might as well close this one.

Issue #653, opened 2024-10-11
[Feature]: Submit `rotate-keys-wrapper` transactions after DKG


djordon:
# Feature - Submit `rotate-keys-wrapper` transactions after DKG

## 1. Description

After an initial round of distributed key generation (DKG), the signers will be able to function just fine (well, after https://github.com/stacks-network/sbtc/issues/652 gets done). The problem is that no one will be able to deposit into the system without the signers informing the world about their latest aggregate key for use on bitcoin. The mechanism to inform the world about the aggregate key is the `rotate-keys` contract call. After a successful `rotate-keys` contract call, the aggregate key will be part of the contract state for anyone to read. So we should always submit a rotate-keys transaction after running DKG.

### 1.1 Context & Purpose

In order to make a deposit into sBTC, the depositor needs to know the current aggregate key of the signers. And in order for anyone to know the aggregate key, the signers need to tell them. The only way to do that right now is for them to broadcast that information as part of a `rotate-keys` contract call.

## 2. Technical Details:

This ticket is just for the construction and submission of a rotate-keys contract call. They exact criteria for running validation on such a contract call will be left for another ticket.

### 2.1 Acceptance Criteria:
   - [ ] We sign and submit a `rotate-keys` contract call after DKG using the aggregate key that DKG generated.

## 3. Related Issues and Pull Requests (optional):

This should follow up after https://github.com/stacks-network/sbtc/issues/590 gets done.

aldur:
@djordon please reword this so that it does _not_ include validation since that will be taken care of in #663.

matteojug:
Done in https://github.com/stacks-network/sbtc/pull/788

PR #788, opened 2024-11-07
Submit rotate-key tx

matteojug:
## Description

First part of: https://github.com/stacks-network/sbtc/issues/653
The second part is related to emily sidecar picking up the event and populating `rotate_keys_transactions` table; ~currently the coordinator keeps broadcasting rotate keys tx (that fails, after the first one) because it doesn't find any rotation in db~ (changed implementation, no longer true).

## Changes

Submit rotate key if the latest DKG differs from chain registry data.

## Testing Information

Tested in devenv and changed an integration test to check for the tx.

## Checklist:

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -42,6 +42,7 @@ use crate::stacks::api::SubmitTxResponse;
 use crate::stacks::contracts::AsTxPayload;
 use crate::stacks::contracts::CompleteDepositV1;
 use crate::stacks::contracts::ContractCall;
+use crate::stacks::contracts::RotateKeysV1;
 use crate::stacks::contracts::SmartContract;
 use crate::stacks::contracts::SMART_CONTRACTS;
 use crate::stacks::wallet::MultisigTx;
@@ -284,25 +285,57 @@ where
         tracing::debug!("We are the coordinator, we may need to coordinate DKG");
         // If Self::get_signer_set_and_aggregate_key did not return an
         // aggregate key, then we know that we have not run DKG yet. Since
-        // we are the signer, we should coordinate DKG.
+        // we are the coordinator, we should coordinate DKG.
         let aggregate_key = match maybe_aggregate_key {
             Some(key) => key,
             // This function returns the new DKG aggregate key.
-            None => self.coordinate_dkg(&bitcoin_chain_tip).await?,
+            None => {
+                self.coordinate_dkg(&bitcoin_chain_tip).await?;
+                let (maybe_aggregate_key, _) = self
+                    .get_signer_set_and_aggregate_key(&bitcoin_chain_tip)
+                    .await?;
+                maybe_aggregate_key.ok_or(Error::NoDkgShares)?
+            }
         };
 
+        let mut just_deployed_contracts = false;
         if !self.sbtc_contracts_deployed {
             let contract_deployments = self
                 .deploy_smart_contracts(&bitcoin_chain_tip, &aggregate_key)
                 .await;
             match contract_deployments {
-                Ok(()) => self.sbtc_contracts_deployed = true,
+                Ok(()) => {
+                    just_deployed_contracts = true;
+                    self.sbtc_contracts_deployed = true;
+                }
                 Err(error) => {
                     // No need to continue if the contracts are not deployed.
                     tracing::error!(%error, "could not deploy smart contracts");
                     return Err(error);
                 }
             };
+        }
+
+        // If we have just deployed the smart contracts, trying to broadcast the
+        // key rotation tx would result in transaction rejected, reason: NoSuchContract
+        if self.sbtc_contracts_deployed && !just_deployed_contracts {
+            let last_key_rotation = self
+                .context
+                .get_storage()
+                .get_last_key_rotation(&bitcoin_chain_tip)
+                .await?;
+            if last_key_rotation.is_none() {
+                // We don't have a key rotation but we have an aggregate_key from DKG,
+                // so we need to broadcast a rotate key tx
+                let wallet = self.get_signer_wallet(&bitcoin_chain_tip).await?;
+
+                self.construct_and_sign_rotate_key_transaction(
+                    &bitcoin_chain_tip,
+                    &aggregate_key,
+                    &wallet,
+                )
+                .await?;
+            }
         }
 
         self.construct_and_sign_bitcoin_sbtc_transactions(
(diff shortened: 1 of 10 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -305,6 +316,28 @@ where
             };
         }
 
+        // If we have just deployed the smart contracts, trying to broadcast the
+        // key rotation tx would result in transaction rejected, reason: NoSuchContract
+        if self.sbtc_contracts_deployed && !just_deployed_contracts {
```
cylewitruk, line 321: suggestion: i personally might have put this into a new function just to keep things concise (i probably would have done the same for the actual deployment above too but that was another PR)

cylewitruk, line 321: QQ: What's the thought on intended behavior if this transaction gets submitted but not mined by the next bitcoin block, then the next coordinator does the same, etc.? It won't be an RBF I guess because the signatures-portion of the stacks tx isn't deterministic in our case so the txid will be new -- is the idea that it just hangs around until garbage collected? and could that cause problems with nonces?

Just thinking out loud :thinking:

cylewitruk (commented):
Looks good to me, but i'll let daniel take a quick look too in-case there's something specific i'm not thinking of

signer/src/transaction_coordinator.rs
djordon, line 321: I think I would this check should be a little different. I think the check should be
1. whether the contract has been deployed and
2. whether the current aggregate key on the stacks node is the same as the latest one in the database.

The (2) matches what the signers should do during validation.

signer/src/transaction_coordinator.rs
```diff
@@ -284,19 +285,29 @@ where
...
         let aggregate_key = match maybe_aggregate_key {
             Some(key) => key,
             // This function returns the new DKG aggregate key.
-            None => self.coordinate_dkg(&bitcoin_chain_tip).await?,
+            None => {
+                self.coordinate_dkg(&bitcoin_chain_tip).await?;
+                let (maybe_aggregate_key, _) = self
+                    .get_signer_set_and_aggregate_key(&bitcoin_chain_tip)
+                    .await?;
+                maybe_aggregate_key.ok_or(Error::NoDkgShares)?
```
djordon, line 297: Ahh there is a race condition with the transaction signer. Well, I don't think that we should bail here, we could still deploy the smart contracts.

djordon (approved):
This looks good.

I made some suggestions, but they are all suggestions. I think we are due for a small refactor with the contract deploy and the rotate keys stuff in the `process_new_blocks` function. But my vote would be to do that in a follow-up PR.

signer/src/transaction_coordinator.rs
djordon: > ... is the idea that it just hangs around until garbage collected? and could that cause problems with nonces?

Hmmm, I don't think so, but I'm not sure. Generally speaking, if we have two transactions from the same account using the same nonce which have the same cost but have different fees then I think the miner will just choose the one with the greater fee. I would assume that Stacks nodes allow for transactions with the same nonce to exist in the mempool, but I don't know for sure. If the natural thing here is indeed the case (same cost, same nonce, different fees) are allowed then we should be okay.

cylewitruk: Yeah you're right, the code is [here](https://github.com/stacks-network/stacks-core/blob/40059a57cd27e740c5e9d91a833fb2c975b0bf0b/stackslib/src/core/mempool.rs#L2122) for RBF. Looks like it's only a check for address + nonce + fee is greater than previous. if fee isn't > last fee it will be rejected.

signer/src/transaction_coordinator.rs
matteojug: Yeah this is actually not solving the underlying issue, reverting the change.

signer/src/transaction_coordinator.rs
matteojug: In addition to what was said above, if for some strange timing reason we have two valid rotate keys tx (with different, valid nonces) then the second one will fail with `ERR_AGG_PUBKEY_REPLAY`.

signer/src/transaction_coordinator.rs
matteojug: Nvm, we may need this, added a comment about it (and removed the bailing).

signer/src/transaction_coordinator.rs
matteojug: Addressed in [0febb85](https://github.com/stacks-network/sbtc/pull/788/commits/0febb856000e877689aac10d7cb00002b1eff071)

matteojug: Refactored a bit in [0febb85](https://github.com/stacks-network/sbtc/pull/788/commits/0febb856000e877689aac10d7cb00002b1eff071)

signer/src/transaction_coordinator.rs
```diff
@@ -321,6 +330,48 @@ where
...
+        // If the latest DKG aggregate key matches on-chain data, nothing to do here
+        if Some(last_dkg.aggregate_key) == current_aggregate_key {
+            return Ok(());
+        }
+
+        let wallet = self.get_signer_wallet(bitcoin_chain_tip).await?;
+
+        self.construct_and_sign_rotate_key_transaction(
+            bitcoin_chain_tip,
+            aggregate_key,
```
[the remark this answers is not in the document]
matteojug: Right, and according to the docs of `StacksTransactionSignRequest` (where this end up) `It is used to identify the signing set for the transaction.`, so even if we just did a DKG with some new signers we want to use the "old" signing set to do the handoff. (I'll change it, with a fallback to `aggregate_key` if we have no `current_aggregate_key`).

Something I find confusing though is that, while we do validation right checking for dkg shares associated with that aggregate key, then when signing we use `SignerWallet::load`, that appears to load always the last key rotation. I guess this is going to work (probably almost) always, but it's based on some potentially non obvious assumptions.

signer/src/transaction_coordinator.rs
```diff
@@ -284,27 +285,35 @@ where
...
-                Err(error) => {
-                    // No need to continue if the contracts are not deployed.
-                    tracing::error!(%error, "could not deploy smart contracts");
-                    return Err(error);
-                }
-            };
+        if self
+            .deploy_smart_contracts(&bitcoin_chain_tip, &aggregate_key)
+            .await?
+        {
```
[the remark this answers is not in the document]
matteojug: Yeah makes sense, and the result is definitely cleaner. Refactored as suggested in [c026b1a](https://github.com/stacks-network/sbtc/pull/788/commits/c026b1a2d37f0d24b52d61720fc7c0f11556a34c).

djordon (approved):
Awesome, looks good ✅

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
djordon: Yeah, doing rotate keys correctly is not obvious, and I think this is partly because we use confusing terminology. Also, things are structured confusingly too. `SignerWallet::load` should probably be renamed or something too. Yeah, it's not the best.

Issue #583, opened 2024-09-30
[Feature]: Add sweep transaction ID to complete-deposit contract call


djordon:
# Feature - Add sweep transaction ID to complete-deposit contract call

## 1. Description

When we mint sBTC we currently only send what is absolutely necessary in the `complete-deposit` contract call. This includes the deposit outpoint, the amount of sBTC to mint, and the recipient of the minted sBTC. When https://github.com/stacks-network/sbtc/issues/488 gets done we'll add the block hash and block height of the sweep transaction to get protection against forks too. This ticket is to also add the bitcoin transaction ID of the sweep transaction to the `complete-deposit` contract call.

### 1.1 Context & Purpose

Although the signers need this transaction ID for validating a `complete-deposit` contract call transaction before it gets broadcast, it isn't needed in the contract call itself. So this transaction ID will only serves as a receipt for what the signers have done, allowing outside observers the "easily" validate the signers' actions. ~~It would also allow the signers to simplify their internal logic when trying to figure out which transactions they still need to send~~ (with #585 this statement is not really true).

## 2. Technical Details:

Update the `complete-deposit` to take a bitcoin transaction ID of the transaction that swept in the funds.

### 2.1 Acceptance Criteria:
   - [ ] Update the `complete-deposit` public function to take a bitcoin transaction ID of the transaction that swept in the funds.
   - [ ] Update the signer codebase to send this data.
   - [ ] The smart contract verifies that the provided ID is the correct length.

## 3. Related Issues and Pull Requests (optional):

https://github.com/stacks-network/sbtc/issues/488

PR #795, opened 2024-11-08
Add missing parameter to complete deposit call

matteojug:
## Description

Followup for https://github.com/stacks-network/sbtc/issues/583

## Changes

Add missing `sweep_txid` arg to complete deposit contract call.
I didn't fix `accept-withdrawal-request` since the error is probably more eloquent than using a `None` when we will need to address withdrawals; also, what should we put in there? Isn't it the same as `bitcoin-txid`?

## Testing Information

Tested in devenv with demo script.

## Checklist:


djordon (approved):
✅

Issue #797, opened 2024-11-08
[Feature]: Add rotate-keys event parsing in event observer


matteojug:
# Feature - Add rotate-keys event parsing in event observer 

## 1. Description

Events from `rotate-keys` are not processed by the stacks event observer.

### 1.1 Context & Purpose

After https://github.com/stacks-network/sbtc/issues/653, we now publish rotate keys tx. But the code to parse and handle such events in the event observer is missing (since we didn't have such events before). We need to parse those, and write the events to the rotate keys table.

## 3. Related Issues and Pull Requests (optional):

 - https://github.com/stacks-network/sbtc/issues/689
 - https://github.com/stacks-network/sbtc/issues/340
 - https://github.com/stacks-network/sbtc/issues/653

PR #799, opened 2024-11-08
add parsing and handling of rotate-keys events

Jiloc:
## Description 
Closes: #797 

## Changes
 - Add parsing of `rotate-keys` print envents in the `new_block_handler`
     - Write the rotate-keys tx to the db
 - Add `KeyRotation` to `RegistryEvent`
 - Add `KeyRotationEvent` struct
 - Add event parsing logic

## Testing Information
- add unit tests

## Checklist:

### DIFF
--- a/signer/src/stacks/events.rs
+++ b/signer/src/stacks/events.rs
@@ -23,13 +23,11 @@ use blockstack_lib::burnchains::Txid as StacksTxid;
 use clarity::vm::types::CharType;
 use clarity::vm::types::PrincipalData;
 use clarity::vm::types::SequenceData;
-use clarity::vm::types::SequencedValue;
 use clarity::vm::types::TupleData;
 use clarity::vm::ClarityName;
 use clarity::vm::Value as ClarityValue;
+use secp256k1::PublicKey;
 use stacks_common::types::chainstate::StacksBlockId;
-
-use crate::keys::PublicKey;
 
 /// An error when trying to parse an sBTC event into a concrete type.
 #[derive(Debug, thiserror::Error)]
@@ -54,7 +52,7 @@ pub enum EventError {
     /// This error is thrown when trying to convert a public key from a
     /// Clarity buffer into a proper public key. It should never be thrown.
     #[error("Could not convert a public key in clarity event into the expected public key {0}")]
-    ClarityPublicKeyConversion(#[source] crate::error::Error),
+    ClarityPublicKeyConversion(#[source] secp256k1::Error),
     /// This should never happen, but happens when one of the given topics
     /// is not on the list of expected topics.
     #[error("Got an unexpected event topic: {0}")]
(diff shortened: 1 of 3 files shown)


Jiloc:
I realized that the `rotate-keys` event now outputs additional data that we aren’t currently saving in the database. I was thinking of tracking it in the separate issue I am going to open to extend the database schema to capture the new fields from all print events.

signer/src/stacks/events.rs
```diff
@@ -23,11 +23,14 @@ use blockstack_lib::burnchains::Txid as StacksTxid;
 use clarity::vm::types::CharType;
 use clarity::vm::types::PrincipalData;
 use clarity::vm::types::SequenceData;
+use clarity::vm::types::SequencedValue;
 use clarity::vm::types::TupleData;
 use clarity::vm::ClarityName;
 use clarity::vm::Value as ClarityValue;
 use stacks_common::types::chainstate::StacksBlockId;
 
+use crate::keys::PublicKey;
```
djordon, line 32: I think we should change this to use the `secp256k1::PublicKey` type instead.

So, this file uses no types from the rest of the signer crate, it only uses rust-bitcoin types, stacks-core types, and bitvec types (and the rust-bitcoin crate has a dependency on the `secp256k1` crate so that is fine). The reason for us limiting the types used here is that we want to make it easy for people to build on sBTC without pulling in all of the dependencies that we have in the signer crate. 

Eventually we should move this module to the sBTC crate, but first we'd have to remove our use of the `blockstack_lib::burnchains::Txid`, which pulls in a lot of stuff. We may also want to remove our use of bitvec too.

signer/src/stacks/events.rs
```diff
@@ -255,6 +280,17 @@ impl RawTupleData {
         }
     }
 
+    /// Extract the list value from the given field
+    fn remove_list(&mut self, field: &'static str) -> Result<Vec<ClarityValue>, EventError> {
+        match self.data_map.remove(field) {
+            Some(ClarityValue::Sequence(SequenceData::List(mut list))) => {
+                Ok(list.drained_items())
```
djordon, line 287: I think we can do `list.data` here.

signer/src/stacks/events.rs
```diff
@@ -255,6 +280,17 @@ impl RawTupleData {
...
+                // Ok(list)
```
djordon, line 288: nit: we can remove this.

signer/src/stacks/events.rs
```diff
@@ -569,6 +605,51 @@ impl RawTupleData {
             signer_bitmap: BitArray::new(bitmap.to_le_bytes()),
         }))
     }
+
+    /// This function if for transforming the print events of the
```
djordon, line 609: nit: 
```suggestion
    /// This function is for transforming the print events of the
```
Looks like a copy pasta from my typos 😂.

signer/src/stacks/events.rs
```diff
@@ -783,6 +868,61 @@ mod tests {
         };
     }
 
+    #[test]
+    fn test_key_rotation_event() {
+        let new_keys: Vec<PublicKey> = vec![
+            fake::Faker.fake_with_rng(&mut OsRng),
+            fake::Faker.fake_with_rng(&mut OsRng),
+            fake::Faker.fake_with_rng(&mut OsRng),
+        ];
```
djordon, line 877: Oh generating public keys is a little more annoying, I'd generate a `secp256k1::Keypair` and get a public key through that.

djordon (commented):
This looks great, I had some small comments around a type change but looks good otherwise.

signer/src/stacks/events.rs
Jiloc: That's a great point! For now I only refactored with the use of `secp256k1::PublicKey`

signer/src/stacks/events.rs
Jiloc: Great suggestion thanks!

djordon (approved):
Looks good ✅ 🚢!

Issue #802, opened 2024-11-09
[Feature]: Implement validation of deposit fees in bitcoin transactions


djordon:
# Feature - Implement validation of deposit fees in bitcoin transactions

## 1. Description

Before a signer signs a bitcoin transaction they need to make sure that the fee associated with each  deposit transaction input is below the associated max fee. 

### 1.1 Context & Purpose

We need to respect the max-fee, and this is the signers' validation check.

### 2.1 Acceptance Criteria:
   - [ ] We have code that validates an assessed fee against the max-fee.

PR #803, opened 2024-11-09
feat: implement validation of deposit fees

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/802.

## Changes

* Add a function for validating the fees of a deposit that is used in a sweep transaction.
* Add a new `FeeAssessment` trait to make it easy to assess fees when we have other types.
* Add the max-fee to the `DepositRequestReport`.

## Testing Information

This PR adds unit tests and updates the integration tests to make sure that we've updated the query correctly.

## Checklist:


signer/src/bitcoin/utxo.rs
```diff
@@ -925,12 +928,21 @@ impl<'a> UnsignedTransaction<'a> {
...
-/// bitcoin transaction that has already been confirmed. This
-/// implementation is located in this module because it the assumptions for
-/// how the transaction is organized follows the logic in
+/// A trait for figuring out the fees assessed to deposit prevouts and
+/// withdrawal outputs in a bitcoin transaction.
+///
+/// This trait and the default implementations includes functions for
+/// apportioning fees to a bitcoin transaction that has already been
+/// confirmed. This implementation is located in this module because it the
+/// assumptions for how the transaction is organized follows the logic in
```
cylewitruk, line 937:
```suggestion
/// confirmed. This implementation is located in this module because the
/// assumptions it makes for how the transaction is organized follows the logic in
```

signer/src/bitcoin/utxo.rs
```diff
@@ -982,18 +993,18 @@ impl BitcoinTxInfo {
     ///
     /// The logic for the fee assessment is from
     /// <https://github.com/stacks-network/sbtc/issues/182>.
-    pub fn assess_output_fee(&self, vout: usize) -> Option<Amount> {
+    fn assess_output_fee(&self, vout: usize, fee: Amount) -> Option<Amount> {
```
cylewitruk, line 996: suggestion: I'd rename the `fee` parameter to `tx_fee` or something so that it's clearer as to _which_ fee this is.

Same above with `assess_input_fee()`

signer/src/bitcoin/utxo.rs
```diff
@@ -925,12 +928,21 @@ impl<'a> UnsignedTransaction<'a> {
...
+///
+/// This trait and the default implementations includes functions for
+/// apportioning fees to a bitcoin transaction that has already been
+/// confirmed. This implementation is located in this module because it the
+/// assumptions for how the transaction is organized follows the logic in
 /// [`UnsignedTransaction::new`].
-impl BitcoinTxInfo {
+pub trait FeeAssessment {
+    /// Returns all transaction inputs as a slice.
+    fn inputs(&self) -> &[TxIn];
```
cylewitruk, line 941: suggestion: I might have actually made a separate trait with `inputs()` and `outputs()` and had `FeeAssessment: InputsOutputs` or whatever, as that seems like it could be something generally useful.

signer/src/bitcoin/validation.rs
```diff
@@ -295,10 +298,33 @@ impl DepositRequestReport {
 
         Ok(())
     }
+
+    /// Validate that the fees assessed to the deposit prevout is below the
+    /// max fee.
+    pub fn validate_fee<F>(&self, tx: &F, tx_fee: u64) -> Result<(), BitcoinDepositInputError>
```
cylewitruk, line 304: suggestion: It would probably have been sufficient here to just use a `ValidateFeeResult` enum or something since we're not propagating any errors, and easier for the consumer to match on. Just a thought.

cylewitruk (commented):
Looks fine, just had a few suggestions

signer/src/bitcoin/utxo.rs
djordon: Great idea.

signer/src/bitcoin/validation.rs
djordon: Yeah that maybe be better. Let me have a look, I believe I have some code that does propagate these errors (but that code is in flux so maybe it's okay).

djordon: I left this as is for now, it's a little easier. But after we settle on how we'll use the code then it may make more sense to change.

signer/src/bitcoin/utxo.rs
djordon: Changed in https://github.com/stacks-network/sbtc/pull/803/commits/84f9fd7bc05073500cd61d46d0a567d77b4f6b18.

signer/src/bitcoin/utxo.rs
djordon: Updated in https://github.com/stacks-network/sbtc/pull/803/commits/84f9fd7bc05073500cd61d46d0a567d77b4f6b18

signer/src/bitcoin/utxo.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/803/commits/84f9fd7bc05073500cd61d46d0a567d77b4f6b18

matteojug (approved):

PR #807, opened 2024-11-09
chore: use nextest

djordon:
## Description

It looks like it will saves us some CI time.

## Changes

* Use nextest in CI

## Checklist:


Makefile
```diff
@@ -43,6 +43,10 @@ test: $(AUTOGENERATED_BLOCKLIST_CLIENT_CLIENT) $(AUTOGENERATED_EMILY_CLIENT)
 	cargo test -- --test-threads=1
 	pnpm --recursive test
 
+test-ci: $(AUTOGENERATED_BLOCKLIST_CLIENT_CLIENT) $(AUTOGENERATED_EMILY_CLIENT)
+	cargo nextest run --no-fail-fast
```
matteojug, line 47: Using nextest locally I get some flaky failures for `network::libp2p::network::tests::swarm_rejects_connections_from_unknown_peers` withouth `--test-threads 1`. I guess here it may work because of CI runner specs, but should we mark that test as integration to avoid random failures?

cylewitruk: We should probably move these tests to `integration` and have the `cfg_attr` thing on them, because if they're run in parallel then they're discovering eachother via mDNS and breaking the assumptions of the individual tests, which is much more likely to happen when `nextest` is running them as it's more aggressive in its parallelism by default.

matteojug (approved):

Issue #809, opened 2024-11-10
[Feature]: Get last fee for bitcoin transactions (replaces #541)


cylewitruk:
# Feature - Get last fee for bitcoin transactions (replaces #541)

## 1. Description

Before we can replace-by-fee (RBF) a bitcoin transaction, we need to fetch the fee that was paid for the last transaction *package* that is currently in the mempool.

### 1.1 Context & Purpose

In order to RBF a transaction you need to know how much you paid both in total and per vbyte in order to know whether the new fee would be accepted. Since we could be replacing an entire package, we need to know the fees paid for all signer transactions in the mempool.

## 2. Technical Details:

The last transaction package can be retrieved using information in the `sweep_transactions` table. The first transaction in the transaction package "chain" can be identified by using the current signer UTXO; the `signer_prevout_txid` will be UTXO's `txid`. From there a recursive query over `sweep_transactions` can be used to find the rest of the transactions in the chain.

The `SbtcRequests` has a property `signer_state` of the type `SignerBtcState` which is populated in `transaction_coordinator::get_pending_requests()`. The current implementation expects to use the `BitcoinInteract::get_last_fees()` method, however this should be changed to calculate the last fees based on information in the `sweep_transactions` table.

The resulting `Fees` object has two fields which should be calculated as follows:

- `total`: the _total_ amount of fees for all of the transactions in the package.
- `rate`: the `total` / the sum of `vbytes` of the transactions in the package.

### 2.1 Acceptance Criteria:

- [ ]  The "previous fees" are calculated based on information in the `sweep_transactions` table.

## 3. Related Issues and Pull Requests (optional):

- Replaces #541

PR #810, opened 2024-11-10
feat: get last bitcoin fees from sweep tables (#809)

cylewitruk:
## Description

Closes: #809

## Changes

Removes the `BitcoinInteract::get_last_fees()` (described in #541, replaced by #809) and replaces it with logic to retrieve the `Fees` from the latest sweep transaction package.

## Testing Information

RBF tests updated to use the `sweep_transactions` method.

## Checklist:

### DIFF
--- a/signer/migrations/0003__create_tables.sql
+++ b/signer/migrations/0003__create_tables.sql
@@ -196,6 +196,8 @@ CREATE TABLE sbtc_signer.sweep_transactions (
     amount BIGINT NOT NULL,
     -- The fee paid for the transaction.
     fee BIGINT NOT NULL,
+    -- The _vsize_ of the transaction in bytes.
+    vsize INTEGER NOT NULL,
     -- The Bitcoin block hash at which this package was created.
     created_at_block_hash BYTEA NOT NULL,
     -- The Bitcoin market fee rate at the time this package was created.
--- a/signer/src/message.rs
+++ b/signer/src/message.rs
@@ -151,6 +151,8 @@ pub struct SweepTransactionInfo {
     pub amount: u64,
     /// The fee paid for this transaction.
     pub fee: u64,
+    /// The virtual size of this transaction (in bytes).
+    pub vsize: u32,
     /// The Bitcoin block hash at which this transaction was created.
     pub created_at_block_hash: bitcoin::BlockHash,
     /// The market fee rate at the time of this transaction.
@@ -164,6 +166,10 @@ pub struct SweepTransactionInfo {
 impl SweepTransactionInfo {
     /// Creates a [`SweepTransactionInfo`] from an [`UnsignedTransaction`] and a
     /// Bitcoin block hash.
+    ///
+    /// **Note that it is important that this function is called after the
+    /// transaction has been signed**, as otherwise the `vbytes` size will not
+    /// be correct.
     pub fn from_unsigned_at_block(
         block_hash: &bitcoin::BlockHash,
         unsigned: &crate::bitcoin::utxo::UnsignedTransaction,
@@ -208,6 +214,8 @@ impl SweepTransactionInfo {
                 .signers_script_pubkey(),
             amount: unsigned.output_amounts(),
             fee: unsigned.tx_fee,
+            // the vsize should never be larger than u32::MAX
+            vsize: unsigned.tx.vsize() as u32,
             market_fee_rate: unsigned.signer_utxo.fee_rate,
             created_at_block_hash: *block_hash,
             swept_deposits,
--- a/signer/src/storage/in_memory.rs
+++ b/signer/src/storage/in_memory.rs
@@ -796,6 +796,31 @@ impl super::DbRead for SharedStore {
 
         Ok(package)
     }
+
+    async fn get_sweep_transaction_package(
+        &self,
+        prevout_txid: &model::BitcoinTxId,
+    ) -> Result<Vec<model::SweepTransaction>, Error> {
+        let store = self.lock().await;
+        let sweep_txs = &store.sweep_transactions;
+        let first = sweep_txs
+            .iter()
+            .find(|tx| tx.signer_prevout_txid == *prevout_txid);
+
+        if first.is_none() {
+            return Ok(vec![]);
+        }
+
+        let package = std::iter::successors(first, |sweep| {
+            sweep_txs
+                .iter()
+                .find(|tx| tx.signer_prevout_txid == sweep.txid)
+        })
+        .cloned()
+        .collect::<Vec<_>>();
+
+        Ok(package)
+    }
 }
 
 impl super::DbWrite for SharedStore {
@@ -1076,5 +1101,42 @@ impl super::DbWrite for SharedStore {
         store.sweep_transactions.push(tx.clone());
 
         Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::storage::DbRead as _;
+    use fake::{Fake, Faker};
+    use model::*;
+    use rand::rngs::OsRng;
+
+    #[tokio::test]
+    async fn can_get_sweep_transaction_package() {
+        let mut txs: Vec<SweepTransaction> = Faker.fake_with_rng(&mut OsRng);
+        let txids = txs.iter().map(|tx| tx.txid).collect::<Vec<_>>();
+        let utxo_txid: BitcoinTxId = Faker.fake_with_rng(&mut OsRng);
+
+        for (i, tx) in txs.iter_mut().enumerate() {
+            if i == 0 {
+                tx.signer_prevout_txid = utxo_txid;
+            } else {
+                tx.signer_prevout_txid = txids[i - 1];
+            }
+        }
+
+        let store = Store::new_shared();
+        {
+            let mut store = store.lock().await;
+            store.sweep_transactions = txs.clone();
+        }
+
+        let package = store
+            .get_sweep_transaction_package(&utxo_txid)
+            .await
+            .unwrap();
+
+        assert_eq!(package, txs);
     }
 }
--- a/signer/src/storage/mod.rs
+++ b/signer/src/storage/mod.rs
@@ -246,6 +246,13 @@ pub trait DbRead {
         chain_tip: &model::BitcoinBlockHash,
         context_window: u16,
     ) -> impl Future<Output = Result<Option<model::SweepTransaction>, Error>> + Send;
+
+    /// Get the sweep transaction package for the given previous transaction id
+    /// (i.e. the previous output that was spent from the then-UTXO).
+    fn get_sweep_transaction_package(
+        &self,
+        prevout_txid: &model::BitcoinTxId,
+    ) -> impl Future<Output = Result<Vec<model::SweepTransaction>, Error>> + Send;
 }
 
 /// Represents the ability to write data to the signer storage.
--- a/signer/src/storage/model.rs
+++ b/signer/src/storage/model.rs
@@ -11,6 +11,8 @@ use serde::Deserialize;
 use serde::Serialize;
 use stacks_common::types::chainstate::{BurnchainHeaderHash, StacksBlockId};
 
+use crate::bitcoin::utxo;
+use crate::bitcoin::utxo::Fees;
 use crate::block_observer::Deposit;
 use crate::error::Error;
 use crate::keys::PublicKey;
@@ -43,6 +45,10 @@ pub struct SweepTransaction {
     #[sqlx(try_from = "i64")]
     #[cfg_attr(feature = "testing", dummy(faker = "0..i32::MAX as u64"))]
     pub fee: u64,
+    /// The virtual size of this transaction (in bytes).
+    #[sqlx(try_from = "i32")]
+    #[cfg_attr(feature = "testing", dummy(faker = "0..i32::MAX as u32"))]
+    pub vsize: u32,
     /// The Bitcoin block hash at which this transaction was created.
     pub created_at_block_hash: BitcoinBlockHash,
     /// The market fee rate at the time of this transaction.
@@ -76,11 +82,23 @@ impl From<&crate::message::SweepTransactionInfo> for SweepTransaction {
             signer_prevout_script_pubkey: info.signer_prevout_script_pubkey.clone().into(),
             amount: info.amount,
             fee: info.fee,
+            vsize: info.vsize,
             market_fee_rate: info.market_fee_rate,
             created_at_block_hash: info.created_at_block_hash.into(),
             swept_deposits: info.swept_deposits.iter().map(Into::into).collect(),
             swept_withdrawals: info.swept_withdrawals.iter().map(Into::into).collect(),
         }
+    }
+}
+
+impl utxo::GetFees for Vec<SweepTransaction> {
+    /// Return the total fee of all the transactions in the vector.
+    fn get_fees(&self) -> Fees {
+        let total = self.iter().map(|tx| tx.fee).sum();
+        let total_size: u32 = self.iter().map(|tx| tx.vsize).sum();
+        let rate: f64 = total as f64 / total_size as f64;
+
+        Fees { total, rate }
     }
 }
 
@@ -962,6 +980,9 @@ pub type Bytes = Vec<u8>;
 mod tests {
     use fake::Fake;
     use rand::SeedableRng;
+    use test_case::test_case;
+
+    use crate::bitcoin::utxo::GetFees;
 
     use super::*;
 
@@ -978,5 +999,28 @@ mod tests {
         let block_hash = BitcoinBlockHash::from(stacks_hash);
         let round_trip = BurnchainHeaderHash::from(block_hash);
         assert_eq!(stacks_hash, round_trip);
+    }
+
+    #[test_case(&[(1000, 500)], Fees { total: 500, rate: 0.5 })]
+    #[test_case(&[(1000, 500), (2000, 1000)], Fees { total: 1500, rate: 0.5 })]
+    #[test_case(&[(1000, 250), (2000, 1000)], Fees { total: 1250, rate: 0.4166666666666667 })]
+    #[test_case(&[(1000, 125), (1250, 125), (1500, 175)], Fees { total: 425, rate: 0.11333333333333333 })]
+    fn get_sweep_transaction_package_fees(sweeps: &[(u32, u64)], expected: Fees) {
+        // (vsize, fee)
+        let mut rng = rand::rngs::StdRng::seed_from_u64(1);
+
+        let mut sweep_txs = vec![];
+        for (vsize, fee) in sweeps {
+            let tx = SweepTransaction {
+                vsize: *vsize,
+                fee: *fee,
+                ..fake::Faker.fake_with_rng(&mut rng)
+            };
+            sweep_txs.push(tx);
+        }
+
+        let fees = sweep_txs.get_fees();
+
+        assert_eq!(fees, expected);
     }
 }
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -420,6 +420,7 @@ impl PgStore {
               , signer_prevout_script_pubkey
               , amount
               , fee
+              , vsize
               , created_at_block_hash
               , market_fee_rate
             FROM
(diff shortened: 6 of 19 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -351,12 +352,6 @@ where
         let transaction_package = pending_requests.construct_transactions()?;
 
         for mut transaction in transaction_package {
-            self.send_message(
```
cylewitruk, line 354: This is moved now because we need to store the size of the transaction after signing, which happens in `sign_and_broadcast()`.

signer/src/transaction_coordinator.rs
```diff
@@ -678,12 +673,20 @@ where
...
-
+        // Broadcast the transaction to the Bitcoin network.
         self.context
             .get_bitcoin_client()
             .broadcast_transaction(&transaction.tx)
             .await?;
 
+        // Publish the transaction to the P2P network so that peers get advance
+        // knowledge of the sweep.
+        self.send_message(
```
cylewitruk, line 684: Tbh I'm not sure which is better, publishing this before or after the bitcoin broadcast.. :thinking:

signer/migrations/0003__create_tables.sql
```diff
@@ -196,6 +196,8 @@ CREATE TABLE sbtc_signer.sweep_transactions (
     amount BIGINT NOT NULL,
     -- The fee paid for the transaction.
     fee BIGINT NOT NULL,
+    -- The _vsize_ of the transaction in bytes.
```
djordon, line 199: I would rephrase this to say something like 'The bitcoin "virtual size" of the transaction'. I don't think that saying that it is in bytes is incorrect.

signer/src/message.rs
```diff
@@ -164,6 +166,10 @@ pub struct SweepTransactionInfo {
 impl SweepTransactionInfo {
     /// Creates a [`SweepTransactionInfo`] from an [`UnsignedTransaction`] and a
     /// Bitcoin block hash.
+    ///
+    /// **Note that it is important that this function is called after the
+    /// transaction has been signed**, as otherwise the `vbytes` size will not
+    /// be correct.
```
djordon, line 172: When creating an `UnsignedTransaction` we compute what the `vsize` would be if it were signed, but then throw that value away since we didn't need it at the time. We could keep it around if we want to.

signer/src/message.rs
```diff
@@ -208,6 +214,8 @@ impl SweepTransactionInfo {
                 .signers_script_pubkey(),
             amount: unsigned.output_amounts(),
             fee: unsigned.tx_fee,
+            // the vsize should never be larger than u32::MAX
```
djordon, line 217: Oh because the max block vsize is currently `u32::MAX` bytes, right?

signer/tests/integration/rbf.rs
```diff
@@ -188,7 +192,7 @@ pub fn transaction_with_rbf(
         })
         .collect();
 
-    faucet.generate_blocks(1);
+    let block_hash = faucet.generate_blocks(1);
```
djordon, line 195: nit: I would add a `.pop().unwrap()` here, since it makes reading line 260 a little easier.

signer/migrations/0003__create_tables.sql
```diff
@@ -196,6 +196,8 @@ CREATE TABLE sbtc_signer.sweep_transactions (
...
+    vsize INTEGER NOT NULL,
```
djordon, line 200: We should make this a `BIGINT` since the max is `u32::MAX`. Hmmm, it feels like a waste of space, but it is more correct.

signer/src/storage/postgres.rs
```diff
@@ -1749,6 +1750,60 @@ impl super::DbRead for PgStore {
...
+                    sweep_txs last 
+                        ON tx.signer_prevout_txid = last.txid
+            )
+            SELECT
+                txid
+              , number
+            FROM
+                sweep_txs
+            ORDER BY
+                number ASC;
```
djordon, line 1788: Doesn't this find all transactions after the `prevout_txid`? I mean, including the ones issued as part of subsequent transaction packages?

djordon: Also, this returns the transaction that includes the given txid. This isn't bad per se, but when computing the last fee we'll include a transaction that has been confirmed, which will give us a bad last fee. And if we have RBF'ed a transaction package more than once, this will include all packages not just the one that we care about, which is the last one.

djordon (commented):
This looks okay, but I don't think the query does what we want (or we need to change our usage).

signer/src/storage/postgres.rs
cylewitruk: Yeah you're right, there's some funk here. I'll add a couple negative tests to ensure those scenarios can't happen and update the query.

signer/src/message.rs
cylewitruk: That and that it would be completely insane if we had a transaction of this size; max bitcoin block size is ~4mb.

signer/src/storage/postgres.rs
cylewitruk: I updated the query which I think solves the above issues (and current tests pass), but I still need to write the negative tests

signer/migrations/0003__create_tables.sql
cylewitruk: Yeah you're right, since witness data is 1/4 weight

signer/src/message.rs
cylewitruk: Ah good to know, I'll make that change

cylewitruk: Updated in [a399760](https://github.com/stacks-network/sbtc/pull/810/commits/a39976075a79f2c064b214d368485b787a1139e0)

signer/src/message.rs
cylewitruk: This is changed in [a399760](https://github.com/stacks-network/sbtc/pull/810/commits/a39976075a79f2c064b214d368485b787a1139e0) from the above comment

signer/tests/integration/rbf.rs
cylewitruk: Updated in [9b38cb1](https://github.com/stacks-network/sbtc/pull/810/commits/9b38cb1d33633cbc3655b95e53caa443d6e9ed16)

signer/migrations/0003__create_tables.sql
cylewitruk: [77d4884](https://github.com/stacks-network/sbtc/pull/810/commits/77d4884094bc91a71e5c7ee44a96cc22b72b895e)

signer/migrations/0003__create_tables.sql
djordon: Hmm, yeah. And we should probably cap it ourselves anyway.

signer/src/storage/mod.rs
```diff
@@ -271,6 +271,13 @@ pub trait DbRead {
         chain_tip: &model::BitcoinBlockHash,
         context_window: u16,
     ) -> impl Future<Output = Result<Option<model::SweepTransaction>, Error>> + Send;
+
+    /// Get the sweep transaction package for the given previous transaction id
+    /// (i.e. the previous output that was spent from the then-UTXO).
```
djordon, line 276: I would add a little here and maybe rephrase it. I would say something like "This function takes the transaction ID of the signers' unspent output on the canonical Bitcoin blockchain and returns the chain of mempool transactions that spend that output." And maybe add something like this "The provided txid refers to an on-chain unspent output created by the signers, which may be spent by a transaction currently in the mempool."

Man that was wordy.

signer/src/transaction_coordinator.rs
```diff
@@ -950,13 +953,28 @@ where
...
+            .get_storage()
+            .get_sweep_transaction_package(&utxo.outpoint.txid.into())
+            .await?;
+
+        // If the last sweep package is empty, then we don't have any fees to
+        // report, so we return None. Otherwise, we return the calculated fees
+        // from the last sweep package.
+        let last_fees = last_sweep_package
+            .is_empty()
+            .then(|| last_sweep_package.get_fees());
```
djordon, line 977: I think this is backwards. If `last_sweep_package.is_empty()` then `last_fees` should be `None`

djordon (commented):
I think there is only one issue with the `last_fee` part in the coordinator. Looks good otherwise.

signer/src/transaction_coordinator.rs
cylewitruk: I wasn't done! :stuck_out_tongue_closed_eyes: But yeah you're right, refactoring gone bad. But now there's tests that catch this and verify the whole `get_btc_state()` function both when sweeps exist and when they don't.
[6cf7d78](https://github.com/stacks-network/sbtc/pull/810/commits/6cf7d789d6db074533b7484aca5902d9eb310a99)

signer/src/storage/mod.rs
cylewitruk: Updated the comments a bit in [0d4aed9](https://github.com/stacks-network/sbtc/pull/810/commits/0d4aed96c6ec3706a382ee98797298178da90c9f).

While what you describe is the _intended use_, it's not really true for the provided functionality of the method. It purposely places the burden of making sure you're providing the correct `prevout_txid` on the caller and thus itself provides no guarantees of canonicality or that the `prevout_txid` is indeed confirmed.

cylewitruk: Actually renamed the method in [413ae4c](https://github.com/stacks-network/sbtc/pull/810/commits/413ae4ce3d7195a326e8a9695866d605d5ec9963) to better reflect what it does

signer/src/transaction_coordinator.rs
djordon: Oh whoops, I didn't check to see if you were done yet.

djordon (approved):
Looks good ✅ 🚢. We should check if the integration tests in the `transaction_coordinator` run in CI.

djordon:
I decided to go and check myself and I didn't find those tests in the CI output logs. Also, we should have CI explicitly output which tests were ignored.

signer/src/transaction_coordinator.rs
```diff
@@ -1380,4 +1413,295 @@ mod tests {
     async fn should_get_signer_utxo_donations() {
         test_environment().assert_get_signer_utxo_donations().await;
     }
+
+    /// This test asserts that the `get_btc_state` function returns the correct
+    /// `SignerBtcState` when there are no sweep transactions available, i.e.
+    /// the `last_fees` field should be `None`.
+    #[cfg_attr(not(feature = "integration-tests"), ignore)]
+    #[test(tokio::test)]
```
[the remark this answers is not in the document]
cylewitruk: It's from `test_log::test` -- lets you run the test with `RUST_LOG=xx` and get actual logs output. You'll find it on a number of tests which I've needed to debug :sweat_smile:

signer/src/transaction_coordinator.rs
```diff
@@ -1380,4 +1413,295 @@ mod tests {
     async fn should_get_signer_utxo_donations() {
         test_environment().assert_get_signer_utxo_donations().await;
     }
+
+    /// This test asserts that the `get_btc_state` function returns the correct
+    /// `SignerBtcState` when there are no sweep transactions available, i.e.
+    /// the `last_fees` field should be `None`.
+    #[cfg_attr(not(feature = "integration-tests"), ignore)]
```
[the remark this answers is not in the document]
cylewitruk: @djordon hmm, I'm not sure of the best approach. Here we're testing `get_btc_state()` which is private, so it needs to be here, but otoh it also needs postgres, so it needs to be run as an integration-test. Should I just make it public and move the test, or do you have a better suggestion?

signer/src/storage/in_memory.rs
```diff
@@ -810,6 +810,31 @@ impl super::DbRead for SharedStore {
 
         Ok(package)
     }
+
+    async fn get_latest_unconfirmed_sweep_transactions(
+        &self,
+        prevout_txid: &model::BitcoinTxId,
+    ) -> Result<Vec<model::SweepTransaction>, Error> {
```
matteojug, line 817: nit: wrt the pg one (and the trait docstring), this doesn't filter out confirmed ones IIUC. I guess it's fine, but maybe we should add a comment about it and revisit later?

signer/src/storage/model.rs
```diff
@@ -84,6 +91,17 @@ impl From<&crate::message::SweepTransactionInfo> for SweepTransaction {
     }
 }
 
+impl utxo::GetFees for Vec<SweepTransaction> {
+    /// Return the total fee of all the transactions in the vector.
+    fn get_fees(&self) -> Fees {
+        let total = self.iter().map(|tx| tx.fee).sum();
+        let total_size: u32 = self.iter().map(|tx| tx.vsize).sum();
```
matteojug, line 98: qq: do we risk to overflow when summing `vsize`s?

signer/src/storage/model.rs
```diff
@@ -84,6 +91,17 @@ impl From<&crate::message::SweepTransactionInfo> for SweepTransaction {
...
+        let rate: f64 = total as f64 / total_size as f64;
```
matteojug, line 99: should we add a `total_size == 0` check here? I guess it shouldn't happen, merely as a safety guard

signer/src/storage/postgres.rs
```diff
@@ -1749,6 +1750,88 @@ impl super::DbRead for PgStore {
...
+        // chained from the first one. As a precaution, this query also filters
+        // out any sweep transactions that have already been confirmed on the
+        // Bitcoin blockchain.
+        let infos: Vec<(model::BitcoinTxId, i32)> = sqlx::query_as(
+            "
+            WITH RECURSIVE sweep_txs AS (
+                SELECT
+                    txid
+                  , signer_prevout_txid
+                  , 1 AS number
```
matteojug, line 1792: nit: do we need the number?

signer/src/storage/postgres.rs
```diff
@@ -1749,6 +1750,88 @@ impl super::DbRead for PgStore {
...
+                FROM sweep_transactions tx
+                INNER JOIN sweep_txs last 
+                    ON tx.signer_prevout_txid = last.txid
+            )
+            SELECT
+                sweep_txs.txid
+              , number
+            FROM sweep_txs
+            LEFT JOIN bitcoin_transactions AS btc_tx 
+                ON btc_tx.txid = sweep_txs.txid
```
matteojug, line 1812: What happens in case of forks here? eg a sweep tx is confirmed on a block, but then we get two block on another fork (making this second one the canonical one): the sweep tx may still be in the mempool for this second (canonical) fork, but here we would exclude it.

djordon:
> @djordon hmm, I'm not sure of the best approach. Here we're testing `get_btc_state()` which is private, so it needs to be here, but otoh it also needs postgres, so it needs to be run as an integration-test. Should I just make it public and move the test, or do you have a better suggestion?

Yeah I'd make it public.

signer/src/storage/model.rs
cylewitruk: In theory yes, but then we're up to 4MB worth of transaction data (well, roughly -- witness data is discounted to 1/4). So in practical terms, this should never happen.

But it's a good point, I'll refactor this a little to return a result, I guess it could potentially happen with `Fake`'d stuff.

signer/src/storage/postgres.rs
djordon: Ahhh, good point... we need to join on the canonical bitcoin blockchain... nuts.

signer/src/storage/model.rs
cylewitruk: Right, I don't think it's even possible to create a zero-sized `bitcoin::Transaction` -- but I'll toss that in while I'm doing the above refactor just so it's 100% sound.

signer/src/storage/in_memory.rs
cylewitruk: The
```sql
LEFT JOIN bitcoin_transactions AS btc_tx 
    ON btc_tx.txid = sweep_txs.txid
WHERE btc_tx.txid IS NULL
```
should take care of that wrt the sweep transactions themselves. If you're thinking about the `prevout_txid` which you pass in, it's expected that you as the caller have done the work to ensure that that transaction you're referring to is correct.

signer/src/network/in_memory2.rs
```diff
@@ -83,10 +87,22 @@ impl SignerNetwork {
...
+    /// Create a new in-memory signer network with a single signer instance.
+    /// You can use this if you do not need to simulate multiple signers.
+    pub fn single() -> Self {
+        let (wan_tx, _) = tokio::sync::broadcast::channel(DEFAULT_WAN_CAPACITY);
+        let (signer_tx, _) = tokio::sync::broadcast::channel(DEFAULT_SIGNER_CAPACITY);
+        Self {
+            wan_tx,
+            signer_tx,
+            sent: Arc::new(RwLock::new(VecDeque::new())),
+        }
```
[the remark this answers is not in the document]
cylewitruk: Haha, yes that works too :angel: [f9c4ead](https://github.com/stacks-network/sbtc/pull/810/commits/f9c4ead98c5c5d68bec6ccae0ce37b31b0c45354)

signer/src/storage/in_memory.rs
matteojug: Uhm I'm not sure I'm following: here I was referring to the in-memory impl not filtering out the confirmed sweep tx, while the pg does it (correctly, according to the docstring). Am I missing something?

signer/src/storage/postgres.rs
matteojug: Ah I missed that, now it makes sense 😅

signer/src/storage/postgres.rs
cylewitruk: Welp... I guess sooner or later someone was going to have to write a canonical function for transactions... :tada:

signer/src/storage/in_memory.rs
cylewitruk: Ohh sorry I thought you meant with regards to the PG one. Right, i'll have to check the in-mem one or just mark it `unimplemented!()`, I think I forgot to circle back to that impl after the latest updates.

signer/src/storage/model.rs
cylewitruk: That too, so good catch

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
djordon: 👍🏽 good to know.

signer/src/storage/model.rs
cylewitruk: Refactored in [8018721](https://github.com/stacks-network/sbtc/pull/810/commits/8018721e7104fdbe5b368af70835ffbc23f437c5)

signer/src/storage/model.rs
cylewitruk: Added in [8018721](https://github.com/stacks-network/sbtc/pull/810/commits/8018721e7104fdbe5b368af70835ffbc23f437c5)

signer/src/storage/model.rs
```diff
@@ -93,12 +93,22 @@ impl From<&crate::message::SweepTransactionInfo> for SweepTransaction {
...
     /// Return the total fee of all the transactions in the vector.
-    fn get_fees(&self) -> Fees {
-        let total = self.iter().map(|tx| tx.fee).sum();
-        let total_size: u32 = self.iter().map(|tx| tx.vsize).sum();
+    fn get_fees(&self) -> Result<Fees, Error> {
+        let total: u64 = self
+            .iter()
+            .map(|tx| tx.fee)
+            .try_fold(0u64, |acc, fee| acc.checked_add(fee))
+            .ok_or(Error::ArithmeticOverflow)?;
```
djordon, line 101: Hmmm, these fees cannot add up to anything more than `u64::MAX`, even if we added up all bitcoins (measured in sats).

signer/src/storage/model.rs
```diff
@@ -93,12 +93,22 @@ impl From<&crate::message::SweepTransactionInfo> for SweepTransaction {
...
+
+        let total_size: u64 = self.iter().map(|tx| tx.vsize as u64).sum();
+
+        if total_size == 0 {
+            return Err(Error::DivideByZero);
+        }
```
djordon, line 111: This can only happen if the vector is empty, so we could just return zero.

signer/src/storage/model.rs
cylewitruk: Nope, but the code otherwise is _technically_ not panic-safe, which is what we've been striving for.

signer/src/storage/model.rs
cylewitruk: I don't think zero is what we want -- `compute_transaction_fee()` matches on an `Option<>` and `None` would not be the same as `Fees::ZERO`. 

But I changed this to return `None` ([891e46a](https://github.com/stacks-network/sbtc/pull/810/commits/891e46aa7ce8aaf05a0fb7e87912aca2a5e96b05)) if `total_size` is zero since I could describe that mode in the trait description and it matches its use. Yeah it should only happen if the vec is empty in prod code, but it's still not panic-safe by simply checking `self.len()` (you could create items all of which have `vsize == 0`).

signer/src/storage/postgres.rs
```diff
@@ -1701,6 +1702,99 @@ impl super::DbRead for PgStore {
 
         self.get_sweep_transaction_by_txid(&txid).await
     }
+
+    async fn get_latest_unconfirmed_sweep_transactions(
+        &self,
+        chain_tip: &model::BitcoinBlockHash,
+        context_window: u16,
+        prevout_txid: &model::BitcoinTxId,
+    ) -> Result<Vec<model::SweepTransaction>, Error> {
```
matteojug, line 1711: nit: this fn may return a suffix of the chain of sweep tx starting from `prevout_txid`, if some prefix of the chain is confirmed. The result is that the first sweep tx of the vec may not be spending `prevout_txid`. I guess it's fine with its current usage, but I wanted to point it out in case we plan to rely on that assumption (feel free to resolve this convo).

matteojug (approved):
LGTM, but I think you should revert the changes in generated code (`make lint` should do the trick).

cylewitruk:
@matteojug yeah, crap, those generated files snuck in on me. `make lint` did indeed seem to resolve it :+1:

signer/src/storage/model.rs
djordon: Great points!

signer/tests/integration/postgres.rs
```diff
@@ -2311,6 +2314,246 @@ async fn can_store_and_get_latest_sweep_transaction() {
...
+    assert_eq!(sweep_transactions.len() - 1, sweeps.len());
+    assert_eq!(&sweep_transactions[1..], &sweeps[..]);
+
+    // =========================================================================
+    // Step 5 - create a fork from block 1, competing with block "2a"
+    // -------------------------------------------------------------------------
+    // We create two new Bitcoin blocks, chained off of the first Bitcoin block
+    // and thus competing with block 2a. By creating two new blocks, this chain
+    // now becomes the new canonical (longest) chain.
+    // =========================================================================
```
djordon, line 2518: I like this style of comment a lot. 👍🏽

djordon (approved):

Issue #791, opened 2024-11-07
[Design]: Cap on max sBTC pegged per transaction / account


aldur:
***Completing the issue description and arriving at a conclusion is the deliverable of this issue.***

# Design - Cap on max sBTC pegged per transaction / account

If #562 is implemented, then a whale might take over the full amount of sBTC available for peg-in. 
This would be detrimental to other users, DeFi, etc.

## 2. Context & Purpose

***Relevant Research Discussions***

***External Resources***

### 3.1 Proposed Component Design

Similarly to #562, Signers could discretionarily refuse peg-in requests over a given threshold. This would not solve the sybil attack, nor user peg-ins spanning multiple transactions. The implementation of the Signer binary can expose such feature through a configuration option.

### 3.1.2 Considerations & Alternatives

As for #562, the max amount could be written on-chain for public display.

### 3.2 Areas of Ambiguity

---

**Closing Checklist**
- [ ] The design proposed in this issue is clearly documented in the description of this ticket.
- [ ] Everyone necessary has reviewed the resolution and agrees with the proposal.
- [ ] This ticket has or links all the information necessary to familiarize a contributor with the design decision, why it was made, and how it'll be included.

djordon:
Capping the maximum of sBTC pegged in per transaction is simple to implement in the signers. We can do it per account too, but it's a bit more annoying.

cylewitruk:
My concern with the account-level cap is the tracking of withdrawals. If the user has sent/spent the sBTC then should their current peg be reduced? If not, then they may not have any sBTC left to peg-out and thus reduce their peg-in limit and be able to peg-in more.

Otoh, it's also really easy to just do multiple peg-ins to multiple addresses and consolidate thereafter.

So the question I have is if this is even feasible (without some advanced tracking of sBTC across the entire blockchain)?

andrerserrano:
@cylewitruk an account-level cap is not the intent of this issue. The intention of this issue is simply meant to encourage broad participation in the sBTC launch. The idea is to prevent a single entity from buying out one Bitcoin block and taking the entire cap (roughly 1,000 BTC), squeezing everyone else out in the process. By creating a mechanism that requires at least several Bitcoin blocks to fill, it may result in wider participation and increase the overall distribution of sBTC at launch.

cylewitruk:
@andrerserrano yeah I get the high-level goal, but this issue specifically mentions an account-level cap so I was just giving my spontaneous thoughts on that particular point.

[...]

Issue #562, opened 2024-09-24
[Design]: Iterative Cap on Pegged sBTC


AshtonStephens:
# Design - Iterative Cap on Pegged sBTC

This ticket holds the design of Iterative Cap on Pegged sBTC and how it fits into sBTC-v1.

## 1. Summary

TL;DR Both the GTM strategy and a slow-rollout strategy would benefit from a cap on the maximum amount of sBTC that can be pegged-in. The cap can iteratively be increased and, eventually, lifted.

## 2. Context & Purpose

***Relevant Research Discussions***

***External Resources***

### 3.1 Proposed Component Design

The signer binary can be programmed to stop processing deposits if the amount of sBTC pegged in is over a configured cap. 

### 3.1.2 Considerations & Alternatives

An alternative design would be to publish the cap on-chain. Since, however, signers can always decide _not_ to process deposits, the additional implementation required seems to not pay dividends.

### 3.2 Areas of Ambiguity

---

**Closing Checklist**
- [ ] The design proposed in this issue is clearly documented in the description of this ticket.
- [ ] Everyone necessary has reviewed the resolution and agrees with the proposal.
- [ ] This ticket has or links all the information necessary to familiarize a contributor with the design decision, why it was made, and how it'll be included.

djordon:
Having an iterative cap on the amount pegged into sBTC is relatively easy to implement in the signers. It is easier than an account cap but harder than a per transaction cap (both proposed in https://github.com/stacks-network/sbtc/issues/791). I give all of these a complexity of 2 (or maybe 3).

PR #811, opened 2024-11-11
feat: suggestion on sbtc limits handling in signer

cylewitruk:
## Description

This is just a skeleton suggestion on how we could handle sBTC limits within the signer based on information in #562 and #791 to spark discussion.

This does not provide any actual implementation of limit fetching/handling.

### DIFF
--- a/signer/src/context/signer_state.rs
+++ b/signer/src/context/signer_state.rs
@@ -13,12 +13,67 @@ use crate::keys::PublicKey;
 #[derive(Debug, Default)]
 pub struct SignerState {
     current_signer_set: SignerSet,
+    current_limits: RwLock<SbtcLimits>,
 }
 
 impl SignerState {
     /// Get the current signer set.
     pub fn current_signer_set(&self) -> &SignerSet {
         &self.current_signer_set
+    }
+
+    /// Get the current sBTC limits.
+    pub fn get_current_limits(&self) -> SbtcLimits {
+        // We should never fail to acquire a lock from the RwLock so that it panics.
+        self.current_limits
+            .read()
+            .expect("BUG: Failed to acquire read lock")
+            .clone()
+    }
+
+    /// Update the current sBTC limits.
+    pub fn update_current_limits(&self, new_limits: SbtcLimits) {
+        // We should never fail to acquire a lock from the RwLock so that it panics.
+        let mut limits = self
+            .current_limits
+            .write()
+            .expect("BUG: Failed to acquire write lock");
+        *limits = new_limits;
+    }
+}
+
+/// Represents the current sBTC limits.
+#[derive(Debug, Clone, Default)]
+pub struct SbtcLimits {
+    /// Represents the total cap for all pegged-in BTC/sBTC.
+    total_cap: Option<f64>,
+    /// Represents the total pegged-in BTC/sBTC cap per Stacks account.
+    per_account_cap: Option<f64>,
+    /// Represents the maximum amount of BTC allowed to be pegged-in per transaction.
+    per_deposit_cap: Option<f64>,
+    /// Represents the maximum amount of sBTC allowed to be pegged-out per transaction.
+    per_withdrawal_cap: Option<f64>,
+}
+
+impl SbtcLimits {
+    /// Get the total cap for all pegged-in BTC/sBTC.
+    pub fn total_cap(&self) -> Option<f64> {
+        self.total_cap
+    }
+
+    /// Get the total pegged-in BTC/sBTC cap per Stacks account.
+    pub fn per_account_cap(&self) -> Option<f64> {
+        self.per_account_cap
+    }
+
+    /// Get the maximum amount of BTC allowed to be pegged-in per transaction.
+    pub fn per_deposit_cap(&self) -> Option<f64> {
+        self.per_deposit_cap
+    }
+
+    /// Get the maximum amount of sBTC allowed to be pegged-out per transaction.
+    pub fn per_withdrawal_cap(&self) -> Option<f64> {
+        self.per_withdrawal_cap
     }
 }
 
--- a/signer/src/main.rs
+++ b/signer/src/main.rs
@@ -15,6 +15,7 @@ use signer::block_observer;
 use signer::blocklist_client::BlocklistClient;
 use signer::config::Settings;
 use signer::context::Context;
+use signer::context::SbtcLimits;
 use signer::context::SignerContext;
 use signer::emily_client::EmilyClient;
 use signer::error::Error;
@@ -83,6 +84,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         context.state().current_signer_set().add_signer(signer);
     }
 
+    // TODO: Retrieve sBTC limits from another source (like Emily). This will
+    // set the initial limits for the signer all to `None` (unlimited).
+    context.state().update_current_limits(SbtcLimits::default());
+
     // Run the application components concurrently. We're `join!`ing them
     // here so that every component can shut itself down gracefully when
     // the shutdown signal is received.
@@ -99,6 +104,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         run_shutdown_signal_watcher(context.clone()),
         // The rest of our services which run concurrently, and must all be
         // running for the signer to be operational.
+        run_checked(run_update_limits, &context),
         run_checked(run_stacks_event_observer, &context),
         run_checked(run_libp2p_swarm, &context),
         run_checked(run_block_observer, &context),
@@ -309,4 +315,31 @@ async fn run_transaction_coordinator(ctx: impl Context) -> Result<(), Error> {
     };
 
     coord.run().await
+}
+
+/// Run the limit updater event-loop.
+async fn run_update_limits(ctx: impl Context) -> Result<(), Error> {
+    let mut interval = tokio::time::interval(Duration::from_secs(300)); // 5 Minutes
+    let mut term = ctx.get_termination_handle();
+
+    let poll = async {
+        loop {
+            interval.tick().await;
+
+            // TODO: Get limits from source of truth instead of hardcoding
+            //let limits = SbtcLimits::default();
+            //ctx.state().update_current_limits(limits);
+        }
+    };
+
+    tokio::select! {
+        _ = term.wait_for_shutdown() => {
+            tracing::info!("shutting down the limit updater");
+        },
+        _ = poll => {
+            tracing::info!("limit updater has stopped");
+        }
+    }
+
+    Ok(())
 }
--- a/signer/src/transaction_signer.rs
+++ b/signer/src/transaction_signer.rs
@@ -742,6 +742,12 @@ where
             .await?
             .unwrap_or(false);
 
+        // TODO: Validate that the deposit is within the current sBTC limits:
+        // - Total cap for all pegged-in BTC/sBTC.
+        // - Total cap per-Stacks-principal for pegged-in BTC/sBTC.
+        // - Total cap per-deposit.
+        // let limits = self.context.state().get_current_limits();
+
         let is_accepted = can_sign && self.can_accept_deposit_request(&request).await?;
 
         let msg = message::SignerDepositDecision {
@@ -778,6 +784,10 @@ where
         withdrawal_request: model::WithdrawalRequest,
         bitcoin_chain_tip: &model::BitcoinBlockHash,
     ) -> Result<(), Error> {
+        // TODO: Validate that the withdrawal is within the current sBTC limits:
+        // - Total cap per-withdrawal.
+        // let limits = self.context.state().get_current_limits();
+
         // TODO: Do we want to do this on the sender address or the
         // recipient address?
         let is_accepted = self


docker/docker-compose.yml
```diff
@@ -364,7 +367,7 @@ services:
     profiles:
       - default
     extra_hosts:
-      - "host.docker.internal:host-gateway"
+      - "host.docker.internal:10.0.2.2"
```
cylewitruk, line 370: I'll revert this before an eventual merge. Leaving this comment here so I can't forget :)

docker/sbtc/signer/signer-config.toml
```diff
@@ -26,7 +26,7 @@
 # Environment: SIGNER_EMILY__ENDPOINTS
 # Environment Example: '{"endpoint":"https://api.emilyexample.com","api_key":"1234567890abcdef"},..'
 endpoints = [
-    { endpoint = "http://emily-server:3031" },
+    { endpoint = "http://localhost:3031" },
```
cylewitruk, line 29: I'll revert this before an eventual merge, comment so I don't forget.

signer/src/network/libp2p/network.rs
```diff
@@ -104,7 +104,7 @@ impl MessageTransfer for P2PNetwork {
     }
 
     fn receiver_stream(&self) -> BroadcastStream<SignerSignal> {
-        let (sender, receiver) = tokio::sync::broadcast::channel(10);
+        let (sender, receiver) = tokio::sync::broadcast::channel(1_000);
```
cylewitruk, line 107: Unrelated, will revert this before merge

signer/src/request_decider.rs
```diff
@@ -126,17 +126,35 @@ where
             .await?;
 
         for deposit_request in deposit_requests {
-            self.handle_pending_deposit_request(deposit_request, &chain_tip)
-                .await?;
+            let outpoint = deposit_request.outpoint();
+            let _ = self
```
cylewitruk, line 130: Changed this logic so that we don't bail at the first sign of trouble (i.e. we keep processing the remaining requests).

signer/src/request_decider.rs
```diff
@@ -236,8 +254,8 @@ where
         // TODO: Do we want to do this on the sender address or the
         // recipient address?
         let is_accepted = self
-            .can_accept(&withdrawal_request.sender_address.to_string())
-            .await;
+            .can_accept_withdrawal_request(&withdrawal_request)
+            .await?;
```
Jiloc, line 258: nit: I like that the` can_accept_*` functions now consistently return a Result. However, it seems this particular call can't fail. The `?` threw me off, making me wonder where it might fail. Maybe consider using expect to make it clear this is infallible? On the cons side, If the function ever changes to return an error, we'd need to update the expect. Just a thought, happy to leave it as-is if you prefer!

signer/src/request_decider.rs
Jiloc: sanity check: are we potentially breaking any assumptions in the code by only processing and propagating a subset of the deposit/withdrawal requests in a block? I don't think so, since without this change, if we fail mid-iteration, we'd still end up in a similar state. Curious if you have thoughts or know of any cases where this might cause issues.

Jiloc (commented):
Just a few sanity checks. My main question is why this is handled in a separate sidecar that updates every 5 minutes, rather than during the creation and processing of new transactions. That said, the functionality remains unchanged, so I’m completely okay with it. The only potential edge case I see is if we lowered the cap while signers were validating new transactions, but that’s highly unlikely, and we’re not planning anything like that. Overall, I’m fine with the current implementation!

docker/docker-compose.yml
```diff
@@ -134,6 +134,9 @@ services:
       - DYNAMODB_ENDPOINT=http://emily-dynamodb:8000
       - HOST=0.0.0.0
       - PORT=3031
+      - DEFAULT_PEG_CAP=100000000000
+      - DEFAULT_PER_DEPOSIT_CAP=100000000
+      - DEFAULT_PER_WITHDRAWAL_CAP=100000000
```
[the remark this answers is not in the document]
cylewitruk: Nah these are just some default values for dev, so we can change them at will.

signer/src/emily_client.rs
```diff
@@ -326,6 +336,27 @@ impl EmilyInteract for EmilyClient {
...
             .map_err(Error::EmilyApi)
     }
+
+    async fn get_limits(&self) -> Result<SbtcLimits, Error> {
+        let limits = limits_api::get_limits(&self.config)
+            .await
+            .map_err(EmilyClientError::GetLimits)
+            .map_err(Error::EmilyApi)?;
+
+        let total_cap = limits.peg_cap.and_then(|cap| cap.map(Amount::from_sat));
```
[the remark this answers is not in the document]
cylewitruk: I mentioned that to Ashton and he said we're free to choose. It's not a float so the only remaining option is satoshis really (which everything else we use is defined in as well)

signer/src/request_decider.rs
cylewitruk: Yeah the infallible scenario is just because this is anemic compared to its deposit equivalent, so I fully expect this to be built out as well when we get there.

cylewitruk:
> Just a few sanity checks. My main question is why this is handled in a separate sidecar that updates every 5 minutes, rather than during the creation and processing of new transactions. That said, the functionality remains unchanged, so I’m completely okay with it. The only potential edge case I see is if we lowered the cap while signers were validating new transactions, but that’s highly unlikely, and we’re not planning anything like that. Overall, I’m fine with the current implementation!

So my thinking was largely that this value should almost never change and to avoid extra calls to Emily that maybe this would be once an hour or something instead of every block (or 5m as it's coded right now). I also wanted to avoid piling additional things on-top of the block-observer checks in general. I don't have a strong opinion either way tbh.

signer/src/request_decider.rs
Jiloc: Dunno why, but it looks like a good idea not to halt everything for a single request error :) I can't think at any broken assuptions either, so we are good!

Jiloc:
That makes sense. The more I think about it, the only way to make this check synchronous with the deposit request would be to have it act an emergency stop mechanism to halt deposits immediately if something unusual happens. However, this isn’t meant to serve that purpose, so I agree that we’re fine as it is. Keeping the block-observer "lean" is definitely a good goal.

cylewitruk:
Note that this PR is still missing tests, hence the draft status. I'm going to work on those next.

signer/src/bitcoin/utxo.rs
```diff
@@ -162,18 +165,39 @@ impl SbtcRequests {
...
+        // 1. The user's max fee must be >= our minimum required fee for deposits
+        //     (based on fixed deposit tx size)
+        // 2. The deposit amount must be less than the per-deposit limit
+        // 3. The total amount being minted must stay under the maximum allowed mintable amount
         let minimum_deposit_fee = self.compute_minimum_fee(SOLO_DEPOSIT_TX_VSIZE);
-        let deposits = self
-            .deposits
-            .iter()
-            .filter(|req| req.max_fee.min(req.amount) >= minimum_deposit_fee)
-            .map(RequestRef::Deposit);
```
cylewitruk, line 174: I might embed this logic into the `SbtcLimits` impl instead since we likely want the same kind of logic everywhere.

docker/docker-compose.yml
Jiloc: reverted in https://github.com/stacks-network/sbtc/pull/811/commits/f14503f8fc405b5995400433c7b477df82198cf3

docker/sbtc/signer/signer-config.toml
Jiloc: reverted in https://github.com/stacks-network/sbtc/pull/811/commits/f14503f8fc405b5995400433c7b477df82198cf3

signer/src/bitcoin/validation.rs
```diff
@@ -91,10 +100,117 @@ pub fn is_unique(package: &[TxRequestIds]) -> bool {
...
+        let cache = self.fetch_all_reports(ctx, btc_ctx).await?;
+
+        self.validate_max_mintable(ctx, &cache).await?;
+
+        Ok(cache)
+    }
+
+    async fn fetch_all_reports<C>(
+        &self,
+        ctx: &C,
```
cylewitruk, line 123: suggestion: looks like you could just take a `DbRead` here instead of a full `Context`

signer/src/context/signer_state.rs
```diff
@@ -13,13 +14,98 @@ use crate::keys::PublicKey;
...
+        self.per_withdrawal_cap
+    }
+
+    /// Get the maximum amount of sBTC that can currently be minted.
+    pub fn max_mintable_cap(&self) -> Option<Amount> {
+        self.max_mintable_cap
+    }
+
+    /// Set the total cap for sBTC that can currently be minted.
+    pub fn set_max_mintable_cap(&mut self, max_mintable_cap: Option<Amount>) {
```
cylewitruk, line 106: When I had designed this I intended for this struct to be immutable, because when you pull it out you get a clone and any changes won't be reflected here. Just noting this, and if we keep this method here we should probably make it clear on the `get_current_limits` method that any changes made to the struct won't be reflected in global state.

signer/src/block_observer.rs
```diff
@@ -155,6 +156,11 @@ where
                         tracing::warn!(%error, "could not load latest deposit requests from Emily");
                     }
 
+                    if let Err(error) = self.update_sbtc_limits().await {
```
cylewitruk, line 159: I think maybe this should be moved above the `load_latest_deposit_requests` block above.

signer/src/block_observer.rs
```diff
@@ -472,6 +478,34 @@ impl<C: Context, B> BlockObserver<C, B> {
...
+            }
+        };
+        limits.set_max_mintable_cap(Some(max_mintable));
+
+        let signer_state = self.context.state();
+        if limits == signer_state.get_current_limits() {
+            tracing::trace!(%limits, "sBTC limits have not changed");
+        } else {
+            tracing::debug!(%limits, "updated sBTC limits from Emily");
+            signer_state.update_current_limits(limits);
```
cylewitruk, line 505: If you constructed a new `SbtcLimits` struct here then you wouldn't need the `set_max_mintable_cap()` method (see earlier comment)

signer/src/bitcoin/utxo.rs
```diff
@@ -162,18 +165,39 @@ impl SbtcRequests {
...
 
+        // If the caps are not set, we default to the maximum amount of sats.
```
cylewitruk, line 175: I might embed this logic into the SbtcLimits impl instead since we likely want the same kind of logic everywhere.

signer/src/context/signer_state.rs
```diff
@@ -13,13 +14,98 @@ use crate::keys::PublicKey;
...
+            .current_limits
+            .write()
+            .expect("BUG: Failed to acquire write lock");
+        *limits = new_limits;
+    }
+}
+
+/// Represents the current sBTC limits.
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct SbtcLimits {
```
matteojug, line 48: qq: is there a reason not to have everything pub in here and avoid the getters?

signer/tests/integration/emily.rs
```diff
@@ -627,3 +629,36 @@ async fn get_deposit_request_works() {
...
+#[tokio::test]
+async fn get_limits_works() {
+    let url = Url::parse("http://localhost:3031").unwrap();
+    let emily_client = EmilyClient::try_from(&url).unwrap();
+    let mut config = emily_client::apis::configuration::Configuration::new();
+    // Url::parse defaults `path` to `/` even if the parsed url was without the trailing `/`
+    // causing the api calls to have two leading slashes in the path (getting a 404)
+    config.base_path = url.to_string().trim_end_matches("/").to_string();
+    emily_client::apis::limits_api::set_limits(
+        &config,
```
matteojug, line 643: nit?
```suggestion
    emily_client::apis::limits_api::set_limits(
        &emily_client.config(),
```

matteojug (approved):
mostly nits, LGTM

signer/src/bitcoin/validation.rs
Jiloc: done in https://github.com/stacks-network/sbtc/pull/811/commits/086c1a9a380d8416c63ea1edca2d3307453a0bf1

signer/src/context/signer_state.rs
Jiloc: I removed the method and did what you proposed below

signer/src/block_observer.rs
Jiloc: done in https://github.com/stacks-network/sbtc/pull/811/commits/086c1a9a380d8416c63ea1edca2d3307453a0bf1

signer/src/bitcoin/utxo.rs
Jiloc: i gave a default of `Amount::MAX_MONEY` to all getters https://github.com/stacks-network/sbtc/pull/811/commits/086c1a9a380d8416c63ea1edca2d3307453a0bf1

signer/src/block_observer.rs
```diff
@@ -472,6 +478,34 @@ impl<C: Context, B> BlockObserver<C, B> {
...
+                    .context
+                    .get_stacks_client()
+                    .get_sbtc_total_supply(&self.context.config().signer.deployer)
+                    .await?;
+                // The maximum amount of sBTC that can be minted is the total cap
+                // minus the current supply.
+                total_cap.checked_sub(sbtc_supply).unwrap_or(Amount::ZERO)
+            }
+        };
+        limits.set_max_mintable_cap(Some(max_mintable));
```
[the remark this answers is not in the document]
Jiloc: I removed the setter, and all getters directly return a MAX_MONEY in case their value is None

signer/src/block_observer.rs
Jiloc: Fixed in https://github.com/stacks-network/sbtc/pull/811/commits/086c1a9a380d8416c63ea1edca2d3307453a0bf1

Issue #785, opened 2024-11-06
[Chore]: Production docker compose


AshtonStephens:
# Chore - Production docker compose 

## 1. Description

Create a docker compose that spins up all the necessary containers for a signer to run sBTC in production.

## 2. Why It's Necessary

People running signers need to be able to spin up the infrastructure quickly and smoothly. This GitHub issue is to make the docker compose that sets all the entities up.

## 3. Execution Checklist

- [ ] Create a docker compose that can run all the necessary production infrastructure
- [ ] Have that docker compose in a PR and merged

AshtonStephens:
@aldur what do we envision to be the immediate purpose of this docker compose? Is this to give people an idea of what's necessary or be exactly what they anticipate using in prod?

Additionally, I'd like to propose that we don't include a stacks node in this, and we expect them to have that already running so we separate this compose from any changes made to the stacks core images. WDYT?

aldur:
> @aldur what do we envision to be the immediate purpose of this docker compose? Is this to give people an idea of what's necessary or be exactly what they anticipate using in prod?

Let's assume that signers will use this either verbatim or as a base (e.g., if deploying through K8).

> Additionally, I'd like to propose that we don't include a stacks node in this, and we expect them to have that already running so we separate this compose from any changes made to the stacks core images. WDYT?

I agree. Would it be feasible for signers that are not already running their node to rely on a node provider?

PR #813, opened 2024-11-11
feat: Add production like docker compose for docs

AshtonStephens:
## Description

Closes: #785

> **Note: This is _NOT_ meant to be run! It is solely for documentation, but is close to accurate!**

This PR creates a docker compose and configuration files meant to be used solely in documentation. The primary goal is
to give those who will be running signers the right idea of what will be required. This docker compose calls upon docker
images that aren't published yet.

I fully expect us to need to make decent changes to these, but the communication aspect won't be greatly changed by
the changes we'll need to make. 

I've opted to leave in some of the `TODO(123)` comments in the signer configuration because I think leaving them in
makes it clearer what work needs to be done and what they can expect will have changes before they deploy.

## Changes

Adds three files:

```
prodlike
├── bitcoin
│   ├── Dockerfile.bitcoin
│   └── bitcoin.conf
├── config
│   ├── blocklist-client-config.toml
│   └── signer-config.toml
└── docker-compose.testnet.yml
```

## Testing Information

Not tested.

## Checklist:


docker/prodlike/bitcoin/Dockerfile.bitcoin
```diff
@@ -0,0 +1,8 @@
+FROM debian:bookworm
+
+COPY --from=dobtc/bitcoin:25.1 /opt/bitcoin-*/bin /usr/local/bin
+
+RUN apt-get update && apt-get install -y curl jq zstd
+RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
cylewitruk, line 6: Just a little note that this doesn't reduce the final image size as it's still included in a previous layer, unless the image is built with `--squash`.

docker/prodlike/config/signer-config.toml
```diff
@@ -0,0 +1,261 @@
...
+# You may specify multiple Emily API servers if you have them. They will be
+# tried round-robin until one succeeds.
+#
+# Format: ["http(s)://<host>:<port>", ..]
+# Default: <none>
+# Required: true
+# Environment: SIGNER_EMILY__ENDPOINTS
+endpoints = [
+    "https://emily-sbtc.com"
+]
```
AshtonStephens, line 44: This needs to be changed to consume the configuration changes made in this ticket:
- #861

docker/prodlike/config/signer-config.toml
```diff
@@ -0,0 +1,261 @@
...
+# Environment Example: tcp://10.0.0.1:28332,tcp://10.0.0.2:28332
+block_hash_stream_endpoints = [
+    "tcp://localhost:28332"
+]
+
+# !! ==============================================================================
+# !! Block Notifier Configuration
+# !! ==============================================================================
+# Electrum server connection confirmation.
+[block_notifier]
```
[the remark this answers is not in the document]
AshtonStephens: wdym? What part should be deleted? I see this in the default config:
- https://github.com/stacks-network/sbtc/blob/main/signer/src/config/default.toml#L67-L82 - [permalink](https://github.com/stacks-network/sbtc/blob/b2e7f99900c930eac0d70f47ed585b9ace2342d0/signer/src/config/default.toml#L67-L82)
```toml
# !! ==============================================================================
# !! Block Notifier Configuration
# !! ==============================================================================
# Electrum server connection confirmation.
[block_notifier]
# The URI of the Electrum server to connect to.
#
# Format: "<protocol>://<host>:<port>"
# Default: <none>
# Required: true
# Environment: SIGNER_BLOCK_NOTIFIER__SERVER
server = "tcp://localhost:60401"
retry_interval = 10
max_retry_attempts = 5
ping_interval = 60
subscribe_interval = 10
```

docker/prodlike/docker-compose.prod.yml
```diff
@@ -0,0 +1,59 @@
...
+      - 5432:5432
+
+  sbtc-signer:
+    image: blockstack/sbtc:signer-latest
+    entrypoint: "/bin/bash -c '/usr/local/bin/signer -c /signer-config.toml --migrate-db'"
+    depends_on:
+      - postgres
+    environment:
+      RUST_LOG: info
+      SIGNER_SIGNER__P2P__LISTEN_ON: tcp://0.0.0.0:4122
```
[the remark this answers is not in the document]
AshtonStephens: Done.

aldur:
Based on the amount of changed files this might beed to be re-based.

docker/prodlike/config/signer-config.toml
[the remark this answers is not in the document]
djordon: I wonder if this has been repurposed.

docker/prodlike/bitcoin/bitcoin.conf
```diff
@@ -0,0 +1,41 @@
+regtest=1 #chain=regtest
+
+[regtest]
+printtoconsole=1
+disablewallet=0
```
djordon, line 5: I think this should be 1 even in regest, I think we want the wallet disabled.

djordon: Ahhh, this is the same thing we use in devnet. Okay, it's probably fine to use either.

---
Thinking out loud:

Having the wallet enabled is probably necessary in devnet because of the stacks-core minor or for bridge demos. But shouldn't be not necessarily on testnet. Also, our integration tests run bitcoin core with the wallet disabled, and I think it's important during our integration testing. This may explain why I had some trouble running integration tests against devnet.

docker/prodlike/config/signer-config.toml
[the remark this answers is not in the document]
djordon: It wasn't repurposed, we need to delete this in a separate PR.

docker/prodlike/config/signer-config.toml
```diff
@@ -0,0 +1,261 @@
...
+#
+# Format: ["<protocol>:<ip>:<port>", "<protocol>:<ip>:<port>", ...]
+# Required: false
+# Environment: SIGNER_SIGNER__P2P__SEEDS
+# Environment Example: tcp://seed-1:4122,tcp://seed-2:4122
+# TODO(429): Add well-known seed nodes
+# TODO(715): Add well-known seed nodes
+seeds = [
+    "<protocol>:<ip>:<port>",
+    "provided:provided-ip:provided-port"
```
djordon, line 213: I'd just keep the seeds empty.

docker/prodlike/config/signer-config.toml
```diff
@@ -0,0 +1,261 @@
...
+# db_endpoint = "postgresql://postgres:postgres@localhost:5432/signer"
+
+# A complete list of (compressed) public keys for known bootstrap signer
+# peers who are approved to be in the sBTC signer set.
+#
+# Required: true Environment: SIGNER_SIGNER__BOOTSTRAP_SIGNING_SET
+# TODO(715): Change after initial signing set has been determined.
+bootstrap_signing_set = [
+    "03providedsigningsetpublickey023a1d53bc96ad670bfe03adf8a06c52e6380",
+    "02providedsigningsetpublickey023b28143130a18099ecf094d36fef0f6135c",
```
djordon, line 160: I'd leave this empty

docker/prodlike/config/signer-config.toml
```diff
@@ -0,0 +1,261 @@
...
+#
+# This may be either in 32- or 33-byte format. If you generated the key using
+# `stacks-cli` or other ecosystem tools, it is likely that the key is in 33-byte
+# format which includes a stacks-proprietary suffix byte. The sBTC signer doesn't
+# make use of this byte and it will be trimmed automatically if provided.
+#
+# Format: "<hex-encoded-private-key>" (64 or 66 hex-characters)
+# Required: true
+# Environment: SIGNER_SIGNER__PRIVATE_KEY
+private_key = "your-private-key"
```
djordon, line 124: I'd change this to `"0000000000000000000000000000000000000000000000000000000000000001"
`

docker/prodlike/config/signer-config.toml
```diff
@@ -0,0 +1,261 @@
...
+subscribe_interval = 10
+
+# !! ==============================================================================
+# !! Stacks Node Configuration
+# !! ==============================================================================
+[stacks]
+# The RPC URL(s) of the Stacks node(s) to connect to. At least one must be
+# provided. If multiple nodes are provided they will be tried in order when
+# making requests.
+endpoints = ["stacks-node-rpc"]
```
djordon, line 103: nit: I'd change to be empty.

docker/prodlike/docker-compose.testnet.yml
```diff
@@ -0,0 +1,58 @@
...
+      - 5432:5432
+
+  sbtc-signer:
+    image: blockstack/sbtc:signer-0.0.1-dev-4
+    entrypoint: "/bin/bash -c '/usr/local/bin/signer -c /signer-config.toml --migrate-db'"
+    depends_on:
+      - postgres
+    environment:
+      RUST_LOG: info
+      SIGNER_SIGNER__DB_ENDPOINT: postgresql://postgres:postgres@postgres-1:5432/signer
```
djordon, line 28:
```suggestion
      SIGNER_SIGNER__DB_ENDPOINT: postgresql://postgres:postgres@postgres:5432/signer
```

docker/prodlike/docker-compose.testnet.yml
```diff
@@ -0,0 +1,58 @@
...
+    ports:
+      - 5432:5432
+
+  sbtc-signer:
+    image: blockstack/sbtc:signer-0.0.1-dev-4
+    entrypoint: "/bin/bash -c '/usr/local/bin/signer -c /signer-config.toml --migrate-db'"
+    depends_on:
+      - postgres
+    environment:
+      RUST_LOG: info
```
djordon, line 27:
```suggestion
      RUST_LOG: info,signer=debug
```

docker/prodlike/docker-compose.testnet.yml
```diff
@@ -0,0 +1,58 @@
...
+    volumes:
+      - ./bitcoin/bitcoin.conf:/root/.bitcoin/bitcoin.conf
+    entrypoint:
+      - /bin/bash
+      - -c
+      - |
+        set -e
+        bitcoind
+    profiles:
+      - default
```
djordon, line 57: Do we need to add this to the above ones? Maybe we should remove this profiles stanza altogether.

docker/prodlike/config/signer-config.toml
```diff
@@ -0,0 +1,261 @@
...
+# Required: true
+# Possible values: mainnet, testnet, regtest
+# Environment: SIGNER_SIGNER__NETWORK
+network = "mainnet"
+
+# The address that deployed the sbtc smart contracts.
+#
+# Required: true
+# TODO(715): Change after SCs have been deployed.
+deployer = "SNYourDeployerAddress"
```
djordon, line 141: I'd change this to `ST000000000000000000002AMW42H`, which is the testnet burn address.

djordon (commented):
Largely looks good. I had some docker compose fixes and questions. 

I also made some suggestions for the signer config, so that people can run the signer binary with that config but where nothing happens because things aren't configured correctly. But then I thought that maybe the point was that everything fails when it isn't configured correctly. If so feel free to disregard.

docker/prodlike/config/signer-config.toml
AshtonStephens: Ah yes, agreed. It made more sense for docs imo

docker/prodlike/docker-compose.testnet.yml
AshtonStephens: Done.

docker/prodlike/config/signer-config.toml
AshtonStephens: That's an interesting address, I wonder if the ending is meant to mean something...

docker/prodlike/config/signer-config.toml
AshtonStephens: Done.

docker/prodlike/config/signer-config.toml
AshtonStephens: Done.

docker/prodlike/config/signer-config.toml
AshtonStephens: Done.

docker/prodlike/config/signer-config.toml
AshtonStephens: Done.

docker/prodlike/docker-compose.testnet.yml
AshtonStephens: Done.

docker/prodlike/docker-compose.testnet.yml
AshtonStephens: Removed.

docker/prodlike/config/signer-config.toml
AshtonStephens: Done.

djordon: It means 2 AM on Wednesday we'll be 42 hours away from deploying.

djordon (approved):
Looks good ✅

djordon:
> Btw it seems that some commit after [7a4b569](https://github.com/stacks-network/sbtc/commit/7a4b5690c8321ae858655a72d6e1ed28b25dbe0b) broke something, let me double check

Wait so using the latest flag worked by this specific flag does not?

Issue #814, opened 2024-11-11
[Feature]: Add API key security to private endpoints on Emily openapi spec


AshtonStephens:
# Feature - Add API key security to private endpoints on Emily openapi spec

## 1. Description

Some of the endpoints to the Emily API are private - this GH issue is to add API keys to the relevant autogenerated openapi calls.

Note that while this was not created during planning, it falls under the umbrella of #692

### 1.1 Context & Purpose
Same as above.

## 2. Technical Details:
The openapi template will need to require api key level access in order to make the API calls. We need to have the openapi template specify API key level call permissions such that AWS knows how to consume the API keys.

### 2.1 Acceptance Criteria:
   - [ ] The openapi template includes Api key security requirements for some of the API calls.
   - [ ] There is a PR created and merged for this feature.

Note that this does NOT include hooking the API Keys into the AWS cdk definition.

## 3. Related Issues and Pull Requests (optional):

- #692

PR #816, opened 2024-11-11
814 feature add api key security to private endpoints on emily openapi spec

AshtonStephens:
## Description

Closes: #814 

Adds api key level security to private Emily API endpoints.

## Changes

- Adds openapi template builder modifiers so the API keys can be added
- Adds the security requirement to the API endpoints 

## Testing Information

- Tested with integration tests and building.

## Checklist:

### DIFF
--- a/emily/handler/src/api/handlers/health.rs
+++ b/emily/handler/src/api/handlers/health.rs
@@ -15,7 +15,8 @@ use crate::common::error::Error;
         (status = 404, description = "Address not found", body = ErrorResponse),
         (status = 405, description = "Method not allowed", body = ErrorResponse),
         (status = 500, description = "Internal server error", body = ErrorResponse)
-    )
+    ),
+    security(("ApiGatewayKey" = []))
 )]
 pub async fn get_health() -> impl warp::reply::Reply {
     Error::NotImplemented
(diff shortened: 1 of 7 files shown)


emily/handler/src/api/handlers/health.rs
```diff
@@ -15,7 +15,8 @@ use crate::common::error::Error;
         (status = 404, description = "Address not found", body = ErrorResponse),
         (status = 405, description = "Method not allowed", body = ErrorResponse),
         (status = 500, description = "Internal server error", body = ErrorResponse)
-    )
+    ),
+    security(("ApiGatewayKey" = []))
```
matteojug, line 19: Do we want this to be private?

matteojug (commented):
This look good, but it seems to me it's only changing the openapi spec. Is there some aws-magic going on here that actually enforces the api key check?

emily/handler/src/api/handlers/health.rs
AshtonStephens: I thought so?

matteojug (approved):

emily/handler/src/api/handlers/health.rs
matteojug: It seems the kind of endpoint that may be useful for external clients, @aldur WDYT?

AshtonStephens: What do you think about us keeping it private for now so we don't need to worry about exposing too much info from this endpoint and then making it public later if we decide we want to?

Issue #760, opened 2024-11-02
[Feature]: Deconstruct bitcoin transactions when populating sweep tables


djordon:
# Feature - Deconstruct bitcoin transactions when populating sweep tables

## 1. Description

This is a part of https://github.com/stacks-network/sbtc/issues/758, and shares the same description, context and purpose.

## 2. Technical Details:

After we get a transaction from bitcoin core, each signer can identify each of the inputs with a deposit request or a signer UTXO by the outpoint. Similarly, the first two outputs are the signers new UTXO and an `OP_RETURN` output. The withdrawal outputs can be tricky since there is some ambiguity; a `scriptPubKey` and an amount does not uniquely identify a withdrawal request. For withdrawals, we can use the merkle root stored in the `OP_RETURN` output (https://github.com/stacks-network/sbtc/issues/483) to help identify the withdrawals in the transaction.

### 2.1 Acceptance Criteria:
   - [ ] Write a function that transforms a bitcoin transaction into the sweep transaction object introduced in https://github.com/stacks-network/sbtc/pull/732.

## 3. Related Issues and Pull Requests (optional):

* relates to https://github.com/stacks-network/sbtc/issues/585,
* relates to https://github.com/stacks-network/sbtc/issues/709,
* a subtask of https://github.com/stacks-network/sbtc/issues/758.

PR #822, opened 2024-11-11
feat: deconstruct and store bitcoin transactions into inputs and outputs

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/760.

## Changes

* Remove the signer_outputs field from the `SweepTransaction` and `SweepTransactionInfo` types.
* Add a trait for deconstructing a bitcoin transaction into inputs and outputs that are properly labeled for usage by the signers.
* Add new tables containing these deconstructed components and add two functions for writing to these tables.
* Deconstruct relevant transactions in the block observer and write the outputs into the database.

## Testing Information

This PR adds an integration test that checks the logic in the block observer is correct for donations and sweep transactions.

## Checklist:


signer/src/block_observer.rs
```diff
@@ -392,52 +393,36 @@ where
                 continue;
             }
 
+            // This function is called after we have received a
```
cylewitruk, line 396: Looks like we've changed the validation here a bit, like the previous version was grabbing the previous tx from bitcoin core and validating etc. I wasn't involved in the original issue so I'm not really sure what the exact expectations are here, just want to raise a :eyes: in-case this should be double-checked.

signer/tests/integration/block_observer.rs
```diff
@@ -331,3 +354,350 @@ async fn link_blocks() {
 
     testing::storage::drop_db(db).await;
 }
+
```
cylewitruk, line 357: Hm, I think I would have preferred that these returned a `Vec<TxOutput>` and `Vec<TxPrevout>` and used `.is_empty()` instead of `.is_none()` etc. below, to encourage re-use, since the only filter is on the type (another future test might expect two of them, for example).

signer/src/bitcoin/utxo.rs
```diff
@@ -1068,6 +1080,142 @@ impl BitcoinTxInfo {
...
+        // This transaction might not be related to the signers at all. If
+        // not then we can exit early.
+        let mut outputs = self.outputs().iter();
+        if !outputs.any(|tx_out| signer_script_pubkeys.contains(&tx_out.script_pubkey)) {
+            return Vec::new();
+        }
+
+        // If the signers did not create this transaction, but the signers
+        // control at least one output then the outputs that the signers
+        // control are donations. So we scan the outputs and exit early.
```
matteojug, line 1146: maybe nit: for completeness we could mention that those cannot be deposits since deposits outputs are not "directly" controlled by signers aggregate key

signer/src/block_observer.rs
matteojug: LGTM!

signer/src/bitcoin/utxo.rs
```diff
@@ -1068,6 +1080,142 @@ impl BitcoinTxInfo {
...
+            .enumerate()
+            .filter_map(|(index, _)| match index {
+                0 => self.vout_to_output(index, TxOutputType::SignersOutput),
+                1 => self.vout_to_output(index, TxOutputType::SignersOpReturn),
+                _ => self.vout_to_output(index, TxOutputType::Withdrawal),
+            })
+            .collect()
+    }
+
+    /// Take an output index and the known output type and return
```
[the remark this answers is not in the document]
djordon: https://github.com/stacks-network/sbtc/pull/822/commits/54ca3d35eb95111bab5e2dc37eb048201b4ecf1b

signer/src/bitcoin/utxo.rs
```diff
@@ -931,11 +945,18 @@ impl<'a> UnsignedTransaction<'a> {
 /// A trait where we return all inputs and outputs for a bitcoin
 /// transaction.
 pub trait InputsOutputs {
```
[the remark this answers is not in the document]
djordon: https://github.com/stacks-network/sbtc/pull/822/commits/54ca3d35eb95111bab5e2dc37eb048201b4ecf1b

signer/src/storage/in_memory.rs
```diff
@@ -1088,7 +1088,11 @@ impl super::DbWrite for SharedStore {
         Ok(())
     }
 
-    async fn write_signer_txo(&self, _signer_output: &model::SignerOutput) -> Result<(), Error> {
+    async fn write_tx_output(&self, _output: &model::TxOutput) -> Result<(), Error> {
+        Ok(())
```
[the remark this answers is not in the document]
cylewitruk: Haha okay 😅

cylewitruk (approved):

PR #649, opened 2024-10-10
feat: adds a little cli we can use for testing/demo to execute the needed operations

cylewitruk:
## Description

This just adds a little CLI we can build on for tasks needed for the demo (in the immediate future). Maybe we find it useful and want to keep some sort of CLI, in which case we should rename it :)

## Changes

- Creates a `bin` directory and moves `main.rs` there but renamed as `sbtc_signer.rs`.
- Adds `demo_cli.rs` as a bin
- Added both bins to the `Cargo.toml`

## Checklist:

- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #825, opened 2024-11-11
chore: demo-cli

matteojug:
## Description

Add `demo-cli` tool used during demo.

## Changes

Includes `demo-cli` binary originally from https://github.com/stacks-network/sbtc/pull/649 (with some small changes). We can use `./signers.sh demo` to issue a donation and create a deposit request, and `./signers.sh info` to get signers related info to be used to configure the bridge.

## Testing Information

Tested on devenv.

## Checklist:

### DIFF
--- a/signer/Cargo.toml
+++ b/signer/Cargo.toml
@@ -77,3 +77,7 @@ test-log = { version = "0.2.16", default-features = false, features = ["trace"] 
 
 [build-dependencies]
 tonic-build.workspace = true
+
+[[bin]]
+name = "demo-cli"
+path = "src/bin/demo_cli.rs"
--- a/signer/src/bin/demo_cli.rs
+++ b/signer/src/bin/demo_cli.rs
@@ -1,0 +1,332 @@
+use std::str::FromStr;
+
+use bitcoin::hex::DisplayHex;
+use bitcoin::{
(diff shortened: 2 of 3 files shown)


djordon:
> What would you guys think of moving `main.rs` into `bin/signer.rs` to be consistent with where binaries live? Before I had called it `sbtc_signer.rs` and named it `sbtc-signer` (because otherwise the binary name isn't very clear in `ps` commands when e.g. you've also got a stacks signer running). But that needed some other changes in the `Makefile` etc.
> 
> Thoughts?

Well, I think we should place it where most people would expect it to be. If I were new I'd look for a `src/main.rs` and then if I don't see it I'd look in the `src/bin` folder (mainly because I think of the `src/bin` folder as the place where other binaries go, not where the main binary "lives"). But I don't feel strongly about it, maybe my reading of the convention is incorrect, so I'm curious about what others think.

cylewitruk:
Just read through https://doc.rust-lang.org/cargo/guide/project-layout.html and you're right, the _main_ binary should be located at `/src/main.rs` :+1:

cylewitruk (approved):
LGTM, not going to bother nitpicking anything since this is a dev tool and we can easily iterate on it.

PR #827, opened 2024-11-12
[chore] Introduce cargo vet

MCJOHN974:
## Description

Add cargo vet to our CI

## Changes
Just added cargo vet to CI

## Motivation
Cargo vet is a tool to check if third-party crates we use passed "trusted" audit. Also, it checks if version of given crate we use passed the audit. https://mozilla.github.io/cargo-vet/index.html

Benefits it can give to us are increased security


Makefile
```diff
@@ -58,6 +58,10 @@ lint: $(AUTOGENERATED_BLOCKLIST_CLIENT_CLIENT) $(AUTOGENERATED_EMILY_CLIENT)
 format: $(AUTOGENERATED_BLOCKLIST_CLIENT_CLIENT) $(AUTOGENERATED_EMILY_CLIENT)
 	cargo fmt
 
+vet: $(AUTOGENERATED_BLOCKLIST_CLIENT_CLIENT) $(AUTOGENERATED_EMILY_CLIENT)
+	cargo install cargo-vet
```
matteojug, line 62: nit: should we move the install under the `install` target above?

supply-chain/config.toml
djordon, line 1: How did you create this file. We should document that somewhere, probably at the top of the file itself and/or the repo README.

.github/workflows/unit-tests.yaml
```diff
@@ -40,6 +40,9 @@ jobs:
       - name: Run lints
         run: make lint
 
+      - name: Run vetting
+        run: make vet
+
```
djordon, line 45: Shouldn't this be in another workflow that runs in parallel to this one.

djordon (approved):
Looks good ✅

.github/workflows/cargo-vet.yaml
```diff
@@ -8,7 +8,7 @@ concurrency:
 
 jobs:
   cargo-vet:
-    name: Run Unit Tests
+    name: Run cargo vet
```
ilyungjoey, line 11: .github/workflows/cargo-vet.yaml

Issue #801, opened 2024-11-08
[Feature]: Extend DB schemas to save all data from new-block print events


Jiloc:
## 1. Description

In recent PRs we added new fields to the print events from the sBTC contracts that we are not currently saving into the DB. 

## 2. Technical Details:
We are currently missing the following fields for each event: [withdrawal-accept](https://github.com/stacks-network/sbtc/blob/main/contracts/contracts/sbtc-registry.clar#L188-L190), [completed-deposit](https://github.com/stacks-network/sbtc/blob/main/contracts/contracts/sbtc-registry.clar#L247-L249), [rotate-keys](https://github.com/stacks-network/sbtc/blob/main/contracts/contracts/sbtc-registry.clar#L281)

### 2.1 Acceptance Criteria:
   - [ ] Extended the `CompletedDepositEvent`, `WithdrawalAcceptEvent` and `KeyRotationEvent` structs to contain the missing data
   - [ ] Extended the corresponding events parsing methods in [RawTupleData ](https://github.com/stacks-network/sbtc/blob/main/signer/src/stacks/events.rs#L276-L572)
   - [ ] Extended the [db migrations](https://github.com/stacks-network/sbtc/blob/main/signer/migrations/0003__create_tables.sql) to contain the new data
   - [ ] Extended the `completed_deposit_event`, `write_withdrawal_accept_event` and `write_rotate_keys_transaction` to save the new data.

PR #833, opened 2024-11-13
chore: mock RotateKeysTransactions with consistent addresses

Jiloc:
## Description

Change the mocking of the `RotateKeysTransaction` `address` field in the tests added in #801. The tests were valid but used the bitcoin `aggregate_key` to compute the `address`.  The changes aim to improve clarity and correctness and avoid confusion for future contributors.

## Changes

- Replaced the incorrect mocking of StacksAddress with the correct value from the wallet when available, or used `StacksAddress::from_public_keys` when it was not.

## Checklist:


signer/src/testing/dummy.rs
```diff
@@ -362,14 +365,24 @@ impl fake::Dummy<SignerSetConfig> for RotateKeysTransaction {
...
+                    .collect::<Vec<_>>(),
+            )
+            .expect("failed to create StacksAddress"),
         ));
+
         RotateKeysTransaction {
             txid: fake::Faker.fake_with_rng(rng),
             address,
-            aggregate_key,
+            aggregate_key: PublicKey::combine_keys(signer_set.iter()).unwrap(),
```
djordon, line 385: Although this is what it was before, you should actually make this a random public key, since that is what it basically is. It is never the combined keys of the signer set public keys.

djordon (approved):
Looks good, left one suggestion about setting the aggregate key.

signer/src/testing/dummy.rs
Jiloc: +1 great point!

Issue #798, opened 2024-11-08
[Feature]: Signers create bitcoin sighashes for bitcoin sign requests


djordon:
# Feature - Signers create bitcoin sighashes for bitcoin sign requests

## 1. Description

This is ticket corresponds to point (4) in 3.1 in https://github.com/stacks-network/sbtc/issues/782. Signers generate sighash digests, from the incoming sign request.

### 1.1 Context & Purpose

Each signer generates the bitcoin transaction and associated sighashes from the sign request.

## 2. Technical Details:

Each signer needs to load the current bitcoin state and use that to generate an `UnsignedTransaction`. Once they have that they can easily create sighashes.

### 2.1 Acceptance Criteria:
   - [ ] Signers fetch the BTC state from the database and construct a transaction and sighash.

## 3. Related Issues and Pull Requests (optional):

Relates to:
* https://github.com/stacks-network/sbtc/issues/793
* https://github.com/stacks-network/sbtc/issues/286
* https://github.com/stacks-network/sbtc/issues/782
* https://github.com/stacks-network/sbtc/issues/794

PR #838, opened 2024-11-14
feat: run blocklist client for deposits that we cannot sign for

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/839, which is a small part of https://github.com/stacks-network/sbtc/issues/798.

## Changes

* Rename the `is_accepted` to `can_accept`, to note whether the deposit was blocked by the blocklist client.
* Always use the blocklist client for each deposit request, even if the signer is not part of the signer set that controls the public key locking the deposit.

## Testing Information

This is largely a refactor, with the only functional change being that we always call the blocklist client. For that change I updated the relevant integration tests.

## Checklist:

### DIFF
--- a/signer/src/storage/in_memory.rs
+++ b/signer/src/storage/in_memory.rs
@@ -314,7 +314,11 @@ impl super::DbRead for SharedStore {
                     .deposit_request_to_signers
                     .get(&(deposit_request.txid, deposit_request.output_index))
                     .map(|signers| {
-                        signers.iter().filter(|signer| signer.can_accept).count() >= threshold
+                        signers
+                            .iter()
+                            .filter(|signer| signer.can_accept && signer.can_sign)
+                            .count()
+                            >= threshold
                     })
                     .unwrap_or_default()
             })
--- a/signer/src/storage/in_memory.rs
+++ b/signer/src/storage/in_memory.rs
@@ -314,7 +314,7 @@ impl super::DbRead for SharedStore {
                     .deposit_request_to_signers
                     .get(&(deposit_request.txid, deposit_request.output_index))
                     .map(|signers| {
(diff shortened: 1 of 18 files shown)


signer/src/storage/in_memory.rs
```diff
@@ -314,7 +314,7 @@ impl super::DbRead for SharedStore {
                     .deposit_request_to_signers
                     .get(&(deposit_request.txid, deposit_request.output_index))
                     .map(|signers| {
-                        signers.iter().filter(|signer| signer.is_accepted).count() >= threshold
+                        signers.iter().filter(|signer| signer.can_accept).count() >= threshold
```
Jiloc, line 317: Should this now be `signers.iter().filter(|signer| signer.can_accept && signer.can_sign).count() >= threshold` to match the Postgres implementation behavior?

Jiloc: If that's the case, we should update the docstring (and maybe the name) of `get_pending_accepted_deposit_request` to specify that it returns requests that can be both accepted and signed to avoid possible confusion!

https://github.com/stacks-network/sbtc/blob/c16e287bdfc39cbe21c17083eff2939748d008cc/signer/src/storage/mod.rs#L62-L64

Jiloc (commented):
I just have a doubt about matching the `get_pending_accepted_deposit_requests` implementations and, maybe, updating its description. Everything else looks good!

signer/src/storage/in_memory.rs
djordon: Fixed them both up in https://github.com/stacks-network/sbtc/pull/838/commits/c00ac5404f4c3a32cb80f04a7a4116c897be4178

Jiloc (approved):
All good!

Issue #782, opened 2024-11-06
[Design]: Signing bitcoin transactions


djordon:
# Design - Signing bitcoin transactions

This ticket holds the design of the signing and validating protocol and how it fits into constructing bitcoin transactions.

## 1. Summary

Our designs for constructing, signing, and validating bitcoin transactions has not been completed yet. Our current code base does no validation for the bitcoin transaction yet. The proposed approach is to have most of the validation logic be handled implicitly by matching on signature digests.

## 2. Context & Purpose

We need to have a protocol where bitcoin transactions are validated by a consensus majority of the signers before a bitcoin transaction can be created and signed.

### 3.1 The proposed Signer bitcoin protocol

When it comes time to construct a bitcoin transaction, the following will happen:

1. The coordinator fetches all relevant deposit and withdrawal requests. It also fetches the last fees paid for transactions in the mempool.
2. The coordinator divides the request IDs into an ordered subset of request IDs. The process for doing so is outlined in [#52](https://github.com/stacks-network/sbtc/issues/52).
3. The coordinator constructs a sign request message containing the ordered list of request IDs, the last fees paid, and the current bitcoin market fee rate and sends that to the other signers.
4. Each signer, upon receiving the sign request message uses their own knowledge of the requests identified by the IDs to construct a chain of bitcoin transactions called a transaction package. Each transaction in the package is formed from one of the disjoint subsets created in step (2).
6. Once a signer has a transaction package, they then responds to the coordinator with the transaction digests that they are willing and able to sign. Each signer stores these digests, along with whether they will sign for them, in their database.
7. The coordinator also constructs a transaction package using the same request IDs from (2) and leads a signing round for the digests that it creates.
8. When deciding whether to participate in a signing round, each signer matches on the digest-to-sign that they receive from the coordinator.
9. After the transaction is signed, the coordinator broadcasts the transaction.

### 3.2 Protocol sensitivity

Each digest from (5), called a [sighash](https://developer.bitcoin.org/devguide/transactions.html?highlight=sighash#signature-hash-types), is constructed as a `SIGHASH_ALL` digest that protects all inputs and outputs from modification. Each signer constructs such a digest for each input to the transaction. A signer will only agree to sign a digest if the following is true:

1. All inputs and outputs in the transaction associated with the digest is “acceptable” to the signer (as in the block checker doesn’t block any of the requests),
[...]

PR #840, opened 2024-11-14
refactor: use `SIGHASH_ALL` explicitly

djordon:
## Description

This was discussed in https://github.com/stacks-network/sbtc/issues/782, and came up as part of https://github.com/stacks-network/sbtc/issues/798.

## Changes

* Explicitly use `TapSighashType::All`, which maps to `SIGHASH_ALL`. This makes the sighash type explicit, which appears to increase the transaction size by at least one byte.

Before we used `TapSighashType::Default`, which is `TapSighashType::All` already. But I have a preference for making the sighash type explicit, and unfortunately this preference has a cost of increasing the transaction size by one byte. Bitcoin usually doesn't change that much and changing the default here seems like a big change so maybe we can rely on the default behavior. If someone else has an opinion here, let me know. 

## Testing Information

This is just a refactor, if tests pass then we're good.

## Checklist:


cylewitruk (approved):
Looks fine :+1:

matteojug (approved):
LGTM, about whether we should make the change or not I am not sure: it is a relatively small increase in cost and we remove some potentially risky thing, but otoh I'm not sure what's the actual possibility of bitcoin lib changing this without a major release and big warnings all around.

cylewitruk:
Late to the party -- I'm in favor of being explicit tbh, primarily because it removes all doubt from the reader of the code.

Issue #800, opened 2024-11-08
[Feature]: Add new fields to the sbtc clarity map


djordon:
# Feature - Add new fields to sbtc clarity maps

## 1. Description

We added new input parameters to some of the contract calls as part of https://github.com/stacks-network/sbtc/issues/583, https://github.com/stacks-network/sbtc/issues/547, and https://github.com/stacks-network/sbtc/issues/488. We should update the `withdrawal-requests` and `completed-deposits` maps in the registry with the `sweep-txid`, `sweep-burn-hash`, and `sweep-burn-height` values when completing the deposit and accepting the withdrawal request.

### 1.1 Context & Purpose

Adding this data is mainly for "completeness" of the protocol, and to give some flexibility of the signers (or any other observers) getting up-to-date data, even if they miss a webhook from stacks-core.

### 2.1 Acceptance Criteria:
   - [ ] Store the `sweep-txid`, `sweep-burn-hash`, and `sweep-burn-height` data in the `completed-deposits` map.
   - [ ] Store the `sweep-txid`, `sweep-burn-hash`, and `sweep-burn-height` data in the `withdrawal-requests` map.

## 3. Related Issues and Pull Requests (optional):

relates to
* https://github.com/stacks-network/sbtc/issues/583,
* https://github.com/stacks-network/sbtc/issues/547,
* https://github.com/stacks-network/sbtc/issues/488

PR #848, opened 2024-11-15
Updated Clarity Maps

setzeus:
## Description
Updates two structs in the sbtc-registry contract to stored data already passed through.
 
Closes: #800 

## Changes
The 'sbtc-registry' Clarity smart contract

## Testing Information
All unit tests effected are now updated.

## Checklist:

### DIFF
--- a/contracts/contracts/sbtc-registry.clar
+++ b/contracts/contracts/sbtc-registry.clar
@@ -40,12 +40,16 @@
 ;; Data structure to map request-id to status
 ;; If status is `none`, the request is pending.
 ;; Otherwise, the boolean value indicates whether
-;; the deposit was accepted.
-(define-map withdrawal-status uint {
-  is-accepted: bool,
-  sweep-txid: (optional (buff 32)),
-  sweep-burn-hash: (optional (buff 32)),
-  sweep-burn-height: (optional uint),
+;; the withdrawal was accepted.
+(define-map withdrawal-status uint bool)
+
+;; Data structure to map successful withdrawal requests
+;; to their respective sweep transaction. Stores the 
+;; txid, burn hash, and burn height.
+(define-map completed-withdrawal-sweep uint {
+  sweep-txid: (buff 32),
+  sweep-burn-hash: (buff 32),
+  sweep-burn-height: uint,
 })
 
 ;; Internal data structure to store completed
@@ -86,6 +90,12 @@
     }))
     none
   )
+)
+
+;; Get a completed withdrawal sweep data by its request ID.
+;; This function returns the fields of the withdrawal-sweeps map.
+(define-read-only (get-completed-withdrawal-sweep-data (id uint))
+  (map-get? completed-withdrawal-sweep id)
 )
 
 ;; Get a completed deposit by its transaction ID & vout index.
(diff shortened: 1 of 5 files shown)


djordon (commented):
Can we also update the `get-withdrawal-request` read only function to flatten out the return tuple. Right now the `status` is a tuple that also contains another `status` field, but I think we want 
```clarity
(merge request (map-get? withdrawal-status id))
```
right?

contracts/contracts/sbtc-registry.clar
```diff
@@ -41,14 +41,22 @@
 ;; If status is `none`, the request is pending.
 ;; Otherwise, the boolean value indicates whether
 ;; the deposit was accepted.
-(define-map withdrawal-status uint bool)
+(define-map withdrawal-status uint {
+  status: bool,
```
djordon, line 45: We should rename this `is-accepted`

djordon (commented):
You're right, the type is fine! But let's rename the field to `is-accepted`.

hstove:
@setzeus I think Daniel meant that the `status` prop should be changed to `is-accepted`, but the `withdrawal-status` map should still be named that IMO.

hstove:
I don't mean to shoehorn in complexity here, but because I've been thinking about costs lately I wanted to mention this.

This modification (increasing the size of these maps) greatly increases `read_length`, especially with the change to `withdrawal-status` (it's like a 100x+ increase in read_length).

It's still a `read_count` of 1, which is the most important metric to optimize, but IMO it's worth considering breaking the `sweep` fields into a separate map. These fields are never used during contract calls (writes), they're just used to store information for off-chain usage.

Even though read_count is all that matters really today, this might not be the case in the future, which is why I think this is worth considering.

djordon:
> I don't mean to shoehorn in complexity here, but because I've been thinking about costs lately I wanted to mention this.
> 
> This modification (increasing the size of these maps) greatly increases `read_length`, especially with the change to `withdrawal-status` (it's like a 100x+ increase in read_length).
> 
> It's still a `read_count` of 1, which is the most important metric to optimize, but IMO it's worth considering breaking the `sweep` fields into a separate map. These fields are never used during contract calls (writes), they're just used to store information for off-chain usage.
> 
> Even though read_count is all that matters really today, this might not be the case in the future, which is why I think this is worth considering.

Great point, let's break the sweep fields into a separate map.

setzeus:
Ah makes sense @hstove, I'll create a separate map.

hstove (approved):

djordon:
@hstove does it make sense to break out things on the deposit side as well? I was thinking that we'd have a map with a boolean value for the status and another with the additional data, just like we do for withdrawals.

hstove:
Yep I like it 👍🏼

djordon (commented):
Awesome, this is the structure that I think we should go with. But with these latest changes I think we should rename the `completed-deposit-sweep` map to `completed-deposits`, have the existing `get-completed-deposit` return the completed-deposits map and add `get-deposit-status` and that returns the `deposit-status`. We'd also have to update our current usage of `get-completed-deposit` to use the `get-deposit-status` function.

djordon (approved):
Looks great! ✅ 🚢

PR #850, opened 2024-11-15
fix: no more flaky `load_latest_deposit_requests_persists_requests_from_past` test

djordon:
## Changes

* Try to make sure that the test is not flaky anymore. Locally it seems to be solid

## Testing Information

The issue with the test was that the block observer wasn't always up-to-date with the blockchain for a crucial part of the test. This fixes that. I cannot get these tests to fail.œ

## Checklist:

### DIFF
--- a/signer/tests/integration/block_observer.rs
+++ b/signer/tests/integration/block_observer.rs
@@ -1,7 +1,6 @@
 use std::collections::HashSet;
 use std::ops::Deref;
 use std::sync::atomic::AtomicBool;
-use std::sync::atomic::AtomicU8;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


matteojug (approved):

cylewitruk:
Just a tip (not necessarily for this PR) -- you could also go the route using [`TestContext::wait_for_signal`](https://github.com/stacks-network/sbtc/blob/dba2823fc736e6c23a2cae2b7f9b7a4283407444/signer/src/testing/context.rs#L120), and adding an event like I did [here](https://github.com/stacks-network/sbtc/blob/dba2823fc736e6c23a2cae2b7f9b7a4283407444/signer/src/transaction_signer.rs#L161) for the block observer. I think this can help us keep this kind of boilerplate out of tests as much as possible, and this kind of event is only ever fired once so there's no performance implications.

Example of usage: https://github.com/stacks-network/sbtc/blob/dba2823fc736e6c23a2cae2b7f9b7a4283407444/signer/src/block_observer.rs#L519

Issue #826, opened 2024-11-12
[Feature]: Create AWS machine image


MCJOHN974:
# Feature - AWS image

## 1. Description

Create AWS machine image which will be synced with testnet and have all software needed to deploy an sBTC signer on testnet.

### 1.1 Context & Purpose

We want to deploy the sBTC to testnet with a similar setup to how we're recommending things be deployed on mainnet. For that we will build an AMI with the necessary applications to run and configure an sBTC signer.

## 2. Technical Details:

The high level steps to do this is as follows:
1. Launch an EC2 instance.
2. Install the necessary applications, which includes bitcoin-core and stacks-core.
3. Make sure the chainstate is up-to-date on the machine.
4. Save the AMI.

We'd like to automate the above process, so I'd recommend using packer for the automated build process, but the first time around you would do this all manually first.

A more detailed breakdown of the steps that are necessary are as follows.

1. Get an AWS account with enough permissions to do the things that need doing.
2. Go to AWS website, launch EC2 instance. 
    1. c6i.xl with 100 GB HD. Use NVMe drives.
    2. setup VPC security groups for SSHing into the machine.
    3. create a keypair.
3. Install dependencies: docker, watchtower, ansible.
5. Add in docker compose file and other config files. This docker compose file will have service entries for everything that is necessary on testnet.
6. Do `docker compose pull` to store the images on the machine.
7. Get the primary testnet chainstate data. This can be done by setting up an EBS volume and attaching it to the image or by pulling it from S3 as part of the AMI build process. The stacks chainstate data can be found here https://archive.hiro.so/testnet/stacks-blockchain/, but there isn't a similar archive for the bitcoin blockchain.
8. Do `docker compose up`. Not sure if we want to do this as part of the automated build process.
10. Build AMI from running EC2 instance.
11. Write files for automatically doing steps (2-10) using packer.

### 2.1 Acceptance Criteria:
   - [ ] We have an AMI with everything that is necessary for primary testnet.
   - [ ] We have a packer file that allows us to recreate the image if we need to do so.

djordon:
This has been finished.

PR #853, opened 2024-11-15
[feature] Build AMI with packer

MCJOHN974:
## Description

Closes: #826

## Changes

Adds packer file for creating an image

## Checklist:

WIP

### DIFF
--- a/packer/template.pkr.hcl
+++ b/packer/template.pkr.hcl
@@ -45,6 +45,8 @@ build {
       "sudo usermod -aG docker ubuntu",
       "git clone https://github.com/stacks-network/sbtc.git",
       "cd sbtc",
+      "",
+      "curl -O https://archive.hiro.so/testnet/stacks-blockchain/testnet-stacks-blockchain-2.5.0.0.7-20240917.tar.gz",
       "sudo docker-compose -f docker/docker-compose.yml --profile sbtc-signer up",
     ]
   }
--- a/packer/template.pkr.hcl
+++ b/packer/template.pkr.hcl
@@ -41,7 +41,7 @@ source "amazon-ebs" "ubuntu" {
   ssh_username            = "ubuntu"
   ami_block_device_mappings {
     device_name           = "/dev/sda1"
-    volume_size           = 100
+    volume_size           = 200
     volume_type           = "gp3"
     delete_on_termination = true
   }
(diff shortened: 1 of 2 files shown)


packer/template.pkr.hcl
```diff
@@ -0,0 +1,53 @@
...
+  sources = ["source.amazon-ebs.ubuntu"]
+
+  provisioner "shell" {
+    inline = [
+      "sudo apt-get update",
+      "sudo apt-get install -y docker.io docker-compose git",
+      "sudo usermod -aG docker ubuntu",
+      "git clone https://github.com/stacks-network/sbtc.git",
+      "cd sbtc",
+      "",
```
djordon, line 48: Is this necessary?

packer/template.pkr.hcl
```diff
@@ -0,0 +1,53 @@
...
+      "curl -O https://archive.hiro.so/testnet/stacks-blockchain/testnet-stacks-blockchain-2.5.0.0.7-20240917.tar.gz",
+      "sudo docker-compose -f docker/docker-compose.yml --profile sbtc-signer up",
```
djordon, line 50: You probably want to run this in detached mode, does it work without?

packer/template.pkr.hcl
```diff
@@ -0,0 +1,53 @@
...
+
+build {
+  sources = ["source.amazon-ebs.ubuntu"]
+
+  provisioner "shell" {
+    inline = [
+      "sudo apt-get update",
+      "sudo apt-get install -y docker.io docker-compose git",
+      "sudo usermod -aG docker ubuntu",
+      "git clone https://github.com/stacks-network/sbtc.git",
```
djordon, line 46: As a follow up to this, we can now use a proper "testnet" docker-compose file that points to the latest images. Then no more building the docker images as part of the AMI creation.

packer/template.pkr.hcl
```diff
@@ -0,0 +1,53 @@
...
+  default = "us-west-2"
+}
+
+source "amazon-ebs" "ubuntu" {
+  ami_name                = "sbtc-signer-image-{{timestamp}}"
+  instance_type           = "c6i.xlarge"
+  region                  = var.aws_region
+  source_ami_filter {
+    filters = {
+      name                = "ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-*"
```
djordon, line 22: We might want to consider using ubuntu 24.04 at this point, unless there is a good reason not to.

djordon (commented):
After a quick read through, this looks great. I just had a few questions.

packer/template.pkr.hcl
```diff
@@ -0,0 +1,53 @@
...
+      root-device-type    = "ebs"
+      virtualization-type = "hvm"
+    }
+    owners      = ["099720109477"] # Canonical
+    most_recent = true
+  }
+  ssh_username            = "ubuntu"
+  ami_block_device_mappings {
+    device_name           = "/dev/sda1"
+    volume_size           = 100
```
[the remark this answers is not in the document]
cylewitruk: I noticed that the hiro archive testnet data seems to grow at a rate of about 100mb/day (stacks chainstate) -- just fyi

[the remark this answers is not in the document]
aldur: Yeah but we will be storing logs, etc. Disk is cheap and it seems like a no-brainer to add more just in case.

packer/template.pkr.hcl
djordon: We should also checkout to a specific commit here.

Issue #623, opened 2024-10-07
[Chore]: switch to using FIRE coordinator


xoloki:
# Chore - switch to using FIRE coordinator

## 1. Description

Currently the WSTS code is using the FROST coordinator state machine. Switch to the FIRE coordinator. This should not require many changes, since they both implement the same trait. But the config object will need to be created using `with_timeouts`.

## 2. Why It's Necessary

The FROST coordinator  does not handle timeouts or misbehavior, so DKG and signing rounds can stall indefinitely even when a valid quorum exists. 

## 3. Execution Checklist

- [ ] update the typedef to point to `FireCoordinator`
- [ ] add timeouts to main config and map them to coordinator config init
- [ ] add integration test that shows DKG and sign complete with offline or misbehaving signers

PR #860, opened 2024-11-15
Use the fire coordinator to handle missing or malicious signers

xoloki:
## Description
This PR changes `signer` to use the `WSTS` `fire::Coordinator` rather than the `frost::Coordinator`.  Whereas the `frost::Coordinator` will wait for all signers to send nonces when starting a signing round, the `fire::Coordinator` will take the first `threshold` of responders and attempt to sign with them.  

If a signer sends a nonce but does not send a signature share before the configured timeout, or sends a signature share that fails to verify, it is excluded from further signing, and another iteration of the current round is attempted.  So at every iteration, we either get a good signature, or we reduce the number of available signers.  Thus, we are guaranteed  to converge to a good signature or a failure. 

The `fire::Coordinator` can also run `DKG` with less than full participation, but we do not anticipate doing so in `sBTC` `v1`.  The timeouts are also configurable, but this PR does not attempt to configure them; rather they are set to `None` as per the existing `wsts_state_machine::CoordinatorStateMachine::new` `fn`.

Closes: #623 

## Testing Information
Existing unit and integration tests should demonstrate sufficiently that this code is working.  It was necessary to set the signing threshold to `3/7` for some unit test to work around test harness incompatibility with the `fire::Coordinator`; this was because the test harness assumed that all signers would send signature shares in every round, which is no longer the case.

## Checklist:

### DIFF
--- a/signer/src/testing/transaction_coordinator.rs
+++ b/signer/src/testing/transaction_coordinator.rs
@@ -137,6 +137,9 @@ where
         mut self,
         delay_to_process_new_blocks: Duration,
     ) {
+        let _ = tracing_subscriber::fmt()
+            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+            .try_init();
         let mut rng = rand::rngs::StdRng::seed_from_u64(46);
         let network = network::InMemoryNetwork::new();
         let signer_info = testing::wsts::generate_signer_info(&mut rng, self.num_signers as usize);
--- a/signer/src/testing/transaction_coordinator.rs
+++ b/signer/src/testing/transaction_coordinator.rs
@@ -137,9 +137,6 @@ where
         mut self,
         delay_to_process_new_blocks: Duration,
     ) {
-        let _ = tracing_subscriber::fmt()
-            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
-            .try_init();
         let mut rng = rand::rngs::StdRng::seed_from_u64(46);
         let network = network::InMemoryNetwork::new();
         let signer_info = testing::wsts::generate_signer_info(&mut rng, self.num_signers as usize);
--- a/signer/src/testing/wsts.rs
+++ b/signer/src/testing/wsts.rs
@@ -124,8 +124,6 @@ impl Coordinator {
             signer_public_keys,
         };
 
-        tracing::info!("signer_key_ids {:?}", &config.signer_key_ids);
-
         let wsts_coordinator = fire::Coordinator::new(config);
 
         Self {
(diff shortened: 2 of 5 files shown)


signer/src/testing/transaction_coordinator.rs
```diff
@@ -137,6 +137,9 @@ where
         mut self,
         delay_to_process_new_blocks: Duration,
     ) {
+        let _ = tracing_subscriber::fmt()
+            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+            .try_init();
```
djordon, line 142: We should remove this. The issue is that only one tracing logger can be instantiated, so if another test adds this we'll get an error in CI.

One alternative is to have a static type that globally instantiates a logging thingy. Alternatively, there is this `test_log` utility that was added recently (I haven't played with it much, it seems to only log errors).

signer/src/wsts_state_machine.rs
```diff
@@ -253,66 +254,17 @@ impl CoordinatorStateMachine {
...
-
-        // Okay we've already gotten the private keys, and we've set the
-        // `party_polynomials` variable in the `WstsCoordinator`. Now we
-        // can just set the aggregate key and move the state to the `IDLE`,
-        // which is the state after a successful DKG round.
-        coordinator.set_aggregate_public_key(Some(aggregate_key.into()));
+        coordinator
+            .set_key_and_party_polynomials(aggregate_key.into(), party_polynomials)
+            .map_err(Error::wsts_coordinator)?;
+        coordinator.current_dkg_id = 1;
```
djordon, line 267: 😍

signer/src/testing/wsts.rs
```diff
@@ -107,7 +107,7 @@ impl Coordinator {
         let num_keys = num_signers;
         let dkg_threshold = num_keys;
         let signer_key_ids = (0..num_signers)
-            .map(|signer_id| (signer_id, std::iter::once(signer_id).collect()))
+            .map(|signer_id| (signer_id, std::iter::once(signer_id + 1).collect()))
```
djordon, line 110: Why the plus 1?

signer/src/transaction_signer.rs
```diff
@@ -800,7 +800,7 @@ mod tests {
             context,
             context_window: 6,
             num_signers: 7,
-            signing_threshold: 5,
+            signing_threshold: 3,
```
djordon, line 803: Nice, but what is the "real" fix? What are the underlying changes that we need to make for these tests to pass as is? We can make those changes in a separate PR.

For this PR I'd add open a tech debt ticket with a brief description of the issue, and add a TODO comment here (and in the equivalent change in the transaction coordinator), pointing to the ticket number.

signer/src/testing/transaction_coordinator.rs
xoloki: Done

signer/src/transaction_signer.rs
xoloki: Done

signer/src/testing/wsts.rs
djordon: 😮 👍🏽

djordon (approved):
Looks good! ✅ 🚢

Issue #748, opened 2024-10-31
[Feature]: Emily client authentication


cylewitruk:
# Feature - Emily client authentication

## 1. Description

The Emily API has some protected endpoints and we currently do not support initializing the Emily client with authentication information. This needs to be done to secure the protected endpoints from tampering from non-signers.

## 2. Technical Details:

The API credentials should be configurable in the config, and these credentials should be used for the relevant Emily API calls.

AshtonStephens:
I'll do this one; I'll need to spin up dev infrastructure to test this so it'll go after #692

aldur:
Note: just marked this as "high" since it is required for client-side authentication.

PR #861, opened 2024-11-17
748 feature emily client authentication

AshtonStephens:
## Description

Adds API Keys to the Emily Client section of the signer config.

Closes: #748 

## Changes

Adds a new configuration option for the EmilyConfig to include an api key for every endpoint.
- Updated the configuration documentation
- Updated relevant configuration creation functions

### Unit tests
I added an additional configuration unit test.

### Integration
I didn't include this in the PR, but I tested this with integration tests against a version of the API deployed that required API Keys. 

For each place in the `signer` integration tests that instantiated an API I replaced the old logic with an actual API and key.

```rust
    let emily_client =
        EmilyClient::try_from(&Url::parse("http://localhost:3031").unwrap()).unwrap();
```
Converted to
```rust
    // We need to populate our databases, so let's fetch the data.
    let emily_config = EmilyEndpointConfig {
        endpoint: Url::parse("https://API_ID.execute-api.us-west-2.amazonaws.com/dev").unwrap(),
        api_key: Some("ACTUAL_API_KEY".to_string()),
    };
    let emily_client =
        EmilyClient::try_from(&emily_config).unwrap();
```
And in doing this everything worked perfectly except the signer integration test `signer::emily::deposit_flow` which needed the following on line `572` so that Emily had time to catch up:

```rust
    tokio::time::sleep(Duration::from_millis(50)).await;
```

## Checklist:

### DIFF
--- a/signer/src/config/default.toml
+++ b/signer/src/config/default.toml
@@ -18,17 +18,18 @@
 # !! Emily API Configuration
 # !! ==============================================================================
 [emily]
-# The URI(s) of the Emily API server to connect to.
+# The endpoint(s) of the Emily API server to connect to.
 #
 # You may specify multiple Emily API servers if you have them. They will be
 # tried round-robin until one succeeds.
 #
-# Format: ["http(s)://<host>:<port>", ..]
+# Format: [ { endpoint = "http(s)://<host>:<port>", api_key = "1234567890abcdef" }, ..]
 # Default: <none>
 # Required: true
 # Environment: SIGNER_EMILY__ENDPOINTS
+# Environment Example: '{"endpoint":"https://api.emilyexample.com","api_key":"1234567890abcdef"},..'
 endpoints = [
-    "http://localhost:3031"
+    { endpoint = "http://localhost:3031" },
 ]
 
 # !! ==============================================================================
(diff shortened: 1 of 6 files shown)


signer/src/config/default.toml
```diff
@@ -18,17 +18,18 @@
...
 # tried round-robin until one succeeds.
 #
-# Format: ["http(s)://<host>:<port>", ..]
+# Format: [ { endpoint = "http(s)://<host>:<port>", api_key = "1234567890abcdef" }, ..]
 # Default: <none>
 # Required: true
 # Environment: SIGNER_EMILY__ENDPOINTS
+# Environment Example: '{"endpoint":"https://api.emilyexample.com","api_key":"1234567890abcdef"},..'
 endpoints = [
-    "http://localhost:3031"
```
cylewitruk, line 31: Hmm, I think I would have preferred to use URI syntax here instead, either authentication (like `http://123123@localhost:3031`) or querystring (`?api_key=123123`), both of which can be extracted from `url::Url`.

AshtonStephens: Hm, that seems odd because it implies something about the structure of the underlying API and the way it expects keys. Why would you prefer this?

AshtonStephens: I think going forward if there's a preference for the way something should be done we should make sure to add that to the ticket, otherwise I'd think all implementations (within reason) should be fair game. This to me looks like a style nit, but if you feel strongly I can go back and change it.

aldur (approved):

PR #872, opened 2024-11-18
refactor: break up the transaction signer into two

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/871.

## Changes

* Create a new event loop only for deciding on requests (and handling other signers decisions). Remove the equivalent code from the tx-signer.
* Move over the old tests.
* Add and use new event variants for the new event type.

## Testing Information

Some of the integration tests come with a slight update. But almost everything else is a copy paste.

## Checklist:


cylewitruk (approved):
Looks good in general, has been something I've wanted to address (as well as splitting up the coordinator a bit). I don't _love_ `RequestDecider` as a name but we can always change that later :)

djordon:
> Looks good in general, has been something I've wanted to address (as well as splitting up the coordinator a bit). I don't _love_ `RequestDecider` as a name but we can always change that later :)

Well, I don't like it much either, I'm very much open to suggestions. Oh, I should ask ChatGPT for some suggestions. Also, I'm going to hold off on merging it until after the parent PRs have been reviewed.

signer/src/request_decider.rs
```diff
@@ -0,0 +1,435 @@
...
+    pub context_window: u16,
+}
+
+impl<C, N, B> RequestDeciderEventLoop<C, N, B>
+where
+    C: Context,
+    N: MessageTransfer,
+    B: BlocklistChecker,
+{
+    /// Run the signer event loop
```
matteojug, line 57: nit: update comment

signer/src/testing/request_decider.rs
```diff
@@ -0,0 +1,541 @@
...
+use crate::testing::storage::model::TestData;
+
+use hashbrown::HashSet;
+use rand::SeedableRng as _;
+use tokio::sync::broadcast;
+use tokio::time::error::Elapsed;
+
+use super::context::*;
+
+/// A test harness for the signer event loop.
```
matteojug, line 33: nit: update comment

matteojug (approved):

djordon:
Hmmm, how about `BlocklistEventLoop` or `TxCheckerEventLoop`?

signer/src/request_decider.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/872/commits/771e5076cff87a960485fea53ff6981529e8c1b0.

signer/src/testing/request_decider.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/872/commits/abb9db7529ca6ce90001ca9433bf0ca72e675117

Issue #794, opened 2024-11-08
[Feature]: Signers store bitcoin sighashes and decisions


djordon:
# Feature - Signers store bitcoin sighashes and decisions

## 1. Description

This is ticket corresponds to point (5) in 3.1 in https://github.com/stacks-network/sbtc/issues/782. Signers stores sighash digests, along with whether they will sign for them, in their databases.

### 1.1 Context & Purpose

Each signer records their decision for easy look-up before participating in a signing round.

## 2. Technical Details:

We need a new table that stores the sighashes and the signers decisions.

### 2.1 Acceptance Criteria:
   - [ ] Signers store enough information in the database to know whether they will participate in a signing round for a sighash.

## 3. Related Issues and Pull Requests (optional):

Relates to:
* https://github.com/stacks-network/sbtc/issues/793
* https://github.com/stacks-network/sbtc/issues/286
* https://github.com/stacks-network/sbtc/issues/782

PR #877, opened 2024-11-18
794 save bitcoin sighashes to db

Jiloc:
## Description

Extend DB schemas and add queries to saving and retrieving the sighashes of the deposit and withdrawal transactions.

Connected to: #794 

## Changes
- add `bitcoin_txs_sighashes` and `bitcoin_withdrawals_outputs` tables
- move `BitcoinTxSigHash` and `BitcoinWithdrawalOutput` into models.rs
- add write and get queries for single and multiple sighashes and outputs.

## Testing Information
- add integration tests for get and write queries

## Checklist:


signer/migrations/0003__create_tables.sql
```diff
@@ -319,3 +319,30 @@ CREATE TABLE sbtc_signer.swept_deposits (
...
+    sighash BYTEA NOT NULL,
+    prevout_type sbtc_signer.prevout_type NOT NULL,
+    validation_result TEXT NOT NULL,
+    is_valid_tx BOOLEAN NOT NULL,
+    construction_version TEXT NOT NULL,
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL
+);
+
+CREATE TABLE sbtc_signer.bitcoin_withdrawals_outputs (
+    request_id BIGINT PRIMARY KEY,
```
Jiloc, line 338: I noticed that wherever the withdrawal's request_id is used in ther other tables as a primary key, it is always paired with the block_hash. Not sure if it's required here, or if it's necessary to declare a foreign key to specific table

djordon: It should be paired with the block hash because withdrawal requests are only unique that way. Reorgs on Stacks can lead to a withdrawal request with the same ID but where the contents of the request can be different from the earlier one.

signer/migrations/0003__create_tables.sql
```diff
@@ -319,3 +319,31 @@ CREATE TABLE sbtc_signer.swept_deposits (
...
+
+CREATE TABLE sbtc_signer.bitcoin_txs_sighashes (
+    txid BYTEA PRIMARY KEY,
+    chain_tip BYTEA NOT NULL,
+    prevout_txid BYTEA NOT NULL,
+    prevout_output_index INTEGER NOT NULL,
+    sighash BYTEA NOT NULL,
+    prevout_type sbtc_signer.prevout_type NOT NULL,
+    validation_result TEXT NOT NULL,
+    is_valid_tx BOOLEAN NOT NULL,
```
cylewitruk, line 332: Could you pad these tables out with some comments so it's clear what everything is?

signer/migrations/0003__create_tables.sql
```diff
@@ -319,3 +319,31 @@ CREATE TABLE sbtc_signer.swept_deposits (
 -- unique per 'sweep_transaction_id'.
 CREATE UNIQUE INDEX uix_swept_deposits_req_txid_req_output_index_pkgd_txid
     ON sbtc_signer.swept_deposits(deposit_request_txid, deposit_request_output_index, sweep_transaction_txid);
+
+
+CREATE TABLE sbtc_signer.bitcoin_txs_sighashes (
```
cylewitruk, line 324: I'd do `tx` instead of `txs` just to be consistent with the `bitcoin_tx_inputs` etc. tables.

signer/migrations/0003__create_tables.sql
```diff
@@ -319,3 +319,31 @@ CREATE TABLE sbtc_signer.swept_deposits (
...
+    txid BYTEA PRIMARY KEY,
+    chain_tip BYTEA NOT NULL,
+    prevout_txid BYTEA NOT NULL,
+    prevout_output_index INTEGER NOT NULL,
+    sighash BYTEA NOT NULL,
+    prevout_type sbtc_signer.prevout_type NOT NULL,
```
cylewitruk, line 330: Do we need `prevout_type` here? We should be able to get it by joining to `bitcoin_tx_*` tables,  right?

signer/src/storage/mod.rs
```diff
@@ -464,4 +477,28 @@ pub trait DbWrite {
...
     ) -> impl Future<Output = Result<(), Error>> + Send;
+
+    /// Write the bitcoin tx sighash to the database.
+    fn write_bitcoin_tx_sighash(
+        &self,
+        sighash: &model::BitcoinTxSigHash,
+    ) -> impl Future<Output = Result<(), Error>> + Send;
+
+    /// Write the bitcoin transactions sighaes to the database.
+    fn write_bitcoin_txs_sighashes(
```
cylewitruk, line 488: I would probably just have the write-multiple versions of these methods, and take a slice instead of vec. Then if you just want to write one item then you just do `&[item]`.

signer/src/storage/postgres.rs
```diff
@@ -2618,6 +2670,247 @@ impl super::DbWrite for PgStore {
...
+        let mut chain_tip = Vec::with_capacity(sighashes.len());
+        let mut prevout_txid = Vec::with_capacity(sighashes.len());
+        let mut prevout_output_index = Vec::with_capacity(sighashes.len());
+        let mut sighash = Vec::with_capacity(sighashes.len());
+        let mut prevout_type = Vec::with_capacity(sighashes.len());
+        let mut validation_result = Vec::with_capacity(sighashes.len());
+        let mut is_valid_tx = Vec::with_capacity(sighashes.len());
+        let mut construction_version = Vec::with_capacity(sighashes.len());
+
+        for tx_sighash in sighashes {
```
cylewitruk, line 2729: Just a note, nothing to change for now -- this works and does achieve atomicity, but it makes for a pretty convoluted implementation (both rust and sql). I think we should opt for readability/maintainability and embrace transactions instead.

Jiloc: Yeah, I am not a bit fan of the implementation either. But I thought of keeping it consistent with the other similar methods

signer/migrations/0003__create_tables.sql
djordon: Hmmm, right now the `bitcoin_tx_*` tables only have confirmed inputs and outputs and I'm not sure if we should add this data there. @cylewitruk are you thinking that we should add this stuff there?

signer/src/storage/postgres.rs
djordon: Man, good points on all sides here. I don't like our messy existing implementations either (and I added them 😞), and I think we should try to be consistent too.

Slight note that I've been wary of using sqlx transactions because I got bit by a bug in an earlier version of the library. It's probably fixed by now, but that was part of the why I've (perhaps mistakenly) tried not to use them.

signer/migrations/0003__create_tables.sql
cylewitruk: @djordon I guess I was thinking that these prevouts should already be confirmed and stored in those tables when processing in the block observer? I've probably got my head spun around though :sweat_smile:

djordon: Ohh, yeah I see. Hmm, I don't think deposits get added to those tables, just donations and sweep/sBTC transactions.

Just checked and we don't add deposits to those tables. But maybe we should?

cylewitruk: Yeah maybe we should for consistency, like to be able to say that those tables include all relevent inputs/outputs for all sbtc-relevant txs

signer/src/storage/postgres.rs
cylewitruk: Transactions are such a fundamental/primary concept/feature of rdbms that if we don't trust the library to properly handle them then I wouldn't trust anything else in it tbh 😅

djordon: I haven't seen any issues with using their `PgPool` executor though, and I haven't seen that transaction bug recently. So I don't think that we should switch libraries, at least not now.

But, that's a good point. I was just explaining why I shied away from them.

signer/src/storage/model.rs
```diff
@@ -1109,6 +1112,109 @@ impl ScriptPubKey {
...
+#[cfg_attr(feature = "testing", derive(fake::Dummy))]
+pub enum ConstructionVersion {
+    /// The first version for constructing a UTXO
+    V0,
+}
+
+/// The sighash and enough metadata to piece together what happened.
+#[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord, sqlx::FromRow)]
+#[cfg_attr(feature = "testing", derive(fake::Dummy))]
+pub struct BitcoinTxSigHash {
```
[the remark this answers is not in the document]
Jiloc: Roger for sqlx! Just to clarify, when you say "to move it back" you mean to move `BitcoinTxSigHash`  and `BitcoinWithdrawalOuput` in validation.rs?

signer/src/storage/postgres.rs
Jiloc: I'll just throw two alternative implementations for you guys to consider! Personally, I lean slightly toward the first one for its simplicity and readability.  It is theorerically slower, but I can't imagine a scenario where we'd need to handle hundreds of deposits or withdrawals in a single batch, so the performance difference should be negligible... The second one is still a bit weird, but maybe less that the current implementation?

```
    async fn write_bitcoin_withdrawals_outputs(
        &self,
        withdrawal_outputs: &[model::BitcoinWithdrawalOutput],
    ) -> Result<(), Error> {
        if withdrawal_outputs.is_empty() {
            return Ok(());
        }

        let mut transaction = self.0.begin().await.map_err(Error::SqlxQuery)?;

        for output in withdrawal_outputs {
            sqlx::query(
                r#"
                INSERT INTO sbtc_signer.bitcoin_withdrawals_outputs (
                    bitcoin_txid
                  , bitcoin_chain_tip
                  , output_index
                  , request_id
                  , stacks_txid
                  , stacks_block_hash
                  , validation_result
                  , is_valid_tx
                )
                VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
                ON CONFLICT DO NOTHING
                "#,
            )
            .bind(output.bitcoin_txid)
            .bind(output.bitcoin_chain_tip)
            .bind(i32::try_from(output.output_index).map_err(Error::ConversionDatabaseInt)?)
            .bind(i64::try_from(output.request_id).map_err(Error::ConversionDatabaseInt)?)
            .bind(output.stacks_txid)
            .bind(output.stacks_block_hash)
            .bind(output.validation_result)
            .bind(output.is_valid_tx)
            .execute(&mut *transaction)
            .await
            .map_err(Error::SqlxQuery)?;
        }
        transaction.commit().await.map_err(Error::SqlxQuery)?;

        Ok(())
    }
```

```
    async fn write_bitcoin_withdrawals_outputs(
        &self,
        withdrawal_outputs: &[model::BitcoinWithdrawalOutput],
    ) -> Result<(), Error> {
        if withdrawal_outputs.is_empty() {
            return Ok(());
        }

        // Build the VALUES part of the query dynamically
        let mut param_count = 1;
        let values: Vec<String> = withdrawal_outputs
            .iter()
            .map(|_| {
                let value = format!(
                    "(${},${},${},${},${},${},${},${})",
                    param_count,
                    param_count + 1,
                    param_count + 2,
                    param_count + 3,
                    param_count + 4,
                    param_count + 5,
                    param_count + 6,
                    param_count + 7
                );
                param_count += 8;
                value
            })
            .collect();

        let query = format!(
            r#"
            INSERT INTO sbtc_signer.bitcoin_withdrawals_outputs (
                bitcoin_txid,
                bitcoin_chain_tip,
                output_index,
                request_id,
                stacks_txid,
                stacks_block_hash,
                validation_result,
                is_valid_tx
            )
            VALUES {}
            ON CONFLICT DO NOTHING
            "#,
            values.join(",")
        );

        // Build query parameters
        let mut query = sqlx::query(&query);
        for output in withdrawal_outputs {
            query = query
                .bind(output.bitcoin_txid)
                .bind(output.bitcoin_chain_tip)
                .bind(i32::try_from(output.output_index).map_err(Error::ConversionDatabaseInt)?)
                .bind(i64::try_from(output.request_id).map_err(Error::ConversionDatabaseInt)?)
                .bind(output.stacks_txid)
                .bind(output.stacks_block_hash)
                .bind(output.validation_result)
                .bind(output.is_valid_tx);
        }
        query.execute(&self.0).await.map_err(Error::SqlxQuery)?;

        Ok(())
    }
```

signer/src/storage/model.rs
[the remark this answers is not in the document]
djordon: Yeah that was what I was referring to. But actually, it makes sense to have them here too.

signer/src/storage/postgres.rs
djordon: Between the two I'd go with the first one. I'd default to not using a transaction too for the reasons stated above.

signer/migrations/0003__create_tables.sql
Jiloc: fixed in https://github.com/stacks-network/sbtc/pull/877/commits/0703ec69542199460a6c69a92ea352d432afc624

signer/migrations/0003__create_tables.sql
Jiloc: fixed in https://github.com/stacks-network/sbtc/commit/0703ec69542199460a6c69a92ea352d432afc624

signer/migrations/0003__create_tables.sql
Jiloc: Note: Since we’ve removed the get query for this object, no instances of this struct are being created from the database. So, keeping or removing the field, without it being present in another table, won’t cause any issues for now. 

To keep things moving, I propose leaving the field for now (it's just an extra integer, so it shouldn’t cause any issues) and opening a separate issue to save all sBTC transactions in the block observer, what do you guys think?

signer/src/storage/mod.rs
Jiloc: soved in https://github.com/stacks-network/sbtc/pull/877/commits/d0d00ff4170e84eb67a207b1d71663e901da911d

djordon (approved):
Looks good ✅

signer/migrations/0003__create_tables.sql
```diff
@@ -310,11 +310,61 @@ CREATE TABLE sbtc_signer.swept_deposits (
...
 CREATE UNIQUE INDEX uix_swept_deposits_req_txid_req_output_index_pkgd_txid
     ON sbtc_signer.swept_deposits(deposit_request_txid, deposit_request_output_index, sweep_transaction_txid);
+
+
+CREATE TABLE sbtc_signer.bitcoin_tx_sighashes (
+    -- The transaction ID of the bitcoin transaction.
+    txid BYTEA PRIMARY KEY,
+    -- The bitcoin chain tip when the sign request was submitted.
+    chain_tip BYTEA NOT NULL,
+    -- The txid that created the output that is being spent.
```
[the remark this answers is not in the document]
Jiloc: done in https://github.com/stacks-network/sbtc/pull/877/commits/9528d26029571f2bd56d08ba1a879084e8991eed , open to suggestions for a shorter name :P

signer/migrations/0003__create_tables.sql
Jiloc: Issue opened https://github.com/stacks-network/sbtc/issues/931

signer/migrations/0003__create_tables.sql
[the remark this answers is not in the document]
djordon: No that's perfect

signer/src/storage/postgres.rs
cylewitruk: The first one will be much more performant than the second one, a lot of write overhead comes from the begin/commit and index statistics updates, which when batching is limited to once if you use a transaction, otherwise each individual statement has its own begin/commit.

cylewitruk: But we can do this later :)

djordon: Yeah sure, and we can revamp the old ones too.

Issue #847, opened 2024-11-14
[Feature]: Add missing proto conversions for serialized messages


matteojug:
# Feature - Add missing proto conversions for serialized messages

## 1. Description

Subtask of https://github.com/stacks-network/sbtc/issues/412.

Some proto definitions for internal types used in communications are missing conversion functions (eg, SignerState).

### 1.1 Context & Purpose

To switch to protobuf-based messages we need to have all the messages defined and conversions to internal types in place.

### 2.1 Acceptance Criteria:
   - [ ] Criterion 1
   - [ ] Criterion 2
   - [ ] Criterion 3

matteojug:
Done in https://github.com/stacks-network/sbtc/pull/879

PR #879, opened 2024-11-18
Add missing proto conversions

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/847

## Changes

Add the necessary conversions between proto and internal types used in messages. 

## Testing Information

Tested e2e on https://github.com/stacks-network/sbtc/pull/880. Add top level serde tests.

## Checklist:


signer/src/stacks/contracts.rs
```diff
@@ -1009,13 +1009,13 @@ impl AsContractCall for RejectWithdrawalV1 {
 pub struct RotateKeysV1 {
     /// The new set of public keys for all known signers during this
     /// PoX cycle.
-    new_keys: BTreeSet<PublicKey>,
+    pub new_keys: BTreeSet<PublicKey>,
```
matteojug, line 1012: Now we need to deserialize it from (potentially) arbitrary values. We could add custom methods to avoid making those `pub`.

signer/src/proto/convert.rs
```diff
@@ -404,6 +1701,8 @@ mod tests {
     #[test_case(PhantomData::<(SignerWithdrawalDecision, proto::SignerWithdrawalDecision)>; "SignerWithdrawalDecision")]
     #[test_case(PhantomData::<(BitcoinTransactionSignAck, proto::BitcoinTransactionSignAck)>; "BitcoinTransactionSignAck")]
     #[test_case(PhantomData::<(StacksTransactionSignature, proto::StacksTransactionSignature)>; "StacksTransactionSignature")]
+    #[test_case(PhantomData::<(SignerMessage, proto::SignerMessage)>; "SignerMessage")]
+    #[test_case(PhantomData::<(Signed<SignerMessage>, proto::Signed)>; "Signed")]
```
cylewitruk, line 1705: My concern here was that maybe we're not ensuring that we're testing all of the possible `Payload` combinations, and thus the code above, since it will be faked (randomized). Not sure if there's other tests somewhere covering this though.

signer/src/proto/convert.rs
```diff
@@ -379,6 +1669,13 @@ mod tests {
         }
     }
 
+    impl Dummy<Unit> for Point {
+        fn dummy_with_rng<R: rand::Rng + ?Sized>(_: &Unit, rng: &mut R) -> Self {
+            let public_key: PublicKey = Faker.fake_with_rng(rng);
+            public_key.into()
```
djordon, line 1675: I would change this to be something like
```rust
let number: [u8; 32] = Faker.fake_with_rng(rng);
let scalar = p256k1::scalar::Scalar::from(number);
Point::from(scalar)
```
This approach catches the case where we would generate a point that is not a public key (even though we will never trigger it by chance).

signer/src/proto/convert.rs
djordon: Yeah, I have the same concern. We should add test cases for the types individually.

djordon (commented):
🙏🏽 This largely looks good. I think we should add more specific test cases. I'd focus on the WSTS ones since we also store those in the database.

Wait, those are the more annoying ones since we do not implement `Faker` for them already 😞. Okay, maybe open a ticket to round out our test coverage here?

signer/src/proto/convert.rs
matteojug: Added the tests placeholder in here, I'll enable them (with the required mocking) in a separate PR (before merging https://github.com/stacks-network/sbtc/pull/880)

signer/src/proto/convert.rs
```diff
@@ -1365,18 +1365,25 @@ impl TryFrom<proto::Signed> for Signed<SignerMessage> {
...
+            true => 3u8,  // SECP256K1_TAG_PUBKEY_ODD
+            false => 2u8, // SECP256K1_TAG_PUBKEY_EVEN
+        }];
+        let x: [u8; 32] = value.x_coordinate.required()?.into();
+        let compressed: Compressed = [prefix, &x]
+            .concat()
+            .as_slice()
             .try_into()
             .map_err(|_| Error::TypeConversion)?;
         Self::try_from(&compressed).map_err(|_| Error::TypeConversion)
```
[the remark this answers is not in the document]
matteojug: Nice! I couldn't find the right way to do the conversion and I was quite annoyed by the previous version. Now it's way better, thanks!

signer/src/proto/convert.rs
matteojug: Done in https://github.com/stacks-network/sbtc/pull/899

PR #885, opened 2024-11-19
fix: bail on `None` from the stream

djordon:
## Description

This was supposed to be added in https://github.com/stacks-network/sbtc/pull/870, but I missed it.

## Changes

* Bail on more errors when polling the stream.

## Checklist:

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -931,8 +931,15 @@ where
         loop {
             // Let's get the next message from the network or the
             // TxSignerEventLoop.
+            //
+            // If signal_stream.next() returns None then one of the
+            // underlying streams has closed. That means either the
+            // network stream, the internal message stream, or the
+            // termination handler stream has closed. This is all bad,
+            // so we trigger a shutdown.
             let Some(msg) = signal_stream.next().await else {
-                continue;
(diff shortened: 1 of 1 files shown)


cylewitruk (approved):

signer/src/transaction_coordinator.rs
```diff
@@ -931,8 +931,15 @@ where
...
+            //
+            // If signal_stream.next() returns None then one of the
+            // underlying streams has closed. That means either the
+            // network stream, the internal message stream, or the
+            // termination handler stream has closed. This is all bad,
+            // so we trigger a shutdown.
             let Some(msg) = signal_stream.next().await else {
-                continue;
+                self.context.get_termination_handle().signal_shutdown();
+                return Err(Error::SignerShutdown);
```
Jiloc, line 942: If this is unexpected, should we log an error?

djordon: It's not obvious, but we already do. Couldn't hurt to add another one though, will update.

djordon: logs added in https://github.com/stacks-network/sbtc/pull/885/commits/a23b1418c11e54a6a64ab94f0b7e0f0b74f7b43b.

Jiloc (approved):
all good!

PR #889, opened 2024-11-19
Testnet docker-compose setup

aldur:


docker/testnet/stacks/Config.toml
```diff
@@ -0,0 +1,89 @@
...
+
+[[burnchain.epochs]]
+epoch_name = "2.5"
+start_height = 6
+
+# Use the value here for the configuration, replace whatever was in your
+# file before if it is different.
+[[burnchain.epochs]]
+epoch_name = "3.0"
+start_height = 56_457
```
matteojug, line 89: nit: should we also add 3.1?

docker/testnet/docker-compose.testnet.yml
```diff
@@ -0,0 +1,47 @@
+services:
+  postgres:
+    image: postgres:16.3-alpine
+    stop_grace_period: 5s
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: postgres
```
matteojug, line 7: we did remove the port exposure, should we also suggest to use some different password here?

docker/testnet/stacks/Config.toml
```diff
@@ -0,0 +1,89 @@
...
+auth_token = "XXX"
+private_neighbors = false
+
+# Add sbtc-signer as an event observer.
+# NOTE: See the readme for how to set this value correctly, your endpoint
+# may not be named `sbtc-signer:8801`.
+#
+# It's perfectly fine to have multiple event observers, just add this
+# section below your existing one, it'll adds the event observer to the
+# events_observers list.
```
matteojug, line 30: Should we add a note here that while it's fine, you one of them is down/not responding it will block the stacks node? (unless they already fixed this in stacks-core)

docker/testnet/config/signer-config.toml.sample
```diff
@@ -0,0 +1,71 @@
...
+# TODO: Change this to local stacks container
+endpoints = [
+    "http://TODO:20443",
+]
+
+nakamoto_start_height = 56457
+
+[signer]
+# TODO: Change this to signer private key
+private_key = "0000000000000000000000000000000000000000000000000000000000000001"
```
matteojug, line 31: nit: in the readme we say to use the `.env`, if so should we restore the `# NOTE: Set through Environment: SIGNER_SIGNER__PRIVATE_KEY`?

matteojug (approved):
Just some nits

docker/testnet/stacks/Config.toml
aldur: Yes we should! Done

docker/testnet/docker-compose.testnet.yml
aldur: We could, but then users would need to change the user/config on the provided PG dump as well.

docker/testnet/stacks/Config.toml
aldur: They didn't, adding a note

docker/testnet/config/signer-config.toml.sample
aldur: Done!

docker/testnet/docker-compose.testnet.yml
matteojug: If we are talking only about the password then it's fine: pgdump will include statements like `ALTER SCHEMA sbtc_signer OWNER TO postgres;`, but not the `postgres` user itself; this applies for "proper" pgdumps (which we should start to use instead of the disk image).
The would still need to update the db config. Maybe we can mention it on the readme (_if you want, you can change the db password and update here and here_).

BowTiedDevOps: What about setting it in `.env`, and then document that they'd only have to change it there? It can be then reused in the compose file and config, example:
```sh
# .env
PG_PASSWORD=postgres # PostgreSQL Database Password, modify this to a password of your choice
```

```yml  
# docker-compose.yml
services:
  postgres:
    environment:
      POSTGRES_PASSWORD: ${PG_PASSWORD:-postgres}

  sbtc-signer:
    environment:
      PG_PASSWORD: ${PG_PASSWORD:-postgres}
    volumes:
      - ./config/signer-config.toml:/signer-config.toml.in
    entrypoint: 
      - /bin/bash
      - -c
      - |
        envsubst < /signer-config.toml.in > /signer-config.toml
        /usr/local/bin/signer -c /signer-config.toml --migrate-db -o json | tee /logs/sbtc_signer.logs
```

```toml
# signer-config.toml
[signer]
db_endpoint = "postgresql://postgres:$PG_PASSWORD@postgres:5432/signer"
```

docker/testnet/config/signer-config.toml.sample
BowTiedDevOps: Can be done in .env as well for security, see my above comment

BowTiedDevOps:
Overall looks good - my only suggestions here would be setting some sensitive fields (i.e private key, postgres password, Emily key, bitcoin rpc url or auth) along with not-so-sensitive ones (configurations etc) in `.env` and parsing them through the docker compose file to either the service's `environment:` or the configuration files of the services through `environment:` and `envsubst`. This way, everything that can or has to be modified by users can be done in 1 place. This would make it easier for people to understand/follow the process, and also simplifies the docs page.

Issue #855, opened 2024-11-15
[Feature]: Add an endpoint in Emily for getting the sBTC cap


djordon:
# Feature - Add an endpoint in Emily for getting the sBTC cap

## 1. Description

Add an endpoint to emily for getting any sBTC related caps. This is just for retrieving the cap, not for setting it.

### 1.1 Context & Purpose

This is part of https://github.com/stacks-network/sbtc/issues/854

## 2. Technical Details:

Make a dynamodb table with the caps. And add accessors for getting the caps. 
The format of the caps should follow what is laid out in https://github.com/stacks-network/sbtc/pull/811/files#diff-13e21a1d09b66d39f5304cab3fecb3591df347ab1952c80747cc94e2be0fb2f5R46-R56, except use unsigned integer types.

### 2.1 Acceptance Criteria:
   - [ ] We have an endpoint that retries the caps.

## 3. Related Issues and Pull Requests (optional):

We have two design tickets for this:
* https://github.com/stacks-network/sbtc/issues/854
* https://github.com/stacks-network/sbtc/issues/562
* https://github.com/stacks-network/sbtc/issues/791

Issue #857, opened 2024-11-15
(optional) we have a simple way of changing the cap in Emily


djordon:

AshtonStephens:
Done with #896

PR #896, opened 2024-11-19
855 Feature add an endpoint in emily for getting the sbtc cap

AshtonStephens:
## Description

Adds an endpoint in Emily for getting the sbtc cap.

It creates three API calls:

| uri | method | returns |
|-|-|-|
| `/limits` | `GET` | The limits of the whole protocol in the form of a `Limits` object. |
| `/limits` | `POST` | Sets the limits for the whole protocol in the form of a `Limits` object. | 
| `/limits/{account}` | `GET` | Gets the limits for just the account in the form of the `AccountLimits` object |
| `/limits/{account}` | `POST` | Sets the limits for just the account in the form of the `AccountLimits` object | 

Closes: #855, #857

Note that these caps are accessible from the base API via an API key - the same API keys that the signers have access to. This means that the signers will be able to change the cap themselves. This is a problem that I'll solve in a follow on PR by deploying a version of the API that has access to these endpoints separately. I've already partially done this in my local branch but I'm opting to push this first in order to merge iterative improvements.

I've also chosen to make the caps only accessible to the signers because the endpoint suggests that account level caps are supported where they are not.

#### Limits

```rust
/// Represents the current sBTC limits.
pub struct Limits {
    /// Represents the total cap for all pegged-in BTC/sBTC.
    pub peg_cap: Option<u64>,
    /// Per deposit cap. If none then there is no cap.
    pub per_deposit_cap: Option<u64>,
    /// Per withdrawal cap. If none then there is no cap.
    pub per_withdrawal_cap: Option<u64>,
    /// Represents the individual limits for requests coming from different accounts.
    pub account_caps: HashMap<String, AccountLimits>,
}
```

Note that when stetting the limit with the `POST /limit` api call individual account caps can be updated by specifying them in the `account_caps` field. All accounts that are specified will be updated with new caps, and any that are not specified will be left alone.

#### AccountLimits

```rust
/// The representation of a limit for a specific account.
pub struct AccountLimits {
    /// Represents the current sBTC limits.
    pub peg_cap: Option<u64>,
    /// Per deposit cap. If none then the cap is the same as the global per deposit cap.
    pub per_deposit_cap: Option<u64>,
    /// Per withdrawal cap. If none then the cap is the same as the global per withdrawal cap.
    pub per_withdrawal_cap: Option<u64>,
}
```

To "delete" a cap for a specific account, one just needs to set every cap field on the account to be `null`. Internally the entry will not be deleted, but the response will filter it out when responding to the API call.

## Changes

- Updates the CDK template to spin up a limits table
- updates the lambda configuration to consume this table
- Creates a new limits endpoint
- Adds integration tests for the new limits endpoint
- Updates the openapi spec with the new endpoints.

## Testing Information

Tested with integration tests provided in the PR.

## Checklist:


cylewitruk (approved):
Looks good, agreed that @AshtonStephens will make some iterative changes to this in follow-up PR's and that we can start integrating this now.

PR #905, opened 2024-11-20
feat: send sbtc requests context with signers and store sighashs to db

Jiloc:
## Description

Closes: #794 

## Checklist:


signer/src/message.rs
```diff
@@ -322,6 +335,121 @@ pub struct BitcoinTransactionSignAck {
...
+    pub fee_rate: f64,
+    /// The total fee amount and the fee rate for the last transaction that
+    /// used this UTXO as an input.
+    pub last_fees: Option<FeesMessage>,
+    // /// The total fee paid in sats for the transaction.
+    // pub last_fee_total: Option<u64>,
+    // /// The fee rate paid in sats per virtual byte.
+    // pub last_fee_rate: Option<f64>,
+}
+
```
Jiloc, line 452: What other information should be sent to the signers?

signer/src/transaction_coordinator.rs
```diff
@@ -400,14 +402,28 @@ where
...
+                .map(|tx| SbtcRequestsContext::from(&tx.requests))
+                .collect(),
+            fee_rate: pending_requests.signer_state.fee_rate,
+            last_fees: pending_requests.signer_state.last_fees.map(Into::into),
+        };
+
+        // Share the list of requests with the signers.
+        self.send_message(sbtc_requests, bitcoin_chain_tip).await?;
+        // Wait for the signers to receive and process the requests.
+        tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
```
Jiloc, line 426: Currently, the simplest solution is to wait for a fixed time.

An possible approach could be switching to an ACKs and maintaining a count for each signhash indicating whether it is accepted or rejected. We would stop as soon as one of the following conditions is met:

All signhashes exceed the threshold of sign promises -> proceed with the signing round.
Any signhash reaches a rejection count that prevents consensus -> skip the signing round (should we invalidate the request?).
Maximum time has elapsed -> attempt the signing round anyway?

djordon, line 426: We might want to update the comment a little. We wait because if we proceed immediately and send the next message, there is a greater change that the other signers will receive subsequent messages before the `SbtcRequestsContextMessage` one.

signer/src/message.rs
```diff
@@ -322,6 +335,47 @@ pub struct BitcoinTransactionSignAck {
     pub txid: bitcoin::Txid,
 }
 
+/// The message version of an the [`OutPoint`].
+#[derive(Debug, Copy, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct OutPointMessage {
```
djordon, line 340: I'd rename this to `BitcoinOutPoint`.

djordon: Actually, we can keep it as is. When we switch to protobufs we can remove this and use the `proto::OutPoint` type.

signer/src/message.rs
```diff
@@ -322,6 +335,47 @@ pub struct BitcoinTransactionSignAck {
...
+
+impl From<OutPointMessage> for OutPoint {
+    fn from(val: OutPointMessage) -> Self {
+        OutPoint { txid: val.txid, vout: val.vout }
+    }
+}
+
+/// The transaction context needed by the signers to reconstruct the transaction.
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct SbtcRequestsContextMessage {
```
djordon, line 364: I'd rename this to `BitcoinPreSignRequest` or something similar.

Later, the context object in the validation file will separate into two, and the meaning of this will change slightly. The signer be validating the information in this object received from the coordinator with "context" information that the receiving signer generates. It'll be similar to how things are done on the stacks side in `contracts.rs`.

signer/src/message.rs
```diff
@@ -452,6 +507,48 @@ impl wsts::net::Signable for WstsMessage {
...
+        for request in &self.requests {
+            request.hash(hasher);
+        }
+        hasher.update(self.fee_rate.to_be_bytes());
+        // if let Some(fee_total) = self.last_fee_total {
+        //     hasher.update(fee_total.to_be_bytes());
+        // }
+        // if let Some(fee_rate) = self.last_fee_rate {
+        //     hasher.update(fee_rate.to_be_bytes());
+        // }
```
djordon, line 548: ?

signer/src/transaction_coordinator.rs
djordon: I would simplify it a bit for our initial implementation and just wait for the threshold number of Acks (with a timeout) and proceed with a signing round after the timeout.

We should design the Ack message though to be like you said, with a vector transactions IDs, sighashs, and a boolean for whether they will sign it. And I don't think that we need the transaction IDs so maybe we don't Ack with those.

djordon: But we don't need to do all of that in this PR so this is fine.

signer/src/transaction_signer.rs
```diff
@@ -309,7 +317,64 @@ where
...
+            .ok_or_else(|| Error::NoChainTip)?;
+
+        let (maybe_aggregate_key, _signer_set) = self
+            .get_signer_set_and_aggregate_key(bitcoin_chain_tip)
+            .await?;
+        let aggregate_key = maybe_aggregate_key.ok_or(Error::NoDkgShares)?;
+
+        let signer_state = self
+            .get_btc_state(bitcoin_chain_tip, &aggregate_key)
+            .await?;
```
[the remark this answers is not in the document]
djordon: No, each signer should create the BTC state with their view of the signers' UTXO but use the fees sent in the request. Later we can add some loose validation of the fees themselves but we should always use the ones sent by the coordinator.

signer/src/transaction_signer.rs
```diff
@@ -309,7 +317,64 @@ where
...
+
+        self.context
+            .get_storage_mut()
+            .write_bitcoin_withdrawals_outputs(withdrawals_outputs)
+            .await?;
+
+        Ok(())
+    }
+
+    #[tracing::instrument(skip(self, request))]
```
djordon, line 377: I'd `skip_all` here.

signer/src/transaction_signer.rs
```diff
@@ -309,7 +317,64 @@ where
         })
     }
 
-    #[tracing::instrument(skip_all)]
+    #[tracing::instrument(skip(self, request))]
```
djordon, line 320: I'd `skip_all` here too.

signer/tests/integration/transaction_signer.rs
```diff
@@ -508,3 +514,107 @@ async fn signing_set_validation_check_for_stacks_transactions() {
...
+    let signer_digest = digests.signer_sighash();
+    let deposit_digest = digests.deposit_sighashes();
+    assert_eq!(deposit_digest.len(), 1);
+    let deposit_digest = deposit_digest[0];
+
+    // Check that the sighashes are stored in the database
+    assert!(db
+        .will_sign_bitcoin_tx_sighash(&signer_digest.sighash.into())
+        .await
+        .is_ok());
```
djordon, line 614: I don't think that this checks that the sighashes are stored in the database, just that the query returned successfully.

signer/tests/integration/setup.rs
```diff
@@ -400,6 +401,108 @@ pub async fn backfill_bitcoin_blocks(db: &PgStore, rpc: &Client, chain_tip: &bit
     }
 }
 
+pub async fn fill_signers_utxo<R: rand::RngCore + ?Sized>(
+    db: &PgStore,
+    bitcoin_block: model::BitcoinBlock,
+    aggregate_key: &PublicKey,
+    mut rng: &mut R,
+) {
```
djordon, line 409: Hmmm, a similar method should be added to the `TestSweepSetup` struct, since that generates a UTXO for the signers to use. But we don't have methods for writing to all of the necessary tables there yet.

djordon: Well after the base branch gets merged in we can use the new test setup struct.

signer/src/transaction_signer.rs
```diff
@@ -740,6 +849,21 @@ where
...
+    // -------------- \\
+
+    /// Public testing handler for [`Self::handle_sbtc_requests_context_message`]
+    #[cfg(feature = "testing")]
+    pub async fn handle_sbtc_requests_context_message_tester(
+        &mut self,
+        request: &message::SbtcRequestsContextMessage,
+        bitcoin_chain_tip: &model::BitcoinBlockHash,
+    ) -> Result<(), Error> {
+        self.handle_sbtc_requests_context_message(request, bitcoin_chain_tip)
```
djordon, line 864: Let's just make `handle_sbtc_requests_context_message` public and use that.

signer/src/message.rs
djordon: I think we need to hash them still though. It's a shame since we'll be deleting it soonish.

djordon (commented):
This looks good to me. Let's hold off on merging this until the base branch, `794-save-bitcoin-sighashes-to-db` gets merged.

djordon:
This closes https://github.com/stacks-network/sbtc/issues/913 as well.

signer/src/transaction_coordinator.rs
```diff
@@ -418,14 +419,29 @@ where
...
+                .collect(),
+            fee_rate: pending_requests.signer_state.fee_rate,
+            last_fees: pending_requests.signer_state.last_fees.map(Into::into),
+        };
+
+        // Share the list of requests with the signers.
+        self.send_message(sbtc_requests, bitcoin_chain_tip).await?;
+        // Wait to reduce chance that the other signers will receive the subsequent
+        // messages before the BitcoinPreSignRequest one.
+        tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
```
djordon, line 444: Given the variability of our network, we're probably going to want to make this a configurable parameter with a default on testnet/mainnet that is ~5-10 seconds or so. For devnet and local testing we can keep this at 1 second.

djordon (approved):
Looks good.

We're going to need to fiddle with the timeout for the pre-sign message but we should do that in a separate PR.

signer/src/transaction_coordinator.rs
djordon: Well, I'm not sure if we want to do the intermediate work and make this configurable or go straight for having the coordinator wait for `Ack`s. Your call.

Jiloc: done in https://github.com/stacks-network/sbtc/pull/905/commits/468153f8dcb67255e836ccdfe65d4e449fd63993 . 
I should have done it in a separate branch, sorry, my mistake!

signer/src/transaction_coordinator.rs
```diff
@@ -387,6 +392,63 @@ where
...
+                // so we trigger a shutdown.
+                let Some(_) = signal_stream.next().await else {
+                    tracing::warn!("signal stream returned None, shutting down");
+                    self.context.get_termination_handle().signal_shutdown();
+                    return Err(Error::SignerShutdown);
+                };
+                num_acks += 1;
+            }
+            Ok(())
+        };
```
Jiloc, line 444: At the moment they are simple ACKs. But, they can be easily extended to return an ordered vector of bool representing the will to vote per each request, in the same order as they were received.

signer/src/transaction_coordinator.rs
```diff
@@ -427,6 +489,13 @@ where
         // Construct the transaction package and store it in the database.
         let transaction_package = pending_requests.construct_transactions()?;
 
+        self.construct_and_send_bitcoin_presign_request(
+            bitcoin_chain_tip,
+            &pending_requests,
+            &transaction_package,
+        )
+        .await?;
```
Jiloc, line 497: The coordinator will return without sending the signing request if we don't receive enough Acks from the signers in the expected timeout

PR #906, opened 2024-11-20
fix: remove the foreign key constraint on rotate-keys event table

djordon:
## Description

The other event tables do not have such a constraint, because we cannot know whether the transaction has been written to the table before the event.

## Changes

* Remove the foreign key constraint

## Checklist:


Jiloc (approved):
LGTM

PR #907, opened 2024-11-20
fix: handle networking bug

djordon:
## Description

This might address a bug we're seeing on devnet.

I'm not entirely sure this fixes the issue, or if the issue was here at all. But this adds more logging and probably makes things better.

## Changes

* Bump the stream capacity
* More eagerly send messages along.

## Checklist:

### DIFF
--- a/signer/src/context/mod.rs
+++ b/signer/src/context/mod.rs
@@ -12,7 +12,6 @@ use crate::bitcoin::BitcoinInteract;
 use crate::config::Settings;
 use crate::emily_client::EmilyInteract;
 use crate::error::Error;
-use crate::network::MessageTransfer;
 use crate::stacks::api::StacksInteract;
 use crate::storage::DbRead;
 use crate::storage::DbWrite;
@@ -59,16 +58,14 @@ pub trait Context: Clone + Sync + Send {
     /// later return `Some(_)`. But if [`StreamExt::next`] yields `None`
     /// three times then the stream is "fused" and will return `None`
     /// forever after.
-    fn as_signal_stream<F, M>(&self, network: &M, predicate: F) -> ReceiverStream<SignerSignal>
+    fn as_signal_stream<F>(&self, predicate: F) -> ReceiverStream<SignerSignal>
     where
-        M: MessageTransfer,
         F: Fn(&SignerSignal) -> bool + Send + Sync + 'static,
     {
         let (sender, receiver) = tokio::sync::mpsc::channel(SIGNER_CHANNEL_CAPACITY);
 
         let mut watch_receiver = self.get_termination_handle();
         let mut signal_stream = self.get_signal_receiver();
-        let mut network_stream = network.as_receiver();
 
         tokio::spawn(async move {
             loop {
@@ -98,25 +95,6 @@ pub trait Context: Clone + Sync + Send {
                             Err(error @ RecvError::Lagged(_)) => {
                                 tracing::warn!(%error, "internal signal stream lagging");
                                 continue
-                            }
-                        }
-                    }
-                    item = network_stream.recv() => {
-                      match item {
-                            Some(msg) => {
-                                let signal = P2PEvent::MessageReceived(msg).into();
-                                if predicate(&signal) {
-                                    // See comment above, we can bail.
-                                    if sender.send(signal).await.is_err() {
-                                        break;
-                                    }
-                                }
-                            }
-                            // This means the channel has been closed.
-                            // Since this is the network stream we can bail
-                            // here, we're probably on our way down.
-                            None => {
-                                break;
                             }
                         }
                     }
--- a/signer/src/context/mod.rs
+++ b/signer/src/context/mod.rs
@@ -66,30 +66,27 @@ pub trait Context: Clone + Sync + Send {
     {
         let (sender, receiver) = tokio::sync::mpsc::channel(SIGNER_CHANNEL_CAPACITY);
 
-        let mut term = self.get_termination_handle().as_receiver();
+        let mut watch_receiver = self.get_termination_handle();
         let mut signal_stream = self.get_signal_receiver();
         let mut network_stream = network.as_receiver();
+
         tokio::spawn(async move {
             loop {
                 tokio::select! {
-                    item = term.recv() => {
-                        match item {
-                            Some(signal) => {
-                                // An error means that the channel has been
-                                // closed. This is most likely due to the
-                                // receiver being closed so we can bail.
-                                if let Err(_) = sender.send(signal).await {
-                                    break;
-                                }
-                            }
-                            None => break,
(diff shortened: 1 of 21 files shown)


signer/src/network/libp2p/network.rs
```diff
@@ -104,11 +105,26 @@ impl MessageTransfer for P2PNetwork {
     }
 
     fn receiver_stream(&self) -> BroadcastStream<SignerSignal> {
-        let (sender, receiver) = tokio::sync::broadcast::channel(10);
+        // This is the same capacity of `signal_tx` in the SignerContext,
+        // which is typically used to create this P2PNetwork.
+        let (sender, receiver) = tokio::sync::broadcast::channel(SIGNER_CHANNEL_CAPACITY);
```
cylewitruk, line 110: I think that if you changed this to an unbounded mpsc stream that it would work. I think the problem is that the poller of the `BroadcastStream` returned below is still polling items one-at-a-time with (sometimes long) processing times in-between each item, which allows this channel to quickly fill up especially when coordinated rounds are flying around.

But that still leaves the potential problem of OOM if this channel grows too much since it accumulates _all_ P2P messages and not only the ones the consumer is interested in, which is why I'm not sure if trying to aggregate the "inbox" up to this abstraction vs. the actual consumer is the best solution since the consumer, knowing exactly what they're interested in, can more effectively manage the items that flow into their inbox vs. get dropped and also potentially take action if they see that the inbox is getting too big.

I hope I'm making sense? :sweat_smile:

djordon: Yeah you're making sense. 🤔

signer/src/context/mod.rs
```diff
@@ -58,14 +59,70 @@ pub trait Context: Clone + Sync + Send {
...
+                                tracing::warn!("internal signal stream closed");
+                                break;
+                            }
+                            Err(error @ RecvError::Lagged(_)) => {
+                                tracing::warn!(%error, "internal signal stream lagging");
+                                continue
+                            }
+                        }
+                    }
+                    item = network_stream.recv() => {
```
cylewitruk, line 107: Hm, is this section needed right now due to the fact that tests use their special network thing right now?

signer/src/network/libp2p/network.rs
```diff
@@ -103,37 +103,54 @@ impl MessageTransfer for P2PNetwork {
         }
     }
 
-    fn receiver_stream(&self) -> BroadcastStream<SignerSignal> {
-        let (sender, receiver) = tokio::sync::broadcast::channel(10);
+    fn as_receiver(&self) -> tokio::sync::mpsc::Receiver<Msg> {
+        // This is the same capacity of `signal_tx` in the SignerContext,
+        // which is typically used to create this P2PNetwork.
+        let (sender, receiver) = tokio::sync::mpsc::channel(SIGNER_CHANNEL_CAPACITY);
```
cylewitruk, line 109: Let's see how this does

signer/src/context/mod.rs
djordon: Oh yeah I think you're right. Libp2p sends messages on the signal stream, and so do our tests.

signer/src/network/mod.rs
```diff
@@ -39,7 +38,7 @@ pub trait MessageTransfer: Clone {
     /// Return a stream of the same messages from the
     /// [`MessageTransfer::receive`]. The inner type of each
     /// [`SignerSignal`] should be a [`P2PEvent::MessageReceived`].
-    fn receiver_stream(&self) -> BroadcastStream<SignerSignal>;
```
cylewitruk, line 42: Much happier that this is `Msg` now :pray:

signer/src/context/mod.rs
cylewitruk: I think the crux is rather that the in-memory networks _don't_ publish `P2PMessageReceived` etc :grimacing:

djordon: I think so. This catch might be the source of all issues. Since we're going through this stream now, I don't think that we'll have this issue anymore. But I'm only 30% sure that I've thought this through.

signer/src/transaction_coordinator.rs
```diff
@@ -170,13 +191,13 @@ where
...
-        let mut signal_stream = self.context.as_signal_stream(&self.network);
+        let mut signal_stream = self.context.as_signal_stream(run_loop_message_filter);
 
         loop {
             match signal_stream.next().await {
-                Some(Ok(SignerSignal::Command(SignerCommand::Shutdown))) => break,
-                Some(Ok(SignerSignal::Command(SignerCommand::P2PPublish(_)))) => {}
-                Some(Ok(SignerSignal::Event(event))) => {
+                Some(SignerSignal::Command(SignerCommand::Shutdown)) => break,
+                Some(SignerSignal::Command(SignerCommand::P2PPublish(_))) => {}
```
matteojug, line 199: nit: this case should be impossible given the filter above, should we add a catchall instead?

signer/src/transaction_coordinator.rs
```diff
@@ -170,13 +191,13 @@ where
     #[tracing::instrument(skip_all, name = "tx-coordinator")]
     pub async fn run(mut self) -> Result<(), Error> {
         tracing::info!("starting transaction coordinator event loop");
-        let mut signal_stream = self.context.as_signal_stream(&self.network);
+        let mut signal_stream = self.context.as_signal_stream(run_loop_message_filter);
 
         loop {
             match signal_stream.next().await {
```
matteojug, line 197: nit: `while let Some(message) = signal_stream.next().await {` like in the other run loop? also in `signer/src/transaction_signer.rs`

signer/src/transaction_coordinator.rs
djordon: yeah, the main difference is that you can see which commands you're skipping here at first glance, and know that you aren't skipping something important. Then again, we have a catch all for the events. If we had only three events I would replace the catchall with the explicit variant.

signer/src/transaction_coordinator.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/907/commits/5ecf544445e816671902ebb68dcda3fab7ba2545.

signer/src/network/mod.rs
djordon: Yeah I was very happy. But now https://github.com/stacks-network/sbtc/pull/907/commits/b1d0873c4c169a8b84e4580b223b96716f6a10c4 removed all of that nonsense.

cylewitruk (approved):
There's definitely some cleanup to do here (mostly mine), but given that this branch has been proven in the initial testnet setup, I rather we isolate further changes to iterative changes in future PRs, so approving this :+1:

Issue #897, opened 2024-11-19
[Feature]: Vary Emily API access across `Signer`, `Admin`, and `Public` endpoints


AshtonStephens:
# Feature - Vary Emily API access across `Signer`, `Admin`, and `Public` endpoints

## 1. Description

The Emily API needs to have three levels of access for different api endpoints.
- `signer` level access to update the state of the blockchain
- `admin` level access to update the caps on the sbtc protocol amongst anything else necessary.
- `public` level access that has access to all the public facing api calls

### 1.1 Context & Purpose

Born from #896.

The Emily API now has more than two levels of access across the api endpoints. Originally there were the signers that maintained the state and therefore had write access to state maintaining api calls and then public endpoints that anyone could modify. #896 introduces an API endpoint with set access that the signers should *not* be able to modify.

In that PR we give the signers access to those endpoints via the openapi definition and api keys, but we don't want to do that.

This ticket is to separate the permissions for the various calls such that only users with the right permissions can access the different API calls.

## 2. Technical Details:

1. The openapi spec should be updated so that only the accessible endpoints show up within the client compiled for a specific version of the spec
2. The cdk should provide the api with different keys depending on the endpoint type.

### 2.1 Acceptance Criteria:
   - [ ] There are multiple openapi specs and clients with the varying versions of the api across access 
   - [ ] The cdk deploys the different apis and provides them different api keys
   - [ ] The PR for this features is created and merged.

## 3. Related Issues and Pull Requests (optional):
- Issues: 
    - #855
    - #857 
- PR that lead to this feature: #896

PR #909, opened 2024-11-20
IGNORE - 897's Autogenerated sources (separated)

AshtonStephens:
## Description

Closes: #897 

This is Just the autogenerated sources modifications.


AshtonStephens:
Closing because I need to modify this so that the public and signer endpoints are the same, and it makes it look like a ton of things are in review.

PR #920, opened 2024-11-22
fix: demo branch fixes

matteojug:
## Description

Closes: #?

## Checklist:

### DIFF
--- a/contracts/contracts/sbtc-registry.clar
+++ b/contracts/contracts/sbtc-registry.clar
@@ -266,8 +266,6 @@
     (try! (is-protocol-caller))
     ;; Check that the aggregate pubkey is not already in the map
     (asserts! (map-insert aggregate-pubkeys new-aggregate-pubkey true) ERR_AGG_PUBKEY_REPLAY)
-    ;; Check that the new address (multi-sig) is not already in the map
-    (asserts! (map-insert multi-sig-address new-address true) ERR_MULTI_SIG_REPLAY)
     ;; Update the current signer set
     (var-set current-signer-set new-keys)
     ;; Update the current multi-sig address
--- a/contracts/contracts/sbtc-registry.clar
+++ b/contracts/contracts/sbtc-registry.clar
@@ -56,10 +56,6 @@
 ;; stored to avoid replay
 (define-map aggregate-pubkeys (buff 33) bool)
 
-;; Data structure to store the current signer set,
-;; stored to avoid replay
-(define-map multi-sig-address principal bool)
-
 ;; Data structure to store the active protocol contracts
 (define-map protocol-contracts principal bool)
 (map-set protocol-contracts .sbtc-bootstrap-signers true)
--- a/signer/Cargo.toml
+++ b/signer/Cargo.toml
@@ -48,7 +48,7 @@ tokio = { workspace = true, features = ["signal", "macros", "rt-multi-thread"] }
 tokio-stream.workspace = true
 tracing.workspace = true
 tracing-attributes.workspace = true
-tracing-subscriber = { workspace = true, features = ["json", "fmt"]}
+tracing-subscriber = { workspace = true }
 url.workspace = true
 # wsts.workspace = true
 wsts = { git = "https://github.com/Trust-Machines/wsts.git", rev = "ebd7d7775ad5e44cdbf4f5c1fb468bdf6c467265" }
--- a/signer/Cargo.toml
+++ b/signer/Cargo.toml
@@ -48,7 +48,7 @@ tokio = { workspace = true, features = ["signal", "macros", "rt-multi-thread"] }
 tokio-stream.workspace = true
 tracing.workspace = true
 tracing-attributes.workspace = true
-tracing-subscriber.workspace = true
+tracing-subscriber = { workspace = true, features = ["json", "fmt"]}
 url.workspace = true
 # wsts.workspace = true
 wsts = { git = "https://github.com/Trust-Machines/wsts.git", rev = "ebd7d7775ad5e44cdbf4f5c1fb468bdf6c467265" }
--- a/signer/src/emily_client.rs
+++ b/signer/src/emily_client.rs
@@ -323,6 +323,7 @@ impl EmilyInteract for EmilyClient {
     async fn set_chainstate(&self, chainstate: Chainstate) -> Result<Chainstate, Error> {
         chainstate_api::set_chainstate(&self.config, chainstate)
             .await
+            .inspect_err(|e| tracing::info!(?e, "Error for set_chainstate"))
             .map_err(EmilyClientError::AddChainstateEntry)
             .map_err(Error::EmilyApi)
     }
--- a/signer/src/main.rs
+++ b/signer/src/main.rs
@@ -279,7 +279,7 @@ async fn run_transaction_signer(ctx: impl Context) -> Result<(), Error> {
         network,
         context: ctx.clone(),
         context_window: 10000,
-        threshold: 2,
+        threshold: config.signer.bootstrap_signatures_required.into(),
         rng: rand::thread_rng(),
         signer_private_key: config.signer.private_key,
         wsts_state_machines: HashMap::new(),
@@ -290,7 +290,8 @@ async fn run_transaction_signer(ctx: impl Context) -> Result<(), Error> {
 
 /// Run the transaction coordinator event-loop.
 async fn run_transaction_coordinator(ctx: impl Context) -> Result<(), Error> {
-    let private_key = ctx.config().signer.private_key;
+    let config= ctx.config().clone();
+    let private_key = config.signer.private_key;
     let network = P2PNetwork::new(&ctx);
 
     let coord = transaction_coordinator::TxCoordinatorEventLoop {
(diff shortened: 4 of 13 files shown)


signer/src/main.rs
```diff
@@ -41,22 +48,30 @@ struct SignerArgs {
...
     // Parse the command line arguments.
     let args = SignerArgs::parse();
 
+    // Initialize logging
+    let tracing_filter = tracing_subscriber::EnvFilter::from_default_env();
+    if let Some(LogOutputFormat::Json) = args.output_format {
+        tracing_subscriber::fmt()
+            .json()
+            .with_env_filter(tracing_filter)
+            .init();
```
djordon, line 68: There is also `signer::logging::setup_logging(_, false)`, which is what I typically use for JSON logging

signer/src/network/libp2p/event_loop.rs
```diff
@@ -239,19 +302,18 @@ fn handle_mdns_event(swarm: &mut Swarm<SignerBehavior>, ctx: &impl Context, even
...
-                tracing::debug!(%peer_id, "ignoring identify message from unknown peer");
-                return;
-            }
-            tracing::debug!(%peer_id, "Received identify message from peer; adding to confirmed external addresses");
-            swarm.add_external_address(info.observed_addr.clone());
+            tracing::debug!(%peer_id, ?info, "Received identify message from peer");
+            swarm
+                .behaviour_mut()
+                .kademlia
+                .add_address(&peer_id, strip_peer_id(&info.observed_addr));
```
djordon, line 316: Why not verify the peer?

signer/src/transaction_signer.rs
```diff
@@ -182,6 +183,16 @@ where
 
     #[tracing::instrument(skip_all, fields(chain_tip = tracing::field::Empty))]
     async fn handle_signer_message(&mut self, msg: &network::Msg) -> Result<(), Error> {
+        if matches!(
+            msg.payload,
+            message::Payload::SignerDepositDecision(_)
+                | message::Payload::SignerWithdrawalDecision(_)
+                | message::Payload::StacksTransactionSignature(_)
+                | message::Payload::BitcoinTransactionSignAck(_)
+        ) {
```
djordon, line 192: This one isn't so bad, but I find this style harder to read than the two step approach:
```rust
// The tx-signer does not handle these message types at all, the first two are
// handled by the request-decider (we should change to tx-checker or something)
// and the last two by the tx-coordinator.
let is_ignored_message = matches!(
    msg.payload,
    message::Payload::SignerDepositDecision(_)
    | message::Payload::SignerWithdrawalDecision(_)
    | message::Payload::StacksTransactionSignature(_)
    | message::Payload::BitcoinTransactionSignAck(_)
);

if is_ignored_message {
    return Ok(());
}
```

signer/src/transaction_signer.rs
```diff
@@ -501,6 +519,10 @@ where
                     .await?;
             }
             WstsNetMessage::DkgPublicShares(dkg_public_shares) => {
+                tracing::info!(
+                    "handling DkgPublicShares from signer {}",
+                    dkg_public_shares.signer_id
+                );
```
djordon, line 525: For these logs we should probably do:
```rust
                tracing::info!(
                    signer_id = %dkg_public_shares.signer_id,
                    "handling DkgPublicShares",
                );
```

signer/src/transaction_signer.rs
```diff
@@ -488,10 +499,17 @@ where
                     self.signer_private_key,
                 )?;
                 self.wsts_state_machines.insert(msg.txid, state_machine);
+                // Let's give the others some slack
+                tracing::debug!(
+                    "Sleeping a bit to give the other peers some slack to get DkgBegin"
+                );
+                tokio::time::sleep(Duration::from_secs(3)).await;
+
```
djordon, line 507: Oh this is a band-aid until we revamp DKG?

signer/Cargo.toml
```diff
@@ -48,7 +48,7 @@ tokio = { workspace = true, features = ["signal", "macros", "rt-multi-thread"] }
 tokio-stream.workspace = true
 tracing.workspace = true
 tracing-attributes.workspace = true
-tracing-subscriber.workspace = true
+tracing-subscriber = { workspace = true, features = ["json", "fmt"]}
```
djordon, line 51: These features, and more, are already pulled in.

signer/src/main.rs
```diff
@@ -41,22 +48,30 @@ struct SignerArgs {
     /// pending migrations to the database on startup.
     #[clap(long)]
     migrate_db: bool,
+
+    #[clap(short = 'o', long = "output-format", default_value = "pretty")]
+    output_format: Option<LogOutputFormat>,
```
djordon, line 53: It feels more natural to make this a config option. Same thing for the migrate db field too, although that one probably shouldn't be an option at all.

signer/src/main.rs
```diff
@@ -290,17 +305,18 @@ async fn run_transaction_signer(ctx: impl Context) -> Result<(), Error> {
...
         network,
         context: ctx,
         context_window: 10000,
         private_key,
-        signing_round_max_duration: Duration::from_secs(15),
-        threshold: 2,
-        dkg_max_duration: Duration::from_secs(15),
+        signing_round_max_duration: Duration::from_secs(30),
+        threshold: config.signer.bootstrap_signatures_required,
+        dkg_max_duration: Duration::from_secs(120),
```
djordon, line 319: We should open an issue to make these config options... maybe?

djordon (commented):
Had a few questions and comments.

signer/src/transaction_signer.rs
matteojug: Correct -- we bumped this to 10s in the current deployment and everyone got their DKG shares

signer/src/network/libp2p/event_loop.rs
djordon: Oh I see. Let's add a comment here to that effect.

signer/Cargo.toml
cylewitruk: [ca98a84](https://github.com/stacks-network/sbtc/pull/920/commits/ca98a8401b5f2522f8b3120120ddabd6c16e63f8)

contracts/contracts/sbtc-registry.clar
djordon, line 1: Let's remove these changes until after we see more movement on whether this is a good idea in https://github.com/stacks-network/sbtc/issues/928. I think it is, but want to hear more thoughts on this.

signer/src/main.rs
cylewitruk: Hm, I think I disagree, I often like to quickly change what I'm outputting depending on how I'm running something, and this is a lot easier than messing with long env vars... 
If we were writing to file then yeah I would agree as that's just in general more static, but here we're writing to stdout/err and for me feels more natural to be able to steer that in the same place as it's being executed.

signer/src/main.rs
cylewitruk: Yeah probably :+1:

signer/src/transaction_signer.rs
```diff
@@ -488,10 +498,17 @@ where
                     self.signer_private_key,
                 )?;
                 self.wsts_state_machines.insert(msg.txid, state_machine);
+                // Let's give the others some slack
+                tracing::debug!(
+                    "Sleeping a bit to give the other peers some slack to get DkgBegin"
+                );
+                tokio::time::sleep(Duration::from_secs(10)).await;
```
djordon, line 505: This may make some of our tests fail, maybe make this configurable or something?

signer/src/transaction_signer.rs
cylewitruk: [7d7c91b](https://github.com/stacks-network/sbtc/pull/920/commits/7d7c91b65e3aed2ee28f0f1bf2b9a5a7fd106bce)

contracts/contracts/sbtc-registry.clar
cylewitruk: Based on discussion in standup I think we agreed to merge this and repurpose #928 rather towards reverting this change in-case we decide it needs to be there for some reason.

signer/src/main.rs
cylewitruk: [8f788e8](https://github.com/stacks-network/sbtc/pull/920/commits/8f788e8307b5a98edd2082386bfd8d0123bcf40a)

I liked that json better, but the pretty was _too_ pretty :laughing:

contracts/contracts/sbtc-registry.clar
djordon: Yup.

signer/src/main.rs
djordon: 🙏🏽

signer/src/main.rs
djordon: Okay sounds good. Might be worth opening an issue for though... maybe

signer/src/network/libp2p/event_loop.rs
djordon: Spoke about this offline and we should add module documentation for all of the things that we've learned about our libp2p setup.

signer/src/transaction_signer.rs
cylewitruk: [147f3f3](https://github.com/stacks-network/sbtc/pull/920/commits/147f3f3c3816765267fcb735c1aeda1c5769228e)

Updated #947 as well in-case this ends up lingering around, but added a comment that it may become deprecated with upcoming dkg changes.

djordon (approved):
Looks good!

signer/src/transaction_signer.rs
djordon: 🙏🏽

signer/src/emily_client.rs
```diff
@@ -323,6 +323,7 @@ impl EmilyInteract for EmilyClient {
     async fn set_chainstate(&self, chainstate: Chainstate) -> Result<Chainstate, Error> {
         chainstate_api::set_chainstate(&self.config, chainstate)
             .await
+            .inspect_err(|e| tracing::info!(?e, "Error for set_chainstate"))
```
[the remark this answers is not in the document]
djordon: Whatever, let's just merge and fix later.

PR #927, opened 2024-11-22
Embed public and private shares into DkgBegin messages

xoloki:
## Description

This change fixes the race condition that arises during `DKG` from the interplay of our networking stack, usage of `WSTS`, and `WSTS` itself.   It uses recent `WSTS` changes to embed public and private shares into the various `DkgBegin` messages.  This avoids the case where a signer receives `DKG` public or private shares before the corresponding `DkgBegin`.

Closes: #929 

## Checklist:

### DIFF
--- a/signer/src/proto/convert.rs
+++ b/signer/src/proto/convert.rs
@@ -665,20 +665,21 @@ impl From<DkgPrivateBegin> for proto::DkgPrivateBegin {
     }
 }
 
-impl From<proto::DkgPrivateBegin> for DkgPrivateBegin {
-    fn from(value: proto::DkgPrivateBegin) -> Self {
+impl TryFrom<proto::DkgPrivateBegin> for DkgPrivateBegin {
+    type Error = Error;
+    fn try_from(value: proto::DkgPrivateBegin) -> Result<Self, Self::Error> {
         let mut dkg_public_shares = HashMap::new();
         if let Some(shares) = value.dkg_public_shares {
             for (id, share) in shares.shares {
-                dkg_public_shares.insert(id, DkgPublicShares::try_from(share).unwrap());
+                dkg_public_shares.insert(id, DkgPublicShares::try_from(share)?);
             }
         }
-        DkgPrivateBegin {
+        Ok(DkgPrivateBegin {
             dkg_id: value.dkg_id,
             signer_ids: value.signer_ids,
             key_ids: value.key_ids,
             dkg_public_shares,
-        }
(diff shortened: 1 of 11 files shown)


xoloki:
@matteojug can you point me to the added `sleep` statements?  I want to try disabling them.

djordon:
I think this is the one: https://github.com/stacks-network/sbtc/pull/927/files#diff-2d16f548dd0cc67e9c08474479e9c73ff36c6004bafe17d3ab0af140d14d5954R557-R562

xoloki:
> I think this is the one: https://github.com/stacks-network/sbtc/pull/927/files#diff-2d16f548dd0cc67e9c08474479e9c73ff36c6004bafe17d3ab0af140d14d5954R557-R562

Ok, the sleep is disabled in `main.rs` but the code is left intact.  I ran a devnet demo with no problem, please try to reproduce the error on this branch @matteojug @cylewitruk @djordon

signer/src/proto/convert.rs
```diff
@@ -649,20 +650,34 @@ impl From<proto::DkgBegin> for DkgBegin {
...
         }
     }
 }
 
 impl From<proto::DkgPrivateBegin> for DkgPrivateBegin {
     fn from(value: proto::DkgPrivateBegin) -> Self {
+        let mut dkg_public_shares = HashMap::new();
+        if let Some(shares) = value.dkg_public_shares {
+            for (id, share) in shares.shares {
+                dkg_public_shares.insert(id, DkgPublicShares::try_from(share).unwrap());
```
xoloki, line 673: @djordon I had to unwrap here, since you can't return an error from `impl From`.  We should probably change this and the others to `impl TryFrom`, and deal with the fallout.

xoloki: Ok, `unwrap` calls are now gone

djordon:
@xoloki I think it's safe to close this for now right? I'm hoping that we don't need to touch DKG code for some time, and that if when we do revisit it, we can remove the coordinator altogether by doing something like https://github.com/stacks-network/sbtc/issues/1286.

xoloki:
> @xoloki I think it's safe to close this for now right? I'm hoping that we don't need to touch DKG code for some time, and that if when we do revisit it, we can remove the coordinator altogether by doing something like #1286.

Yeah I'll close this now.  I'll leave the `wsts` PR open since Jacinta approved it, but I'll hold off on merging it until I have a reason.

Issue #932, opened 2024-11-24
[Bug]: Deposit max fee must not exceed amount


djordon:
# Bug - Deposit max fee must not exceed amount

## 1. Description

For deposits, the signers can construct an "invalid" bitcoin transaction because the users max fee may exceed the deposit amount.

### 1.1 Context & Purpose

We neither check whether the max-fee is less than the amount during deposit validation nor do we filter on the minimum of the max-fee and the amount. But we explicitly check this during validation of bitcoin and stacks transactions.

Without fixing this someone can make a deposit where the coordinator would attempt to create a bitcoin transaction where the assessed fee is greater than the amount. This would invalidate the transaction in the eyes of the other signers, affecting all requests serviced by that transaction too. The next coordinator would try to do the same thing too.

The path advocated for here is to consider the "real max-fee" to be the minimum of the specified max-fee and deposit amount and filter on their "real max-fee".

## 2. Technical Details:

Change the filtering in the `utxo` module to filter on the minimum of the max-fee and the amount for deposits.

### 2.1 Acceptance Criteria:
   - [ ] Signers don't construct transactions that will fail validation because of fees.

PR #933, opened 2024-11-24
fix: deposit max-fees must not exceed the deposit amount

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/932

## Changes

* Make sure we filter on the "real max-fee", which is the minimum of their max-fee and the deposit amount.
* Update the `SOLO_DEPOSIT_TX_VSIZE` to be based off of the most expensive recipient. This ensures that filtering based off of fees for deposits are conservative.

## Testing Information

This still needs a test or two.

## Checklist:


signer/src/bitcoin/utxo.rs
```diff
@@ -54,10 +54,11 @@ const DEFAULT_INCREMENTAL_RELAY_FEE_RATE: f64 =
...
 /// This constant represents the virtual size (in vBytes) of a BTC
-/// transaction that includes two inputs and one output. The inputs
-/// consist of the signers' input UTXO and a UTXO for a deposit request.
-/// The output is the signers' new UTXO.
-const SOLO_DEPOSIT_TX_VSIZE: f64 = 234.0;
+/// transaction that includes two inputs and one output. The inputs consist
+/// of the signers' input UTXO and a UTXO for a deposit request. The output
+/// is the signers' new UTXO. The deposit request is such that the sweep
+/// transaction has the largest size of solo deposit sweep transactions.
+const SOLO_DEPOSIT_TX_VSIZE: f64 = 267.0;
```
Jiloc, line 61: sanity check and...personal curiosity: how is this value computed? Just asking to better understand the context

Jiloc (commented):
Just a personal question. The implementation looks good

signer/src/bitcoin/utxo.rs
djordon: We have a "test", `create_deposit_only_tx`, that creates a sweep transaction moving only one deposit and zero withdrawals that outputs the size of the sweep transaction after it has been signed. This gives us the maximum size of a sweep transaction with only one deposit. Adding additional deposits decreases the costs, and adding additional withdrawals used to decrease the cost as well. 🤔 But now maybe not, I need to check again.

Jiloc (approved):

signer/src/bitcoin/utxo.rs
djordon: I just checked and costs decrease with an added withdrawal.

Issue #946, opened 2024-11-25
Implement Acks for BitcoinPreSignRequest


Jiloc:
Currently, the coordinator waits for 1 second after sending a BitcoinPreSignRequest. This seems sufficient for the development environment, but not for testnet or mainnet.

The proposed approach is to implement an Ack process between the coordinator and the signers. The coordinator would wait for at least the minimum required number of signers to send an ack back before proceeding, or wait for up to 10 seconds.

_Originally posted in https://github.com/stacks-network/sbtc/pull/905#discussion_r1855837508_

PR #951, opened 2024-11-25
wait for signers acks after sending bitcoin pre request

Jiloc:
## Description

Closes: #946 

## Changes
- added a new `TxSignerEvent::BitcoinPreSignRequestReceived` event type.
- added a new `bitcoin_presign_request_max_duration` field to `TxCoordinatorEventLoop`.
-  moved the sending of` BitcoinPreSignRequest` to `construct_and_send_bitcoin_presign_request` and switched from a simple timeout to waiting for Acks immediately after.


signer/src/transaction_coordinator.rs
```diff
@@ -388,6 +392,68 @@ where
...
+                // so we trigger a shutdown.
+                let Some(_) = signal_stream.next().await else {
+                    tracing::warn!("signal stream returned None, shutting down");
+                    self.context.get_termination_handle().signal_shutdown();
+                    return Err(Error::SignerShutdown);
+                };
+                num_acks += 1;
+            }
+            Ok(())
+        };
```
Jiloc, line 449: Currently, there is a simple ack counter, but this may be extended to have an ordered vector of bool indicating the will to sign the corresponding sighash.

PR #956, opened 2024-11-25
feat: wait for acks from signers after BitcoinPreSignRequest

Jiloc:
## Description

Closes: #946 

## Changes
- added a new `TxSignerEvent::BitcoinPreSignRequestReceived` event type.
- added a new `bitcoin_presign_request_max_duration` field to `TxCoordinatorEventLoop`.
-  moved the sending of` BitcoinPreSignRequest` to `construct_and_send_bitcoin_presign_request` and switched from a simple timeout to waiting for Acks immediately after.


signer/src/transaction_signer.rs
```diff
@@ -277,6 +277,8 @@ where
             (message::Payload::BitcoinPreSignRequest(requests), _, _) => {
                 self.handle_bitcoin_pre_sign_request(requests, &msg.bitcoin_chain_tip)
                     .await?;
+                self.context
+                    .signal(TxSignerEvent::BitcoinPreSignRequestReceived.into())?;
```
djordon, line 281: If we send a pre-sign ack message then we should remove this.

signer/src/transaction_coordinator.rs
```diff
@@ -390,6 +392,68 @@ where
...
+        };
+
+        // Define the filter for presign request received events
+        let presign_ack_filter = |event: &SignerSignal| -> bool {
+            matches!(
+                event,
+                SignerSignal::Event(SignerEvent::TxSigner(
+                    TxSignerEvent::BitcoinPreSignRequestReceived
+                ))
+            )
```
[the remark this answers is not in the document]
Jiloc: changed in https://github.com/stacks-network/sbtc/pull/956/commits/eec1c22ee13bac3395a6568823e6bc7f8cd8621f

signer/src/transaction_signer.rs
Jiloc: removed in https://github.com/stacks-network/sbtc/pull/956/commits/eec1c22ee13bac3395a6568823e6bc7f8cd8621f

Jiloc:
> As an aside we have pre-sign and sign messages and an Ack message already. We should consider renaming the the Ack one and removing the sign one.

Are you referring to [BitcoinTransactionSignRequest](https://github.com/stacks-network/sbtc/blob/main/protobufs/stacks/signer/v1/requests.proto#L10-L15) and [BitcoinTransactionSignAck](https://github.com/stacks-network/sbtc/blob/main/protobufs/stacks/signer/v1/decisions.proto#L36-L39) ? They look like they are used in some part of the wsts and in some tests, but seems like dead code. If that's the case, I can create a PR that clean up that part! Note that's a consensus breaking change, so hopeful we can pull it in asap

protobufs/stacks/signer/v1/messages.proto
```diff
@@ -33,6 +33,8 @@ message SignerMessage {
     SweepTransactionInfo sweep_transaction_info = 9;
     // Information about a new sweep transaction
     BitcoinPreSignRequest bitcoin_pre_sign_request = 10;
+    // Represents an acknowledgment of a BitcoinPreSignRequest
+    BitcoinPreSignAck bitcoin_pre_sign_ack = 11;
```
cylewitruk, line 37: note: protocol breaking

signer/src/message.rs
```diff
@@ -404,6 +417,7 @@ impl wsts::net::Signable for Payload {
             Self::StacksTransactionSignature(msg) => msg.hash(hasher),
             Self::SweepTransactionInfo(msg) => msg.hash(hasher),
             Self::BitcoinPreSignRequest(msg) => msg.hash(hasher),
+            Self::BitcoinPreSignAck(_) => hasher.update("SIGNER_BITCOIN_PRE_SIGN_ACK"),
```
cylewitruk, line 420: suggestion: just for consistency's sake I would probably have implemented this in the same way as the others, even if it's a little more code.

signer/src/transaction_coordinator.rs
```diff
@@ -390,6 +392,75 @@ where
...
+                .collect(),
+            fee_rate: pending_requests.signer_state.fee_rate,
+            last_fees: pending_requests.signer_state.last_fees.map(Into::into),
+        };
+        let presign_ack_filter = {
+            let target_tip = *bitcoin_chain_tip;
+            move |event: &SignerSignal| {
+                matches!(
+                    event,
+                    SignerSignal::Event(SignerEvent::TxSigner(TxSignerEvent::MessageGenerated(
```
cylewitruk, line 420: I think you still need to also include `P2PEvent::MessageReceived(Signed { etc. })`

signer/src/transaction_coordinator.rs
```diff
@@ -390,6 +392,75 @@ where
...
+        let future = async {
+            let mut num_acks = 1;
+            while num_acks < self.threshold {
+                // If signal_stream.next() returns None then one of the
+                // underlying streams has closed. That means either the
+                // network stream, the internal message stream, or the
+                // termination handler stream has closed. This is all bad,
+                // so we trigger a shutdown.
+                let Some(_) = signal_stream.next().await else {
+                    tracing::warn!("signal stream returned None, shutting down");
```
cylewitruk, line 449: I think it should suffice to just return an error here, the main loop will exit if it receives `None`.

cylewitruk:
# Protocol Breaking
The proto changes here are backwards compatible, but this PR changes the assumptions of the transaction coordinator, now expecting signers to respond to the Bitcoin pre-sign request with an ACK before continuing.

signer/src/transaction_coordinator.rs
```diff
@@ -390,6 +392,75 @@ where
...
+
+        // Create a signal stream with the defined filter
+        let signal_stream = self.context.as_signal_stream(presign_ack_filter);
+
+        // Send the presign request message
+        self.send_message(sbtc_requests, bitcoin_chain_tip).await?;
+
+        tokio::pin!(signal_stream);
+        let future = async {
+            let mut num_acks = 1;
```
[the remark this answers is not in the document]
cylewitruk: Yeah, we probably shouldn't make the assumption that our local signer will successfully participate.

signer/src/transaction_coordinator.rs
cylewitruk: hm, yeah that's a bit worrying... even counting from 1 I guess it should fail if the tests are testing with several signers and a threshold greater than 1.

matteojug: With 2 out of 3, considering it starts from 1 and the coordinator signer it should reach the required 2 acks I guess

cylewitruk: Yeah I guess so. I thought I've seen tests running like 5-7 though. But maybe not for the coordinator.

signer/src/transaction_coordinator.rs
```diff
@@ -390,6 +392,75 @@ where
...
+            while num_acks < self.threshold {
+                // If signal_stream.next() returns None then one of the
+                // underlying streams has closed. That means either the
+                // network stream, the internal message stream, or the
+                // termination handler stream has closed. This is all bad,
+                // so we trigger a shutdown.
+                let Some(_) = signal_stream.next().await else {
```
djordon, line 448: I think the signal stream always sends `Shutdown` messages, so we probably need to inspect them and exit early if so.

signer/src/transaction_coordinator.rs
```diff
@@ -390,6 +392,75 @@ where
...
+                    event,
+                    SignerSignal::Event(SignerEvent::TxSigner(TxSignerEvent::MessageGenerated(
+                        Signed {
+                            inner: SignerMessage {
+                                bitcoin_chain_tip,
+                                payload: Payload::BitcoinPreSignAck(_),
+                            },
+                            ..
+                        },
+                    ))) if *bitcoin_chain_tip == target_tip
```
djordon, line 428: I don't think that we should do the chain-tip filtering here, unless we also log that we've received a message where the chain tip is not what we are expecting. For example, in `drive_wsts_state_machine` we log that the coordinator observes "concurrent WSTS activity"

signer/src/main.rs
```diff
@@ -307,11 +307,11 @@ async fn run_transaction_coordinator(ctx: impl Context) -> Result<(), Error> {
         context_window: 10000,
         private_key,
         signing_round_max_duration: Duration::from_secs(30),
+        bitcoin_presign_request_max_duration: Duration::from_secs(30),
```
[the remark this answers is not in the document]
Jiloc: Even here, I was just matching the signing round duration. I believe we won't have a clear idea of the optimal timing until we test this on testnet

signer/src/message.rs
Jiloc: fixed in https://github.com/stacks-network/sbtc/pull/956/commits/d86363dbb1f0164a5944ab6ac71c399390691dcb

signer/src/transaction_coordinator.rs
Jiloc: fixed in https://github.com/stacks-network/sbtc/pull/956/commits/d86363dbb1f0164a5944ab6ac71c399390691dcb

signer/src/transaction_coordinator.rs
Jiloc: Do we also need to change the logic in `sign_stacks_transaction` if that's the case? https://github.com/stacks-network/sbtc/blob/main/signer/src/transaction_coordinator.rs#L730-L733

signer/src/transaction_coordinator.rs
Jiloc: I wrote it this way because that's how we handle this error in `sign_stacks_transaction`, but I can change it if it's sure we don't need it! https://github.com/stacks-network/sbtc/blob/main/signer/src/transaction_coordinator.rs#L730-L733

signer/src/transaction_coordinator.rs
```diff
@@ -390,6 +392,75 @@ where
...
+                // underlying streams has closed. That means either the
+                // network stream, the internal message stream, or the
+                // termination handler stream has closed. This is all bad,
+                // so we trigger a shutdown.
+                let Some(_) = signal_stream.next().await else {
+                    tracing::warn!("signal stream returned None, shutting down");
+                    self.context.get_termination_handle().signal_shutdown();
+                    return Err(Error::SignerShutdown);
+                };
+                num_acks += 1;
```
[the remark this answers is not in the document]
Jiloc: changed in https://github.com/stacks-network/sbtc/pull/956/commits/d86363dbb1f0164a5944ab6ac71c399390691dcb . I use a hashmap to keep track of the number of public_signer_keys from whom I received an ack

signer/src/transaction_coordinator.rs
```diff
@@ -390,6 +392,75 @@ where
...
+            }
+            Ok(())
+        };
+
+        // Wait for the future to complete with a timeout
+        tokio::time::timeout(self.signing_round_max_duration, future)
+            .await
+            .map_err(|_| Error::CoordinatorTimeout(self.signing_round_max_duration.as_secs()))?
```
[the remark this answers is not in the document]
Jiloc: fixed in https://github.com/stacks-network/sbtc/pull/956/commits/2807fc1e49aeb40088756f30a720f47a41982550

signer/src/transaction_signer.rs
```diff
@@ -277,6 +278,8 @@ where
             (message::Payload::BitcoinPreSignRequest(requests), _, _) => {
                 self.handle_bitcoin_pre_sign_request(requests, &msg.bitcoin_chain_tip)
                     .await?;
+                self.send_message(BitcoinPreSignAck, &msg.bitcoin_chain_tip)
+                    .await?;
```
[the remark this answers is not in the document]
Jiloc: changed in https://github.com/stacks-network/sbtc/pull/956/commits/d86363dbb1f0164a5944ab6ac71c399390691dcb

signer/src/transaction_coordinator.rs
Jiloc: The idea is to have an identified to be sure that the ack we received is connected to this request. We considered adding a unique identifier to the `BitcoinPreSignAck` to link it with the corresponding `BitcoinPreSignRequest`. But, we agreed that using the `bitcoin_chain_tip` as the key check is sufficient, since we are going to only send a single PreSignRequest per new block.

I’ll add a log for cases where the chain tip doesn’t match the expected value though. But if you feel that this doesn't make any sense I can get rid of that!

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
Jiloc: fixed in https://github.com/stacks-network/sbtc/pull/956/commits/d86363dbb1f0164a5944ab6ac71c399390691dcb

signer/src/proto/convert.rs
```diff
@@ -1372,6 +1373,19 @@ impl TryFrom<proto::BitcoinPreSignRequest> for BitcoinPreSignRequest {
...
+        proto::BitcoinPreSignAck {}
+    }
+}
+
+impl TryFrom<proto::BitcoinPreSignAck> for BitcoinPreSignAck {
+    type Error = Error;
+    fn try_from(_: proto::BitcoinPreSignAck) -> Result<Self, Self::Error> {
+        Ok(BitcoinPreSignAck {})
+    }
+}
```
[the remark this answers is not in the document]
Jiloc: fixed in https://github.com/stacks-network/sbtc/pull/956/commits/d86363dbb1f0164a5944ab6ac71c399390691dcb

signer/src/message.rs
```diff
@@ -374,6 +383,10 @@ pub struct BitcoinPreSignRequest {
     pub last_fees: Option<Fees>,
 }
 
+/// An acknowledgment of a [`BitcoinPreSignRequest`].
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct BitcoinPreSignAck;
+
```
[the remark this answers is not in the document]
Jiloc: For now I left it there because it won't compile unless #957 is merged, but good to know!

signer/src/transaction_coordinator.rs
djordon: Probably, we ignore those messages there too (the ignoring happens in the `filter_map` https://github.com/stacks-network/sbtc/blob/e647881c744be3e2034da7c69b45c33123ac516b/signer/src/transaction_coordinator.rs#L716-L719C31). Here we don't ignore them and increment the Ack counter. Well it used to so at least. The consequence is that shutdown could take longer so not filtering is fine.

signer/src/transaction_coordinator.rs
djordon: Yeah, that's the right idea. I was suggesting that we are just as noisy here as we are in other places where we assume the same thing. So yeah, definitely do the filtering, but log when the bitcoin chain tip is not what we expect.

Jiloc: implemented in https://github.com/stacks-network/sbtc/pull/956/commits/1818278009b40028f7518c3c446eb9dab7ec0118

signer/src/transaction_coordinator.rs
```diff
@@ -390,6 +392,95 @@ where
...
+            let mut acknowledged_signers = hashbrown::HashSet::new();
+
+            while acknowledged_signers.len() < self.threshold as usize {
+                match signal_stream.next().await {
+                    None => {
+                        tracing::warn!("signer signal stream closed unexpectedly, shutting down");
+                        return Err(Error::SignerShutdown);
+                    }
+                    Some(SignerSignal::Command(SignerCommand::Shutdown)) => {
+                        tracing::warn!("signer shutdown signal received, shutting down");
```
djordon, line 443: This should be an INFO message. It might be a graceful shutdown and all is well.

signer/src/transaction_coordinator.rs
```diff
@@ -390,6 +392,95 @@ where
...
+        // Create a signal stream with the defined filter
+        let signal_stream = self.context.as_signal_stream(presign_ack_filter);
+
+        // Send the presign request message
+        self.send_message(sbtc_requests, bitcoin_chain_tip).await?;
+
+        tokio::pin!(signal_stream);
+        let future = async {
+            let target_tip = *bitcoin_chain_tip;
+            let mut acknowledged_signers = hashbrown::HashSet::new();
```
djordon, line 434: We should use the standard library `HashSet`, I believe that a lot of the hashbrown implementation has already been ported over there. And I want to get WSTS to use the one in the standard library as well.

signer/src/transaction_coordinator.rs
```diff
@@ -390,6 +392,95 @@ where
...
 
+    /// Constructs a BitcoinPreSignRequest from the given transaction package and
+    /// sends it to the signers. Waits for acknowledgments from the signers until
+    /// the threshold is met or a timeout occurs.
+    /// If the signal stream closes unexpectedly, triggers a shutdown.
+    #[tracing::instrument(skip_all)]
+    async fn construct_and_send_bitcoin_presign_request(
+        &mut self,
+        bitcoin_chain_tip: &model::BitcoinBlockHash,
+        pending_requests: &SbtcRequests,
```
djordon, line 403: nit: we only need the `SignerBtcState` here. Also, it might be a little confusing using this type with the `sbtc_requests` variable below.

djordon (commented):
Looks good, I basically left a few nits.

signer/src/transaction_coordinator.rs
cylewitruk: Yeah, hashbrown is pretty much a part of stdlib now, so if we don't need hashbrown for specific reasons we should use std, agreed

signer/src/transaction_coordinator.rs
Jiloc: done in https://github.com/stacks-network/sbtc/pull/956/commits/c803d9d18bc48fc3aae449ca3c59233a26f44228

signer/src/transaction_coordinator.rs
Jiloc: done in https://github.com/stacks-network/sbtc/pull/956/commits/c803d9d18bc48fc3aae449ca3c59233a26f44228. Should we open a tech debt issue about removing all of our use of `hashbrown::HashSet`  and `HashMap`? Just to make it clear

signer/src/transaction_coordinator.rs
Jiloc: done in https://github.com/stacks-network/sbtc/pull/956/commits/c803d9d18bc48fc3aae449ca3c59233a26f44228

signer/src/transaction_coordinator.rs
Jiloc: Switched to just returning an error in https://github.com/stacks-network/sbtc/pull/956/commits/fd4873d66ef4e4e368ae594b19bfec7e5f1bac1a

signer/src/transaction_coordinator.rs
djordon: Maybe, but it's here because WSTS uses it. We cannot unilaterally remove it.

djordon (approved):
Approved!

PR #957, opened 2024-11-26
fix: add integrity protection for all messages

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/517.

## Changes

* Remove the digest field from the `StacksTransactionSignRequest`. It was always unnecessary.
* Sign messages through the protobuf encoding instead of through the `wsts::net::Signable` trait.
* Change `SignEcdsa::sign_ecdsa` to be infallible.
* Remove unnecessary `Serialize`/`Deserialize` implementations.

## Testing Information

This is a refactor, so we should be good here.

## Checklist:


signer/src/ecdsa.rs
```diff
@@ -57,21 +69,24 @@ pub struct Signed<T> {
...
     /// Verify the signature over the inner data.
     pub fn verify(&self) -> bool {
-        self.inner
-            .verify(&self.signature, &self.signer_pub_key.into())
+        let msg = secp256k1::Message::from_digest(self.inner.digest());
+        let Ok(sig) = secp256k1::ecdsa::Signature::from_compact(&self.signature) else {
+            return false;
+        };
+
+        self.signer_pub_key.verify(SECP256K1, &msg, &sig).is_ok()
```
matteojug, line 84: I still need to finish going through this, but I guess we may have an issue with this approach: if I'm reading https://protobuf.dev/programming-guides/encoding/#implications correctly, we don't have any guarantee that `self.inner.digest()` will result in the same bytes as what the original sender signed; other than the specs, looking into the specific lib we use they don't explicitly guarantee that either.

djordon: Right! I had forgotten about this. Okay, I'll revert a lot of these changes.

aldur:
As instructed, marked as draft to avoid merging it early.

signer/src/ecdsa.rs
```diff
@@ -23,31 +23,43 @@
...
 //!
 //! // Sign the message.
-//! let signed_msg = msg
-//!     .sign_ecdsa(&private_key)
-//!     .expect("Failed to sign message");
+//! let signed_msg = msg.sign_ecdsa(&private_key);
 //!
 //! // Verify the signed message.
-//! assert!(signed_msg.verify());
+//! assert_eq!(signed_msg.recover_ecdsa(), Ok(private_key.into()));
```
matteojug, line 51: nit: is this last comment correct?

matteojug (approved):
Approving as we are relying on `prost` being deterministic enough. Opened https://github.com/stacks-network/sbtc/issues/1001 as a followup to ensure we checked all the types.

signer/src/ecdsa.rs
djordon: Nope, fixing them now.

djordon: Fixed in https://github.com/stacks-network/sbtc/pull/957/commits/304999adfa97b97d2ad69909203cf65b29132821.

PR #958, opened 2024-11-26
feat: verify all p2p messages

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/995.

We now verify all network messages, not just the WSTS ones.

## Changes

* Send `RecoverableSignatures` in signed messages, remove the public key. This reduces the size of our messages over the network without limiting our ability to verify the messages.
* Always verify that the message was signed with the given public key upon receipt of the message.

## Testing Information

This might needs another test, but I think our existing tests cover this case.

## Checklist:


cylewitruk:
So I guess I'm mostly curious if we feel we need this? We already have Strict message validation enabled in the gossipsub config so all messages are already signed using the signer's pubkey (now that we use the signers' keys for identity) and verified at receipt/propagation.

More info here: https://github.com/libp2p/specs/blob/master/pubsub%2FREADME.md

djordon:
> So I guess I'm mostly curious if we feel we need this?

No we don't, not strictly speaking at least. We have multiple layers of security: one or two at the networking layer, another with WSTS checks, validation of bitcoin and stacks transactions, trusted peers, and so on. But I think it's best to go with a "[defense in depth](https://en.wikipedia.org/wiki/Defense_in_depth_(computing))" approach (other [bitcoin bridges](https://blog.rootstock.io/noticia/the-security-architecture-of-rootstock-and-the-principle-of-defense/) do this too). So I'm for a simple change like this for that reason.

But AR also flagged missing integrity checks on our messages (in https://github.com/stacks-network/sbtc/issues/517) so I think we should care that we sign our entire messages on top of the networking layer. And if we care about signing the entire message, we should verify them as well.

cylewitruk (commented):
@djordan from our brief discussion yesterday -- have you had time to think through based on what this does vs libp2p and determine if you think this adds additional security?

My concern at this point is that we're introducing p2p-protocol breaking changes here, so this might need to wait until the next release?

djordon:
> @djordan from our brief discussion yesterday -- have you had time to think through based on what this does vs libp2p and determine if you think this adds additional security?
> 
> My concern at this point is that we're introducing p2p-protocol breaking changes here, so this might need to wait until the next release?

I have. It seems like the high level idea of the `StrictSign` sign policy in libp2p is almost exactly what we are doing here. That leaves us with a few options:
1. Remove signing of our own messages and get rid of the `Signed<T>` type.
2. Remove our use of the `StrictSign` libp2p policy.
3. Keep the signed type and the `StrictSign` libp2p policy and verify the signatures in the signed type.

I don't think we should do (2). And I'm not sure if we should do (3), but if we do it should be because we're adding extra security. The changes here aren't doing that. The one thing that we could do is add a timestamp to all messages and reject the ones that are at least 10 minutes old. That wouldn't be the changes here though.

aldur:
As instructed, marked as draft to avoid merging it early.

djordon:
> It seems like the high level idea of the `StrictSign` sign policy in libp2p is almost exactly what we are doing here.

Wait, I got this wrong, I think that they do different things. The libp2p strict policy ensures that the source of the message is from who it's supposed to be from. Adding this check helps ensure that a signer cannot replay someone else's messages as if they are their own.

cylewitruk:
> > It seems like the high level idea of the `StrictSign` sign policy in libp2p is almost exactly what we are doing here.
> 
> Wait, I got this wrong, I think that they do different things. The libp2p strict policy ensures that the source of the message is from who it's supposed to be from. Adding this check helps ensure that a signer cannot replay someone else's messages as if they are their own.

Okay, i'll have a fresh look through 👍

signer/src/ecdsa.rs
```diff
@@ -8,85 +8,128 @@
...
-        let msg = secp256k1::Message::from_digest(self.inner.digest());
-        let Ok(sig) = secp256k1::ecdsa::Signature::from_compact(&self.signature) else {
-            return false;
-        };
+impl Signed<SignerMessage> {
+    /// Determines the public key for which `sig` is a valid signature for
+    /// `msg`.
+    pub fn recover_ecdsa(&self) -> Result<PublicKey, Error> {
+        let msg = secp256k1::Message::from_digest(self.inner.to_digest());
+        self.signature.recover_ecdsa(&msg)
```
matteojug, line 42: nit: reuse `recover_ecdsa_with_digest` with `self.inner.to_digest()`

signer/src/error.rs
```diff
@@ -14,6 +13,10 @@ use crate::storage::model::SigHash;
 /// Top-level signer error
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+    /// Indicates that a sweep transaction with the specified txid could not be found.
+    #[error("sweep transaction not found: {0}")]
+    DecodeProtobuf(#[from] prost::DecodeError),
+
```
matteojug, line 19: Is this some leftover?

signer/src/ecdsa.rs
```diff
@@ -8,85 +8,128 @@
...
+    /// One of the assumptions is that repeated serialization of prost
+    /// types generate the same bytes. It also assumes that fields are
+    /// serialized in order by their tag. This is not true for protobufs
+    /// generally, but it is for the prost protobuf implementation. The
+    /// implementation was checked by inspecting the output of `cargo
+    /// expand`.
+    /// <https://protobuf.dev/programming-guides/serialization-not-canonical/>
+    /// <https://protobuf.dev/programming-guides/encoding/#order>
+    /// <https://github.com/tokio-rs/prost/blob/v0.12.6/prost/src/message.rs#L108-L134>
+    pub fn decode_with_digest(data: &[u8]) -> Result<(Self, [u8; 32]), Error> {
```
matteojug, line 86: IIUC another assumption here is that when we encode (in `to_digest` below):
```
        let ans = proto::Signed {
            signature: None,
            signer_message: Some(self.clone().into()),
        };
```
we are zero-length encoding `signature` (and `Signed` itself)? I guess this is a proto spec, but it seems we are messing with proto internals to do something we probably shouldn't.
At this point, wouldn't it be better to have `Signed` with `signer_message: bytes` directly? We would need to do a double encoding/decoding ourselves, but we can avoid some assumptions about the layout (and doing prost work).

signer/src/ecdsa.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/958/commits/3078ca8c7a44e4526cbf15dbc82c66f3b6e67416. But I actually need to remove that function entirely.

signer/src/ecdsa.rs
djordon: That's a good point. Let me try that.

signer/src/error.rs
djordon: Kinda, I had forgotten that I added it. Will update.

signer/src/ecdsa.rs
djordon: Wait, I'm having cold feet about both approaches.

djordon: > IIUC another assumption here is that when we encode (in `to_digest` below):
> 
> ```
>         let ans = proto::Signed {
>             signature: None,
>             signer_message: Some(self.clone().into()),
>         };
> ```
> 
> we are zero-length encoding `signature` (and `Signed` itself)? I guess this is a proto spec, but it seems we are messing with proto internals to do something we probably shouldn't.

Yeah, I believe that this assumption is part of the protobuf spec and so should be constant across all implementations.

> At this point, wouldn't it be better to have `Signed` with `signer_message: bytes` directly? We would need to do a double encoding/decoding ourselves, but we can avoid some assumptions about the layout (and doing prost work).

I really want to keep the protobuf definition self describing, but it appears that we pay a steep price for doing so. So I'll try changing it.

matteojug (approved):

PR #961, opened 2024-11-26
feat: change emily config to uri

cylewitruk:
## Description

Changes the Emily endpoint config to a URI in the format `http(s)://[api-key@]<host>:<port>`.

Making this change primarily due to the fact that things aren't working well with environment variables and the URI format already supports authentication information.

## Changes

- Modified the emily endpoint format in config
- Updated tests
- Updated demo-cli
- Added the env vars to the local signer `.env` files 1-3.

## Checklist:

### DIFF
--- a/signer/src/bin/demo_cli.rs
+++ b/signer/src/bin/demo_cli.rs
@@ -117,14 +117,17 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .expect("No Emily endpoints configured");
     let emily_client_config = Configuration {
         base_path: emily_api_endpoint_config
-            .endpoint
             .to_string()
             .trim_end_matches("/")
             .to_string(),
-        api_key: emily_api_endpoint_config
-            .api_key
-            .clone()
-            .map(|key_string| ApiKey { prefix: None, key: key_string }),
+        api_key: if emily_api_endpoint_config.username().is_empty() {
+            None
+        } else {
+            Some(ApiKey {
+                prefix: None,
+                key: emily_api_endpoint_config.username().to_string(),
+            })
+        },
         ..Default::default()
     };
 
--- a/signer/src/bin/demo_cli.rs
+++ b/signer/src/bin/demo_cli.rs
@@ -110,24 +110,30 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Setup our Emily client configuration by getting the first configured endpoint
     // and using that to populate the client.
-    let emily_api_endpoint_config = settings
+    let mut emily_api_endpoint = settings
         .emily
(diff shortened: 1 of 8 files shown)


cylewitruk:
> Much better! Semantically it might be a bit confusing because it's an API key and we are shoving it under the username syntax of the URI, but it makes things a lot simpler, so I am happy with it.

Yeah, I figured this was at least more natural than a query param or messing with the url and it's consistent with how we specify other endpoints in the config, so I chose the lesser evil 😊

signer/src/bin/demo_cli.rs
```diff
@@ -117,14 +117,17 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .expect("No Emily endpoints configured");
     let emily_client_config = Configuration {
         base_path: emily_api_endpoint_config
-            .endpoint
             .to_string()
             .trim_end_matches("/")
             .to_string(),
```
matteojug, line 122: Will the `to_string` include the username (ie, the api key)?

cylewitruk: Yeah it will, we don't have an api key in devenv so it wasn't a problem, but I guess you run this against testnet with keys don't you... will add the same check here

matteojug: I was wondering if that "username" may accidentally get leaked in some logs

cylewitruk: Nah, it's not an actual username, this is just a URI-format hijack (like `postgres://user:pass@host.port/database`). We take this directly and map it to the generated emily code's `api_key` -- it's only the format that has changed.

cylewitruk: I think this should do it [2fc4d2e](https://github.com/stacks-network/sbtc/pull/961/commits/2fc4d2e66a098bf13ad1d3be406cf0fa11744999)

matteojug: I think you should port that same change also in `signer/src/emily_client.rs`

matteojug (approved):
LGTM! Should we mark "breaking changes" PRs (in this case, config related one) somehow? So that we remember there's thing we need to do when updating signers

cylewitruk:
> LGTM! Should we mark "breaking changes" PRs (in this case, config related one) somehow? So that we remember there's thing we need to do when updating signers

Yeah, I brought that up this morning before I read this in the sbtc channel with my fears of protocol changes, I suggested that we have an additional section in the PR template to annotate additional release considerations -- but we should probably have a label as well so that we can easily check all merged PRs before release. I wonder if we can also get the automated release notes to add a "Breaking Changes" section somehow...

PR #962, opened 2024-11-26
Add ability to pass config file via command line arguments

MCJOHN974:
## Description

Closes: I think there is no issue for it

## Changes

Use command line arguments in `SETTINGS` initialization to allow passing custom config files, as @aldur requested

## Usage

Now you can pass custom config file via `-c` or `--config` options:

```
<path-to-blocklist-binary> -c my-custom-config.toml
```
or
```
cargo run -- --config my-custom-config.toml
```

## Checklist:

### DIFF
--- a/blocklist-client/src/config.rs
+++ b/blocklist-client/src/config.rs
@@ -16,7 +16,7 @@ struct Cli {
     config: Option<PathBuf>,
 }
 
-/// Statically configured settings for the Blocklist client
+/// Command line arguments for the blocklist client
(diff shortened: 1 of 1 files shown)


cylewitruk:
> @aldur told me that this command line option worked somehow before, but, I didn't find any code which process arguments explicitly. Maybe it was some magic in `Settings` internals?? Anyway I think explicit processing of arguments will be easier for future changes

I think he's thinking of the signer binary 😊

aldur:
I think the reason I was confused is that the docker-config file we were using did provide `-c`.

Thanks @MCJOHN974 for jumping at this!

PR #963, opened 2024-11-26
fix: remove trailing slash from blocklist client

matteojug:
## Description

Avoid double slashes in api calls.

## Testing Information

Tested locally with
```
[blocklist_client]
endpoint = "http://blocklist-client:3030"
```
and adding the API key.

## Checklist:

### DIFF
--- a/signer/src/blocklist_client.rs
+++ b/signer/src/blocklist_client.rs
@@ -44,10 +44,18 @@ impl BlocklistClient {
     pub fn new(ctx: &impl Context) -> Option<Self> {
         let config = ctx.config().blocklist_client.as_ref()?;
 
-        let config = Configuration {
+        let mut config = Configuration {
             base_path: config.endpoint.to_string(),
             ..Default::default()
         };
+
+        // Url::parse defaults `path` to `/` even if the parsed url was without the trailing `/`
+        // causing the api calls to have two leading slashes in the path (getting a 404)
+        config.base_path = config
+            .base_path
+            .to_string()
+            .trim_end_matches("/")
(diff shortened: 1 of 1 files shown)


signer/src/blocklist_client.rs
```diff
@@ -44,11 +44,19 @@ impl BlocklistClient {
     pub fn new(ctx: &impl Context) -> Option<Self> {
         let config = ctx.config().blocklist_client.as_ref()?;
 
-        let config = Configuration {
+        let mut config = Configuration {
```
cylewitruk, line 47: I know this isn't your doing, but it feels pretty extreme that we're talking a whole `Context` to build this when all we really need is the url -- makes the tests below a lot more complicado than they need to be 😬

cylewitruk (approved):
Looks good, with the exception of that lingering `Context` thing that I mentioned (but has nothing to do with these changes)

signer/src/blocklist_client.rs
```diff
@@ -44,11 +44,19 @@ impl BlocklistClient {
...
             base_path: config.endpoint.to_string(),
             ..Default::default()
         };
 
+        // Url::parse defaults `path` to `/` even if the parsed url was without the trailing `/`
```
aldur, line 52: I was just wondering who I had to blame for this behavior, stdlib that thinks it is smarter than us!

aldur (approved):

signer/src/blocklist_client.rs
matteojug: Yeah, I will fix this later

Issue #856, opened 2024-11-15
[Feature]: In the signers we use the caps to filter out deposits


djordon:
# Feature - In the signers we use the caps to filter out deposits

## 1. Description

This is the part where we filter out the deposits based on the cap amounts that we receive from Emily. We'll also need to get the global amount pegged in as part of completing this ticket. For that we can use the stacks-node to get an idea of what we have minted already. That is probably a better source of truth than the signer's database.

### 1.1 Context & Purpose

This is part of https://github.com/stacks-network/sbtc/issues/854

## 2. Technical Details:

Update the signer to filter out pending and accepted deposits that are above the caps.

### 2.1 Acceptance Criteria:
   - [x] The signer to filter out pending and accepted deposits that are above the transaction cap.
   - [x] The signer to filter out pending and accepted deposits that are above the global cap.

## 3. Related Issues and Pull Requests (optional):

We have two design tickets for this:
* https://github.com/stacks-network/sbtc/issues/854
* https://github.com/stacks-network/sbtc/issues/562
* https://github.com/stacks-network/sbtc/issues/791

PR #967, opened 2024-11-26
Feat: validate deposits are below total and per-request-cap

Jiloc:
# Issue Description

Closes: #856

## Changes
  - `SbtcRequests` now filters out deposit requests that would cause the total transaction amount to exceed the cap.  
      - The filtering logic is straightforward and processes deposits in the order they were received.
  - Signers revalidate `BitcoinPreSignRequest` to ensure the total sum of deposits does not exceed the cap.  
  - Individual transactions are also checked against the per-request cap in the validate function. (Maybe I should remove it from there and just merge the check on the `can_accept` field of the `SignerDepositDecision` - like in the original implementation from Cyle)
  - A new `ValidationCache` struct caches `DepositRequestReport` and `SignerVotes` from the database.  
     - This prevents redundant queries during pre-validation and per-request validation steps.

## Testing Information
  - Added unit tests for filtering deposit requests that exceed the total cap when creating new transactions.  
  - Added unit tests for validating full transactions on the signer side.
  - Adapted existing tests to align with the new logic.

  - Still need to add tests for validating individual requests for the `AmountTooHigh` error. (That may be removed)

### DIFF
--- a/signer/src/bitcoin/utxo.rs
+++ b/signer/src/bitcoin/utxo.rs
@@ -2692,4 +2692,68 @@ mod tests {
         // number-of-requests`.
         assert!(combined_fee <= (fee + Amount::from_sat(3u64)));
     }
+
+    #[test_case(vec![
+        create_deposit(10_000, 100_000, 0),
+        create_deposit(10_000, 100_000, 0),
+        create_deposit(10_000, 100_000, 0),
+        create_deposit(10_000, 100_000, 0),
+        create_deposit(10_000, 100_000, 0),
+    ], 3, 30_000, 30_000; "should_accept_deposits_until_max_mintable_reached")]
+    #[test_case(vec![
+        create_deposit(10_000, 100_000, 0),
+        create_deposit(10_000, 100_000, 0),
+    ], 1, 10_000, 15_000; "should_accept_all_deposits_when_under_max_mintable")]
+    #[test_case(vec![
+        create_deposit(10_000, 100_000, 0),
+    ], 0, 0, 0; "should_handle_empty_deposit_list")]
+    #[test_case(vec![
+        create_deposit(10_000, 0, 0),
+        create_deposit(8_000, 100_000, 0),
+        create_deposit(5_000, 100_000, 0),
+    ], 1, 8000, 10000; "should_skip_invalid_fee_and_accept_valid_deposits")]
+    #[test_case(vec![
+        create_deposit(10_001, 100_000, 0),
+    ], 0, 0, 10_000; "should_reject_single_deposit_exceeding_max_mintable")]
+    #[tokio::test]
+    async fn test_construct_transactions_filters_deposits_over_max_mintable(
+        deposits: Vec<DepositRequest>,
+        num_accepted_requests: usize,
+        accepted_amount: u64,
+        max_mintable: u64,
+    ) {
+        // Each deposit and withdrawal has a max fee greater than the current market fee rate
+        let public_key = XOnlyPublicKey::from_str(X_ONLY_PUBLIC_KEY1).unwrap();
+        let requests = SbtcRequests {
+            deposits: deposits,
+            withdrawals: vec![],
+            signer_state: SignerBtcState {
+                utxo: SignerUtxo {
+                    outpoint: generate_outpoint(300_000, 0),
+                    amount: 300_000_000,
+                    public_key,
+                },
+                fee_rate: 25.0,
+                public_key,
+                last_fees: None,
+                magic_bytes: [0; 2],
+            },
+            num_signers: 10,
+            accept_threshold: 8,
+            max_mintable: max_mintable,
+        };
+        let txs = requests.construct_transactions().unwrap();
+        let nr_requests = txs.iter().map(|tx| tx.requests.len()).sum::<usize>();
+        let total_amount: u64 = txs
+            .iter()
+            .map(|tx| {
+                tx.requests
+                    .iter()
+                    .map(|req| req.as_deposit().unwrap().amount)
+            })
+            .flatten()
+            .sum();
+        assert_eq!(nr_requests, num_accepted_requests);
+        assert_eq!(total_amount, accepted_amount);
+    }
 }
--- a/signer/src/bitcoin/utxo.rs
+++ b/signer/src/bitcoin/utxo.rs
@@ -133,6 +133,8 @@ pub struct SbtcRequests {
     pub accept_threshold: u16,
     /// The total number of signers.
     pub num_signers: u16,
+    /// The maximum amount of sBTC that can be minted in sats.
+    pub max_mintable: u64,
 }
 
 impl SbtcRequests {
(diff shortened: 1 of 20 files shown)


signer/src/request_decider.rs
```diff
@@ -329,20 +311,6 @@ where
...
-                tracing::info!(
-                    txid = %req.txid,
-                    outpoint = %req.outpoint(),
-                    amount = %req.amount,
-                    "deposit request exceeds per-deposit cap"
-                );
-                return Ok(false);
-            }
-        }
-
```
Jiloc, line 345: I'm having second thoughts about removing this check. I believe the per-withdrawal and per-deposit cap validation should remain here, and I'll reintroduce the code.
I'll update the comments to clarify that can_accept means the request was both approved by the blocklist and within the cap limits.

cylewitruk: Yeah I think it's good to have the checks here as well, because if enough signers vote `can_accept = false` then the coordinator won't even try to package them.

Note that I think we need to check `amount + max_fee` here, otherwise the user could potentially specify `10 000` with max fee of `10 000 000` and circumvent the check?

I might consider re-adding the limits state in context and updating the limits from Emily once in the block observer upon new Bitcoin block though, because now the limit-pulling from Emily will be scattered a bit all over the place. If it's updated when receiving a new Bitcoin block then the limits will apply to all validations done by the signer within that block (request-decider, tx-signer and tx-coordinator).

signer/src/bitcoin/utxo.rs
```diff
@@ -160,16 +162,24 @@ impl SbtcRequests {
...
-            .filter(|req| req.max_fee >= minimum_deposit_fee)
-            .map(RequestRef::Deposit);
 
+        let mut amount_to_mint = 0;
+        let deposits = self.deposits.iter().filter_map(|req| {
+            let valid_fee = req.max_fee >= minimum_deposit_fee;
+            let valid_amount = amount_to_mint + req.amount <= self.max_mintable;
+
+            if valid_fee && valid_amount {
+                amount_to_mint += req.amount;
```
cylewitruk, line 177: The actual minting amount can be greater than this, right? See my reply to your PR comment. I _think_ we need to check `amount + (max_fee - min_deposit_fee)`

signer/src/bitcoin/validation.rs
```diff
@@ -91,10 +104,121 @@ pub fn is_unique(package: &[TxRequestIds]) -> bool {
...
+        cache: &ValidationCache<'a>,
+    ) -> Result<(), Error>
+    where
+        C: Context + Send + Sync,
+    {
+        let sbtc_supply = ctx
+            .get_stacks_client()
+            .get_sbtc_total_supply(&ctx.config().signer.deployer)
+            .await?;
+        let total_cap = cache.sbtc_limits.total_cap().unwrap_or(Amount::MAX_MONEY);
```
cylewitruk, line 197: I'd add some comments here explaining that if the limit is `None` in Emily then we use `MAX_MONEY` to simulate no cap, etc. (it might not be obvious to a future reader why we're doing this)

signer/src/bitcoin/validation.rs
```diff
@@ -91,10 +104,121 @@ pub fn is_unique(package: &[TxRequestIds]) -> bool {
...
+        let max_mintable = total_cap
+            .checked_sub(sbtc_supply)
+            .unwrap_or(Amount::ZERO)
+            .to_sat();
+        cache
+            .deposit_reports
+            .values()
+            .try_fold(0u64, |acc, (report, _)| {
+                acc.checked_add(report.amount)
```
cylewitruk, line 206: Do we need to account for max fee here?

Note that there's other places where we don't account for max_fee, but I'll wait on commenting on those until we flesh out if it's needed or not in the previous comments.

signer/src/transaction_coordinator.rs
```diff
@@ -1232,13 +1233,30 @@ where
         if deposits.is_empty() && withdrawals.is_empty() {
             return Ok(None);
         }
+        // Get the amount of sBTC that can be currently minted.
+        // This is the total cap received from Emily minus the current supply.
+        let sbtc_supply = self
```
cylewitruk, line 1238: Looks like this logic is done in several places, might be best to add a common function somewhere?

signer/tests/integration/bitcoin_validation.rs
```diff
@@ -663,13 +738,27 @@ async fn outcome_is_independent_of_input_order() {
...
 
-    let ctx = TestContext::builder()
+    let mut ctx = TestContext::builder()
         .with_storage(db.clone())
         .with_first_bitcoin_core_client()
         .with_mocked_stacks_client()
         .with_mocked_emily_client()
         .build();
 
+    ctx.with_emily_client(|client| {
```
cylewitruk, line 748: suggestion: just to make these cleaner (i.e. not take so much space) I might add a common function somewhere in `testing` that takes either a context or the stacks/emily clients and sets up this expect, like a `configure_zero_cap_limits()` or something, looks like we use exactly this block in a number of places both in this file and elsewhere.

signer/src/request_decider.rs
matteojug: mmm something to consider is that after https://github.com/stacks-network/sbtc/pull/968 we are no longer voting on requests we already voted for, so if we want to reconsider deposits after we change the limit we should *not* vote against them in here.

cylewitruk: My only concern with not voting against them is that I guess the coordinators will perpetually try to package them until the total supply decreases (which can't currently happen since we don't have withdrawals) and they manage to get squeezed in, so the list of pending deposits could potentially grow pretty big.

Jiloc: As discussed at the standup, we are going to remove this check from here, but we will introduce a max lock filter (in a separate PR) and a per-deposit-cap filter in the coordinator (implemented in https://github.com/stacks-network/sbtc/pull/967/commits/fe55499a492a0b1b5f94ac40f2c1438f50478ef3 ).

signer/src/bitcoin/validation.rs
Jiloc: comments added wherever we call an `unwrap_or` in in https://github.com/stacks-network/sbtc/pull/967/commits/fe55499a492a0b1b5f94ac40f2c1438f50478ef3

signer/src/bitcoin/validation.rs
```diff
@@ -142,53 +266,32 @@ impl BitcoinPreSignRequest {
...
-            };
-
-            let votes = db
-                .get_deposit_request_signer_votes(&txid, output_index, aggregate_key)
-                .await?;
-
+            let key = (&outpoint.txid, outpoint.vout);
+            let (report, votes) = cache
+                .deposit_reports
+                .remove(&key)
```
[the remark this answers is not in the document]
Jiloc: this is "consuming" what was in the cache so that we avoid copying, the cache will be empty will be empty after this

signer/src/bitcoin/utxo.rs
```diff
@@ -160,16 +162,24 @@ impl SbtcRequests {
...
-        let deposits = self
-            .deposits
-            .iter()
-            .filter(|req| req.max_fee >= minimum_deposit_fee)
-            .map(RequestRef::Deposit);
 
+        let mut amount_to_mint = 0;
+        let deposits = self.deposits.iter().filter_map(|req| {
+            let valid_fee = req.max_fee >= minimum_deposit_fee;
+            let valid_amount = amount_to_mint + req.amount <= self.max_mintable;
```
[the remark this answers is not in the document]
Jiloc: added in https://github.com/stacks-network/sbtc/pull/967/commits/fe55499a492a0b1b5f94ac40f2c1438f50478ef3

signer/tests/integration/bitcoin_validation.rs
Jiloc: Now that the cap are cached, those are not needed anymore fortunately!

signer/src/transaction_coordinator.rs
Jiloc: I moved the entire logic of updating the caps in the block observer, as you suggested. That function will now both call Emily and also get the sbtc supply, if needed

Issue #934, opened 2024-11-25
[Feature]: Send-Many Functions


setzeus:
# Feature - Adding A 'Send-Many' Function To The sBTC Token Contract

## 1. Description
As pointed out by @friedger, the 'send-many' feature is seen in most sip-10 projects & critical to many projects that need to send tokens to more than one person at a time. Instead of forcing users & contracts to call many times, we should add a 'send-many' function to the existing contract.

### 1.1 Context & Purpose
As mentioned above, it's not part of sip-10, but most fungible tokens these days include a 'send-many' function for common use-cases; there are a lot of pros & no immediate cons that come to mind in implementing this.

## 2. Technical Details:
Need to add a single function to the sBTC Token Contract.

### 2.1 Acceptance Criteria:
   - [ ] Can send multiple transactions with one contract call

PR #969, opened 2024-11-27
Added 'send-many' Function To sBTC Token

setzeus:
## Description
This PR adds the ability to send multiple(many), specifically up to 200, sBTC transactions at once. This feature is particularly useful for DeFi apps that need to send to many recipients at once. A good example of this is a stacking app that wants to pay stacking rewards in sBTC.

Closes: #934 

## Changes
Adds one public & one private function to the sbtc-token.clar Clarity contract. 

## Testing Information
A unit test has been added here.

## Checklist:


contracts/contracts/sbtc-token.clar
```diff
@@ -110,6 +112,22 @@
 	)
 )
 
+(define-public (transfer-many (recipients (list 200 { amount: uint, sender: principal, to: principal, memo: (optional (buff 34)) })))
+	(fold complete-individual-transfer recipients (ok u0))
+)
+
+(define-private (complete-individual-transfer (individual-transfer { amount: uint, sender: principal, to: principal, memo: (optional (buff 34)) }) (helper-response (response uint uint)))
```
cylewitruk, line 119: Style Q: is using super-long lines the general style convention in clarity, or can this be multi-lined in an easy-to-read way?

contracts/contracts/sbtc-token.clar
```diff
@@ -110,6 +112,22 @@
 	)
 )
 
+(define-public (transfer-many (recipients (list 200 { amount: uint, sender: principal, to: principal, memo: (optional (buff 34)) })))
```
cylewitruk, line 115: Q: Should these be within the `;; sip-010-trait` section since they're not part of the trait?

contracts/contracts/sbtc-token.clar
```diff
@@ -110,6 +112,22 @@
...
+	(fold complete-individual-transfer recipients (ok u0))
+)
+
+(define-private (complete-individual-transfer (individual-transfer { amount: uint, sender: principal, to: principal, memo: (optional (buff 34)) }) (helper-response (response uint uint)))
+    (match helper-response 
+        index
+            (begin 
+                (unwrap! (transfer (get amount individual-transfer) (get sender individual-transfer) (get to individual-transfer) (get memo individual-transfer)) (err (+ ERR_TRANSFER_INDEX_PREFIX (+ u10 index))))
```
cylewitruk, line 123: This is largely just food for thought, but figured I'd throw out the idea anyway --

It looks like we don't preserve the eventual error code from `transfer()` with this, right? I wonder if we could encode both the transfer index and the transfer error code by assigning a range for transfers:

```
(define-constant ERR_TRANSFER_INDEX_PREFIX u1000)

(define-private (complete-individual-transfer (individual-transfer { amount: uint, sender: principal, to: principal, memo: (optional (buff 34)) }) (helper-response (response uint uint)))
    (match helper-response 
        index
            (let ((transfer-result (transfer (get amount individual-transfer) (get sender individual-transfer) (get to individual-transfer) (get memo individual-transfer))))
                (match transfer-result
                    transfer-ok (ok (bit-or (bit-shift-left (+ ERR_TRANSFER_INDEX_PREFIX index) u4) u0))
                    transfer-err (err (bitwise-or (bit-shift-left (+ ERR_TRANSFER_INDEX_PREFIX index) u4) transfer-err))
                )
            )
        err-response
            (err err-response)
    )
)
´´´

Which I think should give something like:

```
Index: 0
Result: ok
Encoded Value: 10000 (1000 << 4 | 0)

Index: 1
Result: ERR_NOT_OWNER
Encoded Value: 10011 (1001 << 4 | 1)

Index: 2
Result: ok
Encoded Value: 10020 (1002 << 4 | 0)
```

File: contracts/contracts/sbtc-token.clar (hunk 2 above)
Comment (setzeus) [reply contracts/contracts/sbtc-token.clar:115]: Copy. Moved around.

File: contracts/contracts/sbtc-token.clar (hunk 3 above)
Comment (setzeus) [reply contracts/contracts/sbtc-token.clar:123]: Right. I see the idea. we use this fold method a few other times in this contract so if we change it here we should change it everywhere else. 

I'll open an issue for this as a 'nice-to-have' & think on how I'd implement.

Comment (friedger) [reply contracts/contracts/sbtc-token.clar]: What is the purpose of returning the index in the error?  Is it so that other contracts can fix the call? 

If the index is only about user experience, tools like stxer.xyz can help users to understand errors.  

I'd prefer a simpler call with just a match and (transfer..)

File: contracts/tests/sbtc-token.test.ts [hunk 4]
```diff
@@ -101,6 +101,76 @@ describe("sBTC token contract", () => {
...
+        topic: "completed-deposit",
+        bitcoinTxid: new Uint8Array(32).fill(0),
+        outputIndex: 0n,
+        amount: 1000n,
+        burnHash,
+        burnHeight: BigInt(burnHeight),
+        sweepTxid: new Uint8Array(32).fill(1),
+      });
+      const receipt1 = txOk(
+        token.transferMany({
```
friedger, line 139: Add a test with a contract as contract-caller, so that it is shown that sBTC is a "safe transferable" token.

contracts/contracts/sbtc-token.clar
setzeus: Yeap. So that an operator can quickly remove the errored transaction & re-submit.

cylewitruk: Yeah this was primarily just an idea on how to give the caller clear information directly in the result (based on what they submitted), and thinking that the errors for transfer may expand in the future. But just an idea, I'll resolve this 😊

friedger: For me, the extra costs are not justified for the convenience of the user

contracts/contracts/sbtc-deposit.clar
```diff
@@ -98,7 +98,7 @@
     (match helper-response 
         index
             (begin 
-                (try! (unwrap! (complete-deposit-wrapper (get txid deposit) (get vout-index deposit) (get amount deposit) (get recipient deposit) (get burn-hash deposit) (get burn-height deposit) (get sweep-txid deposit)) (err (+ ERR_DEPOSIT_INDEX_PREFIX (+ u10 index)))))
+                (unwrap! (complete-deposit-wrapper (get txid deposit) (get vout-index deposit) (get amount deposit) (get recipient deposit) (get burn-hash deposit) (get burn-height deposit) (get sweep-txid deposit)) (err (+ ERR_DEPOSIT_INDEX_PREFIX (+ u10 index))))
```
cylewitruk, line 101: sorry maybe i missed this, long line :)

contracts/tests/sbtc-token.test.ts
setzeus: Was exploring this today & I think the best way to do this is probably not to create an actual contract but to create one at runtime when tests are executed. Have you done this before in Clarinet?

setzeus: I added a test contract 'sbtc-token-test' that tests a contract call properly call the main function in this PR (transfer-many).

setzeus:
Seeing something odd happening with Clarigen if you have a few minutes to check this out @hstove:

Whenever I run 'pnpm test' an additional block is appended to the 'default.simnet-plan.yaml' file, which now breaks the tests that use simnet.block to check the right height (specifically the three tests you currently see failing in this PR). I don't believe 'pnpm test' had this behavior before as these tests were currently passing? Did I cause this locally somehow?

contracts/tests/sbtc-token.test.ts
```diff
@@ -101,6 +101,197 @@ describe("sBTC token contract", () => {
...
+          recipients: [
+            {
+              amount: 100n,
+              sender: alice,
+              to: bob,
+              memo: null
+            },
+            {
+              amount: 100n,
```
djordon, line 148: I would change this a separate value from the one we use for bob.

contracts/tests/sbtc-token.test.ts
```diff
@@ -101,6 +101,197 @@ describe("sBTC token contract", () => {
...
+              sender: alice,
+              to: charlie,
+              memo: null
+            },
+          ],
+        }),
+        bob
+      );
+      // Many-transfer errors start at 6, error happens at index 0
+      expect(receipt1.value).toEqual(errors.token.ERR_TRANSFER);
```
djordon, line 228: I would add another case where the second transfer fails and the first succeeds so that we get the right error number (`errors.token.ERR_TRANSFER + 1`).

djordon (approved):
Looks good. Left some comments but you can feel free to ignore them.

Issue #971, opened 2024-11-27
[Chore]: Clarity Bugs


setzeus:
# Chore - Clarity Bugs
## 1. Description
@friedger was kind enough to point out two existing bugs in the Clarity contracts:

- [ ] An extra response wrapper in the 'complete-deposit-wrapper' function
- [ ] A recalculated value that's already saved as a variable

## 2. Why It's Necessary
The first issue could (possibly?) lead to an exploit since it's possible the contract call fails since the sbtc-registry contracts updates with 'map-insert'. 

The second issue is strictly optimization.

## 3. Execution Checklist

- [ ] Remove the (ok ...) wrapper from the 'complete-deposit' contract-call
- [ ] Reuse 'version-int' variable

PR #972, opened 2024-11-27
Minor Clarity Issues

setzeus:
## Description
This small PR addresses two bugs for the Clarity contracts recently opened under a single issue.

Closes: #971 

## Changes
Both changes mentioned in linked issue are implemented. Removed the unnecessary (ok ...) wrapper & reused the 'version-int' variable.

## Testing Information
No new tests were added but all unit tests still intact.

## Checklist:

### DIFF
--- a/contracts/contracts/sbtc-deposit.clar
+++ b/contracts/contracts/sbtc-deposit.clar
@@ -63,7 +63,7 @@
         (try! (contract-call? .sbtc-token protocol-mint amount recipient))
 
         ;; Complete the deposit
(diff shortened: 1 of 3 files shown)


djordon (commented):
Looks good!

But isn't there something to test here? Before the failure mode would occur if the contract caller wasn't the "protocol caller". And I think sBTC would still be being minted in that case, since that happens before the protocol caller check. I might be misreading or misunderstanding, but if not we should add an explicit test that contract calls from non-protocol callers would not lead to sBTC being minted.

djordon:
I misread the issue, we do that check before hand so there is nothing to really test.

djordon (approved):
✅ Yeah looks good!

Issue #973, opened 2024-11-27
[Feature]: Add test for reclaim path spending


djordon:
# Feature - Add test for reclaim path spending

## 1. Description

Bitcoin deposits for sBTC are taproot scriptPubKeys with two spend path scripts and key-spend paths with a (probably) unknown discrete logarithm. We've tested the deposit path spending, but not the reclaim path spending.

### 1.1 Context & Purpose

We don't have any examples of 

## 2. Technical Details:

Add an integration test to serve as an example of how to reclaim a deposit

### 2.1 Acceptance Criteria:
   - [ ] We have an integration test that tests at least one common reclaim script

PR #977, opened 2024-11-28
test: add test for reclaim path spending

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/973

## Changes

* Add a reclaim script integration test.

## Checklist:

### DIFF
--- a/sbtc/tests/integration/validation.rs
+++ b/sbtc/tests/integration/validation.rs
@@ -96,7 +96,7 @@ fn tx_validation_from_mempool() {
 /// The test proceeds as follows:
 /// 1. Create and submit a transaction where the lock script has a
 ///    non-minimal push for the deposit.
-/// 2. Confirm the transaction and try spend it immediately. The
+/// 2. Confirm the transaction and try to spend it immediately. The
 ///    transaction spending the "deposit" should be rejected.
 ///
 /// We do not attempt to create an actual P2TR deposit, but an
@@ -160,7 +160,7 @@ fn minimal_push_check() {
     regtest::p2tr_sign_transaction(&mut tx0, 0, &[utxo], &depositor.keypair);
     rpc.send_raw_transaction(&tx0).unwrap();
 
-    // 2. Confirm the transaction and try spend it immediately. The
+    // 2. Confirm the transaction and try to spend it immediately. The
     //    transaction spending the "deposit" should be rejected.
     faucet.generate_blocks(1);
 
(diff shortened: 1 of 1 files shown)


sbtc/tests/integration/validation.rs
```diff
@@ -386,3 +400,166 @@ fn op_csv_disabled() {
         err => panic!("{err}"),
     };
 }
+
+/// This validates that a user can reclaim their funds after the locktime.
+///
+/// The test proceeds as follows:
+/// 1. Spend all the user's funds for am sBTC deposit. Check that the
```
cylewitruk, line 407:
```suggestion
/// 1. Spend all the user's funds for a sBTC deposit. Check that the
```

sbtc/tests/integration/validation.rs
```diff
@@ -386,3 +400,166 @@ fn op_csv_disabled() {
...
+///    balance is zero.
+/// 2. Wait locktime blocks after the first confirmation.
+/// 3. Create and submit another transaction reclaiming the funds.
+/// 4. Confirm the transaction and check that the balance is what it is
+///    supposed to be, less the bitcoin transaction fees.
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
+#[test]
+fn reclaiming_rejected_deposits() {
```
cylewitruk, line 415: Should we also have a negative test? i.e. `cannot_reclaim_deposit_at_or_before_locktime()`? Or maybe this is covered somewhere else? Otherwise I guess it should be fairly easy to include it in this test prior to advancing past the lock time.

djordon: Yeah let's test that case here, the one we have in `op_csv_disabled` is insufficient.

sbtc/tests/integration/validation.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/977/commits/33fb28ccd494a137025020c0ed4a2286a0ec215e.

sbtc/tests/integration/validation.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/977/commits/33fb28ccd494a137025020c0ed4a2286a0ec215e

sbtc/tests/integration/validation.rs
```diff
@@ -386,3 +400,177 @@ fn op_csv_disabled() {
...
+        .unwrap();
+    // Now we can set the witness data.
+    let witness_data = [
+        signature.serialize().to_vec(),
+        reclaim_script.to_bytes(),
+        control_block.serialize(),
+    ];
+    reclaim_tx.input[0].witness = Witness::from_slice(&witness_data);
+
+    // Not enough time has passed so the submitting the reclaim transaction
```
cylewitruk, line 555: I think that we should do a `faucet.generate_blocks(lock_time as u64 - 1)` here, and change the `generate_blocks()` below to `1`, just to show that things work as expected around the boundary.

djordon: Great point. Doing so highlights an off-by-one error. Maybe we should increase out buffer by one?

djordon: Well not an error, but a way of thinking. You'll see

Issue #988, opened 2024-11-28
[Feature]: (High) Hardcoded Postgres login credentials


technovision99:
## 1. Description
The signer documentation https://github.com/stacks-network/sbtc/tree/main/signer#inspecting-the-signer-database states that `Once connected to the database, the signer migrator will create a default user sbtc_user with password sbtc_password which the signer uses to manage the signer state.`
Using these default credentials can allow anyone to compromise the Postgres instance in the event that it is ever exposed over the internet. 

## Recommendation
Switch the default user and password to configurable credentials.

cylewitruk:
The referenced `README.md` is severely out-of-date, thanks for pointing that out. The script that is currently run is:

```sql
-- Create the user if it doesn't exist
DO
$do$
BEGIN
   IF NOT EXISTS (
      SELECT FROM pg_catalog.pg_roles
      WHERE rolname = 'sbtc_user'
   ) THEN
      CREATE USER sbtc_user WITH PASSWORD 'sbtc_password';
   END IF;
END
$do$;
```

Which is largely intended for development purposes. I would be an advocate of not running this script at all unless explicitly requested via a CLI flag i.e. `--create-development-db-user` or something. Thoughts?

PR #994, opened 2024-11-29
fix: remove default user

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/988

## Changes

Remove the default user -- other than being bad because of hardcoded creds, we weren't even using it anyway as far as I'm aware.

## Testing Information

Devenv + demo script run was okay.

## Checklist:


djordon (approved):

matteojug:
Commented out the migration that create the user to avoid the possibility of racing during the startup to use that user to do bad things.

aldur:
If someone has provisioned their PG DB with that user we will delete it though, which can be a surprising side effect. Maybe just add a line to the docs?

signer/migrations/0006__remove_default_user.sql
```diff
@@ -0,0 +1,13 @@
+-- Delete default user if it exists
+DO
+$do$
+BEGIN
```
cylewitruk, line 4: I'm not sure we should do this, because it might blindly delete `sbtc_user` from a live signer if they ended up using that name.

matteojug: Good point, removed the script directly -- then we can communicate this to the current signers and let them decide if they want to remove the default user or not.

cylewitruk (approved):

Issue #979, opened 2024-11-28
[Feature]: Check message source is in current signing set


djordon:
# Feature - Check message source is in current signing set

## 1. Description

We currently check that the peer that propagated the message is in the current signing set, but we need to also check that the origin is in the current signing set as well.

### 1.1 Context & Purpose

Our network is supposed to be closed. That means that all messages must originate and propagate from trusted peers.

## 2. Technical Details:

I think we need this

https://github.com/stacks-network/sbtc/blob/06af219b0198369587ad97207fd78dc320ea3691/signer/src/network/libp2p/event_loop.rs#L342-L352

### 2.1 Acceptance Criteria:
   - [ ] Messages are only accepted from trusted peers

PR #997, opened 2024-12-01
feat: check message source is in the current signing set

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/979

## Changes

* Check that the peer that published a message is in the current signing set.

## Testing Information

This adds two tests, one for happy path gossiping and another for the check.

## Checklist:

### DIFF
--- a/signer/src/network/libp2p/event_loop.rs
+++ b/signer/src/network/libp2p/event_loop.rs
@@ -338,8 +338,22 @@ fn handle_gossipsub_event(
             message,
             ..
         } => {
-            if !ctx.state().current_signer_set().is_allowed_peer(&peer_id) {
+            let current_signer_set = ctx.state().current_signer_set();
+            if !current_signer_set.is_allowed_peer(&peer_id) {
                 tracing::warn!(%peer_id, "ignoring message from unknown peer");
+                return;
+            }
+
+            // The message may have originated from someone else, let's
+            // check that peer ID too. If we haven't been told the source
+            // then we distrust the message and ignore it.
+            let Some(origin_peer_id) = message.source else {
+                tracing::warn!(%peer_id, "origin peer id unknown, ignoring message");
+                return;
+            };
+
+            if !current_signer_set.is_allowed_peer(&origin_peer_id) {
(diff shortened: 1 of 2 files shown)


signer/src/network/libp2p/network.rs
```diff
@@ -206,4 +209,239 @@ mod tests {
...
+        });
+
+        // We wait for our first message, but not for too long, life is short.
+        let received_messages = signer_msg_stream
+            .take(1)
+            .take_until(tokio::time::sleep(Duration::from_secs(4)))
+            .collect::<Vec<_>>()
+            .await;
+
+        assert!(received_messages.is_empty());
```
matteojug, line 445: nit: while I'm pretty confident about the small change, can we add in here (even in a different PR) a check that signer 2 is getting signer 3 messages, and that if signer 2 sends a message signer 1 will receive it? To ensure that the test is not passing just because signer 1 is not connected at all

matteojug (approved):

signer/src/network/libp2p/event_loop.rs
```diff
@@ -338,11 +338,25 @@ fn handle_gossipsub_event(
             message,
             ..
         } => {
-            if !ctx.state().current_signer_set().is_allowed_peer(&peer_id) {
+            let current_signer_set = ctx.state().current_signer_set();
+            if !current_signer_set.is_allowed_peer(&peer_id) {
```
cylewitruk, line 342: note: this check is technically unnecessary since we can't receive a message via a peer if it's not connected, and you can't connect if not a valid peer. The origin check below was a good catch, though!

cylewitruk (approved):

signer/src/network/libp2p/event_loop.rs
djordon: Oh right. I'll add a comment that it is unnecessary.

signer/src/network/libp2p/network.rs
djordon: Good idea, I think I'll do this here.

Issue #722, opened 2024-10-28
[Tech Debt]: Fix DepositUpdate in handle_completed_deposit to contain the exact `btc_fee`


Jiloc:
## Description

PR that introduced tech-debt: #712

When we create `DepositUpdate` messages for Emily to update the deposit to `Confirmed`, we are currently hardcoding the [paid btc fee to 1](https://github.com/stacks-network/sbtc/blob/332e042e9881f3135152d5755934aeeb51dba094/signer/src/api/new_block.rs#L271). 

There could be multiple ways to extract the fee. One clean way could be to modify the [`complete-deposit`](https://github.com/stacks-network/sbtc/blob/main/contracts/contracts/sbtc-registry.clar#L233) function in the `sbtc-registry` contract to also print this information exactly like the [`complete-withdrawal-accept`](https://github.com/stacks-network/sbtc/blob/main/contracts/contracts/sbtc-registry.clar#L184) function event does.

This way we could simply read the value from the print event like we do [when we construct a `WithdrawalUpdate`](https://github.com/stacks-network/sbtc/blob/332e042e9881f3135152d5755934aeeb51dba094/signer/src/api/new_block.rs#L312)

djordon:
I see, so we should update this when we do https://github.com/stacks-network/sbtc/issues/583.

aldur:
Thank you! As discussed, moved this to https://github.com/stacks-network/sbtc/milestone/12.

djordon:
I think that it's useful to think about this ticket out as two separate issues:
1. Making sure that the real `btc_fee` fee is sent to Emily.
2. Adding the `btc_fee` to the clarity contract for `complete-deposit` contract calls.

We can and should do (1) now. A signer can compute the fee using the information in the `complete-deposit` event and the original deposit request UTXO. If a signer does not have the deposit request record in their database then it just shouldn't bother updating Emily. We can figure out whether we should do (2) later.

PR #1005, opened 2024-12-01
set `DepositUpdate` sweep tx fee in `handle_completed_deposit` 

Jiloc:
## Description

Closes: #722

## Changes
- Introduced a new `get_deposit_request` query.
- Updated `handle_completed_deposit` to fetch the deposti request connected to the event and compute the fee as the difference between the deposited amount and the amount of sBTC minted.
 
## Testing Information
- adapted existing unit and integration tests to work with the new change
- added new unit test and integration tests


djordon (commented):
Nice, I think we can simplify this a bit. If we subtract the amount in the `CompletedDepositEvent` from the deposit request amount then that will give us the fee.

cylewitruk:
I've kinda been thinking in the lines of removing the sweep tables since we have the information elsewhere (i.e. in the bitcoin_* tables). This creates a new dependency on them -- can we acoid that?

djordon:
> the sweep tables since we have the information elsewhere (i.e. in the bitcoin_* tables). This creates a new dependency on them -- can we avoid that?

Yeah, we should only need the `deposit_requests` table here.

signer/src/error.rs
```diff
@@ -26,6 +26,10 @@ pub enum Error {
     #[error("sweep transaction not found: {0}")]
     MissingSweepTransaction(bitcoin::Txid),
 
+    /// Indicates that a deposit request with the specified txid and vout could not be found.
+    #[error("deposit request not found: {0}, {1}")]
+    MissingDepositRequest(bitcoin::Txid, u32),
```
djordon, line 31: Use the `OutPoint` type here, it's simpler.

signer/tests/integration/stacks_events_observer.rs
```diff
@@ -415,6 +421,20 @@ async fn test_new_blocks_sends_update_deposits_to_emily() {
...
+    sweep.txid = deposit_completed_event.sweep_txid.into();
+    sweep.created_at_block_hash = deposit_completed_event.sweep_block_hash.into();
+    // A real sweep transaction would have the relevant deposits. For this test, we don't need them.
+    sweep.swept_deposits = vec![];
+    sweep.swept_withdrawals = vec![];
+    context
+        .get_storage_mut()
+        .write_sweep_transaction(&sweep)
+        .await
+        .expect("failed to insert dummy sweep transaction");
```
djordon, line 436: Can't this all be removed?

djordon (approved):
Looks good, had a few nits.

signer/src/error.rs
Jiloc: done in https://github.com/stacks-network/sbtc/pull/956/commits/c803d9d18bc48fc3aae449ca3c59233a26f44228

signer/tests/integration/stacks_events_observer.rs
Jiloc: yeah, I had to swap it with a deposit request 🤦 . Done in https://github.com/stacks-network/sbtc/pull/1005/commits/2b6e492f8f4a0868f541cf59b8f33bec06c25e67

signer/src/error.rs
Jiloc: changed in https://github.com/stacks-network/sbtc/pull/1005/commits/23b091cb0204cde75f0f5f0d96662a983b6febe0

djordon (approved):
Looks good! 🚢

Issue #1009, opened 2024-12-02
[Feature]: Endianness of `sweep_block_hash` in stored events is incorrect


djordon:
# Feature - Endianness of `sweep_block_hash` in stored events is incorrect

## 1. Description

The bytes of the `sweep_block_hash` column in the `completed_deposit_events` and `withdrawal_accept_events` tables are in reverse from how such fields are typically stored. We do not use that column currently, but we'll need to reverse the bytes before writing them going forward.

### 1.1 Context & Purpose

We should have a consistent endianness for all of the different hashes and IDs in the signer database. We cannot do joins otherwise.

## 2. Technical Details:

The `sweep_block_hash` column takes values directly from the webhook event emitted by the stacks node when an sBTC transaction is executed. The signers are the ones that originally populate the returned `sweep_block_hash` field, see https://github.com/stacks-network/sbtc/blob/0dd15d3868c31ef8efa567d95bdcc414f107b5b9/signer/src/stacks/contracts.rs#L325-L328

The `From<BitcoinBlockHash>` implementation for `BurnchainHeaderHash` reverses the underlying bytes, but we do not reverse them back when parsing the event. If we first parsed the bytes into a `BurnchainHeaderHash` and converted that to a `BitcoinBlockHash` then things would be fine.

Note that we need to do this for `CompletedDepositEvent` and `WithdrawalAcceptEvent`.

### 2.1 Acceptance Criteria:
   - [ ] We have store the correct endianness for `sweep_block_hash` values.

PR #1010, opened 2024-12-02
feat: store completed-deposit and withdrawal-accept events with consistent burnhash ordering

cylewitruk:
## Description

This was just a find while I was looking into SQL issues.

Closes: #1009 

This reverses the (reversed) byte order of the `burn-hash` event field for the `completed-deposit` and `withdrawal-accept` Stacks events, which is otherwise unusable in db joins. It doesn't seem that we use these fields once stored at the moment, but it's good if it's stored in a way which is easily usable.

What I'm not really sure of are potential implications with Emily, since it looks like we send the sweep block hash there as well, which is likely reversed atm, but also likely unused.

## Changes

- Adds a new function for reversing `BYTEA` fields. This is not an amazing function -- it's good for migrations, but I wouldn't use it for any larger queries.
- Migrates old `completed_deposit_events` entries, reversing the `sweep_block_hash` field so that it matches how blocks are stored in the rest of the database.
- Migrates old `withdrawal_accept_events` entries, reversing the `sweep_block_hash` field so that it matches how blocks are stored in the rest of the database.
- Reverses new entries for both events prior to storage so that they are stored in the same ordering as `bitcoin_blocks` etc.

## Testing Information

Have run the migration against a local copy of the db several times without issue. All integration tests pass, but I'm not sure that we use this field anywhere otherwise where it's expected in reversed format -- if we do then we don't test it.

## Checklist:


Jiloc (approved):
looks good!

cylewitruk:
I'll wait and see if we also want to adjust the withdrawal events in this PR as well

signer/src/stacks/events.rs
```diff
@@ -327,10 +327,13 @@ impl RawTupleData {
         let vout = self.remove_u128("output-index")?;
         let txid_bytes = self.remove_buff("bitcoin-txid")?;
         let mut sweep_block_hash = self.remove_buff("burn-hash")?;
-        sweep_block_hash.reverse();
         let sweep_block_height = self.remove_u128("burn-height")?;
         let sweep_txid = self.remove_buff("sweep-txid")?;
 
+        // The `sweep_block_hash` we receive is reversed, so we reverse it here
+        // so that we store it in an ordering consistent with the rest of our db.
```
djordon, line 334: I would reword this slightly. Something like: "The `bitcoin::BlockHash` type takes block hashes in big-endian format, however, the stacks node returns the `sweep_block_hash` in little-endian format. Before submitting the transaction that generated this event the signers reversed the endianness to conform to what the node expects."

djordon: Hmmm, actually scratch the above, I think we should handle this a little differently. I mean, we can merge this as is, but we should eventually change this. Instead the type for the `sweep_block_hash` should be a `stacks_common::types::chainstate::BurnchainHeaderHash`. When we go to write it to our database, we should convert it to a `crate::storage::models::BitcoinBlockHash`, which reverses the bytes for us.

I think this entire module should be moved to the sBTC crate so that other developers can use it for parsing sBTC events from their stacks nodes (not sure how useful it would be but that is the intention). It doesn't use any signer specific types and is generally devoid of signer specific logic. With that in mind, I think it would be best to remove any sbtc-signer specific logic here.

Moreover, using `BurnchainHeaderHash` feels the correct type, since we first convert our our `bitcoin::BlockHash` wrapper type into a `BurnchainHeaderHash` and use those bytes. So expecting a `BurnchainHeaderHash` on the other side feels more correct.

So it seems like a lot of hoops for all of this but it seems like the right thing to do. Thoughts?

djordon (commented):
Had some thoughts about the right way to handle this here, curious about what you think.

signer/src/stacks/events.rs
cylewitruk: Yeah I saw that you had mentioned this approach in #1009 after I had already done this.

My spontaneous thought was that since this is currently "signer" code, I didn't want anything/anyone to have to think about the fact that this is reversed, and the `BurnchainHeaderHash` is implicitly doing things that aren't obvious, so doing it here keeps at at the in/out boundary where it can easily be documented and internally then everything is just a regular old `*BlockHash`.

I'm not necessarily against using `BurnchainHeaderHash` either, especially if we move this stuff into a separate crate -- but if we do that then I might be for having an internal signer representation of the same structs which is how the _signer_ expects to work with things. What's most "correct" I think depends largely on the philosophy we want to take -- should discrepancies be resolved at the boundaries, or should they be propagated inwards.

djordon: Okay, we can merge this, I opened https://github.com/stacks-network/sbtc/issues/1016 for the move (I need to flesh it out).

Issue #960, opened 2024-11-26
[Tech Debt]: Introduce Cargo Feature Flag to Safeguard Incomplete Withdrawal Functionality


Jiloc:
## Description

PR that introduced tech-debt: #849 

Add a Cargo feature flag to disable the processing of withdrawal events by default. While we already prevent withdrawals from being processed [during validation](https://github.com/stacks-network/sbtc/blob/main/signer/src/bitcoin/validation.rs#L701), this additional safeguard ensures the functionality remains disabled unless explicitly enabled for development purposes. 
This prevents accidental activation due to unintentional code changes.

aldur:
> PR that introduced tech-debt: https://github.com/stacks-network/sbtc/pull/849

Did we add code for withdrawals in this PR, specifically?

Jiloc:
Not really, it’s the PR that added validation, so I probably shouldn’t have marked it specifically. I believe the (partial) withdrawal support itself was introduced across multiple PRs alongside the deposits functionality. For example, when we added support in the block_event to capture deposits prints events from the smart contracts, we also did the same for withdrawals. If you think it’s worth investigating further, I can compile a list of related PRs

matteojug:
Currently the withdrawal functionality (and access to the relative code paths) are prevent via:
 - There are no calls to `write_withdrawal_request`: we never (outside tests) call the function to store the withdrawal requests in the db.
 - Even with those requests in the db, they would fail bitcoin validation: we purposely fail validation for withdrawals tx

PR #970, opened 2024-11-27
[feature] Remove withdrawals related code from signer

MCJOHN974:
## Description
This PR ensures that signers don't post withdrawal related utxo's to bitcoin
Closes: #960

## Changes
Removed creating of withdrawal related utxo's.
Added tests for it

## Testing Information

Added a test which can be run with `cargo test --package signer --lib -- transaction_coordinator::tests::should_ignore_withdrawals --exact --show-output`

## Checklist:

- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #1011, opened 2024-12-02
[feature] Ignore withdrawals in transaction coordinator

MCJOHN974:
## Description

Follow up after #970
Closes: https://github.com/stacks-network/sbtc/issues/960
## Changes
See #970
## Testing Information
See #970
## Checklist:

### DIFF
--- a/signer/src/testing/transaction_coordinator.rs
+++ b/signer/src/testing/transaction_coordinator.rs
@@ -137,6 +137,7 @@ where
 {
     /// Asserts that TxCoordinatorEventLoop::get_pending_requests ignores withdrawals
     pub async fn assert_ignore_withdrawals(mut self) {
+        // Setup network and signer info
         let mut rng = rand::rngs::StdRng::seed_from_u64(46);
         let network = network::InMemoryNetwork::new();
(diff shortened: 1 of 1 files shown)


signer/src/testing/transaction_coordinator.rs
```diff
@@ -133,6 +135,81 @@ impl<Storage>
...
+            output: vec![bitcoin::TxOut {
+                value: bitcoin::Amount::from_sat(1_337_000_000_000),
+                script_pubkey: aggregate_key.signers_script_pubkey(),
+            }],
+            ..EMPTY_BITCOIN_TX
+        };
+        test_data.push_bitcoin_txs(
+            &bitcoin_chain_tip,
+            vec![(model::TransactionType::SbtcTransaction, tx_1.clone())],
+        );
```
djordon, line 160: We should add a comment above here about why this is necessary, which I think is that the coordinator needs to be able to find their UTXO when we call the `get_pending_requests` function below.

djordon (approved):
Looks good! I would add more comments to the tests, I left one specific example but add some comments that you think would be helpful for you two months from now if you had to look back.

Issue #1013, opened 2024-12-03
[Feature]: Expose tx-coordinator fields as configurable parameters


djordon:
# Feature - Expose tx-coordinator fields as configurable parameters

## 1. Description

There are a few fields in the `TxCoordinatorEventLoop` that we would like to modify in testnet (or maybe mainnet) without rebuilding the signer binary. These fields are:
1. `context_window` but with a default of 500.
2. `signing_round_max_duration`
3. `bitcoin_presign_request_max_duration`
4. `dkg_max_duration`

Each of these should be config parameters, similar to `bitcoin_processing_delay`, except these should be optional.

### 1.1 Context & Purpose

Since these parameters affect how well the sBTC signer protocol operates, we would like the be able to tailor them to our needs depending on the environment. Having them as config variables makes this makes it easy for us to quickly modify them. The down side is that these are parameters that we probably should not be changed from their defaults, so we should document that fact in our config and allow for them to be unspecified there as well.

## 2. Technical Details:

Config processing happens in `signer/src/config/mod.rs`, and the tests there read from `signer/src/config/default.toml`. Add new optional field(s) to the `SignerConfig` struct for these new parameters. The defaults should be same ones that are set in `main.rs` for the `TxCoordinatorEventLoop`.

### 2.1 Acceptance Criteria:
   - [ ] When running the binary, the fields in the `TxCoordinatorEventLoop` identified above get set from the values in config file if they are set there.

## 4. Appendix

It might make a lot of sense to group these parameters into a `[signer.coordinator]` field or something like that. But whatever we do we should make it very clear in the config documentation that the operator should be very careful when setting them. Also, when validating these values we must reject all zero durations.

PR #1014, opened 2024-12-03
[Feature]: Expose tx-coordinator fields as configurable parameters

MCJOHN974:
## Description

Closes: #1013

## Changes

Added requested parameters to signer config, switched from hardcoded parameters to using config while creating coordinator in `main.rs`

## Testing Information

Added 3 new tests in `signer/src/config/mod.rs::tests` and modified `signer/src/config/mod.rs::tests::default_config_toml_loads`

## Checklist:

### DIFF
--- a/signer/Cargo.toml
+++ b/signer/Cargo.toml
@@ -76,7 +76,8 @@ ripemd = "0.1.3"
 sbtc = { path = "../sbtc", default-features = false, features = ["testing"] }
 test-case = "3.1"
 test-log = { version = "0.2.16", default-features = false, features = ["trace"] }
-
+toml_edit = "0.22.22"
+tempfile = "3.6" 
 
 [build-dependencies]
 tonic-build.workspace = true
--- a/signer/src/config/default.toml
+++ b/signer/src/config/default.toml
@@ -143,7 +143,32 @@ bootstrap_signatures_required = 2
 # Seconds to wait before processing a new Bitcoin block.
 # Required: true Environment: SIGNER_SIGNER__BITCOIN_PROCESSING_DELAY
 # TODO(715): Add default delay (15 seconds? see #701)
-bitcoin_processing_delay = 0
+bitcoin_processing_delay = 15
+
+# How many bitcoin blocks back from the chain tip the signer will
+# look for requests.
+# Optional.
+# Warning: this parameter may significantly affect how well the sBTC signer protocol operates
+# change it carefully
+context_window = 500
+# The maximum duration of a signing round before the coordinator will
+# time out and return an error.
+# Optional.
+# Warning: this parameter may significantly affect how well the sBTC signer protocol operates
+# change it carefully
+signer_round_max_duration = 30
+# The maximum duration of a pre-sign request before the coordinator will
+# time out and start sending the requests to the signers.
+# Optional.
+# Warning: this parameter may significantly affect how well the sBTC signer protocol operates
+# change it carefully
+bitcoin_presign_request_max_duration = 30
+# The maximum duration of distributed key generation before the
+# coordinator will time out and return an error.
+# Optional.
+# Warning: this parameter may significantly affect how well the sBTC signer protocol operates
+# change it carefully
+dkg_max_duration = 120
 
 # !! ==============================================================================
 # !! Stacks Event Observer Configuration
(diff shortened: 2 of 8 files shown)


signer/Cargo.toml
```diff
@@ -76,7 +76,8 @@ ripemd = "0.1.3"
 sbtc = { path = "../sbtc", default-features = false, features = ["testing"] }
 test-case = "3.1"
 test-log = { version = "0.2.16", default-features = false, features = ["trace"] }
-
+toml_edit = "0.22.22"
+tempfile = "3.6"
```
djordon, line 80: Why are these necessary?

signer/src/config/default.toml
```diff
@@ -145,6 +145,31 @@ bootstrap_signatures_required = 2
 # TODO(715): Add default delay (15 seconds? see #701)
 bitcoin_processing_delay = 0
 
+# How many bitcoin blocks back from the chain tip the signer will
+# look for requests.
+# Optional.
+# Warning: this parameter may significantly affect how well the sBTC signer protocol operates
+# change it carefully
```
djordon, line 152: You should use the same style of formatting as the other fields in this file. Also you should detail how to set them from the environment and which values are invalid.

signer/src/config/mod.rs
```diff
@@ -328,6 +365,12 @@ impl Settings {
             .prefix_separator("_");
 
         let mut cfg_builder = Config::builder();
+
+        cfg_builder = cfg_builder.set_default("signer.dkg_max_duration", 120)?;
+        cfg_builder = cfg_builder.set_default("signer.bitcoin_presign_request_max_duration", 30)?;
+        cfg_builder = cfg_builder.set_default("signer.signer_round_max_duration", 30)?;
+        cfg_builder = cfg_builder.set_default("signer.context_window", 500)?;
```
djordon, line 372: Have you tried commenting them out in the file and seeing what happens? Also, I think you can mark them as optional fields in the struct and that might work as well. But then we have to deal with the default value so maybe this is clearer.

djordon (commented):
Largely looks good. Had a few comments

signer/src/config/default.toml
djordon: This basically good, but I think that we should note that any value less than 1 is invalid.

signer/src/config/mod.rs
```diff
@@ -596,6 +678,61 @@ mod tests {
...
+        assert_eq!(settings.signer.dkg_max_duration, Duration::from_secs(120));
+    }
+
+    #[test]
+    fn zero_durations_fails_in_signer_config() {
+        fn test_one(field: &str) {
+            clear_env();
+            std::env::set_var(format!("SIGNER_SIGNER__{}", field.to_uppercase()), "0");
+            let _ = Settings::new_from_default_config().expect_err(&format!(
+                "Duration for {field} dkg_max_duration must be non zero"
```
djordon, line 728:
```suggestion
                "Duration for {field} must be non zero"
```

signer/src/config/default.toml
```diff
@@ -145,6 +145,27 @@ bootstrap_signatures_required = 2
...
+# look for requests.
+# Required: false
+# Environment: SIGNER_SIGNER__CONTEXT_WINDOW
+context_window = 500
+# The maximum duration of a signing round before the coordinator will
+# time out and return an error.
+# Required: false
+# Environment: SIGNER_SIGNER__SIGNER_ROUND_MAX_DURATION
+signer_round_max_duration = 30
+# The maximum duration of a pre-sign request before the coordinator will
```
djordon, line 158: Oh, please add a new line between each parameter and the comments for the next parameter.

djordon (commented):
We need to update `docker/sbtc/signer/signer-config.toml` and `docker/prodlike/config/signer-config.toml` too. We can comment out the parameters in those configs so that the defaults get used.

djordon:
> > We need to update docker/sbtc/signer/signer-config.toml and docker/prodlike/config/signer-config.toml too. We can comment out the parameters in those configs so that the defaults get used.
> 
> As I see this files don't have parameters for this, thus I understand your suggestion as "introduce this parameters to this files". Since they are optional, from point of view "code should work" code will work without adding it. From point of view "code should be readable" I agree that it will be easier to edit this parameters if you already have them in config file, but... why then we made them optional?
> 
> @djordon

Making them optional makes it very easy for the signer operator to say "use the default" without the operator having to look up what the default should be. If we make it required, the operator needs to look up the right value for the environment that they are operating in (which is either mainnet, testnet, or devnet). But, if they are optional we can let the binary figure out what the right default value should be (the code here has a static default for all environments/networks, but that doesn't need to be the case).

djordon (approved):

Issue #1007, opened 2024-12-01
[Feature]: The coordinator shouldn't bail so easily when processing new blocks


djordon:
# Feature - The coordinator shouldn't bail so easily when processing new blocks

## 1. Description

The main function where the coordinator does all of it's work is in the `TxCoordinatorEventLoop::process_new_blocks` function. Right now it bails if the call to `construct_and_sign_bitcoin_sbtc_transactions` results in an error but it shouldn't. Instead it should continue to `construct_and_sign_stacks_sbtc_response_transactions`, since there may be stacks  transactions that need to be generated and submitted.

### 1.1 Context & Purpose

An error on the bitcoin side of things shouldn't affect the coordinator from doing it's duties on the stacks side of things.

## 2. Technical Details:

In `TxCoordinatorEventLoop::process_new_blocks`, we should catch any `Err`, log it, and continue to `TxCoordinatorEventLoop::construct_and_sign_stacks_sbtc_response_transactions`

### 2.1 Acceptance Criteria:
   - [ ] An error in `TxCoordinatorEventLoop::construct_and_sign_bitcoin_sbtc_transactions` should not stop the calls to `TxCoordinatorEventLoop::construct_and_sign_stacks_sbtc_response_transactions`.

PR #1015, opened 2024-12-03
feat: do not bail when coordinator fails to create bitcoin transactions

Jiloc:
Closes: #1007

## Changes
- Modify `process_new_blocks` to catch and log errors coming from `construct_and_sign_bitcoin_sbtc_transactions` so that it can continue processing stacks transactions.

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -324,12 +324,16 @@ where
         self.check_and_submit_rotate_key_transaction(&bitcoin_chain_tip, &aggregate_key)
             .await?;
 
-        self.construct_and_sign_bitcoin_sbtc_transactions(
-            &bitcoin_chain_tip,
-            &aggregate_key,
-            &signer_public_keys,
-        )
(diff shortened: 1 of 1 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -324,12 +324,16 @@ where
...
-        )
-        .await?;
+        if let Err(error) = self
+            .construct_and_sign_bitcoin_sbtc_transactions(
+                &bitcoin_chain_tip,
+                &aggregate_key,
+                &signer_public_keys,
+            )
+            .await
+        {
```
djordon, line 334: nit: can we restyle this to something like
```rust
let bitcoin_processing_fut = self
    .construct_and_sign_bitcoin_sbtc_transactions(
        &bitcoin_chain_tip,
        &aggregate_key,
        &signer_public_keys,
    );

if let Err(error) = bitcoin_processing_fut.await {
    ...
```
I find it much easier to read.

djordon (approved):

Issue #1016, opened 2024-12-03
[Feature]: Move sBTC event module to the sbtc crate


djordon:
# Feature - Move sBTC event module to the sbtc crate

## 1. Description

The `signer/src/stacks/events.rs` module was meant to live else where.

### 2.1 Acceptance Criteria:
   - [ ] Criterion 1
   - [ ] Criterion 2
   - [ ] Criterion 3

djordon:
Hey @MCJOHN974, I just took another look at the `events` module and there are a few additional changes that I think we should make when we move that module to the `sbtc` crate. They are
1. Create a local `StacksTxId` type that wraps a 32 byte array and use that instead of the `blockstack_lib::burnchains::Txid` type. This allows us to not require stackslib as a dependency of the sbtc crate.
2. Use `u128` as the `signer_bitmap` type instead of the `BitArray` type. This allows us to remove the bitvec dependency from the `sbtc` crate.

On the signer side of things we'll convert the new `StacksTxId` type in the sbtc crate into the signer's `StacksTxId` type. Similarly, the signer versions of the event types should use `BitArray<[u8; 32]>` as the type.

PR #1027, opened 2024-12-04
[feature] Move sBTC event module to sbtc crate

MCJOHN974:
## Description
WIP
Closes: #1016

## Checklist:

### DIFF
--- a/signer/src/testing/dummy.rs
+++ b/signer/src/testing/dummy.rs
@@ -311,7 +311,7 @@ impl fake::Dummy<fake::Faker> for WithdrawalAcceptEvent {
                 vout: rng.next_u32(),
             },
             fee: rng.next_u32() as u64,
-            sweep_block_hash: block_hash(config, rng),
+            sweep_block_hash: block_hash(config, rng).into(),
             sweep_block_height: rng.next_u32() as u64,
             sweep_txid: txid(config, rng),
         }
@@ -322,7 +322,7 @@ impl fake::Dummy<fake::Faker> for WithdrawalRejectEvent {
     fn dummy_with_rng<R: Rng + ?Sized>(config: &fake::Faker, rng: &mut R) -> Self {
         let bitmap = rng.next_u64() as u128;
         WithdrawalRejectEvent {
-            txid: StacksTxid(config.fake_with_rng(rng)),
+            txid: config.fake_with_rng::<[u8; 32], _>(rng).into(),
             block_id: stacks_common::types::chainstate::StacksBlockId(config.fake_with_rng(rng)),
             request_id: rng.next_u32() as u64,
             signer_bitmap: bitmap,


signer/src/stacks/events.rs
djordon, line 1: Hmmm, after moving the new conversion `impl`s, can't we delete this file and use the sbtc version here in this crate?

djordon: No. I mean can't we delete this file? Doing so would mean that some functions will be missing, and we'd have to import them form the sbtc crate.

Deleting this file would also mean deleting the new code that you added so you should first move the new code. You should also move the local versions of the structs that we will want to keep.

sbtc/src/events.rs
```diff
@@ -25,13 +25,19 @@ use clarity::vm::types::TupleData;
 use clarity::vm::ClarityName;
 use clarity::vm::Value as ClarityValue;
 use secp256k1::PublicKey;
-use stacks_common::impl_byte_array_newtype;
+use stacks_common::*;
```
djordon, line 28: I don't think that we should use `*` imports.

sbtc/src/events.rs
```diff
@@ -11,24 +11,33 @@
...
+/// Stacks transaction identifier. Wrapper over a 32 byte array.
+pub struct StacksTxid(pub [u8; 32]);
+// impl_byte_array_newtype!(StacksTxid, u8, 32);
+
+
+impl_array_newtype!(StacksTxid, u8, 32);
+impl_array_hexstring_fmt!(StacksTxid);
+impl_byte_array_newtype!(StacksTxid, u8, 32);
+impl_byte_array_message_codec!(StacksTxid, 32);
+impl_byte_array_serde!(StacksTxid);
```
djordon, line 40: Do we need all of these? I'm under the impression that we don't, and that we can just have the `StacksTxid(pub [u8; 32])` type. Maybe just derive `Copy, Clone, Debug`.

sbtc/src/events.rs
```diff
@@ -350,8 +355,9 @@ impl RawTupleData {
                 // that gets emitted here.
                 vout: u32::try_from(vout).map_err(EventError::ClarityIntConversion)?,
             },
-            sweep_block_hash: BitcoinBlockHash::from_slice(&sweep_block_hash)
-                .map_err(EventError::ClarityHashConversion)?,
+            // TODO: I don't like this unwrap()
+            sweep_block_hash: BurnchainHeaderHash::from_bytes(&sweep_block_hash).unwrap(),
+            //.map_err(EventError::ClarityHashConversion)?,
```
djordon, line 360: Let's map the error. If `EventError` doesn't have a variant for the returned error then add one.

signer/src/storage/model.rs
```diff
@@ -1195,6 +1195,120 @@ pub struct BitcoinWithdrawalOutput {
...
+    /// withdrawal request.
+    pub sweep_txid: BitcoinTxid,
+}
+
+impl From<sbtc::events::WithdrawalAcceptEvent> for WithdrawalAcceptEvent {
+    fn from(sbtc_event: sbtc::events::WithdrawalAcceptEvent) -> WithdrawalAcceptEvent {
+        let mut reversed_raw_hash = *sbtc_event.sweep_block_hash.as_bytes();
+        reversed_raw_hash.reverse();
+        let raw_hash = Hash::from_byte_array(reversed_raw_hash);
+        let sweep_hash = BitcoinBlockHash::from_raw_hash(raw_hash);
```
djordon, line 1254: You can replace this with:
```suggestion
        let sweep_hash = BitcoinBlockHash::from(sbtc_event.sweep_block_hash);
```

signer/src/storage/model.rs
```diff
@@ -1195,6 +1195,120 @@ pub struct BitcoinWithdrawalOutput {
...
+    /// deposit.
+    pub sweep_txid: BitcoinTxid,
+}
+
+impl From<sbtc::events::CompletedDepositEvent> for CompletedDepositEvent {
+    fn from(sbtc_event: sbtc::events::CompletedDepositEvent) -> CompletedDepositEvent {
+        let mut reversed_raw_hash = *sbtc_event.sweep_block_hash.as_bytes();
+        reversed_raw_hash.reverse();
+        let raw_hash = Hash::from_byte_array(reversed_raw_hash);
+        let sweep_hash = BitcoinBlockHash::from_raw_hash(raw_hash);
```
djordon, line 1298: Same here, you can replace this with:
```suggestion
        let sweep_hash = BitcoinBlockHash::from(sbtc_event.sweep_block_hash);
```

signer/src/storage/model.rs
```diff
@@ -1195,6 +1195,120 @@ pub struct BitcoinWithdrawalOutput {
...
+use bitcoin::Txid as BitcoinTxid;
+use bitcoin::WitnessProgram;
+use bitcoin::WitnessVersion;
+
+use blockstack_lib::burnchains::Txid as StacksTxid;
+use clarity::vm::types::CharType;
+use clarity::vm::types::SequenceData;
+use clarity::vm::types::TupleData;
+use clarity::vm::ClarityName;
+use clarity::vm::Value as ClarityValue;
```
djordon, line 1215: Imports go at the top of the file.

signer/src/storage/model.rs
djordon, line 1: We also need structs for the other event types.

djordon (commented):
This largely looks good, but we need more structs and their conversions. Also, I would imagine that things don't compile.

Issue #985, opened 2024-11-28
[Feature]: (Partial) Use of `usize` can cause state divergence between signers 


technovision99:
## 1. Description

The use of `usize` in the following function: https://github.com/stacks-network/sbtc/blob/f48d6fe506b06d426cceb8fba5cce21829938a3e/signer/src/transaction_coordinator.rs#L1577-L1597 is not necessary (instead it could just be set to a `u32` to accommodate both 32 and 64 bit archs)  and could cause different signers to process who the coordinator is differently. This could lead to signers being completely out of sync and valid messages being dropped/unsigned.

cylewitruk:
I guess that `u64` shouldn't be a problem either given emulation and 8 bytes would keep the behavior we've been testing with, are you thinking something like this instead (given u64)?

```rust
let index = u64::from_be_bytes(digest[0..8].try_into().unwrap())
```

technovision99:
That should be good, since the use of `u32` and the first 4 bytes of the hash are deterministic among both 32 and 64 bit signers.

PR #1029, opened 2024-12-04
fix: always use 32-bit math for determining coordinator index

cylewitruk:
## Description

Closes: #985

## Changes

Changes the `coordinator_public_key()` function of the coordinator to always use 32-bit math instead of `usize` (which can either be 32- or 64-bit) when determining the current coordinator's public key from bitcoin block hash.

## Testing Information

All existing tests pass.

## Checklist:

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -1620,7 +1620,8 @@ where
     }
 }
 
-/// Check if the provided public key is the coordinator for the provided chain tip
+/// Check if the provided public key is the coordinator for the provided chain
+/// tip
 pub fn given_key_is_coordinator(
     pub_key: PublicKey,
     bitcoin_chain_tip: &model::BitcoinBlockHash,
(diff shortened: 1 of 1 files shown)


djordon (approved):
Looks good ✅

signer/src/transaction_coordinator.rs
```diff
@@ -1634,21 +1635,28 @@ pub fn coordinator_public_key(
...
-    // Also, do humans even make machines where the pointer width is
-    // greater than 32 bytes?
-    let index = usize::from_be_bytes(*digest.first_chunk()?);
+
+    // Use the first 4 bytes of the digest to create a u32 index. Since `digest`
+    // is 32 bytes and we explicitly take the first 4 bytes, this is safe.
+    #[allow(clippy::expect_used)]
+    let u32_bytes = digest[..4]
+        .try_into()
+        .expect("BUG: failed to take first 4 bytes of digest");
```
matteojug, line 1649: We have three copies of the old `usize::from_be_bytes(*digest.first_chunk().expect()` around in test code, we should update those too (eventually reusing the same code)

djordon: Looks like we may need to in order for tests to pass.

cylewitruk: Yeah was just looking at that, the process_complete_deposit messes up

cylewitruk: @djordon @matteojug pushed fixes btw

djordon (approved):

matteojug (approved):

Issue #1036, opened 2024-12-05
[Feature]: Restrict Nodes that can perform a Reorg


AshtonStephens:
# Feature - Restrict Nodes that can perform a Reorg

## 1. Description

Restrict Nodes that can perform a Reorg to stabilize system during onboarding.

### 1.1 Context & Purpose
When sbtc signers have been spinning up they've sent bad information to the emily API, which has caused it to wig out with the bootup of the signer. This ticket is to add a temporary feature to only trust a single sBTC signer, specified by an api key, that can initiate a reorg.

### 2.1 Acceptance Criteria:
   - [ ] An api key can be specified in the cdk template that has unique power to intiate a reorg
   - [ ] Set chainstate calls provide an OK response for all cases that would have done a reorg in the past but actually aren't performing the reorg
   - [ ] feature is pr-ed, and merged

PR #1037, opened 2024-12-05
Allow reorg with special api key

AshtonStephens:
## Description

Closes: #1036 

## Changes

Adds a new env variable to the AWS lambda to pick up the api key that is allowed to perform reorgs. This will make it such that only a specific trusted api key can cause a reorg to mess up the chainstate.

This was done by adding a new parameter to all the functions that can initiate a reorg to detect the api key and check it against the approved api key.

## Testing Information

This was tested with a deployed version of the emily api in a non-production account. It was first tested with a disallowed api key on the chainstate integration tests and ONLY the reorg tests failed (great) and then tested with the allowed api key and the tests succeeded (great).

## Checklist:


emily/handler/src/api/handlers/chainstate.rs
```diff
@@ -171,22 +180,28 @@ pub async fn update_chainstate(
...
+                    canonical_tip: chainstate.clone(),
+                    conflicting_chainstates,
+                };
+                // Execute the reorg.
+                execute_reorg_handler(context, execute_reorg_request)
+                    .await
+                    .inspect_err(|e| warn!("Failed executing reorg with error {}", e))?;
+            // Log error.
+            } else {
+                info!("Inconsistent chainstate found for {entry:?} but we pretend it's okay.");
```
cylewitruk, line 203: I'd probably change this to like trace or debug since it'll probably happen a lot?

cylewitruk (approved):
I think this looks okay, just one comment re: logging (so it's easier to filter out this stuff), but no need to fix it.

As mentioned in other channels, I don't think we should use env vars for the trusted api key if this "fix" were to live long enough for a mainnet release, just because env vars have a tendency to "leak".

emily/handler/src/api/handlers/chainstate.rs
AshtonStephens: Okay

AshtonStephens: Done.

PR #1048, opened 2024-12-05
feat: remove need of context window for signers utxo

### LINKED ISSUE #1004: [Feature]: Remove dependency of context window for fetching the signers' UTXO
# Feature - Remove dependency of context window for fetching the signers' UTXO

## 1. Description

We use a context window for fetching the signers' UTXO, which means we only look so far back for it when trying to find it. The issue is that if the signers do not have any activity for `context_window` number of blocks, then the signers will not find their UTXO when trying to sweep any funds. Let's change that to make finding the signers UTXO independent of the context window.

### 1.1 Context & Purpose

We should always be able to find out what the signers' UTXO is, no matter what. Let's make it so.

### 2.1 Acceptance Criteria:
   - [ ] The context window is not an input for finding the signers' UTXO.

* Kinda related https://github.com/stacks-network/sbtc/issues/993

## 4. Appendix

The issue for a simple fix is that we do not know which outputs are truly unspent without identifying a blockchain, and when identifying a block chain we usually need some limiting criteria for when to stop looking for blocks. We an use a heuristic to get a minimum block height when looking for the UTXO. If we assume that reorgs only affect a maximum of, say, 10 blocks then we can find a minimum search height. We would just have to look at the block height of our most recent TXO minus 10 blocks to compute a minimum search height when looking for the UTXO.

djordon:
## Description

Closes #1004 

## Changes

* Add a new table valued function for getting the bitcoin blockchain until a certain height.
* Change the `get_signer_utxo` to not require the context window.

## Testing Information

I added integration tests, but they need comments.

## Checklist:

### DIFF
--- a/signer/migrations/0007__create_bitcoin_blockchain_from.sql
+++ b/signer/migrations/0007__create_bitcoin_blockchain_from.sql
@@ -1,0 +1,43 @@
+-- Table-Valued Function (TVF) for fetching a Bitcoin blockchain from a given
+-- block hash, only fetching blocks to a specific height.
+--
+-- - chain_tip: The block hash to start the blockchain from ("chain tip").
+-- - min_block_height: The minimum height of all blocks that are returned.
+CREATE FUNCTION sbtc_signer.bitcoin_blockchain_from (
+    chain_tip BYTEA,
+    min_block_height INT
+)
+RETURNS TABLE (
+    block_hash BYTEA,
+    parent_hash BYTEA,
+    block_height BIGINT
+) 
+AS $$
+BEGIN
+    RETURN QUERY
+    WITH RECURSIVE blockchain AS (
+        SELECT
+            blocks.block_hash
+          , blocks.parent_hash
+          , blocks.block_height
+        FROM sbtc_signer.bitcoin_blocks as blocks
+        WHERE blocks.block_hash = chain_tip
+
+        UNION ALL
+
+        SELECT
+            parent.block_hash
+          , parent.parent_hash
+          , parent.block_height
+        FROM sbtc_signer.bitcoin_blocks AS parent
+        JOIN blockchain AS last
+          ON parent.block_hash = last.parent_hash
+        WHERE last.block_height >= min_block_height
+    )
+    SELECT
+        block_hash
+      , parent_hash
+      , block_height
+    FROM blockchain;
+END;
+$$ LANGUAGE plpgsql;
--- a/signer/migrations/0007__create_bitcoin_blockchain_from.sql
+++ b/signer/migrations/0007__create_bitcoin_blockchain_from.sql
@@ -35,9 +35,9 @@ BEGIN
         WHERE last.block_height >= min_block_height
     )
     SELECT
-        block_hash
-      , parent_hash
-      , block_height
-    FROM blockchain;
+        blocks.block_hash
+      , blocks.parent_hash
+      , blocks.block_height
+    FROM blockchain as blocks;
 END;
 $$ LANGUAGE plpgsql;
--- a/signer/migrations/0007__create_bitcoin_blockchain_from.sql
+++ b/signer/migrations/0007__create_bitcoin_blockchain_from.sql
@@ -1,5 +1,5 @@
 -- Table-Valued Function (TVF) for fetching a Bitcoin blockchain from a given
--- block hash, only fetching blocks to a specific height.
+-- block hash, only fetching blocks down to a specific height.
 --
 -- - chain_tip: The block hash to start the blockchain from ("chain tip").
 -- - min_block_height: The minimum height of all blocks that are returned.
@@ -32,7 +32,7 @@ BEGIN
         FROM sbtc_signer.bitcoin_blocks AS parent
         JOIN blockchain AS last
           ON parent.block_hash = last.parent_hash
-        WHERE last.block_height >= min_block_height
+        WHERE last.block_height > min_block_height
     )
     SELECT
         blocks.block_hash
--- a/signer/migrations/0007__create_bitcoin_blockchain_from.sql
+++ b/signer/migrations/0007__create_bitcoin_blockchain_from.sql
@@ -3,7 +3,7 @@
 --
 -- - chain_tip: The block hash to start the blockchain from ("chain tip").
 -- - min_block_height: The minimum height of all blocks that are returned.
-CREATE FUNCTION sbtc_signer.bitcoin_blockchain_from (
+CREATE FUNCTION sbtc_signer.bitcoin_blockchain_until (
     chain_tip BYTEA,
     min_block_height BIGINT
 )
--- a/signer/migrations/0007__create_bitcoin_blockchain_from.sql
+++ b/signer/migrations/0007__create_bitcoin_blockchain_from.sql
@@ -5,7 +5,7 @@
 -- - min_block_height: The minimum height of all blocks that are returned.
 CREATE FUNCTION sbtc_signer.bitcoin_blockchain_from (
     chain_tip BYTEA,
-    min_block_height INT
+    min_block_height BIGINT
 )
 RETURNS TABLE (
     block_hash BYTEA,
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -415,22 +415,66 @@ impl PgStore {
         Ok(pg_utxo.map(SignerUtxo::from))
     }
 
+    /// Return a block height of the donation UTXOs that have been
+    /// confirmed on the block of least height and remain unspent.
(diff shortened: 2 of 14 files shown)


signer/src/storage/postgres.rs
```diff
@@ -403,15 +409,109 @@ impl PgStore {
...
     }
 
+    /// Return a block height that is less than or equal to the block that
+    /// confirms the signers' UTXO.
+    ///
+    /// The signers outstanding UTXO is a UTXO that is confirmed on the
+    /// canonical bitcoin blockchain. But regardless of which blockchain
+    /// that is, we know that it is very unlikely to have been affected by
+    /// a reorg of more than 10 blocks deep. That means that if we look back
+    /// at least 10 blocks before the best candidate for the signers' UTXO,
```
matteojug, line 428: nit
```suggestion
    /// a reorg of more than `MAX_REORG_BLOCK_COUNT` blocks deep. That means that if we look back
    /// at least `MAX_REORG_BLOCK_COUNT` blocks before the best candidate for the signers' UTXO,
```

signer/src/storage/postgres.rs
```diff
@@ -403,15 +409,109 @@ impl PgStore {
...
+    /// we are very likely to find a transaction output that is on the best
+    /// bitcoin blockchain.
+    pub async fn minimum_utxo_height(
+        &self,
+        output_type: model::TxOutputType,
+    ) -> Result<Option<i64>, Error> {
+        // Get the block height of the unspent transaction that was most
+        // recently confirmed. Note that we are not filtering by the
+        // blockchain identified by a chain tip, we just want the UTXO with
+        // maximum height, event if it has been reorged.
```
matteojug, line 438:
```suggestion
        // maximum height, even if it has been reorged.
```

signer/src/storage/postgres.rs
```diff
@@ -403,15 +409,109 @@ impl PgStore {
...
+    /// confirms the signers' UTXO.
+    ///
+    /// The signers outstanding UTXO is a UTXO that is confirmed on the
+    /// canonical bitcoin blockchain. But regardless of which blockchain
+    /// that is, we know that it is very unlikely to have been affected by
+    /// a reorg of more than 10 blocks deep. That means that if we look back
+    /// at least 10 blocks before the best candidate for the signers' UTXO,
+    /// we are very likely to find a transaction output that is on the best
+    /// bitcoin blockchain.
+    pub async fn minimum_utxo_height(
```
matteojug, line 431: I'm wondering if this _pick one, then pick its parent_ is enough: if we have a >10 blocks without activity, then a deposit and two consecutive sweeps (spanning a total of 3 blocks), and those 3 blocks are reorged invalidating the deposit, we will lose track of the original utxo. Is it correct?

About how we may end up in that, we could assume we have the deposit input to be using one of two possible spending path, and in the competing branch (that eventually becomes the canonical one) a tx is spending that same output with another spending path. All of this to say, should we use the buffer `MAX_REORG_BLOCK_COUNT` to constraint `prev_confirmed_height_candidate`?

signer/migrations/0007__create_bitcoin_blockchain_from.sql
```diff
@@ -0,0 +1,43 @@
+-- Table-Valued Function (TVF) for fetching a Bitcoin blockchain from a given
+-- block hash, only fetching blocks to a specific height.
```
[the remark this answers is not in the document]
cylewitruk: or, down to? :)

cylewitruk (commented):
Could we also add the following indexes as we're filtering and ordering on these columns?

```
CREATE INDEX ix_bitcoin_tx_inputs_prevout_type ON sbtc_signer.bitcoin_tx_inputs(prevout_type);
CREATE INDEX ix_bitcoin_tx_ouputs_ouput_type ON sbtc_signer.bitcoin_tx_ouputs(output_type);
CREATE INDEX ix_bitcoin_blocks_block_height ON sbtc_signer.bitcoin_blocks(block_height);
```

djordon:
> Could we also add the following indexes as we're filtering and ordering on these columns?
> 
> ```
> CREATE INDEX ix_bitcoin_tx_inputs_prevout_type ON sbtc_signer.bitcoin_tx_inputs(prevout_type);
> CREATE INDEX ix_bitcoin_tx_ouputs_ouput_type ON sbtc_signer.bitcoin_tx_ouputs(output_type);
> CREATE INDEX ix_bitcoin_blocks_block_height ON sbtc_signer.bitcoin_blocks(block_height);
> ```

Sure thing

signer/src/storage/postgres.rs
djordon: Oh you're right. And I don't think that using the buffer on `prev_confirmed_height_candidate` is enough either. I was really hoping to not go back up the transaction TXO chain but it looks like that is what we need to do.

Basically, we need to start from the UTXO with greatest height and go back at least 10 blocks (which may be more than 10 transactions). It's annoying and might be complex but it's what we should do.

Also, I need to add an additional test for the case that you just spoke about.

matteojug: Oh right, also if we don't actually walk up the transactions chain we could jump from a reorged to a different reorged branch, still missing a canonically confirmed sweep tx.

djordon: I updated the logic in https://github.com/stacks-network/sbtc/pull/1048/commits/71d852f97bdf9b3da7d24e195c37990287bfb7d0. I'm also going to take another look at the logic (and comments) when I get some fresh eyes. There may be another approach that is simpler.

signer/src/storage/postgres.rs
```diff
@@ -1566,14 +1686,34 @@ impl super::DbRead for PgStore {
...
+
+        // We do not have a signer output. Maybe we have a donation, so
+        // let's try that.
+        let output_type = model::TxOutputType::Donation;
+        let Some(min_block_height) = self.minimum_utxo_height(output_type).await? else {
+            return Ok(None);
+        };
+
+        self.get_utxo(chain_tip, output_type, min_block_height)
+            .await
```
djordon, line 1716: The logic for fetching donations needs some work still (it's "wrong" here). But the stuff above should be correct.

djordon: I think the logic is correct after https://github.com/stacks-network/sbtc/pull/1048/commits/c6c914f39252f59c3c3715f53cabfc5b0ec826dd. Now for more tests

djordon: I just realized that the above was wrong. Fixed in https://github.com/stacks-network/sbtc/pull/1048/commits/a5fdb3f05665e78e5de1c319fb0ea57464319228... let's hope that this is the last fix.

Also, I cannot help but think that there is a simpler way to get the signers' UTXO. Any ideas?

signer/src/storage/postgres.rs
```diff
@@ -446,68 +488,120 @@ impl PgStore {
...
+                    bti.txid
+                  , bti.prevout_txid
+                  , MIN(bb.block_height) AS block_height
+                FROM sbtc_signer.bitcoin_tx_inputs AS bti
+                JOIN sbtc_signer.bitcoin_transactions USING (txid)
+                JOIN sbtc_signer.bitcoin_blocks AS bb USING (block_hash)
+                WHERE bti.prevout_type = 'signers_input'
+                  AND bb.block_height <= $1
+                GROUP BY bti.txid, bti.prevout_txid
+                LIMIT $2
```
matteojug, line 561: I'm not sure about `LIMIT $2` (`max_transactions`)  here: we may have unrelated (wrt the tx chain we want to walk, eg in case of forks) txs in here that "consume" that limit, and in the worst case scenario where we have actually need to walk `max_transactions` to find one with the block height less than the threshold we may not have all of them in `signer_inputs`.

djordon: Yup, we spoke offline, and this is still an issue.

matteojug (approved):

signer/src/storage/postgres.rs
```diff
@@ -418,25 +415,70 @@ impl PgStore {
...
+    /// # Notes
+    ///
+    /// * This function only returns `Ok(None)` if there have been no
+    ///   confirmed sweep transactions.
+    /// * As the signers sweep funds between BTC and sBTC, they leave a
+    ///   chain of transactions, where each transaction spends the signers'
+    ///   sole UTXO and creates a new one. This function "crawls" the chain
+    ///   of transactions, starting at the most recently confirmed one,
+    ///   until it goes back at least [`MAX_REORG_BLOCK_COUNT`] blocks
+    ///   worth of transactions. A blocks with height greater than or equal
```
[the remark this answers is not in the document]
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1048/commits/b7f797d352ed877dce1b6534903d7579fbf3d811.

PR #1056, opened 2024-12-06
feat:do not update chainstate on deposit and withdrawal updates

Jiloc:
## Description

We don't need to start a chainstate update from the udpate_deposits and withdrawals endpoints. They were originally intended to prevent update messages to be rejected by emily if she didn't know the relative chainstate yet. 

 
## Changes
- remove the calls to `inferred_chainstates` from the withdrawals and deposits update endpoints.
- remove linked tests


Jiloc:
We have opted to retain this feature. Singers will not have permission to initiate a reorganization. However, if they send an update for a deposit or withdrawal related to a stacks block that the sidecar has not yet transmitted to Emily, Emily will append the unseen block to the chain state. (provided it does not trigger a reorg)

Issue #1026, opened 2024-12-04
[Bug] Cannot fetch token total supply before the token exists


matteojug:
If we have a `total_cap` limit in place, but we still need to deploy the contracts, we cannot proceed because when we call `update_sbtc_limits` in block observer it will then query the smart contract for the total supply, and this call will fail with something like
```
{"okay":false,"cause":"Unchecked(NoSuchContract(\"SN1PM3B00AKVTVT5QMS0XZZ0VERETABE4PHY4TVC8.sbtc-token\"))"}
```
since the token smart contract doesn't exists yet; then we `continue`, without sending the `BitcoinBlockObserved` event, thus never waking up the coordinator.

PR #1058, opened 2024-12-07
fix: don't ask for total supply if token doesn't exist

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/1026

## Changes

In block observer `update_sbtc_limits`, check token total supply only if we have a total cap and the token contract exists.

Also remove some duplicated code in favor of a dummy const.

## Testing Information

Added an integration test checking that block observer fully process a block in the different scenarios.

## Checklist:

### DIFF
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -30,6 +30,7 @@ use crate::emily_client::EmilyInteract;
 use crate::error::Error;
 use crate::stacks::api::StacksInteract;
 use crate::stacks::api::TenureBlocks;
+use crate::stacks::contracts::SmartContract;
 use crate::storage;
 use crate::storage::model;
 use crate::storage::DbRead;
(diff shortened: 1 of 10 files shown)


signer/src/block_observer.rs
```diff
@@ -489,18 +490,21 @@ impl<C: Context, B> BlockObserver<C, B> {
...
-                    .context
-                    .get_stacks_client()
-                    .get_sbtc_total_supply(&self.context.config().signer.deployer)
-                    .await?;
-                // The maximum amount of sBTC that can be minted is the total cap
-                // minus the current supply.
-                total_cap.checked_sub(sbtc_supply).unwrap_or(Amount::ZERO)
-            }
+
+        let max_mintable = if limits.total_cap_exists() && self.is_token_deployed().await? {
```
Jiloc, line 494: Just a thought, but what if we move the `sbtc_contracts_deployed` field from the `TxCoordinatorEventlLoop` to the `SignerState` so that we can query it from here to avoid an extra call to the node?

https://github.com/stacks-network/sbtc/blob/main/signer/src/transaction_coordinator.rs#L1480

Jiloc (commented):
PR looks great, I just have a question about a possible change

signer/src/testing/stacks.rs
```diff
@@ -26,6 +28,21 @@ pub const DUMMY_SORTITION_INFO: SortitionInfo = SortitionInfo {
...
+/// Some dummy tenure info
+pub const DUMMY_TENURE_INFO: RPCGetTenureInfo = RPCGetTenureInfo {
+    consensus_hash: ConsensusHash([0; 20]),
+    tenure_start_block_id: StacksBlockId([0; 32]),
+    parent_consensus_hash: ConsensusHash([0; 20]),
+    parent_tenure_start_block_id: StacksBlockId([
+        0x55, 0xc9, 0x86, 0x1b, 0xe5, 0xcf, 0xf9, 0x84, 0xa2, 0x0c, 0xe6, 0xd9, 0x9d, 0x4a, 0xa6,
+        0x59, 0x41, 0x41, 0x28, 0x89, 0xbd, 0xc6, 0x65, 0x09, 0x41, 0x36, 0x42, 0x9b, 0x84, 0xf8,
+        0xc2, 0xee,
+    ]), // StacksBlockId::first_mined()
```
djordon, line 40: Curious, why the change? And what is the significance of this block?

Also, we should remove the comment right?

djordon (approved):
Looks good ✅

signer/src/block_observer.rs
djordon: We'd have to also check that the other smart contracts being deployed as well, since this only checks for the token contract. The remaining three smart contracts get submitted after the token one, and they can be confirmed much later.

djordon: Oh wait, but this could work if the `tx-coordinator` sets some state variable for each contract when it checks (which is what you had in mind to begin with).

signer/src/testing/stacks.rs
matteojug: We cannot use `StacksBlockId::first_mined()` since is not a const fn; those bytes are the result of calling `StacksBlockId::first_mined()`. I'll nit the comment to avoid confusion about it.

signer/src/block_observer.rs
matteojug: Moved to the state -- I've used a single flag, set by the coordinator once all the smart contracts are confirmed as deployed.

signer/src/testing/stacks.rs
matteojug: Done in [e6ff96b](https://github.com/stacks-network/sbtc/pull/1058/commits/e6ff96b95dc0bc93ff2d531d62eb859e3df1d60c)

djordon (approved):

Jiloc (approved):
looks good!

Issue #1055, opened 2024-12-06
[Chore]: Reduce default context window


matteojug:
# Chore - Reduce default context window

## 1. Description

As per https://github.com/stacks-network/sbtc/issues/978, we should reduce the context window size to reduce pressure on db. We may also need to update the bridge `RECLAIM_LOCK_TIME` (but that's already pretty low, so it should be already okay).

## 2. Related

 - https://github.com/stacks-network/sbtc/issues/993
 - https://github.com/stacks-network/sbtc/issues/978

Issue #993, opened 2024-11-29
[Feature]: Add configurable cap for deposit lock time


cylewitruk:
# Feature - Add configurable cap for deposit lock time

## 1. Description

At present, the deposit lock time can be arbitrarily specified. This implicitly means that the transaction coordinators will perpetually attempt to package deposits which have a very long lock time.

To prevent this, we should add a configurable max-lock-time (suggestion for default: 144 blocks = ~1 day) to prevent the list of pending serviceable requests from becoming too long (especially in the situation where we've hit the total-supply cap).

By doing this, it also allows us to reduce the `context window` parameters to a more reasonable number than the currently hardcoded `10_000` bitcoin blocks, as we'll know the maximum block age of potential deposits.

cylewitruk:
@djordon we touched on this last night, and the signer utxo was a cause for concern since looking for the spendable utxo also uses the context window. Was there something else to add here?

djordon:
I don't think so.

Just thought about the signer's UTXO a bit more, and I think that we can implement that in such a way where we don't require a context window.

matteojug:
As per offline discussion -- we don't actually need this here: given it's just a kind of UX safety net, it's better to handle it in the bridge (we already control the lock time via an env, [source](https://github.com/stacks-network/sbtc-bridge/blob/a19bb7f035623c468fb4a97c715f6709142b66fd/src/actions/get-sbtc-bridge-config.ts#L14) ) and can adjust that based on the context window we decide to use.
People are free to ignore the ~hardcoded~ suggested locktime value at risk of having the funds locked for longer than needed.

PR #1059, opened 2024-12-07
chore: reduce default context window

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/1055

## Changes

Reduce default context window to 500 ~144~ (as suggested in https://github.com/stacks-network/sbtc/issues/993, ~1d in mainnet, 12h in testnet)~. ~We can bump this up (500?) if the proposed one it's too tight~ bumped.

## Testing Information

Tests green and devenv demo script run okay.

## Checklist:

### DIFF
--- a/docker/prodlike/config/signer-config.toml
+++ b/docker/prodlike/config/signer-config.toml
@@ -163,7 +163,7 @@ bitcoin_block_horizon = 1500
 #
 # Required: false
 # Environment: SIGNER_SIGNER__CONTEXT_WINDOW
-# context_window = 144
+# context_window = 500
 
 # The maximum amount of time, in seconds, a signing round will take before
 # the coordinator will time out and return an error. This value must be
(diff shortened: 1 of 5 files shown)


cylewitruk (approved):
Looks good -- this is only like 12hrs in testnet but tbh I think that's fine.

djordon (approved):
Looks good. I would keep them at 500, at least on testnet so that we can test out performance implication (if we haven't already).

Large windows provide a slightly better user experience for peg-ins during periods of sudden bitcoin fee spikes, since the fee may exceed the users max-fee until things become more normal. So if we don't notice the performance hit with a larger context window then I'd vote that we go for it.

cylewitruk:
> Looks good. I would keep them at 500, at least on testnet so that we can test out performance implication (if we haven't already).
> 
> Large windows provide a slightly better user experience for peg-ins during periods of sudden bitcoin fee spikes, since the fee may exceed the users max-fee until things become more normal. So if we don't notice the performance hit with a larger context window then I'd vote that we go for it.

That makes sense 👍 I had tested this with both 500 and 1000 against a real signer db about a week ago, and both of those values were relatively fast.

PR #1065, opened 2024-12-08
chore: update stacks core git commit to the 3.1.0.0.0 release

cylewitruk:
## Description

This PR updates our stacks-core commit to 49777d3fd73a6dbb610be80c376b7d9389c9871a, which is the release tag for 3.1.0.0.0.

This should resolve issues that have been observed with regards to Stacks epoch 3.1 enum variants missing in responses from the Stacks node API, which at the moment causes the "process bitcoin blocks" function to bail before persisting new block information (after the 3.1 stacks-node update).

## Testing

All unit and integration tests pass locally.


djordon (approved):
Looks good ✅, I only checked that 3.1.0.0.0 has the same commit as `49777d3fd73a6dbb610be80c376b7d9389c9871a`.

cylewitruk:
👍 I didn't try this against devenv yet, won't merge until I do and verify the epoch errors dissapear.

cylewitruk:
Looks like this does indeed fix the epoch errors 👍

PR #1066, opened 2024-12-09
feat: do not update chainstate from signers

Jiloc:
## Description

Now that we have a sidecar, we don't need the signers to send and control chain state updates to Emily. 

## Changes

- Remove set_chainstate calls from the `new_block_handler`. 

## Testing Information
- Remove tests relative to the chainstate updates


cylewitruk:
Does this need to wait to be merged until everything emily-related is running? Or how should the deploy process look between signer vs emily vs sidecar?

Jiloc:
In dev-testnet we are already running the sidecar, so once approved and the singers updated in dev, it will keep working like nothing changed.

For primary testnet, we should deploy the sidecar before upgrading to this change.

For mainnet the deployment order should be Emily -> sidecar -> signers

cylewitruk:
Okay -- I think we should wait to merge this then until primary testnet is running the sidecar, because once merged it can quickly end up there.

AshtonStephens (approved):
LGTM

AshtonStephens (changes requested):
Revoking approval until we know the sidecar is working.

aldur:
@Jiloc according to your update from yesterday this is now being used in test, right?

Jiloc:
> @Jiloc according to your update from yesterday this is now being used in test, right?

@aldur Not exactly. This change is not in the testnet yet. But the sidecar is now deployed in testnet (and devnet). Emily is now configured to solely listen to it, for reorgs. So there is a short and a long explaination:

- Short answer: Now that Emily only listens to the sidecar for reorgs, these chainstate updates are not necessary and don't have any significant effect.

- Long answer about how it currently works:
  - If there's a fork or reorg that goes back in time, Emily will rely solely on the sidecar's chainstate.
  - Emily still accepts chainstate updates from the signers, but only if the update does not cause a reorg (i.e., it can be appended to the current chainstate).
    - The reason for still accepting the chainstates is to ensure that deposits and withdrawals updates from the signers are processed. Emily discards updates if she doesn’t recognize the corresponding Stacks block. This could happen if the sidecar is slightly behind.
    - However, in practice, Emily also infers the chainstate (as long as it doesn't cause a reorg) directly from the update requests themselves. So the signer’s chainstate updates are just redundant and without any real goal.

matteojug (approved):

PR #1081, opened 2024-12-09
Different Endpoint Permissions Emily

AshtonStephens:
## Description

Closes: #897

## Changes

This PR creates a number of different endpoints for Emily in order to facilitate different API keys for each api. AWS CDK does not have convenient interfaces for working with openapi json files, so each deployed API needs to have API keys with one set of permissions. We really want two sets of permission groups for API keys here: one that's admin like, and the other that's for the signers.

This PR splits the API into three different openapi versions:
- `public` -> for the signers and the regular folks
- `private` -> for the stacks node that updates chainstates and can initiate a reorg
- `testing` -> has the testing related endpoints (currently only wipe)

Additionally, the cdk template now needs to deploy each of these interfaces separately, but doesn't deploy the testing endpoint if the environment isn't of a development stage. The following is the naming framework for the different endpoints.

- `if stage != prod` -> `[stage].[purpose].[customRootDomainNameRoot]`
- `if stage == prod` -> `[purpose].[customRootDomainNameRoot]`

Lastly, a result of this is that there are now three autogenerated rust clients for Emily corresponding to each of the endpoint types.

## Testing Information

This PR was tested with both integration testing, and by deploying the api to the dev account under the stage `temp`. The interfaces all seem to work.

## Checklist:


signer/tests/integration/emily.rs
```diff
@@ -604,22 +603,23 @@ async fn get_deposit_request_works() {
...
+    //     &emily_client.config(),
+    //     Limits {
+    //         peg_cap: Some(Some(100)),
+    //         per_deposit_cap: Some(Some(90)),
+    //         per_withdrawal_cap: Some(Some(80)),
+    //         ..Default::default()
+    //     },
+    // )
+    // .await
+    // .unwrap();
```
Jiloc, line 622: nit: should this test also use the testing_emily_client or the private one?

Jiloc (approved):
lgtm! Possible goals for future refinement, not in this PR:

- make `set_chainstate` unavailable from the public endpoint
- remove the `TRUSTED_REORG_API_KEY` parameter now that we have a private api for that
    -  this will require to somehow modify the update_withdrawal/update_deposit handlers, so that it won't cause reorgs in the public implementation.
- update the Emily documentation to explain the difference between the two APIs, i.e. public is for bridge and signers, private is for the sidecar and admin (i.e. setting limits)

signer/tests/integration/emily.rs
AshtonStephens: It can set it with the testing one, but the actual signer will still use the structures from the public one. If you'd rather I can have this pull in the test api client and then create the limits and use the other for the rest of the test

Jiloc: This is postponed for a separete PR

Issue #685, opened 2024-10-20
[Feature]: Add a bitcoin block start height to the signers


djordon:
# Feature - Add a bitcoin block start height to the signers

## 1. Description

We should have a bitcoin start height for which all signers will have bitcoin and stacks block headers for. This issue is about making sure that the signers have a common idea for when to begin their bridging activities, and provides a foundational assumption that we can use in other areas.

### 1.1 Context & Purpose

Having a common bitcoin and stacks height solves a few problems.
* The signers need a good stopping point when backfilling bitcoin and stacks blocks.
* It also allows us to make simplifying assumptions in the block observer when fetching deposit requests. Specifically, it allows us to know that we can safely write the deposit request into the database in the code that originally closed https://github.com/stacks-network/sbtc/issues/522.

I expect that it will solve other problems since we can rely on the fact that the signers start with the same basic blockchain information.

## 2. Technical Details:

The natural choice for this bitcoin block start height is the stacks Nakamoto start height, since that is easy to fetch and can be used consistently in both testnet and mainnet. It also frees us from adding another variable in the config.

### 2.1 Acceptance Criteria:
   - [x] We have a process that fetches bitcoin blocks on boot up until the bitcoin start height.

## 3. Related Issues and Pull Requests (optional):

* The stacks side requires https://github.com/stacks-network/sbtc/issues/559.
* This also relates to https://github.com/stacks-network/sbtc/issues/637.

## 4. Appendix (new)

The block observer loads all blocks that need processing in memory before proceeding. This is not the best thing for us to do because we could need to process thousands of blocks at one time. Since each block takes up around ~1.5 MBs this could take up a lot of memory. Instead we should use the [getblockheader](https://bitcoincore.org/en/doc/25.0.0/rpc/blockchain/getblockheader/) RPC to get the headers and fetch the actual block when processing it.

djordon:
The stacks side of this was not completed in https://github.com/stacks-network/sbtc/pull/1093

aldur:
Status update: the Bitcoin side is out there, the Stacks side is still missing.

djordon:
Update here after chatting about this with @matteojug about the current state and what we want:
1. The bitcoin side has a configurable parameter for how far back the signer will fill their database.
2. The stacks side has no such parameter and always gets data back to nakamoto.
3. Having the disconnect where we do one thing on stacks and another on bitcoin is actually okay, but it feels wrong.
[...]

PR #1087, opened 2024-12-10
feat: bitcoin + stacks blockchain sync on startup

cylewitruk:
## Description

Closes: #685

## Changes

Will update, draft for now

## Checklist:


signer/migrations/0003__create_tables.sql
```diff
@@ -26,8 +26,6 @@ CREATE TABLE sbtc_signer.stacks_blocks (
     bitcoin_anchor BYTEA NOT NULL,
     created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL
 );
--- Index to serve queries filtering on `parent_hash`. This is commonly used when
```
cylewitruk, line 29: This index is added in a later script, but the comment got left here apparently, just cleaning up.

signer/src/block_observer.rs
```diff
@@ -276,20 +278,50 @@ impl<C: Context, B> BlockObserver<C, B> {
     /// Process the bitcoin block. Also process all recent stacks blocks.
     #[tracing::instrument(skip_all, fields(block_hash = %block.block_hash()))]
     async fn process_bitcoin_block(&self, block: bitcoin::Block) -> Result<(), Error> {
+        let storage = self.context.get_storage_mut();
+        let bitcoin_client = self.context.get_bitcoin_client();
+        let stacks_client = self.context.get_stacks_client();
+
         tracing::info!("processing bitcoin block");
+
+        let until_bitcoin_height = stacks_client
```
cylewitruk, line 287: Hm, maybe we should just store this in the `SignerState` at startup since it should never change while running..

signer/src/block_observer.rs
```diff
@@ -367,125 +399,6 @@ impl<C: Context, B> BlockObserver<C, B> {
...
 
-    /// Extract all BTC transactions from the block where one of the UTXOs
-    /// can be spent by the signers.
-    ///
-    /// # Note
-    ///
-    /// When using the postgres storage, we need to make sure that this
-    /// function is called after the `Self::write_bitcoin_block` function
-    /// because of the foreign key constraints.
-    pub async fn extract_sbtc_transactions(
```
cylewitruk, line 378: Note: these are just moved to top-level fn's in the file

signer/src/block_observer.rs
cylewitruk: Changed in [c7fb734](https://github.com/stacks-network/sbtc/pull/1087/commits/c7fb734e8b90856847a461c248e117146aa25ce9)

signer/src/blockchain_sync.rs
```diff
@@ -0,0 +1,336 @@
...
+
+        // We use `get_tenure` to fetch all of the Nakamoto blocks for each
+        // tenure, walking backwards, which accepts a `StacksBlockId` as its
+        // parameter. As we're walking backwards, we grab the first block in
+        // the tenure and use its parent block ID for the next iteration, which
+        // should give us the previous tenure and all of its blocks.
+        let first_block = tenure
+            .blocks()
+            .first()
+            .ok_or(Error::EmptyTenure(next_tenure_tip))?;
```
matteojug, line 250: in `fetch_unknown_ancestors` we use the last block, so I guess one of the two is wrong? the wrong one will still work since we are advancing, just very slowly (going through all the blocks instead of jumping right at the end)

matteojug, line 250: nit: we can move the `first.ok_or` up to replace `is_empty()`

signer/src/blockchain_sync.rs
```diff
@@ -0,0 +1,336 @@
...
+            &[tenure],
+        )
+        .await?;
+
+        // If the block height of the anchor block is equal to the nakamoto
+        // activation height then we're done.
+        if bitcoin_block.block_height == nakamoto_activation_height {
+            tracing::info!("nakamoto activation height reached, stacks block-sync completed");
+            return Ok(());
+        }
```
matteojug, line 268:
```suggestion
        // If the block height of the anchor block is equal to the nakamoto
        // activation height then we're done.
        if bitcoin_block.block_height <= nakamoto_activation_height {
            tracing::info!("nakamoto activation height reached, stacks block-sync completed");
            return Ok(());
        }
```
since we can have gaps? also to be consistent with `fetch_unknown_ancestors`

signer/src/context/signer_state.rs
```diff
@@ -55,6 +56,17 @@ impl SignerState {
...
+    /// Get the Nakamoto activation height.
+    pub fn nakamoto_activation_height(&self) -> u64 {
+        self.nakamoto_activation_height.load(Ordering::SeqCst)
+    }
+
+    /// Set the Nakamoto activation height.
+    pub fn set_nakamoto_activation_height(&self, height: u64) {
+        self.nakamoto_activation_height
+            .store(height, Ordering::SeqCst);
+    }
```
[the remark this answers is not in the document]
cylewitruk: Sure, maybe just `sbtc_activation_height` or `sbtc_bitcoin_activation_height`?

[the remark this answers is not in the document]
matteojug: As long as it has `bitcoin` and `height` it works for me 😄

signer/src/blockchain_sync.rs
```diff
@@ -0,0 +1,336 @@
...
+) -> Result<(), Error> {
+    let term = ctx.get_termination_handle();
+    let storage = ctx.get_storage_mut();
+    let bitcoin_client = ctx.get_bitcoin_client();
+    let mut next_bitcoin_block = *chain_tip;
+
+    loop {
+        if term.shutdown_signalled() {
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
cylewitruk: Leftover from an earlier refactor, will fix

signer/src/blockchain_sync.rs
```diff
@@ -0,0 +1,336 @@
...
+
+/// Helper function to determine the Nakamoto activation height (in Bitcoin
+/// block height).
+#[tracing::instrument(skip_all)]
+pub async fn determine_nakamoto_activation_height(ctx: &impl Context) -> Result<u64, Error> {
+    let mut term = ctx.get_termination_handle();
+    let stacks_client = ctx.get_stacks_client();
+
+    // If the nakamoto start height is provided in the config, use that value.
+    // Otherwise, get the value from the Stacks node's PoX-info endpoint.
```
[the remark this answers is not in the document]
cylewitruk: Yeah, missed update

signer/src/blockchain_sync.rs
```diff
@@ -0,0 +1,336 @@
...
+
+            tracing::info!(
+                current = %node_info.burn_block_height,
+                target = %nakamoto_activation_height,
+                "waiting for stacks node to reach nakamoto activation height"
+            );
+        }
+
+        tokio::select! {
+            _ = term.wait_for_shutdown() => return Ok(()),
```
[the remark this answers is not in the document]
cylewitruk: Yeah will update all of these

signer/src/blockchain_sync.rs
cylewitruk: Yeah good point 👍

signer/src/blockchain_sync.rs
```diff
@@ -0,0 +1,336 @@
...
+/// `is_fully_synced` field is `true`.
+#[tracing::instrument(skip_all)]
+pub async fn wait_for_stacks_node_to_report_full_sync(ctx: &impl Context) -> Result<(), Error> {
+    let mut term = ctx.get_termination_handle();
+    let stacks_client = ctx.get_stacks_client();
+
+    let mut interval = tokio::time::interval(Duration::from_secs(5));
+    loop {
+        match stacks_client.get_node_info().await {
+            Ok(node_info) if node_info.is_fully_synced => break,
```
matteojug, line 321: Given that in the nodes stalled on devnet this returns `true`, can we rely on this?

djordon (commented):
Hmmm, I think this solves one issue but opens another. I'm not entirely sure that I got this right but we may have an issue in the following scenario:
1. The signer is up and running, it has complete blockchain data in its database.
2. The signer binary stops for some reason, it's down for some number of hours.
3. The signer binary resumes, starts backfilling bitcoin blockchain data, but then stops again before the sync completes.
4. Some time later (I think any amount of time later) it resumes yet again.

After (4), I don't think that the code here will fetch all blockchain data. I think this is because the backfilling is done from the wrong direction (from the chain tip) but it needs to be done from the last known block or our "genesis" block.

signer/src/blockchain_sync.rs
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    Ok(())
+}
+
+/// Waits for the Stacks node to report that it is fully-synced. It does this
+/// by polling the node's `info` endpoint every 5 seconds until the response's
+/// `is_fully_synced` field is `true`.
+#[tracing::instrument(skip_all)]
+pub async fn wait_for_stacks_node_to_report_full_sync(ctx: &impl Context) -> Result<(), Error> {
```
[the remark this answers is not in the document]
cylewitruk: In general it's not a risk per-se, but if the Stacks node isn't synced then we know for a fact that the signer is going to spam a bunch of errors while either trying to play coordinator with outdated information (likely spamming the network with a bunch of stuff that will be rejected) or trying to validate things from signers which it doesn't have information for.

So this is more of a cheap insurance that the Stacks node is functioning and up-to-date before starting up and joining p2p.

I guess my counter-question would be why not ensure things are in a good state before starting? For a relatively short pause in operation, this shouldn't result in much of a pause (if any). If we join p2p and start signing sweeps but can't handle validating mints, wouldn't it be better to not do either? 🤷🏻‍♂️

signer/src/blockchain_sync.rs
cylewitruk: Most likely not -- all of my assumptions about this flag have been thrown out the window 😬 We'll nees to ask what it _actually_ means... it could be that it's in sync with bitcoin, or it could mean that it's in sync with other nodes when it comes to receiving (but not processing) stacks blocks... or something completely different

signer/src/blockchain_sync.rs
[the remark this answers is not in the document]
djordon: Yeah I gave the wrong impression here. I'm not saying we shouldn't wait, but I was thinking: _shouldn't we be waiting more often_? That's what I meant by 
> I mean, are these risks that are unique to booting the sBTC signer, or are these ever-present issues?

I was also thinking about whether we should be waiting _differently_. For example, the signer should probably be updating their database with bitcoin data, so that the backfill is faster when their Stacks node comes back online. And maybe they should vote on deposits. But yeah there are definitely things that they shouldn't do, like signing rounds and DKG (because of the key rotation contract call logic).

signer/src/main.rs
```diff
@@ -98,23 +102,58 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    // progress if it cannot send events to the observer and thus become out-
+    // of-sync.
+    let signer_api_handle = tokio::spawn(run_api(context.clone()));
+
+    // Pause until the Stacks node is fully-synced, otherwise we will not be
+    // able to properly back-fill blocks and the sBTC signer will generally just
+    // not work.
+    tracing::info!("waiting for stacks node to report full-sync");
+    wait_for_stacks_node_to_report_full_sync(&context).await?;
+    tracing::info!("stacks node reports that is is up-to-date");
```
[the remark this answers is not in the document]
cylewitruk: Yeah this is something that worried me a little too. I didn't see any FKs when I glanced through but it's possible there's something I missed or an edge-case somewhere.

It didn't really feel in-scope but one thing I think we should possibly do is change the event observers to only persist the events to the db to be processed in another task (which maybe waits for an "all-green" signal).

That would both help make sure the stacks node always gets a timely response so we don't block it, and lets us control when the events get processed.

djordon:
Closing, this was opened 5 months ago. We have the bitcoin side, and we're working on the stacks side in #1589

PR #1089, opened 2024-12-10
fix: add missing emily api keys for devenv

matteojug:
## Description

Closes: #?

## Testing Information

Devenv no longer fails to update emily.

## Checklist:


AshtonStephens (approved):
lgtm

docker/sbtc/signer/signer-config.toml
```diff
@@ -26,7 +26,7 @@
 # Environment: SIGNER_EMILY__ENDPOINTS
 # Environment Example: '"https://1234567890abcdef@api.emilyexample.com",..'
 endpoints = [
-    "http://emily-server:3031",
+    "http://testApiKey@emily-server:3031",
```
Jiloc, line 29: note that this api key will be able to trigger a reorg, so devenv will work slightly different from production where only the sidecar has this permission

cylewitruk: but we'll very soon remove this from the signer i guess, so maybe it doesn't matter too much right now.

PR #1092, opened 2024-12-10
[sbtc] Remove unused constants in sBTC token contract

friedger:
## Description

Remove unused constants in sBTC token contract

## Changes

Removes two constants from contract that were not used.

## Testing Information

Unit test adapted.
More unit tests added.

## Checklist:

### DIFF
--- a/contracts/contracts/sbtc-token.clar
+++ b/contracts/contracts/sbtc-token.clar
@@ -140,7 +140,9 @@
 (define-public (transfer (amount uint) (sender principal) (recipient principal) (memo (optional (buff 34))))
 	(begin
 		(asserts! (or (is-eq tx-sender sender) (is-eq contract-caller sender)) ERR_NOT_OWNER)
-		(ft-transfer? sbtc-token amount sender recipient)
+		(try! (ft-transfer? sbtc-token amount sender recipient))
+		(match memo to-print (print to-print) 0x)
+		(ok true)
 	)
 )
 
--- a/contracts/contracts/sbtc-token.clar
+++ b/contracts/contracts/sbtc-token.clar
@@ -1,7 +1,5 @@
 (define-constant ERR_NOT_OWNER (err u4)) ;; `tx-sender` or `contract-caller` tried to move a token it does not own.
-(define-constant ERR_NOT_AUTH (err u5)) ;; `tx-sender` or `contract-caller` is not the protocol caller
-(define-constant ERR_TRANSFER_INDEX_PREFIX (unwrap-err! ERR_TRANSFER (err true)))
-(define-constant ERR_TRANSFER (err u6))
+(define-constant ERR_TRANSFER_INDEX_PREFIX u1000)
 
 (define-fungible-token sbtc-token)
 (define-fungible-token sbtc-token-locked)
--- a/contracts/tests/sbtc-deposit.test.ts
+++ b/contracts/tests/sbtc-deposit.test.ts
@@ -98,7 +98,7 @@ describe("sBTC deposit contract", () => {
         deployer
       );
       expect(receipt0.value).toEqual(true);
-      simnet.mineEmptyBlock()
+      simnet.mineEmptyBlock();
       const receipt1 = txErr(
         deposit.completeDepositWrapper({
           txid: new Uint8Array(32).fill(1),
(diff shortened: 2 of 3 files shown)


contracts/contracts/sbtc-token.clar
```diff
@@ -107,17 +94,17 @@
 					sender: principal, 
 					to: principal, 
 					memo: (optional (buff 34)) })))
-	(fold complete-individual-transfer recipients (ok u0))
+	(fold transfer-many-iter recipients (ok u0))
```
friedger, line 97: This uses the same naming pattern as for protocol-mint-many

contracts/tests/sbtc-token.test.ts
```diff
@@ -320,8 +321,8 @@ describe("sBTC token contract", () => {
         }),
         alice
       );
-      // Many-transfer errors start at 6, error happens at index 1
-      expect(receipt1.value).toEqual(7n);
+      // Many-transfer errors start at 1000, error happens at index 1
+      expect(receipt1.value).toEqual(1001n);
```
friedger, line 325: I changed this for readability

contracts/tests/sbtc-token.test.ts
```diff
@@ -399,17 +400,159 @@ describe("sBTC token contract", () => {
...
-          memo: new Uint8Array(1).fill(0),
+          memo: null,
         }),
         bob
       );
       expect(receipt1.value).toEqual(errors.token.ERR_NOT_OWNER);
     });
   });
+
+  describe("protocol actions", () => {
```
friedger, line 450: This only works in non-mainnet

setzeus (approved):
PR consists of removing unused constant, unused comments & minimal formatting changes (such as error codes in 'sbtc-token').

Only one question re: failed costs at 400 but looks good to me otherwise (assuming 'git is dirty' issue is fixed cc @djordon)

contracts/tests/sbtc-deposit.test.ts
```diff
@@ -321,7 +321,7 @@ describe("optimization tests", () => {
     const { burnHeight, burnHash } = getCurrentBurnInfo();
 
     const totalAmount = 1000000n;
-    const runs = 650;
+    const runs = 400;
```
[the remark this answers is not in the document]
setzeus: Hmm, not entirely sure where that diff comes from? Should we change the list parameter size to 400 as well in the contracts?

contracts/contracts/sbtc-token.clar
setzeus: Copy, works.

contracts/tests/sbtc-token.test.ts
setzeus: Copy, no objection.

contracts/tests/sbtc-token.test.ts
setzeus: Not sure I follow? Does this mean this logic won't hold in mainnet/is broken or that this test only works in non-mainnet tests?

contracts/tests/sbtc-deposit.test.ts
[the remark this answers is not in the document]
djordon: We updated the contract calls to take in more data, write more, and emit more data. I would imagine the diff comes from that. I'm not sure if this will be a production issue as well, where 650 is unusable. If so, we should lower it.

djordon:
@friedger have you tried running `make build` locally and pushing the changes up? Sometimes that work, but sometimes not.

contracts/tests/sbtc-token.test.ts
friedger: This test only works in testnet because the deployer has protocol permissions

friedger:
I don't know why the check fails due to dirty git. Where get files changed?

contracts/tests/sbtc-deposit.test.ts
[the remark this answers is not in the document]
friedger: Where does the 650 come from?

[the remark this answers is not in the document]
setzeus: @hstove I remember you working on these optimization PRs, any idea here?

contracts/tests/sbtc-token.test.ts
setzeus: Ah, right, I follow.

djordon:
We run `make build` in CI, which generates some files. Here there are some documentation files that get generated based on the clarity source, so if the source changes the documentation will automatically get updated.

contracts/tests/sbtc-deposit.test.ts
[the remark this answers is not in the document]
djordon: My guess is that it is close to maximum allowed before the report returned an error.

PR #1093, opened 2024-12-11
feat: add bitcoin block start height to the signers

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/685 as well.

This PR focuses on the bitcoin side of things. The OOM possibility applies much less to Stacks blocks.

## Changes

* Add an optional config parameter that defines the minimum bitcoin block height for the signers. This means that the signers will have blocks for heights at least as high as the block set here. If the above parameter is unset, we use the nakamoto start height as the fallback value.
* Fetch bitcoin blocks all the way back to the last known block in the database. If there are no blocks in the database, then stop at the minimum block height defined above.
* Fetch historical bitcoin blocks by first fetching (slimmed down) block headers and then, once all block headers are known, fetch the bitcoin blocks one at a time. This limits memory usage of the previous approach.
* Remove the now unused horizon parameter.
* Add a function for calling the `getblockhash` RPC from bitcoin-core.
* Add a new query for checking whether a block is in the database.

## Testing Information

This needs more local testing.

## Checklist:

### DIFF
--- a/docker/sbtc/signer/signer-config.toml
+++ b/docker/sbtc/signer/signer-config.toml
@@ -148,7 +148,7 @@ dkg_begin_pause = 2
 # node if not present.
 #
 # Required: false
-# Environment: SIGNER_SIGNER__sbtc_bitcoin_start_height
+# Environment: SIGNER_SIGNER__SBTC_BITCOIN_START_HEIGHT
 # sbtc_bitcoin_start_height = 231
 
 # When defined, this field sets the scrape endpoint as an IPv4 or IPv6
--- a/docker/sbtc/signer/signer-config.toml
+++ b/docker/sbtc/signer/signer-config.toml
@@ -104,12 +104,6 @@ network = "regtest"
 # TODO(715): Add default delay (15 seconds? see #701)
 bitcoin_processing_delay = 3
 
-# The number of blocks back the block observer should look for unprocessed
-# blocks before proceeding.
-# Required: true
-# Environment: SIGNER_SIGNER__BITCOIN_BLOCK_HORIZON
-bitcoin_block_horizon = 1500
-
 # How many bitcoin blocks back from the chain tip the signer will look for
 # requests. Must be strictly positive.
 #
--- a/docker/sbtc/signer/signer-config.toml
+++ b/docker/sbtc/signer/signer-config.toml
@@ -148,8 +148,8 @@ dkg_begin_pause = 2
 # node if not present.
 #
 # Required: false
-# Environment: SIGNER_SIGNER__SBTC_START_HEIGHT
-# sbtc_start_height = 231
+# Environment: SIGNER_SIGNER__sbtc_bitcoin_start_height
+# sbtc_bitcoin_start_height = 231
 
 # When defined, this field sets the scrape endpoint as an IPv4 or IPv6
 # socket address for exporting metrics for Prometheus.
--- a/docker/sbtc/signer/signer-config.toml
+++ b/docker/sbtc/signer/signer-config.toml
@@ -142,6 +142,22 @@ bitcoin_processing_delay = 3
 # Environment: SIGNER_SIGNER__DKG_BEGIN_PAUSE
 dkg_begin_pause = 2
 
+# The minimum bitcoin block height for which the sbtc signers will backfill
+# bitcoin blocks to. The signers may not work if operated before this
+# height. Defaults to the Nakamoto start height returned from the stacks
+# node if not present.
+#
+# Required: false
+# Environment: SIGNER_SIGNER__SBTC_START_HEIGHT
+# sbtc_start_height = 231
+
+# When defined, this field sets the scrape endpoint as an IPv4 or IPv6
+# socket address for exporting metrics for Prometheus.
+#
+# Required: false
+# Environment: SIGNER_SIGNER__PROMETHEUS_EXPORTER_ENDPOINT
+# prometheus_exporter_endpoint = "[::]:9184"
+
 # The address that deployed the sbtc smart contracts.
 #
 # Required: true
--- a/signer/src/bitcoin/mod.rs
+++ b/signer/src/bitcoin/mod.rs
@@ -2,12 +2,12 @@
 
 use std::future::Future;
 
-use bitcoin::block::Header;
 use bitcoin::BlockHash;
 use bitcoin::Txid;
 
 use bitcoincore_rpc_json::GetMempoolEntryResult;
 use bitcoincore_rpc_json::GetTxOutResult;
+use rpc::BitcoinBlockHeader;
 use rpc::BitcoinTxInfo;
 use rpc::GetTxResponse;
 
@@ -57,7 +57,7 @@ pub trait BitcoinInteract: Sync + Send {
     fn get_block_header(
         &self,
         block_hash: &BlockHash,
-    ) -> impl Future<Output = Result<Option<Header>, Error>> + Send;
+    ) -> impl Future<Output = Result<Option<BitcoinBlockHeader>, Error>> + Send;
 
     /// get tx
     fn get_tx(
--- a/signer/src/bitcoin/mod.rs
+++ b/signer/src/bitcoin/mod.rs
@@ -2,6 +2,7 @@
 
 use std::future::Future;
 
+use bitcoin::block::Header;
 use bitcoin::BlockHash;
 use bitcoin::Txid;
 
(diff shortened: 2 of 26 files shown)


signer/src/block_observer.rs
```diff
@@ -243,60 +247,104 @@ impl<C: Context, B> BlockObserver<C, B> {
...
+    ///    encounters a known block header or if the height of the block is
+    ///    less than or equal to the `sbtc_start_height`.
+    ///
+    /// If there are many unknown blocks then this function can take some
+    /// time. Since each header is 80 bytes, we should be able to fetch
+    /// headers for the entire bitcoin blockchain (~900k blocks at the time
+    /// of writing) into memory.
     #[tracing::instrument(skip_all, fields(%block_hash))]
-    async fn next_blocks_to_process(
+    pub async fn next_headers_to_process(
```
cylewitruk, line 286: I really think that this sync/process logic should be somewhere else and callable both at startup and on observed block.

When running the initial sync here, then the signer is online and will attempt to respond to potential signing requests which will unnecessarily fail (bad chain tip). It's not a problem per-se I don't think, but it just doesn't feel "clean" and mucks up the logs.

Maybe I'm just oldschool, but I like to ensure that things are in a good/known state before starting up -- Stacks core does this too, for example, making sure it's synced bitcoin block headers before it starts doing stuff.

cylewitruk: Just want to add that this isn't a blocker for me to merge the PR, but something I think we should discuss

signer/src/block_observer.rs
```diff
@@ -243,60 +247,104 @@ impl<C: Context, B> BlockObserver<C, B> {
...
-                tracing::debug!(%block_hash, "already processed block");
+        let start_height = self.context.state().get_sbtc_start_height();
+        let mut headers = std::collections::VecDeque::new();
+        let db = self.context.get_storage();
+        let bitcoin_client = self.context.get_bitcoin_client();
+
+        while !db.is_known_bitcoin_block_hash(&block_hash.into()).await? {
+            let Some(header) = bitcoin_client.get_block_header(&block_hash).await? else {
+                tracing::warn!(%block_hash, "bitcoin-core does not know about block header");
+                return Ok(headers.into());
```
matteojug, line 300: Is this an `ok` situation? Could it be the case (eg, bitcoin node restarted/catching up while signer stays up) that we end up with gaps in the chain?
If it's `ok`, we can just `break`.

signer/tests/integration/block_observer.rs
```diff
@@ -811,5 +808,95 @@ async fn block_observer_handles_update_limits(deployed: bool, sbtc_limits: SbtcL
...
+        context: ctx.clone(),
+        bitcoin_blocks: (),
+    };
+
+    let chain_tip_block_hash = rpc.get_best_block_hash().unwrap();
+    let headers = block_observer
+        .next_headers_to_process(chain_tip_block_hash)
+        .await
+        .unwrap();
+    assert!(!headers.is_empty());
```
matteojug, line 842: nit: should we set some specific starting height in here? otherwise it's not clear what we are using

signer/tests/integration/block_observer.rs
```diff
@@ -811,5 +808,95 @@ async fn block_observer_handles_update_limits(deployed: bool, sbtc_limits: SbtcL
...
+    db.write_bitcoin_block(&block).await.unwrap();
+
+    // We know about the chain tip now, so we should return an empty vector
+    // of next headers to processes.
+    let headers = block_observer
+        .next_headers_to_process(chain_tip_block_hash)
+        .await
+        .unwrap();
+    assert!(headers.is_empty());
+
```
matteojug, line 900: nit: after the empty check, we could produce a new block here and check that the next `next_headers_to_process` only contain that one -- just as an extra sanity check

signer/src/block_observer.rs
```diff
@@ -243,60 +247,104 @@ impl<C: Context, B> BlockObserver<C, B> {
         Ok(())
     }
 
-    /// Find the parent blocks from the given block that are also missing from our database
+    /// Set the sbtc start height, if it has not been set already.
+    async fn set_sbtc_start_height(&self) -> Result<(), Error> {
```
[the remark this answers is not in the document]
djordon: Good point, will update.

signer/src/block_observer.rs
djordon: No, it is not! I need to return an empty `Vec`.

signer/src/block_observer.rs
djordon: Makes sense, I'm not opposed. I just didn't want to make the change here. Most of the decisions here are about keeping the diff as small as possible before launch.

signer/src/block_observer.rs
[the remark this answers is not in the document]
djordon: I'll get to this in a second. I updated everything else though... I think

djordon:
I think I updated most comments, will step away for a sec

signer/src/block_observer.rs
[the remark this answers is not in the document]
cylewitruk: We can do this in a later PR, feel free to resolve this

signer/src/bitcoin/mod.rs
```diff
@@ -52,6 +53,12 @@ pub trait BitcoinInteract: Sync + Send {
         block_hash: &BlockHash,
     ) -> impl Future<Output = Result<Option<bitcoin::Block>, Error>> + Send;
 
+    /// Fer the header of the block identified by the given block hash.
```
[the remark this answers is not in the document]
cylewitruk: I see this is resolved in [5d36eea](https://github.com/stacks-network/sbtc/pull/1093/commits/5d36eea859aab5b806f9e4706d6fa99c9d87d19a)

signer/src/block_observer.rs
cylewitruk: I see this is resolved in [5d36eea](https://github.com/stacks-network/sbtc/pull/1093/commits/5d36eea859aab5b806f9e4706d6fa99c9d87d19a):

```
let Some(header) = bitcoin_client.get_block_header(&block_hash).await? else {
    tracing::error!(%block_hash, "bitcoin-core does not know about block header");
    return Err(Error::BitcoinCoreUnknownBlockHeader(block_hash));
};
```

signer/src/context/signer_state.rs
```diff
@@ -19,6 +19,8 @@ pub struct SignerState {
     current_signer_set: SignerSet,
     current_limits: RwLock<SbtcLimits>,
     sbtc_contracts_deployed: AtomicBool,
+    sbtc_start_height: AtomicU64,
```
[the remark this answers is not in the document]
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/1093/commits/dde0f988566738f30862fe8c1bdd949a462070dc

signer/src/block_observer.rs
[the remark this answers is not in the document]
djordon: @cylewitruk reminded me offline that setting this on boot isn't so simple, so let's push this off until sometime later.

docker/sbtc/signer/signer-config.toml
```diff
@@ -148,6 +142,22 @@ bitcoin_block_horizon = 1500
...
 dkg_begin_pause = 2
 
+# The minimum bitcoin block height for which the sbtc signers will backfill
+# bitcoin blocks to. The signers may not work if operated before this
+# height. Defaults to the Nakamoto start height returned from the stacks
+# node if not present.
+#
+# Required: false
+# Environment: SIGNER_SIGNER__SBTC_START_HEIGHT
+# sbtc_start_height = 231
```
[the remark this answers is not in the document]
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1093/commits/dde0f988566738f30862fe8c1bdd949a462070dc

signer/tests/integration/block_observer.rs
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1093/commits/5d36eea859aab5b806f9e4706d6fa99c9d87d19a

signer/tests/integration/block_observer.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/1093/commits/5d36eea859aab5b806f9e4706d6fa99c9d87d19a

signer/src/block_observer.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/1093/commits/5d36eea859aab5b806f9e4706d6fa99c9d87d19a.

signer/src/bitcoin/mod.rs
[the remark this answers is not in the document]
djordon: https://github.com/stacks-network/sbtc/pull/1093/commits/5d36eea859aab5b806f9e4706d6fa99c9d87d19a

signer/src/block_observer.rs
djordon: Yeah we need to push this out of this PR for the same reasons mentioned in https://github.com/stacks-network/sbtc/pull/1093#discussion_r1880266567, which is that doing this start-up isn't so simple. Probably doable, just not super straightforward.

matteojug (approved):

PR #1095, opened 2024-12-11
Backwards compatible api keys

AshtonStephens:
## Description

Closes: #N/A

## Changes

- Fixes odd API key naming scheme I accidentally introduced in #897 
- Uses the right number of api keys per api
- Makes the api key names backwards compatible for the public endpoint

## Checklist:

### DIFF
--- a/emily/cdk/lib/emily-stack.ts
+++ b/emily/cdk/lib/emily-stack.ts
@@ -420,11 +420,12 @@ export class EmilyStack extends cdk.Stack {
             ]
         });
 
-        let num_api_keys = EmilyStackUtils.getNumSignerApiKeys();
         let api_keys = [];
-        for (let i = 0; i < num_api_keys; i++) {
+        for (let i = 0; i < numApiKeys; i++) {
             // Create an API Key
-            const apiKeyId: string = `ApiKey-${apiUsagePlan}-${i}`;
+            const apiKeyId: string = apiPurpose === "public"
+                ? `ApiKey-${apiUsagePlan}-${i}`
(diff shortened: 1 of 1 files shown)


emily/cdk/lib/emily-stack.ts
```diff
@@ -420,11 +420,12 @@ export class EmilyStack extends cdk.Stack {
...
 
-        let num_api_keys = EmilyStackUtils.getNumSignerApiKeys();
         let api_keys = [];
-        for (let i = 0; i < num_api_keys; i++) {
+        for (let i = 0; i < numApiKeys; i++) {
             // Create an API Key
-            const apiKeyId: string = `ApiKey-${apiUsagePlan}-${i}`;
+            const apiKeyId: string = apiPurpose === "public"
+                ? `ApiKey-${i}`
+                : `ApiKey-${apiPurposeTitleCase}-${i}`;
```
Jiloc, line 428: I am unclear about how the naming convention relates to API key generation. Is this using the previous naming convention, as CDK will avoid creating API keys that were already generated in earlier deployments based on that name? Is it implicit that we are going to do the same for the private-api keys?

AshtonStephens: yes. The private api keys will not have the same issue because its never been deployed. But keeping the same name will keep the api keys the same.

AshtonStephens: This change keeps the api keys having the same name for the public api, and appends the word `Private` for the private endpoint

Jiloc (approved):
lgtm!

Issue #1096, opened 2024-12-11
[Bug]: Length bug in rotate-keys-wrapper


djordon:
# Bug - Length bug in rotate-keys-wrapper

## 1. Description

The real limit on the number of keys that the `rotate-keys-wrapper` contract call accepts is 15, while the signature says that it accepts 128.

### 1.1 Context & Purpose

If we ever get to more than 15 signers, this would be an issue.

## 2. Technical Details:

Make sure that we can make `rotate-keys-wrapper` contract calls with 128 keys.

### 2.1 Acceptance Criteria:
   - [ ] We can make `rotate-keys-wrapper` contract calls with 128 keys.

## 3. Related Issues and Pull Requests (optional):

https://github.com/stacks-network/sbtc/issues/504#issuecomment-2532465931

Issue #504, opened 2024-09-10
[Feature]: Inconsistent code comment


evonide:
# (Low, non-security) Inconsistent code comment

## 1. Description

The comment [here](https://github.com/stacks-network/sbtc/blob/3da6e48a7bc471e1edd1bd21aa24cb10d760555e/contracts/contracts/sbtc-deposit.clar#L61-L63) is wrong since the list is 650 in length.

friedger:
Similar here: https://github.com/stacks-network/sbtc/blob/3da6e48a7bc471e1edd1bd21aa24cb10d760555e/contracts/contracts/sbtc-bootstrap-signers.clar#L103

Only 15 signers can be used to bootstrap, not 128.

djordon:
> Similar here:
> 
> https://github.com/stacks-network/sbtc/blob/3da6e48a7bc471e1edd1bd21aa24cb10d760555e/contracts/contracts/sbtc-bootstrap-signers.clar#L103
> 
> Only 15 signers can be used to bootstrap, not 128.

Oh, do you mean that the file should be renamed?

friedger:
> > Similar here:
> > https://github.com/stacks-network/sbtc/blob/3da6e48a7bc471e1edd1bd21aa24cb10d760555e/contracts/contracts/sbtc-bootstrap-signers.clar#L103
> > 
> > Only 15 signers can be used to bootstrap, not 128.
> 
> Oh, do you mean that the file should be renamed?

No, the list suggest 128, but the iterator only accepts 15 signers

friedger:
Similar here:

https://github.com/stacks-network/sbtc/blob/main/contracts/contracts/sbtc-bootstrap-signers.clar#L45

The comment suggest that contract-caller is checked, however, the tx-sender is checked. Which one is the correct check?

setzeus:
> > > Similar here:
> > > https://github.com/stacks-network/sbtc/blob/3da6e48a7bc471e1edd1bd21aa24cb10d760555e/contracts/contracts/sbtc-bootstrap-signers.clar#L103
> > > 
> > > Only 15 signers can be used to bootstrap, not 128.
> > 
> > 
> > Oh, do you mean that the file should be renamed?
> 
> No, the list suggest 128, but the iterator only accepts 15 signers

I think we agreed on 128 as a max? So as a way to future-proof we set the upper limit in the 'sbtc-registry' to 128. At least in the contracts there's no limit of 15 signers anywhere else?

[...]

PR #1097, opened 2024-12-11
fix: length bug in rotate-keys-wrapper contract call

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/504 and closes https://github.com/stacks-network/sbtc/issues/1096.

## Changes

* Made sure the highlighted code comments were accurate
* Fixed a lingering bug in the number of keys that are acceptable in `rotate-keys-wrapper` contract calls.
* Reduce the number of acceptable deposits in one `complete-deposits-wrapper` contract call down to 500. `pnpm test:report` complete successfully with this change.

## Testing Information

This adds a test against the actual limit, that was missing when some of the function signatures were changed to accept 128 keys as opposed to 15.

## Checklist:

### DIFF
--- a/contracts/contracts/sbtc-bootstrap-signers.clar
+++ b/contracts/contracts/sbtc-bootstrap-signers.clar
@@ -114,15 +114,15 @@
 )
 
 ;; Concatenate a pubkey buffer with a length prefix.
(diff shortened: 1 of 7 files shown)


setzeus (approved):
LGTM, thanks for catching!

Issue #1102, opened 2024-12-12
[Feature]: Use consistent hash endianness in contract calls


djordon:
# Feature - Use consistent hash endianness in contract calls

## 1. Description

Bitcoin block hashes that are sent in clarity contract calls must be expressed in little-endian byte since that is order that stacks-core has chosen. However, bitcoin transaction IDs are sent in the representation that we use internally, which is big-endian byte order. It would better for all bitcoin hashes to use the same byte order, so let's align them.

### 1.1 Context & Purpose

Make sure that all hash bytes are sent in the same order.

## 2. Technical Details:

Since we need bitcoin block hashes to be expressed in little-endian byte order, and we want them to all have the same endian order, let's submit things in little endian byte order.

### 2.1 Acceptance Criteria:
   - [ ] All bitcoin hash bytes are sent in contract calls in the little endian byte order.

PR #1103, opened 2024-12-12
feat: use consistent hash endianness in contract calls

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1102

## Changes

* Add traits for converting to and from little-endian byte order. Implement them for bitcoin `Txid` and `BlockHash` types.
* Convert bitcoin txid and block hash types to little-endian byte order for all contract calls transactions.

## Testing Information

I tested this out on local devenv.

## Checklist:

### DIFF
--- a/signer/src/stacks/events.rs
+++ b/signer/src/stacks/events.rs
@@ -30,23 +30,26 @@ use clarity::vm::Value as ClarityValue;
 use secp256k1::PublicKey;
 use stacks_common::types::chainstate::StacksBlockId;
 
-/// This trait adds a function for converting bytes from little-endian
-/// order into a type. This is because the signers convert
+/// This trait adds a function for converting bytes from little-endian byte
+/// order into a bitcoin hash types. This is because the signers convert
 /// [`bitcoin::Txid`] and [`bitcoin::BlockHash`] bytes into little-endian
-/// order before submitting the contract call.
+/// order before submitting contract calls.
 ///
 /// Both [`bitcoin::BlockHash`] and [`bitcoin::Txid`] are hash types that
-/// store bytes internally in big-endian order, and bitcoin-core transmits
-/// hashes in big-endian byte order[1] through the RPC interface. Note that
-/// the wire and zeromq interfaces transmit things in little-endian
-/// order[2].
+/// store bytes as SHA256 output, which is in big-endian order. Stacks-core
+/// stores hashes in little-endian byte order[2], implying that clarity
+/// functions, like `get-burn-block-info?`, return bitcoin block hashes in
+/// little-endian byte order. Note that Bitcoin-core transmits hashes in
+/// big-endian byte order[1] through the RPC interface, but the wire and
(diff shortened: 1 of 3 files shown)


signer/src/stacks/events.rs
```diff
@@ -30,6 +30,44 @@ use clarity::vm::Value as ClarityValue;
...
+/// big-endian byte order[1] through the RPC interface, but the wire and
+/// zeromq interfaces transmit hashes in little-endian order[3].
+///
+/// [^1]: See the Note in
+///     <https://github.com/bitcoin/bitcoin/blob/62bd61de110b057cbfd6e31e4d0b727d93119c72/doc/zmq.md>.
+/// [^2]: <https://github.com/stacks-network/stacks-core/blob/70d24ea179840763c2335870d0965b31b37685d6/stacks-common/src/types/chainstate.rs#L427-L432>
+/// [^3]: <https://developer.bitcoin.org/reference/block_chain.html#block-chain>
+///       <https://developer.bitcoin.org/reference/p2p_networking.html>
+/// <https://learnmeabitcoin.com/technical/general/byte-order/>
+pub trait FromLittleEndianOrder: Sized {
```
cylewitruk, line 52: suggestion: would prefer that this kind of trait be placed more "central" since we may want to use it for other things.

signer/src/storage/model.rs
```diff
@@ -623,6 +623,30 @@ pub struct QualifiedRequestId {
...
+/// big-endian byte order[1] through the RPC interface, but the wire and
+/// zeromq interfaces transmit hashes in little-endian order[3].
+///
+/// [^1]: See the Note in
+///     <https://github.com/bitcoin/bitcoin/blob/62bd61de110b057cbfd6e31e4d0b727d93119c72/doc/zmq.md>.
+/// [^2]: <https://github.com/stacks-network/stacks-core/blob/70d24ea179840763c2335870d0965b31b37685d6/stacks-common/src/types/chainstate.rs#L427-L432>
+/// [^3]: <https://developer.bitcoin.org/reference/block_chain.html#block-chain>
+///       <https://developer.bitcoin.org/reference/p2p_networking.html>
+/// <https://learnmeabitcoin.com/technical/general/byte-order/>
+pub trait ToLittleEndianOrder: Sized {
```
cylewitruk, line 645: suggestion: same as above, would prefer it be more central for re-use.

cylewitruk (commented):
This looks fine to me, i'm just trying to look through to see if maybe there might be a missed reversal somewhere

cylewitruk (approved):
Alright I think this looks fine, a bit mind-wrangling at times but yeah (was before too).. didn't find anywhere else where reversing is happening anyway

matteojug (approved):

signer/src/storage/model.rs
djordon: Yeah, we're due for a refactor. We can move this elsewhere

PR #1108, opened 2024-12-12
chore: change sidecar port 5000->20540

Jiloc:
## Description

Update the port used by the emily-sidecar from 5000 to 20540. The reason for this change is that port 5000 is commonly used by various other applications and services, which can lead to conflicts during deployment or local development. Switching to a less commonly used port (20540) reduces the risk of such conflicts and ensures smoother operation.

## Testing Information
- tested that the sidecar was still receiving update from the stacks-node and forwarding chainstate updates to emily in devenv

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -138,7 +138,7 @@ services:
       target: emily-sidecar
     restart: on-failure
     ports:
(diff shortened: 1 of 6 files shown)


cylewitruk (approved):
Thank you 🙏

PR #1109, opened 2024-12-12
chore: improve sidecar logs

Jiloc:
## Description

Improve logs of the sidecar, previously it was flooded by the events received by the stacks node, and difficult to read.

## Changes
- logs event validation errors
- logs sidecar configuration 
- simplify the information logged to just the chainstate and not the full events.


emily_sidecar/app.py
```diff
@@ -24,6 +24,9 @@ class Meta:
 url = os.getenv("EMILY_CHAINSTATE_URL",  "http://host.docker.internal:3031/chainstate")
 deployer_address = os.getenv("DEPLOYER_ADDRESS", "SN3R84XZYA63QS28932XQF3G1J8R9PC3W76P9CSQS")
 
+logger.info(f"Using chainstate URL: {url}")
```
aldur, line 27: nit: in theory you should use the "%s" format because otherwise you are formatting the f-string even if the logs aren't shown.

aldur (approved):

emily_sidecar/app.py
Jiloc: Good to know, I never thought about it!

Issue #1114, opened 2024-12-13
[Bug]: DKG can be triggered whenever


djordon:
# Bug - DKG can be triggered whenever

## 1. Description

The coordinator can trigger DKG whenever they want. This is problematic because we currently assume that the aggregate key locking all UTXOs are the same. If the keys get rotated, then we won't properly sign for the signer's UTXO or any other deposits. Instead, we should make sure that DKG cannot be triggered again, and even if it is, we should make sure that we properly sign for any UTXOs.

### 1.1 Context & Purpose

DKG should only ever be triggered once, at least until we finish key rotation support. Until then we need to make sure that it doesn't run more than once.

## 2. Technical Details:

We should stop DKG from running more than once.

### 2.1 Acceptance Criteria:
   - [ ] Stop DKG from running more than once.
   - [ ] Sign deposits UTXOs and the signers' UTXO using the correct DKG secret shares.

PR #1115, opened 2024-12-13
fix: run DKG only once

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1114

## Changes

* Construct the WSTS coordinator state machine using the correct DKG shares.
* Do not run DKG if it has been run before.
* Remove some currently unused code. The tests that exercise that code path have already been ignored.

## Testing Information

No tests yet, hence the draft status.

## Checklist:

### DIFF
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -1460,8 +1460,8 @@ impl super::DbRead for PgStore {
         X: Into<PublicKeyXOnly> + Send,
     {
         // The aggregate_key column stores compressed public keys, which
-        // always include a parity byte. Since the input is an x-only
-        // public key, we don't have a parity byte, so we lop it off when
+        // always include a parity byte. Since the input here is an x-only
+        // public key we don't have a parity byte, so we lop it off when
         // filtering.
         let key: PublicKeyXOnly = aggregate_key.into();
         sqlx::query_as::<_, model::EncryptedDkgShares>(
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -1459,6 +1459,9 @@ impl super::DbRead for PgStore {
     where
         X: Into<PublicKeyXOnly> + Send,
     {
+        // The first two characters are for the aggregate key column
+        // represent the parity byte. Since the input is an x-only public
+        // key, we don't have a parity byte, so we lop it off.
         let key: PublicKeyXOnly = aggregate_key.into();
         sqlx::query_as::<_, model::EncryptedDkgShares>(
             r#"
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -1459,9 +1459,10 @@ impl super::DbRead for PgStore {
     where
         X: Into<PublicKeyXOnly> + Send,
     {
-        // The first two characters are for the aggregate key column
-        // represent the parity byte. Since the input is an x-only public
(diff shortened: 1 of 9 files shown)


signer/src/storage/postgres.rs
```diff
@@ -1452,10 +1452,17 @@ impl super::DbRead for PgStore {
...
+    async fn get_encrypted_dkg_shares<X>(
         &self,
-        aggregate_key: &PublicKey,
-    ) -> Result<Option<model::EncryptedDkgShares>, Error> {
+        aggregate_key: X,
+    ) -> Result<Option<model::EncryptedDkgShares>, Error>
+    where
+        X: Into<PublicKeyXOnly> + Send,
+    {
+        // The first two characters are for the aggregate key column
```
matteojug, line 1462: nit?
```suggestion
        // The first two characters in the aggregate key column
```

signer/tests/integration/stacks_events_observer.rs
```diff
@@ -109,7 +109,7 @@ async fn test_new_blocks_sends_update_deposits_to_emily() {
         .expect("Wiping Emily database in test setup failed.");
 
     let body = COMPLETED_DEPOSIT_WEBHOOK.to_string();
-    let new_block_event = serde_json::from_str::<NewBlockEvent>(&body).unwrap();
+    serde_json::from_str::<NewBlockEvent>(&body).unwrap();
```
matteojug, line 112: nit: remove this?

matteojug (commented):
LGTM

signer/tests/integration/stacks_events_observer.rs
djordon: I kept it because I want to check that we deserialize the JSON correctly. I suppose that test is implicitly handled in this test anyway, and it is handled in our unit tests.

djordon: I'll just remove it

signer/src/storage/postgres.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/1115/commits/856d1f4273c79d9beb6b3eb96f8c13752f6c3f30 and https://github.com/stacks-network/sbtc/pull/1115/commits/86642684946fb78d68643d7417010881dcdfdf3d

signer/tests/integration/stacks_events_observer.rs
djordon: Removed in https://github.com/stacks-network/sbtc/pull/1115/commits/5a352d6912b9512f87c044c1dbb22f41c69cc950

matteojug (approved):
Tested in devnet

djordon:
Oh I figured out how to test this and it isn't so bad. Will add tomorrow.

signer/src/storage/postgres.rs
```diff
@@ -1467,10 +1475,10 @@ impl super::DbRead for PgStore {
               , signer_set_public_keys
               , signature_share_threshold
             FROM sbtc_signer.dkg_shares
-            WHERE aggregate_key = $1;
+            WHERE substring(aggregate_key FROM 2) = $1;
```
cylewitruk, line 1478: Just a nit -- I think I would have preferred that we stored this in an extra column instead of messing around with it at query-time. I think we do something similar and even more convoluted somewhere else..

PR #1117, opened 2024-12-13
Add test for first signer set

friedger:
## Description

This PR adds a test for clarity to calculate the current signer from the first signer set at launch of sbtc 

## Changes

* added new file

## Checklist:

### DIFF
--- a/contracts/tests/sbtc-first-signer-set.test.ts
+++ b/contracts/tests/sbtc-first-signer-set.test.ts
@@ -4,7 +4,8 @@ import { describe, expect, test } from "vitest";
 import { constructMultisigAddress, signers } from "./helpers";
 
 describe("sBTC first signer set", () => {
-  test("Deployer address is correct", () => {
+  test("Deployer address is same as SignerWallet address if sorted", () => {
+    const signatureThreshold = 10n;
     const keys = [
       "03431638d7656f2b313776ab8e5fbffb994e0635cb53965d33cdb819384400286f",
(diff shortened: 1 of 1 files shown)


contracts/tests/sbtc-first-signer-set.test.ts
```diff
@@ -0,0 +1,54 @@
...
+      "03ed8edbcbb26010ae3a992ea1e4fffab3805e1303edbd3c130307b4cbcd38a481",
+    ]
+      .sort() // sort the keys hex string wise
+      .map((k) => hexToBytes(k));
+
+    // testnet signer address from contract
+    const currentSigner = rov(
+      signers.pubkeysToPrincipal(keys, signatureThreshold)
+    );
+    expect(currentSigner).toEqual("SN3VDXK3WZZSA84XXFKAFAF15NNZX32CTSJV6BTWG");
```
djordon, line 33: I suppose having it here is fine, but this isn't the sbtc signers' address on testnet. I would update the comment above to say that this is what the signers' address would be on testnet if the above public keys were used there. The mainnet one is the correct one though.

djordon (commented):
This largely looks good. The only worry is that this test gives the wrong impression about the testnet address for the sbtc signers. I would add a comment that clarifies that the testnet address here is not the one to actually expect on testnet.

aldur:
Closing this, we can re-open if comments are addressed.

PR #1119, opened 2024-12-13
Fix: revision in CI Docker image

aldur:
## Description

The Dockerfile we are using in CI prevented the `build.rs` file from getting the current `git_commit`, which was resulting in an empty revision (e.g. in metrics, this was also breaking Cloudwatch collection).

(For context, this took me longer than I expected because the code did not check the result of `git rev-parse` and would produce an empty string on failure).

This fixes it.

Before merging, please that the build resulting from [this manual run](https://github.com/stacks-network/sbtc/actions/runs/12337468550/job/34431241890) embeds the correct revision.


signer/build.rs
```diff
@@ -11,11 +13,15 @@ pub fn set_up_build_info() {
...
         .args(["rev-parse", "HEAD"])
         .output()
-        .map(|output| String::from_utf8_lossy(&output.stdout).to_string())
-        .unwrap_or_default();
+    {
+        Ok(output) if output.status.success() && !output.stdout.is_empty() => {
+            String::from_utf8_lossy(&output.stdout).to_string()
+        }
+        _ => std::env::var("GIT_COMMIT").unwrap_or_default(),
+    };
```
matteojug, line 24: nit: for readabiliy split we could split this in something like
```
let git_output = std::process ... .output();
let git_hash = match git_output {
...
}
```

matteojug (approved):
Tested locally, devenv local and devenv pulling the built image from CI and got a revision in all of them (CI one):
```
starting the sBTC signer rust_version="rustc 1.80.0 (051478957 2024-07-21)" revision="fix/revision_in_docker_image" arch="x86_64" env_abi="gnu"
```

PR #1121, opened 2024-12-13
Deposit minimums in limits

AshtonStephens:
## Description

> Note: This changes autogenerated sources so it looks bigger than it is. Just review the first commit.

Closes: #N/A

## Changes

Adds minimums to limits endpoint.

## Testing Information

Updated integration tests. Currently running the tests.

## Checklist:


emily/handler/src/database/accessors.rs
```diff
@@ -669,7 +671,9 @@ pub async fn get_limits(context: &EmilyContext) -> Result<Limits, Error> {
     // Get the global limit for the whole thing.
     Ok(Limits {
         peg_cap: global_cap.peg_cap,
+        per_deposit_minimum: default_global_cap.per_deposit_minimum,
         per_deposit_cap: global_cap.per_deposit_cap,
+        per_withdrawal_minimum: default_global_cap.per_withdrawal_minimum,
```
Jiloc, line 676: Should these be `global_cap` instead of `default_global_cap`?

Jiloc: already fixed!

.generated-sources/emily/client/rust/private/docs/Limits.md
```diff
@@ -7,7 +7,9 @@ Name | Type | Description | Notes
 **account_caps** | [**std::collections::HashMap<String, models::AccountLimits>**](AccountLimits.md) | Represents the individual limits for requests coming from different accounts. | 
 **peg_cap** | Option<**u64**> | Represents the total cap for all pegged-in BTC/sBTC. | [optional]
 **per_deposit_cap** | Option<**u64**> | Per deposit cap. If none then there is no cap. | [optional]
+**per_deposit_minimum** | Option<**u64**> | Per deposit minimum. If none then there is no minimum. | [optional]
 **per_withdrawal_cap** | Option<**u64**> | Per withdrawal cap. If none then there is no cap. | [optional]
+**per_withdrawal_minimum** | Option<**u64**> | Per withdrawal minimum. If none then there is no minimum. | [optional]
```
djordon, line 12: Why add withdrawal minimums?

djordon (commented):
Why add withdrawal minimums? I don't think that there should be configurable withdrawal minimums, but instead we should have the same minimums enforced on bitcoin.

.generated-sources/emily/client/rust/private/docs/Limits.md
AshtonStephens: I just added all around minimums - we can delete it later if we decide we don't want the option.

aldur: As a rule, I would keep scope to the minimum.

AshtonStephens: You would keep withdrawal minimum in scope, or you would remove the withdrawal minimum because it's not in scope? @Jiloc and I aren't sure which it is

AshtonStephens:
Waiting on tests

.generated-sources/emily/client/rust/private/docs/Limits.md
AshtonStephens: Done.

Jiloc (approved):
lgtm!

PR #1124, opened 2024-12-13
add per_deposit_minimum validation

Jiloc:
## Description

Add support for the `per_deposit_minimum` limit from Emily

## Changes

- Retrieve the new values from Emily
- Validate both in the coordinator and the signers that the amount is above the minimum

## Testing Information
- Add unit tests

## Checklist:

### DIFF
--- a/signer/src/bitcoin/utxo.rs
+++ b/signer/src/bitcoin/utxo.rs
@@ -175,15 +175,18 @@ impl SbtcRequests {
         // Filter deposit requests based on two constraints:
         // 1. The user's max fee must be >= our minimum required fee for deposits
         //     (based on fixed deposit tx size)
-        // 2. The deposit amount must be less than the per-deposit limit
-        // 3. The total amount being minted must stay under the maximum allowed mintable amount
+        // 2. The deposit amount must be greater than or equal to the per-deposit minimum
+        // 3. The deposit amount must be less than or equal to the per-deposit cap
+        // 4. The total amount being minted must stay under the maximum allowed mintable amount
         let minimum_deposit_fee = self.compute_minimum_fee(SOLO_DEPOSIT_TX_VSIZE);
         let max_mintable_cap = self.sbtc_limits.max_mintable_cap().to_sat();
         let per_deposit_cap = self.sbtc_limits.per_deposit_cap().to_sat();
+        let per_deposit_minimum = self.sbtc_limits.per_deposit_minimum().to_sat();
 
         let mut amount_to_mint: u64 = 0;
         let deposits = self.deposits.iter().filter_map(|req| {
             let is_fee_valid = req.max_fee.min(req.amount) >= minimum_deposit_fee;
+            let is_above_per_deposit_minimum = req.amount >= per_deposit_minimum;
             let is_within_per_deposit_cap = req.amount <= per_deposit_cap;
             let is_within_max_mintable_cap =
                 if let Some(new_amount) = amount_to_mint.checked_add(req.amount) {
@@ -191,7 +194,11 @@ impl SbtcRequests {
                 } else {
                     false
                 };
-            if is_fee_valid && is_within_per_deposit_cap && is_within_max_mintable_cap {
+            if is_fee_valid
+                && is_above_per_deposit_minimum
+                && is_within_per_deposit_cap
+                && is_within_max_mintable_cap
+            {
                 amount_to_mint += req.amount;
                 Some(RequestRef::Deposit(req))
             } else {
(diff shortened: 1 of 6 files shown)


signer/src/bitcoin/validation.rs
```diff
@@ -1088,10 +1130,19 @@ mod tests {
...
+            TX_FEE,
+            &SbtcLimits::new(
+                None,
+                Some(Amount::from_sat(per_deposit_minimum)),
+                Some(Amount::from_sat(per_deposit_cap)),
+                None,
+                None,
+                None,
+            ),
+        );
```
djordon, line 1145: nit: it's much easier to read if we just split this out like:
```rust
let limits = SbtcLimits::new(
    None,
    Some(Amount::from_sat(per_deposit_minimum)),
    Some(Amount::from_sat(per_deposit_cap)),
    None,
    None,
    None,
);
let status = mapping
    .report
    .validate(mapping.chain_tip_height, &tx, TX_FEE, limits);
```
or if we add a test helper implementation, we could do:
```rust
#[test_case(DepositReportErrorMapping {
    report: DepositRequestReport { ... },
    status: InputValidationResult::AmountTooLow,
    chain_tip_height: 2,
    sbtc_limits: SbtcLimits::per_deposit(100_000_000, u64::MAX),
} ; "amount-too-low")]
fn deposit_report_validation(mapping: DepositReportErrorMapping) {
    ...
    let status = mapping
        .report
        .validate(mapping.chain_tip_height, &tx, TX_FEE, &mapping.sbtc_limits:);
```

signer/src/bitcoin/utxo.rs
```diff
@@ -175,23 +175,30 @@ impl SbtcRequests {
...
+        // 4. The total amount being minted must stay under the maximum allowed mintable amount
         let minimum_deposit_fee = self.compute_minimum_fee(SOLO_DEPOSIT_TX_VSIZE);
         let max_mintable_cap = self.sbtc_limits.max_mintable_cap().to_sat();
         let per_deposit_cap = self.sbtc_limits.per_deposit_cap().to_sat();
+        let per_deposit_minimum = self.sbtc_limits.per_deposit_minimum().to_sat();
 
         let mut amount_to_mint: u64 = 0;
         let deposits = self.deposits.iter().filter_map(|req| {
             let is_fee_valid = req.max_fee.min(req.amount) >= minimum_deposit_fee;
+            let is_above_per_deposit_minimum = req.amount >= per_deposit_minimum;
```
djordon, line 189: I think that this closure has gotten complex enough where we should break it out into a separate function or create some struct with some function. I feel that doing so would be much easier to understand and test.

djordon (commented):
This largely looks good, but I don't think that we should have withdrawal minimums. Well, I don't think that they should be configurable minimums, but instead be the same minimums enforced on bitcoin.

For enforcing the deposit minimums, we should just refactor the code in the utxo module to isolate the limits logic in a proper function which gets used in the `filter` closure. Doing so will make it easier to understand and test.

signer/src/bitcoin/utxo.rs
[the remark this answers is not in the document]
djordon: Actually, the above is incorrect, let me explain. There needs to be two checks; one `req.amount >= per_deposit_minimum` check and `eq.max_fee.min(amount) - minimum_deposit_fee > 546` check. The second check is fixing a bug that we have here. The clarity smart contracts have a check that the amounts are greater than a `DUST_LIMIT` constant which is 546. So we need to do a filter here.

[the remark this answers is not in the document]
djordon: Actually, we don't need to fix the dust limit thing here since it's a bit more involved. I opened https://github.com/stacks-network/sbtc/issues/1135 for this.

signer/src/bitcoin/validation.rs
```diff
@@ -1078,8 +1084,44 @@ mod tests {
...
+            reclaim_script: ScriptBuf::new(),
+            signers_public_key: *sbtc::UNSPENDABLE_TAPROOT_KEY,
+        },
+        status: InputValidationResult::AmountTooLow,
+        chain_tip_height: 2,
+    }, 100_000_000, u64::MAX ; "amount-too-low")]
+    fn deposit_report_validation(
+        mapping: DepositReportErrorMapping,
+        per_deposit_minimum: u64,
+        per_deposit_cap: u64,
```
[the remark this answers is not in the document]
Jiloc: makes lots of sense, changed in https://github.com/stacks-network/sbtc/pull/1124/commits/b813d2062d178cb5b961b065585c12d0ea04bb0b

signer/src/context/signer_state.rs
```diff
@@ -82,8 +82,12 @@ impl SignerState {
 pub struct SbtcLimits {
     /// Represents the total cap for all pegged-in BTC/sBTC.
     total_cap: Option<Amount>,
+    /// Represents the minimum amount of BTC allowed to be pegged-in per transaction.
+    per_deposit_minimum: Option<Amount>,
     /// Represents the maximum amount of BTC allowed to be pegged-in per transaction.
     per_deposit_cap: Option<Amount>,
+    /// Represents the minimum amount of sBTC allowed to be pegged-out per transaction.
+    per_withdrawal_minimum: Option<Amount>,
```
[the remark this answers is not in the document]
Jiloc: removed in https://github.com/stacks-network/sbtc/pull/1124/commits/b813d2062d178cb5b961b065585c12d0ea04bb0b

signer/src/bitcoin/validation.rs
Jiloc: now the limits are included in the DepositReportErrorMapping  https://github.com/stacks-network/sbtc/pull/1124/commits/b813d2062d178cb5b961b065585c12d0ea04bb0b

signer/src/bitcoin/utxo.rs
Jiloc: Done in https://github.com/stacks-network/sbtc/pull/1124/commits/555ffbaec8d59796bac43e4a72bb8cf3be602a43

signer/src/bitcoin/utxo.rs
```diff
@@ -109,6 +109,70 @@ pub trait GetFees {
...
+
+impl<'a> DepositFilter<'a> {
+    /// Create a new [`DepositFilter`] instance.
+    pub fn new(
+        requests: &'a Vec<DepositRequest>,
+        minimum_fee: u64,
+        per_deposit_minimum: u64,
+        per_deposit_cap: u64,
+        max_mintable_cap: u64,
+    ) -> Self {
```
djordon, line 130: nit: wouldn't it be simpler to have this function take the limits struct for some of these? 

It also might make sense to implement `validate_deposit` on the `SbtcLimits` struct itself, since it can be viewed as a separate kind of check. And I would imagine we'd put a `validate_withdrawal` function there too. But this is also fine.

djordon (commented):
Largely looks good, just had a couple of minor comments.

signer/src/bitcoin/utxo.rs
```diff
@@ -109,6 +109,70 @@ pub trait GetFees {
     fn get_fees(&self) -> Result<Option<Fees>, Error>;
 }
 
+/// Filter out the deposit requests that do not meet the amount or fee requirements.
+pub struct DepositFilter<'a> {
+    requests: &'a Vec<DepositRequest>,
```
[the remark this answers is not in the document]
Jiloc: done in https://github.com/stacks-network/sbtc/pull/1124/commits/6df60b07836f4a3f49d1523eb37f8f72de9e1452

signer/src/bitcoin/utxo.rs
Jiloc: I tried to move it in `SbtcLimits` like you suggested but, when I looked a bit at the things to move around to support also the filtering of withdrawals, it felt too much validation logic needed to be moved in the signer_state file, and I didn't like it 🥲. Eventually, I renamed the struct `DepositFilter` -> `SbtcRequestsFilter`, thinking that we can implement the withdrawals filter in there once we'll know move onto that

djordon: Oh okay, yeah I can see how that can happen, makes sense.

signer/src/bitcoin/utxo.rs
```diff
@@ -109,6 +109,64 @@ pub trait GetFees {
...
+    /// 1. The user's max fee must be >= our minimum required fee for deposits
+    ///     (based on fixed deposit tx size)
+    /// 2. The deposit amount must be greater than or equal to the per-deposit minimum
+    /// 3. The deposit amount must be less than or equal to the per-deposit cap
+    /// 4. The total amount being minted must stay under the maximum allowed mintable amount
+    fn validate_deposit(
+        &self,
+        amount_to_mint: &mut Amount,
+        req: &'a DepositRequest,
+    ) -> Option<RequestRef<'a>> {
```
djordon, line 134: This only validates the deposit amounts, so I would name this slightly different, so that it's clearer what is being validated.

At first I thought that this should return a boolean, but that would make it's use more annoying down the line so returning a `Option<RequestRef<'a>>` makes more sense.

signer/src/bitcoin/utxo.rs
```diff
@@ -2783,136 +2831,6 @@ mod tests {
...
-        ),
-        true; "deposit amounts over the dust limit accepted")]
-    #[test_case(
-        create_deposit(
-            DEPOSIT_DUST_LIMIT + SOLO_DEPOSIT_TX_VSIZE as u64 - 1,
-            10_000,
-            0
-        ),
-        false; "deposit amounts under the dust limit rejected")]
-    fn deposit_requests_respect_dust_limits(req: DepositRequest, is_included: bool) {
```
djordon, line 2873: Why delete these?

Jiloc: I moved them in the filter_deposits test

https://github.com/stacks-network/sbtc/pull/1124/files#diff-aef027d6a4d764f2511fb7b7065d4b2b7ecff26b7ff1e59deaa25955705f1736R2908-R2914
and 
https://github.com/stacks-network/sbtc/pull/1124/files#diff-aef027d6a4d764f2511fb7b7065d4b2b7ecff26b7ff1e59deaa25955705f1736R2983

djordon: I see, well I would keep them both. Also, it looks like you missed the "deposit amounts over the dust limit accepted" one right?

Issue #1122, opened 2024-12-13
[Feature]: Add a way to filter deposits by recipient in Emily


djordon:
# Feature - Add a way to filter deposits by recipient in Emily

## 1. Description

It's annoying to get deposits that a user may care about. I think we need to get everything and filter by hand. Let's not do that

### 1.1 Context & Purpose

We want to get all deposits that may be relevant to a user. This will help.

### 2.1 Acceptance Criteria:
   - [ ] We support requests that only return deposits that peg-in to a particular recipient.

PR #1125, opened 2024-12-13
feat: Query By Recipient

AshtonStephens:
## Description

Closes: #1122 

## Changes

This PR adds a new endpoint to emily. Unfortunately, the client autogeneration stopped working when I made the status optional in the `get_deposits` query, and I couldn't find any autogenerated solution, so to resolve this I've added a new endpoint:

`https://{base_url}/deposit/recipient/{recipient}`.

It works much like the other query but has the recipient in the path. This is a departure from the rest implementation, but it's the only way I could get it working.

In order to support this, there's now a new deposit table index that can be accessed by recipient.

## Testing Information

Tested with an integration test in this PR.

## Checklist:


Jiloc:
Given that we are not rushing this feature, would it make sense to have another time-gated effort to try it to make it work with the `get_deposits` endpoint as a query? If you think that's just time wasted, I am fine with going forward as it is

Jiloc (commented):
Given that we are not rushing this feature, would it make sense to have another time-gated effort to try it to make it work with the get_deposits endpoint as a query? If you think that's just time wasted, I am fine with going forward as it is

Jiloc:
> Given that we are not rushing this feature, would it make sense to have another time-gated effort to try it to make it work with the get_deposits endpoint as a query? If you think that's just time wasted, I am fine with going forward as it is

As discussed, an important time was already allocated for this and there was no apparent workaround.  We are fine with the current approach after @MCJOHN974 comments are solved

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -31,6 +31,18 @@ pub struct GetDepositsQuery {
...
+/// Query structure common for all paginated queries.
+#[derive(Clone, Default, Debug, PartialEq, Hash, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct BasicPaginationQuery {
+    /// Next token for the search.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub next_token: Option<String>,
+    /// Maximum number of results to show.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub page_size: Option<i32>,
```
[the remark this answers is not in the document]
AshtonStephens: The dynamodb interface uses i32, so this is just to keep it consistent without casting.

https://github.com/stacks-network/sbtc/blob/query-deposits-by-recipient/emily/handler/src/database/entries/mod.rs

I've changed it everywhere to be a u16 so that it can cast safely to an i32 for the underlying library.

emily/handler/tests/integration/deposit.rs
```diff
@@ -314,7 +324,7 @@ async fn get_deposits() {
     // If the number of elements is an exact multiple of the chunk size the "final"
     // query will still have a next token, and the next query will now have a next
     // token and will return no additional data.
-    let expected_chunks = expected_deposit_infos.len() as i32 / chunksize + 1;
+    let expected_chunks: i32 = expected_deposit_infos.len() as i32 / chunksize + 1;
```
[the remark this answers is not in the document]
AshtonStephens: Making the same change everywhere

emily/handler/src/common/error.rs
```diff
@@ -106,6 +106,10 @@ pub enum Error {
     /// in an update not being performed.
     #[error("Version conflict")]
     VersionConflict,
+
+    /// Bad request
+    #[error("Bad request {0}")]
+    BadRequest(String),
```
Jiloc, line 112: Just wondering, how are we using this currently?

AshtonStephens: Looks like we don't use it, do you want me to remove it in this PR?

Jiloc: No, that's fine, I'll probably use it for the create_deposit validation PR as I am sending back a lot of 400s

Jiloc (approved):

Issue #1100, opened 2024-12-11
[Feature]: Retrieve and Store sBTC Deposit Amount 


Jiloc:
# Feature - Retrieve and Store sBTC Deposit Amount

## 1. Description

Currently, the `amount` field in `ScriptParameters` is hardcoded to `0` in the Emily `deposit.rs` handler. This prevents accurate reporting of the deposit amount in the bridge and other user-facing interfaces. The feature request aims to extract the correct deposit amount from the provided transaction data and store it appropriately in the system.

### 1.1 Context & Purpose
The hardcoded value was noted during development as a placeholder because the deposit amount is encoded in the original transaction, and Emily doesn't currently have access to it. This limitation impacts Emily's ability to provide accurate information for deposits.

## 2. Technical Details:
One. simple, possible solution could be to add the deposit amount to the request (`CreateDepositRequestBody`) and save it alongside the deposit request. Note that the amount will not be validated by Emily (nor will be used by the signers, that will keep using the original tx), but will just be stored for visualization purposes.

### 2.1 Acceptance Criteria:
   - [ ] Add the amount to the CreateDepositRequestBody.
   - [ ] Replace the hardcoded `amount: 0` value with the extracted deposit amount.

## 3. Related Issues and Pull Requests (optional):
- [Reference code with the hardcoded `amount: 0`](https://github.com/stacks-network/sbtc/blob/fdbf78481cbd28d420bab98dcd98725dc19bc8f8/emily/handler/src/api/handlers/deposit.rs#L295).

djordon:
We decided to have Emily take the entire transaction hex, so we can get the amount from there. So this is no longer needed.

PR #1149, opened 2024-12-16
Feat/emily retrieve and store deposit amount

Jiloc:
## Description

Closes: #1100 

## Changes
- add `amount: Option<u64>` to `CreateDepositRequest`
    - As explained in the linked issue, Emily doesn't have enough information to retrieve by herself the amount. 
    - Emily does not validate (or uses) the amount. It may just be used by the bridge itself to show more precise information instead of the currently hardcoded 0.
    - se it to Option, so that it won't break retrocompability with the bridges that don't upgrade to this change.
   
## Testing Information
 - updated all unit and integration tests to work with the new parameter


Jiloc:
create_deposit will be modified to get the entire tx information so that Emily can validate the txid, the real amount will be extracted from there

Issue #1155, opened 2024-12-18
[Feature]: Poll `completed-deposits` contract map prior to submitting complete-doposit contract call


cylewitruk:
# Feature - Poll `completed-deposits` contract map prior to submitting complete-doposit contract call

## 1. Description

To help avoid wasting fees on failed complete-deposit contract calls due to replay, we could use the `getmapentry` Stacks RPC to poll the `completed-deposits` map in the registry contract to check if the node's view of the contract has already seen the contract call, and avoid submitting it again.

The map could be checked together with the deposit outpoint:

```clarity
;; Data structure to map successful deposit requests
;; to their respective sweep transaction. Stores the
;; txid, burn hash, and burn height.
(define-map completed-deposits {txid: (buff 32), vout-index: uint}
	{
		amount: uint,
		recipient: principal,
		sweep-txid: (buff 32),
		sweep-burn-hash: (buff 32),
		sweep-burn-height: uint,
	}
)
```

PR #1158, opened 2024-12-18
RFC: cache + index for slow queries

cylewitruk:
## Description

This is a suggestion for how me might be able to, in a minimalistic way (pre-holidays), reduce pressure from the two slow SQL queries that we currently see logged often. This would help the stacks signing rounds to complete much more quickly.

_This is intended only as a band-aid until the underlying issues can be addressed more properly (which I'd rather avoid before the holidays)._

This includes two changes:

1.  An index for `sbtc_signer.stacks_blocks(block_height DESC, block_hash DESC)` which may help the query in the `get_last_key_rotation()` method. (addressed this separately in #1159, will remove)
2.  Local caching of the results of `get_swept_deposit_requests()` within the validation of Stacks complete-deposit contract call signing requests, using a LRU with capacity of 5. During Stacks signing rounds, we see this query reaching up to 5-6s in some cases, despite a context window of 1000.

Point 1 is simple, but point 2 has the potential caveat that once the (non-empty) result for a chain tip is cached, if new stacks block events arrive after this then they will still be included in the results (as they are cached), which means that we may attempt to mint if enough signers have the same view. This would ultimately be stopped by the dedup logic in the contract, but could cost fees (which could be mitigated by #1155). I'm not sure of the likelihood of this actually happening, though, assuming that the signing round times are shortened.


cylewitruk:
Closing this since we picked up the "real" solution

Issue #1160, opened 2024-12-18
[feature] - Move sBTC webhooks module to the sbtc crate


MCJOHN974:
# Feature - Move sBTC event module to the sbtc crate

## 1. Description

The `signer/src/stacks/webhooks.rs` module was meant to live in the sbtc crate, where it can be used more easily by those who want to interact with the protocol. Let's move it there.

### 1.1 Context & Purpose

This ticket is very similar to https://github.com/stacks-network/sbtc/issues/1016, and has the same purpose. The difference is that the webhooks module requires a stackslib dependency so we should be mindful about that in our implementation. Specifically, we should feature gate the webhooks module and any additional dependencies.

## 2. Technical Details:

Move the module from the `signer` crate to the `sbtc` crate.

### 2.1 Acceptance Criteria:
   - [ ] The `signer/src/stacks/webhooks.rs` module lives in the sbtc crate.
   - [ ] We use the now foreign types in the signer crate

## 3. Related Issues and Pull Requests (optional):

#1016 was tracking a similar refactoring

## 4. Appendix

We'll want to add a `webhooks` feature to the sbtc crate. This way people can use it without pulling in too many dependencies.

PR #1163, opened 2024-12-18
refactor: move webhooks module to the sbtc crate

MCJOHN974:
## Description

Closes: #1160

## Changes
Moved `webhooks` module to the `sbtc` crate. Feature gated it. All code in `signer` now use imported from `sbtc` types.

## Testing Information
All old tests from `signer` applicable.
## Checklist:

### DIFF
--- a/sbtc/src/webhooks.rs
+++ b/sbtc/src/webhooks.rs
@@ -31,17 +31,7 @@ use stacks_common::types::chainstate::BlockHeaderHash;
 use stacks_common::types::chainstate::BurnchainHeaderHash;
 use stacks_common::types::chainstate::StacksBlockId;
 
-/// Webhooks error
-#[derive(Debug, thiserror::Error)]
-pub enum Error {
-    /// This is thrown when failing to parse a hex string into bytes.
-    #[error("could not decode the hex string into bytes: {0}")]
-    DecodeHexBytes(#[source] hex::FromHexError),
-    /// An error when attempting to generically decode bytes using the
-    /// trait implementation.
-    #[error("got an error when attempting to call StacksMessageCodec::consensus_deserialize {0}")]
-    StacksCodec(#[source] blockstack_lib::codec::Error),
-}
+use crate::error::Error;
 
 /// This struct represents the body of POST /new_block events from a stacks
 /// node.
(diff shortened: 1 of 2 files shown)


sbtc/src/webhooks.rs
```diff
@@ -31,7 +31,17 @@ use stacks_common::types::chainstate::BlockHeaderHash;
...
-use crate::error::Error;
+/// Webhooks error
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    /// This is thrown when failing to parse a hex string into bytes.
+    #[error("could not decode the hex string into bytes: {0}")]
+    DecodeHexBytes(#[source] hex::FromHexError),
+    /// An error when attempting to generically decode bytes using the
+    /// trait implementation.
+    #[error("got an error wen attempting to call StacksMessageCodec::consensus_deserialize {0}")]
```
djordon, line 42: Oops, this was in the original. Would you mind fixing here.
```suggestion
    #[error("got an error when attempting to call StacksMessageCodec::consensus_deserialize {0}")]
```

sbtc/src/webhooks.rs
```diff
@@ -31,7 +31,17 @@ use stacks_common::types::chainstate::BlockHeaderHash;
 use stacks_common::types::chainstate::BurnchainHeaderHash;
 use stacks_common::types::chainstate::StacksBlockId;
 
-use crate::error::Error;
+/// Webhooks error
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
djordon, line 36: Can we rename this `WebhookError`?

djordon: Actually, can we add these error variants to the crate error and feature gate them as well? Everything should work after importing the crate `Error`. @cylewitruk any opinions here?

djordon (commented):
This largely looks good.

The only question is where to place the Error variants. If we keep them in the webhooks module then we should rename the type. But there are only two variants so I think we can add them to the crate error (feature gated though).

sbtc/src/webhooks.rs
djordon: The proposed change won't force users to import more types, but instead force them to import a type from a different place. Which reminds me, we should probably move this enum to the `error` module.

djordon:
Yeah, it's `load_latest_deposit_requests_persists_requests_from_past` 😞, we need to fix it.

djordon (approved):
Looks good ✅!

Issue #1164, opened 2024-12-18
[Feature]: Return a proper errors instead of a `debug_assert`s


djordon:
# Feature - Return a proper errors instead of a `debug_assert`s

## 1. Description

We have some `debug_assert`s in the codebase that turn out to be conditions that can be tripped in production. We should turn some of these into proper Errors. The one that we should definitely handle is the one here:

https://github.com/stacks-network/sbtc/blob/155f0d4b26bcd99fce8f8b0871ad119431c6229a/signer/src/transaction_signer.rs#L413-L416

### 1.1 Context & Purpose

We think that one of the `debug_assert`s is a proper error so it should be treated as such. For the others, we should at least have a look to see if they should be removed or upgraded and act on them accordingly.

## 2. Technical Details:

Generally speaking, we create new error variants for the `debug_assert` condition that we want to convert (if an error variant does not exist for the condition already) and return them when the proper condition is triggered.

### 2.1 Acceptance Criteria:
   - [ ] The debug assertion highlighted above is replaced with a proper error.

PR #1165, opened 2024-12-18
[feature] return a proper errors instead of debug asserts

MCJOHN974:
## Description

Closes: #1164

## Checklist:

### DIFF
--- a/signer/src/error.rs
+++ b/signer/src/error.rs
@@ -454,7 +454,7 @@ pub enum Error {
     #[error("invalid configuration")]
     InvalidConfiguration,
 
-    /// We throw this when signer produced txid, and txid coordinator sent are different
+    /// We throw this when signer produced txid and coordinator produced txid differ.
     #[error(
         "Signer and coordinator txid mismatch. Signer produced txid {0}, but coordinator send txid {1}"
     )]


djordon:
> I will replace other `debug_assert_*` in another commits, WIP

Before replacing them, try to figure out whether we should just remove them instead. That's a little harder to do, but give it a shot.

signer/src/error.rs
```diff
@@ -237,6 +241,11 @@ pub enum Error {
     #[error("invalid tweak? seriously? {0}")]
     InvalidPublicKeyTweak(#[source] secp256k1::Error),
 
+    /// This happens when a tweak produced by [`XOnlyPublicKey::add_tweak`] was computed incorrectly.
+    /// One if possible reasons is that you tweaked same key twice.
```
djordon, line 245: I would remove the second line, I'm not sure that it would point us in the right direction for the cause of the error. It's not obvious what the error really means here, as in I'm not sure why an error is possible to begin with.
```suggestion
    /// This happens when a tweak produced by [`XOnlyPublicKey::add_tweak`] was computed incorrectly.
```

signer/src/error.rs
```diff
@@ -444,6 +454,15 @@ pub enum Error {
     #[error("invalid configuration")]
     InvalidConfiguration,
 
+    /// We throw this when signer produced txid, and txid coordinator sent are different
```
djordon, line 457: I would rephrase this a little:
```suggestion
    /// We throw this when signer produced txid and coordinator produced txid differ.
```

Issue #1170, opened 2024-12-19
[Feature]: Materialized database views for canonical chains


cylewitruk:
# Feature - Materialized database views for canonical chains

## 1. Description

Instead of recursively walking the blockchains in real-time for many of our queries, we should materialize the data after the processing of each Bitcoin block so that queries can use the materialized "view" instead of recursive CTE's.

### 1.1 Context & Purpose
In production, we have a lot of signers consistently reporting slow queries (see the `slow-query-stats` dashboard in cloudwatch). But in general, we have two queries which have an average slow-query reported time of `3s`, and several others which often report closer to `1s`.

PR #1171, opened 2024-12-19
feat: add sql-side of a pseudo-materialized-view for the blockchains

cylewitruk:
## Description

Closes: #1170

## Changes

Adds the SQL-side of pseudo-materialized views for the canonical blockchains, keyed on the bitcoin chain tip. It's intended that the `materialize_canonical_chains` function is called after each Bitcoin block has been processed (but that's not included here).

## Testing Information

The added script has been run and tested against a local copy of the signer database from testnet using the `get_last_key_rotation()` query. The updated query I used (below) reduced the execution time on my local machine from 0.2 seconds to 0.002:

```sql
select 
	rkt.txid,
	rkt.address,
	rkt.aggregate_key,
	rkt.signer_set,
	rkt.signatures_required
from rotate_keys_transactions rkt 
inner join stacks_transactions st on st.txid = rkt.txid
inner join canonical_chains cc 
	on cc.stacks_block_hash = st.block_hash 
	and cc.bitcoin_chain_tip = '\x9F654F6D199EB6691B560A2CE2511F600D21E2F350549855488A75CC7FAF7D59'
order by 
	cc.stacks_block_height desc, 
	cc.stacks_block_hash desc, 
	rkt.created_at desc
limit 1
```

### DIFF
--- a/signer/migrations/0011__materialized_chains.sql
+++ b/signer/migrations/0011__materialized_chains.sql
@@ -1,0 +1,112 @@
+-- This table holds the canonical chain information for Bitcoin and Stacks
+-- chains keyed by the bitcoin chain tip.
+CREATE TABLE sbtc_signer.canonical_chains (
+    run_id INT,
+    bitcoin_chain_tip BYTEA,
+    bitcoin_block_hash BYTEA,
+    bitcoin_block_height BIGINT,
+    stacks_block_hash BYTEA,
+    stacks_block_height BIGINT
+);
+
+-- Indexes to support common queries on the canonical chains table.
+CREATE INDEX ix_canonical_chains_run_id ON sbtc_signer.canonical_chains(run_id);
+CREATE INDEX ix_canonical_chains_bitcoin_chain_tip ON sbtc_signer.canonical_chains(bitcoin_chain_tip);
+CREATE INDEX ix_canonical_chains_bitcoin_block_hash ON sbtc_signer.canonical_chains(bitcoin_block_hash);
+CREATE INDEX ix_canonical_chains_bitcoin_block_height ON sbtc_signer.canonical_chains(bitcoin_block_height);
+CREATE INDEX ix_canonical_chains_stacks_block_hash ON sbtc_signer.canonical_chains(stacks_block_hash);
+CREATE INDEX ix_canonical_chains_stacks_block_height ON sbtc_signer.canonical_chains(stacks_block_height);
+
+-- New sequence which will be used to generate the run_id for each materialized
+-- view canonical chain.
+CREATE SEQUENCE sbtc_signer.canonical_chains_run_id_seq;
+
+-- Function to materialize the canonical chains for a given bitcoin chain tip
+-- and the maximum depth of the bitcoin blockchain to consider.
+CREATE OR REPLACE FUNCTION sbtc_signer.materialize_canonical_chains(chain_tip BYTEA, max_depth INT)
+RETURNS INTEGER AS $$
+DECLARE
+    rows_written INTEGER;
+	existing_rows INTEGER;
+	new_run_id BIGINT;
+BEGIN
+	-- Get the next value for the run_id
+    new_run_id := nextval('sbtc_signer.canonical_chains_run_id_seq');
+
+	-- Check if rows exist with the given chain_tip
+    SELECT COUNT(*) INTO existing_rows
+    FROM sbtc_signer.canonical_chains
+    WHERE bitcoin_chain_tip = chain_tip;
+
+    -- If rows exist, return an error code
+    IF existing_rows > 0 THEN
+        RETURN -1; -- Error code indicating rows already exist
+    END IF;
+
+	-- Materialize the canonical chains from the given bitcoin chain tip.
+    WITH RECURSIVE 
+    bitcoin AS (
+        SELECT 
+            block_hash,
+            parent_hash,
+            block_height,
+			1 as depth
+        FROM sbtc_signer.bitcoin_blocks
+        WHERE block_hash = chain_tip
+
+        UNION ALL
+
+        SELECT 
+            parent.block_hash,
+            parent.parent_hash,
+            parent.block_height,
+			last.depth + 1
+        FROM sbtc_signer.bitcoin_blocks parent
+        JOIN bitcoin last ON parent.block_hash = last.parent_hash
+		WHERE last.depth < max_depth
+    ),
+    stacks AS (
+        (SELECT 
+            blocks.block_hash,
+            blocks.parent_hash,
+            blocks.block_height,
+            blocks.bitcoin_anchor
+        FROM sbtc_signer.stacks_blocks blocks
+        JOIN bitcoin ON blocks.bitcoin_anchor = bitcoin.block_hash
+        ORDER BY bitcoin.block_height DESC, bitcoin.block_hash DESC, blocks.block_height DESC, blocks.block_hash DESC
+        LIMIT 1)
+
+        UNION ALL
+
+        SELECT 
+            parent.block_hash,
+            parent.parent_hash,
+            parent.block_height,
+            parent.bitcoin_anchor
+        FROM sbtc_signer.stacks_blocks parent
+        JOIN stacks last ON parent.block_hash = last.parent_hash
+        JOIN bitcoin ON bitcoin.block_hash = parent.bitcoin_anchor
+    )
+    INSERT INTO sbtc_signer.canonical_chains (
+		run_id,
+        bitcoin_chain_tip,
+        bitcoin_block_hash,
+        bitcoin_block_height,
+        stacks_block_hash,
+        stacks_block_height
+    )
+    SELECT 
+		new_run_id,
+        chain_tip,
+        bb.block_hash AS bitcoin_block_hash,
+        bb.block_height AS bitcoin_block_height,
+        sb.block_hash AS stacks_block_hash,
+        sb.block_height AS stacks_block_height
+    FROM bitcoin bb
+    LEFT JOIN stacks sb ON sb.bitcoin_anchor = bb.block_hash
+    ORDER BY bb.block_height DESC, sb.block_height DESC;
+
+    GET DIAGNOSTICS rows_written = ROW_COUNT;
+    RETURN rows_written;
+END;
+$$ LANGUAGE plpgsql;


signer/migrations/0011__materialized_chains.sql
```diff
@@ -0,0 +1,115 @@
+-- This table holds the canonical chain information for Bitcoin and Stacks
+-- chains keyed by the bitcoin chain tip.
+CREATE TABLE sbtc_signer.canonical_chains (
+    run_id INT,
+    bitcoin_chain_tip BYTEA,
+    bitcoin_block_hash BYTEA,
+    bitcoin_block_height BIGINT,
+    stacks_block_hash BYTEA,
```
matteojug, line 15: So apparently we don't have any fk on stacks blocks `bitcoin_anchor` (to bitcoin blocks): given how we process stuff below (joining on the anchor) we should be fine, but if we were to change that nothing would prevent -- data-integrity-wise -- to have null `bitcoin_block_hash`, which could be annoying maybe? So I guess the question is: should we throw some `NOT NULL` in here (for everything but stacks stuff)?

signer/migrations/0011__materialized_chains.sql
```diff
@@ -0,0 +1,115 @@
...
+
+	-- Check if rows exist with the given chain_tip
+    SELECT COUNT(*) INTO existing_rows
+    FROM sbtc_signer.canonical_chains
+    WHERE bitcoin_chain_tip = chain_tip;
+
+    -- If rows exist, return an error code
+    IF existing_rows > 0 THEN
+        RETURN -1; -- Error code indicating rows already exist
+    END IF;
```
matteojug, line 80: I'm not sure about this: we get new blocks from the block observer when it thinks we have a better block, but we use our own logic to determine the chain tip -- in particular wrt the height tie breaking; so I think it can happen that we call this function twice for the same chain tip. I'm moderately sure it doesn't matter for the bitcoin side (like, we just need to handle the error correctly outside), but it could matter for the stacks side since we may have new stacks blocks?
Also, do we need to lock/run this whole function in a transaction? Because it's fine for the coordinator, but the signers may start to query this for a bitcoin tip that is still being worked on.

signer/migrations/0011__materialized_chains.sql
```diff
@@ -0,0 +1,148 @@
...
+-- New sequence which will be used to generate the run_id for each materialized
+-- view canonical chain.
+CREATE SEQUENCE sbtc_signer.canonical_chains_run_id_seq;
+
+-- Function to materialize the canonical chains for a given bitcoin chain tip
+-- and the maximum depth of the bitcoin blockchain to consider.
+--
+-- This function returns the number of rows written to the `canonical_chains`
+-- table upon success, and `-1` if rows already exist for the given chain tip.
+CREATE OR REPLACE FUNCTION sbtc_signer.materialize_canonical_chains(chain_tip BYTEA, max_depth INT)
```
matteojug, line 62: Do we need to lock/run this whole function in a transaction? Because it's fine for the coordinator, but the signers may start to query this for a bitcoin tip that is still being worked on.

cylewitruk: postgres runs all statements within a function in a transaction (i.e. everything between begin/end), so it should be fine. [Link](https://cooxe.dev/2023-10-06-transaction-in-postgresql-function.html)

signer/migrations/0011__materialized_chains.sql
cylewitruk: I had "overwriting" of the chain-tip in my first iteration but removed it, primarily to avoid that some rogue code accidentally calls it twice for the same chain-tip (we don't have any PK's/UK's here) -- which is why I chose to return an error code so that the calling code could detect this and choose to separately remove the chain tip and replace it. That's how I'm thinking anyway 🤷

signer/migrations/0011__materialized_chains.sql
cylewitruk: Yeah we could do that, bitcoin blocks should _never_ be null here, so better to protect it 👍

cylewitruk: [34a7709](https://github.com/stacks-network/sbtc/pull/1171/commits/34a77091b9a6f3aeda8f4ee2f2214b1b400078c7)

signer/migrations/0011__materialized_chains.sql
cylewitruk: Oh, and the default isolation level is `READ COMMITTED`, so a reader will never get partially-written/uncommitted data.

signer/migrations/0011__materialized_chains.sql
matteojug: That makes sense 👍

signer/migrations/0011__materialized_chains.sql
```diff
@@ -0,0 +1,148 @@
...
+-- materialized. If no Stacks block(s) exist for a given Bitcoin block, the
+-- `stacks_block_hash` and `stacks_block_height` columns will be `NULL` for that
+-- Bitcoin block hash.
+CREATE TABLE sbtc_signer.canonical_chains (
+    run_id INT NOT NULL,
+    bitcoin_chain_tip BYTEA NOT NULL,
+    bitcoin_block_hash BYTEA NOT NULL,
+    bitcoin_block_height BIGINT NOT NULL,
+    stacks_block_hash BYTEA,
+    stacks_block_height BIGINT
```
djordon, line 16: No primary key? I suppose one gets created for you using the OID or something(?), but it still feels icky.

signer/migrations/0011__materialized_chains.sql
```diff
@@ -0,0 +1,148 @@
...
+-- Note that this view will return multiple canonical chains, so you need to
+-- filter the results to the specific chain tip which you are interested in.
+CREATE VIEW sbtc_signer.canonical_chains_view AS
+SELECT
+    bitcoin_chain_tip,
+    bitcoin_block_hash,
+    bitcoin_block_height,
+    stacks_block_hash,
+    stacks_block_height
+FROM sbtc_signer.canonical_chains;
```
djordon, line 30: I think I would forgo this views. It seems like very thin layers on top of the `canonical_chains` table, so I would always query that directly.

signer/migrations/0011__materialized_chains.sql
```diff
@@ -0,0 +1,148 @@
...
+-- used in queries, the `bitcoin_chain_tip` column should be used to filter the
+-- results to the specific chain tip which you are interested in. The results
+-- are not ordered by block height, so you should do that yourself if you
+-- require the results to be ordered.
+CREATE VIEW sbtc_signer.canonical_bitcoin_chain_view AS
+SELECT DISTINCT
+    bitcoin_chain_tip,
+    bitcoin_block_hash,
+    bitcoin_block_height
+FROM sbtc_signer.canonical_chains;
```
djordon, line 43: I would probably remove this one too, but the `DISTINCT` adds some value here. I'm okay with either.

djordon (commented):
Very nice! I had some nits. 

Also, I'm not sure if we want two tables, one table with bitcoin blocks and another with bitcoin and stacks blocks. That might make sense, but maybe it's not worth it.

signer/migrations/0011__materialized_chains.sql
```diff
@@ -0,0 +1,148 @@
...
+        UNION ALL
+
+        SELECT 
+            parent.block_hash,
+            parent.parent_hash,
+            parent.block_height,
+			last.depth + 1
+        FROM sbtc_signer.bitcoin_blocks parent
+        JOIN bitcoin last ON parent.block_hash = last.parent_hash
+		WHERE last.depth < max_depth
```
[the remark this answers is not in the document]
cylewitruk: I'm actually hoping to eventually kill those functions in favor of only using the materialized data (they have horrible performance when used in joins because they're a "black box" for the query optimizer), in which case I'd prefer to keep the logic for creating them consolidated in one place.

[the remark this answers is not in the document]
cylewitruk: I'm actually hoping to eventually kill those functions in favor of only using the materialized data (they have horrible performance when used in joins because they're a "black box" for the query optimizer), in which case I'd prefer to keep the logic for creating them consolidated in one place.

signer/migrations/0011__materialized_chains.sql
```diff
@@ -0,0 +1,148 @@
...
+    ),
+    stacks AS (
+        (SELECT 
+            blocks.block_hash,
+            blocks.parent_hash,
+            blocks.block_height,
```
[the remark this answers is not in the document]
cylewitruk: Lol okay, for consistency's sake -- 

But in general I'm not sure what tooling you're using, but tbh the leading commas are kind of a pain to format, give tab warnings in my editor and imo are more difficult to read. Leading commas are usually used in scripts where you want to quickly be able to comment out part of the select without messing with fixing commas, but we're not doing that here.. so I'd definitely be interested in discussing the formatting of SQL.

signer/migrations/0011__materialized_chains.sql
cylewitruk: So a PK here isn't the easiest due to the nature of the table, like this is a materialization of multiple versions of other data. A PK would essentially need to be on all columns excluding heights, which has the inherent problem that PKs can't be NULL (which the Stacks blocks can be). We have the other check in the materialization function which is sort-of like our integrity check, at least for ensuring that the same chain tip can't be accidentally materialized twice.

Otherwise, the only thing that a PK really is is a clustered unique index, i.e. defines the ordering of storage on disk plus a unique index. So what I've done is added a unique index on `canonical_chain_id, bitcoin_block_hash, stacks_block_hash`, which will allow exactly _one_ NULL stacks block per bitcoin block (i.e. no stacks blocks for that bitcoin block), and the ordering on disk will be insert-order, which will implicitly be `canonical_chain_id ASC, bitcoin_block_hash DESC, stacks_block_hash DESC` due to how the materialization function works.

One thing to consider is that the UK adds ~4MB for the table for a recent (this morning) production db dump, for a context window of 1000 (bitcoin blocks -- 32,624 stacks blocks).

signer/migrations/0011__materialized_chains.sql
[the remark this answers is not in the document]
cylewitruk: Changed in [545214b](https://github.com/stacks-network/sbtc/pull/1171/commits/545214b76c2d53b3ff73a6c88cee274c2ad16eea)

signer/migrations/0011__materialized_chains.sql
```diff
@@ -0,0 +1,148 @@
...
+	existing_rows INTEGER;
+	new_run_id BIGINT;
+BEGIN
+	-- Get the next value for the run_id
+    new_run_id := nextval('sbtc_signer.canonical_chains_run_id_seq');
+
+	-- Check if rows exist with the given chain_tip
+    SELECT COUNT(*) INTO existing_rows
+    FROM sbtc_signer.canonical_chains
+    WHERE bitcoin_chain_tip = chain_tip;
```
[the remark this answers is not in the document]
cylewitruk: This is reworked in [545214b](https://github.com/stacks-network/sbtc/pull/1171/commits/545214b76c2d53b3ff73a6c88cee274c2ad16eea)

signer/migrations/0011__materialized_chains.sql
```diff
@@ -0,0 +1,148 @@
...
+-- table upon success, and `-1` if rows already exist for the given chain tip.
+CREATE OR REPLACE FUNCTION sbtc_signer.materialize_canonical_chains(chain_tip BYTEA, max_depth INT)
+RETURNS INTEGER AS $$
+DECLARE
+    rows_written INTEGER;
+	existing_rows INTEGER;
+	new_run_id BIGINT;
+BEGIN
+	-- Get the next value for the run_id
+    new_run_id := nextval('sbtc_signer.canonical_chains_run_id_seq');
```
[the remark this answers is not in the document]
cylewitruk: At least for databases which use separate sequences -- but this is now removed due to changes in the data model in [545214b](https://github.com/stacks-network/sbtc/pull/1171/commits/545214b76c2d53b3ff73a6c88cee274c2ad16eea)

signer/migrations/0011__materialized_chains.sql
cylewitruk: After [545214b](https://github.com/stacks-network/sbtc/pull/1171/commits/545214b76c2d53b3ff73a6c88cee274c2ad16eea) you'll see why I had this here :)

signer/migrations/0011__materialized_chains.sql
cylewitruk: Same, should be clearer now why these were here.

cylewitruk:
> Also, I'm not sure if we want two tables, one table with bitcoin blocks and another with bitcoin and stacks blocks. That might make sense, but maybe it's not worth it.

I think that keeping them consolidated at the moment makes the most sense -- at least unless/until we find a motivation for them to be separated. This solution is a little bit more OLAP mindset than OLTP, where we try to do more work up-front to avoid needing more work (joins) later.

signer/migrations/0011__materialized_chains.sql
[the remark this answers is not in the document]
djordon: Oh okay, makes sense. I did not know that they were a black box for the query optimizer.

signer/migrations/0011__materialized_chains.sql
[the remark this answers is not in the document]
djordon: Yeah, @matteojug has been asking to discuss this as well. Let's set aside a few moments.

cylewitruk:
One thing we need to consider is that functions like `get_last_key_rotation()` don't use a context window, which means that we wouldn't be able to use this together with a context window as the last key rotation might slide out-of-scope since it (should) happen so seldom. The idea with the context window in this case is to try to minimize both processing time and storage.

So maybe we should rethink that (and similar) functions separately -- i.e. if we should get it from the contract at each block instead, or have a separate materialization of context-window-free data -- or if we should just always materialize back to the last utxo or key rotation (whichever comes last) and let it take the space it needs?

Just thinking out loud..

signer/migrations/0011__materialized_chains.sql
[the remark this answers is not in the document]
cylewitruk: Yeah, a little more info [here](https://www.postgresql.org/docs/16/xfunc-optimization.html) and [here](https://www.postgresql.org/docs/16/xfunc-volatility.html) -- we can't use `IMMUTABLE` and I did try `STABLE` flags on the functions but it didn't help :(

signer/migrations/0011__materialized_chains.sql
```diff
@@ -0,0 +1,148 @@
...
+    ),
+    stacks AS (
+        (SELECT 
+            blocks.block_hash,
+            blocks.parent_hash,
+            blocks.block_height,
+            blocks.bitcoin_anchor
+        FROM sbtc_signer.stacks_blocks blocks
+        JOIN bitcoin ON blocks.bitcoin_anchor = bitcoin.block_hash
+        ORDER BY bitcoin.block_height DESC, bitcoin.block_hash DESC, blocks.block_height DESC, blocks.block_hash DESC
```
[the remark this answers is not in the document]
cylewitruk: Yeah the behavior is different, with the goal here to materialize the actual canonical chains in relation to the Bitcoin canonical chain. Since Stacks can only fork together with Bitcoin now, I'm not sure that this view would ever be different than what we want, especially if both the sBTC signer and Stacks node are using the same Bitcoin node.

We also have a different method of determining the canonical Bitcoin chain than either Stacks-core or Bitcoin (they both use the same calculation, we don't have the same tie-breaking mechanism), so I think there's also always the risk during forks/reorgs that the sBTC signer and the Stacks+Bitcoin nodes have different ideas of what the canonical Bitcoin tip is, with today's implementation, anyway.

But yeah, _if_ for some reason one of our queries wants the Stacks chain-tip regardless the Bitcoin chain then it could still get it from the data in the db.

djordon:
This is likely going to be important soonish, but for the time being let's close this and revisit when we have a bit more information about how things are performing after some of our recent changes.

Issue #1161, opened 2024-12-18
[Bug]: Return an error for deposit creation in Emily if deposit with the same details already exists


matteojug:
# Bug - Return error for deposit creation in Emily if deposit with the same details already exists

## 1. Description
 
Follow-up for https://github.com/stacks-network/sbtc/issues/1116, we should also return an error when re-creating a deposit with the same details.

PR #1179, opened 2024-12-23
[chore] Reject duplicate deposit requests when first one already accepted | confirmed | failed

MCJOHN974:
## Description

Closes: #1161

## Changes

Now `create_deposit` requests returns 409 (Conflict) if we already have same deposit accepted

## Testing Information

Added integration test for this.

Locally I run it the following way:
```
 docker compose -f docker/docker-compose.test.yml build   // rebuild emily
 docker compose --file docker/docker-compose.test.yml up.  // start emily
 // open new terminal
 cargo test --package emily-handler --test integration -- deposit::overwrite_deposit --exact --show-output --ignored // run the test
 ```

## Checklist:

### DIFF
--- a/emily/handler/tests/integration/deposit.rs
+++ b/emily/handler/tests/integration/deposit.rs
@@ -630,18 +630,12 @@ async fn overwrite_deposit(status: Status, should_reject: bool) {
     assert_eq!(response.bitcoin_txid, bitcoin_txid);
     assert_eq!(response.status, status);
 
-    if should_reject {
+    assert_eq!(
         apis::deposit_api::create_deposit(&configuration, create_deposit_body)
             .await
-            .expect_err(&format!(
-                "We should reject duplicate deposits, if old one is {:#?}",
-                status
-            ));
-    } else {
-        apis::deposit_api::create_deposit(&configuration, create_deposit_body)
-            .await
-            .expect("Received an error after making a valid get deposit api call.");
-    }
+            .is_err(),
+        should_reject
+    );
 
     let response = apis::deposit_api::get_deposit(
         &configuration,

### CODE REVIEW

File: emily/handler/src/api/handlers/deposit.rs [hunk 1]
```diff
@@ -213,7 +213,24 @@ pub async fn create_deposit(
...
+                && deposit.bitcoin_tx_output_index == body.bitcoin_tx_output_index
+                && deposit.status == Status::Accepted
+            {
+                let responce = json(&serde_json::json!({
+                    "error": "Conflict",
+                    "message": "This deposit already exists and accepted."
+                }));
+                return Ok(with_status(responce, StatusCode::CONFLICT));
+            }
+        }
```
Jiloc, line 231: I am not 100% sure, but I believe the duplicate check should go after the reorg check below
```
        let api_state = accessors::get_api_state(&context).await?;
        api_state.error_if_reorganizing()?;
```
@AshtonStephens what do you think?

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -213,7 +213,24 @@ pub async fn create_deposit(
         context: EmilyContext,
         body: CreateDepositRequestBody,
     ) -> Result<impl warp::reply::Reply, Error> {
-        // Set variables.
+        // Reject dups.
+        let (entries, _next_token) =
+            accessors::get_deposit_entries(&context, &Status::Accepted, None, None).await?;
```
Jiloc, line 218: I think we should prevent deposits from being overridden in any state.

One case we encountered was when a Pending deposit was created multiple times. Each time it was overridden, a new higher stacks_block_height was linked to the deposit. Sometimes the new stacks_block_height was higher than the block_height in which the deposit would be marked as Approved or Completed by the signers, and it would cause Emily to reject the updates.

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -213,7 +213,24 @@ pub async fn create_deposit(
...
+        let deposits: Vec<DepositInfo> = entries.into_iter().map(|entry| entry.into()).collect();
+        for deposit in deposits {
+            if deposit.bitcoin_txid == body.bitcoin_txid
+                && deposit.bitcoin_tx_output_index == body.bitcoin_tx_output_index
+                && deposit.status == Status::Accepted
+            {
+                let responce = json(&serde_json::json!({
```
Jiloc, line 225: typo: responce -> response

emily/handler/src/api/handlers/deposit.rs
Jiloc: A possible solution could be to use the function `accessors::get_deposit_entry` to get the specific entry whatever its status, instead of iterating all entries

Jiloc (commented):
Welcome to the Emily side of things @MCJOHN974 ! I left a few comments about changes required to handle duplicates deposits in all statuses

emily/handler/src/api/handlers/deposit.rs
matteojug, line 218: You can use `get_deposit_entry` to directly get the deposit

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -213,7 +213,24 @@ pub async fn create_deposit(
...
+                    "error": "Conflict",
+                    "message": "This deposit already exists and accepted."
+                }));
```
matteojug: Looking around we return `Error` directly (`Error::Conflict` here) and let the `impl Error` handle the required serialization.

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -213,7 +213,24 @@ pub async fn create_deposit(
...
+                return Ok(with_status(responce, StatusCode::CONFLICT));
+            }
+        }
+
+        // Create deposit.
```
matteojug, line 233: nit: the previous `// Set variables.` seems fine here, why the change?

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -213,7 +213,24 @@ pub async fn create_deposit(
         context: EmilyContext,
         body: CreateDepositRequestBody,
     ) -> Result<impl warp::reply::Reply, Error> {
-        // Set variables.
+        // Reject dups.
```
matteojug, line 216: nit: expand a bit, since it's not all the duplicates

emily/handler/tests/integration/deposit.rs
```diff
@@ -548,3 +548,81 @@ async fn update_deposits_updates_chainstate() {
         );
     }
 }
+
+#[cfg_attr(not(feature = "integration-tests"), ignore)]
+#[tokio::test]
+async fn overwrite_deposit() {
```
matteojug, line 554: Once we add the reject also for the other states, can we add test cases for those as well? Other than the `Fullfilment`, everything else can be the same I guess

matteojug (changes requested):

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -213,7 +213,24 @@ pub async fn create_deposit(
...
+        let (entries, _next_token) =
+            accessors::get_deposit_entries(&context, &Status::Accepted, None, None).await?;
+        let deposits: Vec<DepositInfo> = entries.into_iter().map(|entry| entry.into()).collect();
+        for deposit in deposits {
+            if deposit.bitcoin_txid == body.bitcoin_txid
+                && deposit.bitcoin_tx_output_index == body.bitcoin_tx_output_index
+                && deposit.status == Status::Accepted
```
matteojug, line 223: (answering to your Q) yeah we probably want to reject also for `Confirmed` and `Failed`. Not 100% sure about `Reprocessing` ones, but since we are not using that state at the moment let's not include it. @Jiloc WDYT?

emily/handler/src/api/handlers/deposit.rs
Jiloc: Ok, sorry for the confusion then! I didn't know we wanted to still have room to modify a pending deposit for possible malformed requests. 

The thing I would suggest is:
- if the request exists and is not in Pending then reject
- if the request exists and it is in Pending, override it, but preserve its original `stacks_block_height` and `stacks_block_hash` https://github.com/stacks-network/sbtc/blob/viktar/emily-returns-error-on-cloned-tx/emily/handler/src/api/handlers/deposit.rs#L238-L239

emily/handler/src/api/handlers/deposit.rs
Jiloc: The Reprocessing state is not currently in use. Everything that we were previously moving to Reprocessing is now moved back to Pending for simplicity. https://github.com/stacks-network/sbtc/pull/1074 

I am not sure if we'll ever re-introduce it. Given that we want to be able to modify pending deposits, the most correct approach would be to keep it overridable like the Pending state..

emily/handler/src/api/handlers/deposit.rs
Jiloc: This is the only official doc i found about it https://docs.google.com/document/d/13AN9GdLfPYKKPkeHNU4niq_CyJCDxFR7DGBLzLc5gxU/edit?tab=t.0#heading=h.9q7n8dly2bue

But in short - when Emily receives a conflicting chainstate update from the sidecar (e.g., a Stacks chainstate update for a block height Emily has already recorded, but with a different block hash), Emily interprets this as a chain reorganization. As a result, any updates received after that block are deemed invalid. Emily then deletes all deposit and withdrawal updates recorded after the block_height of the new block. (only the updates, the deposits/withdrawal records will still be in Dynamo)

emily/handler/src/api/handlers/deposit.rs
Jiloc: Great, looks good now! We just need to address the second part of the request: if the deposit already exists and is in a pending or reprocessing state, ensure that the values for [stacks_block_height and stacks_block_hash](https://github.com/stacks-network/sbtc/blob/viktar/emily-returns-error-on-cloned-tx/emily/handler/src/api/handlers/deposit.rs#L236-L237) are set as they are in the existing deposit

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -213,6 +213,21 @@ pub async fn create_deposit(
...
     ) -> Result<impl warp::reply::Reply, Error> {
+        // Reject if we already have a deposit with the same txid and output index and it is NOT pending or reprocessing.
+        let entry = accessors::get_deposit_entry(
+            &context,
+            &DepositEntryKey {
+                bitcoin_txid: body.bitcoin_txid.clone(),
+                bitcoin_tx_output_index: body.bitcoin_tx_output_index,
+            },
+        )
+        .await;
```
matteojug, line 224: As `get_deposit_entry` can fail for different reasons, we should propagate the error if it's not `Error::NotFound`, otherwise in case of unrelated errors (eg, dynamodb throttling) we may skip the check.

emily/handler/tests/integration/deposit.rs
```diff
@@ -548,3 +548,129 @@ async fn update_deposits_updates_chainstate() {
...
+    ] {
+        match status {
+            Status::Accepted | Status::Confirmed | Status::Failed => {
+                test_duplicate(status, true).await;
+            }
+            Status::Pending | Status::Reprocessing => {
+                test_duplicate(status, false).await;
+            }
+        }
+    }
```
matteojug, line 675: We can use `test_case` for this to improve readability (both of code and test results), eg https://github.com/stacks-network/sbtc/blob/83b316c5d26a3434a1b53d558bc7f899ce6c03f2/emily/handler/src/database/entries/deposit.rs#L651-L657

emily/handler/tests/integration/deposit.rs
```diff
@@ -548,3 +548,114 @@ async fn update_deposits_updates_chainstate() {
         );
     }
 }
+
+use test_case::test_case;
```
matteojug, line 552: nit: move up

emily/handler/tests/integration/deposit.rs
```diff
@@ -548,3 +548,114 @@ async fn update_deposits_updates_chainstate() {
...
+        apis::deposit_api::create_deposit(&configuration, create_deposit_body)
+            .await
+            .expect_err(&format!(
+                "We should reject duplicate deposits, if old one is {:#?}",
+                status
+            ));
+    } else {
+        apis::deposit_api::create_deposit(&configuration, create_deposit_body)
+            .await
+            .expect("Received an error after making a valid get deposit api call.");
```
matteojug, line 644: nit: wrong err message, also we could simplify this block by `assert_eq` that `create_deposit(...).await.is_err()` matches `should_reject`.
Note that we already have the info related to which test case is running (eg, `emily-handler::integration deposit::overwrite_deposit::should_not_reject_pending_duplicate`), so we would still have everything we need to find the failing case.

matteojug (approved):

PR #1180, opened 2024-12-30
fix: add default '200' handler for signer api

cylewitruk:
## Description

The sBTC signers stalled their Stacks nodes in prod when the nodes emitted "attachment" events (which is a known bug on the node-side), and the sBTC signers couldn't handle it.

## Changes

This adds a default handler to the axum server which will simply return 200 OK for unknown routes. I don't necessarily think that we should keep this around long-term, or rather that stacks events maybe should be behind a `/stacks/..` path where we have a default handler. But this change in particular is intended to be a band-aid until the event-filtering for `attachment` events are properly handled in stacks-core.

## Testing Information

New test added which checks before/after these changes to ensure that the correct status code is returned.

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -732,7 +732,7 @@ dependencies = [
  "rustversion",
  "serde 1.0.203",
  "sync_wrapper 0.1.2",
- "tower",
+ "tower 0.4.13",
  "tower-layer",
  "tower-service",
 ]
@@ -764,7 +764,7 @@ dependencies = [
  "serde_path_to_error",
  "sync_wrapper 1.0.1",
  "tokio",
- "tower",
+ "tower 0.4.13",
  "tower-layer",
  "tower-service",
  "tracing",
@@ -3207,7 +3207,7 @@ dependencies = [
  "serde_json",
  "tokio",
  "tokio-stream",
- "tower",
+ "tower 0.4.13",
  "tracing",
 ]
 
@@ -6238,6 +6238,7 @@ dependencies = [
  "toml_edit 0.22.22",
  "tonic",
  "tonic-build",
+ "tower 0.5.2",
  "tower-http",
  "tracing",
  "tracing-attributes",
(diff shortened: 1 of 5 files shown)


matteojug (commented):
Since we know the issue is with a specific endpoint (the new attachments one), I would prefer to add an handler for it (like we did for the sidecar in https://github.com/stacks-network/sbtc/pull/1107/files#diff-dda86696b10335c16238c0db02cd3fa8bee7a7e2280c08fb3a1a5ca07442b892R78 ) instead of defaulting to ok for everything, with the rationale that we may prefer to break things if we get more unexpected events, instead of silently ignoring them. WDYT?

matteojug:
Addressed in https://github.com/stacks-network/sbtc/pull/1185

cylewitruk:
> Since we know the issue is with a specific endpoint (the new attachments one), I would prefer to add an handler for it (like we did for the sidecar in https://github.com/stacks-network/sbtc/pull/1107/files#diff-dda86696b10335c16238c0db02cd3fa8bee7a7e2280c08fb3a1a5ca07442b892R78 ) instead of defaulting to ok for everything, with the rationale that we may prefer to break things if we get more unexpected events, instead of silently ignoring them. WDYT?

@matteojug just to reply since I was away --
Since we know that we're explicitly interested only in specific events, my preference is actually the opposite, with the rationale that we shouldn't need to explicitly account for bugs (current or future) in the Stacks node's event filtering. As long as we're getting the events we want, I don't think the sBTC signer should risk blocking the Stacks node (and likely requiring a new hotfix) if the node has a bug and incorrectly sends us events we don't care about. But like I wrote above, I would prefer in that case that we moved the stacks endpoints to `/stacks/*` and had a default 200 handler there, not for the whole API, but that would be `breaking-local`.

matteojug:
I was more concerned about legittimate changes in the behaviour of the event observer (eg, sending the events to a different endpoint) that with a default 200 we would just ignore. I agree that we don't want to account for bugs, but to err on the safe side if anything where to change in the event observer imho it's better if we stop everything and reassess how we need to handle new things, instead of potentially losing important information (and act on wrong information).

About moving the event observer to a different prefix, +1 on that (assuming we can do that as you mentioned).

PR #1185, opened 2025-01-07
fix: add handler for new attachments

matteojug:
## Description

The sBTC signers stalled their Stacks nodes in prod when the nodes emitted "attachment" events (which is a known bug on the node-side), and the sBTC signers couldn't handle it.

## Changes

This adds a dummy endpoint responding with 200 OK.

## Testing Information

Add simple test, and tested that everything else still works via devenv demo script.

## Checklist:


djordon (approved):
Looks good

Issue #1187, opened 2025-01-07
[Bug]: Enforce decoding order for Signed messages


djordon:
# Bug - Enforce decoding order for Signed messages

## 1. Description

Signers are supposed to sign all non-signature bytes of the messages that they send, but it is currently possible for a signer to send messages that are not "fully" signed. There is no direct exploit of this bug but it opens the possibility of an attack vector so it needs to be closed.

### 1.1 Context & Purpose

Signers always send signed messages over the peer-to-peer network, where the wire format is a protobuf encoded message. Unfortunately, the protobufs spec is not strict enough for our serialization and deserialization needs so we rely on a specific "canonical" protobuf format (the "sbtc protobuf codec") in the signer codebase. And we use the fact that messages are serialized using the "sbtc protobuf codec" for proper deserialization of signed messages but do not reject all signed messages that violate the codec.

## 2. Technical Details:

Protobuf messages are supposed to be encoded in field tag order. This is the default behavior of the code generated from our protobuf library, [prost](https://github.com/tokio-rs/prost). However, it is possible for a malicious signer to modify the code and break the codec. The change needed is to enforce the tag order expectation when deserializing fields of `Signed` protobuf message.

### 2.1 Acceptance Criteria:
   - [ ] Each field of a `Signed` message is deserialized in tag order, and messages that are not encoded correctly are rejected.

PR #1188, opened 2025-01-07
fix: enforce codec deserialization

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1187.

I don't consider this protocol breaking, since we are enforcing something that was already supposed to be.

## Changes

1. Add a bit more documentation of the sbtc protobuf codec.
2. Enforce the tag order requirement of `Signed` protobuf messages.
3. Go through encoding/decoding code path in tests.

## Checklist:


signer/src/codec.rs
```diff
@@ -1,13 +1,32 @@
 //! # Canonical encoding and decoding for the sBTC signer
 //!
-//! The purpose of this module is to define how to encode and decode
-//! signer messages as byte sequences.
+//! The purpose of this module is to define how to encode and decode signer
+//! messages as byte sequences.
 //!
-//! This is achieved by
+//! ## Codec specification
```
matteojug, line 6: nit: while point 2 & 3 are enforced on the read path by prost (IIUC), that is not the case for point 1 (in general, other than for `Signed<SignerMessage>`; eg in nested messages) and maybe also for the last part of point 4.
Should we specify it, to avoid giving the impression that we have such guarantees in place?

signer/src/ecdsa.rs
```diff
@@ -587,4 +609,119 @@ mod tests {
 
         assert_eq!(signed_message_v2, original_proto);
     }
+
+    #[test]
+    fn poc_bypass_sign2() {
```
matteojug, line 614: nit: better name?

signer/src/ecdsa.rs
```diff
@@ -587,4 +609,119 @@ mod tests {
...
+        // tags. If we were, then the signature field would be over the
+        // message's `type_tag`, bypassing the verification of the signature.
+        let hasher = sha2::Sha256::new_with_prefix(msg.type_tag());
+        let empty_prefix_hash = secp256k1::Message::from_digest(hasher.finalize().into());
+
+        let signature = private_key.sign_ecdsa(&empty_prefix_hash);
+        let empty_prefix_sign: Vec<u8> =
+            proto::EcdsaSignature::from(signature).encode_length_delimited_to_vec();
+
+        let mut hack1 = Vec::new();
```
matteojug, line 675: nit: better name? also `hack2`

matteojug (approved):

signer/src/ecdsa.rs
djordon: Agreed, I'll think of something.

signer/src/codec.rs
djordon: I thought about mentioning it in these docs (and did here https://github.com/stacks-network/sbtc/pull/1188/commits/8fb948b63ce760f93d2194b452b2ade40be54315 before I removed it). Yeah, I'll mention it in these docs to avoid giving the wrong impression.

signer/src/ecdsa.rs
djordon: Roger, will update.

signer/src/codec.rs
djordon: I think I may have partially misunderstood what you were saying. So I believe `prost` generates code that enforces all 4 points all of the time, but you're right, we do not have such guarantees in place. We do not test against it... maybe we should though... brb.

signer/src/ecdsa.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/1188/commits/a48c56450cc061cf738dde7e868bb8171beb6049

signer/src/ecdsa.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/1188/commits/b587d01aae07b48128c710896bcb519f82cd3c1e.

signer/src/codec.rs
djordon: Added tests in https://github.com/stacks-network/sbtc/pull/1188/commits/07f8d0089062e5ac535571b5a665c3776c8af1c0.

matteojug: Sorry, I should have worded it better; having those extra tests is def nice, but I was referring to something slightly different: while it's true (and now tested) that prost enforces those requirements when serializing the messages, we are not enforcing some of those requirements when deserializing:
 - we don't check tag order other than for top level `Signed<SignerMessage>`
 - we don't check that we don't deserialize empty values -- although this may be just in the proto spec so nothing else to do here

While the above shouldn't be important outside the digest/signature computation, I'm worried that as is written it may give the impression that (eg) the tag order check is enforced on deserialization for *any* message.

djordon: Oh, gotcha. I'll make sure to note that the current deserialization code does not enforce these rules.

djordon: Updated in https://github.com/stacks-network/sbtc/pull/1188/commits/7224280045181eb0022d5476982667a5351702ce

matteojug (approved):

Issue #1116, opened 2024-12-13
[Bug]: Ensure integrity of deposit requests submitted


djordon:
# Bug - Ensure integrity of deposit requests

## 1. Description

When someone submits deposit request data to Emily, we do not have a way of ensuring that the sender is the same entity that created the deposit UTXO. This means that someone can create properly formed deposit requests to Emily that do not match any actual deposits on bitcoin, with the outcome being that actual deposits will be deemed invalid.

### 1.1 Context & Purpose

We may want to make some design decisions to increase the integrity of data submitted to Emily, so that we know that data stored there comes from the correct source.

## 2. Technical Details:

There are three solutions that come to mind and all are partly undesirable:
1. Require a signature in the deposit request object sent to Emily. The signature must be verifiable with a public key that locked one of the inputs into the deposit transaction.
2. Put all deposit information on chain as part of an `OP_RETURN` output.
3. Give Emily access to bitcoin-core to do a full verification of the deposit request.
4. Require a stacks transaction from an account that mirrors one of the `scriptPubKeys` locking an input from the deposit transaction.

Option (1) Is the easiest lift for us, but seems impractical: what is going to construct this signature? I don't see many wallets adding such functionality. Option (2) is doable, but it also requires wallets to support the creation of `OP_RETURN` outputs. Option (3) is like option (1) but better. Option (4) is quite cumbersome.

I think (2) is the way to go, and allows us to remove functionality from Emily. We can fit the required data into an `OP_RETURN` output if we are creative.

### 2.1 Acceptance Criteria:
   - [ ] It not possible for someone other than the depositor to inform the sBTC system of their deposits.

## 4. Appendix

Option (2) above is more tricky than it looks, and it requires us to be a little creative because we only have 80 bytes to work with in an `OP_RETURN` output. A straightforward porting of the data in our current deposit transactions will show that we don't have enough space. So an alternative layout for the deposit is something like this:

```text
0                    20             21        23               44                    64
|--------------------|--------------|---------|----------------|---------------------|
  signer pubkey hash   sbtc version   max fee   stacks address   reclaim pubkey hash
```

Meaning:
1. "signer pubkey hash" is the HASH160 of the signers x-only aggregate key, 
2. "sbtc version" is just a version byte for the layout.
3. "max fee" is the users max fee interpreted as multiples of 10000 sats.
5. "stacks address" is the recipient address. It is always a standard address.
[...]

PR #1197, opened 2025-01-09
fix: emily validate deposits

Jiloc:
## Description

Closes: #1116, 

## Changes
- add a new `transaction_hex` mandatory field to the `create_deposit` endpoint 
- add validation on all request fields and the new one
- add validation to check if the raw transaction data is an sbtc transaction and its content corresponds to the txid, deposit and reclaim scripts
- extract deposit amount from the tx  - previously we were hardcoding a 0. 
~~- add validation to check if the deposit amount is withint the per-deposit limits:~~
     ~~- the validation is simple by design. We want to avoid requests that are clearly malicious to be accepted by Emily, but the refined validation (e.g. amount - paid fee > dust fee) are still expected from the signers.~~
~~- save recipients with their real string value instead of their hex representation. e.g. `SP5ZH8581RX20N9MJXZ0V7G0QKCBKGDKM4EK3700` insead of `05160bf8a0a80e3a205534977e0d9e00bcd8b9c1b3a1`~~
 
## Testing Information
- added real txs fixtures and unit tests for the validation
- modified `tx_setup` to create transactions with mulitple deposits
- updated unit and integration tests
 
## Checklist:


emily/handler/src/api/handlers/deposit.rs
```diff
@@ -245,11 +248,19 @@ pub async fn create_deposit(
             Err(e) => return Err(e),
         }
 
-        let status = Status::Pending;
+        let tx = (&body).try_into()?;
+        let limits = accessors::get_limits(&context).await?;
+        body.validate(&tx, &limits)?;
```
djordon, line 253: Might be better to have `validate` return a `DepositInfo` object and use that to get the deposit amount.

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -43,8 +53,89 @@ pub struct CreateDepositRequestBody {
...
+    type Error = Error;
+
+    fn try_into(self) -> Result<Transaction, Error> {
+        parse_hex(
+            &self.transaction_hex,
+            "invalid transaction_hex",
+            encode::deserialize_hex,
+        )
+    }
+}
```
djordon, line 138: Having `CreateDepositRequestBody` implement `TryInto<Transaction>` doesn't unintuitive since it's not really a `bitcoin::Transaction`. Instead, implement a function on the `CreateDepositRequestBody` that returns the transaction as a result.

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -43,8 +53,89 @@ pub struct CreateDepositRequestBody {
...
+    F: Fn(&str) -> Result<T, E>,
+{
+    parser(input).map_err(|_| Error::HttpRequest(StatusCode::BAD_REQUEST, error_msg.to_string()))
+}
+
+impl CreateDepositRequestBody {
+    /// Validates that the deposit request is valid.
+    /// This includes validating the request fields, if their content matches the transaction
+    /// and if the amount is within the limits.
+    pub fn validate(&self, tx: &Transaction, limits: &Limits) -> Result<(), Error> {
```
djordon, line 76: Consider converting the `transaction_hex` field to a transaction object within this function. We don't really need to take the `bitcoin::Transaction` as an input, we already basically have it. If we return the `sbtc::deposits::DepositInfo` then we will have all the information that we need down the line.

emily/handler/src/api/handlers/deposit.rs
Jiloc: Great idea thanks! This will help a lot even the integration tests to have access to `DepositInfo`

Jiloc: fixed in 447d36a1c7ca0dcacf516dfce3fa0d1d3a7a6b98

emily/handler/src/api/models/deposit/requests.rs
Jiloc: fixed in 447d36a1c7ca0dcacf516dfce3fa0d1d3a7a6b98

emily/handler/src/api/models/deposit/requests.rs
Jiloc: fixed in 447d36a1c7ca0dcacf516dfce3fa0d1d3a7a6b98

Jiloc:
We'll keep this PR in Draft till after the rotate-key release

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -43,6 +53,82 @@ pub struct CreateDepositRequestBody {
...
+        let amount = tx
+            .tx_out(self.bitcoin_tx_output_index as usize)
+            .map_err(|_| {
+                Error::HttpRequest(
+                    StatusCode::BAD_REQUEST,
+                    "invalid bitcoin_output_index".to_string(),
+                )
+            })?
+            .value
+            .to_sat();
```
djordon, line 109: nit: you can remove this if you move up the `validate_tx` call, which I think also has a dedicated error message for this check:
```suggestion
        let deposit_info = deposit_req
            .validate_tx(&tx)
            .map_err(|e| Error::HttpRequest(StatusCode::BAD_REQUEST, e.to_string()))?;
```
Then below just do
```rust
...
if deposit_info.amount < min {
    ...
```

Jiloc: I noticed this, but I preferred the current approach because I wanted the deposit's amount to be validated before the transaction matching, and I found the error message for the existing validation (`could not get outpoint f75cb869600c6a75ab90c872435da38d54d53c27afe5e03ac7dedae7822958de:10 from BTC transaction: invalid output index`) too verbose. Said that, I am aware that my motivations are just a small subjective preference, so I am open to change if you think the other approach would be clearer!

fabergat:
The CDK part looks good to me

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -55,6 +65,82 @@ pub struct CreateDepositRequestBody {
...
+            ));
+        }
+
+        let cap = limits.per_deposit_cap.unwrap_or(Amount::MAX_MONEY.to_sat());
+        if amount > cap {
+            return Err(Error::HttpRequest(
+                StatusCode::BAD_REQUEST,
+                format!("deposit amount exceeds cap ({})", cap),
+            ));
+        }
```
matteojug, line 138: I'm a bit worried about validating the limits here (other than the dust), as there's a time difference between when the bridge checks for the limits and when Emily does. And in particular, if the request is not on emily the bridge cannot offer a recovery via reclaim. So maybe it's better to not enforce the limits checks in here.

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -91,3 +177,117 @@ pub struct UpdateDepositsRequestBody {
...
+        include_str!("../../../../tests/fixtures/create-deposit-mismatch-txid.json");
+
+    const CREATE_DEPOSIT_MISMATCH_RECLAIM_SCRIPT: &str =
+        include_str!("../../../../tests/fixtures/create-deposit-mismatch-reclaim-script.json");
+
+    const CREATE_DEPOSIT_MISMATCH_DEPOSIT_SCRIPT: &str =
+        include_str!("../../../../tests/fixtures/create-deposit-mismatch-deposit-script.json");
+
+    mod helpers {
+        use super::*;
```
matteojug, line 214: nit: I haven't seen those around, we usually just throw the helpers in the same `tests` module.

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -55,6 +65,82 @@ pub struct CreateDepositRequestBody {
...
+/// Deposit amounts that is less than this amount will be rejected by the
+/// smart contract.
+pub const DEPOSIT_DUST_LIMIT: u64 = 546;
+
+fn parse_hex<T, F, E>(input: &str, error_msg: &str, parser: F) -> Result<T, Error>
+where
+    F: Fn(&str) -> Result<T, E>,
+{
+    parser(input).map_err(|_| Error::HttpRequest(StatusCode::BAD_REQUEST, error_msg.to_string()))
+}
```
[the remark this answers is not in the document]
Jiloc: I like both suggestions :)

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -55,6 +65,82 @@ pub struct CreateDepositRequestBody {
     pub reclaim_script: String,
     /// Deposit script.
     pub deposit_script: String,
+    /// The raw transaction hex.
+    pub transaction_hex: String,
```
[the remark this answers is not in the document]
matteojug: Are we comparing txid and the resulting txid from `transaction_hex` somewhere? I couldn't find it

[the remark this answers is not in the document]
Jiloc: https://github.com/stacks-network/sbtc/blob/9e1c3a62b400419e9d9b87537e38c2f467fa9fc4/emily/handler/src/api/models/deposit/requests.rs#L144-L146

It's the validate_tx function from the sbtc crate that matches all the sbtc deposits info with the tx info

emily/handler/src/api/models/deposit/requests.rs
matteojug: I'm wondering if we want to keep the dust limit check or not (I know previously I said so, sorry 🙏 ): it's kinda a constant of the system so I would be inclined to keep it, but OTOH there's the usual: what if someone else (the bridge, some external wallet) doesn't enforce it? Then we are back to the usual _more work for the signers but users can reclaim easily_ thingy

Jiloc: 🤣 no worries, it makes sense to me. Any kind of enforcement on Emily's side makes sense only if the create deposit call could be sent and validated before the bitcoin TX is sent, but given that it's not the case..

emily/handler/src/api/models/deposit/requests.rs
djordon: Oh okay, no worries. What is here is fine then.

emily/handler/src/api/models/deposit/requests.rs
Jiloc: removed in https://github.com/stacks-network/sbtc/pull/1197/commits/f8894f27d7143d9af6789f88b31acd4751527d01

emily/handler/tests/integration/deposit.rs
```diff
@@ -279,36 +321,46 @@ async fn get_deposits() {
...
 
+    let amounts = vec![DEPOSIT_AMOUNT_SATS; 4];
     // Setup test deposit transaction.
-    let DepositTxnData {
-        recipient: expected_recipient,
-        reclaim_script,
-        deposit_script,
-    } = DepositTxnData::new(DEPOSIT_LOCK_TIME, DEPOSIT_MAX_FEE, DEPOSIT_AMOUNT_SATS, 0);
+    let deposit_txn_data =
+        (0..2).map(|_| DepositTxnData::new(DEPOSIT_LOCK_TIME, DEPOSIT_MAX_FEE, &amounts));
```
matteojug, line 329: nit: this seems a bit convoluted, why not create this directly in the loop below?

signer/src/block_observer.rs
```diff
@@ -822,14 +822,14 @@ mod tests {
...
-        let tx_setup0 = sbtc::testing::deposits::tx_setup(lock_time, max_fee, amount);
+        let tx_setup0 = sbtc::testing::deposits::tx_setup(lock_time, max_fee, &[amount]);
         let deposit_request0 = CreateDepositRequest {
             outpoint: bitcoin::OutPoint {
                 txid: tx_setup0.tx.compute_txid(),
                 vout: 0,
             },
-            deposit_script: tx_setup0.deposit.deposit_script(),
-            reclaim_script: tx_setup0.reclaim.reclaim_script(),
+            deposit_script: tx_setup0.deposits.first().unwrap().deposit_script(),
```
matteojug, line 831: nit: we should probably add a `first_deposit` and some other util to `TxSetup` to streamline stuff like this -- not in this PR

matteojug (approved):
LGTM

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -55,6 +64,65 @@ pub struct CreateDepositRequestBody {
...
+}
+
+/// This is the dust limit for deposits in the sBTC smart contracts.
+/// Deposit amounts that is less than this amount will be rejected by the
+/// smart contract.
+pub const DEPOSIT_DUST_LIMIT: u64 = 546;
```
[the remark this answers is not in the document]
Jiloc: removed in https://github.com/stacks-network/sbtc/pull/1197/commits/e9d40ad084cfb5fd39f426b9d21c0f66a603332f

signer/tests/integration/setup.rs
```diff
@@ -90,6 +91,8 @@ pub struct TestSweepSetup {
     /// threshold is the bitcoin signature threshold, which for v1 matches
     /// the signatures required on stacks.
     pub signatures_required: u16,
+    /// The deposit transaction
+    pub deposit_tx: Transaction,
```
djordon, line 95: This is already in `deposit_tx_info`, which has a field for the raw transaction.

djordon (commented):
I partially skimmed this over, but I will to give this one more full fresh look today.

djordon (approved):
Looks good 🚀

signer/tests/integration/setup.rs
Jiloc: removed in https://github.com/stacks-network/sbtc/pull/1197/commits/4858deb32900306a63459eb12f38e66f1ee82f6b

Issue #1200, opened 2025-01-09
[Feature]: Signers allow second DKG after specific block height


djordon:
# Feature - Signers allow second DKG after specific block height

## 1. Description

The signers currently allow DKG only if it hasn't run before. The change is for the signers to allow DKG a second time if it hasn't run twice and only allow a second run if the current bitcoin block height is greater than some configured block height.

### 1.1 Context & Purpose

We want to support full DKG. But that is difficult so we are tackling it in phases. Phase 0 was running DKG once. Phase 1 was supporting DKG a second time, and this ticket is part of phase 1.

## 2. Technical Details:

The current check stops the signer from participating in DKG if there exists keys in the `dkg_shares` table, since that indicates that DKG has succeeded. We need to change this to stop the signer from participating in DKG if the number of rows in the DKG shares table greater than 1 or if the current block height is less than the some configured block height (the one from https://github.com/stacks-network/sbtc/issues/1199).

### 2.1 Acceptance Criteria:
   - [ ] In `TxSignerEventLoop`, the signer will participate in DKG if it has never been run, or if it has run once and the current block height is greater than the one that they have "configured".

Although the above acceptance criteria mentions the configured variable, we can settle for some variable in the `TxSignerEventLoop` struct that gets set to whatever variable if the config work isn't done. That way we can do this task in parallel to https://github.com/stacks-network/sbtc/issues/1199.

## 3. Related Issues and Pull Requests (optional):

This is related to https://github.com/stacks-network/sbtc/issues/1199.

Issue #1201, opened 2025-01-09
[Feature]: Add bitcoin block height configuration variable for second DKG


djordon:
# Feature - Add bitcoin block height configuration variable for second DKG

## 1. Description

We will attempt to run DKG after a certain bitcoin block height has been found. This block height will be a configuration variable set in the signers config so that we can test this out in different variables.

### 1.1 Context & Purpose

We want to support full DKG. But that is difficult so we are tackling it in phases. Phase 0 was running DKG once. Phase 1 was supporting DKG a second time, and this ticket is part of phase 1.

## 2. Technical Details:

Add a new required signer config variable for the second DKG run. For the example production configs, set the bitcoin block height to 879578, which is ~2025-01-16 14:30:00 UTC. Or we could use block height 880088, which is ~2025-01-20 15:40:00 UTC

### 2.1 Acceptance Criteria:
   - [ ] We have a required configuration variable for the second DKG run.

## 3. Related Issues and Pull Requests (optional):

This relates to https://github.com/stacks-network/sbtc/issues/1200, https://github.com/stacks-network/sbtc/issues/1199 and https://github.com/stacks-network/sbtc/issues/1152, which together add support for the second DKG run.

PR #1205, opened 2025-01-10
feat: #1200 feature signers allow second dkg after specific block height

cylewitruk:
## Description

Closes: #1200 

## Changes

Changes the check in the `handle_wsts_message` method for `DkgBegin` messages in the transaction signer to conditionally allow a second DKG run if:

- There has not been a DKG run yet (i.e. no entries in the signer's `dkg_shares` db table), **OR**
- If DKG has been run exactly once (one entry in the signer's `dkg_shares` db table) **AND** a DKG re-run height is configured **AND** the current bitcoin chain tip height is >= the configured re-run height

This copies a snippet of the config code in #1201 to be able to retrieve the config value for the DKG re-run height.

## Testing Information

This currently lacks additional tests because my integration env is borked for some reason; looking into that now.

## Checklist:

### DIFF
--- a/signer/src/transaction_signer.rs
+++ b/signer/src/transaction_signer.rs
@@ -900,21 +900,13 @@ async fn assert_allow_dkg_begin(
                 "DKG rerun height has been met and we are below the target number of rounds; proceeding with DKG"
             );
         }
-        (current, _, None) if current >= 1 => {
+        // Note that we account for all (0, _, _) cases above (i.e. first DKG round)
+        (_, _, None) => {
             tracing::warn!(
                 ?dkg_min_bitcoin_block_height,
                 %dkg_target_rounds,
                 dkg_current_rounds = %dkg_shares_entry_count,
                 "attempt to run multiple DKGs without a configured re-run height; aborting"
-            );
-            return Err(Error::DkgHasAlreadyRun);
-        }
-        _ => {
-            tracing::warn!(
-                ?dkg_min_bitcoin_block_height,
-                %dkg_target_rounds,
-                dkg_current_rounds = %dkg_shares_entry_count,
-                "multiple DKG shares already exist; aborting"
             );
             return Err(Error::DkgHasAlreadyRun);
         }
--- a/signer/src/transaction_signer.rs
+++ b/signer/src/transaction_signer.rs
@@ -879,6 +879,7 @@ async fn assert_allow_dkg_begin(
                 tracing::warn!(
                     ?dkg_min_bitcoin_block_height,
                     %dkg_target_rounds,
+                    dkg_current_rounds = %dkg_share_entry_count,
                     "The target number of DKG shares has been reached; aborting"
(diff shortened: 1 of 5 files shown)


cylewitruk:
Still needs tests but should be OK for review (existing tests pass) -- so don't approve yet

signer/src/transaction_signer.rs
```diff
@@ -499,15 +499,52 @@ where
...
+                    dkg_rerun_min_block_height,
+                ) {
+                    (0, _, _) => {
+                        tracing::info!("no DKG shares exist; proceeding with DKG");
+                    }
+                    (current, target, Some(dkg_rerun_height)) => {
+                        if current >= target.get() {
+                            tracing::warn!(
+                                "The target number of DKG shares has been reached; aborting"
+                            );
```
matteojug, line 531: nit: adding `current` and `target` in the log could help in debugging (eg, someone forgot to update their confing); also in the `warn` right below.

djordon (commented):
Looks good, but can we separate the logic in the transaction_signer into a separate function?

signer/src/transaction_signer.rs
cylewitruk: Addressed in [68c948d](https://github.com/stacks-network/sbtc/pull/1205/commits/68c948d30d76e2f33652751eab42a2b80d50c0cb) + [a1f93f4](https://github.com/stacks-network/sbtc/pull/1205/commits/a1f93f4a4497213cd225c1106e18126170ef252e)

cylewitruk:
> Looks good, but can we separate the logic in the transaction_signer into a separate function?

I moved the assertion to a separate method now (primarily so I can use a simpler unit test instead of a bigger integration-y test). For a future task we should probably make each of the match arms their own "handle" method.

cylewitruk:
Tests added now 👍

signer/src/transaction_signer.rs
```diff
@@ -846,6 +839,90 @@ where
...
+                return Err(Error::DkgHasAlreadyRun);
+            }
+            tracing::info!(
+                ?dkg_min_bitcoin_block_height,
+                %dkg_target_rounds,
+                dkg_current_rounds = %dkg_shares_entry_count,
+                "DKG rerun height has been met and we are below the target number of rounds; proceeding with DKG"
+            );
+        }
+        (1, _, None) => {
```
matteojug, line 903: qq: why the `1` here? I guess it's related to having those last two cases as separate ones; is it just for better conf debugging?

matteojug (approved):

signer/src/transaction_signer.rs
cylewitruk: Actually a miss -- updated in [ec63981](https://github.com/stacks-network/sbtc/pull/1205/commits/ec63981d4b083be5819e8bfc47c1b78d392b584e)

Issue #1202, opened 2025-01-10
[Feature]: New signer state machine for each input


djordon:
# Feature - New signer state machine for each input

## 1. Description

Whenever a signer gets a sign request over a particular message, the signers create a state machine if its for a transaction ID that they have never seen before. But this ties the signing shares to transaction when it should be tied to the input that is being signed since a transaction can contained inputs locked by different keys.

### 1.1 Context & Purpose

We need to support running deposits locked with different aggregate keys.

## 2. Technical Details:

When a signer receives a `NonceRequest` message they should create a new state machine associated with the message that they being are asked to sign. Then, when the signer receives a subsequent `SignatureShareRequest` message, they provide their signature shares if a state machine exists for the message to be signed and exit otherwise. At the end of they should delete the state machine from memory. This is similar to the current flow, but oriented around messages instead of txids.

### 2.1 Acceptance Criteria:

- [ ]  The signers create new state machines for each message that they sign

## 3. Related Issues and Pull Requests (optional):

Issue #1193, opened 2025-01-09
[Bug]: Cap the number of signer state machines to avoid OOM


djordon:
# Bug - Cap the number of signer state machines to avoid OOM

## 1. Description

The signers create a new state machine for each transaction that needs signing. These state machines are never removed and remain in memory for the lifetime of the application.

### 1.1 Context & Purpose

We should avoid OOMing, so let's make sure that they can only take up a maximum amount of memory.

## 2. Technical Details:

There are a few things that we could do but the best is probably switch to using a map data structure that caps the number of entries at any given time.

### 2.1 Acceptance Criteria:
   - [ ] It is not possible to OOM the signers because we keep too many signer state machines.

## 4. Appendix

As for the data structure that caps the number of signer state machines, there is an [LruCache](https://docs.rs/lru/0.12.5/lru/struct.LruCache.html) type in the `lru` crate. There are a few other well used crates for caches as well, but this seems like the oldest and more widely used. We could also implement one ourselves with something like
```rust
struct TimeCache {
    entries: BTreeMap<Id, SignerStateMachine>,
    inserted_at: BTreeMap<std::time::Instant, Id>,
    max_entries: NonZeroUsize,
}
```
where `Id` is the what we use to identify a `SignerStateMachine`. In this formulation, we always remove the oldest entry from the map when we attempt to insert a new object.

With whatever type we go with, we need to make sure that the capacity is enough to support the maximum number of signer state machines that could be created during a coordinator's tenure, which is capped following https://github.com/stacks-network/sbtc/pull/1181. We need that to support concurrent bitcoin signing rounds sent from the same coordinator.

PR #1209, opened 2025-01-13
feat: always create new state machines

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1202 and closes https://github.com/stacks-network/sbtc/issues/1193

## Changes

* Cap the number of signer state machines in use at one time to the maximum number of signing rounds that we can have during the tenure of one bitcoin block (plus 1 for DKG).
* Create a new state machine for each bitcoin input that the signer will sign. These state machines are identified by the signature hash to sign.
* Identify DKG signer state machines by the bitcoin chain tip when they received the DKG begin message.

## Testing Information

I added tests that a new state machines will be created for a "valid" sighashes, and for the chain tip when DKG is run.

## Checklist:


signer/src/main.rs
```diff
@@ -315,14 +316,27 @@ async fn run_transaction_signer(ctx: impl Context) -> Result<(), Error> {
     let config = ctx.config().clone();
     let network = P2PNetwork::new(&ctx);
 
+    // This is the maximum number of inputs that can be signed during the
+    // tenure of a single bitcoin transaction plus one for DKG.
+    let max_state_machines = signer::MAX_MEMPOOL_PACKAGE_TX_COUNT
+        .saturating_mul(ctx.config().signer.max_deposits_per_bitcoin_tx.get() as u64 + 1)
```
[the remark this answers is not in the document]
matteojug: I'm also a bit worried about this `min`: `MAX_SIGNER_STATE_MACHINES` has definitely enough buffer to be reasonably safe, while with this exact count we may be forgetting something (or some funny race condition may mess with us) without any slack.

signer/src/lib.rs
```diff
@@ -108,6 +108,16 @@ pub const DEPOSIT_DUST_LIMIT: u64 = 546;
 /// <https://github.com/bitcoin/bitcoin/blob/v25.0/src/policy/policy.h#L60-L61>
 pub const MAX_MEMPOOL_PACKAGE_SIZE: u64 = 101000;
 
+/// This is an upper bound on the number of signer state machines that we
+/// "could" need if we wanted to sign all inputs in parallel and running
+/// DKG.
+///
+/// The smallest vsize of a deposit input is 91 vbytes, so if the entire
```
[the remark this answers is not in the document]
djordon: I need to change these from the vsize for deposits to the minimum vsize of any input that the signers could sign for. I'll make that a constant, since that will be a bitcoin limit instead of an sbtc limit.

signer/src/main.rs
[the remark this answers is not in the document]
djordon: Yeah, good point. I'll build in some slack reasonable slack or just use `MAX_SIGNER_STATE_MACHINES`.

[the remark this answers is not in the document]
cylewitruk: Tbh I think an x2 or x3 would probably be sufficient, just in-case there's something weird across blocks or something. But `MAX_SIGNER_STATE_MACHINES` is fine too, that's only like 1100, so as long as they aren't too big then it shouldn't be an issue.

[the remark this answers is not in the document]
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1209/commits/49af8271768dca888a97a87c252e8aad1f548942

signer/src/transaction_signer.rs
```diff
@@ -626,22 +603,21 @@ where
...
-                        .get_signer_set_and_aggregate_key(bitcoin_chain_tip)
-                        .await?;
-
-                    let state_machine = SignerStateMachine::load(
-                        &db,
-                        maybe_aggregate_key.ok_or(Error::NoDkgShares)?,
-                        self.threshold,
-                        self.signer_private_key,
-                    )
+                let id = validation_sighash?.into();
```
matteojug, line 606: heads-up/FYI/maybe this will bite us in the future: previously with the `contains_key` we would call `get_signer_set_and_aggregate_key` only once per bitcoin tx, while now we will call it once per input, and since that one calls `get_last_key_rotation` we may increase the signing round time by a couple of seconds (or more) until we address that.

signer/src/transaction_signer.rs
```diff
@@ -626,22 +603,21 @@ where
...
+                let state_machine = SignerStateMachine::load(
+                    &db,
+                    maybe_aggregate_key.ok_or(Error::NoDkgShares)?,
+                    self.threshold,
+                    self.signer_private_key,
+                )
+                .await?;
+
+                self.wsts_state_machines.put(id, state_machine);
+                self.relay_message(id, msg.txid, &msg.inner, bitcoin_chain_tip)
```
matteojug, line 620: IIRC with certain failures wsts would try to sign again (something about sign iteration N or something). If that's the case, will replacing the state machine cause any troubles?

signer/src/transaction_signer.rs
```diff
@@ -652,21 +628,28 @@ where
...
-                self.relay_message(msg.txid, &msg.inner, bitcoin_chain_tip)
-                    .await?;
+                let sighash = Self::validate_bitcoin_sign_request(&db, &request.message).await?;
+                let id = sighash.into();
+                let response = self
+                    .relay_message(id, msg.txid, &msg.inner, bitcoin_chain_tip)
+                    .await;
+
+                self.wsts_state_machines.pop(&id);
+                response?;
```
matteojug, line 638: Are we sure we can safely pop here? eg, if `relay_message` failed something, will the wsts coordinator try again? If so, will it do it from scratch or we need to have the state machine still ready?

djordon: I didn't think that the WSTS coordinator retried anything, but let me check.

signer/src/transaction_signer.rs
djordon: In https://github.com/stacks-network/sbtc/pull/1210 I replace that look-up using `get_signer_set_and_aggregate_key` with a look-up using the `bitcoin_tx_sighashes.sighash` column. That should be much faster, since the `sighash` column is the primary key for that table.

signer/src/transaction_signer.rs
djordon: Yeah, I cannot see how our use of the coordinator actually retries things automatically. Please correct me if I'm wrong. cc @xoloki

djordon: > IIRC with certain failures wsts would try to sign again (something about sign iteration N or something). If that's the case, will replacing the state machine cause any troubles?

So I think replacing things will be fine, assuming that the coordinator sends another nonce request message. I'm more worried about the case where they send another signature share request. I'm also more worried about keeping these state machines around. I think that the expectation should be that one nonce request leads to at most one signature share response. If we error here then the coordinator has to try again from scratch. Then we aggressively clean up signer state machines so that the private nonces can get cleared from memory and cannot be recovered.

signer/src/transaction_signer.rs
djordon: hmmm, I think my above message is better tailored to the question here. Basically, I think that we can safely pop here. I don't think that our use of WSTS tries again but if it did we should require new nonces (I think that's the safer thing to do). So I'm okay with deleting the state machine and requiring the creation of a new one with fresh nonces.

signer/src/transaction_signer.rs
djordon: Yeah I recall seeing it as well. I just took another look and here is what I think about the failure modes with overwriting the state machines. There are two cases to consider where there is an existing state machine for some sighash:
1. The coordinator tries again and starts with a new nonce request.
2. The coordinator tries again asking for signing shares with the old nonce.

## Coordinator tries again with nonce request

Having a fresh signer state machine is fine, except if a coordinator sends `NonceRequest1`, `SignatureShareRequest1`, `NonceRequest2`, `SignatureShareRequest2` and the signer gets `NonceRequest1`, `NonceRequest2`, `SignatureShareRequest1`, `SignatureShareRequest2`. I think that's a failure mode that we'll need to tolerate. Plus it seems like it's unlikely to occur.

## Coordinator tries again with a signing share request

In this case the coordinator sends `NonceRequest1`, `SignatureShareRequest1`, `SignatureShareRequest2`. Well, we can ignore this case because we don't explicitly overwrite the state machine here.

(I had written this up and then realized that it isn't applicable to the question but I'm including it here just cause). The coordinator sending this strikes me as bad behavior that we should not support. And I think the coordinator might be out of luck under, since the old private nonces get overwritten when the state machine produces a successful signature share. If a signer does not produce a successful signature share (so the old private nonces are still around), then that is because the original message sent by the coordinator did not pass some simple validation in the signer state machine.

signer/src/transaction_signer.rs
```diff
@@ -723,17 +736,17 @@ where
             .into();
 
         match db.will_sign_bitcoin_tx_sighash(&sighash).await? {
-            Some(true) => Ok(()),
+            Some(true) => Ok(sighash),
             Some(false) => Err(Error::InvalidSigHash(sighash)),
```
cylewitruk, line 740: Is this the only reason for `will_sign = false`?

signer/tests/integration/transaction_signer.rs
```diff
@@ -331,3 +345,232 @@ pub async fn assert_should_be_able_to_handle_sbtc_requests() {
...
+    // Now let's see what happens when we receive a nonce request message
+    // for a sighash that we do not know about. Since the nonce request is
+    // not in the database we should return an error, and the state machine
+    // should not be in the local cache.
+    let random_message: [u8; 32] = Faker.fake_with_rng(&mut rng);
+    match &mut nonce_request_msg.inner {
+        wsts::net::Message::NonceRequest(NonceRequest { message, .. }) => {
+            *message = random_message.to_vec();
+        }
+        _ => panic!("You forgot to update the variant"),
```
cylewitruk, line 457: lolwut?

signer/src/transaction_signer.rs
djordon: Yeah. Maybe we should change the variant name, but the message for this error is "the given sighash is known and failed validation".

signer/tests/integration/transaction_signer.rs
djordon: lol I just didn't know what to say here. I was thinking that if you see this panic, then you've updated the message variant and forgot to update this code here to catch the proper variant, or something like that.

cylewitruk (approved):
Lgtm!

Issue #1199, opened 2025-01-09
[Feature]: Coordinator attempts second DKG after a block height


djordon:
# Feature - Coordinator attempts second DKG after a block height

## 1. Description

The second DKG run will be gated by our bitcoin clock. That is, we will not run DKG again until after the arrival of a given bitcoin block with a specific height. This height will be a required configurable value set in the signers config, since we need to support this in multiple environments. After the arrival of the block at the configured height, the coordinator should attempt to run DKG.

With the current codebase the second DKG run will fail, but that is fine. We will fix that as part of a separate issue.

### 1.1 Context & Purpose

We want to support running key rotation. Full key rotation is quite complicated, so we are breaking it out into small and simple phases. We've already done phase 0, which is support running DKG once. The second phase is to run DKG a second time. This second DKG will be gated by bitcoin time.

## 2. Technical Details:

We need to add a config variable and modify the coordinator to try to run DKG if
1. DKG has never been run, or
2. DKG has been run exactly once and the current block height is greater the configured bitcoin block height.

### 2.1 Acceptance Criteria:
   - [ ] The coordinator will attempt to run DKG a second time if the current block height is greater than the configured block height.
   - [ ] The coordinator wall still run DKG once if it has never been run before.

PR #1212, opened 2025-01-14
feat: #1199 feature coordinator attempts subsequent DKG run according to configuration

cylewitruk:
## Description

Closes: #1199

## Changes

Introduces similar logic to #1200 for the transaction coordinator when determining if DKG should be run or not. If the current state of the signer and the configuration allow, the coordinator will attempt a new round of DKG.

## Testing Information

Added a new test `run_subsequent_dkg` which is very similar to the existing `run_dkg_from_scratch` test, but with its starting point that one DKG round has been completed successfully.

## Checklist:

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -1938,7 +1938,7 @@ mod tests {
             .with_mocked_clients()
             .modify_settings(|s| {
                 s.signer.dkg_min_bitcoin_block_height =
-                    dkg_min_bitcoin_block_height.map(NonZeroU64::new).flatten();
+                    dkg_min_bitcoin_block_height.and_then(NonZeroU64::new);
                 s.signer.dkg_target_rounds = NonZeroU32::new(dkg_target_rounds).unwrap();
             })
             .build();
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -303,25 +303,24 @@ where
             return Ok(());
         }
 
-        tracing::debug!("we are the coordinator, we may need to coordinate DKG");
+        tracing::debug!("we are the coordinator");
         metrics::counter!(Metrics::CoordinatorTenuresTotal).increment(1);
-        // If Self::get_signer_set_and_aggregate_key did not return an
-        // aggregate key, then we know that we have not run DKG yet. Since
-        // we are the coordinator, we should coordinate DKG.
-        let aggregate_key = match maybe_aggregate_key {
-            Some(key) => key,
-            // This function returns the new DKG aggregate key.
-            None => {
-                let dkg_result = self.coordinate_dkg(&bitcoin_chain_tip).await?;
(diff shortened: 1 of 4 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -1859,4 +1918,61 @@ mod tests {
...
+        dkg_target_rounds: u32,
+        chain_tip_height: u64,
+        should_allow: bool,
+    ) {
+        let context = TestContext::builder()
+            .with_in_memory_storage()
+            .with_mocked_clients()
+            .modify_settings(|s| {
+                s.signer.dkg_min_bitcoin_block_height =
+                    dkg_min_bitcoin_block_height.map(NonZeroU64::new).flatten();
```
djordon, line 1941: nit: I think this can be simplified a little:
```rust
dkg_min_bitcoin_block_height.and_then(NonZeroU64::new);
```

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -1115,6 +1117,242 @@ async fn run_dkg_from_scratch() {
...
+        RotateKeysV1::CONTRACT_NAME
+    );
+    assert_eq!(
+        contract_call.function_name.to_string(),
+        RotateKeysV1::FUNCTION_NAME
+    );
+    let rotate_keys = RotateKeysV1::new(
+        &signer_wallet,
+        signers.first().unwrap().0.config().signer.deployer,
+        all_aggregate_keys.iter().next().unwrap(),
```
djordon, line 1347: I think this may be fragile, right? This has both aggregate keys, and there shouldn't be any guarantee that the first one, when sorted, is the most recently generated one. Better to just check that it is the latest aggregate key.

djordon (commented):
This basically looks good. I think we need to change the integration test ever so slightly.

signer/src/transaction_coordinator.rs
cylewitruk: [680d822](https://github.com/stacks-network/sbtc/pull/1212/commits/680d82203c7c447d51ab8f84aa8d0b1685c533b5)

signer/tests/integration/transaction_coordinator.rs
cylewitruk: Well, it wasn't, but I did get it to fail by setting the agg key explicitly to `aggregate_key_1` -- not sure why it doesn't fail like this 🤔 Anyway, changed it to explicitly use the new key in [680d822](https://github.com/stacks-network/sbtc/pull/1212/commits/680d82203c7c447d51ab8f84aa8d0b1685c533b5)

djordon: I think that it was the random number generator that made it less fragile than it looks. The aggregate key that is constructed using the pseudo-randomness from `rand::rngs::StdRng::seed_from_u64(51)` must have been quite large. Switching to the `OsRng` gives failures at the expected frequency.

djordon (approved):
Looks good ✅

Issue #1214, opened 2025-01-14
[Bug]: Do not create a `BitcoinPreSignRequest` with no requests


djordon:
# Bug - Do not create a `BitcoinPreSignRequest` with no requests

## 1. Description

Signers are not supposed to create an empty presign requests but there is a bug where we allow this. Such requests are deemed as invalid by the other signers but we shouldn't even try, it wastes network bandwidth, adds logging noise, and, most importantly, is actually invalid.

### 1.1 Context & Purpose

The coordinator should not try to do invalid things. This is an invalid thing, so let's not do it.

## 2. Technical Details:

Just return early in `TxCoordinatorEventLoop::construct_and_send_bitcoin_presign_request`.

### 2.1 Acceptance Criteria:
   - [ ] The coordinator does not attempt to create or send `BitcoinPreSignRequest` messages with empty requests.

## 3. Related Issues and Pull Requests (optional):

Relates to https://github.com/stacks-network/sbtc/issues/1189

PR #1215, opened 2025-01-14
fix: exit early if no requests during presign request creation

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1214

## Changes

* Return early in `TxCoordinatorEventLoop::construct_and_send_bitcoin_presign_request` if there are no transactions to handle.

## Testing Information

I did not add a test because it would take quite a while to setup. Perhaps I should create a ticket to add a test at some later date.

## Checklist:


djordon:
I created https://github.com/stacks-network/sbtc/issues/1217 for adding a test for this change.

cylewitruk (approved):
lgtm 👍

PR #1221, opened 2025-01-15
refactor: pass around block refs

djordon:
## Changes

* Change the signature of a few functions in the `transaction_signer` module to expect a `BitcoinBlockRef`.
* Remove unnecessary db look-ups since we pass the block ref around now.

## Testing Information

This is a refactor, so if tests pass then we should be good.

## Checklist:


cylewitruk (commented):
You know, I was actually thinking about this when doing the `info` endpoint since I wanted something that returned both the canonical hash/height, but opted to keep it simple and do an extra call.

What I was thinking would that it would be best if methods took an `AsRef<BitcoinBlockHash>` that need a block hash and we implemented that for `BitcoinBlockRef`.

Thoughts?

signer/src/storage/postgres.rs
```diff
@@ -799,6 +799,22 @@ impl super::DbRead for PgStore {
...
+        &self,
+    ) -> Result<Option<model::BitcoinBlockRef>, Error> {
+        sqlx::query_as::<_, model::BitcoinBlockRef>(
+            "SELECT
+                block_hash
+              , block_height
+             FROM sbtc_signer.bitcoin_blocks
+             ORDER BY block_height DESC, block_hash DESC
+             LIMIT 1",
+        )
```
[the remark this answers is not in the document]
djordon: > Don't think it preventing merging, but why we prefer manually writing SQL instead of using some query-builder crate? As pros using query builder I see that 1) Code looks more rusty. 2) We need think about SQL injections.

Well, my personal preference is to see the SQL, even for simple queries. The folks who started on this code base all felt the same or were indifferent. Some ORM libraries, like [diesel](https://diesel.rs/), support compile time checks, which I find valuable. The original code here did compile time checks with plain SQL but that turned out to be more of a pain so I removed it.

As for SQL injections, I think we'd need to worry about that with either approach.

signer/src/transaction_signer.rs
```diff
@@ -240,26 +240,21 @@ where
...
 
             (message::Payload::WstsMessage(wsts_msg), _, _) => {
-                self.handle_wsts_message(
-                    wsts_msg,
-                    &msg.bitcoin_chain_tip,
-                    msg.signer_public_key,
-                    &chain_tip_report,
-                )
-                .await?;
+                self.handle_wsts_message(wsts_msg, msg.signer_public_key, &chain_tip_report)
```
matteojug, line 250: Previously here we were using the chain tip from the msg, now we are using the local chain tip. This kinda sounds better, but I'm not sure if it's messing things up (eg, when loading the `StateMachine`).
Kinda also applies to `BitcoinPreSignRequest` below, while the `StacksTransactionSignRequest` should be fine since we check that the message bitcoin ti[ is canonical.

djordon: Yeah it makes a difference for sure. We probably need to make some other changes around there too, so I'll take a look.

But in general, I'm okay with us ignoring the old coordinator it the cases where we weren't.

signer/src/transaction_signer.rs
```diff
@@ -557,28 +540,28 @@ where
...
                     signer_id = %dkg_private_shares.signer_id,
                     "handling DkgPrivateShares"
                 );
-                let id = StateMachineId::from(bitcoin_chain_tip);
+                let id = StateMachineId::from(&chain_tip.block_hash);
                 self.validate_sender(&id, dkg_private_shares.signer_id, &msg_public_key)?;
-                self.relay_message(id, msg.txid, &msg.inner, bitcoin_chain_tip)
+                self.relay_message(id, msg.txid, &msg.inner, &chain_tip.block_hash)
                     .await?;
             }
```
matteojug, line 566: IIUC here (and in `DkgPublicShares`) we are now loading the "current" (chain tip wise) state machine, even for messages coming from a stale signer (eg, sending a message for an old chain tip).
I'm wondering if we should just ignore messages that are not from our chain tip directly before calling `handle_wsts_message`. Otherwise, we could also load the correct state machine, but at that point I'm not sure we care about it, and a rouge signer could mess with the LRU ordering (not an issue now, but it could be if we start to do signature rounds concurrently I guess).

matteojug (approved):

signer/src/transaction_signer.rs
djordon: Yeah, good point. Fixed up in https://github.com/stacks-network/sbtc/pull/1221/commits/7be5559387c22ae874c582fdfd5ff04036584275.

PR #1224, opened 2025-01-15
Update to wsts 11

### LINKED ISSUE #1162: [Bug]: Signers can crash each other with invalid `DkgPrivateShares`
# Bug - Signers can crash each other with invalid `DkgPrivateShares`

## 1. Description

This issue is that a malicious signer can cause other signers to panic by sending them empty `DkgPrivateShares` objects. This is a bug in WSTS and is tracked there by https://github.com/Trust-Machines/wsts/issues/109.

### 1.1 Context & Purpose

A malicious signer can construct an empty (or nearly empty) `DkgPrivateShares` object and broadcast it to other signers. When they receive it they will crash, and their hopes and dreams will crash with it.

## 2. Technical Details:

The fix here will be on the WSTS side and it is here https://github.com/Trust-Machines/wsts/pull/111. We still need to update the signer crate to use this version, so this ticket tracks that.

### 2.1 Acceptance Criteria:
   - [ ] A signer cannot crash other signers if they broadcast empty `DkgPrivateShares`.

https://github.com/Trust-Machines/wsts/issues/109

### LINKED ISSUE #1166: [Bug]: Missing length check when parsing `SignatureShareRequest`
# Bug - Missing length check when parsing `SignatureShareRequest`

## 1. Description

We do not check the length of the contents of `SignatureShareRequest`s. This can lead to resource exhaustion of the signer, since a malicious coordinator can jam up the system by broadcasting signature share requests with lots of junk. What's more, going through the `SignatureShareRequest` takes a lot more time than generating the request itself, so doing basic validation is crucial here.

### 1.1 Context & Purpose

We do not want to make it easy for a malicious signer to lock the protocol. Doing basic validation on the request is necessary to ensure that an attacker cannot compromise the protocol. 

## 2. Technical Details:

There is a ticket for this in WSTS at https://github.com/Trust-Machines/wsts/issues/108. The fix in this repo is to update to use it.

### 2.1 Acceptance Criteria:
   - [ ] A malicious signer cannot exhaust other signer resources by sending bad `SignatureShareRequest` messages.

### LINKED ISSUE #1167: [Bug]: Correctly handle WSTS nonce issues
# Bug - Correctly handle WSTS nonce issues

## 1. Description

Whenever a signer gets a sign request over a particular message, the signers create a state machine if its for a transaction ID that they have never seen before. This leads to the signer having fresh state with fresh nonces. The issue is that we do this over the incoming transaction ID and not the message itself, leading to the same state machine being reused. When that happens nonces can be replayed leading to various attacks against the protocol.

### 1.1 Context & Purpose

Reusing the same WSTS state machine can lead to issues down the line due to nonce reuse. This ticket addresses such issues by ensuring the same nonces can't be reused.

## 2. Technical Details:

When a signer receives a `NonceRequest` message they should create new nonces.

### 2.1 Acceptance Criteria:

- [ ]  The signers use a new nonce for each input that they are signing

xoloki:
## Description

Update the `wsts` dependency to `11.0.0`

Closes #1162, closes #1166, and closes #1167 

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8045,7 +8045,8 @@ dependencies = [
 [[package]]
 name = "wsts"
 version = "11.0.0"
-source = "git+https://github.com/Trust-Machines/wsts.git?rev=612c0230d3469e5807575d32da7f6c992e2f6177#612c0230d3469e5807575d32da7f6c992e2f6177"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "609a3172eb7ed12203fa7d2dd2a3ddff9ddb7c14b7b2b10039293362d41a3cc0"
 dependencies = [
  "aes-gcm",
  "bs58 0.5.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4558,9 +4558,9 @@ checksum = "b15813163c1d831bf4a13c3610c05c0d03b39feb07f7e09fa234dac9b15aaf39"
 
 [[package]]
 name = "p256k1"
-version = "7.1.0"
+version = "7.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a40a031a559eb38c35a14096f21c366254501a06d41c4b327d2a7515d713a5b7"
+checksum = "f8641765697df03a529cd21181fbccd1cad9f66086c24252b91c16bf38cdddd6"
 dependencies = [
  "bindgen",
  "bitvec",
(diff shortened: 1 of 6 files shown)


djordon (commented):
This looks good, but let's hold off on merging until we know whether we want to include it or not.

cylewitruk:
Is this still relevant?

djordon:
> Is this still relevant?

Yes, it is. The updates to WSTS in 0.0.9-rc6 do not include many fixes that have landed on WSTS main.

@xoloki isn't the version here also stale? It includes the listed fixes but it needs to include one more right? As in, don't we need `v12.0.0` or something like that? Also, you'll likely see CI fail after merging in main because of `cargo-vet`, so you'll need to update that here as well.

djordon:
@xoloki it's safe to close this right?

xoloki:
> @xoloki it's safe to close this right?

Yeah I'll close it now

PR #1223, opened 2025-01-15
feat: integration test improvements

cylewitruk:
## Description

Putting this up as a draft just for visibility -- not a priority, this was just some fun I had an evening.

This is a PR to try to make some QoL improvements (i.e. both cognitive and speed-wise) regarding the running of our integration tests in general, primarily with focus on being able to run integration tests in parallel.

It makes use of [`testcontainers`](https://github.com/testcontainers/testcontainers-rs) to run Bitcoin core in containers, one for each test, in Rust.

PR #1238, opened 2025-01-18
refactor: ci/test improvements

cylewitruk:
## Description

Our build/CI times were starting to annoy me, so I took a stab at refactoring this a bit. Got a bit bigger than I thought it would, but this is the result of a fair bit of trial-and-error -- at least it deletes more than it adds 😄

A lot of the changed files are just moved or deleted files, or removing the `#[cfg_attr(not(feature = "integration-tests"), ignore)]` attribute.

This PR brings CI times down from [~10 minutes](https://github.com/stacks-network/sbtc/actions/runs/12817682729) to [~4 minutes](https://github.com/stacks-network/sbtc/actions/runs/12844130479). This could probably be further reduced to closer to ~3m with #1223 (which reduces runtime of integration tests from ~ 3m to ~1m).

### Workspace
- Refactored the `Makefile` a fair bit; all of the targets should work as expected now (unless I missed to test something)
- The `Makefile` now handles the 3x emily API clients (private, public, testing).
- Removed the `integration-tests` feature:
  - This is unnecessary as unit tests can be selected by running with the `--lib` cargo/nextest flag, and integration tests with the `--test integration` argument. 
  - This lets us avoid an extra build with the feature vs. not. 
  - The integration tests and `Makefile` targets have been updated to reflect this.
- All testing uses `nextest` now.
- Consolidated some dependencies and their features into the workspace `Cargo.toml`.
  - Had to make some updates updates to `blocklist-client` and emily integration tests due to the `config` crate version/API differences.
- Bumped rust to `1.81` so that `make emily-as-lambda` works.
  - Minor refactor in libp2p `event_loop.rs` due to new `clippy` lint.
- Code generation:
  - We no longer try to generate code all the time (we have it checked in). `make build` or the respective specialized codegen targets will still do this if needed, though.
  - ...but we now have checks which will fail if the generated code is out-of-date vs. its generators.
  - Moved all generators out of `.generated-sources` and into the main workspace. Now only actually generated code lives there.
  - Made the `openapi-generator-cli` less noisy (error logs only).
  - Removed the `clean` commands for generators since we don't want to delete the generated sources anymore (checked in).

### CI
- Consolidated "on push" jobs to the `on-push.yaml` workflow. 
  - This makes it easier with job dependencies.
  - I generally find it more coherent to keep related jobs from the same trigger in the same workflow (in this case `on-push`).
  - Yes, the jobs still run in parallel, and you still see the individual jobs in PRs (see the "Checks" at the bottom of this PR).
- Bumped job runners to `ubuntu-24.04`.
- Consolidates tool versions to environment variables at the top.
- Use `nextest` [archives](https://nexte.st/docs/ci-features/archiving/) to separate the building of test artifacts from the running of them. Now we run unit and integration jobs concurrently.
- Use the [`mold`](https://github.com/rui314/mold) linker -- generally faster than `ld`.
- Optimize how integration tests are run:
  - We now work on starting the integration environment (takes about ~1m) in parallel to building the test artifacts so that the environment should be ready at just about the same time as the build artifacts (nextest archives) are ready, and once the environment is ready we wait for the artifacts (if needed). [example](https://github.com/stacks-network/sbtc/actions/runs/12843507730/job/35815739292)
  - Removed the `emily-aws-setup` container in CI and run the `initialize.py` directly. The container build took unnecessary time.
  - Use `nextest` [partitioning](https://nexte.st/docs/ci-features/partitioning/) to split-up the integration tests over parallel jobs.
- Added checks for stale generated code for blocklist and emily clients (in addition to contracts).
- All project lints are consolidated into the `lints` job.
- Added commit hashes for all used actions (#1235).
- Sets all default permissions for the workflow to `read` (#1234).
- ~~Unit & integration tests now run with `--retries 2`, so if there are any flaky tests the run will have a chance at succeeding (note that flaky tests are reported in the logs).~~
- Commented the workflow jobs.

Well I think that's the gist of it... Probably forgot something 😅 

## Checklist:


aldur (commented):
I _think_ this is OK and have given it a quick look given limited time available.

I have seen a lot of things I like -- my only wish would have been for this to be smaller and split into multiple PRs.

cylewitruk:
> I _think_ this is OK and have given it a quick look given limited time available.
> 
> I have seen a lot of things I like -- my only wish would have been for this to be smaller and split into multiple PRs. 

Yeah, it's bigger than I thought it would be, it was a lot of trial and error and things are sort-of intertwined. 

Now that I have it buttoned up and know what works I could probably split it over multiple successive PR's, but that'll take time to do, so I opted to PR it as-is and move on to things in the backlog.

.github/workflows/on-push.yaml
```diff
@@ -0,0 +1,254 @@
...
+        with:
+          cache-key: "rust-tests"
+      # - uses: pnpm/action-setup@v4
+      # - uses: actions/setup-node@v3
+      #   with:
+      #     node-version: ${{ env.NODE_VERSION }}
+      #     cache: "pnpm"
+      # - uses: arduino/setup-protoc@v3
+      #   with:
+      #     version: "25.3"
```
djordon, line 85: Why not just delete, same thing for a the commented out `Install Package Dependencies`

.github/workflows/on-push.yaml
```diff
@@ -0,0 +1,254 @@
...
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: nextest@0.9.82
+      # - name: Install Package Dependencies
+      #   run: make install
+      - name: Build Tests
+        run: make test-build
+      - name: Create Nextest Archives
+        run: make nextest-archive
```
djordon, line 94: Why is this necessary?

.github/workflows/on-push.yaml
```diff
@@ -0,0 +1,254 @@
...
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: nextest@0.9.82
+      - name: Install Package Dependencies
+        run: make install
+      - name: Start Integration Test Environment
+        run: make integration-env-up-ci
+      - name: Wait for Test Artifacts
+        uses: kachick/wait-other-jobs@v3.6.0
+        timeout-minutes: 5
```
djordon, line 175: Why is necessary? Oh because of the `--detach` the above step finishes nearly immediately? Before we would get lucky, maybe?

.github/workflows/on-push.yaml
```diff
@@ -0,0 +1,254 @@
...
+                }
+              }
+            ]
+      - name: Download Nextest Archives
+        uses: actions/download-artifact@v4
+        with:
+          name: nextest-archives
+          path: .
+      - name: Run Integration Tests (Rust)
+        run: cargo-nextest nextest --color always run --no-fail-fast --test-threads 1 --retries 2 --partition hash:${{ matrix.partition }}/2 --archive-file nextest-archive-serial.tar.zst
```
djordon, line 197: Let's remove the retries. If we construct a flaky test, let's go with handling flaky tests very quickly by either removing or fixing them when they arise.

djordon: What does `--partition hash:${{ matrix.partition }}/2` do, and why is `--archive-file nextest-archive-serial.tar.zst` necessary?

.github/workflows/on-push.yaml
djordon: Oh I see, this compiles the tests and stuff, or something like that.

djordon (commented):
Thank you, 🙏🏽. I need to give some of these changes another look, but this largely looks good. I mainly just have a bunch of questions.

docker/docker-compose.ci.yml
```diff
@@ -43,13 +43,3 @@ services:
...
-
-  emily-aws-setup:
-    build:
-      context: ..
-      dockerfile: docker/sbtc/emily-aws-setup/Dockerfile.ci
-      target: emily-aws-setup
-    depends_on:
-      - emily-dynamodb
-    environment:
-      - DYNAMODB_ENDPOINT=http://emily-dynamodb:8000
```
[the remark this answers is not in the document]
cylewitruk: Turns out we never actually needed it (in CI) 🤷🏻‍♂️

.github/workflows/on-push.yaml
```diff
@@ -0,0 +1,254 @@
...
+      - uses: actions/checkout@v4
+        with:
+          submodules: "recursive"
+      - uses: pnpm/action-setup@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: "pnpm"
+      - name: Install Nextest
+        run: curl -LsSf https://get.nexte.st/latest/linux | tar zxf - -C /usr/local/bin
```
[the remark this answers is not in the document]
cylewitruk: Will need to check if there was a reason I did it this way or if it was just a miss

.github/workflows/on-push.yaml
cylewitruk: Yeah should be removed, miss

.github/workflows/on-push.yaml
cylewitruk: Yup -- we do one single build of all of the test artifacts and then package them into two archives: one for tests that can be run in parallel (unit tests right now), and one for tests that must be run serially.

The running of the tests then no longer need rust, only the nextest binary. So we use this to ship the artifacts to the jobs which actually run the tests without needing to build them again.

.github/workflows/on-push.yaml
cylewitruk: Well before we ran integration tests serially after unit tests.

You'll notice that "Run Unit Tests" has a direct dependency on "Build Test Aetifacts". That works well because the latter doesn't really take any startup time.

For integration tests, we need to start the integration env, i.e. pull docker images, start containers, and build `emily-server`, which takes ~1m.

So, instead of having that 1m come _after_ the tests are built, we immediately start setting up the environment in parallel to the test artifacts being built, which also takes ~1m.

So this action waits for the artifacts to be ready (the nextest archives) if the environment would become ready before the artifacts are available for download.

Hope that makes sense?

.github/workflows/on-push.yaml
cylewitruk: This splits the runs into 2 separate job runs based on the job matrix `[1,2]` -- i.e. we get two deterministic buckets of integration tests, splitting them roughly 50/50, so instead of one job taking ~3m, we get two jobs taking ~1.5m running in parallel.

More information in the nextest archive docs linked in PR description.

signer/tests/integration/bitcoin_client.rs
```diff
@@ -26,8 +25,7 @@ async fn test_get_block_not_found() {
 
 // TODO: Figure out how to let this (and similar tests) run against the wallet
 // generated by `initialize_blockchain()`. See comment in the test below.
-//#[ignore = "This test needs to be run against a 'fresh' bitcoin core instance"]
-#[cfg_attr(not(feature = "integration-tests"), ignore)]
+#[ignore = "This test needs to be run against a 'fresh' bitcoin core instance"]
```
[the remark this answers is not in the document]
cylewitruk: Hm, I wonder if I messed this one up if it was commented out first. Will try uncommenting it

signer/Cargo.toml
```diff
@@ -15,19 +14,19 @@ testing = ["fake", "mockall", "sbtc/testing"]
...
 emily-client.workspace = true
 clap.workspace = true
 clarity.workspace = true
-config = "0.14"
+config.workspace = true
 futures.workspace = true
 hashbrown.workspace = true
 libp2p.workspace = true
-lru = { workspace = true, default-features = false }
+lru.workspace = true
```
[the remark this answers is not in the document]
cylewitruk: Yeah, I probably did this before I realized what we discussed

signer/src/bitcoin/fees.rs
```diff
@@ -217,7 +217,6 @@ mod tests {
     }
 
     #[tokio::test]
-    #[cfg_attr(not(feature = "integration-tests"), ignore)]
     async fn prod_estimate_fee_rate_works() {
         let client = reqwest::Client::new();
```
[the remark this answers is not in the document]
cylewitruk: It actually passes because it uses `mempool.space` and `bitcoiner.live` -- but I agree that this should be an integration test and not a unit test.

No, I don't think this would be run currently in CI as either a unit or integration test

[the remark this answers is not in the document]
cylewitruk: Actually, I think this should actually be both moved and ignored since I don't think we actually use this atm.

.github/workflows/on-push.yaml
djordon: Yup, that makes sense. Before this change, we would rely on the fact that it takes time to run unit tests.

signer/src/bitcoin/fees.rs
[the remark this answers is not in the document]
djordon: Yeah that makes sense to me. Let's do that.

.github/workflows/on-push.yaml
```diff
@@ -0,0 +1,254 @@
...
+
+  # Runs various lints and checks for the project, including Rustfmt, Clippy,
+  # Protobuf lints, and pnpm lints.
+  lint:
+    name: Run Lints
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: rui314/setup-mold@v1
+        with:
+          make-default: true
```
djordon, line 26: Why the mold linker? Is it just faster?

cylewitruk: Yup, generally speaking a lot faster than ld

.github/workflows/on-push.yaml
cylewitruk: > Before this change, we would rely on the fact that it takes time to run unit tests before running our integration tests.

Well, before this change we would actually do the following, serially:

1. Run `clippy` (build 1, no `--tests` flag)
2. Run unit tests (build 2, `--tests` flag)
3. Setup test environment (build 3, `emily-server` with `--bin` flag + docker pulls)
4. Run integration tests (build 4, `--tests` flag with `integration-tests` feature)

So this is all about doing as much of the above in parallel as possible, and avoiding re-compiling things where possible.

.github/workflows/on-push.yaml
cylewitruk: [7d93766](https://github.com/stacks-network/sbtc/pull/1238/commits/7d93766c0ebfc80a7d841f846629944a75d0e0ec)

signer/Cargo.toml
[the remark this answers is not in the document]
cylewitruk: [53ae218](https://github.com/stacks-network/sbtc/pull/1238/commits/53ae2182f57bfbca9e2faebe5871438a1621454d)

signer/tests/integration/bitcoin_client.rs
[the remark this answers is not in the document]
cylewitruk: Yeah it shouldn't be ignored -- removed the ignore completely in [bdbac6a](https://github.com/stacks-network/sbtc/pull/1238/commits/bdbac6a81698e2f29b11c15075527a7906d3fda1)

signer/src/bitcoin/fees.rs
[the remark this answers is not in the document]
cylewitruk: Moved and ignored in [08dc1b8](https://github.com/stacks-network/sbtc/pull/1238/commits/08dc1b8e181d2f38e8aee93319a369bbab574ea5)

.github/workflows/on-push.yaml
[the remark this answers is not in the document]
cylewitruk: Looks like I did it for no reason, the action works 👍  [b4856c8](https://github.com/stacks-network/sbtc/pull/1238/commits/b4856c8af65953ab0df5547c7146f6fcef1c6786)

cylewitruk:
Added `permissions: read-all` to the `on-push` workflow in lieu of #1247 being merged

cylewitruk:
Added commit hashes for all actions in lieu of #1249 being PR'd

.github/workflows/on-push.yaml
cylewitruk: Retries removed in [6500f48](https://github.com/stacks-network/sbtc/pull/1238/commits/6500f487ecd5a526a213469e8de4db58c3d1bf81)

Makefile
```diff
@@ -86,51 +90,82 @@ integration-test-full: integration-env-down integration-env-up integration-test
...
-	AWS_SECRET_ACCESS_KEY=xxxxxxxxxxxx \
-	AWS_REGION=us-west-2 \
-	TRUSTED_REORG_API_KEY=testApiKey \
-	cargo run --bin emily-server -- \
-		--host 127.0.0.1 --port 3031 --dynamodb-endpoint http://localhost:8000 > ./target/emily-server.log 2>&1 &
+		AWS_SECRET_ACCESS_KEY=xxxxxxxxxxxx \
+		AWS_REGION=us-west-2 \
+		TRUSTED_REORG_API_KEY=testApiKey \
+		cargo run --bin emily-server -- \
+			--host 127.0.0.1 --port 3031 --dynamodb-endpoint http://localhost:8000 > ./target/emily-server.log 2>&1 &
```
matteojug, line 105: not strictly related to this but still: why can't we use `integration-env-up` directly even in CI?

Makefile
```diff
@@ -1,80 +1,84 @@
...
+	docker compose --file docker/docker-compose.test.yml up -d
+
+integration-test:
+	cargo nextest run $(CARGO_FLAGS) --test integration --no-fail-fast --test-threads 1
+
+integration-test-build:
+	cargo test build $(CARGO_FLAGS) --test integration --no-run --locked
 
 integration-env-down:
 	docker compose --file docker/docker-compose.test.yml down -v
```
matteojug, line 84: nit: can we add a `-t 0` here? I don't think we care about graceful shutdown for this

Makefile
```diff
@@ -86,51 +90,82 @@ integration-test-full: integration-env-down integration-env-up integration-test
...
+# ##############################################################################
 
-# Project Names
-## Cargo crates
-EMILY_HANDLER_PROJECT_NAME := emily-handler
-EMILY_OPENAPI_SPEC_PROJECT_NAME := emily-openapi-spec
-## Node projects
-EMILY_CDK_PROJECT_NAME := emily-cdk
+devenv-no-sbtc-up:
+	docker compose -f docker/docker-compose.yml --profile default --profile bitcoin-mempool up
```
matteojug, line 119: Can we add `-d` here (and maybe also in `devenv-sbtc-up`)? To align with `devenv-up` (and avoid having logs spammed to the console).

Even better if we could supply some extra flags when invoking the targets in this section (eg, being able to `make devenv-down -- -v -t 0`, not sure if it's something we can do with make)

.github/workflows/on-push.yaml
```diff
@@ -0,0 +1,247 @@
...
+        run: make contracts
+      - name: Ensure Git is Clean (Contracts)
+        id: git-status-contracts
+        run: git diff --no-ext-diff --exit-code
+        continue-on-error: true
+      - name: Git is Dirty (Contracts)
+        if: steps.git-status-contracts.outcome == 'failure'
+        run: |
+          echo "::error title=Contracts are dirty:: Make sure you ran 'make contracts' before pushing."
+          exit 1
```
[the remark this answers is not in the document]
cylewitruk: I did it this way so that it's very clear in CI what succeeds/fails (on the step-level) -- otherwise I would have done it in a single make step

.github/workflows/on-push.yaml
```diff
@@ -0,0 +1,247 @@
+name: On Push
+on: [push]
+
+permissions: read-all
```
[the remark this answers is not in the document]
cylewitruk: We don't, I could mess around with this and find the exact permissions, but since it's a public repo I don't really see any risks with giving it read to everything and it keeps things simple; write is the scary one (actually surprised I didn't need to give it write for something for uploading artifacts 🤔 )

Makefile
cylewitruk: i think we could, but it would be slower

Makefile
cylewitruk: we can[816e2b3](https://github.com/stacks-network/sbtc/pull/1238/commits/816e2b37e0cf5209a25cd5c50313c911d058dc33)
[f8327b8](https://github.com/stacks-network/sbtc/pull/1238/commits/f8327b8d23c94156aa0c8c73050c12f7c7a6536e)
[cb4947b](https://github.com/stacks-network/sbtc/pull/1238/commits/cb4947bb4397208f4f93ef9fd26d1ef4a1b396af)

Went ahead and added it to all of the `down`s with the exception of `devenv-sbtc-down` (since i use this and want it to shutdown gracefully because i generally start it up again).

Makefile
cylewitruk: [816e2b3](https://github.com/stacks-network/sbtc/pull/1238/commits/816e2b37e0cf5209a25cd5c50313c911d058dc33)
[f8327b8](https://github.com/stacks-network/sbtc/pull/1238/commits/f8327b8d23c94156aa0c8c73050c12f7c7a6536e)
[cb4947b](https://github.com/stacks-network/sbtc/pull/1238/commits/cb4947bb4397208f4f93ef9fd26d1ef4a1b396af)

Yeah `make` doesn't support arguments like that... have to pass it like `make xxx ARGS="-t 0 -v"` etc.

.github/workflows/on-push.yaml
```diff
@@ -0,0 +1,247 @@
...
+  lint:
+    name: Run Lints
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: rui314/setup-mold@f80524ca6eeaa76759b57fb78ddce5d87a20c720 #v1
+        with:
+          make-default: true
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
+        with:
+          submodules: "recursive"
```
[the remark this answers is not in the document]
matteojug: mmm I would prefer to add it when it would become required, also since it's unclear if we would want to clone recursively or not at this point.
If you want to keep it, the checkout in `check-generated-code` is missing this setting.

Makefile
```diff
@@ -1,80 +1,84 @@
...
-AUTOGENERATED_EMILY_CLIENT := $(AUTOGENERATED_SOURCES)/emily/client/rust/src/lib.rs
-EMILY_LAMBDA_BINARY := target/lambda/emily-handler/bootstrap.zip
-EMILY_CDK_TEMPLATE := emily/cdk/cdk.out/EmilyStack.template.json
+AUTOGENERATED_SOURCES := .generated-sources
 
 # Don't use the install target here so you can rerun install without
 # Makefile complaints.
 export DATABASE_URL=postgres://user:password@localhost:5432/signer
 
+# The package flags for cargo commands. If we do not specify
```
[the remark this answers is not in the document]
cylewitruk: Whoops, don't even remember what I was going to write 😅 

[3637f08](https://github.com/stacks-network/sbtc/pull/1238/commits/3637f08de55649f46397adc0ae4774de271f6c33)

Makefile
matteojug: Makes sense, it's just a bit annoying that we run something that is slightly different that the setup we use to run those locally.

.github/workflows/on-push.yaml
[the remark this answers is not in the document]
cylewitruk: Removed it in [c966944](https://github.com/stacks-network/sbtc/pull/1238/commits/c966944f4318732bd82364b007e3ae02c4854c82)

blocklist-openapi-gen/package.json
djordon, line 1: The old file for this generated crate, the one in `.generated-sources/blocklist-openapi-gen/package.json`, has two more commands, both `clean` related. That clean command found all files of a certain depth and deleted them. Removing might be for the best. Just calling it out.

emily/openapi-gen/package.json
djordon, line 1: This one had a `clean-rust` command, which I think means that it was never really used in our `Makefile`.

Makefile
```diff
@@ -1,83 +1,87 @@
...
+
+# Creates nextest archives
+nextest-archive:
+	cargo nextest archive $(CARGO_FLAGS) --lib --archive-file $(NEXTEST_ARCHIVE_FILE) ${CARGO_BUILD_ARGS}
+	cargo nextest archive $(CARGO_FLAGS) --archive-file $(NEXTEST_SERIAL_ARCHIVE_FILE) --test integration ${CARGO_BUILD_ARGS}
+
+# Runs nextest archives
+nextest-archive-run:
+	cargo nextest run --no-fail-fast --retries 2 --archive-file $(NEXTEST_ARCHIVE_FILE)
+	cargo nextest run --no-fail-fast --test-threads 1 --retries 2 --archive-file $(NEXTEST_SERIAL_ARCHIVE_FILE)
```
djordon, line 63: Don't the retries need to be removed here as well? Or maybe not since these don't get used in CI.

Makefile
```diff
@@ -1,83 +1,87 @@
...
 	pnpm install
 
-build: $(AUTOGENERATED_BLOCKLIST_CLIENT_CLIENT) $(AUTOGENERATED_EMILY_CLIENT)
-	cargo build ${CARGO_BUILD_ARGS}
-	pnpm --recursive build
-	# TODO(719): No need to do this once rustfmt 2.0.0 ships
-	# Format generated sources:
-	cargo fmt -p emily-openapi-spec -p emily-client -p blocklist-api -p blocklist-openapi-gen
+build: blocklist-client-codegen emily-client-codegen contracts
+	cargo build --all-targets $(CARGO_FLAGS) ${CARGO_BUILD_ARGS}
```
djordon, line 22: Do we want all targets here? From the [docs](https://doc.rust-lang.org/cargo/commands/cargo-build.html#target-selection), this is equivalent to `--lib --bins --tests --benches --examples`, which seems like more than we'd usually want. Maybe just the default `cargo build` would be better and you can supply the build arg with `CARGO_BUILD_ARGS=--all-targets make build`, if you want all targets. Thoughts?

djordon (approved):
Looks good. But I had only two things, one about building all targets and another about the retries, which I don't think get used in CI.

I did not take a close look at the Makefile changes for the emily client or the blocklist client so some extra eyes there would be nice.

Makefile
cylewitruk: Yeah -- I had actually toyed with the idea of doing the reverse, i.e. using the ci version locally; I might still try to do this, but in another PR..

blocklist-openapi-gen/package.json
cylewitruk: Yeah, this is noted in the PR desc -- now that the code is comitted, it doesn't really make sense to have a command that deletes it all

Makefile
cylewitruk: Yeah, locally it should be fine to have retries, then it's highlighted in a separate yellow "FLAKY" section in the nextest report, which imo is kinda nice to know if the test flat-out fails or if it is indeed flaky.

blocklist-openapi-gen/package.json
djordon: Oh right 😅.

emily/openapi-gen/package.json
djordon: I don't think that command should be added back, just noting it.

Makefile
cylewitruk: Tbh i was on the fence about removing this altogether because it's so ambiguous and we don't use it in any of our actual build processes.

`cargo build` will default build dev profile without the `--tests` flag, which is less than useful on a dev machine and not optimal for a release build. `--all-targets` gives us `--bin --lib --tests` which feels more relevant (we don't have examples or benches so those are moot).

Now we have separate `make contracts` etc. so `make build` isn't needed for codegen, and our other rules will spec/codegen if (and only if) the sources are outdated now.

So, I honestly don't know what's best 🤷🏻‍♂️ I never actually run this personally (except before when contract stuff was outdated).

djordon: I don't use it either (at least not anymore). It is something that I've told people to do to get up and running for development, since it was used to build generated sources and smart contract stuff. But the part that seems unnecessary is `--bin`, which as time goes on will probably be even more unnecessary.

I don't actually feel strongly about it. I use `cargo build` and `pnpm build` when I need to.

I'll resolve this and leave the final decision to you.

cylewitruk: I'll leave it for now and we can change it in another PR if we decide we want to, I don't have any strong opinions either

Issue #1237, opened 2025-01-17
[Bug]: Missing Signer Votes


Jiloc:
# Bug - Missing Signer Votes

### 1.1 Context & Purpose
In the current devnet setup, we observed issues where some signer votes are discarded because the signers receive votes before their Bitcoin node has updated with the latest block containing the deposit. This causes signers that have not yet received the request to fetch it from Emily but still "skip" saving it because their Bitcoin node is unable to validate the deposit, ultimately resulting in the vote being discarded. Once the new block is received (~15 seconds later), they process subsequent votes correctly, but the initial votes are lost. This causes unnecessary delays and inefficiencies in the signing process.

The issue appears related to timing and synchronization of Bitcoin block updates across signers. While all votes are broadcasted and received correctly, some signers lagging behind on Bitcoin block updates fail to validate deposits and discard votes prematurely.

## 2. Technical Details:

- The issue occurs when a signer receives a vote before its Bitcoin node has processed the latest block containing the deposit.
- Discarded votes lead to delays until the new block is processed, causing signers to re-process and validate subsequent votes.
- Current behavior relies on a 30-second sleep (`bitcoin_processing_delay`) after the request decider sends out votes, but this does not address the root synchronization issue.

### 2.1 Acceptance Criteria:
   - [ ] Introduce a mechanism to re-propagate votes for a few subsequent blocks (e.g., N=3 or 5) to allow lagging signers to catch up with Bitcoin block updates.
   - [ ] Ensure signers do not prematurely discard votes when their Bitcoin node is lagging?
   - [ ] Optimize vote propagation to avoid excessive message spam while maintaining reliability (e.g., consolidate votes into a single message after processing all requests?).

PR #1243, opened 2025-01-20
fix: missing signers' votes

Jiloc:
## Description

Closes: #1237

## Changes
- add new `deposit_decisions_retry_window` param to the signers' settings (defaults to 3)
- add query to get the all the `DepositSigner` entries from the DB from the last N blocks
- add resend mechanism in the request_decider to resend all the Requests of the last 3 blocks when a new bitcoin block arrives
 
## Testing Information
- add query integration tests

## Checklist:


signer/src/storage/postgres.rs
```diff
@@ -1921,6 +1921,54 @@ impl super::DbRead for PgStore {
...
+            WHERE deposit_signers.signer_pub_key = $3
+            "#,
+        )
+        .bind(chain_tip)
+        .bind(i32::from(context_window))
+        .bind(signer_public_key)
+        .fetch_all(&self.0)
+        .await
+        .map_err(Error::SqlxQuery)
+    }
```
Jiloc, line 1971: A simple optimization could be to specialize a bit the query and only get decision of the deposits which have less than the total number of signers entries in the table.

signer/src/request_decider.rs
```diff
@@ -248,6 +276,20 @@ where
...
 
+    /// Send the given deposit decisions to the other signers for redundancy.
+    #[tracing::instrument(skip_all)]
+    pub async fn handle_deposit_decisions_to_retry(
+        &mut self,
+        decisions: Vec<model::DepositSigner>,
+        chain_tip: &BitcoinBlockHash,
+    ) -> Result<(), Error> {
+        for decision in decisions.into_iter().map(SignerDepositDecision::from) {
+            self.send_message(decision, chain_tip).await?;
```
matteojug, line 281: Since this retry is a "best effort" type of thing, should we just log the error in here instead of bailing at the first one?

signer/src/storage/postgres.rs
matteojug: This is doing something slightly different from what we initially talked about: here we pick the deposits that were confirmed (on the canonical chain) in the last three blocks, but this means that if we fetch from Emily a deposit that was already confirmed but submitted (to emily) only recently we may not retry it.
This is fine wrt the original issue (signers getting a vote for an viewed-as-unconfirmed deposit), but if we were to have a real _retry 3 times_ approach we would also add redundancy to lost votes messages (for other, unrelated reasons) -- but it would require adding more info somewhere else in the db, so maybe it's not really worth it.

Also a possible simplification here could be to ignore the canonical chain, but given the context window is so small the recursive part should be an issue.

signer/src/storage/postgres.rs
```diff
@@ -1921,6 +1921,54 @@ impl super::DbRead for PgStore {
...
+
+                UNION ALL
+
+                -- Recursive member: Fetch the parent block using the last block's parent_hash
+                SELECT parent.block_hash, parent.block_height, parent.parent_hash,
+                       parent.created_at, last.depth + 1
+                FROM sbtc_signer.bitcoin_blocks parent
+                JOIN context_window last ON parent.block_hash = last.parent_hash
+                WHERE last.depth < $2
+            ),
```
matteojug, line 1947: Can we use `bitcoin_blockchain_of` here?

signer/src/request_decider.rs
```diff
@@ -248,6 +276,20 @@ where
...
     }
 
+    /// Send the given deposit decisions to the other signers for redundancy.
+    #[tracing::instrument(skip_all)]
+    pub async fn handle_deposit_decisions_to_retry(
+        &mut self,
+        decisions: Vec<model::DepositSigner>,
+        chain_tip: &BitcoinBlockHash,
+    ) -> Result<(), Error> {
+        for decision in decisions.into_iter().map(SignerDepositDecision::from) {
```
matteojug, line 280: Should we do this concurrently? Instead of awaiting each one.

signer/tests/integration/postgres.rs
```diff
@@ -1400,6 +1400,71 @@ async fn fetching_deposit_request_votes() {
...
+    let signer_set = testing::wsts::generate_signer_set_public_keys(&mut rng, num_signers);
+
+    let test_data = TestData::generate(&mut rng, &signer_set, &test_model_params);
+    test_data.write_to(&pg_store).await;
+
+    let chain_tip = test_data
+        .bitcoin_blocks
+        .iter()
+        .max_by_key(|x| (x.block_height, x.block_hash))
+        .unwrap();
```
matteojug, line 1430: nit: we can use `pg_store.get_bitcoin_canonical_chain_tip`

signer/tests/integration/postgres.rs
```diff
@@ -1400,6 +1400,71 @@ async fn fetching_deposit_request_votes() {
...
+                signer_pub_key: *signer_pub_key,
+                can_accept: true,
+                can_sign: true,
+            };
+            pg_store
+                .write_deposit_signer_decision(&decision)
+                .await
+                .unwrap();
+        }
+    }
```
matteojug, line 1446: Note that `TestData::generate` already [generates signers decision](https://github.com/stacks-network/sbtc/blob/main/signer/src/testing/storage/model.rs#L401-L412) (randomly) based on `num_signers_per_request`, and given that `write_deposit_signer_decision` `ON CONFLICT DO NOTHING` this snippet shouldn't have any effect (IIUC)

signer/tests/integration/postgres.rs
```diff
@@ -1400,6 +1400,71 @@ async fn fetching_deposit_request_votes() {
...
+
+    let signer_pub_key = signer_set.first().unwrap();
+
+    let deposit_decisions = pg_store
+        .get_deposit_signer_decisions(&chain_tip.block_hash, 3, signer_pub_key)
+        .await
+        .unwrap();
+
+    assert_eq!(deposit_decisions.len(), 3);
+    for deposit in test_data.deposit_requests[3..].iter() {
```
matteojug, line 1456: qq: why the `[3..]`?

signer/tests/integration/postgres.rs
```diff
@@ -1400,6 +1400,71 @@ async fn fetching_deposit_request_votes() {
...
+    }
+
+    let signer_pub_key = signer_set.first().unwrap();
+
+    let deposit_decisions = pg_store
+        .get_deposit_signer_decisions(&chain_tip.block_hash, 3, signer_pub_key)
+        .await
+        .unwrap();
+
+    assert_eq!(deposit_decisions.len(), 3);
```
matteojug, line 1455: We should also check that we are fetching only our own (by `signer_pub_key`) decisions.

signer/tests/integration/postgres.rs
```diff
@@ -1400,6 +1400,71 @@ async fn fetching_deposit_request_votes() {
...
+    for deposit in test_data.deposit_requests[3..].iter() {
+        assert!(deposit_decisions
+            .iter()
+            .find(|decision| {
+                decision.txid == deposit.txid && decision.output_index == deposit.output_index
```
matteojug, line 1460: This should also check that the matching decision is our own (based on `signer_pub_key`)

signer/src/storage/postgres.rs
Jiloc: Great thinking on the edge case of missing older transactions! A simple approach that avoids modifying the database could be to use the `created_at` column of the bitcoin block in the `bitcoin_blocks` table relative to the selected context_window. With each new block, we can retrieve the creation time of the block received N-updates ago and fetch all decisions with a `created_at` greater than that block's creation time. The only downside is that if blocks arrive very close to each other, we might end up retrying a vote an additional time. What do you think?

signer/src/request_decider.rs
Jiloc: Yeah, I thought about it, and got too scared with all the previous interactions with Emily and concurrent requests 🥲 , but in this case it make sense. We are not processing concurrent messages on the P2P network

signer/src/storage/postgres.rs
matteojug: Oh that's a nice way to avoid extra stuff! And an extra retry shouldn't really be an issue, we would just send 4x more votes message but I don't think that's overwhelming -- but maybe we should double check, just in case.
I'm okay with leveraging `created_at`, but maybe let's check also with @djordon or @cylewitruk , just to be sure it's not an overkill or we are missing some other corner case.

signer/tests/integration/postgres.rs
```diff
@@ -1400,6 +1400,56 @@ async fn fetching_deposit_request_votes() {
...
+    assert_eq!(deposit_decisions.len(), 3);
+    // Test data contains 5 deposit requests, we should get decisions for
+    // the last 3.
+    for deposit in test_data.deposit_requests[2..].iter() {
+        assert!(deposit_decisions.iter().any(|decision| {
+            decision.txid == deposit.txid
+                && decision.output_index == deposit.output_index
+                && decision.signer_pub_key == *signer_pub_key
+        }));
+    }
```
matteojug, line 1448: This doesn't always work: `TestData` generates a chain with forks, so picking the last 3 deposits requests doesn't guarantee that those 3 are all on the canonical chain; it does work with `51` as random seed, but if you change that to `52` the test fails.

signer/src/testing/storage/model.rs
```diff
@@ -515,6 +524,8 @@ pub struct Params {
     pub num_withdraw_requests_per_block: usize,
     /// The number of signers to hallucinate per request
     pub num_signers_per_request: usize,
+    /// Wheter to generate consecutive blocks or not
+    pub concatenate_blocks: bool,
```
matteojug, line 528: nit: `consecutive_blocks`? just to make it clear what's the effect

signer/src/request_decider.rs
```diff
@@ -248,6 +268,21 @@ where
...
+            if let Err(error) = self.send_message(decision, chain_tip).await {
+                tracing::warn!(%error, "error sending deposit decision to retry, skipping");
+            }
```
matteojug, line 281: nit: consider using `inspect_err` for readability

signer/src/testing/storage/model.rs
```diff
@@ -66,8 +68,14 @@ impl TestData {
         let mut test_data = Self::new();
 
         for _ in 0..params.num_bitcoin_blocks {
-            let (next_chunk, _) = test_data.new_block(rng, signer_keys, params, None);
+            let parent = if params.concatenate_blocks {
+                test_data.bitcoin_blocks_ref.last()
+            } else {
+                None
+            };
```
matteojug, line 75: nit: we can use a local and avoid `bitcoin_blocks_ref`

signer/tests/integration/postgres.rs
```diff
@@ -1381,6 +1389,106 @@ async fn fetching_deposit_request_votes() {
...
+
+    let signer_set = testing::wsts::generate_signer_set_public_keys(&mut rng, num_signers);
+
+    let mut test_data = TestData::generate(&mut rng, &signer_set, &test_model_params);
+    test_data.write_to(&pg_store).await;
+
+    let signer_pub_key = signer_set.first().unwrap();
+
+    // move the oldest deposit request to the last block
+    // so that we change its created_at time to be the latest
```
matteojug, line 1418: nit: can you expand a bit what the resulting scenario will look like? So that is easier to follow the setup below

signer/tests/integration/postgres.rs
```diff
@@ -1381,6 +1389,106 @@ async fn fetching_deposit_request_votes() {
...
+    test_data.deposit_requests.rotate_left(1);
+    let mut new_time_block = OffsetDateTime::now_utc() - time::Duration::minutes(15);
+    let mut new_time_request = OffsetDateTime::now_utc() - time::Duration::minutes(12);
+    // Update Bitcoin blocks
+    for (block, deposit) in test_data
```
matteojug, line 1423: nit: can we split this into two different for loops? It doesn't seem we are doing anything with _both_ of them, and since we also changed the coupling above with the `rotate_left` is a bit confusing.

matteojug (approved):

PR #1246, opened 2025-01-20
add sleep after new block to the request_decider

Jiloc:
## Description

Part of : #1237

## Changes
 - originally part of #1243 , this change add a new `requests_processing_delay` to the request_decider. This will give additional time to the bitcoin nodes to get in sync before receiving the vote decisions from the signers. 
 
## Checklist:


djordon (approved):
Looks good.

Let's wait to get more eyes on this before merging, in case there is some edge case that we're blanking on that would make this a bad idea.

matteojug (commented):
Thinking out loud: assuming everyone gets the bitcoin block at exactly the same time, with the sleep as is here we may risk having the coordinator starting the round before the decisions manage to get to it (like, everyone broadcast its own decision, then the coordinator starts but it only has its own decision).
I think this is slightly worse than before, but it could be equivalent.

Jiloc:
How about we separate the two timeouts: one in the RequestDecider after a new block, and another in the TxCoordinator after receiving the signal from the RequestDecider? We could set them both to 15 seconds.

While this isn't an ideal or final solution, I believe it's a reasonable compromise. It should still allow us to receive the decisions and blocks in time most of the time.

#1243  will remain the final solution. And if we implement a filter to only resend messages that didn't receive all replies, we could significantly reduce the number of repeated messages.

signer/src/config/mod.rs
```diff
@@ -223,9 +226,14 @@ pub struct SignerConfig {
     pub bootstrap_signatures_required: u16,
     /// The number of seconds the coordinator will wait
     /// before processing a new Bitcoin block
-    /// (allowing it to propagate to the others signers)
+    /// (allowing it to propagate the request decisions the others signers)
```
matteojug, line 229: very nit:
```suggestion
    /// (allowing the request decisions to propagate to the others signers)
```

signer/src/config/mod.rs
```diff
@@ -223,9 +226,14 @@ pub struct SignerConfig {
...
     #[serde(deserialize_with = "duration_seconds_deserializer")]
     pub bitcoin_processing_delay: std::time::Duration,
+    /// The number of seconds the request decider will wait
+    /// before processing the new sbtc requests
+    /// (allowing it to propagate the bitcoin block the others signers)
```
matteojug, line 234: very nit:
```suggestion
    /// (allowing the bitcoin block to propagate to the others signers)
```

signer/src/config/mod.rs
```diff
@@ -223,9 +226,14 @@ pub struct SignerConfig {
...
+    #[serde(deserialize_with = "duration_seconds_deserializer")]
+    pub requests_processing_delay: std::time::Duration,
```
matteojug, line 236: Do we want this to be a required param? (probably yes, just to confirm)
If so, can you add the `breaking-local` tag to the PR?

matteojug (approved):

signer/src/config/mod.rs
Jiloc: Alternatively, we could give a default value?

matteojug: It makes sense to have this required (as we did for the other sleep), the only nuisance is that we don't know what's the best value for this on mainnet so we may require signers to update the config more than once. But even with defaults we would still need them to update the config (or do a new release, that is even worse). Imho it's okay to have it required.

PR #1249, opened 2025-01-21
feat: Pin action by hash in CI workflows

fabergat:
## Description

Pin action by hash in CI workflows

Closes: https://github.com/stacks-network/sbtc/issues/1235

## Changes

- Add hash to every CI action

## Checklist:

### DIFF
--- a/.github/workflows/buf-ci.yaml
+++ b/.github/workflows/buf-ci.yaml
@@ -14,8 +14,8 @@ jobs:
     name: Run protobuf lints
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v4
-      - uses: bufbuild/buf-action@v1
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
+      - uses: bufbuild/buf-action@3fb70352251376e958c4c2c92c3818de82a71c2b #v1.0.2
         with:
           version: 1.42.0
           setup_only: true
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -15,26 +15,26 @@ jobs:
     name: Run Unit Tests
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
 
-      - uses: pnpm/action-setup@v4
+      - uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2 #v4.0
 
       - name: Set Node Version
-        uses: actions/setup-node@v3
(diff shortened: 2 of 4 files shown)


aldur (approved):

.github/workflows/unit-tests.yaml
```diff
@@ -15,26 +15,26 @@ jobs:
...
       - name: Setup rust toolchain
         run: rustup show
 
-      - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@f0deed1e0edfc6a9be95417288c0e1099b1eeec3 #v2.7.7
 
-      - uses: taiki-e/install-action@v2
+      - uses: taiki-e/install-action@da41fb311fbbcecf899732e575aaeaa2fe65c934 #v2.47.21
         with:
           tool: nextest@0.9.82
```
aldur, line 39: Can this be pinned as well?

.github/workflows/buf-ci.yaml
```diff
@@ -14,8 +14,8 @@ jobs:
     name: Run protobuf lints
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v4
-      - uses: bufbuild/buf-action@v1
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
```
djordon, line 17: Random question, is it possible for someone to somehow overwrite the action associated with this "commit"?

For example, can someone rewrite their git history to remove this commit and create a tag with this same "commit string"? I wasn't sure that the answer is "no".

Regardless of the answer, these changes make sense to me.

djordon (commented):
I was curious and just have a question.

.github/workflows/unit-tests.yaml
fabergat: Tried but the action is not supporting it

.github/workflows/buf-ci.yaml
fabergat: Technically to have the same SHA you should have the same content

djordon: Yeah make sense, but I don't just mean the same SHA. I was thinking more along the lines that someone creates a tag with a string that looks like a SHA. So instead of creating a tag with a normal string like "v1.2.3", you use the string "11bd71901bbe5b1630ceea73d27597364c9af683". So long as this commit does not exist, I would think that the tag string would be valid. And if the commit gets rewritten maybe that SHA is eligible for being a tag but I don't know, hence my question.

.github/workflows/unit-tests.yaml
fabergat: What did you pass in the env? I don't have access but if it works, it could be good!!

.github/workflows/buf-ci.yaml
fabergat: ah now I got your point, yes I guess it is technically possible

Issue #1017, opened 2024-12-03
[Feature]: Add a more comprehensive health endpoint to the signer app


djordon:
# Feature - Add a more comprehensive health endpoint to the signer app

## 1. Description

In order to get the status of a signer we have the status endpoint, `GET /`, and then we have to manually check the database, the connected stacks node, and the connected bitcoin node. Instead we should add an endpoint that returns the relevant information and update our vector config regularly fetch this information for reporting to CloudWatch.

### 1.1 Context & Purpose

We don't have a good way of knowing whether a signer is behind and the source of the issue, since it could be due to bitcoin-core, stacks-core, postgres, or some connectivity issue between them. This endpoint provides a simple way to help with debugging these issues.

## 2. Technical Details:

Add an endpoint to the signer that, using the `ApiState`, returns a JSON with the following:
* The chain-tip hash and height from bitcoin-core.
* The bitcoin-core reported chain (either main, test, signet, regtest).
* The chain-tip hash and height from stacks-core
* The bitcoin chain-tip hash and height from the database.
* The stacks chain-tip hash and height from the database.
* The current timestamp
* (optional) the git sha that the binary was built from.
* (optional) the CPU target architecture of the signer binary build.
* (optional) the target environment ABI of the signer binary build.
* (optional) the version of rustc used to build the signer binary.

If we cannot get a piece of information from the intended source then we return `null` for the field. The optional fields require a little bit of work done in the `build.rs`.

### 2.1 Acceptance Criteria:

   - [ ] We have a new `GET /health` endpoint that returns the above required information.
   - [ ] (Optional) we return the optional information as well in the `GET /health` response.
   - [ ] We update the vector config so that it poll this information using the [HTTP server source](https://vector.dev/docs/reference/configuration/sources/http_server/) and forwards it to our logging sinks.

## 4. Appendix

We should use the `GET /v3/tenures/info` endpoint for getting Stacks information, since it's not clear how well the `GET /v2/info` endpoint works post Nakamoto. For bitcoin, we can get what we want from the [`getblockchaininfo`](https://bitcoincore.org/en/doc/25.0.0/rpc/blockchain/getblockchaininfo/) RPC endpoint.

PR #1251, opened 2025-01-21
feat: signer `/info` api endpoint

cylewitruk:
## Description

Closes: #1017 

## Changes

- Adds a new `/info` endpoint to the signer's API (I'm thinking we reserve `/health` for connectivity checks and reporting connection errors, etc.).
- Adds `getnetworkinfo` and `getblockchaininfo` RPC methods to `BitcoinInteract`.
- Some other minor tweaks

Example response from `devenv`:

```json
{
    "bitcoin": {
        "signer_tip": {
            "block_hash": "4829942fe1d8a876b9550080274ae43c7c9a83517cb19a2e7b08c4163fa6018b",
            "block_height": 237
        },
        "node_tip": {
            "block_hash": "4829942fe1d8a876b9550080274ae43c7c9a83517cb19a2e7b08c4163fa6018b",
            "block_height": 237
        },
        "node_chain": "regtest",
        "node_version": 250100,
        "node_subversion": "/Satoshi:25.1.0/"
    },
    "stacks": {
        "signer_tip": {
            "block_hash": "87abe90e746124ca9df157c7e4e1b0785c3cb6a712ce466dd630fecbb90b1dbd",
            "block_height": 47
        },
        "node_tip": {
            "block_hash": "87abe90e746124ca9df157c7e4e1b0785c3cb6a712ce466dd630fecbb90b1dbd",
            "block_height": 47
        },
        "node_bitcoin_block_height": 237,
        "node_version": "stacks-node 0.0.1 (:49777d3fd73a6dbb610be80c376b7d9389c9871a, debug build, linux [aarch64])"
    },
    "dkg": {
        "rounds": 1,
        "current_aggregate_key": "03cffc8638c02dde18c99a84ebf25c6fce3faf7843ea67152a69b6dfbca80c35cc",
        "contract_aggregate_key": "03cffc8638c02dde18c99a84ebf25c6fce3faf7843ea67152a69b6dfbca80c35cc"
    },
    "config": {
        "network": "regtest",
        "deployer": "SN3R84XZYA63QS28932XQF3G1J8R9PC3W76P9CSQS",
        "bootstrap_signatures_required": 2,
        "bitcoin_processing_delay": 3,
        "context_window": 1000,
        "signer_round_max_duration": 5,
        "bitcoin_presign_request_max_duration": 5,
        "dkg_max_duration": 5,
        "sbtc_bitcoin_start_height": null,
        "dkg_begin_pause": 1,
        "max_deposits_per_bitcoin_block": 25,
        "dkg_min_bitcoin_block_height": null,
        "dkg_target_rounds": 1
    },
    "build_info": {
        "rust_version": "rustc 1.84.0 (9fc6b4312 2025-01-07)",
        "git_revision": "add5a8868d5f5d1df9b58e369e8032cc65b79c3d",
        "target_arch": "aarch64",
        "target_env_abi": null
    },
    "timestamp": "2025-01-24 11:15:22.802073 +00:00:00"
}
```
Values which are not available (i.e. either missing or an error occurred) are returned as `null`, my thinking is that this should report as much information as possible and be infallible, and we use a `/health` endpoint for having simple db/bitcoin/stacks connectivity checks and report errors.

## Testing Information

Added additional unit tests and tested pretty extensively against `devenv`.

## Checklist:


signer/src/error.rs
```diff
@@ -593,6 +593,13 @@ pub enum Error {
         /// Maximum sBTC mintable
         max_mintable: u64,
     },
+
+    /// An error which can be used in test code instead of `unimplemented!()` or
+    /// other alternatives, so that an an actual error is returned instead of
+    /// panicking.
+    #[cfg(test)]
+    #[error("Dummy (for testing purposes)")]
+    Dummy,
```
matteojug, line 602: mmm panicking if we hit some unimplemented test path could be better, otherwise we may "fail" silently since we handle the error in multiple places (I'm thinking like in the tx coordinator event loop) and not test what we actually want to test.

signer/src/api/info.rs
```diff
@@ -0,0 +1,505 @@
...
+        // so this lint is not applicable.
+        #[allow(clippy::const_is_empty)]
+        let target_env_abi = if crate::TARGET_ENV_ABI.is_empty() {
+            None
+        } else {
+            Some(crate::TARGET_ENV_ABI.to_string())
+        };
+
+        Self {
+            bitcoin: BitcoinInfo {
```
matteojug, line 79: nit: we could derive default for `BitcoinInfo` and get the all `None` for free; also for `StacksInfo`

signer/src/api/info.rs
```diff
@@ -0,0 +1,505 @@
...
+impl Default for InfoResponse {
+    fn default() -> Self {
+        // This is a build-time constant and varies per build environment,
+        // so this lint is not applicable.
+        #[allow(clippy::const_is_empty)]
+        let target_env_abi = if crate::TARGET_ENV_ABI.is_empty() {
+            None
+        } else {
+            Some(crate::TARGET_ENV_ABI.to_string())
+        };
```
matteojug, line 76: qq: why are we using this, vs just `crate::TARGET_ENV_ABI.to_string()` (as we do for the others below)?

signer/src/api/info.rs
```diff
@@ -0,0 +1,505 @@
...
+                    return;
+                };
+
+                self.bitcoin.signer_tip = Some(ChainTipInfo {
+                    block_hash: bitcoin_block.block_hash,
+                    block_height: bitcoin_block.block_height,
+                });
+
+                match storage
+                    .get_stacks_chain_tip(&bitcoin_block.block_hash)
```
matteojug, line 158: This is (currently) a very expensive query (no context window constraint), are we okay with having it triggerable so easily?

signer/src/api/info.rs
```diff
@@ -0,0 +1,505 @@
...
+impl InfoResponse {
+    /// Populates the local Bitcoin and Stacks chain tip information.
+    async fn populate_local_chain_info(&mut self, storage: &impl DbRead) {
+        match storage.get_bitcoin_canonical_chain_tip().await {
+            Ok(Some(local_bitcoin_chain_tip)) => {
+                let bitcoin_block = storage
+                    .get_bitcoin_block(&local_bitcoin_chain_tip)
+                    .await
+                    .inspect_err(|e| {
+                        tracing::error!("error reading bitcoin block from the database: {}", e)
```
matteojug, line 142: I'm wondering if it could be helpful to also have those errors in the response, so that we could get useful info without having to dig into logs.

signer/src/api/info.rs
```diff
@@ -0,0 +1,505 @@
...
+        }
+    }
+
+    /// Populates the Stacks node tip information from the provided Stacks client.
+    /// This uses the `/v2/info` RPC endpoint to populate the information.
+    async fn populate_stacks_node_info(&mut self, stacks_client: &impl StacksInteract) {
+        match stacks_client.get_node_info().await {
+            Ok(node_info) => {
+                self.stacks.node_tip = Some(ChainTipInfo {
+                    block_hash: node_info.stacks_tip.0.into(),
```
matteojug, line 218: If this is the block hash and not the block id it could explain the mismatch you mentioned in the PR

signer/src/error.rs
cylewitruk: Yeah I agree, it should be used correctly -- in this case it's used in the tests for the `info_handler()` for simulating an error from the mocked `*Interact` implementations, where I don't want it to panic, and I didn't really want to just choose a random unrelated error either.

signer/src/api/info.rs
cylewitruk: This is the only one which (should) return an empty string if not running in i.e. a container. Instead of `""` I prefer it uses a proper `null`.

signer/src/api/info.rs
cylewitruk: I thought about it, but landed in that we should probably have a separate `/health` endpoint which tests connectivity to the different dependencies and returns errors if encountered.

signer/src/api/info.rs
cylewitruk: Yeah, I assume it's something like that... I might be able to get it if I add a new Stacks RPC that can search for the block (iirc they have some search criteria that might work).

Right now I feel pretty content with at least having the block height and adding that kind of improvement later vs. spending time on it now.

signer/src/api/info.rs
cylewitruk: Was this one of them? I can't remember -- this one has the `LIMIT 1` so I thought it wouldn't be a problem.

matteojug: We have the limit outside, but no limits on the CTE. Checked the execution time of `get_stacks_chain_tip` and it seems the limit it's not helping much:
```
---- postgres_bech::bench_get_stacks_tip::playnet stdout ----
Executed query in 77.858541ms

---- postgres_bech::bench_get_stacks_tip::mainnet stdout ----
Executed query in 123.353708ms

---- postgres_bech::bench_get_stacks_tip::synth_small stdout ---- // bitcoin_blocks: 50, stacks_blocks_per_anchor: 20
Executed query in 1.705084ms

---- postgres_bech::bench_get_stacks_tip::synth stdout ---- // bitcoin_blocks: 1000, stacks_blocks_per_anchor: 20
Executed query in 61.17175ms
```

signer/src/api/info.rs
matteojug: Yeah makes sense, also knowing the difference it's "easy" to crosscheck anyway. Maybe we could rename `block_hash` under `signer_tip` to make it clear that is referring to the block index hash.

signer/src/api/info.rs
```diff
@@ -0,0 +1,505 @@
...
+    pub build_info: BuildInfo,
+    pub timestamp: String,
+}
+
+#[derive(Debug, Serialize)]
+pub struct BuildInfo {
+    pub rust_version: String,
+    pub git_revision: String,
+    pub target_arch: String,
+    pub target_env_abi: Option<String>,
```
djordon, line 33: You can use `&'static str` for all of these, it should work just fine.

signer/src/api/info.rs
```diff
@@ -0,0 +1,505 @@
...
+                self.stacks.node_version = Some(node_info.server_version);
+            }
+            Err(e) => {
+                tracing::error!("error getting stacks node tip: {}", e);
+            }
+        }
+    }
+
+    /// Populates the DKG information from the provided storage.
+    async fn populate_dkg_info(&mut self, storage: &impl DbRead, config: &Settings) {
```
djordon, line 231: I was going to say that we should fetch the aggregate key from the stacks node, but maybe that isn't necessary. I don't have a strong feeling about it so I'm not suggesting one or the other.

signer/src/api/info.rs
```diff
@@ -0,0 +1,505 @@
...
+            Ok(Some(keys)) => {
+                self.dkg.current_aggregate_key = Some(keys.aggregate_key.to_string());
+                match storage.get_encrypted_dkg_shares_count().await {
+                    Ok(count) => {
+                        self.dkg.rounds = count;
+                    }
+                    Err(e) => {
+                        tracing::error!(
+                            "error reading encrypted DKG shares count from the database: {}",
+                            e
```
djordon, line 246: nit:
```suggestion
                            "error reading encrypted DKG shares count from the database: {e}",
                            
```
Same thing elsewhere.

signer/src/storage/model.rs
```diff
@@ -835,7 +836,7 @@ impl From<&BitcoinBlock> for BitcoinBlockRef {
 }
 
 /// The Stacks block ID. This is different from the block header hash.
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize)]
```
djordon, line 839: We should add `#[serde(transparent)]` to these.

djordon (commented):
Took a quick look, will have to circle back.

signer/src/bitcoin/rpc.rs
```diff
@@ -567,6 +569,35 @@ impl BitcoinCoreClient {
...
+        let block = self
+            .inner
+            .get_block_header_info(&best_tip)
+            .map_err(Error::BitcoinCoreRpc)?;
+
+        Ok((
+            best_tip,
+            u64::try_from(block.height).map_err(|_| Error::TypeConversion)?,
+        ))
+    }
```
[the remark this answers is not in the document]
cylewitruk: Removed in [658342a](https://github.com/stacks-network/sbtc/pull/1251/commits/658342adb8964f44e42ed4988d32dab7bf7eb22a)

signer/src/api/info.rs
cylewitruk: [4b13bb1](https://github.com/stacks-network/sbtc/pull/1251/commits/4b13bb1ff6b758507cff56fab35497308e98ab94)

signer/src/api/info.rs
cylewitruk: Yeah that works, updated in [6b19e02](https://github.com/stacks-network/sbtc/pull/1251/commits/6b19e0241bd2f88e178927bbe53c24e763ea77c2)

signer/src/api/info.rs
cylewitruk: [807b6ba](https://github.com/stacks-network/sbtc/pull/1251/commits/807b6ba69aee8fd1c8f5651870d2177c4056da51)

signer/src/api/info.rs
cylewitruk: [514757b](https://github.com/stacks-network/sbtc/pull/1251/commits/514757b8db1c5c2c94cbd8a7746e2c3b92a8e7b5)

signer/src/storage/model.rs
cylewitruk: [a8582d9](https://github.com/stacks-network/sbtc/pull/1251/commits/a8582d963a087e7a1aa0fc58299183e077e4b195)

signer/src/api/info.rs
cylewitruk: Well, why not both? 😄 That let's us see if there's a diff between the contract and the local node, i.e. maybe rotate keys is failing.

[a8582d9](https://github.com/stacks-network/sbtc/pull/1251/commits/a8582d963a087e7a1aa0fc58299183e077e4b195)

signer/src/api/info.rs
cylewitruk: Those look "OK" anyway, at least for now.

The query made here shouldn't lock/block anything, and there's nothing waiting on it other than the caller, so I think this should be okay for the time being. My hope is that we've managed to refactor all of this db stuff in the semi-near future...

cylewitruk:
Note: I added a `config` section as well which dumps some of the signer config values.

djordon (approved):
Looks good. Left one style nit.

signer/src/api/info.rs
```diff
@@ -0,0 +1,730 @@
...
+
+                self.bitcoin.signer_tip = Some(ChainTipInfo {
+                    block_hash: bitcoin_block.block_hash,
+                    block_height: bitcoin_block.block_height,
+                });
+
+                match storage
+                    .get_stacks_chain_tip(&bitcoin_block.block_hash)
+                    .await
+                {
```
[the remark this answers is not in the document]
cylewitruk: Will fix this up when I land and merge 👍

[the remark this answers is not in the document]
cylewitruk: Updated in [be1e8be](https://github.com/stacks-network/sbtc/pull/1251/commits/be1e8be7b6e9e1cb1ca25a3ac3afd24b61b65740)

Issue #1245, opened 2025-01-20
[Chore]: Use conservative initial sBTC limits for the signers


matteojug:
# Chore - Use conservative initial limits for the signers

## 1. Description

The signers fetch sBTC limits from Emily in the block observer, but until they succeed in doing so they use very permissive limits (via `SignerState` and `SbtcLimits` `default()`).

Instead, they should default to conservative limits (ie, preventing any operation). 

## 2. Why It's Necessary

This change ensure we don't violate limits in case Emily is offline and we don't have previously known limits to fall back on.

PR #1254, opened 2025-01-21
feat: use conservative default sBTC limits

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/1245

## Changes

Set the initial sBTC limits to zero.

Also removed some `setup_logging` left overs.

## Testing Information

Add an integration test (without the limit changes it was minting, now it doesn't).

## Checklist:

### DIFF
--- a/signer/src/context/signer_context.rs
+++ b/signer/src/context/signer_context.rs
@@ -1,6 +1,5 @@
 use std::sync::Arc;
 
-use bitcoin::Amount;
 use tokio::sync::broadcast::Sender;
 use url::Url;
 
@@ -14,7 +13,7 @@ use crate::{
     SIGNER_CHANNEL_CAPACITY,
 };
 
-use super::{Context, SbtcLimits, SignerSignal, SignerState, TerminationHandle};
+use super::{Context, SignerSignal, SignerState, TerminationHandle};
 
 /// Signer context which is passed to different components within the
 /// signer binary.
@@ -91,13 +90,6 @@ where
         if let Some(height) = config.signer.sbtc_bitcoin_start_height {
             state.set_sbtc_bitcoin_start_height(height);
         }
-        state.update_current_limits(SbtcLimits::new(
-            Some(Amount::ZERO),
-            Some(Amount::MAX_MONEY),
-            Some(Amount::ZERO),
-            Some(Amount::ZERO),
-            Some(Amount::ZERO),
-        ));
 
         Self {
             config,
--- a/signer/src/context/signer_context.rs
+++ b/signer/src/context/signer_context.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use bitcoin::Amount;
 use tokio::sync::broadcast::Sender;
 use url::Url;
 
@@ -13,7 +14,7 @@ use crate::{
     SIGNER_CHANNEL_CAPACITY,
 };
 
-use super::{Context, SignerSignal, SignerState, TerminationHandle};
+use super::{Context, SbtcLimits, SignerSignal, SignerState, TerminationHandle};
 
 /// Signer context which is passed to different components within the
 /// signer binary.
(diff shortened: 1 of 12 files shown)


signer/src/context/signer_context.rs
```diff
@@ -90,6 +91,13 @@ where
         if let Some(height) = config.signer.sbtc_bitcoin_start_height {
             state.set_sbtc_bitcoin_start_height(height);
         }
+        state.update_current_limits(SbtcLimits::new(
+            Some(Amount::ZERO),
+            Some(Amount::ZERO),
+            Some(Amount::ZERO),
+            Some(Amount::ZERO),
+            Some(Amount::ZERO),
```
djordon, line 99: Don't we want the minimums to be `MAX_MONEY`? Actually, I support it doesn't matter.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3075,3 +3083,420 @@ async fn test_get_btc_state_with_available_sweep_transactions_and_rbf() {
...
+    //   coordinator)
+    // =========================================================================
+    let signers_key = setup.signers.signer_keys().iter().cloned().collect();
+    loop {
+        let chain_tip: BitcoinBlockHash = faucet.generate_blocks(1).pop().unwrap().into();
+
+        if given_key_is_coordinator(signers[0].2.public_key().into(), &chain_tip, &signers_key) {
+            break;
+        }
+    }
```
djordon, line 3466: We could also just loop through all signers and break when we figure out who the coordinator is and not their index. I just want to limit the number of bitcoin blocks are created, since that slows down tests. Well, at least until Cyle's containerized CI PR lands.

djordon: Oh wait, this isn't so straightforward. Okay let me think.

djordon: Yeah, this logic seems required given what we want to test. Let's roll with what is here and try to change to each test having their own bitcoin node container. If we cannot do that for some reason then we can figure something else out.

djordon (approved):
Looks good!

I suppose this way was more straightforward than changing `SbtcLimits::default` like you mentioned in stand-up?

signer/tests/integration/transaction_coordinator.rs
matteojug: Yeah this is definitely suboptimal, we could try to figure it out beforehand who is the coordinator or change the guard on the limits mock to be able to set it dynamically, but given the CI/integration tests improvements on the horizon I would +1 on waiting to see if things are bad and we need to address this. Also this test uses 3 signers so it shouldn't require too many blocks before getting the right coordinator.

matteojug:
> I suppose this way was more straightforward than changing SbtcLimits::default like you mentioned in stand-up?

Kinda: the end result would have been pretty much the same I guess (modulo the occurrences to replace), it just seemed to me having the explicit limit in the context was cleaner than relying on a _not so natural_ default. But no strong feeling about it, happy to change if you think the default would be a better way.

signer/src/context/signer_context.rs
matteojug: Yeah that's better, changed.

djordon:
> > I suppose this way was more straightforward than changing SbtcLimits::default like you mentioned in stand-up?
> 
> Kinda: the end result would have been pretty much the same I guess (modulo the occurrences to replace), it just seemed to me having the explicit limit in the context was cleaner than relying on a _not so natural_ default. But no strong feeling about it, happy to change if you think the default would be a better way.

I don't have a strong preference either, being explicit seems fine. We may want to remove the default implementation and replace it with different function that does the same thing. Like `SbtcLimits::unlimited` or something.

djordon (approved):
Looks good ✅

Issue #1253, opened 2025-01-21
[Feature]: Support a Ledger friendly NUMS internal key for deposits


MCJOHN974:
# Feature - Support a Ledger friendly NUMS internal key for deposits

## 1. Description

Currently, we have a situation when some people who use ledger can't sign taproot transaction for reclaim script.  To deal with this we need to switch our pubkey used by Signers during signing (deposit reject??) transaction to pubkey derived from some (which?) key, with known derivation path.

### 1.1 Context & Purpose
Allow ledger users to reclaim their BTC.

## 2. Technical Details:
We need:
- To figure out which key we want to use and what it's derivation path is
- Switch current key to key we will find
- Test it somehow

I think this https://github.com/stacks-network/sbtc/blob/a53ab63e397a547831b6a1298cef7490724b7cb1/sbtc/src/lib.rs#L48 is probably the key we want to switch. 

### 2.1 Acceptance Criteria:
   - [ ] 

## 3. Related Issues and Pull Requests (optional):

@aldur and @djordon can you please edit this issue a bit?

## 4. Appendix

The remaining open question for the signers is whether they want to support more than one internal key for deposits. It's relatively straightforward to support more than one internal key.

The current NUMS public key can be found here:

https://github.com/stacks-network/sbtc/blob/891677e7602f97b2e088e9940b2b4536ae9aad1b/sbtc/src/lib.rs#L20-L37

It gets used during validation of the deposit data, which happens here:

https://github.com/stacks-network/sbtc/blob/55ffd3757f720a8675461c1fce3b2fe1ca33428d/sbtc/src/deposits.rs#L161-L167

Regardless of what the signers support, changing the internal key needs happen on the bridge for ledger.

aldur:
Please document exactly _how_ the new keys are being derived and how users (and our-selves) can independently verify that they are unspendable.

PR #1255, opened 2025-01-22
[feature] support reclaim from Ledger

MCJOHN974:
## Description

Closes: #1253

## Checklist:

### DIFF
--- a/sbtc/src/lib.rs
+++ b/sbtc/src/lib.rs
@@ -64,7 +64,7 @@ mod tests {
     use secp256k1::{PublicKey, Secp256k1};
     use std::str::FromStr;
 
-    /// This test proves, that the DERIVED_NUMS_X_COORDINATE is the correct,
+    /// This test proves, that the DERIVED_NUMS_X_COORDINATE is correct,
     /// and is derived from the NUMS_X_COORDINATE using the derivation path "0/0".
     #[test]
     fn test_derivation() {


sbtc/src/lib.rs
```diff
@@ -46,4 +55,34 @@ pub const NUMS_X_COORDINATE: [u8; 32] = [
...
+mod tests {
+    use super::*;
+    use bitcoin::bip32::{ChainCode, ChildNumber, DerivationPath, Fingerprint, Xpub};
+    use secp256k1::{PublicKey, Secp256k1};
+    use std::str::FromStr;
+
+    /// This test proves, that the DERIVED_NUMS_X_COORDINATE is correct,
+    /// and is derived from the NUMS_X_COORDINATE using the derivation path "0/0".
+    #[test]
+    fn test_derivation() {
```
djordon, line 70: One alternative is to define the `UNSPENDABLE_TAPROOT_KEY` using the same logic of this test.

djordon (commented):
Looks good.

One suggestion is to put the logic for how we derive the key in the `LazyLock`'s function definition. The derivation is fast and only happens once. We would still want a test that the output bytes equal what we expect them to be.

Also, after doing some digging, it seems like this change is indeed required. Ledger requires a wallet policy all the time, and so we need to prove that we know the derivation path. I'd still like to hear it from a Ledger engineer if possible (in a reasonable amount of time).

djordon:
> I think this two options are +- equal and I don't prefer much one to another. But I see a small benefit from option I initially chose: we move derivation logic out of main code making it simpler and easier to read. I think in most cases when you are looking through the sbtc code you are not very interested how we choose the internal key, and when you interested in it, you will look at the test
> 
> What benefits you think we will get moving derivation logic from tests to main code?

Well, I don't prefer one way over the other. I can see a benefit of having the derivation logic in the main code path, since it makes it obvious how the key was derived. It would also be clear that this key is "genuinely unspendable". What's here seems pretty obvious as well, and probably more readable too so I'm happy what keeping it as is. It really was just a suggestion.

Issue #1133, opened 2024-12-14
[Feature]: Retry p2p seed peers


djordon:
# Feature - Retry p2p seed peers

### 2.1 Acceptance Criteria:
   - [ ] If a signer cannot reach out to the seed peers, they should try again after some delay.

cylewitruk:
Hm.. this should be done automatically. I'm guessing that we may be incorrectly purging them at some point, where we should probably never automatically remove a seed peer.

djordon:
I see an issue like this locally, where one signer does not connect to any of the others. Well, that's probably a different issue than what happens in production though (if it happens at all). Anyway, this ticket might be a little premature, but I thought that it might be an actual issue given what I see locally.

cylewitruk:
> I see an issue like this locally, where one signer does not connect to any of the others. Well, that's probably a different issue than what happens in production though (if it happens at all). Anyway, this ticket might be a little premature, but I thought that it might be an actual issue given what I see locally.

I'll have a peek tomorrow and see if I can't reproduce; not sure i've seen it 🤔

PR #1257, opened 2025-01-22
feat: new behavior for network bootstrapping

cylewitruk:
## Description

Closes: #1133

## Changes

- Adds a new libp2p swarm behavior for handling network bootstrapping.
- The swarm will now re-try connections to all seed peers if the connection count is 0, and re-try every minute until the swarm has peers.

## Testing Information

Existing tests pass and I've tested this quite a bit in `devenv`.  I'll add some actual tests for the behavior, but PR'ing this now for preview.

## Checklist:


signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,314 @@
...
+                    tracing::debug!(peer_id = %e.peer_id, %addr, "connected to seed");
+                } else {
+                    return;
+                }
+
+                // Update our connected seeds map.
+                match self.connected_seeds.entry(e.peer_id) {
+                    Entry::Occupied(mut entry) => {
+                        if entry.get_mut().insert(addr) {
+                            tracing::trace!("added connected seed: {:?}", e.peer_id);
```
matteojug, line 188: nit: I guess this trace log is also valid for the arm below; we could return if we don't insert here and move the log after the match.

Also, is this tracing log needed? Considering that we already `debug` log (with more info) right above; if we want to keep it, maybe it should also include the `addr` (in case we have multiple addresses for the same peer id); also applied to the same log in `ConnectionClosed` below

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,314 @@
...
+                        entry.insert(set);
+                    }
+                }
+            }
+            FromSwarm::ConnectionClosed(e) => {
+                // If this was a connection to ourselves then we don't need to do
+                // anything.
+                if e.peer_id == self.local_peer_id() {
+                    return;
+                }
```
matteojug, line 203: Shouldn't we also return if we are not the dialer, like above? Otherwise we would decrement `self.connection_count` even when it's some other remote disconnecting from us.

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,314 @@
...
+
+                // If we've connected to ourselves, we don't need to do anything.
+                // This might happen if a node has its own address in the seed list.
+                if e.peer_id == self.local_peer_id() {
+                    return;
+                }
+
+                // Increment our total connection count. We're not interested
+                // in counting connections to ourselves.
+                self.connection_count = self.connection_count.saturating_add(1);
```
matteojug, line 168: Shouldn't this take into account only seed connections? Otherwise it could be possible that some subset of nodes are still connected one another but none is connected to any seed (creating a partition).

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,314 @@
...
+                // Check if the address is a seed address. If it's not, then
+                // we don't need to do anything so we return early.
+                if self.is_seed_address(&addr) {
+                    tracing::debug!(peer_id = %e.peer_id, %addr, "connected to seed");
+                } else {
+                    return;
+                }
+
+                // Update our connected seeds map.
+                match self.connected_seeds.entry(e.peer_id) {
```
matteojug, line 185: nit: we could simplify this match with something like `self.connected_seeds.entry(e.peer_id).or_default().insert(addr);` (eventually with the check and log if inserted).

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,314 @@
...
+        // If we've attempted to bootstrap recently, we wait until the interval
+        // has passed.
+        if let Some(last_bootstrap) = self.last_attempted_at {
+            if last_bootstrap.elapsed() < self.config.bootstrap_interval {
+                return Poll::Pending;
+            }
+        }
+
+        // Queue the bootstrap started event.
+        tracing::debug!(addresses = ?self.config.seed_addresses, "initiating network bootstrapping from seed addresses");
```
matteojug, line 288: nit: since we already log this event in the libp2p event loop, isn't this a duplicated log?

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,314 @@
...
+            let event = ToSwarm::Dial { opts: dial_opts };
+
+            // Queue the dial event.
+            self.pending_events.push_back(event);
+        });
+
+        // Update the last bootstrap attempt time.
+        self.last_attempted_at = Some(Instant::now());
+
+        Poll::Pending
```
matteojug, line 312: qq: why not popping and returning the head of `self.pending_events` here directly?

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,314 @@
...
+            self.is_bootstrapped = false;
+            self.pending_events
+                .push_back(ToSwarm::GenerateEvent(BootstrapEvent::Needed));
+        }
+
+        // If we've attempted to bootstrap recently, we wait until the interval
+        // has passed.
+        if let Some(last_bootstrap) = self.last_attempted_at {
+            if last_bootstrap.elapsed() < self.config.bootstrap_interval {
+                return Poll::Pending;
```
matteojug, line 283: nit: should this return directly the head of `self.pending_events` (if any)?

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    fn local_peer_id(&self) -> PeerId {
+        self.config.local_peer_id
+    }
+
+    pub fn add_seed_addresses<T>(&mut self, seed_addresses: T) -> &mut Self
+    where
+        T: IntoIterator<Item = Multiaddr>,
+    {
+        self.config.seed_addresses.extend(seed_addresses);
```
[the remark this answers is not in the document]
cylewitruk: [2bfd583](https://github.com/stacks-network/sbtc/pull/1257/commits/2bfd583dbdf5796423e7bed561c13c3c6d4831d5)

signer/src/network/libp2p/event_loop.rs
```diff
@@ -166,6 +167,17 @@ pub async fn run(ctx: &impl Context, swarm: Arc<Mutex<Swarm<SignerBehavior>>>) {
                             "autonat server event"
                         );
                     }
+                    SwarmEvent::Behaviour(SignerBehaviorEvent::Bootstrap(event)) => match event {
+                        BootstrapEvent::Complete => {
+                            tracing::info!("network bootstraping complete");
```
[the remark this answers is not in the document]
cylewitruk: [6818b97](https://github.com/stacks-network/sbtc/pull/1257/commits/6818b971a6f0f867f0893bc8e57edbf85df61806)

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        for addr in &addresses {
+            if self.is_seed_address(addr) {
+                tracing::debug!(%addr, "attempting to dial seed address");
+            }
+        }
+
+        // Return the addresses as-is.
+        Ok(addresses)
```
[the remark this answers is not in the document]
cylewitruk: Yup. It's not that relevant for this behavior, but this is where we could filter or explicitly deny outbound connection attempts if we wanted to. I should actually change this to only react (log) on the `ConnectionId`'s created by this behavior.

signer/src/network/libp2p/bootstrap.rs
cylewitruk: These are `trace` for debugging purposes, which we generally don't enable (except when I do to debug this). But I added the additional tracing fields since the interpolation needed to be removed anyway. [67c5dc2](https://github.com/stacks-network/sbtc/pull/1257/commits/67c5dc2c13afa7828389e1a9679b0b78f492788c)

signer/src/network/libp2p/bootstrap.rs
cylewitruk: I think I need to revisit this. The idea is that this behavior should kick in when the current connection count is zero, otherwise it doesn't need to do anything (and let the other behaviors i.e. Kademlia handle the rest). So I should probably be counting the number of connections based on a `HashSet<PeerId>` instead.

signer/src/network/libp2p/bootstrap.rs
cylewitruk: No -- the idea with the seed nodes is that they're a well-known source of additional peers, but there should be no requirement that peers maintain active connections to the seed nodes.

I get what you're thinking, but that's largely due to the fact that we have exactly one seed node today -- we should change this so that we have at least 2 or 3 trusted signers who can commit to maintaining a static IP and be available for inbound connections, and have all signers add them to their seed list.

Also, implementing persistence of discovered peers to the DB would also alleviate any issues around seed nodes restarting, so they could attempt to re-connect to their most recently connected peers and re-discover nodes.

signer/src/network/libp2p/bootstrap.rs
cylewitruk: Yup, good one [76f0815](https://github.com/stacks-network/sbtc/pull/1257/commits/76f08156e6574538cc336264ed22c1d96769d094)

signer/src/network/libp2p/bootstrap.rs
cylewitruk: I'd like to keep this one for the time being -- my plan is that we move more and more of the event loop into behaviors, thus removing that logging.

signer/src/network/libp2p/bootstrap.rs
cylewitruk: Good point, [f87c47e](https://github.com/stacks-network/sbtc/pull/1257/commits/f87c47ee29ba6d110a49fd7a28977f51b24035b1)

signer/src/network/libp2p/bootstrap.rs
cylewitruk: I moved all of the logging into the behavior and removed it from the event loop.

cylewitruk:
@matteojug I ended up doing quite a bit of refactoring, sorry :( So this probably needs a fresh look.

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,446 @@
...
+        addresses: &[Multiaddr],
+        _effective_role: Endpoint,
+    ) -> Result<Vec<Multiaddr>, ConnectionDenied> {
+        let addresses = addresses.to_vec();
+
+        // If the connection ID is in the pending connections set, then we know
+        // that this is a seed peer dial that we've initiated, so we publish
+        // the `DialingSeed` event.
+        if self.pending_connections.contains(&connection_id) {
+            tracing::debug!(%connection_id, addresses = ?addresses, peer_id = ?maybe_peer, "attempting to dial seed peer");
```
matteojug, line 306: nit?
```suggestion
            tracing::debug!(%connection_id, ?addresses, peer_id = ?maybe_peer, "attempting to dial seed peer");
```

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,446 @@
...
+    }
+
+    #[tracing::instrument(skip_all)]
+    fn poll(
+        &mut self,
+        _cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<libp2p::swarm::ToSwarm<Self::ToSwarm, libp2p::swarm::THandlerInEvent<Self>>>
+    {
+        // If we have any pending events, we return them immediately.
+        if let Some(event) = self.pending_events.pop_front() {
```
matteojug, line 356: nit: use `self.next_pending_event()`? It may end up _uglier_, but at least we are using the same interface

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,446 @@
...
+            self.pending_events
+                .push_back(ToSwarm::GenerateEvent(BootstrapEvent::NoConnectedPeers));
+        }
+
+        // If we've attempted to bootstrap recently, we wait until the interval has passed.
+        if let Some(last_bootstrap) = self.last_attempted_at {
+            if last_bootstrap.elapsed() < self.config.bootstrap_interval {
+                return self.next_pending_event();
+            }
+        }
```
matteojug, line 398: Sanity check: we are going to get poll'd even if there's no swarm activity, right? I'm thinking if it could happen that we need to bootstrap again, but we consume all the events so if we don't get poll'd anymore we would be stuck.

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,446 @@
...
+
+        // If this is the first attempt, we wait for the initial delay.
+        if let Some(first_attempt) = self.first_attempt_at {
+            // If the initial delay has not elapsed, we continue waiting and
+            // return the next pending event.
+            if first_attempt.elapsed() < self.config.initial_delay {
+                return self.next_pending_event();
+            }
+        } else if self.config.initial_delay > Duration::from_secs(0) {
```
matteojug, line 407: nit: `Duration::ZERO`?

signer/src/network/libp2p/multiaddr.rs
```diff
@@ -64,14 +69,31 @@ impl MultiaddrExt for Multiaddr {
...
+        }
+
+        let mut addr = Multiaddr::empty();
+        for part in self.iter() {
+            if matches!(part, Protocol::P2p(_)) {
+                break;
+            }
+            addr.push(part.clone());
+        }
+        addr
```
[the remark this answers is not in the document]
cylewitruk: It's an `Arc` internally so it will modify the original address

signer/src/network/libp2p/bootstrap.rs
cylewitruk: Yeah, _we_ are responsible for driving/polling the swarm, which we do at an interval in the event loop, so every time we do that will trigger a poll on all enabled behaviors.

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,446 @@
...
+            .unwrap_or(Poll::Pending)
+    }
+
+    /// Gets the total number of connected peers.
+    pub fn connected_peer_count(&self) -> usize {
+        self.connected_peers.len()
+    }
+
+    /// Gets the [`PeerId`]s of all connected peers.
+    pub fn connected_peers(&self) -> Vec<(PeerId, Multiaddr)> {
```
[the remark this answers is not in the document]
cylewitruk: Yeah we could return a map, not sure why I chose this structure but looking through it I don't see any explicit reason 🤷

signer/src/network/libp2p/multiaddr.rs
```diff
@@ -198,4 +220,52 @@ mod tests {
...
+            .with(Protocol::Tcp(8080))
+            .with(Protocol::P2p(PeerId::random()));
+        let expected = Multiaddr::empty()
+            .with(Protocol::Ip4(IP4_LOOPBACK))
+            .with(Protocol::Tcp(8080));
+        assert_eq!(
+            multiaddr.without_p2p_protocol(),
+            expected,
+            "ip4 + tcp + p2p"
+        );
```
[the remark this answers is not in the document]
cylewitruk: [8f46a05](https://github.com/stacks-network/sbtc/pull/1257/commits/8f46a051cac554db917ba49e32f1bf80d3036e40)

signer/src/network/libp2p/bootstrap.rs
cylewitruk: [832a2de](https://github.com/stacks-network/sbtc/pull/1257/commits/832a2deb73fb1cbdc0f735319d864c89e21d6df8)

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -2,32 +2,42 @@
...
 }
 
 impl Config {
+    /// Creates a new [`Config`] instance with the provided local peer ID.
     pub fn new(local_peer_id: PeerId) -> Self {
         Self {
             local_peer_id,
             seed_addresses: Default::default(),
             bootstrap_interval: Duration::from_secs(60),
+            initial_delay: Duration::from_secs(0),
```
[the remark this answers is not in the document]
cylewitruk: [832a2de](https://github.com/stacks-network/sbtc/pull/1257/commits/832a2deb73fb1cbdc0f735319d864c89e21d6df8)

signer/src/network/libp2p/bootstrap.rs
cylewitruk: [dcf6488](https://github.com/stacks-network/sbtc/pull/1257/commits/dcf6488431f7b145e1d1bf5170c7e6f0cf3ea906)

signer/src/network/libp2p/bootstrap.rs
cylewitruk: Sure: [f982da0](https://github.com/stacks-network/sbtc/pull/1257/commits/f982da0952823fa72b602afd620ff364ca473374)

signer/src/network/libp2p/bootstrap.rs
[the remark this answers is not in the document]
cylewitruk: Changed to a map in [4f7671f](https://github.com/stacks-network/sbtc/pull/1257/commits/4f7671f13da4110052894dba00cec65c74d5dab2).

> Now that this is a vec, we no longer need to collect (let peers = swarm.connected_peers().collect::<Vec<_>>();) in libp2p event loop.

This is `Swarm::connected_peers()` (which returns an iterator), not this function :)

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -0,0 +1,447 @@
...
+        Self {
+            local_peer_id,
+            seed_addresses: Default::default(),
+            bootstrap_interval: Duration::from_secs(60),
+            initial_delay: Duration::ZERO,
+        }
+    }
+
+    /// Adds seed addresses to the configuration.
+    #[allow(dead_code)]
```
matteojug, line 41: nit: we should be able to remove a couple of those `#[allow(dead_code)]` in here

matteojug (dismissed):

signer/src/network/libp2p/swarm.rs
```diff
@@ -322,30 +330,19 @@ impl SignerSwarm {
...
-        // Start listening on the listen addresses.
-        for addr in self.listen_addrs.iter() {
-            self.swarm
-                .lock()
-                .await
-                .listen_on(addr.clone())
-                .map_err(|e| SignerSwarmError::LibP2P(Box::new(e)))?;
-        }
+        {
+            let mut swarm = self.swarm.lock().await;
```
djordon, line 334: We just locked, and unlock the swarm a moment ago for the `local_peer_id`. Maybe we should move that log down here.

signer/src/network/libp2p/swarm.rs
```diff
@@ -129,9 +141,9 @@ impl<'a> SignerSwarmBuilder<'a> {
             seed_addrs: Vec::new(),
             external_addresses: Vec::new(),
             enable_mdns: false,
-            enable_tcp_transport: true,
             enable_quic_transport: false,
             enable_memory_transport: false,
+            initial_bootstrap_delay: Duration::from_secs(0),
```
djordon, line 146: nit: this could be `Duration::ZERO`.

djordon (approved):
Just had some nits, and one question.

signer/src/network/libp2p/swarm.rs
```diff
@@ -213,31 +219,35 @@ impl<'a> SignerSwarmBuilder<'a> {
...
-                .upgrade(Version::V1)
-                .authenticate(noise.clone())
-                .multiplex(yamux.clone())
-                .boxed();
-            transport = transport
-                .or_transport(tcp_transport)
-                .map(|either, _| either.into_inner())
-                .boxed();
-        }
+        let tcp_config = tcp::Config::default().nodelay(true);
```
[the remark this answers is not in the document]
cylewitruk: It disables [Nagle's Algorithm](https://en.wikipedia.org/wiki/Nagle%27s_algorithm) which is enabled by default on most OS's. I think this option might be set by default now, but I added it explicitly to be, well, explicit as to our intentions.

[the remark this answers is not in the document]
djordon: 👍🏽 I see, makes sense. I wasn't aware of Nagle's algorithm and the problems that could cause. And I didn't know that this was actually the default for us here either.

signer/src/network/libp2p/swarm.rs
cylewitruk: [ffff750](https://github.com/stacks-network/sbtc/pull/1257/commits/ffff750d9634716e2f46315d6c1e56ca8440e0b0)

PR #1262, opened 2025-01-23
chore: bump wsts revision

djordon:
## Description

WSTS signer state machines now take a parameter for the DKG key threshold. This PR uses that commit and we make changes after their API changed.

## Changes

* Update the version of WSTS
* Update the protobufs after the new change
* Update our use of WSTS signer state machines after the API change.

## Testing Information

This is a dependency update, where the dependency, WSTS, is more strict. If tests pass then we should be good to go.

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8048,7 +8048,7 @@ dependencies = [
 [[package]]
 name = "wsts"
 version = "10.0.0"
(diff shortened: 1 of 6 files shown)


xoloki (approved):
LGTM

matteojug (approved):
LGTM

PR #1265, opened 2025-01-24
Update to wsts 13

### LINKED ISSUE #1162: [Bug]: Signers can crash each other with invalid `DkgPrivateShares`

### LINKED ISSUE #1166: [Bug]: Missing length check when parsing `SignatureShareRequest`

### LINKED ISSUE #1167: [Bug]: Correctly handle WSTS nonce issues

xoloki:
## Description

Update the `wsts` dependency to `13.0.0`

Closes #1162, closes #1166, and closes #1167 

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ stackslib = { git = "https://github.com/stacks-network/stacks-core", rev = "4977
 stacks-common = { git = "https://github.com/stacks-network/stacks-core", rev = "49777d3fd73a6dbb610be80c376b7d9389c9871a", default-features = false, features = ["canonical"] }
 
 # Trust Machines Dependencies
-wsts = { git = "https://github.com/Trust-Machines/wsts.git", rev = "a7bf38bd54cddf0b78c0f7c521a5ed1537a684fa" }
+wsts = { version = "12.0.0" }
 
 # Crates.io
 aquamarine = { version = "0.6.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ stackslib = { git = "https://github.com/stacks-network/stacks-core", rev = "4977
 stacks-common = { git = "https://github.com/stacks-network/stacks-core", rev = "49777d3fd73a6dbb610be80c376b7d9389c9871a", default-features = false, features = ["canonical"] }
 
 # Trust Machines Dependencies
-wsts = { version = "13.0.0" }
+wsts = { version = "13.0.1" }
 
 # Crates.io
 aquamarine = { version = "0.6.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ stackslib = { git = "https://github.com/stacks-network/stacks-core", rev = "4977
 stacks-common = { git = "https://github.com/stacks-network/stacks-core", rev = "49777d3fd73a6dbb610be80c376b7d9389c9871a", default-features = false, features = ["canonical"] }
 
 # Trust Machines Dependencies
-wsts = { version = "12.0.0" }
+wsts = { version = "13.0.0" }
 
 # Crates.io
 aquamarine = { version = "0.6.0", default-features = false }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7652,9 +7652,9 @@ dependencies = [
 
 [[package]]
 name = "wsts"
-version = "12.0.0"
+version = "13.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5768ec22f0c646d9992acdd5374cf43f6d0d5aa94bc8016ad55d685f49b72cac"
+checksum = "a1f5cbb1c3934f4ee5c835b5b6c9dc8810ee63ed4122537ab0e399aba2557499"
 dependencies = [
  "aes-gcm",
  "bs58 0.5.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7652,8 +7652,9 @@ dependencies = [
 
 [[package]]
 name = "wsts"
-version = "10.0.0"
-source = "git+https://github.com/Trust-Machines/wsts.git?rev=a7bf38bd54cddf0b78c0f7c521a5ed1537a684fa#a7bf38bd54cddf0b78c0f7c521a5ed1537a684fa"
+version = "12.0.0"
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -75,7 +75,7 @@ tracing = { version = "0.1", default-features = false }
 tracing-attributes = "0.1"
 url = "2.5"
 warp_lambda = "0.1.4"
-wsts = "9.2.0"
+wsts = "11.0.0"
```
cylewitruk, line 78: What's the purpose of this one being `11.0.0` and the `signer` once being a git commit hash?

xoloki: I was testing wsts main branch to make sure it was ready for tag and publish.  I just checked in a change that moves the top level dependency to `12.0.0` and the signer dependency to `workspace`.

xoloki: This branch is a fork of the `update-wsts-11` branch which is why the top-level dependency was `11.0.0`

cylewitruk: Ah okay, thx 👍

matteojug (commented):
sBTC side:
 - We can remove `SignerStateWrapper` and party state one now that they implement eq (in https://github.com/stacks-network/sbtc/blob/main/signer/src/proto/convert.rs#L1802-L1834) and move the tests in `convert_protobuf_type3` to `convert_protobuf_type2`

WSTS side:
1. in `check_public_shares`: maybe switch the order of the checks (first len, then verify)? I'm not sure if `verify` does something bad if the poly len is wrong
2. `make_shared_secret_from_key` (calling `ansi_x963_derive_key`) generates a different key now because of the counter and the order of hash updates, is it fine (wrt backward compatibility with data already generated)? Does this makes this a breaking protocol change (ie, if signers are running different versions can they still interact as expected)?
3. maybe nit: `fn gen<RNG: RngCore + CryptoRng>(rng: &mut RNG) -> Scalar`: as this don’t use a secret, doing the generation as in 4.1 (see comment) doesn’t seem to give any better guarantees: if the rng is a bad one, this will still generate bad nonces

xoloki:
@matteojug:

> * We can remove `SignerStateWrapper` and `PartyStateWrapper` now that they implement eq (in https://github.com/stacks-network/sbtc/blob/main/signer/src/proto/convert.rs#L1802-L1834) and move the tests in `convert_protobuf_type3` to `convert_protobuf_type2`
> 

Will do.

> 1. in `check_public_shares`: maybe switch the order of the checks (first len, then verify)? I'm not sure if `verify` does something bad if the poly len is wrong

`verify` only uses the first coefficient of the polynomial, so it won't do anything bad if the length is wrong.

> 2. `make_shared_secret_from_key` (calling `ansi_x963_derive_key`) generates a different key now because of the counter and the order of hash updates, is it fine (wrt backward compatibility with data already generated)? Does this makes this a breaking protocol change (ie, if signers are running different versions can they still interact as expected)?

Good point; this would definitely break if one signer was using old code with a different key derivation.

> 3. maybe nit: `fn gen<RNG: RngCore + CryptoRng>(rng: &mut RNG) -> Scalar`: as this don’t use a secret, doing the generation as in 4.1 (see comment) doesn’t seem to give any better guarantees: if the rng is a bad one, this will still generate bad nonces

Interesting, I read this as basically saying to hash two 32-bytes random values rather than just using a single 32 byte value directly.  But yeah if the RNG is predictable then you really aren't gaining much.

Looking over the rest of the doc, it looks like "secret key" is used mostly to mean the key shares, but in some places it isn't really defined at all.  We could use the key shares, or pass in the ECDSA key.  What do you think?

xoloki:
> Just to double check: we store (encrypted) the whole `SignerState`, that contains the signer `PartyState` and has some suspecting `private_keys`; this is not related to what you're saying, correct?

DkgPrivateShares are not stored in those state objects.  The polynomials and key shares (which are called private keys) are.

> So if we don't have something secret, the security of it should be the same as using directly the RNG without the extra manipulation.
> 
> Given how it is used in `nonce_generate`, I think it should be fine using either (since it's later hashed).

Agreed.

djordon (approved):
Looks good.

PR #1271, opened 2025-01-25
chore: dependency cleanup pre-`cargo-vet`

cylewitruk:
## Description

I decided to spend my Saturday cleaning up our dependencies before we start using `cargo vet` more seriously. Things have gotten a little "messy" through the crunching.

ℹ️ The PR looks big, but the majority is `Cargo.lock` and `cargo vet` config:

```
$ git diff --stat main . | sort -k 3 -n -r
 Cargo.lock                                 | 1846 ++++++++++++----------------
 supply-chain/config.toml                   |  594 ++++-----
 ...
```

#### Summary

- Consolidated all dependencies to the workspace `Cargo.toml`:
  - All workspace members' dependencies now use `workspace = true`.
  - All dependencies are now specified with `default-features = false` and only necessary features explicitly enabled.
  - All dependencies are now specified with full `semver` (i.e. `x.x.x`).
  - We now only have one version of each direct dependency in our workspace.
- Removed unused dependencies and features.
- Updated all dependencies I could with no or minimal changes (with the exception of `mockito`, see below).
- Exclusions for `cargo-vet` have been updated.
- All `cargo` commands in the `Makefile` now use `--locked`.

#### More Details

- I used `cargo machete --with-metadata` to help me find unused dependencies. False-positives were added to crates' `Cargo.toml` under the `[package.metadata.cargo-machete]` stanza -- it seems less smart when it comes to build- and git-deps.
- Based on the [Rust 2024 edition notes](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/cargo-inherited-default-features.html) I was under the impression that we also needed to add `default-features = false` for each workspace member dependency, but I tested this and workspace members do seem to inherit this configuration  even though the docs say it implicitly enables it 🤷 So I've removed it from workspace members.
- Now that all dependencies are updated and added in the `cargo vet` config as exclusions, I figure that from here on out, we should actually make attempts at auditing the changes when we update dependencies via `cargo vet audit` so that it's recorded.
  - Note that using `--locked` now becomes pretty important unless you're explicitly trying to upgrade things as there's no guarantee that our dependencies don't specify just `x` or `x.y` for their dependencies.
- I ran `cargo sort -g -w` to sort and make the `Cargo.toml` files consistent in formatting.

#### Dependency Upgrade Notes
- `fake` can't be upgraded `>2.9.2` as it causes a problem with one of the withdrawal integration tests.
- `config` can't be upgraded `>0.14.1` due to changes in how it reports errors which breaks a number of tests.
- `secp256k1` can't be upgraded `>0.29.0` due to conflict in bitcoin .
- `reqwest` can't be upgraded `>0.11.27` due to breaking changes.
- `hashbrown` can't be upgraded past `>0.14.5` due to version conflicts with fake/dummy among others.
- `utoipa` can't be upgraded past `>4.2.3` due to breaking changes.
- `libp2p` updates are duplicated from #1257.
  - This also closes https://github.com/stacks-network/sbtc/security/dependabot/6.
- `mockito` version consolidation required changes in `blocklist-client` due to having used a much older version, so that's why there's some larger changes there. Please give this some 👀 -- these were not using `Mock::assert()` previously which after adding them revealed some errors in the tests.
- `thiserror` upgrade required a minor formatting change in `signer/src/error.rs`.
- `prost` upgrade renamed one function used in `signer/build.rs`.

## Testing Information

For each dependency bump, I did the following to find & ensure no issues:
```
 cargo clean && cargo cache -a
 make build
 make test
 make integration-test-full
 ```
 
 All tests pass both locally and in CI. I also ensured that `devenv` works as expected.
     
## Checklist:

### CODE REVIEW

File: signer/src/error.rs [hunk 1]
```diff
@@ -530,7 +530,7 @@ pub enum Error {
 
     /// Bitcoin error when attempting to construct an address from a
     /// scriptPubKey.
-    #[error("bitcoin address parse error: {0}; txid {}, vout: {}", .1.txid, .1.vout)]
+    #[error("bitcoin address parse error: {0}; txid {txid}, vout: {vout}", txid = .1.txid, vout = .1.vout)]
```
cylewitruk, line 533: Due to `thiserror` upgrade.

signer/src/network/libp2p/event_loop.rs
```diff
@@ -400,6 +400,17 @@ fn handle_gossipsub_event(
         Event::GossipsubNotSupported { peer_id } => {
             tracing::warn!(%peer_id, "peer does not support gossipsub");
         }
+        Event::SlowPeer { peer_id, failed_messages } => {
```
cylewitruk, line 403: Due to `libp2p` upgrade (copied from #1257).

signer/build.rs
```diff
@@ -62,6 +62,6 @@ pub fn compile_protos() {
         .out_dir(workingdir.join("signer/src/proto/generated/"))
         .include_file("mod.rs")
         .type_attribute("crypto.Uint256", "#[derive(Copy)]")
-        .compile(&protos, &[workingdir.join("protobufs")])
+        .compile_protos(&protos, &[workingdir.join("protobufs")])
```
cylewitruk, line 65: Due to `prost` upgrade.

docker/docker-compose.yml
```diff
@@ -218,8 +218,8 @@ services:
       dockerfile: docker/sbtc/Dockerfile
       target: blocklist-client
     environment:
-      BLOCKLIST_CLIENT_RISK_ANALYSIS__API_URL: ${BLOCKLIST_CLIENT_RISK_ANALYSIS__API_URL:-https://public.chainalysis.com}
-      BLOCKLIST_CLIENT_RISK_ANALYSIS__API_KEY: ${BLOCKLIST_CLIENT_RISK_ANALYSIS__API_KEY:-}
+      BLOCKLIST_CLIENT__RISK_ANALYSIS__API_URL: ${BLOCKLIST_CLIENT__RISK_ANALYSIS__API_URL:-https://public.chainalysis.com}
```
cylewitruk, line 221: Ahh, travelling now but I'm going to revert this. I see that the `Environment` needs `.with_prefix_separator("_")` like we do in the signer. Must have defaulted to `_` in the old version.

matteojug: Still to do?

matteojug (approved):

docker/docker-compose.yml
cylewitruk: Updated in [c15b3e8](https://github.com/stacks-network/sbtc/pull/1271/commits/c15b3e821399fcba3e22752a8a8626de1b7e4bc6)

matteojug (approved):

Issue #1106, opened 2024-12-12
[Feature]: Print version and public key on startup


aldur:
## 1. Description

On startup a Stack signer prints its version and public address. 
These would be helpful for tracing for an sBTC signer as well.

## 2. Technical Details:

The signer might _not_ be aware of its version, or the version used for Git tags might not match the one used in Cargo. That will need to change for this to be effective.

### 2.1 Acceptance Criteria:
   - [ ] At startup, the signer prints its public key and version in the logs.

PR #1279, opened 2025-01-28
feat: print signer version and pubkey on startup

MCJOHN974:
## Description

This PR changes signer binary such that it prints signer version and pubkey to logs on startup

Closes: #1106

## Changes

Added few log lines

## Testing Information

I think testing here is not worth the effort, so just tried to run this locally

## Checklist:


signer/src/main.rs
```diff
@@ -74,11 +75,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
         env_abi = signer::TARGET_ENV_ABI,
+        signer_version = signer::VERSION,
         "starting the sBTC signer",
     );
 
     // Load the configuration file and/or environment variables.
     let settings = Settings::new(args.config)?;
+
+    tracing::info!(
+        signer_public_key = %PublicKey::from_private_key(&settings.signer.private_key),
```
cylewitruk, line 86: I think I would just add this to the above startup log properties tbh, we don't really need to know that the config has been loaded successfully because it will exit with an error if it fails.

signer/src/main.rs
```diff
@@ -74,11 +75,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         revision = signer::GIT_COMMIT,
         arch = signer::TARGET_ARCH,
         env_abi = signer::TARGET_ENV_ABI,
+        signer_version = signer::VERSION,
```
matteojug, line 78: I don't think we are actively using the package version (I guess `revision` is already enough)

signer/src/main.rs
matteojug: Having the above log before the config creation makes sense to me -- we are sure we get some logs in case there were bad issues with the config itself.

I'm fine with having this log on a different logline, but I would:
 - add a method in `SignerConfig` returning the public key, and call it here
 - change the message from `constructed signer settings` to something more tailored to the public key

signer/src/main.rs
cylewitruk: Yeah, I think we should probably remove it, only because that "useless info" may be confusing to someone. Right now it will always say `0.0.1` (even though we're at `0.0.9-rc6`).

This is more of a flaw in our build process -- there is tooling to handle version bumps and getting this properly set, but we're not using any of it atm.

So, if/until we've got that implemented, I would leave this out, just to avoid the aforementioned confusion.

matteojug (approved):

PR #1282, opened 2025-01-29
fix: use the same sorting key for in memory impl of stacks chain tip

matteojug:
## Description

Fix a difference between the in memory and pg implementation of `get_stacks_chain_tip`, follow up for the broken test after the `fake` (non) bump in https://github.com/stacks-network/sbtc/pull/1271.

## Changes

Bump `fake` and fix the in memory implementation of `get_stacks_chain_tip` to align it to the pg implementation, by changing the endianness of the stacks block hash used for tie breaking.

## Testing Information

The "broken" test now passes, and add two test to compare the get chain tip between the pg and in memory stores.

## Checklist:


cylewitruk (approved):
Nice catch!

Issue #1176, opened 2024-12-20
[Feature]: Reject many WSTS messages with incorrect bitcoin chain tips


djordon:
# Feature - Reject many WSTS messages with incorrect bitcoin chain tips

## 1. Description

All sBTC messages include the block hash of the bitcoin chain tip. This information is very useful for preventing replay attacks. The signers currently use it when filtering messages from peers where we care about whether the message is from the coordinator or from a signer with the same bitcoin chain tip. We should probably reject all WSTS messages where the chain tip of the sending signer does not match our view of the bitcoin chain tip.

### 1.1 Context & Purpose

We already filter messages based off of who the sender is and who the coordinator is. A consequence of this is that if a new bitcoin block arrives in the middle of a signing round or DKG, the signers will _usually_ ignore the messages from the old coordinator, leading to an unsuccessful signing round or failed DKG. I say _usually_ because the new coordinator could be the old coordinator. But regardless, we should not finish our signing rounds because a new block may mean a reorg, affecting lots of stuff. Similarly, for DKG a signer should not accept messages from peers who have a different view of the bitcoin blockchain than they do.

## 2. Technical Details:

The messages that should be rejected are `DkgBegin`, `DkgEndBegin`, `DkgPrivateBegin`, `NonceRequest`, and `SignatureShareRequest`.

### 2.1 Acceptance Criteria:
   - [ ] We reject WSTS `DkgBegin`, `DkgEndBegin`, `DkgPrivateBegin`, `NonceRequest`, and `SignatureShareRequest` messages if the chain tip does not match our own.

## 3. Related Issues and Pull Requests (optional):

This was largely addressed in https://github.com/stacks-network/sbtc/issues/578.

PR #1284, opened 2025-01-29
feat: reject coordinator wsts messages from non canonical tips

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/1176

## Changes

Add a canonical tip check in the messages coming from the coordinator mentioned in the issue.

## Testing Information

Add test to check we indeed stop early in case of messages from non canonical tips.

## Checklist:

### DIFF
--- a/signer/src/transaction_signer.rs
+++ b/signer/src/transaction_signer.rs
@@ -512,13 +512,11 @@ where
             WstsNetMessage::DkgBegin(_) => {
                 tracing::info!("handling DkgBegin");
 
-                if chain_tip_report.chain_tip_status != ChainTipStatus::Canonical {
-                    tracing::warn!("received wsts message for a different chain tip");
-                    return Ok(());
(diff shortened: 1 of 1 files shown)


signer/src/transaction_signer.rs
```diff
@@ -512,6 +512,11 @@ where
...
                 tracing::info!("handling DkgBegin");
 
+                if chain_tip_report.chain_tip_status != ChainTipStatus::Canonical {
+                    tracing::warn!("received wsts message for a different chain tip");
+                    return Ok(());
+                }
+
                 if !chain_tip_report.sender_is_coordinator {
                     tracing::warn!("received coordinator message from non-coordinator signer");
                     return Ok(());
```
djordon, line 522: We should probably add a function in the `MsgChainTipReport` that implements this logic, since it appears to be the same throughout.

PR #1287, opened 2025-01-30
refactor: wsts-related cleanup

cylewitruk:
## Description

A breakout PR from #1285 to move some of the more refactor-like changes related to WSTS.

Part of: #1300

⚠️ This PR's branch is targeted by #1285 which needs to merge into this branch before merging this PR into `main`.

## Changes

- Rename `WstsMessage.txid` to `id` and use a `WstsMessageId` enum instead of a `bitcoin::Txid`.
  - Protobufs updated in a backwards-compatible way, deprecating the `txid` field (flagging this `breaking-protocol` anyway, just in-case).
  - Generated proto code updated.
- `StateMachineId` is now an enum instead of a newtype over `[u8; 32]`.
- A fair bit of restructuring/cleanup in `wsts_state_machine.rs`. **This is the bulk of the changes in this PR**
- Some logging improvements.
- Updates `wsts` to [this commit](https://github.com/Trust-Machines/wsts/commit/a7bf38bd54cddf0b78c0f7c521a5ed1537a684fa) (`apply-fix` + [this](https://github.com/Trust-Machines/wsts/commit/a7bf38bd54cddf0b78c0f7c521a5ed1537a684fa#diff-53ae23f5f35def420877ccc8c0fe3662e64e38a1)).
  - ... and brings FROST back to life 🥳 

## Testing Information

All existing tests pass and devenv works like a charm.

## Checklist:

### DIFF
--- a/signer/src/message.rs
+++ b/signer/src/message.rs
@@ -228,31 +228,31 @@ pub struct BitcoinPreSignAck;
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum WstsMessageId {
     /// The WSTS message is related to a Bitcoin transaction signing round.
-    BitcoinTxid(bitcoin::Txid),
+    Sweep(bitcoin::Txid),
     /// The WSTS message is related to a rotate key verification operation.
-    RotateKey(PublicKey),
+    DkgVerification(PublicKey),
     /// The WSTS message is related to a DKG round.
     Dkg([u8; 32]),
 }
 
 impl From<bitcoin::Txid> for WstsMessageId {
     fn from(txid: bitcoin::Txid) -> Self {
-        Self::BitcoinTxid(txid)
+        Self::Sweep(txid)
     }
 }
 
 impl From<crate::storage::model::BitcoinTxId> for WstsMessageId {
     fn from(txid: crate::storage::model::BitcoinTxId) -> Self {
-        Self::BitcoinTxid(txid.into())
+        Self::Sweep(txid.into())
     }
 }
 
 impl std::fmt::Display for WstsMessageId {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            WstsMessageId::BitcoinTxid(txid) => write!(f, "bitcoin-txid({})", txid),
-            WstsMessageId::RotateKey(aggregate_key) => {
-                write!(f, "rotate-key({})", aggregate_key)
+            WstsMessageId::Sweep(txid) => write!(f, "sweep({})", txid),
+            WstsMessageId::DkgVerification(aggregate_key) => {
+                write!(f, "dkg-verification({})", aggregate_key)
             }
             WstsMessageId::Dkg(id) => {
                 write!(f, "dkg({})", hex::encode(id))
--- a/signer/src/message.rs
+++ b/signer/src/message.rs
@@ -1,14 +1,18 @@
 //! Signer message definition for network communication
 
+use rand::rngs::OsRng;
+use rand::Rng;
 use secp256k1::ecdsa::RecoverableSignature;
 
 use crate::bitcoin::utxo::Fees;
 use crate::bitcoin::validation::TxRequestIds;
+use crate::keys::PrivateKey;
 use crate::keys::PublicKey;
 use crate::stacks::contracts::ContractCall;
 use crate::stacks::contracts::StacksTx;
 use crate::storage::model;
 use crate::storage::model::BitcoinBlockHash;
+use crate::storage::model::BitcoinTxId;
 use crate::storage::model::StacksTxId;
 
 /// Messages exchanged between signers
@@ -224,14 +228,81 @@ pub struct BitcoinPreSignRequest {
 #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
 pub struct BitcoinPreSignAck;
 
+/// The identifier for a WSTS message.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum WstsMessageId {
+    /// The WSTS message is related to a Bitcoin transaction.
+    BitcoinTxid(bitcoin::Txid),
+    /// The WSTS message is related to a rotate key operation.
+    RotateKey(PublicKey),
+    /// The WSTS message isn't specifically related to anything.
+    Arbitrary([u8; 32]),
+}
+
+impl From<bitcoin::Txid> for WstsMessageId {
+    fn from(txid: bitcoin::Txid) -> Self {
+        Self::BitcoinTxid(txid)
+    }
+}
+
+impl WstsMessageId {
+    /// Generate a random [`WstsMessageId::Arbitrary`] WSTS message ID
+    pub fn random_arbitrary() -> Self {
+        Self::Arbitrary(OsRng.gen())
+    }
+
+    /// Generate a random [`WstsMessageId::BitcoinTxid`] WSTS message ID
+    pub fn random_bitcoin_txid() -> Self {
+        let random_bytes: [u8; 32] = OsRng.gen();
+        let txid = BitcoinTxId::from(random_bytes).into();
+        Self::BitcoinTxid(txid)
+    }
+
+    /// Generate a random [`WstsMessageId::AggregateKey`] WSTS message ID
+    pub fn random_rotate_key() -> Self {
+        let private = PrivateKey::new(&mut OsRng);
+        let public = PublicKey::from_private_key(&private);
+        Self::RotateKey(public)
+    }
+}
+
+impl std::fmt::Display for WstsMessageId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            WstsMessageId::BitcoinTxid(txid) => write!(f, "bitcoin-txid({})", txid),
+            WstsMessageId::RotateKey(aggregate_key) => {
+                write!(f, "rotate-key({})", aggregate_key)
+            }
+            WstsMessageId::Arbitrary(bytes) => write!(f, "arbitrary({})", hex::encode(bytes)),
+        }
+    }
+}
+
 /// A wsts message.
 #[derive(Debug, Clone, PartialEq)]
 pub struct WstsMessage {
-    /// The transaction ID this message relates to,
-    /// will be a dummy ID for DKG messages
-    pub txid: bitcoin::Txid,
+    /// The id of the wsts message.
+    pub id: WstsMessageId,
     /// The wsts message
     pub inner: wsts::net::Message,
+}
+
+impl WstsMessage {
+    /// Returns the type of the message as a &str.
+    pub fn type_id(&self) -> &'static str {
+        match self.inner {
+            wsts::net::Message::DkgBegin(_) => "dkg-begin",
+            wsts::net::Message::DkgEndBegin(_) => "dkg-end-begin",
+            wsts::net::Message::DkgEnd(_) => "dkg-end",
+            wsts::net::Message::DkgPrivateBegin(_) => "dkg-private-begin",
+            wsts::net::Message::DkgPrivateShares(_) => "dkg-private-shares",
+            wsts::net::Message::DkgPublicShares(_) => "dkg-public-shares",
+            wsts::net::Message::NonceRequest(_) => "nonce-request",
+            wsts::net::Message::NonceResponse(_) => "nonce-response",
+            wsts::net::Message::SignatureShareRequest(_) => "signature-share-request",
+            wsts::net::Message::SignatureShareResponse(_) => "signature-share-response",
+        }
+    }
 }
 
 /// Convenient type aliases
--- a/signer/src/message.rs
+++ b/signer/src/message.rs
@@ -231,10 +231,12 @@ pub struct BitcoinPreSignAck;
 /// The identifier for a WSTS message.
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum WstsMessageId {
-    /// The WSTS message is related to a Bitcoin transaction.
+    /// The WSTS message is related to a Bitcoin transaction signing round.
     BitcoinTxid(bitcoin::Txid),
-    /// The WSTS message is related to a rotate key operation.
+    /// The WSTS message is related to a rotate key verification operation.
     RotateKey(PublicKey),
+    /// The WSTS message is related to a DKG round.
+    Dkg([u8; 32]),
 }
 
 impl From<bitcoin::Txid> for WstsMessageId {
@@ -271,6 +273,9 @@ impl std::fmt::Display for WstsMessageId {
             WstsMessageId::BitcoinTxid(txid) => write!(f, "bitcoin-txid({})", txid),
             WstsMessageId::RotateKey(aggregate_key) => {
                 write!(f, "rotate-key({})", aggregate_key)
+            },
+            WstsMessageId::Dkg(id) => {
+                write!(f, "dkg({})", hex::encode(id))
             }
         }
     }
--- a/signer/src/message.rs
+++ b/signer/src/message.rs
@@ -235,8 +235,6 @@ pub enum WstsMessageId {
     BitcoinTxid(bitcoin::Txid),
     /// The WSTS message is related to a rotate key operation.
     RotateKey(PublicKey),
-    /// The WSTS message isn't specifically related to anything.
-    Arbitrary([u8; 32]),
 }
 
 impl From<bitcoin::Txid> for WstsMessageId {
@@ -245,12 +243,13 @@ impl From<bitcoin::Txid> for WstsMessageId {
     }
 }
 
-impl WstsMessageId {
-    /// Generate a random [`WstsMessageId::Arbitrary`] WSTS message ID
-    pub fn random_arbitrary() -> Self {
-        Self::Arbitrary(OsRng.gen())
+impl From<crate::storage::model::BitcoinTxId> for WstsMessageId {
+    fn from(txid: crate::storage::model::BitcoinTxId) -> Self {
+        Self::BitcoinTxid(txid.into())
     }
+}
 
+impl WstsMessageId {
     /// Generate a random [`WstsMessageId::BitcoinTxid`] WSTS message ID
     pub fn random_bitcoin_txid() -> Self {
         let random_bytes: [u8; 32] = OsRng.gen();
@@ -273,7 +272,6 @@ impl std::fmt::Display for WstsMessageId {
             WstsMessageId::RotateKey(aggregate_key) => {
                 write!(f, "rotate-key({})", aggregate_key)
             }
-            WstsMessageId::Arbitrary(bytes) => write!(f, "arbitrary({})", hex::encode(bytes)),
         }
     }
 }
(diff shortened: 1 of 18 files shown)


signer/src/proto/convert.rs
```diff
@@ -1089,15 +1093,30 @@ impl From<WstsMessage> for proto::WstsMessage {
...
+                WstsMessageId::RotateKey(_) => None,
+            },
+            id: Some(match value.id {
+                WstsMessageId::BitcoinTxid(txid) => {
+                    wsts_message::Id::IdBitcoinTxid(proto::BitcoinTxid {
+                        txid: Some(proto::Uint256::from(BitcoinTxId::from(txid).into_bytes())),
+                    })
+                }
+                WstsMessageId::RotateKey(pubkey) => wsts_message::Id::IdRotateKey(pubkey.into()),
+            }),
```
cylewitruk, line 1111: This is for backwards compatibility. When serializing a `message:WstsMessage` to a `proto::WstsMessage`, it will set both the `txid` and the `id` to a `proto::BitcoinTxid` if the source id type is `BitcoinTxid` (the only current case until #1285 merges and DKG runs).

signer/src/proto/convert.rs
```diff
@@ -1131,8 +1150,22 @@ impl TryFrom<proto::WstsMessage> for WstsMessage {
...
+                    wsts_message::Id::IdBitcoinTxid(txid) => {
+                        WstsMessageId::BitcoinTxid(BitcoinTxId::try_from(txid)?.into())
+                    }
+                    wsts_message::Id::IdRotateKey(pubkey) => {
+                        WstsMessageId::RotateKey(PublicKey::try_from(pubkey)?)
+                    }
+                },
+                None => WstsMessageId::BitcoinTxid(
+                    BitcoinTxId::try_from(value.txid.required()?)?.into(),
+                ),
```
cylewitruk, line 1167: This is also for backwards compatibility. When converting from a `proto::WstsMessage` to a `message:WstsMessage`, if the incoming `id` is `None` then it will try to revert to using the `txid` as a `BitcoinTxId`.

This can be removed once #1285 is merged.

signer/src/wsts_state_machine.rs
cylewitruk, line 1: I refactored this a bit to make it a little more ergonomic to work with by introducing the `WstsCoordinator` trait, which in the coordinator lets either FIRE or FROST be used in `coordinate_signing_round()` (#1285), and reducing code a bit.

signer/src/transaction_signer.rs
```diff
@@ -500,23 +501,36 @@ where
     }
 
     /// Process WSTS messages
-    #[tracing::instrument(skip_all, fields(txid = %msg.txid))]
+    #[tracing::instrument(skip_all, fields(
+        msg_id = %msg.id,
+        msg_type = %msg.type_id(),
+        dkg_signer_id = tracing::field::Empty,
+        dkg_id = tracing::field::Empty,
+        dkg_sign_id = tracing::field::Empty,
```
xoloki, line 509: Why call these fields `dkg_sign_id` and `dkg_iter_id`?  They aren't present in DKG messages.  Same with `dkg_signer_id`, though `signer_id` at least is present in many DKG messages but also elsewhere.

signer/src/transaction_signer.rs
```diff
@@ -534,63 +548,90 @@ where
                 if let Some(pause) = self.dkg_begin_pause {
                     // Let's give the others some slack
                     tracing::debug!(
-                        "Sleeping a bit to give the other peers some slack to get DkgBegin"
+                        "sleeping a bit to give the other peers some slack to get dkg-begin"
```
xoloki, line 551: Why change to `dkg-begin`?

xoloki (approved):
Couple of quick questions, but LGTM

signer/src/transaction_signer.rs
cylewitruk: just to stay consistent with keeping things lowercase in logs, `dkgbegin` or `signaturesharerequest` are harder to read

signer/src/transaction_signer.rs
cylewitruk: Good point, will revisit

cylewitruk:
@xoloki is the wsts version/rev correct?

signer/src/transaction_signer.rs
cylewitruk: Updated in [8cdb739](https://github.com/stacks-network/sbtc/pull/1287/commits/8cdb739276f631677f6a1ae4d8b1604ace59a5cd).

I'm curious if `dkg_id` and `sign_id` are conditionally set depending if it's a dkg round vs. signing? If so then if the default is `0` then we could `record()` them only if `> 0` so that only the relevant one is logged.

signer/src/message.rs
```diff
@@ -224,16 +224,70 @@ pub struct BitcoinPreSignRequest {
 #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
 pub struct BitcoinPreSignAck;
 
+/// The identifier for a WSTS message.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum WstsMessageId {
+    /// The WSTS message is related to a Bitcoin transaction signing round.
+    BitcoinTxid(bitcoin::Txid),
+    /// The WSTS message is related to a rotate key verification operation.
+    RotateKey(PublicKey),
```
cylewitruk, line 233: Tempted to rename the whole `RotateKey` thing to `DkgVerify` or something, thoughts?

matteojug: +1 on that, it avoid misunderstandings

signer/src/wsts_state_machine.rs
```diff
@@ -30,33 +39,435 @@ use wsts::traits::Signer as _;
...
+    /// Creates a new coordinator state machine.
+    fn new<I>(signers: I, threshold: u16, message_private_key: PrivateKey) -> Self
+    where
+        I: IntoIterator<Item = PublicKey>;
+
+    /// Gets the coordinator configuration.
+    fn get_config(&self) -> Config;
+
+    /// Save the state required to reconstruct the state machine.
+    fn save(&self) -> SavedState;
```
matteojug, line 130: Why do have/need this? It seems we call it in the coordinator, but we do nothing with the returned state

signer/src/wsts_state_machine.rs
```diff
@@ -30,33 +39,435 @@ use wsts::traits::Signer as _;
...
+    fn from_message(message: &Message) -> Self
+    where
+        Self: Sized;
+}
+
+impl FromMessage for Packet {
+    fn from_message(message: &Message) -> Self {
+        Packet {
+            msg: message.clone(),
+            sig: Vec::new(),
```
matteojug, line 75: this is a bit strange, does it mean we never care about those signatures when processing the packages?

signer/src/wsts_state_machine.rs
```diff
@@ -30,33 +39,435 @@ use wsts::traits::Signer as _;
...
+    fn process_message(
+        &mut self,
+        message: &Message,
+    ) -> Result<(Option<Packet>, Option<OperationResult>), Error> {
+        let packet = Packet::from_message(message);
+        self.process_packet(&packet)
+    }
+
+    /// Process inbound messages
+    fn process_inbound_messages(
```
matteojug, line 163: Is this used around?

signer/src/wsts_state_machine.rs
```diff
@@ -30,33 +39,435 @@ use wsts::traits::Signer as _;
...
+    fn process_packet(
+        &mut self,
+        packet: &Packet,
+    ) -> Result<(Option<Packet>, Option<OperationResult>), Error>;
+
+    /// Process inbound packets
+    fn process_inbound_packets(
+        &mut self,
+        packets: &[Packet],
+    ) -> Result<(Vec<Packet>, Vec<OperationResult>), Error>;
```
matteojug, line 184: This is only used in `process_inbound_messages`, so if that is not used this is also not used

signer/src/wsts_state_machine.rs
```diff
@@ -30,33 +39,435 @@ use wsts::traits::Signer as _;
...
+
+    fn process_message(
+        &mut self,
+        message: &Message,
+    ) -> Result<(Option<Packet>, Option<OperationResult>), Error> {
+        let packet = Packet::from_message(message);
+        self.0
+            .process_message(&packet)
+            .map_err(Error::wsts_coordinator)
+    }
```
matteojug, line 286: Isn't the trait default already enough for this?

signer/src/proto/convert.rs
```diff
@@ -1089,15 +1093,32 @@ impl From<WstsMessage> for proto::WstsMessage {
...
+
         proto::WstsMessage {
-            txid: Some(BitcoinTxId::from(value.txid).into()),
+            txid: match value.id {
+                WstsMessageId::BitcoinTxid(txid) => {
+                    Some(proto::BitcoinTxid::from(BitcoinTxId::from(txid)))
+                }
+                WstsMessageId::Dkg(_) => None,
+                WstsMessageId::RotateKey(_) => None,
+            },
```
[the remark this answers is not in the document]
cylewitruk: Yeah I considered that, in reality we don't need it, I just erred towards practicing ensuring backwards compatibility 🤷

signer/src/proto/convert.rs
```diff
@@ -1792,11 +1831,14 @@ mod tests {
         U: From<T>,
         E: std::fmt::Debug,
     {
-        let original: T = Unit.fake_with_rng(&mut OsRng);
-        let proto_original = U::from(original.clone());
+        // TODO: proptest
+        for _ in 0..10 {
```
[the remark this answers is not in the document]
cylewitruk: Super slow

signer/src/wsts_state_machine.rs
```diff
@@ -30,33 +39,435 @@ use wsts::traits::Signer as _;
...
+            .map_err(Error::wsts_coordinator)
     }
 }
 
-#[cfg(any(test, feature = "testing"))]
-impl StateMachineId {
-    /// Create a new identifier
-    pub fn new(value: [u8; 32]) -> Self {
-        StateMachineId(value)
+impl WstsCoordinator for FrostCoordinator {
```
matteojug, line 330: nit: would it be possible to template this to avoid having two identical impl for frost and fire with only a single change (`frost::Coordinator` vs `fire::Coordinator`)?

signer/src/proto/convert.rs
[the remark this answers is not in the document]
matteojug: Sorry I just saw your comments above. Anyway, yeah I just wanted to avoid having virtually dead code around.

signer/src/transaction_signer.rs
xoloki: `dkg_id` is always present, in dkg or signing rounds.  `sign_id` is only present in signing rounds.  

they are always set if they are present.

signer/src/transaction_signer.rs
xoloki: I like keeping struct names. I wouldn't have expected to change to lowercase with dashes when grepping.

xoloki:
> @xoloki is the wsts version/rev correct?

This is the rev for the head of the apply-fix branch, which has all the hotfixes plus the framework necessary for this PR.

signer/src/proto/convert.rs
```diff
@@ -1089,15 +1093,32 @@ impl From<WstsMessage> for proto::WstsMessage {
...
+                WstsMessageId::BitcoinTxid(txid) => {
+                    Some(proto::BitcoinTxid::from(BitcoinTxId::from(txid)))
+                }
+                WstsMessageId::Dkg(_) => None,
+                WstsMessageId::RotateKey(_) => None,
+            },
+            id: Some(match value.id {
+                WstsMessageId::BitcoinTxid(txid) => {
+                    wsts_message::Id::IdBitcoinTxid(proto::BitcoinTxid {
+                        txid: Some(proto::Uint256::from(BitcoinTxId::from(txid).into_bytes())),
```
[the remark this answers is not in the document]
cylewitruk: Pretty sure I tried that before, but:
```
the trait bound `crypto::Uint256: From<storage::model::BitcoinTxId>` is not satisfied
the trait `From<storage::model::BitcoinTxId>` is not implemented for `crypto::Uint256`
but trait `From<[u8; 32]>` is implemented for it
for that trait implementation, expected `[u8; 32]`, found `storage::model::BitcoinTxId`
```
And I didn't really feel like adding it just for this 🙈

signer/src/transaction_signer.rs
djordon: My vote is on `dkg-begin`, since it's easier to type.

signer/src/proto/convert.rs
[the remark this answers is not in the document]
djordon: I mentioned this in another [review](https://github.com/stacks-network/sbtc/pull/1285#discussion_r1939470460), basically I think we should set this to `None` all the time. This family of changes are not wire-breaking but are protocol breaking, since DKG will not succeed without the new binary. Same thing for signing via WSTS. So might as well do some clean-up while we're here.

signer/src/testing/wsts.rs
```diff
@@ -147,9 +148,10 @@ impl Coordinator {
             .start_public_shares()
             .expect("failed to start public shares");
 
-        self.send_packet(bitcoin_chain_tip, txid, outbound).await;
+        self.send_packet(bitcoin_chain_tip, txid.into(), outbound)
```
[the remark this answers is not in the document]
cylewitruk: lol it was 1 change or 20 ;p but i made the 20 (or so) here: [716e5c2](https://github.com/stacks-network/sbtc/pull/1287/commits/716e5c2e7cf4617c5cbd4b73acaa4d30c85aa0bd)

signer/src/transaction_coordinator.rs
```diff
@@ -914,9 +915,9 @@ where
         transaction: &mut utxo::UnsignedTransaction<'_>,
     ) -> Result<(), Error> {
         let sighashes = transaction.construct_digests()?;
-        let mut coordinator_state_machine = CoordinatorStateMachine::load(
-            &mut self.context.get_storage_mut(),
-            sighashes.signers_aggregate_key,
+        let mut coordinator_state_machine = FireCoordinator::load(
+            &self.context.get_storage_mut(),
```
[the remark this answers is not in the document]
cylewitruk: nice, missed these 👍 [267c5a4](https://github.com/stacks-network/sbtc/pull/1287/commits/267c5a4694864e3918f6696dcbab6cea8183825e)

signer/src/proto/convert.rs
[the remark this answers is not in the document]
cylewitruk: I'm good with it if you are 👍

signer/src/testing/wsts.rs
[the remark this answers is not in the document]
matteojug: lol sorry, but hey it was 2 of them!

signer/src/wsts_state_machine.rs
cylewitruk: Yup, [76b9b49](https://github.com/stacks-network/sbtc/pull/1287/commits/76b9b4971028e57839d23c41324fdc000dae4dba)

signer/src/wsts_state_machine.rs
cylewitruk: Nope, apparently only for the `SignerStateMachine` and then only in tests, so the coordinator's don't use the interator variants. Removed in [cb95417](https://github.com/stacks-network/sbtc/pull/1287/commits/cb9541747dc90d6fc486f3308536ebddc830b112)

signer/src/wsts_state_machine.rs
cylewitruk: Same as above comment, removed in [cb95417](https://github.com/stacks-network/sbtc/pull/1287/commits/cb9541747dc90d6fc486f3308536ebddc830b112)

signer/src/transaction_signer.rs
```diff
@@ -520,6 +537,8 @@ where
                     return Ok(());
                 }
 
+                tracing::debug!("responding to dkg-begin");
```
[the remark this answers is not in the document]
cylewitruk: Just went with `processing` in general, and also removed the hard-coded message name in favor of the `wsts_msg_type` field.

signer/src/transaction_signer.rs
cylewitruk: We'll go with kebab-case for now. If it's a problem then we can revert it.

signer/src/message.rs
cylewitruk: I gave them some new names in [5f99a29](https://github.com/stacks-network/sbtc/pull/1287/commits/5f99a29c7101d87b2a0ab5c44e2c829c6e2d621c) which I think provide more context.

Again, these aren't used specifically in this PR -- control flow based on them is introduced in #1285.

signer/src/proto/convert.rs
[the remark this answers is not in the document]
cylewitruk: Removed in [d22ed61](https://github.com/stacks-network/sbtc/pull/1287/commits/d22ed614c1628045f313fd2d225c23d67a7ef0fa)

signer/src/wsts_state_machine.rs
cylewitruk: Yeah, I reacted on it too. This is just a helper method to avoid doing the `Packet { .. }` elsewhere, where originally calling code would create the `wsts::Packet` from a `wsts::Mesage` in the above style and call `process_packet()`. So the actual logic shouldn't be changed; it should just be a little cleaner from the caller's perspective.

That's also where the new `process_message()` on the `WstsCoordinator` trait came from as well -- the wsts `process_message()` takes a `Packet`. So the new method instead takes a `Message` and does the redirection to the oddly named wsts `process_message(Packet)`.

@xoloki could check and confirm that there's nothing weird here

cylewitruk:
> > @xoloki is the wsts version/rev correct?
> 
> This is the rev for the head of the apply-fix branch, which has all the hotfixes plus the framework necessary for this PR.
> 
> I'm really not sure what version this should be. That branch is still version `10.0.0`, but we changed the `Coordinator` trait in the last two changes, so maybe it needs to be `10.1.0` ?

Yeah we might need a minor release to include that, if we don't want to use the commit hash. I've reviewed the changes so I'm fine with using the commit hash until it's ready, but yeah it would be nice to have a `10.1.0`.

signer/src/proto/convert.rs
```diff
@@ -1058,6 +1061,7 @@ impl TryFrom<proto::SignatureShareResponse> for SignatureShareResponse {
     }
 }
 impl From<WstsMessage> for proto::WstsMessage {
+    #[allow(deprecated)]
```
matteojug, line 1064: nit: still relevant? also below

matteojug (approved):

signer/src/message.rs
cylewitruk: It's not the best, but it, like the others, now say exactly what they're used for (we don't sign any other kinds bitcoin transactions).

This field went away completely in the message id refactor i was playing with this morning, so when we revisit this I'm sure we'll sort it out.

signer/src/proto/convert.rs
cylewitruk: [68bf8f5](https://github.com/stacks-network/sbtc/pull/1287/commits/68bf8f5337f7f810d56466e20075712341e1125a)

PR #1294, opened 2025-01-31
feat: emily get transactions by reclaim pubkey

Jiloc:
## Description

Closes: #1218 

This PR builds on top of #1197. Most of the changes are autogenerated code. 

## Changes
- at deposit creation time, we now add a new field `reclaim_pubkeys_hash` to Dynamo that contains the ordered SHA256 hash of the public keys extracted by the reclaim_script. 
- the reclaim pubkey is extracted by the reclaim_script. Currently, it supports:
    - the reclaim script generated by the [bridge ](https://github.com/stacks-network/sbtc-bridge/blob/7e959c4c4587a563a22003243dfe7e265305de31/src/util/depositRequest.ts#L79)and leather
    - the multisig reclaim script generated by [Asigna](https://github.com/stacks-network/sbtc-bridge/blob/282374de3df95fe62c9e9469a914694dc50e8d30/src/util/depositRequest.ts#L77-L102)
    
- add new `https://{base_url}/deposit/reclaim-pubkeys/{pubkeys}` endpoint. Follows the same approach as #1125 adding a new secondary global index to DynamoDB.
    - `pubkeys` is a dash-separated list of x-only pubkeys, order indepedent. e.g `/deposit/reclaim-pubkeys/5da66963a375a1b994fbf695ddfa161954ffecdf67d80397650dcb4985f6a09c-883a1b3f430eefac5bed7aa0d428e267a558736346363cbfec6b0e321e31f453`
    - the public keys are ordered and hashed to retrieve the linked deposits

        
## Testing Information
- add unit tests and adapt integration tests

## Checklist:


emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -55,6 +67,119 @@ pub struct CreateDepositRequestBody {
...
+                "invalid transaction input".to_string(),
+            )
+        })?;
+        let input_address = {
+            let params = if is_mainnet {
+                Params::MAINNET
+            } else {
+                Params::REGTEST
+            };
+            Address::from_script(&txin.script_sig, params.clone())
```
djordon, line 162: Are we sure that we want to use Address here instead of the `scriptPubKey`? I feel like we should shy away from addresses if possible and just use the `scriptPubKey`. That's what is in the transaction (when it is an output) so it seems most natural to use that and encode it as hex.

djordon: Also, sorry for looking at this too late, but we might not be able to use either the `scriptPubKey` or the "address" because we do not have them (they're never included in the transaction when they're being spent). This `scriptSig` is not the same as the `scriptPubKey` from the transaction output, it's basically an unlocking script that, when paired with the actual `scriptPubKey`, allows you to spend the funds. It's often empty too. For example, all deposit request outputs and all signer UTXOs have empty `scriptSig`s when spent in a sweep transaction. It's basically only used for legacy `scriptPubKey`s like P2PKH. I think that the only option is to abandon the endeavor or use bitcoin core for the information.

Also, since a transaction's first input does not necessarily map to a "definitive" source address for the deposit, we may need to make some other changes here too. For example, we might want to check all input `scriptPubKey`s and using that if they all match. Or maybe map all input `scriptPubKey`s to the same deposit. All of this is assuming we can get the `scriptPubKey` at all.

djordon (commented):
Hmmm, we might need to find a new way here, since I just realized that we do not have the input address/scriptPubKey in the transaction. I'm not sure anything can be done without reaching out to bitcoin-core.

emily/handler/src/api/models/deposit/requests.rs
Jiloc: Man this is very appreciated! A very simple alternative would be to add an address_hex field in the create_deposit request itself. I am discussing it with @hozzjss and @setbern

djordon: > Man this is very appreciated! A very simple alternative would be to add an address_hex field in the create_deposit request itself. I am discussing it with @hozzjss and @setbern

No problem 😄.

One thing to consider is how to verify that information if we include it. I couldn't think of a way without having the input transactions or bitcoin core (or a trusted equivalent).

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -419,4 +480,230 @@ pub async fn update_deposits(
...
+            //  OP_PUSHBYTES_32, pubkeyN, OP_CHECKSIGADD,
+            //  OP_PUSHNUM_N]
+            let mut data_iter = keys_data.iter();
+            let mut pubkeys = Vec::new();
+            while let Some(&opcode) = data_iter.next() {
+                match opcode {
+                    OP_PUSHBYTES_32 => {
+                        if let Ok(pubkey) = <Vec<u8> as TryInto<[u8; 32]>>::try_into(
+                            data_iter.by_ref().take(32).cloned().collect(),
+                        ) {
```
matteojug, line 529: nit: separate the computation of the value from the `if let` for readability?

emily/handler/tests/integration/deposit.rs
```diff
@@ -532,6 +548,127 @@ async fn get_deposits_for_recipient() {
...
+            let response = apis::deposit_api::get_deposits_for_reclaim_pubkeys(
+                &configuration,
+                pubkey,
+                next_token.as_ref().and_then(|o| o.as_deref()),
+                Some(chunksize),
+            )
+            .await
+            .expect("Received an error after making a valid get deposits for pubkey api call.");
+            gotten_deposit_info_chunks.push(response.deposits);
+            next_token = response.next_token;
```
matteojug, line 644: nit: `response.next_token.flatten()` may simplify the `next_token.as_ref().and_then` thingy

matteojug (approved):

emily/handler/tests/integration/deposit.rs
```diff
@@ -556,17 +554,38 @@ async fn get_deposits_for_reclaim_pubkeys() {
...
+        for pubkey in iter {
+            reclaim_user_script = reclaim_user_script
+                .push_slice(pubkey)
+                .push_opcode(opcodes::OP_CHECKSIGADD);
+        }
+
+        let reclaim_user_script = reclaim_user_script
+            .push_int(pubkeys.len() as i64)
+            .push_opcode(opcodes::OP_NUMEQUAL)
             .into_script();
```
matteojug, line 582: nit: doing so we are no longer testing the bridge/leather case though

Jiloc: good catch! I fixed it in https://github.com/stacks-network/sbtc/pull/1294/commits/32372475b00aa0d11cf3dc9b97c2055903b68953. Now, I only append the extra OPs if there is more than 1 key

matteojug: It would be nice to also have a 1 out of 1 test for asigna (like, in the asigna format), but it's not a blocker.

sbtc/src/testing/deposits.rs
```diff
@@ -33,6 +33,7 @@ fn build_deposit_reclaim_outputs(
     max_fee: u64,
     amounts: &[u64],
     recipient: Option<StacksAddress>,
+    reclaim_user_script: Option<&ScriptBuf>,
```
djordon, line 36: I would change this to always require a `ScriptBuf` and have the caller pass `ScriptBuf::new()` instead of `None`.

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -210,7 +215,7 @@ pub async fn get_deposits_for_recipient(
     recipient: String,
     query: BasicPaginationQuery,
 ) -> impl warp::reply::Reply {
-    debug!("In get deposits for recipient");
+    debug!("in get deposits for recipient: {recipient}");
```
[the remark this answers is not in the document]
Jiloc: we expect them in hex, but we don't validate anywhere before logging

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -419,4 +487,231 @@ pub async fn update_deposits(
...
+                    OP_PUSHBYTES_32 => {
+                        // Collect the next 32 bytes
+                        let pubkey_bytes: Vec<u8> = data_iter.by_ref().take(32).cloned().collect();
+                        let pubkey_result: Result<[u8; 32], _> = pubkey_bytes.try_into();
+
+                        match pubkey_result {
+                            Ok(pubkey) => pubkeys.push(pubkey),
+                            Err(_) => return None, // Malformed pubkey
+                        }
+                    }
```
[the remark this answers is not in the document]
Jiloc: Thanks, and good to know, I didn't look into that!

emily/handler/src/api/handlers/deposit.rs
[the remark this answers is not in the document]
djordon: 👍🏽 gotcha

sbtc/src/testing/deposits.rs
Jiloc: modified in https://github.com/stacks-network/sbtc/pull/1294/commits/2b6ab9d454dede73d2e341fd932aeec28e6a7589, but I still kept it as a reference because internally is still forcefully cloned

djordon: > still kept it as a reference because internally is still forcefully cloned

Oh, usually that's a reason to have it take ownership since you could potentially forgo a clone -- perhaps the caller has an owned thing to pass, in which case there would be no cloning and probably more performant code. This is test code, so it doesn't matter either way, I'm not pressing for a change.

On the flip side, sometimes it's more ergonomic to have a function take a reference when it really needs the owned thing, so not a hard and fast rule to take ownership to forgo the clone either.

djordon (approved):
Looks good 🚀

emily/handler/src/database/accessors.rs
```diff
@@ -96,6 +98,23 @@ pub async fn get_deposit_entries_by_recipient(
     .await
 }
 
+/// Get deposit entries by reclaim pubkey.
+#[allow(clippy::ptr_arg)]
```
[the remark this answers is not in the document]
Jiloc: not really, I think it was a carry-over of a copy and paste of the function definition. I removed it from a few other functions in the same file as well 👍

Issue #1295, opened 2025-01-31
[Feature]: Cache key-rotation data when a new Bitcoin block is observed


djordon:
# Feature - Cache key-rotation data when a new Bitcoin block is observed

### 1.1 Context & Purpose
We have an issue where collective operations take too long because of the time it takes to fetch information from the database. We need to speed it up.

Right now each signer fetches a few pieces of information whenever they receive a message from a signer peer. This information includes:
1. The canonical bitcoin chain tip.
2. The current signing set from the last rotate-keys transaction on Stacks (with a fallback).
3. The current aggregate key, also from the last rotate-keys transaction on Stacks (with a fallback).

The canonical chain tip changes whenever we observe a bitcoin block, so we can cache that after the block observer has finished updating the database. The current signer set and the current aggregate key transactions are also guaranteed to be in the database after the block observer has finished, but the event might appear some time later. Still, we should just cache what we know after the block observer has updated the database.

The error case here is that the event for the transaction is sent after we've fetched the block confirming the transaction that generated the event.

## 2. Technical Details:

At the end of the block observers duties, fetch and store the last key rotation transaction and store it in the `SignerState` object. ~~Do the same for the bitcoin chain tip.~~

### 2.1 Acceptance Criteria:
   - [ ] We attempt to read the last key-rotation transaction from the database at most once per bitcoin block.

## 3. Related Issues and Pull Requests (optional):

PR #1298, opened 2025-02-03
feat: cache key-rotation data block observer

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1295.

## Changes

1. Store the current signer set and aggregate key in the signer state at the end of the block observer.
2. Change the event loops to use the signer state when they need the current signer set or current aggregate key.
3. Changes in the current signer set now affect who the signers can communicate to. This is fine for now, since we do not support changing the current signing set from the bootstrap signer set.
4. Clean up the creation of a block hash stream, particularly in tests.

## Testing Information

This PR adds a new integration test for the expected behavior of the block observer. It also moves around another test, and removes a test that was duplicated. The existing integration tests should catch anything else that might be amiss, but I also tested this locally in our devenv and everything was fine, where I was able to smoothly mint sbtc.

## Checklist:

### DIFF
--- a/signer/src/bitcoin/zmq.rs
+++ b/signer/src/bitcoin/zmq.rs
@@ -50,7 +50,7 @@ pub struct BitcoinCoreMessageStream {
 }
 
 impl BitcoinCoreMessageStream {
-    /// Create a new one using the endpoint(s) in the config.
+    /// Create a new one using the given endpoint.
     pub async fn new_from_endpoint(endpoint: &str) -> Result<Self, Error> {
         let inner_stream = tokio::time::timeout(Duration::from_secs(10), async {
             bitcoincore_zmq::subscribe_async_monitor(&[endpoint])
--- a/signer/src/bitcoin/zmq.rs
+++ b/signer/src/bitcoin/zmq.rs
@@ -51,10 +51,7 @@ pub struct BitcoinCoreMessageStream {
 
 impl BitcoinCoreMessageStream {
     /// Create a new one using the endpoint(s) in the config.
-    pub async fn new_from_endpoint<T>(endpoint: &str, _subscriptions: &[T]) -> Result<Self, Error>
-    where
-        T: AsRef<str>,
-    {
+    pub async fn new_from_endpoint(endpoint: &str) -> Result<Self, Error> {
         let inner_stream = tokio::time::timeout(Duration::from_secs(10), async {
             bitcoincore_zmq::subscribe_async_monitor(&[endpoint])
         })
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -588,7 +588,9 @@ impl<C: Context, B> BlockObserver<C, B> {
         tracing::info!("Updating the signer state with the current signer set");
         let (aggregate_key, public_keys) = self.get_signer_set_and_aggregate_key(chain_tip).await?;
         if let Some(aggregate_key) = aggregate_key {
-            self.context.state().set_current_aggregate_key(aggregate_key);
+            self.context
+                .state()
+                .set_current_aggregate_key(aggregate_key);
         }
 
         self.context.state().update_current_signer_set(public_keys);
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -581,12 +581,10 @@ impl<C: Context, B> BlockObserver<C, B> {
         Ok(())
     }
 
-    async fn update_signer_state(&self, chain_tip: BlockHash) -> Result<(), Error> {
-        tracing::info!("loading sbtc limits from Emily");
-        self.update_sbtc_limits().await?;
+    async fn set_signer_set_and_aggregate_key(&self, chain_tip: BlockHash) -> Result<(), Error> {
+        let (aggregate_key, public_keys) =
+            get_signer_set_and_aggregate_key(&self.context, chain_tip).await?;
 
-        tracing::info!("Updating the signer state with the current signer set");
-        let (aggregate_key, public_keys) = self.get_signer_set_and_aggregate_key(chain_tip).await?;
         if let Some(aggregate_key) = aggregate_key {
             self.context
                 .state()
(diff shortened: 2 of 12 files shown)


signer/tests/integration/block_observer.rs
```diff
@@ -906,3 +851,324 @@ async fn next_headers_to_process_ignores_known_headers() {
...
+    assert_eq!(rotate_keys_signer_set, signer_set);
+
+    testing::storage::drop_db(db).await;
+}
+
+/// This test checks that the signer state is updated with the latest the
+/// sbtc limits, current signer set, and current aggregate key after the
+/// block observer processes a bitcoin block.
+#[tokio::test]
+async fn block_observer_updates_state_after_observing_bitcoin_block() {
```
cylewitruk, line 957: Not suggesting to change it, but man it feels like a lot of work for testing this 🤔 When we get some time we should look into how we can break these kinds of tests into smaller units...

cylewitruk (approved):
Looks good in general, just left a few comments, do with them as you will :)

signer/src/bitcoin/zmq.rs
```diff
@@ -51,10 +51,7 @@ pub struct BitcoinCoreMessageStream {
 
 impl BitcoinCoreMessageStream {
     /// Create a new one using the endpoint(s) in the config.
-    pub async fn new_from_endpoint<T>(endpoint: &str, _subscriptions: &[T]) -> Result<Self, Error>
-    where
-        T: AsRef<str>,
-    {
+    pub async fn new_from_endpoint(endpoint: &str) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
djordon: Oh I'll check and update accordingly.

signer/tests/integration/block_observer.rs
```diff
@@ -906,3 +851,324 @@ async fn next_headers_to_process_ignores_known_headers() {
...
+    let mut ctx = TestContext::builder()
+        .with_storage(db.clone())
+        .with_first_bitcoin_core_client()
+        .with_mocked_emily_client()
+        .with_mocked_stacks_client()
+        .build();
+
+    // We need to set up the stacks client as well. We use it to fetch
+    // information about the Stacks blockchain, so we need to prep it, even
+    // though it isn't necessary for our test.
```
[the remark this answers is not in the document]
djordon: I was thinking that we just have a function that does this. But I don't want to add it here. Probably in my next PR where I'll need to implement these mocks more often 😞.

signer/src/bitcoin/zmq.rs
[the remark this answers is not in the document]
djordon: So `url::Url` does implement `AsRef`, but after making the change locally, it seems kinda unnecessary. We only use this function in two places, one in `main.rs` and another in the test helper function. I try to shy away from generics so I feel like keeping it as-is is the right move. Maybe I'll feel different in the morning though.

signer/src/block_observer.rs
```diff
@@ -157,8 +159,9 @@ where
                         tracing::warn!(%error, "could not process stacks blocks");
                     }
 
-                    if let Err(error) = self.update_sbtc_limits().await {
-                        tracing::warn!(%error, "could not update sBTC limits");
+                    tracing::debug!("Updating the signer state");
```
matteojug, line 162: nit:
```suggestion
                    tracing::debug!("updating the signer state");
```

signer/src/block_observer.rs
```diff
@@ -577,6 +580,101 @@ impl<C: Context, B> BlockObserver<C, B> {
...
+    ///
+    /// The function updates the following:
+    /// * sBTC limits from Emily.
+    /// * The current signer set.
+    /// * The current aggregate key.
+    async fn update_signer_state(&self, chain_tip: BlockHash) -> Result<(), Error> {
+        tracing::info!("loading sbtc limits from Emily");
+        self.update_sbtc_limits().await?;
+
+        tracing::info!("Updating the signer state with the current signer set");
```
matteojug, line 628: nit
```suggestion
        tracing::info!("updating the signer state with the current signer set");
```

signer/src/transaction_coordinator.rs
```diff
@@ -310,15 +309,7 @@ where
...
-            let dkg_result = self.coordinate_dkg(&bitcoin_chain_tip).await?;
-            // TODO: in `run_dkg_from_scratch` test, `dkg_result` differs from
-            // value fetched from the db. Adding a temporary fix for the (probably)
-            // race condition, but we should address this properly.
-            self.get_signer_set_and_aggregate_key(&bitcoin_chain_tip)
-                .await
-                .ok()
-                .and_then(|res| res.0)
-                .unwrap_or(dkg_result)
+            self.coordinate_dkg(&bitcoin_chain_tip).await?
```
matteojug, line 312: Assuming this goes well, shouldn't we refresh the state? It should have a new aggregate key now

signer/tests/integration/block_observer.rs
```diff
@@ -906,3 +851,324 @@ async fn next_headers_to_process_ignores_known_headers() {
 
     testing::storage::drop_db(db).await;
 }
+
+/// The [`TxCoordinatorEventLoop::get_signer_set_and_aggregate_key`]
```
matteojug, line 855: nit: is no longer the `TxCoordinatorEventLoop`

signer/tests/integration/block_observer.rs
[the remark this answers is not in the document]
cylewitruk: Lol yeah, was just a random comment, didn't mean that you should do anything here :)

signer/src/bitcoin/zmq.rs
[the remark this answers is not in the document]
cylewitruk: Up to you, was just a suggestion, no strong opinions here one way or the other :)

signer/src/transaction_coordinator.rs
cylewitruk: Do we want to do that before we see a confirmed rotate-keys event? (which would then be at a new bitcoin block and thus the state be updated)

matteojug: I guess this is only relevant on the first DKG, where after `coordinate_dkg` we would have an aggregate key while before we wouldn't, so going from picking the bootstrapping set to the latest dkg entry.

matteojug: Only applies to bootstrapping, and updating here wouldn't be enough because also the signers would need to refresh the state. The extra complexity doesn't seem to be worthy, resolving -- worst case, we are gonna "pay" this with an extra block latency before submitting the rotate tx

matteojug (approved):

Issue #1299, opened 2025-02-03
[Feature]: More efficient validation of complete-deposit contract calls


djordon:
# Feature - More efficient validation of complete-deposit contract calls

### 1.1 Context & Purpose

One of our slowest queries is the one behind `get_swept_deposit_requests`. The coordinator uses this query to figure out which deposit requests need a corresponding mint, and the other signers use this query during validation of the `complete-deposit` contract call. It would be more performant for the signers to use a more specialized query along with a request to their stacks node when validating the contract call. This approach has the benefit of allowing signers to miss the `completed-deposit` event from the stacks node for validation of contract call transactions.

## 2. Technical Details:

We'll need to add another method to the `StacksInteract` trait that fetches the "status" of a deposit request from the sbtc-registry by making a call to the `get-deposit-status` read-only function. Then we can use this during validation of the stacks transaction. We would also need to use the `get_deposit_request_report` query during validation, which should be much faster since it does not touch the stacks blockchain data.

### 2.1 Acceptance Criteria:
   - [ ] We do not use slow queries during validation of complete-deposit contract calls.

## 3. Related Issues and Pull Requests (optional):

This is tied to some other performance improvement tickets, like https://github.com/stacks-network/sbtc/issues/1295.

PR #1306, opened 2025-02-06
feat: add function for using smart contract for deposit status

djordon:
## Description

This is part of https://github.com/stacks-network/sbtc/issues/1299 but doesn't complete it. The code to actually complete it is here b1125ddf8d7e417050ff1160a12ca4d93c355ec0 (which I reverted). We'd need to fix a few tests with lots of mocks, which is better to do in a separate PR.

## Changes

1. Add a function for fetching the deposit status from the sbtc-registry smart contract.
2. Add some integration tests to actually test the function.

## Testing Information

I set up devenv and ran the new integration test and it worked as expected. But I needed to update devenv a little in order for it to work. Specifically, I needed to update devenv with the versions listed in https://github.com/stacks-network/sbtc/pull/1285. I'll PR that change separately.

## Checklist:


signer/src/stacks/api.rs
```diff
@@ -505,9 +519,18 @@ impl StacksClient {
...
+        let arguments = arguments
+            .iter()
+            .map(|value| value.serialize_to_hex())
+            .collect::<Result<Vec<String>, _>>()
+            .map_err(Error::ClarityValueSerialization)?;
+
         let body = CallReadRequest {
             sender: sender.to_string(),
-            arguments: vec![], // TODO: Add when needed
+            arguments,
```
cylewitruk, line 533: **Suggestion:** Akin to the comment on its cousin withdrawal PR, I might consider adding proper support for handling `404` vs. `error` while anyways in here.

EDIT: But otoh I might actually not do that here as I noticed that it would affect the `SbtcLimits` stuff as well.

signer/src/stacks/api.rs
```diff
@@ -1138,6 +1161,41 @@ impl StacksInteract for StacksClient {
...
 
+    async fn is_deposit_completed(
+        &self,
+        deployer: &StacksAddress,
+        outpoint: &OutPoint,
+    ) -> Result<bool, Error> {
+        // Both ContractName::from and ClarityName::from can panic when
+        // given the "wrong" strings. These particular strings do not
+        // panic, and we test this fact in our unit tests.
+        let contract_name = ContractName::from("sbtc-registry");
```
cylewitruk, line 1172: **Suggestion:** Akin to its cousin PR, I'd love if we started moving towards `const`s for these.

cylewitruk (approved):
Lgtm, just a couple thoughts :+1:

signer/src/stacks/api.rs
matteojug: We can do `ContractName::from(SmartContract::SbtcToken.contract_name())`

matteojug (approved):

signer/src/stacks/api.rs
djordon: > **Suggestion:** Akin to its cousin PR, I'd love if we started moving towards `const`s for these.

> We can do `ContractName::from(SmartContract::SbtcRegistry.contract_name())`

Oh good point. And I had forgotten that we had those constants 😅 already. I'll add some more for the other constants.

djordon: Updated in https://github.com/stacks-network/sbtc/pull/1306/commits/7c8c9dbc5c0af6d869255e4c133249cce41f14d2

PR #1307, opened 2025-02-06
chore: update devenv with less building

djordon:
## Description

Rather than building containers in devenv, we should strive to use the official containers where possible.

## Changes

* Use the bare containers for bitcoin, stacks-core, stacks-signer, stacks-api, and the stacks-explorer.
* Bump the version of the stacks containers. This fixes an issue that I discovered during the testing of https://github.com/stacks-network/sbtc/pull/1306.
* Delete the now-unused docker files for the above services.
* Remove the protoc action, we do not need it, and CI randomly fails when trying to start the action.

Installing `gettext-base` at the beginning of the stacks-core and stacks-signer was surprisingly fast, so I'm okay keeping it there. My original intent was to get things working and then make it better but now I think it's simpler to keep it where it's at.

## Testing Information

I checked that devenv worked after these changes, and it did.

## Checklist:


aldur:
I guess since you are at it you can also delete `https://github.com/stacks-network/sbtc/blob/main/DOCKER.md`.

docker/docker-compose.yml
```diff
@@ -362,11 +358,13 @@ services:
...
       - -c
       - |
+        apt-get update && apt-get install -y gettext-base
+        cd /data/
         set -e
         envsubst < config.toml.in > config.toml
-        bitcoin-cli -rpcwait -rpcconnect=bitcoin getmininginfo
         exec stacks-node start --config config.toml
     profiles:
+        # bitcoin-cli -rpcwait -rpcconnect=bitcoin getmininginfo
```
Jiloc, line 367: nit: can this be removed?

docker/docker-compose.yml
```diff
@@ -471,14 +469,9 @@ services:
       - default
 
   stacks-api:
-    image: stacks-api:latest
+    image: blockstack/stacks-blockchain-api:8.5.0
```
Jiloc, line 472: nit: there is version 8.6.0 available

docker/docker-compose.yml
```diff
@@ -324,10 +323,7 @@ services:
   # Stacks Blockchain.
   # ------------------
   stacks-node:
-    build:
-      context: stacks
-      args:
-        GIT_COMMIT: *STACKS_BLOCKCHAIN_COMMIT
+    image: blockstack/stacks-core:3.1.0.0.5
```
matteojug, line 326: Should we also remove `docker/stacks/Dockerfile`?

Jiloc (commented):
A few nits. apart from the `STACKS_BLOCKCHAIN_COMMIT` question, feel free to risolve and ignore the rest

docker/docker-compose.yml
```diff
@@ -362,11 +358,13 @@ services:
       - /bin/bash
       - -c
       - |
+        apt-get update && apt-get install -y gettext-base
+        cd /data/
         set -e
         envsubst < config.toml.in > config.toml
-        bitcoin-cli -rpcwait -rpcconnect=bitcoin getmininginfo
```
matteojug, line 367: IIUC this was used to ensure the bitcoin node was in a happy state. Is the stacks node going to complain if it goes up when the bitcoin node is still warming up?

docker/docker-compose.yml
```diff
@@ -58,10 +58,7 @@ x-sbtc-signer: &sbtc-signer
     - sbtc-signer
 
 x-stacks-signer: &stacks-signer
-  build:
-    context: stacks
-    args:
-      GIT_COMMIT: *STACKS_BLOCKCHAIN_COMMIT
+  image: blockstack/stacks-core:3.1.0.0.5
```
matteojug, line 61: nit: we could repurpose `STACKS_BLOCKCHAIN_COMMIT` to ensure we use the same version for the node and the signers.

docker/docker-compose.yml
```diff
@@ -324,10 +323,7 @@ services:
   # Stacks Blockchain.
   # ------------------
   stacks-node:
-    build:
-      context: stacks
-      args:
-        GIT_COMMIT: *STACKS_BLOCKCHAIN_COMMIT
```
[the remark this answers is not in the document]
djordon: Oh good point, I think we should remove it.

docker/docker-compose.yml
djordon: I think so, I commented it out to test that things work when it is removed and just forgot to remove it entirely after I saw that things worked.

docker/docker-compose.yml
djordon: 👍🏽 will update

docker/docker-compose.yml
```diff
@@ -506,14 +499,9 @@ services:
       - default
 
   stacks-explorer:
-    image: stacks-explorer
+    image: hirosystems/explorer:1.245.0
```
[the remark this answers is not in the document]
djordon: 👍🏽 will update

docker/docker-compose.yml
djordon: Oh yeah, it might. We can add in a docker compose `depends_on` to make sure that bitcoin core is healthy before starting the stacks node.

docker/docker-compose.yml
djordon: Ahhh, kind of as a global tag. I think doing something like this supported, I'll try it out.

djordon: Fixed in https://github.com/stacks-network/sbtc/pull/1307/commits/d83e29c25b2d880e6e7af07e49651981d772156e

docker/docker-compose.yml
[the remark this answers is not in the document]
djordon: Done in https://github.com/stacks-network/sbtc/pull/1307/commits/c0a00dd25446b390a02d9cece6a19f6e850de67b

docker/docker-compose.yml
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1307/commits/e1cf22a00c37e7dd5abe3b4bf43049c5a60d0893

docker/docker-compose.yml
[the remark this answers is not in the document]
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1307/commits/e1cf22a00c37e7dd5abe3b4bf43049c5a60d0893

docker/docker-compose.yml
djordon: Removed in https://github.com/stacks-network/sbtc/pull/1307/commits/d83e29c25b2d880e6e7af07e49651981d772156e

docker/docker-compose.yml
```diff
@@ -11,7 +10,7 @@ x-common-vars:
   - &BITCOIN_RPC_USER devnet
   - &BITCOIN_RPC_PASS devnet
   - &MINE_INTERVAL ${MINE_INTERVAL:-1s}
-  - &MINE_INTERVAL_EPOCH25 ${MINE_INTERVAL_EPOCH25:-5s} # 5 second bitcoin block times in epoch 2.5
+  - &MINE_INTERVAL_EPOCH25 ${MINE_INTERVAL_EPOCH25:-1s} # 5 second bitcoin block times in epoch 2.5
```
Jiloc, line 13:
```suggestion
  - &MINE_INTERVAL_EPOCH25 ${MINE_INTERVAL_EPOCH25:-1s} # 1 second bitcoin block times in epoch 2.5
```

Jiloc (approved):

docker/docker-compose.yml
djordon: 🙏🏽 fixed in https://github.com/stacks-network/sbtc/pull/1307/commits/a66f97eddf557b8247ec868cfb0fd33bac1503f6

docker/docker-compose.yml
djordon: I couldn't repurpose that variable since I don't think docker compose supports that kind of thing. It also might be a yaml limitation. I went with a common stanza instead, where we only need to update a single place when updating the blockchain.

docker/docker-compose.yml
```diff
@@ -73,11 +74,11 @@ x-stacks-signer: &stacks-signer
     - /bin/bash
     - -c
     - |
+      apt-get update && apt-get install -y gettext-base
+      cd /data/
       set -e
       envsubst < config.toml.in > config.toml
       exec stacks-signer run --config config.toml
-  profiles:
-    - default
```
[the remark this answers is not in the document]
djordon: I should inherit it from the variable stanza thingy, this guy: https://github.com/stacks-network/sbtc/pull/1307/files#diff-423deb13b7c401b1a7f41ee91c77f722e11d2f317d6a66b546524e8a04cc8b03R34-R37

matteojug (approved):

Issue #1288, opened 2025-01-30
[Chore]: Log startup errors


matteojug:
# Chore - Log startup errors

## 1. Description

The errors returned from `main` are not logged to the usual logs, eg if we fail to parse the config (missing some required fields):
https://github.com/stacks-network/sbtc/blob/67778db4399a32f75c98cb930988a42a9f128c79/signer/src/main.rs#L81

We should log whatever error we get from main before exiting.

## 2. Why It's Necessary

Without it, we just see no more logs, which is not really nice.

PR #1311, opened 2025-02-06
[chore] logging startup erros

MCJOHN974:
## Description

Closes: #1288

## Changes

Added logging of `Settings::new` errors.

## Testing Information

Runned it manually. Don't think some tests for this should be added

## Checklist:

### DIFF
--- a/signer/src/main.rs
+++ b/signer/src/main.rs
@@ -78,26 +78,23 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     );
 
     // Load the configuration file and/or environment variables.
-    let settings = Settings::new(args.config).unwrap_or_else(|error| {
+    let settings = Settings::new(args.config).inspect_err(|error| {
         tracing::error!(%error, "failed to construct the configuration");
-        std::process::exit(1);
-    });
+    })?;
     signer::metrics::setup_metrics(settings.signer.prometheus_exporter_endpoint);
 
     // Open a connection to the signer db.
     let db = PgStore::connect(settings.signer.db_endpoint.as_str())
         .await
-        .unwrap_or_else(|err| {
+        .inspect_err(|err| {
             tracing::error!(%err, "failed to connect to the database");
-            std::process::exit(1);
-        });
+        })?;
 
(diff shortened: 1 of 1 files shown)


matteojug (commented):
We also want to logs also other startup errors (db connection, migrations, ...), maybe it's easier to rename the current `main` and create a slim `main` that calls this and log in case of errors. Or add the missing logs on anything that can fail.

signer/src/main.rs
```diff
@@ -78,15 +78,26 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     );
 
     // Load the configuration file and/or environment variables.
-    let settings = Settings::new(args.config)?;
+    let settings = Settings::new(args.config).unwrap_or_else(|error| {
+        tracing::error!(%error, "failed to construct the configuration");
+        std::process::exit(1);
+    });
```
matteojug, line 84: nit: we can use `inspect_err` + `?` so that we still get the usual output on stdout regardless of the logs
```suggestion
    let settings = Settings::new(args.config).inspect_err(|error| {
        tracing::error!(%error, "failed to construct the configuration");
    })?;
```

matteojug (approved):

signer/src/main.rs
cylewitruk: We shouldn't need to use `process::exit()` here as `main()` returns a `Result<>` -- using the `?` coercion is enough.

signer/src/main.rs
```diff
@@ -78,15 +78,26 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            tracing::error!(%err, "failed to connect to the database");
+            std::process::exit(1);
+        });
 
     // Apply any pending migrations if automatic migrations are enabled.
     if args.migrate_db {
-        db.apply_migrations().await?;
+        db.apply_migrations().await.unwrap_or_else(|err| {
+            tracing::error!(%err, "failed to apply database migrations");
+            std::process::exit(1);
```
cylewitruk, line 99: Same with these, these also just use `.await.inspect_err(..)?;`

PR #1315, opened 2025-02-06
feat/modify rotate key to ask for all signatures

Jiloc:
## Description

Make sure that rotate keys is signed by all signers.

## Changes
- modify `estimate_fees` to accept a new parameter `require_all_signatures` that forces the estimate to expect all signers signatures.
- modify coordinator `sign_stacks_transaction` to wait for all signatures in case of a RotateKeyTx.

## Testing Information
- extended unit tests for `estimate_fees` 

## Checklist:

### DIFF
--- a/signer/src/stacks/contracts.rs
+++ b/signer/src/stacks/contracts.rs
@@ -203,6 +203,14 @@ pub enum StacksTx {
     SmartContract(SmartContract),
 }
 
+impl StacksTx {
+    /// Returns true if the transaction is a rotate keys transaction, regardless
+    /// of the version.
+    pub fn is_rotate_keys(&self) -> bool {
+        matches!(self, StacksTx::ContractCall(ContractCall::RotateKeysV1(_)))
+    }
+}
+
 impl AsTxPayload for StacksTx {
     fn tx_payload(&self) -> TransactionPayload {
         match self {
--- a/signer/src/stacks/wallet.rs
+++ b/signer/src/stacks/wallet.rs
@@ -402,7 +402,11 @@ impl MultisigTx {
 /// given payload.
 ///
 /// This function is very unlikely to fail in practice.
-pub fn get_full_tx_size<T>(payload: &T, wallet: &SignerWallet) -> Result<u64, Error>
+pub fn get_full_tx_size<T>(
+    payload: &T,
+    wallet: &SignerWallet,
+    require_all_signatures: bool,
+) -> Result<u64, Error>
 where
     T: AsTxPayload,
 {
@@ -432,11 +436,14 @@ where
         0,
     )?;
 
+    let sigantures_required = if require_all_signatures {
+        wallet.num_signers()
+    } else {
+        wallet.signatures_required
+    } as usize;
+
     let mut multisig_tx = MultisigTx::new_tx(payload, &wallet, 0);
-    for private_key in private_keys
-        .iter()
-        .take(wallet.signatures_required as usize)
-    {
+    for private_key in private_keys.iter().take(sigantures_required) {
         let signature = crate::signature::sign_stacks_tx(multisig_tx.tx(), private_key);
         // This won't fail, since this is a proper signature
         multisig_tx.add_signature(signature)?;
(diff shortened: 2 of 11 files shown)


signer/src/stacks/wallet.rs
```diff
@@ -402,7 +402,11 @@ impl MultisigTx {
 /// given payload.
 ///
 /// This function is very unlikely to fail in practice.
-pub fn get_full_tx_size<T>(payload: &T, wallet: &SignerWallet) -> Result<u64, Error>
+pub fn get_full_tx_size<T>(
+    payload: &T,
+    wallet: &SignerWallet,
+    require_all_signatures: bool,
```
djordon, line 408: I would say just always require all signatures. It's just simpler.

matteojug: IIRC yesterday we were talking about waiting for everybody (only on rotate key) but then adding only the signatures to arrive to the required threshold, to avoid larger transactions and having to change the fee estimate.

Jiloc: Changed in https://github.com/stacks-network/sbtc/pull/1315/commits/d16bf2ae0dcd21ce5bf8a4db3290c549ac7d0a49

signer/src/transaction_coordinator.rs
```diff
@@ -900,6 +901,22 @@ where
...
+        //
+        // Note: this check is naive and assumes that signers are correctly
+        // validating the new aggregate key. A malicious/faulty signer can
+        // still sign the transaction as the signer simply signs the transaction
+        // using their configured private key.
+        let signatures_required = if req.contract_tx.is_rotate_keys() {
+            wallet.signatures_required()
+        } else {
+            wallet.num_signers()
+        } as usize;
```
matteojug, line 918: Shouldn't be the opposite?

signer/src/transaction_coordinator.rs
```diff
@@ -913,7 +930,9 @@ where
         tokio::pin!(signal_stream);
 
         let future = async {
-            while multi_tx.num_signatures() < wallet.signatures_required() {
+            let mut pending_signers = wallet.public_keys().clone();
+
+            while wallet.public_keys().len() - pending_signers.len() < signatures_required {
```
matteojug, line 935: nit
```suggestion
            while wallet.num_signers() - pending_signers.len() < signatures_required {
```

matteojug, line 935: nit: I would add a comment here explaining that this is a "super-condition" wrt `multi_tx.num_signatures() < wallet.signatures_required()`: we remove keys from the pending list (so we only consider the known ones), and we use `recover_ecdsa` on `multi_tx.digest()` so you cannot play as someone else (unless you can sign the correct digest).

signer/src/transaction_coordinator.rs
Jiloc: thanks 🙏 fixed in https://github.com/stacks-network/sbtc/pull/1315/commits/b4a5d72c0112c7e92b0412f86facbb8392bf39ee

signer/src/transaction_coordinator.rs
Jiloc: I let it as it is, just because the `num_signers` would have required a cast from u16 to usize

matteojug (approved):

Jiloc:
> Looks good. If you have the time, I would add a test that an honest coordinator would not submit a rotate-keys contract call transaction if one of the signers does not send a signature.

I added a test, but I don't like how I did it :( Open to suggestions if you have a better way in mind. Expecially to avoid exposing the `config_mut()` to force a validation to fail on one of the signers

matteojug (changes requested):
There may be an issue with this feature (blocking merge pending confirmation)

djordon:
@Jiloc I think we can close this. https://github.com/stacks-network/sbtc/pull/1315 is built off of this branch but fixes the issue that weird issue that can happen if a signer is starting from scratch.

signer/src/stacks/contracts.rs
```diff
@@ -203,6 +203,14 @@ pub enum StacksTx {
     SmartContract(SmartContract),
 }
 
+impl StacksTx {
+    /// Returns true if the transaction is a rotate keys transaction, regardless
+    /// of the version.
+    pub fn is_rotate_keys(&self) -> bool {
+        matches!(self, StacksTx::ContractCall(ContractCall::RotateKeysV1(_)))
+    }
+}
```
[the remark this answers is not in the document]
Jiloc: you are right, could you move this commet in https://github.com/stacks-network/sbtc/pull/1333 now that we closed this one?

Issue #1313, opened 2025-02-06
[Feature]: Validate rotate-keys contract calls with validated DKG shares


djordon:
# Feature - Validate rotate-keys contract calls with validated DKG shares

## 1. Description

When the signers get a rotate keys transaction to sign they validate that the aggregate key in the contract call matches the latest aggregate key from the distributed key generation (DKG) shares. With the upcoming changes we will have both validated and unvalidated signature shares associated with aggregate keys. So the latest DKG shares could be either validate of unvalidated. The change is to make sure that the check is for the latest aggregate key with validated signature shares.

### 1.1 Context & Purpose

Aggregate keys associated with unvalidated shares should not make it to the smart contract, so signers should enforce this.

## 2. Technical Details:

We’ll need a new query to fetch the latest aggregate key with aggregated shares.

### 2.1 Acceptance Criteria:

- [ ]  A signer will not provide a signature for rotate-keys contract calls where the signer has not validated the shares.

## 3. Related Issues and Pull Requests (optional):

Issue #1303, opened 2025-02-04
[Feature]: Generate pseudo-deterministic private shares during DKG


djordon:
# Feature - Generate pseudo-deterministic private shares during DKG

## 1. Description

Right now, the distributed key generation (DKG) private shares for locking the aggregate key are generated randomly using a cryptographically secure pseudo random number generator (CSPRNG) seeded with OS provided randomness. On linux, we use the `getrandom` syscall for our randomness. This is great but it comes at a cost; it's impossible for us to recreate the keys if enough signers lose their secret shares.

This ticket proposes that we seed the randomness for DKG keys using the signer's private key along with the chain tip block hash of the canonical bitcoin blockchain.

### 1.1 Context & Purpose

It's possible for errors to cause many signers to lose their secret shares over time. For example, perhaps some sizable minority of signers have their databases fail. If this were to happen, then the deposited funds would be unspendable. This is bad, so we would like to have some way of recovering the funds in a reasonable amount of time.

The solution proposed here is to open the possibility to recreate the DKG shares. If we seed the random number generator a secure but predictable way, we could re-run DKG and arrive at the DKG shares that were originally generated. This proposal requires all signers to have their original private key and agree to participated in a DKG rerun.

Also, this proposal is not to write the code to re-run DKG, just to give us a path to recovery if we were to need one.

## 2. Technical Details:

Later versions of WSTS allow the user to specify their random number generator when doing DKG. We can use this feature to use WSTS with a random number generate that has been seeded with whatever we like.

The specific proposal is to use the `rand::rngs::StdRng` as the random number generator when building a state machine for DKG. That type is a CSPRNG that implements `rand_core::SeedableRng`, where the seed can be any 32 byte sequence. This ticket proposes that the seed be the SHA256 of the signer's private key with the chain tip block hash of the canonical bitcoin blockchain.

At the time of writing, the code for specifying the random number generators is not on main, but we have a PR at https://github.com/stacks-network/sbtc/pull/1265.

### 2.1 Acceptance Criteria:

- [ ]  When doing DKG, seed the signer state machine with randomness that incorporates the private key and the bitcoin block hash.

## 4. Appendix

[...]

Issue #1286, opened 2025-01-30
[Design]: Redesign of Distributed Key Generation


djordon:
# Design - Redesign of Distributed Key Generation

## 1. Summary

We should redesign Distributed Key Generation (DKG) so that it always finishes successfully when all signers are behaving honestly, while removing the role of a coordinator.

## 2. Context & Purpose

Right now, Distributed Key Generation (DKG) is not something that can be done reliably. There are two main issues:

1. There is a timing issue where the arrival of a bitcoin block during DKG leads to failure.
2. There is another timing issue where some signers might not receive critical information from the coordinator (the signer leading DKG) “in time”. This also leads to DKG failure.

Issue (1) is quite bad, because it could lead to a partial failure. It would be better to have all signers fail or succeed at DKG. Lastly, we would like to remove any special privileges of the coordinator. There is nothing inherently wrong with having a coordinator, but removing the need for one during DKG allows for some simplification and probably enhances security.

***Relevant Research Discussions***

***External Resources***

The current DKG flow follows what is laid out in the FROST protocol [RFC 9591](https://datatracker.ietf.org/doc/rfc9591/). The FROST paper, found in https://eprint.iacr.org/2020/852, speaks about a coordinator algorithm, but does not actually require a coordinator participant for either DKG or signing rounds.

### 3.1 Proposed Component Design

The proposal is to:

- Remove the role of the coordinator during DKG.
- Create a more strict state machine for progressing through DKG.
- Require a configurable threshold of participants to send and receive acknowledgment messages before starting DKG.
- Tie the start of DKG to the block height of the canonical bitcoin blockchain, and accept DKG related messages from any signer until DKG either fails or succeeds.

The above proposals can be refined into several “small” proposals for the new DKG design.

1. Wait for all signers to observe the same bitcoin block before beginning DKG.
2. Add in a bitcoin block height/ block hash column(s) to the `dkg_shares` table. Run DKG if there is no row in the `dkg_shares` table with block height greater than or equal to the `dkg_min_bitcoin_block_height` parameter. Remove the `dkg_target_rounds` config parameter.
3. After observing a bitcoin block and writing blockchain data to the database, we have enough information to tell whether running DKG is allowed. Store that information in the `SignerState` object. Also store the current canonical bitcoin chain tip in the state object.
4. Design and implement a new DKG-only state machine that is a combination of our `SignerStateMachine` and `CoordinatorStateMachine`.
[...]

PR #1316, opened 2025-02-06
feat: validate rotate keys contract calls with validated shares

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1313.

This PR introduces a few changes that we had spoken about earlier. Also, it turns out that https://github.com/stacks-network/sbtc/issues/1313 was already handled the necessary logic required for https://github.com/stacks-network/sbtc/pull/1301, this just adds a test.

## Changes

* Change the `dkg_shares_status_id` column to just be an enum. Also rename it to `dkg_shares_status`.
* Rename the `verified_at_bitcoin_block_*` columns to `started_at_bitcoin_block_*`.
* Removed the foreign key constraints on the `dkg_shares` table. This was kinda accidental, if we want it I don't mind adding it back.
* Only allow going from `unverified` to `verified` or `failed`, no other status transitions are allowed. This allowed for some simplifications.
* The `SignerStateMachine::get_encrypted_dkg_shares` function now takes a bitcoin block ref for when DKG started. It probably makes more sense to put this information in the `SignerStateMachine` but the diff was growing and it looked more complex than the route took here.

## Testing Information

This PR adds a test for the new behavior of the `revoke_dkg_shares` and `verify_dkg_shares` functions. It also adds a new test for the validation condition in `RotateKeysV1` struct.

## Checklist:

### DIFF
--- a/signer/migrations/0012__dkg_verification_extensions.sql
+++ b/signer/migrations/0012__dkg_verification_extensions.sql
@@ -1,75 +1,52 @@
--- Enum table for DKG shares status
-CREATE TABLE sbtc_signer.dkg_shares_status (
-    -- The id of the status, not auto-incremented as we want to control the values.
-    id INTEGER PRIMARY KEY,
-    -- The name of the status.
-    key TEXT NOT NULL,
-    -- Brief description of what the status means.
-    description TEXT NOT NULL
-);
 
--- Insert the initial entries.
-INSERT INTO sbtc_signer.dkg_shares_status (id, key, description) VALUES
-    (0, 'PENDING', 'DKG round successful, pending verification via signing round'),
-    (1, 'VERIFIED', 'Successfully verified via signing round'),
-    (2, 'KEY_REVOKED', 'The DKG key has been revoked and should not be used');
+CREATE TYPE sbtc_signer.dkg_shares_status AS ENUM (
+    'unverified',
+    'verified',
+    'failed'
+);
 
 -- Add the new columns to the `dkg_shares` table. We're not adding indexes for
 -- now because the table is so small that the overhead likely outweighs the
 -- benefits.
 ALTER TABLE sbtc_signer.dkg_shares
-    -- Contains the current 
-    ADD COLUMN dkg_shares_status_id INTEGER DEFAULT 0 REFERENCES sbtc_signer.dkg_shares_status (id),
-    ADD COLUMN verified_at_bitcoin_block_hash BYTEA DEFAULT NULL,
-    ADD COLUMN verified_at_bitcoin_block_height BIGINT DEFAULT NULL,
-    ADD CONSTRAINT fk_dkg_shares_bitcoin_block_hash
-        FOREIGN KEY (verified_at_bitcoin_block_hash)
-        REFERENCES sbtc_signer.bitcoin_blocks (block_hash),
-    ADD CONSTRAINT chk_verified_at
-        CHECK (
-            (dkg_shares_status_id = 1 AND verified_at_bitcoin_block_hash IS NOT NULL AND verified_at_bitcoin_block_height IS NOT NULL) 
-            OR (dkg_shares_status_id <> 1 AND verified_at_bitcoin_block_hash IS NULL AND verified_at_bitcoin_block_height IS NULL)
-        );
+    ADD COLUMN dkg_shares_status sbtc_signer.dkg_shares_status,
+    ADD COLUMN started_at_bitcoin_block_hash BYTEA,
+    ADD COLUMN started_at_bitcoin_block_height BIGINT;
 
--- Set all of the current `dkg_shares` to `3` (revoked) to start with. Confirmed
--- DKG shares will be updated to `1` (verified) in the next step.
+
 UPDATE sbtc_signer.dkg_shares
-SET dkg_shares_status_id = 3;
+SET dkg_shares_status_id = 'unverified';
 
--- Update the `dkg_shares` which have been included in a
--- `rotate_keys_transactions` which can also be tied to a bitcoin block to `1`
--- (verified) and set the `verified_at_*` fields to the bitcoin block
--- hash/height corresponding to the block where these were anchored.
---
--- This update is not fork aware, but at the time of writing there is no forks
--- that should be problematic (i.e. we shouldn't have any rotate-keys events
--- that have been orphaned).
-WITH updated_shares AS (
+-- These are all DKG shares associated scriptPubKeys that have been successfully spent
+UPDATE sbtc_signer.dkg_shares
+SET dkg_shares_status = 'verified'
+FROM sbtc_signer.bitcoin_tx_inputs
+WHERE sbtc_signer.dkg_shares.script_pubkey = sbtc_signer.bitcoin_tx_inputs.script_pubkey;
+
+
+-- Fill in the started at bitcoin blockhash and block height. The timestamp
+-- of when we write the DKG shares row to the database should correspond
+-- with the tenure of the block that started the DKG round.
+WITH block_times AS (
     SELECT 
-        s.aggregate_key,
-        bb.block_hash AS verified_at_bitcoin_block_hash,
-        bb.block_height AS verified_at_bitcoin_block_height
-    FROM sbtc_signer.dkg_shares s
-    INNER JOIN sbtc_signer.rotate_keys_transactions rkt 
-        ON s.aggregate_key = rkt.aggregate_key
-    INNER JOIN sbtc_signer.stacks_transactions stx
-        ON rkt.txid = stx.txid
-    INNER JOIN sbtc_signer.stacks_blocks sb
-        ON stx.block_hash = sb.block_hash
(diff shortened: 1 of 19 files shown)


signer/migrations/0012__dkg_verification_extensions.sql
```diff
@@ -1,75 +1,52 @@
...
 
 -- Add the new columns to the `dkg_shares` table. We're not adding indexes for
 -- now because the table is so small that the overhead likely outweighs the
 -- benefits.
 ALTER TABLE sbtc_signer.dkg_shares
-    -- Contains the current 
-    ADD COLUMN dkg_shares_status_id INTEGER DEFAULT 0 REFERENCES sbtc_signer.dkg_shares_status (id),
-    ADD COLUMN verified_at_bitcoin_block_hash BYTEA DEFAULT NULL,
-    ADD COLUMN verified_at_bitcoin_block_height BIGINT DEFAULT NULL,
-    ADD CONSTRAINT fk_dkg_shares_bitcoin_block_hash
```
cylewitruk, line 25: I think we should keep the FK -- if we don't have the bitcoin block when we're running DKG and we've _somehow_ gotten past the chain tip validations and are writing this, then something is wrong... I'm not sure we should be succeeding DKG in that state?

signer/migrations/0012__dkg_verification_extensions.sql
```diff
@@ -1,75 +1,52 @@
...
-        FOREIGN KEY (verified_at_bitcoin_block_hash)
-        REFERENCES sbtc_signer.bitcoin_blocks (block_hash),
-    ADD CONSTRAINT chk_verified_at
-        CHECK (
-            (dkg_shares_status_id = 1 AND verified_at_bitcoin_block_hash IS NOT NULL AND verified_at_bitcoin_block_height IS NOT NULL)
```
cylewitruk, line 30: These check constraints (if modified slightly after the enum change) ensure that the status can't be `verified` without these also being present, and vice versa. I think we should keep them from a data integrity perspective.

cylewitruk: Ah, nevermind, you added not null below so this can never happen since we write at start block 👍

signer/src/storage/in_memory.rs
```diff
@@ -1284,24 +1284,20 @@ impl super::DbWrite for SharedStore {
     {
         let mut store = self.lock().await;
         if let Some((_, shares)) = store.encrypted_dkg_shares.get_mut(&aggregate_key.into()) {
```
matteojug, line 1286: maybe nit: should we add also for the in memory the constraint of not updating if the state is not pending?

signer/src/storage/model.rs
```diff
@@ -557,16 +553,19 @@ impl From<SignerVotes> for BitArray<[u8; 16]> {
     }
 }
 
-/// Possible states for `dkg_shares` entries.
-#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord, strum::Display)]
+/// The possible states for DKG shares.
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord, sqlx::Type)]
+#[sqlx(type_name = "dkg_shares_status", rename_all = "snake_case")]
 #[cfg_attr(feature = "testing", derive(fake::Dummy))]
+// #[strum(serialize_all = "snake_case")]
```
matteojug, line 560: nit: leftover `strum`

signer/src/storage/model.rs
```diff
@@ -557,16 +553,19 @@ impl From<SignerVotes> for BitArray<[u8; 16]> {
...
 pub enum DkgSharesStatus {
-    /// DKG round successful, pending verification via signing round.
-    Pending,
-    /// Successfully verified via signing round.
-    Verified(BitcoinBlockRef),
-    /// The DKG key has been revoked and should not be used.
-    Revoked,
+    /// The DKG shares have not passed or failed verification.
+    Unverified,
+    /// The DKG shares have passed or verification.
```
matteojug, line 564:
```suggestion
    /// The DKG shares have passed verification.
```

signer/src/transaction_signer.rs
```diff
@@ -1024,7 +1008,12 @@ where
             .get(id)
             .ok_or(Error::MissingStateMachine)?;
 
-        let encrypted_dkg_shares = state_machine.get_encrypted_dkg_shares(&mut self.rng)?;
+        let StateMachineId::Dkg(started_at) = id else {
+            return Err(Error::TypeConversion);
```
matteojug, line 1012: nit: maybe using `UnexpectedStateMachineId` is better suited here

signer/migrations/0012__dkg_verification_extensions.sql
matteojug: I would be for not having it, eg suppose at some point we want to prune our bitcoin view but still keep the shares for whatever reason

cylewitruk: I'm fine either way -- tbh I'm not really sure of the purpose of having the block hash here at all:
- Wrt #1286, It's the block height that's mentioned as a sort of logical clock, I don't see a use for the block hash since the canonical height will never "go backwards".
- If it's for use as a salt for deterministic generation, I would probably just call this "salt" or something instead and treat it as such; then there's no implicit relation that may or may not exist.

djordon: Since we think we'd want it for https://github.com/stacks-network/sbtc/issues/1303, I think it makes the most sense to keep it as-is for now. We can rename it though.

signer/src/stacks/contracts.rs
```diff
@@ -1225,6 +1215,10 @@ pub enum RotateKeysErrorMsg {
     /// There is already a key rotation with the same details.
     #[error("there is already a key rotation with the same details")]
     KeyRotationExists,
+    /// The aggregate key is known but the associated secret shares have
+    /// not passed verification.
+    #[error("the shares associated with the aggregate key have not passes verification")]
+    SecretSharesNotVerified,
```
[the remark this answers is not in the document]
djordon: That makes sense to me, will update.

signer/src/storage/in_memory.rs
djordon: Ah, yeah we should.

signer/src/transaction_signer.rs
djordon: Oh I just forgot to get back to this one and glossed over it all when opening the PR. Will update.

signer/src/error.rs
```diff
@@ -22,11 +22,11 @@ pub enum Error {
 
     /// The DKG shares for the provided aggregate key have not been verified.
     #[error("the dkg shares for the provided aggregate key have not been verified: {0}")]
-    DkgSharesNotVerified(Box<PublicKeyXOnly>),
+    DkgSharesNotVerified(PublicKey),
 
     /// The DKG shares for the provided aggregate key have been revoked.
     #[error("the provided aggregate key has been revoked: {0}")]
-    DkgSharesRevoked(Box<PublicKeyXOnly>),
+    DkgSharesRevoked(PublicKey),
```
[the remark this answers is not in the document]
djordon: Yup, removed in https://github.com/stacks-network/sbtc/pull/1316/commits/9a660388b1d62de852155f4b33c6063f5f189731.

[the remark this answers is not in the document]
djordon: Yup, removed in https://github.com/stacks-network/sbtc/pull/1316/commits/9a660388b1d62de852155f4b33c6063f5f189731.

signer/src/transaction_signer.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/1316/commits/36424ecb8063aeff3df2c5f0558bc3bba8b18ae7.

signer/src/storage/model.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/1316/commits/1fb3ffa10c36b8349fddcd3ad0b08aeee63bf362

signer/src/storage/model.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/1316/commits/1fb3ffa10c36b8349fddcd3ad0b08aeee63bf362

signer/src/storage/in_memory.rs
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1316/commits/97142e3f25098c00851959ea9429a4c79556ae3e

signer/src/stacks/contracts.rs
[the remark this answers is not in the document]
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1316/commits/6570d00ddb4c62063196fffdfcdecf3e1f491e7d

Issue #1300, opened 2025-02-04
[Feature]: DKG key verification step pre-rotate-key


cylewitruk:
# Feature - DKG key verification step

## 1. Description

We should introduce a form of verification that raises confidence that the signers, together with their newly generated key shares, can generate a proper signature and that it correctly unlocks a UTXO locked by the new scriptPubKey.

Quick links:
- [Alternative 1](#alternative-1)
- [Alternative 2](#alternative-2)
- [Alternative 3](#alternative-3)

### 1.1 Context & Purpose

When the signers have re-run DKG and are going to sweep from the current UTXO locked by the `current` key to a new UTXO locked by the `new` key, there are currently no verifications in place to ensure that, according to Bitcoin validation rules, that the new UTXO would indeed be spendable in a new transaction when the input is signed using WSTS prior to moving the funds.

There are three proposals for solving this to various degrees (summarized from discussions on Slack):

### 1.1.1 Alternative 1<a id='alternative-1'>

Require a sweep transaction from `current` to a UTXO locked by `new` to include a deposit that spends the signers' new UTXO. Bitcoin core would then reject the transaction if its validation and/or consensus rules determine that the new UTXO can't be spent by the new transaction.

This is the least-effort change of the proposed solutions, however there are some considerations:

 1. It requires a deposit/withdrawal request locked by the `new` key to trigger this validation. In the case of a compromised key, we may have to (ourselves) make a deposit to trigger this logic. If the caps are reached, we have to modify caps and risk other deposits sneaking in.
 1. Deposit/withdrawal requests are spent with script-path spending, while the signers' UTXO is locked with a key-spend only taproot scriptPubKey, which wouldn't be testing the same thing. Though if a deposit could be spent but not the UTXO it should be a code bug and fixable.
 1. Users' deposit/withdrawal requests are locked by a spend-path that we haven't verified. We could still sweep deposits locked by the `current` key and withdrawals, but deposits locked by the `new` key may get stuck and need to be reclaimed.

### 1.1.2 Alternative 2<a id='alternative-2'>

Require a check that the `new` signers' scriptPubKey is spendable when an input is signed with the new key over WSTS. This would entail creating two **valid** transactions that Bitcoin core would accept into the mempool (i.e. they need to be well-formed and spend valid UTXO's):

- Transaction **T1** spending the `current` UTXO to a new UTXO locked by the `new` scriptPubKey. This would simulate the sweep transaction transferring ownership to the `new` key.
[...]

PR #1317, opened 2025-02-06
feat: add DKG verification window

matteojug:
## Description

Part of https://github.com/stacks-network/sbtc/issues/1300

## Changes

Add `dkg_verification_window`, after which the signer mark the unverified dkg as failed. Also exclude failed shares from `get_encrypted_dkg_shares_count`. 

## Testing Information

Add a test to check that the shares are marked as failed and excluded from the count.

## Checklist:


signer/tests/integration/block_observer.rs
```diff
@@ -1171,3 +1178,225 @@ async fn block_observer_updates_state_after_observing_bitcoin_block() {
 
     testing::storage::drop_db(db).await;
 }
+
+/// This test checks that the block observer correctly update the state of
+/// pending DKG shares once they exit the verification window
+#[tokio::test]
+async fn block_observer_updates_dkg_shares_after_observing_bitcoin_block() {
```
djordon, line 1185: Nice test.

signer/src/block_observer.rs
```diff
@@ -628,6 +634,38 @@ impl<C: Context, B> BlockObserver<C, B> {
...
+        let chain_tip = db
+            .get_bitcoin_block(&chain_tip.into())
+            .await?
+            .ok_or(Error::NoChainTip)?;
+        let verification_window = self.context.config().signer.dkg_verification_window;
+
+        if last_dkg
+            .started_at_bitcoin_block_height
+            .saturating_add(verification_window as u64)
+            < chain_tip.block_height
```
djordon, line 662: nit: for readability, I'd break this up into multiple lines like so:
```rust
let max_verification_height = last_dkg
    .started_at_bitcoin_block_height
    .saturating_add(verification_window as u64);

if max_verification_height < chain_tip.block_height {
```

djordon (approved):
Looks good ✅

signer/src/config/mod.rs
```diff
@@ -494,6 +497,7 @@ impl Settings {
             DEFAULT_MAX_DEPOSITS_PER_BITCOIN_TX,
         )?;
         cfg_builder = cfg_builder.set_default("signer.dkg_target_rounds", 1)?;
+        cfg_builder = cfg_builder.set_default("signer.dkg_verification_window", 10)?;
```
djordon, line 500: I would change this to 100, or 200 or something. I think verification is only attempted by the coordinator, so if we end up in a state of partial verification we need to make sure that each signer has enough attempts at verification.

cylewitruk: I think the idea was that the rule:

```
If latest key is not failed and ≠ contract key, attempt verification
```

Should have every coordinator, even if they've flagged the shares as verified, attempting verification each block, until:
- The contract aggregate key matches ours, OR
- This window has expired

matteojug: Right, and it should already be the case (at least looking at the code in this branch): the coordinator always calls `check_and_submit_rotate_key_transaction`, and early return if the latest aggregate key matches the smart contract; so we should already be trying to validate regardless of the verification, just based on the smart contract (rule 4).

This also means that currently (again, looking at the code in this branch):
 - we are attempting verification even if the latest share failed -- although then we should first hit the `coordinate_dkg`, so we would attempt verification of a failed dkg only if the latest is failed and we already have too many DKGs to run a new one (re, round target), or if the block observer changed the state between the check to run DKGs and the `check_and_submit_rotate_key_transaction` contract call (because of a flash block race maybe?)
 - we not attempting the verification if we have unverified shares (regardless of the smart contract)
plus probably missing some other corner cases of the rule set; talking about this branch just because it may be that this was changed in the siblings PRs

I would refine those checks in a separate PR, just to keep things manageable (and keep the low default for the window).

signer/src/storage/mod.rs
```diff
@@ -232,7 +232,7 @@ pub trait DbRead {
         &self,
     ) -> impl Future<Output = Result<Option<model::EncryptedDkgShares>, Error>> + Send;
 
-    /// Returns the number of DKG shares entries in the database.
+    /// Returns the number of non failed DKG shares entries in the database.
```
cylewitruk, line 235:
```suggestion
    /// Returns the number of non-failed DKG shares entries in the database.
```

signer/src/storage/postgres.rs
```diff
@@ -1531,10 +1531,12 @@ impl super::DbRead for PgStore {
 
     /// Returns the number of rows in the `dkg_shares` table.
```
cylewitruk, line 1532:
```suggestion
    /// Returns the number of unverified rows in the `dkg_shares` table.
```

signer/src/storage/postgres.rs
```diff
@@ -1529,12 +1529,14 @@ impl super::DbRead for PgStore {
...
     }
 
-    /// Returns the number of rows in the `dkg_shares` table.
+    /// Returns the number of non-failed rows in the `dkg_shares` table.
     async fn get_encrypted_dkg_shares_count(&self) -> Result<u32, Error> {
-        let count: i64 = sqlx::query_scalar("SELECT COUNT(*) FROM sbtc_signer.dkg_shares;")
-            .fetch_one(&self.0)
-            .await
-            .map_err(Error::SqlxQuery)?;
+        let count: i64 = sqlx::query_scalar(
```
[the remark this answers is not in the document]
cylewitruk: @MCJOHN974 `sqlx` will complain as it only has built-in support for `i32` (`INTEGER`) and `i64` (`BIGINT`) types, so we have to coerce these after the query.

signer/src/config/mod.rs
matteojug: Missing cases handled in https://github.com/stacks-network/sbtc/pull/1322

signer/src/block_observer.rs
matteojug: Done in https://github.com/stacks-network/sbtc/pull/1317/commits/bf615de39807eb21efad8ced41cb8d8ce27682c6

signer/src/storage/postgres.rs
matteojug: Done in https://github.com/stacks-network/sbtc/pull/1317/commits/660698878265dffdc7768171691efb4bee38f040

signer/src/storage/mod.rs
matteojug: Done in https://github.com/stacks-network/sbtc/pull/1317/commits/660698878265dffdc7768171691efb4bee38f040

Issue #1314, opened 2025-02-06
[Feature]: Do not use DKG shares that have failed validation


djordon:
# Feature - Do not use DKG shares that have failed validation

## 1. Description

We will be testing whether distributed key generation (DKG) shares actually work as part of the production flow. If some DKG shares fail validation, we should never use them. Let’s make it so.

### 1.1 Context & Purpose

If DKG shares fail validation then we shouldn’t use them or sign for them. It’s as if they have been deleted.

## 2. Technical Details:

We’ll need to re-write some of our queries to filter the shares that have failed validation.

### 2.1 Acceptance Criteria:

- [ ]  Shares that have failed validation are never used for signing.

## 3. Related Issues and Pull Requests (optional):

PR #1318, opened 2025-02-07
feat: do not use DKG shares that have failed verification

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1314

## Changes

* Remove the context window from the `BitcoinTxContext`.
* Add two new trait queries for fetching the DKG shares status and the latest verified DKG shares. They need tests.
* Do not sign for deposits locked with an aggregate key whose shares have not passed verification. This is to make sure that bad shares do not make the system worse.
* Only use verified shares when constructing the signer's new UTXO.

## Testing Information

I need to add some tests for almost everything

## Checklist:


signer/src/storage/mod.rs
```diff
@@ -226,12 +226,26 @@ pub trait DbRead {
     where
         X: Into<PublicKeyXOnly> + Send;
 
+    /// Return the applicable DKG shares status for the given aggregate key
+    fn get_dkg_shares_status<X>(
+        &self,
+        aggregate_key: X,
+    ) -> impl Future<Output = Result<Option<model::DkgSharesStatus>, Error>> + Send
+    where
+        X: Into<PublicKeyXOnly> + Send;
```
matteojug, line 235: We could define a default impl using `get_encrypted_dkg_shares`: it would not be as optimized query-wise, but it would be one less query (and reusing one that it's already used/tested around) and it's not like fetching a full row it's our db issue at the moment 😆

signer/src/transaction_signer.rs
```diff
@@ -1148,24 +1162,25 @@ where
...
-                self.context
-                    .get_storage_mut()
-                    .verify_dkg_shares(aggregate_key)
-                    .await?;
-                tracing::info!(
-                    "🔐 DKG shares entry has been marked as verified; it is now able to be used"
-                );
+                match mock_tx.verify_signature(&signature) {
+                    Ok(()) => {
+                        tracing::info!("🔐 \x1b[1;32msignature verification successful\x1b[0m");
```
matteojug, line 1169: nit: remove colors?

matteojug (commented):
LGTM

signer/src/storage/mod.rs
djordon: Yeah that works, I'll do that and remove the new query.

signer/src/transaction_signer.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/1318/commits/402f13ca3827c551bd814b09261ae1cf5578886b

signer/src/storage/mod.rs
djordon: Cleaned up in https://github.com/stacks-network/sbtc/pull/1318/commits/a2f5b6b7816284955840efff7c90b69ca402e022

matteojug (approved):

PR #1320, opened 2025-02-07
Updates `flake.lock` so that `rust-overlay` includes Rust 1.84

aldur:
Context: `flake.lock` pins to a `git rev`. The previous one was pointing at a commit that didn't include the pre-built Rust 1.84 compiler. This fixes it.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1726560853,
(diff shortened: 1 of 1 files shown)


cylewitruk (approved):
Verified here https://github.com/oxalica/rust-overlay/commit/1ff8663cd75a11e61f8046c62f4dbb05d1907b44 lgtm

PR #1324, opened 2025-02-07
feat: state machine for out-of-order message handling during dkg verification

cylewitruk:
## Description

Part of #1300 , and followup to #1287, #1285 and #1301.

ℹ️ A lot of the diff is test-related code.

Due to the FROST coordinator being "passively strict" in its ordering of messages, it's likely that over WAN some signers may receive `NonceResponse` messages from fast signers before having received the `NonceRequest` from the coordinator, which will be ignored and thus the passive coordinator not managing to transition between states correctly. The same situation applies to the `SignatureGatherRequest` vs. its response.

This PR introduces a state machine for handling out-of-order messages received from other signers.

Note that this is only applicable to the "passive FROST coordinator" which each signer runs to track the signing round and generate the signature locally; the `SignerStateMachine` which handles the "participant" side of things doesn't have this issue.

## Changes

The primary addition in this PR is the above mentioned DKG verification state machine (`dkg::verification::StateMachine`). Here's a high-level overview:

- Handles the enqueuing of `NonceRequest`, `NonceResponse`, `SignatureShareRequest` and `SignatureShareResponse` messages used by the `FrostCoordinator`.
- Monitors the state transitions of the `FrostCoordinator` and feeds it only the messages that it can process given its current state.
- Processes all queued messages allowed by the current state of the `FrostCoordinator` and handles the continued processing with the next allowed message types if a state change occurred during the processing of the queue.
- Messages are de-duplicated by `(message type, sender pubkey)` per state machine instance, as we expect to only receive exactly one each of the `*Request` message types, and exactly `[num signers]´ of the `*Response*` message types.

#### Other

- Created a new root mod `dkg` to try to keep things clean. I figure this could also be a good future home for some of the stuff described in #1286.
- Includes a "reasonable" test suite for the state machine + some general test helpers -- it's not 100% coverage, but covers some of the more important aspects. The relevant flows are largely tested in existing integration tests as well.
- Stopped caching the `UnsignedMockTransaction` instances and create them on-the-fly instead for simplicity; they're small and we shouldn't be doing this often anyway.
- Some other minor tweaks along the way.

## Notes

For the `StateMachineId` we're still using just the new aggregate public key -- this means that the state machine can "exist" across block boundaries (and it will work just fine if everyone else is still on-board), with the exception of if a new block comes between `NonceRequest` and `SignatureShareRequest` -- then the "is coordinator" logic will prevent the signers from receiving the `SignatureShareRequest`. The state machine will expire after (a currently hard-coded) 30 seconds, though, so given a "normal" bitcoin block cadence this should resolve within a block or two.

## Testing Information

Have tested this a bit in devenv and it works well. I haven't tried a key rotation + sweep, though...

## Checklist:


cylewitruk:
Hm.. CI failed two tests which sometimes are flaky for me locally:

```
        FAIL [  13.334s] signer::integration transaction_coordinator::sign_bitcoin_transaction_multiple_locking_keys
──── STDOUT:             signer::integration transaction_coordinator::sign_bitcoin_transaction_multiple_locking_keys

running 1 test
2025-02-07T16:42:05.562066Z ERROR signer::transaction_signer: error processing signer message error=the dkg verification state machine raised an error: the state machine has reached the message limit for message type NonceRequest (expected: 1, actual: 2)
2025-02-07T16:42:05.563287Z ERROR signer::transaction_signer: error processing signer message error=the dkg verification state machine raised an error: the state machine has reached the message limit for message type NonceRequest (expected: 1, actual: 2)
2025-02-07T16:42:05.564121Z ERROR signer::transaction_signer: error processing signer message error=the dkg verification state machine raised an error: the state machine has reached the message limit for message type NonceRequest (expected: 1, actual: 2)
2025-02-07T16:42:15.410293Z ERROR signer::transaction_coordinator: error processing requests; skipping this round error=coordinator timed out after 10 seconds
2025-02-07T16:42:15.558477Z ERROR signer::transaction_coordinator: error processing requests; skipping this round error=coordinator timed out after 10 seconds
```

I wonder if this could be a hint to underlying flake 🤔 I'll try removing the error and see what happens.

djordon (commented):
I still have more to review here, just sending what I have now.

signer/src/transaction_signer.rs
```diff
@@ -1066,114 +1123,93 @@ where
...
                     .get_storage_mut()
                     .verify_dkg_shares(aggregate_key)
                     .await?;
+
                 tracing::info!(
                     "🔐 DKG shares entry has been marked as verified; it is now able to be used"
                 );
+
+                // We're done, so remove the state machine entry.
+                self.dkg_verification_state_machines.pop(&state_machine_id);
```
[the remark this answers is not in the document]
cylewitruk: Whoops, yeah I agree, wasn't intended 👍

signer/src/dkg/testing.rs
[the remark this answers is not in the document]
cylewitruk: It crossed my mind, but as (most of this) is so specific to this module, it felt more natural to keep it self-contained -- at least until it proves to be more broadly usable, then it's just to move it.

signer/src/transaction_signer.rs
```diff
@@ -996,15 +1041,17 @@ where
...
+    /// by the given state machine id.
     #[tracing::instrument(skip(self))]
-    async fn store_dkg_shares(&mut self, id: &StateMachineId) -> Result<(), Error> {
+    async fn store_dkg_shares(&mut self, state_machine_id: &StateMachineId) -> Result<(), Error> {
         let state_machine = self
             .wsts_state_machines
-            .get(id)
-            .ok_or(Error::MissingStateMachine)?;
+            .get(state_machine_id)
+            .ok_or(Error::MissingStateMachine(*state_machine_id))?;
```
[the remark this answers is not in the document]
cylewitruk: Ever so slightly 😅  [204b5e5](https://github.com/stacks-network/sbtc/pull/1324/commits/204b5e52006bfe2154570f8f526f3eda0400413c)

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,804 @@
...
+    /// specified timeout, which allows the verification to span multiple
+    /// Bitcoin blocks, being limited by wall-clock time instead of Bitcoin
+    /// block cadence.
+    created_at: Instant,
+    /// Specifies the amount of time elapsed since `created_at` that this
+    /// verification should be valid.
+    timeout: Duration,
+    /// The signature that has been produced by the DKG verification. This is
+    /// only set if/once the DKG verification has completed successfully.
+    state: State,
```
matteojug, line 134: nit: outdated comment?

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,804 @@
...
+}
+
+/// Represents the state of a DKG verification. This implementation is designed
+/// to be able to span across multiple Bitcoin blocks, being limited by
+/// wall-clock time instead of Bitcoin block cadence.
+///
+/// TODO: The signer currently stops messages from being processed if the
+/// received message doesn't match the current chain tip, so that needs to be
+/// changed for cross-block functionality to work.
+#[allow(dead_code)]
```
matteojug, line 109: nit: leftover `#[allow(dead_code)]`?

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,804 @@
...
+        // can receive out-of-order messages, and we need to ensure that we
+        // process all messages that we can before returning, including handling
+        // state transitions.
+        loop {
+            let current_processable_message_type = self.current_processable_message_type();
+            self.process_queued_messages()?;
+            if current_processable_message_type == self.current_processable_message_type() {
+                break;
+            }
+        }
```
matteojug, line 209: if `self.process_queued_messages()` were to return the number of processed messages (or even if we did process at least one message) we could simplify this by iterating until we stop doing things

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,804 @@
...
+    aggregate_key: secp256k1::XOnlyPublicKey,
+    /// The state machine that is being used to verify the signer.
+    coordinator: FrostCoordinator,
+    /// WSTS messages that have been received from other signers. We keep them
+    /// in a [`HashMap`] where the keys are tuples of message types and sender
+    /// public keys, and the values are queued messages. This allows us to
+    /// manage messages by type and sender, ensuring that we can handle
+    /// out-of-order messages and process them correctly, while also
+    /// constraining each signer to one message per message type.
+    wsts_messages: HashMap<(WstsNetMessageType, PublicKey), QueuedMessage>,
```
matteojug, line 122: Given the usage we do below something like:
```suggestion
    wsts_messages: HashMap<WstsNetMessageType, HashMap<PublicKey, QueuedMessage>>,
```
could simplify some operations.

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,804 @@
...
+    /// 4. Once this condition is met, the coordinator will transition to either
+    ///    the `Success` or `Error` state, depending on the result of the
+    ///    signing operation.
+    fn current_processable_message_type(&self) -> WstsNetMessageType {
+        match self.coordinator.state {
+            wsts::state_machine::coordinator::State::Idle => WstsNetMessageType::NonceRequest,
+            wsts::state_machine::coordinator::State::NonceGather(_) => {
+                WstsNetMessageType::NonceResponse
+            }
+            _ => WstsNetMessageType::SignatureShareResponse,
```
matteojug, line 291: As the underlying coordinator can end up also in other states (eg, dkg related ones), can we specify the exact state for `SignatureShareResponse` and return an err for the others?

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,804 @@
...
+                &sender.to_string()[..10]
+            );
+
+            // Pass the message to the coordinator.
+            let (_, result) = self
+                .coordinator
+                .process_message(&msg.message)
+                .map_err(|error| Error::Coordinator(Box::new(error)))?;
+
+            msg.mark_processed();
```
matteojug, line 368: Can we first mark as processed and then process it? With the rationale that if we have a bad message causing errors, we risk to trigger it over and over (as we would never reach the `mark_processed`)

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,804 @@
...
+
+            // Check the result of the operation. If the operation is one of
```
matteojug, line 370: nit: missing words

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,804 @@
...
+            .filter_map(|((msg_type, sender), msg)| {
+                if *msg_type == message_type_to_process && !msg.processed {
+                    Some((sender, msg))
+                } else {
+                    None
+                }
+            });
+
+        // Process all of the messages that we determined could be processed.
+        for (sender, msg) in messages {
```
matteojug, line 355: It could happen that we change the state while we still have messages of a certain type to process (eg, signers can "reset" their state by sending again a message, as in `enqueue_message` we overwrite with non processed messages). It should be fine -- the state machine would just ignore the messages -- but for completeness can we check, at the end of the loop, if the state machine state is still the one we expect, and if not break?

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,804 @@
...
+/// Represents the state of a DKG verification. This implementation is designed
+/// to be able to span across multiple Bitcoin blocks, being limited by
+/// wall-clock time instead of Bitcoin block cadence.
+///
+/// TODO: The signer currently stops messages from being processed if the
+/// received message doesn't match the current chain tip, so that needs to be
+/// changed for cross-block functionality to work.
+#[allow(dead_code)]
+#[derive(Debug)]
+pub struct StateMachine<TError> {
```
matteojug, line 111: Do we need the error template? It's not like we use this with something other than the crate `Error`

signer/src/transaction_signer.rs
```diff
@@ -1047,14 +1097,22 @@ where
...
             dkg_shares.signature_share_threshold,
             signer_private_key,
         )
-        .await
+        .await?;
+
+        let state_machine = dkg::verification::StateMachine::new(
+            coordinator,
+            aggregate_key,
+            Duration::from_secs(30), // TODO: Config?
```
matteojug, line 1112: Can we set this to unlimited or none? Currently we rely on bitcoin blocks and/or other timeouts, having this may conflict with something else

signer/src/wsts_state_machine.rs
```diff
@@ -45,7 +45,19 @@ pub enum StateMachineId {
     /// Identifier for a Bitcoin signing state machines
     BitcoinSign(SigHash),
     /// Identifier for a rotate key verification signing round
-    RotateKey(PublicKeyXOnly, model::BitcoinBlockHash),
+    RotateKey(PublicKeyXOnly),
```
matteojug, line 48: Are we sure about removing the bitcoin block hash (in this moment)? It may have effects we don't want to handle now, eg having old state machines around in some state that may be reused in a new tenure (validating the same aggregate key), potentially stalling the process.

signer/src/dkg/verification.rs
cylewitruk: I had it like that before and switched it to the current form because it actually simplified things a bit :)

signer/src/dkg/verification.rs
cylewitruk: Yup, good point

signer/src/dkg/verification.rs
cylewitruk: I'd prefer to keep it because it keeps this file clean and concise wrt errors, and the supers can use a type alias if it's too tedious to write `StateMachine<Error>`.

cylewitruk: But I'll look at it tomorrow, it might be that the reason I needed it went away with some other changes I made recently

signer/src/dkg/verification.rs
cylewitruk: Yeah or just a bool, even easier, good catch 👍

signer/src/dkg/verification.rs
cylewitruk: [4c38914](https://github.com/stacks-network/sbtc/pull/1324/commits/4c38914acadbc9d76e296f1fc3094ff69c17218a)

signer/src/testing/mod.rs
```diff
@@ -81,3 +81,24 @@ pub fn set_witness_data(unsigned: &mut UnsignedTransaction, keypair: secp256k1::
...
         });
 }
+
+/// Testing helpers for [`Vec`].
+pub trait IterTestExt<T>
+where
+    Self: IntoIterator<Item = T> + Sized,
+{
+    /// Asserts that the vec
+    fn unwrap_one(self) -> T {
```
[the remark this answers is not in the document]
cylewitruk: Went with `single()` like in scala or c#, rust itertools uses the `exactly_one` 🤷 [e1778d5](https://github.com/stacks-network/sbtc/pull/1324/commits/e1778d5f4308be9c1c8b6f251dfbfb4c9063210c)

signer/src/dkg/verification.rs
cylewitruk: Yeah it's not needed, [5f8a13e](https://github.com/stacks-network/sbtc/pull/1324/commits/5f8a13e2890cdf94b834518f78e79a6eb9949d55)

signer/src/error.rs
```diff
@@ -564,6 +559,10 @@ pub enum Error {
     #[error("WSTS coordinator error: {0}")]
     WstsCoordinator(#[source] Box<wsts::state_machine::coordinator::Error>),
 
+    /// WSTS signing error.
+    #[error("WSTS signing error: {0:?}")]
+    WstsUnexpectedResult(Box<wsts::state_machine::OperationResult>),
+
```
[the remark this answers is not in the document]
cylewitruk: [b5e18a1](https://github.com/stacks-network/sbtc/pull/1324/commits/b5e18a18a9541602b30cffa3cf64937d09de33c2)

signer/src/dkg/wsts.rs
```diff
@@ -0,0 +1,47 @@
+//! Containers WSTS helper types and impls for use within the DKG module.
```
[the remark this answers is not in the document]
cylewitruk: [a8f9b0b](https://github.com/stacks-network/sbtc/pull/1324/commits/a8f9b0bfea633bbd0f0748298615b8da2050aff1)

signer/src/dkg/verification.rs
cylewitruk: [d01e277](https://github.com/stacks-network/sbtc/pull/1324/commits/d01e2774daece93a5893a7ae34007d6e3d4241db)

signer/src/dkg/verification.rs
cylewitruk: [d01e277](https://github.com/stacks-network/sbtc/pull/1324/commits/d01e2774daece93a5893a7ae34007d6e3d4241db)

signer/src/dkg/verification.rs
cylewitruk: Also done in [4c38914](https://github.com/stacks-network/sbtc/pull/1324/commits/4c38914acadbc9d76e296f1fc3094ff69c17218a)

signer/src/dkg/verification.rs
cylewitruk: [c9c66c4](https://github.com/stacks-network/sbtc/pull/1324/commits/c9c66c4c683269d36267fffc020ac3ed7959ab81)

signer/src/wsts_state_machine.rs
cylewitruk: Well it's supposed to work in tandem with the internal state machine timeout, as after that timeout the state machine will be invalid. I changed the timeout to the config value for the signing timeout that the coordinator uses for all signing (including this one), which by default is 10s. The tx-signer also resets the state machine on retrieval if it exists and is in an end-state/timed-out now, and also removes its sibling signer-state-machine.

Idunno, I don't think there's anything "bad" that can happen (short of maybe one tenure erroring), but I could have missed an edge case. I'm still missing tests for this logic, though.

signer/src/transaction_signer.rs
cylewitruk: See below comment, they kinda hang together

signer/src/dkg/verification.rs
cylewitruk: This was changed around quite a bit in [c371a97](https://github.com/stacks-network/sbtc/pull/1324/commits/c371a974ef39859a568921efd032ea9ac274abb3), maybe have a new look and see if you think this comment is still relevant?

signer/src/dkg/verification.rs
cylewitruk: This was added in the bigger [c371a97](https://github.com/stacks-network/sbtc/pull/1324/commits/c371a974ef39859a568921efd032ea9ac274abb3)

signer/src/dkg/verification.rs
matteojug: It is still relevant: the possible "issue" (it should be handled by the inner state machine already, it's just to avoid relying on that unverified assumption) is that you may change your state (and thus what messages to process) while iterating over those messages, so you may end up consuming messages for the wrong state.

signer/src/wsts_state_machine.rs
matteojug: To me it seems like a new complexity we need to be sure works as expected with the rest of changes (and the already funky set of timeouts and interactions we have), so unless it's something we need right now I would prefer to keep the current _stick to the coordinator tenure_ behaviour and change it later on, instead of a mix situation of having some of the changes for cross-tenure things but still everything else tied to the coordinator tenure.

signer/src/dkg/verification.rs
cylewitruk: Ohh now I get what you mean. Yeah, that's completely possible, but as you mentioned these will be ignored by the WSTS coordinator as it will have progressed to the "next" state. I didn't want to build in too much actual state logic here and rather delegate it to frost, but we could track the frost state before/after processing and return early if changed.

signer/src/wsts_state_machine.rs
cylewitruk: I'm fine with adding block hash back, I'm genuinely not that concerned about these signing transactions crossing bitcoin block boundaries as we've never really had a problem with pure signing (on the individual-round level, anyway); and if it happens we have the other safeguards & fallbacks.

matteojug: If the new coordinator wouldn't try again I could agree, but it could happen that flashblocks + new coordinator doing new coordinator things would put us in a strange situation (maybe with some state machine still dirty and not in a terminal state yet) -- which could be fine, but I'm not sure and that's kinda my point: this whole DKG verification is already adding complexity, I would really avoid adding now anything extra that is not needed and we are not sure is not harmful.

signer/src/dkg/verification.rs
matteojug: Yeah I was thinking of something like that, ie if `self.current_processable_message_type` changed just return  -- we are already making decisions based on the underlying state, and we would keep the invariant here that is kinda defined at the beginning of this function (we pick the messages based on the inner state).

signer/src/transaction_signer.rs
```diff
@@ -1218,54 +1310,80 @@ where
...
+                tracing::warn!("missing signing round");
+                return Err(Error::MissingStateMachine(*state_machine_id));
+            }
         };
 
+        // Validate that the sender is a valid member of the signing set and
+        // has the correct id according to the signer state machine.
+        if let Some(signer_id) = signer_id {
+            self.validate_sender(state_machine_id, signer_id, &sender)?;
+        }
```
djordon, line 1333: We should do this before feeding any messages to the state machine.

cylewitruk: Yeah, good point 👍  [7b56d5e](https://github.com/stacks-network/sbtc/pull/1324/commits/7b56d5ec06d3645bfd83b51ba729525963c38626)

cylewitruk: Tbh I think we should ask Joey if he can include the `signer_id` in all messages where possible, so that we can do that check. I don't know how much security it adds in reality, but it's a nice "stupidity check" to know that the message is coming from the same signer as the different state machine's think it is.

signer/src/wsts_state_machine.rs
cylewitruk: > I'm fine with adding block hash back

☝️ :) I was just adding sub-thoughts -- I'll add it back

signer/src/transaction_signer.rs
djordon: Yeah, I believe it's the coordinator's messages that are missing the `signer_id`. Basically, WSTS sometimes needs to know who the sender is, because some messages are saying something like signer A's signatures shares are X, or signer B's public commitments are Y, and so on. The coordinator's messages follow a different model, where WSTS does not know who or what the coordinator is because that decision happens outside of WSTS and the "coordinator" does not make the same kinds of statement as "signers". This means that the coordinator does not need to be one of the signers and so does not necessarily have a signer id. Our coordinator plays both roles so it has a signer id be someone could design their system differently.

And we have a separate coordinator check for the coordinator's messages, ensuring that we have this security check all the time, it's just not obvious that we do.

If we do https://github.com/stacks-network/sbtc/issues/1286, then there wouldn't be a coordinator and we would design some things a little differently.

signer/src/wsts_state_machine.rs
cylewitruk: Done in [2320ef9](https://github.com/stacks-network/sbtc/pull/1324/commits/2320ef9061541789bd5b4e23734b402444426807), though it became a `BlockHashRef`, same as the `StateMachineId::Dkg` variant uses now.

signer/src/dkg/verification.rs
cylewitruk: Changed it to what I suggested above, as otherwise there's immutable+mutable borrow issues with the loop. The solution is to `collect()` the `messages_to_process` into a `Vec<_>`, but that's much less efficient.

[1566c74](https://github.com/stacks-network/sbtc/pull/1324/commits/1566c7488ff649309bf1040085f57ea0f44d029a)

signer/src/transaction_signer.rs
cylewitruk: Added support for `None` = `unlimited` and use that in the creation in tx-signer. [85aae81](https://github.com/stacks-network/sbtc/pull/1324/commits/85aae816747a687f1c09e9819204f4c382363152)

signer/src/wsts_state_machine.rs
```diff
@@ -45,7 +45,23 @@ pub enum StateMachineId {
     /// Identifier for a Bitcoin signing state machines
     BitcoinSign(SigHash),
     /// Identifier for a rotate key verification signing round
-    RotateKey(PublicKeyXOnly, model::BitcoinBlockHash),
+    DkgVerification(PublicKeyXOnly, model::BitcoinBlockRef),
+}
+
+impl std::fmt::Display for StateMachineId {
```
cylewitruk, line 51: Could probably have used `strum` for this, but I wasn't sure off the bat how to handle the formatting of the inner's, but if anyone else knows we could change it.

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,766 @@
...
+    fn is_expired(&self) -> bool {
+        match self.timeout {
+            None => false,
+            Some(timeout) => self.created_at.elapsed() > timeout,
+        }
+    }
+
+    /// Asserts that the [`StateMachine`] is in a state where it
+    /// can process messages.
+    fn assert_valid_state(&self) -> Result<(), Error> {
```
xoloki, line 318: Maybe rename this to `check_can_process_messages`?  `Success`, `Idle`, and `Signing` don't seem like "invalid" states.

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,766 @@
...
+        assert_allowed_msg_type!(state_machine, SignatureShareResponse);
+    }
+
+    #[test]
+    fn test_dkg_verification_state_machine() {
+        let mut setup = TestSetup::setup(2);
+        let sender1 = setup.sender(0);
+        let sender2 = setup.sender(1);
+        let mut state_machine = setup.state_machine;
+        let mut signer1 = setup.signers.pop().unwrap();
```
xoloki, line 675: Calling `pop` gives you the signers in reverse order, which was the cause of the issue we saw on Friday.  Since this is test code, why not just use array access here?

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,766 @@
...
+            panic!("expected nonce response, got {:?}", nonce_response2);
+        };
+
+        // Create signature share requests, populated with the nonce responses
+        // from the signers.
+        let sig_share_request = signature_share_request(
+            1,
+            1,
+            1,
+            vec![nonce_response2.clone(), nonce_response1.clone()],
```
xoloki, line 721: If you leave the `pop` call above, please comment here why the reverse ordering is necessary.

signer/src/dkg/wsts.rs
```diff
@@ -0,0 +1,47 @@
+//! Contains WSTS helper types and impls for use within the DKG module.
+
+/// A helper enum to represent the different types of WSTS messages.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum WstsNetMessageType {
```
xoloki, line 5: `wsts::net::Message` is already an `enum`, so why is this necessary?

signer/src/error.rs
```diff
@@ -34,26 +40,19 @@ pub enum Error {
...
 
-    /// No mock transaction was found after DKG successfully completed. Spending
-    /// a signer UTXO locked by the new aggregate key could not be verified.
-    #[error("no mock transaction found when attempting to sign")]
-    MissingMockTransaction,
+    /// The DKG verification state machine is in an end-state.
+    #[error("DKG verification state machine is in an end-state and cannot be used for the requested operation: {0}")]
+    DkgVerificationEnded(PublicKeyXOnly, Box<dkg::verification::State>),
 
     /// The rotate-key frost verification signing round failed for the aggregate
```
xoloki, line 47: Since you renamed the error text, maybe align the comment too?

xoloki (approved):
Looks good, modulo some comments and questions.

signer/src/transaction_signer.rs
```diff
@@ -1218,54 +1383,83 @@ where
...
+            // in the FROST state machine as well for it to properly follow
+            // the signing round.
+            if is_dkg_verification {
+                self.process_dkg_verification_message(
+                    *state_machine_id,
+                    sender,
+                    signer_id,
+                    outbound_message,
+                )
+                .await?;
```
djordon, line 1445: Ahh right. We do not receive our own messages in this case because the FROST state machine lives in the `TxSignerEventLoop` and not in `TxCoordinatorEventLoop`.

signer/src/dkg/testing.rs
```diff
@@ -0,0 +1,312 @@
...
+            eprintln!("!! message count assertions failed:");
+            for result in results {
+                eprintln!(" - {}", result);
+            }
+            false
+        }
+    }
+
+    /// Prints message statistics for this [`StateMachine`] to `stderr`.
+    pub fn print_message_stats(&self) {
```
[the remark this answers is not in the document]
cylewitruk: no, but very useful when you need it

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,766 @@
...
+        // Process all of the messages that we determined could be processed.
+        for msg in messages {
+            // Mark the message as processed so that even if it fails we don't
+            // try to process it again.
+            msg.mark_processed();
+
+            // Record the current state of the coordinator before processing the
+            // message. This is used below to determine if the coordinator has
+            // transitioned to a new state after processing the message.
+            let coordinator_state_pre_processing = self.coordinator.state.clone();
```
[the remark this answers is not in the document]
djordon: Doesn't it need to be inside the loop since the state can change mid-loop? Oh maybe not, maybe it can only change at the end of the loop since we are filtering on the messages that we care about for this state and we require 100% participation.

signer/src/transaction_signer.rs
```diff
@@ -1194,22 +1364,17 @@ where
...
             }
-            Some(OperationResult::SignError(error)) => {
-                tracing::warn!(
-                    ?msg,
-                    %error,
-                    "🔐 failed to complete DKG verification signing round"
-                );
-                self.dkg_verification_results.pop(&id);
+            dkg::verification::State::Error | dkg::verification::State::Expired => {
+                tracing::warn!("🔐 failed to complete DKG verification signing round");
```
matteojug, line 1368: nit: it could be helpful to log also the state here

matteojug (approved):

signer/src/dkg/testing.rs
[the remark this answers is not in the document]
matteojug: something something not solving the world problems in a 1.6k lines diff, but my main issue with this is that we actually have `sender` in `QueuedMessage` just for this, and that seems wrong.

signer/src/wsts_state_machine.rs
djordon: You can, see https://docs.rs/strum_macros/latest/strum_macros/derive.Display.html, but what is here is good too. So I'm okay keeping it as is.

signer/src/dkg/testing.rs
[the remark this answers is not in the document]
cylewitruk: I can remove it, i'll just stash it for if/when I need it again.

signer/src/dkg/verification.rs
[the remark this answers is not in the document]
matteojug: Yeah I think so: it can change, but if it does the `match result` will either break/return or (in the `None` arm) break if changed, so `coordinator_state_pre_processing` should always have the same value -- once it changes, the loop would terminate one way or another.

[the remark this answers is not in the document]
cylewitruk: Yeah it can move, that case will be accounted for.

signer/src/dkg/wsts.rs
cylewitruk: `wsts::net::Message` variants all have inner values, was the issue. I can't e.g. say `.get_messages(wsts::net::Message::NonceRequest)`.

signer/src/error.rs
cylewitruk: [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/verification.rs
cylewitruk: Removed pop: [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/verification.rs
cylewitruk: Ah, good to know. Changed it up a bit in [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/verification.rs
cylewitruk: Went with `assert_valid_state_for_processing`: [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/testing.rs
```diff
@@ -0,0 +1,312 @@
...
+        sign_iter_id,
+        message: vec![0; 5],
+        signature_type: SignatureType::Taproot(None),
+        nonce_responses,
+    })
+}
+
+/// Helpers for the [`StateMachine`] type.
+impl StateMachine {
+    /// Asserts the message counts for the given message type. Returns
```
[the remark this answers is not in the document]
cylewitruk: [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,766 @@
...
+            .process(&sig_share_response1)
+            .expect("should be able to process message");
+
+        // Process signer 2's signature share response with both signers.
+        signer1
+            .process(&sig_share_response2)
+            .expect("should be able to process message");
+        signer2
+            .process(&sig_share_response2)
+            .expect("should be able to process message");
```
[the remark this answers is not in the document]
cylewitruk: They were, probably leftovers from debugging stuff, removed in [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,766 @@
...
+        WstsNetMessageType::NonceResponse, WstsNetMessageType::SignatureShareResponse,
+    };
+
+    #[test]
+    fn test_initial_state() {
+        let signers = TestSetup::setup(5);
+        let state_machine = signers.state_machine;
+
+        assert_eq!(state_machine.signer_count(), 5);
+        assert!(matches!(state_machine.state, State::Idle));
```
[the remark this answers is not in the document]
cylewitruk: [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/verification.rs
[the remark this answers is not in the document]
cylewitruk: [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,766 @@
...
+    /// Processes all queued messages that can be processed given the current
+    /// state of the [`FrostCoordinator`].
+    fn process_queued_messages(&mut self) -> Result<bool, Error> {
+        let message_type_to_process = self.current_processable_message_type()?;
+
+        // Gets references to all pending messages of the given type that are
+        // currently stored. The returned messages are in arbitrary order, but
+        // this doesn't matter here since we're only processing the messages
+        // that the coordinator state machine can handle given its current
+        // state.
```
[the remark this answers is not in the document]
cylewitruk: [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,766 @@
...
+    ///   current state of the [`StateMachine`].
+    /// - Poll [`Self::state`] after a successfull call to this function to get
+    ///   the current state of the [`StateMachine`], which will be updated as
+    ///   messages are processed.
+    /// - If the instance is in an end-state, an error will be returned.
+    /// - Upon success, the resulting signature can be retrieved via
+    ///   [`Self::state`].
+    /// - Will process all eligible pending messages given the current state of
+    ///   the [`FrostCoordinator`].
+    pub fn process_message<M>(&mut self, sender: PublicKey, msg: M) -> Result<(), Error>
```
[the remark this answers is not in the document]
cylewitruk: Reworked it a bit in [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,766 @@
...
+    aggregate_key: secp256k1::XOnlyPublicKey,
+    /// The state machine that is being used to verify the signer.
+    pub(super) coordinator: FrostCoordinator,
+    /// WSTS messages that have been received from other signers. We keep them
+    /// in a [`HashMap`] where the keys are tuples of message types and sender
+    /// public keys, and the values are queued messages. This allows us to
+    /// manage messages by type and sender, ensuring that we can handle
+    /// out-of-order messages and process them correctly, while also
+    /// constraining each signer to one message per message type.
+    pub(super) wsts_messages: HashMap<WstsNetMessageType, Vec<QueuedMessage>>,
```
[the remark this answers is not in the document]
cylewitruk: [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,766 @@
...
+/// wall-clock time instead of Bitcoin block cadence.
+///
+/// NOTE: The signer currently stops messages from being processed if the
+/// received message doesn't match the current chain tip, so that needs to be
+/// changed for cross-block functionality to work.
+#[derive(Debug)]
+pub struct StateMachine {
+    /// The aggregate key that is being verified.
+    aggregate_key: secp256k1::XOnlyPublicKey,
+    /// The state machine that is being used to verify the signer.
```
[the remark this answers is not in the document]
cylewitruk: [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/testing.rs
```diff
@@ -0,0 +1,312 @@
...
+/// Convenience macro for asserting the message counts for a [`StateMachine`].
+///
+/// # Example
+/// ```rust
+/// assert_message_counts!(state_machine,
+///     NonceRequest => all: 0;
+///     NonceResponse => total: 0;
+///     SignatureShareRequest => pending: 0;
+///     SignatureShareResponse => total: 0, pending: 0;
+/// );
```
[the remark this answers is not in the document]
cylewitruk: Fixed it in [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3), now works as expected.

signer/src/dkg/testing.rs
```diff
@@ -0,0 +1,312 @@
...
+            .get(&message_type)
+            .unwrap_or(&vec![])
+            .iter()
+            .filter(|msg| !msg.processed)
+            .count()
+    }
+
+    /// Gets the number of buffered messages of the given type that are
+    /// currently stored in this [`StateMachine`].
+    #[cfg(test)]
```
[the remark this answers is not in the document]
cylewitruk: It's on the `testing` mod, was from me copy/pasting them back and forth before they ended up here permanently. Removed in [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/testing.rs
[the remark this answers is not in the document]
cylewitruk: Removed it all in [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/testing.rs
```diff
@@ -0,0 +1,312 @@
...
+        let total_count = self.total_message_count(message_type);
+        let pending_count = self.pending_message_count(message_type);
+
+        let mut results = Vec::new();
+
+        // Check the total counts.
+        if total_count != expected_total {
+            results.push(format!(
+                "expected {} total messages of type {:?}, got {}",
+                expected_total, message_type, total_count
```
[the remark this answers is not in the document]
cylewitruk: [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/testing.rs
```diff
@@ -0,0 +1,312 @@
...
+    pub fn sender(&self, index: usize) -> PublicKey {
+        self.state_machine
+            .coordinator
+            .get_config()
+            .signer_public_keys
+            .iter()
+            .map(|(_, point)| crate::keys::PublicKey::try_from(point))
+            .collect::<Result<Vec<_>, _>>()
+            .expect("failed to convert public keys")[index]
+    }
```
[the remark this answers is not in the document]
cylewitruk: Deleted this method in favor of another solution. It's in [75c78bc](https://github.com/stacks-network/sbtc/pull/1324/commits/75c78bc3c46868836857bc3750671ccbec4c83c3)

signer/src/dkg/testing.rs
```diff
@@ -0,0 +1,312 @@
...
+
+    /// Gets the number of buffered messages of the given type that are
+    /// currently stored in this [`StateMachine`].
+    #[cfg(test)]
+    pub fn total_message_count(&self, message_type: WstsNetMessageType) -> usize {
+        // The `_ as u32` should be safe here since we know that the number of
+        // signers is far less than `u32::MAX`, and each message is deduplicated
+        // by the sender's public key, which is also validated to be a valid
+        // member of the signer set.
```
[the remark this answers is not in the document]
cylewitruk: [4eed25e](https://github.com/stacks-network/sbtc/pull/1324/commits/4eed25e6ff519d6391efe1338f1db23249abef4d)

signer/src/dkg/verification.rs
```diff
@@ -0,0 +1,766 @@
...
+    #[error("the state machine has reached the message limit for message type {message_type:?} (expected: {expected}, actual: {actual})")]
+    MessageLimitExceeded {
+        /// The type of message that has reached the limit.
+        message_type: WstsNetMessageType,
+        /// The expected number of messages of this type that should be
+        /// received.
+        expected: u32,
+        /// The actual number of messages of this type that have been received.
+        actual: u32,
+    },
```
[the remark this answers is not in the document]
cylewitruk: [9109033](https://github.com/stacks-network/sbtc/pull/1324/commits/91090330b7ab5a734c5f3d3ca466799ed0692741)

signer/tests/integration/block_observer.rs
```diff
@@ -198,9 +198,10 @@ async fn load_latest_deposit_requests_persists_requests_from_past(blocks_ago: u6
 
     let waiting_fut = async {
         let db2 = db2.clone();
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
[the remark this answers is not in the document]
cylewitruk: [e604392](https://github.com/stacks-network/sbtc/pull/1324/commits/e604392aa40a4de7ded848944821c9c31932ad5a)

signer/src/dkg/testing.rs
```diff
@@ -0,0 +1,312 @@
...
+/// # Example
+/// ```rust
+/// assert_message_counts!(state_machine,
+///     NonceRequest => all: 0;
+///     NonceResponse => total: 0;
+///     SignatureShareRequest => pending: 0;
+///     SignatureShareResponse => total: 0, pending: 0;
+/// );
+/// ```
+macro_rules! assert_message_counts {
```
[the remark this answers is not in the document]
cylewitruk: Well, here's my long-winded rebuttal ;)

What you suggest is what I had before, but it quickly turns into this after `rustfmt` has its way with you:
```rust
state_machine
    .assert_message_counts(NonceRequest, 0, 0)
    .unwrap();
state_machine
    .assert_message_counts(NonceResponse, 0, 0)
    .unwrap();
state_machine
    .process_message(sender1, nonce_request)
    .unwrap();
state_machine
    .assert_message_counts(NonceRequest, 1, 0)
    .unwrap();
state_machine
    .assert_message_counts(NonceRequest, 0, 0)
    .unwrap();
```
Or you can do this with a `bool` return, but it's also imo very repetitive and detracting:
```rust
assert!(state_machine.check_message_counts(NonceRequest, 0, 0));
assert!(state_machine.check_message_counts(NonceResponse, 0, 0));
state_machine
    .process_message(sender1, nonce_request)
    .unwrap();
assert!(state_machine.check_message_counts(NonceRequest, 1, 0));
assert!(state_machine.check_message_counts(NonceResponse, 0, 0));
```
In both cases you need to either remember or look at the `check_message_counts()` signature to know what the `1`'s and `0`'s represent, versus the following which is clear, concise and saves my fingers and brain:

```rust
assert_message_counts!(state_machine, 
    NonceRequest => all: 0;
    NonceResponse => all: 0;
);
state_machine
    .process_message(sender1, nonce_request)
    .unwrap();
assert_message_counts!(state_machine, 
    NonceRequest => total: 1, pending: 0;
    NonceResponse => total: 0, pending: 0;
);
```

Another option was to `assert!(..)` within a method call, like the following (which I would have preferred over the above); but this gives you the file and line number for the panic itself, not the file/line in the actual test that failed:
```rust
// testing.rs
pub fn assert_state(state: State) {
    assert!(1 == 2);
}
// verification.rs
state_machine.assert_state(State::Idle);
```
```
--- STDERR:              signer dkg::verification::tests::panic_test ---
thread 'dkg::verification::tests::assert_state' panicked at signer/src/dkg/testing.rs:172:9:
panic
```
**Note:** `RUST_BACKTRACE=full` does not give you file/line info in the stack trace.

> Wrt assert_state above, here we are not relying on the macro to do something we couldn't do with normal rust code (IIUC).

Nope, I just got tired of typing and looking at `assert!(matches!(state_machine.state, ..))`. 
I did try to pull in the `assert_matches` crate once... don't know why the [official `assert_matches`](https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html) has been experimental so long... I'd be for adding our own `assert_matches!()` to the common `testing` lib though and using that instead (later).

signer/src/wsts_state_machine.rs
cylewitruk: Someone can boyscout it when they're passing by this code the next time :)

signer/src/wsts_state_machine.rs
```diff
@@ -45,7 +45,23 @@ pub enum StateMachineId {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            StateMachineId::Dkg(block_hash) => write!(f, "dkg({:?})", block_hash),
+            StateMachineId::BitcoinSign(sighash) => write!(f, "bitcoin-sign({})", sighash),
+            StateMachineId::DkgVerification(pubkey, block_hash) => {
+                write!(
+                    f,
+                    "dkg-verification(key={}, block_hash={}, block_height={})",
+                    pubkey, block_hash.block_hash, block_hash.block_height
```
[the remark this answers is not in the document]
cylewitruk: [9401721](https://github.com/stacks-network/sbtc/pull/1324/commits/9401721b2743e612669b1ba391e45308997cec45)

signer/src/wsts_state_machine.rs
djordon: Yeah, same for `dkg::verification::State`

signer/src/dkg/testing.rs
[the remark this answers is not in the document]
cylewitruk: Oh, one additional note on `assert!(matches!(A, B))` -- it will only tell you that the assertion failed since the assertion is on a boolean, so it won't print the values like in `assert_eq!(A, B)`.

djordon (approved):
The subsequent changes seem fine.

matteojug (approved):

signer/src/dkg/testing.rs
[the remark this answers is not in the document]
matteojug: I was thinking more of something like:
```
assert_message_counts(state_machine, vec![
    (NonceRequest, AssertCount::All(0)),
    (NonceResponse, AssertCount::Total(2))
    // or using a vec also for the counts to support multiple checks
);
```
and then assert inside the function. Uglier? Sure, but there's no magical formula you need to know to get the right format.
But yeah, the missing file/line number would be quite annoying.

PR #1333, opened 2025-02-10
feat: fall-back to current signer set + threshold from contract for stacks tx validation

cylewitruk:
## Description

Changes the tx-signer's `assert_valid_stacks_tx_sign_request()` method to use the current signer set public keys & threshold according to the following rule set. 

⚠️ This PR builds on-top of (_and includes_) #1315 and targets `main`, so changes from both PR's are in the diff.

With only #1315, we would be unable to submit a rotate-keys contract call if any of the signers were missing the dkg shares entry for the _current_ aggregate key. This PR adds a fallback mechanism for this case.

```rust
/// 1. First, we attempt to retrieve the information from the database.
/// 2. If the database does not contain the information, or we receive an
///    error from the database, we fall-back to the contract.
/// 3. If we couldn't get the information from the contract, but the
///    contract calls returned valid "not found" responses (i.e. `None`
///    here), then we're likely bootstrapping and fall-back to the
///    bootstrapping configuration.
```

## Changes

- Adds a new `StacksInteract` method for retrieving the `current-signature-threshold` from `sbtc-registry`.
- Updates the current `StacksInteract` method for retrieving the `current-signing-set` from `sbtc-registry` to return an `Option<>`, and `None` if the Stacks node returns a `404`.
- Updates `assert_valid_stacks_tx_sign_request()` to use a new method `get_current_signer_set_public_keys_and_threshold()` for retrieving the signer set public keys and signature threshold, which uses the above rule set.

## Testing Information

- Added new tests for the `StacksInteract` changes.
- All existing tests pass, including the new test from #1315.
- Ran through `devenv` with a deposit and everything looks good.

## Checklist:

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -11,7 +11,7 @@ x-common-vars:
   - &BITCOIN_RPC_USER devnet
   - &BITCOIN_RPC_PASS devnet
   - &MINE_INTERVAL ${MINE_INTERVAL:-1s}
-  - &MINE_INTERVAL_EPOCH25 ${MINE_INTERVAL_EPOCH25:-5s} # 5 second bitcoin block times in epoch 2.5
+  - &MINE_INTERVAL_EPOCH25 ${MINE_INTERVAL_EPOCH25:-1s} # 1 second bitcoin block times in epoch 2.5
   - &MINE_INTERVAL_EPOCH3 ${MINE_INTERVAL_EPOCH3:-15s} # 15 second bitcoin block times in epoch 3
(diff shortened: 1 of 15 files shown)


cylewitruk:
@djordon @matteojug do we want this functionality (without #1315, i.e. just the fallback logic)? If so, I can open a new PR with this not based on #1315, otherwise we can close this.

djordon:
Closing this, since we are not requiring 15 signers for the rotate-keys contract call, which this PR was needed for (from #1315). However, there are a few ideas in here, around using the stacks node for rotate keys information, that we should probably use going forward as well. We'll create a ticket fleshing that out.

Issue #478, opened 2024-09-02
[Feature]: Validate `reject-withdrawal-request` contract calls


djordon:
## 1. Description

This is a part of https://github.com/stacks-network/sbtc/issues/476.

### 1.1 Context & Purpose

The checks for validating `reject-withdrawal-request` contract call include the following is to create a report struct containing the key pieces of information about a withdrawal request. They include:
1. **The request exists**. Check whether the associated withdrawal request transaction is confirmed on the canonical stacks blockchain. Fail the withdrawal request if it is not on the canonical stacks blockchain.
2. **The de-pegging check**. Check whether the qualified request ID is in the `bitcoin_withdrawals_outputs` table, and that any of the associated `txid`s are confirmed on the canonical bitcoin blockchain. Fail the withdrawal request if such a transaction was found.
3. **The request is final**. Check that we've seen six new bitcoin blocks since observing the bitcoin anchor block associated with the Stacks block confirming the withdrawal request transaction. Fail the withdrawal request if we’ve observed less than 6 bitcoin blocks since the anchor block.
4. **The request is collectively rejected or expired.** Check if the request does not have enough votes or if we've observed 24 new bitcoin blocks since observing the bitcoin anchor block associated with the Stacks block confirming the withdrawal request transaction. Fail otherwise.

I think that that's it. All other things are implicitly handled by the design laid out in https://github.com/stacks-network/sbtc/issues/782.

## 2. Technical Details:

Validation of `reject-withdrawal-request` contract calls, happens inside the `RejectWithdrawalV1::validate` function. The task is to implement that function

### 2.1 Acceptance Criteria:
   - [ ] We have a properly implemented function for validating `reject-withdrawal-request` contract call sign requests.

## 3. Related Issues and Pull Requests (optional):
I think that this ticket is the last part of #255

PR #1336, opened 2025-02-10
feat: implement validatation of reject-withdrawal-request contract calls

xoloki:
## Description

Implement validation of `RejectWithdrawalV1`, which will be used when a coordinator asks signers to sign a stacks contract call to reject a withdrawal request.

Closes: #478 

## Changes
1. Implement `RejectWithdrawalV1::validate`
2. Add `signer/tests/integration/withdraw_reject.rs` which adds tests
3. Update `signer/tests/integration/setup.rs` `TestSweepSetup2`

## Testing Information
All newly added tests pass

## Checklist:


signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1066,118 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+        let Some(stacks_block) = stacks_chain_tip else {
+            eprintln!("no stacks block for request chain tip");
+            return Err(WithdrawalRejectErrorMsg::MissingStacksBlock.into_error(req_ctx, self));
+        };
+        let stacks_block_hash = stacks_block.block_hash;
+
+        eprintln!("get withdrawal signers");
+        let withdrawal_signers = &ctx
+            .get_storage()
+            .get_withdrawal_signers(self.request_id, &stacks_block_hash)
```
djordon, line 1096: Ahh, yeah one thing that needs to change here is the `RejectWithdrawalV1` itself. The `request_id` should be a `QualifiedRequestId` instead of a `u64` and we should probably rename the field too. Then you could use the block hash in the qualified ID here to get the votes.

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1066,118 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+            .map_err(|_| WithdrawalRejectErrorMsg::RequestMissing.into_error(req_ctx, self))?;
+        let signer_bitmap = BitArray::<[u8; 16]>::from(signer_votes);
+        eprintln!("signer_votes  {}", &signer_bitmap);
+        eprintln!("signer_bitmap {}", &self.signer_bitmap);
+
+        if signer_bitmap != self.signer_bitmap {
+            return Err(WithdrawalRejectErrorMsg::BitmapMismatch.into_error(req_ctx, self));
+        }
+
+        // TODO: how many votes are needed to be valid?
```
djordon, line 1179: I think you need `num_signers - threshold + 1` votes against in order to know that the withdrawal request will be fulfilled. And we should use the `withdrawal_signers` variable above for this, since those are the raw actual votes; `get_withdrawal_request_signer_votes` massages missing votes into votes against, which we don't want.

signer/src/stacks/contracts.rs
xoloki: Elsewhere in the code it looks like we name `QualifiedRequestId` fields simply as `id`, and accessor functions `qualified_id()`.  What name would you like to see?

signer/src/stacks/contracts.rs
xoloki: Should we still validate that the bitmap from `get_withdrawal_request_signer_votes` matches `self.signer_bitmap` like `AcceptWithdrawalV1` does?

djordon: Yes, we should. We need to do something for the votes that we are missing, so using `get_withdrawal_request_signer_votes` for creating the bitmap makes sense. But for making sure that we have enough actual votes against, we need to use that other query, the one in `get_withdrawal_signers`.

signer/src/stacks/contracts.rs
xoloki: I tried `qualified_request_id` and it's really verbose.  Any thoughts on `qualified_id` or just `id`?

djordon: I'd probably go with `qualified_id`.

xoloki: I went with `id`, since other code did so.  It seems like the goal was to avoid namespace clashes with the `qualified_id()` accessors.  I fixed it in the code, the protobufs, and the conversions.

signer/src/stacks/contracts.rs
xoloki: Ok, I'm now iterating over the withdrawal signers returned from `get_withdrawal_signers`, and summing up the `is_accepted` count.   Looks like I can get the threshold via ```
  Some(latest_dkg) = db.get_latest_encrypted_dkg_shares
  latest_dkg.signature_share_threshold
```

File: signer/src/stacks/contracts.rs [hunk 3]
```diff
@@ -987,11 +1067,110 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+        // 2. The double spend check. Check whether the qualified request ID is in the bitcoin_withdrawals_outputs table, and that any of the associated txids are confirmed on the canonical bitcoin blockchain. Fail the withdrawal request if such a transaction was found.
+
+        eprintln!("get withdrawal request report");
+        let Some(report) = ctx
+            .get_storage()
+            .get_withdrawal_request_report(
+                &req_ctx.chain_tip.block_hash,
+                &self.id.block_hash,
+                &self.id,
+                &req_ctx.origin,
```
djordon, line 1090: This is supposed to be current signers public key, which you can get from the `ctx.signer.public_key()`.

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1067,110 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+        else {
+            eprintln!("no bitcoin canonical chain tip");
+            return Err(WithdrawalRejectErrorMsg::NoBitcoinChainTip.into_error(req_ctx, self));
+        };
+
+        let Some(tip_bitcoin_block) = ctx.get_storage().get_bitcoin_block(&tip_block_hash).await?
+        else {
+            eprintln!("no bitcoin block for canonical chain tip");
+            return Err(WithdrawalRejectErrorMsg::MissingBitcoinBlock.into_error(req_ctx, self));
+        };
```
djordon, line 1132: Oh, no need for this, you should have what you need in the `ReqContext`.

signer/src/stacks/contracts.rs
xoloki: Neither TestContext nor ReqContext has a field named `signer`.

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1067,110 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+        // 1. The request exists. Check whether the associated withdrawal request transaction is confirmed on the canonical stacks blockchain. Fail the withdrawal request if it is not on the canonical stacks blockchain.
+        //
+        // 2. The double spend check. Check whether the qualified request ID is in the bitcoin_withdrawals_outputs table, and that any of the associated txids are confirmed on the canonical bitcoin blockchain. Fail the withdrawal request if such a transaction was found.
+
+        eprintln!("get withdrawal request report");
+        let Some(report) = ctx
+            .get_storage()
+            .get_withdrawal_request_report(
+                &req_ctx.chain_tip.block_hash,
+                &self.id.block_hash,
```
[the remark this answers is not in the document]
xoloki: Done

signer/src/stacks/contracts.rs
xoloki: The chain tip in `ReqContext` is the sweep block.  In order to finalize the withdrawal request, I have to mine 6 more blocks.

signer/src/stacks/contracts.rs
xoloki: Ok, I fixed the withdrawal request block height in `TestSweepSetup2`, it was using `Fake` data 😹 

I also hardwired the correct `threshold` into the `TestSweetSetup2` version of `ReqContext`. 

Now I just have to tell the test harness to vote against the withdrawal request, and this should be in a much better place.

xoloki:
Ok @djordon  I have it all working now, with minimal changes to `TestSweepSetup2`, squashed and rebased on latest main.

signer/src/stacks/contracts.rs
djordon: Oh yeah it's `ctx.config().signer.public_key()`

signer/src/stacks/contracts.rs
djordon: In our tests we often set it to that but generally, what is in there should be the chain tip.

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1070,144 @@ impl AsContractCall for RejectWithdrawalV1 {
...
     {
-        // TODO(255): Add validation implementation
+        eprintln!("check deployer");
+        if self.deployer != req_ctx.deployer {
+            return Err(WithdrawalRejectErrorMsg::DeployerMismatch.into_error(req_ctx, self));
+        }
+
+        // 1. The request exists. Check whether the associated withdrawal request transaction is confirmed on the canonical stacks blockchain. Fail the withdrawal request if it is not on the canonical stacks blockchain.
+        //
+        // 2. The double spend check. Check whether the qualified request ID is in the bitcoin_withdrawals_outputs table, and that any of the associated txids are confirmed on the canonical bitcoin blockchain. Fail the withdrawal request if such a transaction was found.
```
djordon, line 1083: Can you wrap these please. Feel free to locally use some tool to do it for you. In vscode I use the [rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) extension to do all of that for me.

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1070,113 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+        };
+
+        let Some(tip_bitcoin_block) = ctx.get_storage().get_bitcoin_block(&tip_block_hash).await?
+        else {
+            return Err(WithdrawalRejectErrorMsg::MissingBitcoinBlock.into_error(req_ctx, self));
+        };
+
+        let tip_block_height = tip_bitcoin_block.block_height;
+
+        //let tip_block_height = req_ctx.chain_tip.block_height;
```
djordon, line 1137: nit: remove.

signer/src/stacks/contracts.rs
xoloki: Done

signer/src/stacks/contracts.rs
xoloki: Would you prefer that I pass the `db` into `make_withdrawal_reject2` and set the real chain tip as above?

signer/src/stacks/contracts.rs
xoloki: Done

signer/src/stacks/contracts.rs
xoloki: Done

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1070,118 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+        let Some(report) = ctx
+            .get_storage()
+            .get_withdrawal_request_report(
+                &req_ctx.chain_tip.block_hash,
+                &stacks_chain_tip.block_hash,
+                &self.id,
+                &ctx.config().signer.public_key(),
+            )
+            .await
+            .map_err(|_| WithdrawalRejectErrorMsg::RequestMissing.into_error(req_ctx, self))?
```
djordon, line 1105: An error here should just mean that we're bailing.
```suggestion
            .await?
```

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1070,118 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+            .unwrap();
+        let Some(report) = ctx
+            .get_storage()
+            .get_withdrawal_request_report(
+                &req_ctx.chain_tip.block_hash,
+                &stacks_chain_tip.block_hash,
+                &self.id,
+                &ctx.config().signer.public_key(),
+            )
+            .await
```
djordon, line 1104: nit: can we break this out into two steps for readability? Maybe something like this:
```rust
let maybe_report = ctx
    .get_storage()
    .get_withdrawal_request_report(
        &req_ctx.chain_tip.block_hash,
        &stacks_chain_tip.block_hash,
        &self.id,
        &ctx.config().signer.public_key(),
    );

let Some(report) = maybe_report.await? else {
    return MyError
```

signer/src/stacks/contracts.rs
```diff
@@ -941,6 +970,60 @@ impl WithdrawalErrorMsg {
...
+    WithdrawalRequestFulfilled,
+    /// Withdrawal request unconfirmed
+    #[error("Withdrawal request unconfirmed")]
+    WithdrawalRequestUnconfirmed,
+    /// Withdrawal request not final
+    #[error("Withdrawal request not final")]
+    WithdrawalRequestNotFinal,
+    /// Withdrawal request wasn't rejected by (num_signers - threshold + 1) signers
+    #[error("Withdrawal request not rejected")]
+    WithdrawalRequestNotRejected,
```
djordon, line 1014: No need for Withdrawal in the name of the variants, since the enum is named `WithdrawalRejectErrorMsg`.

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1070,118 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+        else {
+            return Err(WithdrawalRejectErrorMsg::NoBitcoinChainTip.into_error(req_ctx, self));
+        };
+
+        let Some(tip_bitcoin_block) = ctx.get_storage().get_bitcoin_block(&tip_block_hash).await?
+        else {
+            return Err(WithdrawalRejectErrorMsg::MissingBitcoinBlock.into_error(req_ctx, self));
+        };
+
+        let tip_block_height = tip_bitcoin_block.block_height;
```
djordon, line 1142: These are all unnecessary. You should use the chain tip in the `ReqContext`.

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1070,118 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+                WithdrawalRejectErrorMsg::WithdrawalRequestNotFinal.into_error(req_ctx, self)
+            );
+        }
+
+        // 4. The request is collectively rejected or expired.
+        let signer_votes = ctx
+            .get_storage()
+            .get_withdrawal_request_signer_votes(&self.id, &req_ctx.aggregate_key)
+            .await
+            .map_err(|_| WithdrawalRejectErrorMsg::RequestMissing.into_error(req_ctx, self))?;
```
djordon, line 1155: You should just bail here.
```suggestion
            .await?;
```

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1070,118 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+
+        if signer_bitmap != self.signer_bitmap {
+            return Err(WithdrawalRejectErrorMsg::BitmapMismatch.into_error(req_ctx, self));
+        }
+
+        let withdrawal_signers = &ctx
+            .get_storage()
+            .get_withdrawal_signers(self.id.request_id, &self.id.block_hash)
+            .await
+            .map_err(|_| WithdrawalRejectErrorMsg::RequestMissing.into_error(req_ctx, self))?;
```
djordon, line 1166:
```suggestion
            .await?;
```

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1070,118 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+
+        let threshold = req_ctx.signatures_required;
+        let num_signers = ctx.state().current_signer_set().get_signers().len();
+        let reject_threshold = num_signers - threshold as usize + 1;
+
+        if rejected_count < reject_threshold {
+            return Err(
+                WithdrawalRejectErrorMsg::WithdrawalRequestNotRejected.into_error(req_ctx, self)
+            );
+        }
```
djordon, line 1183: This part is tricky since there are two reasons for why we would reject a withdrawal.
1. The withdrawal request has enough votes against.
2. The withdrawal request has expired. This happens after `WITHDRAWAL_BLOCKS_EXPIRY` blocks (the one from my PR [here](https://github.com/stacks-network/sbtc/pull/1352/files#diff-6cd449eb3c30449f70c0d11614a6b525227e18691745335775bf41baec8d7482R121-R126)) from the height of the `bitcoin_block_height` in the `WithdrawalRequestReport` (it's `block_height` on main but we're renaming it).

So we need to check for either condition here.

signer/tests/integration/setup.rs
```diff
@@ -945,10 +950,21 @@ impl TestSweepSetup2 {
...
 
+    pub fn reject_withdrawal_request(&mut self) {
+        for i in 0..self.signers.keys.len() {
+            self.withdrawal_request.signer_bitmap.replace(i, true);
+        }
+    }
+
     pub async fn store_withdrawal_request(&self, db: &PgStore) {
+        let block_height = match self.sweep_block_height() {
+            Some(height) => height,
```
djordon, line 961: I don't think we want this at the sweep block height, but maybe the deposit block height, if it's there.

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -0,0 +1,432 @@
...
+    // corresponds to how the signers voted.
+    setup.reject_withdrawal_request();
+    setup.store_withdrawal_request(&db).await;
+    setup.store_withdrawal_decisions(&db).await;
+
+    // Generate the transaction and corresponding request context.
+    let (reject_withdrawal_tx, req_ctx) = make_withdrawal_reject2(&setup);
+
+    // Generate more blocks then backfill the DB
+    let mut hashes = faucet.generate_blocks(6);
```
djordon, line 174: Ahh, I see why you want to get the chain tip during validation. Instead you should give `make_withdrawal_reject2` the chain tip.

djordon (commented):
I haven't finished looking through the tests but so far looking good.

signer/src/stacks/contracts.rs
djordon: Yeah, that's fine. Or pass it the chain tip.

signer/src/stacks/contracts.rs
xoloki: Done

signer/src/stacks/contracts.rs
xoloki: Done

signer/src/stacks/contracts.rs
xoloki: Done

xoloki: (won't resolve this until you look at it)

signer/src/stacks/contracts.rs
xoloki: Done

signer/src/stacks/contracts.rs
xoloki: Done.  I had to leave the `await` as part of the first step to avoid a use-of-temporary compiler error though.

signer/src/stacks/contracts.rs
xoloki: Done

signer/src/stacks/contracts.rs
xoloki: Done

signer/tests/integration/setup.rs
xoloki: `TestSweetSetup` uses sweep block height, but `TestSweepSetup2` was using `Fake` data, so I changed `TestSweepSetup2` to use the sweep block height if available.

There's a param to include the withdrawal request:```
pub fn submit_sweep_tx(&mut self, rpc: &Client, faucet: &Faucet, with_withdrawals: bool)
```

File: signer/tests/integration/withdrawal_reject.rs (hunk 16 above)
Comment (xoloki) [reply signer/tests/integration/withdrawal_reject.rs]: I passed it `db` so the tests which used `make_withdrawal_reject2` wouldn't have to all find the chain tip.

File: signer/tests/integration/setup.rs (hunk 15 above)
Comment (djordon) [reply signer/tests/integration/setup.rs:961]: Oh it does? Well, just a heads up, I proposed some edits to this struct as well in https://github.com/stacks-network/sbtc/pull/1359.

Comment (djordon) [reply signer/tests/integration/setup.rs:961]: If my PR get's merged first, I don't mind updating this one because of those changes.

File: signer/tests/integration/withdrawal_reject.rs [hunk 17]
```diff
@@ -0,0 +1,453 @@
...
+    backfill_bitcoin_blocks(&db, rpc, &last).await;
+
+    let validation_result = reject_withdrawal_tx.validate(&ctx, &req_ctx).await;
+    match validation_result.unwrap_err() {
+        Error::WithdrawalRejectValidation(ref err) => {
+            assert_eq!(err.error, WithdrawalRejectErrorMsg::RequestNotFinal)
+        }
+        err => panic!("unexpected error during validation {err}"),
+    }
+
```
djordon, line 372: After this we should generate one more block and check that things validate then.

djordon (commented):
We need to make add another check here, specifically, the `reject-withdrawal-request` transaction is valid if the request is expired or collectively rejected. We need to add the expired check, which is check to see if 24 or more blocks have arrived after the `bitcoin_block_height` of the withdrawal request. That field is also in the withdrawal request report.

And we need a few more tests here. In general, we should aim for a test of each of the different validation conditions that we check for, so basically one test per `WithdrawalRejectErrorMsg` variant. That might seem like a lot, but I believe that there are several error variants that we can remove after some of the recent edits; they're currently unused. And we don't need to add a test for the `MissingStacksBlock` at all, I think we can remove that variant in another PR.

signer/src/stacks/contracts.rs
djordon: I don't think that this one is done.

xoloki: Oh sorry I hadn't pushed when I commented.  Now it's:

https://github.com/stacks-network/sbtc/blob/478-validate-reject-withdrawal-request/signer/src/stacks/contracts.rs#L1162

signer/tests/integration/withdrawal_reject.rs
xoloki: This test is just happy_path with one less block, but seems reasonable to mine the extra block and make really sure that it now passes.

xoloki:
> We need to make add another check here, specifically, the `reject-withdrawal-request` transaction is valid if the request is expired or collectively rejected. We need to add the expired check, which is check to see if 24 or more blocks have arrived after the `bitcoin_block_height` of the withdrawal request. That field is also in the withdrawal request report.
> 

I actually had that test, just hadn't pushed my latest code.  It's pushed now and the test is here:

https://github.com/stacks-network/sbtc/blob/478-validate-reject-withdrawal-request/signer/tests/integration/withdrawal_reject.rs#L195

> And we need a few more tests here. In general, we should aim for a test of each of the different validation conditions that we check for, so basically one test per `WithdrawalRejectErrorMsg` variant. That might seem like a lot, but I believe that there are several error variants that we can remove after some of the recent edits; they're currently unused. And we don't need to add a test for the `MissingStacksBlock` at all, I think we can remove that variant in another PR.

Will do.  I was originally trying to map every failure to a variant in `WithdrawalRejectErrorMsg`, but as you pointed out that would make the logs less understandable.

signer/src/stacks/contracts.rs
xoloki: Looking back I don't think this check is sufficient.

xoloki: The logic is that if we negate either of these two conditions, that is sufficient for a valid rejection.  It's equivalent to:

    if !(blocks_observed >= WITHDRAWAL_BLOCKS_EXPIRY || rejected_count >= reject_threshold) {

But what if *both* conditions are negated?  Is it still a valid rejection if we have a threshold but it's expired?  I don't think so.

Maybe what we need is

    if !(blocks_observed >= WITHDRAWAL_BLOCKS_EXPIRY ^ rejected_count >= reject_threshold) {

xoloki: Ok I split the expired test into two: one with a threshold of rejection votes and one without.  Both tests pass.  We already test the case where we aren't expired but we are rejected (that's happy path).  But now we also need a not expired AND not rejected test.  Adding that now.

xoloki: Ok test added.  I think we're good now?

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -0,0 +1,453 @@
...
+        .unwrap();
+
+    let tip_block_height = tip_bitcoin_block.block_height;
+
+    // This is what the current signer thinks is the state of things.
+    let req_ctx = ReqContext {
+        chain_tip: BitcoinBlockRef {
+            block_hash: tip_block_hash,
+            block_height: tip_block_height,
+        },
```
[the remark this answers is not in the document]
xoloki: Done.  So much cleaner!

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -0,0 +1,453 @@
...
+    let mut rng = rand::rngs::StdRng::seed_from_u64(51);
+    let (rpc, faucet) = regtest::initialize_blockchain();
+
+    let amount = 1_000_000;
+    let test_signer_set = TestSignerSet::new(&mut rng);
+    let deposit_amounts = DepositAmounts { amount, max_fee: amount / 2 };
+    let mut setup =
+        TestSweepSetup2::new_setup(test_signer_set.clone(), &faucet, &[deposit_amounts]);
+
+    setup.submit_sweep_tx(rpc, faucet, true);
```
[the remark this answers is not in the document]
xoloki: So to be clear, for these tests we always want to submit the deposit transaction to bitcoin; in some tests we want to submit/confirm the withdrawal to stacks but not bitcoin (happy path), submit/notconfirm to stacks (RequestUnconfirmed), submit/confirm to stacks and submit/confirm to bitcoin (RequestFulfilled).  What about submit/confirm to stacks but submit/notconfirm to bitcoin?  It's not really fulfilled, but we can't exactly reject it can we?

signer/tests/integration/withdrawal_reject.rs
xoloki: And of course it's not passing once I add the block and expect validation :P  Not sure if it's test fixture or the validation is wrong.

xoloki: Oh lol the problem is that I reused the `req_ctx`, which means that the `chain_tip` didn't change even though I mined an extra block ⛓️ 

I knew there was a reason to get the chain tip inside `validate`.  Though I guess if we can't trust `req_ctx` we have bigger problems.

Fixed now.

xoloki:
> > And we need a few more tests here. In general, we should aim for a test of each of the different validation conditions that we check for, so basically one test per `WithdrawalRejectErrorMsg` variant. 
> 
> Will do. I was originally trying to map every failure to a variant in `WithdrawalRejectErrorMsg`, but as you pointed out that would make the logs less understandable.

Ok, I removed all the unused variants.  Looks like we need tests for `RequestFulfilled` and `RequestUnconfirmed` but I'll wait to add those until we reach consensus on 

https://github.com/stacks-network/sbtc/pull/1336#discussion_r1957190815

signer/src/stacks/contracts.rs
djordon: After thinking about this some more, we I propose that we only care about whether the request is expired.

xoloki: That's interesting.  Certainly if we are expired, we should fail validation regardless of how many rejection votes we've received.  But if we aren't expired, but don't have a threshold of rejection votes, shouldn't the rejection request fail validation?

I'll wait to see your logic in your ticket.

djordon: The ticket is here https://github.com/stacks-network/sbtc/issues/1363.

Basically, I think that we should fail validation if the `reject-withdrawal-request` contract call if the withdrawal request has not expired, regardless of the votes against.

Alternatively, we might want to allow for validation of the `reject-withdrawal-request` contract call if we receive a consensus number of votes against or something. But we'd also need to modify how we bitcoin validation as well, but we need to do that regardless.

signer/tests/integration/withdrawal_reject.rs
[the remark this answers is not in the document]
djordon: > So to be clear, for these tests we always want to submit the deposit transaction to bitcoin

Well, we don't need to, but the `TestSweepSetup2` struct does and it's nice to use. And it doesn't hurt that much if we have a deposit lying around on chain.

> in some tests we want to submit/confirm the withdrawal to stacks but not bitcoin (happy path), submit/notconfirm to stacks (RequestUnconfirmed), submit/confirm to stacks and submit/confirm to bitcoin (RequestFulfilled). What about submit/confirm to stacks but submit/notconfirm to bitcoin? It's not really fulfilled, but we can't exactly reject it can we?

Oh the "submit/confirm on stacks" should be a check. We end up burning STX in fees if we allow it. Basically, we should not approve a sign request for a `reject-withdrawal-request` contract call if one has already been confirmed on stacks. There aren't too many helper functions for checking that.

We need to add the same check to validation of `accept-withdrawal-request` contract calls. I'll take that on tonight, the check there will be identical to the one needed here so you could use it.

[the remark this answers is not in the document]
djordon: I created a branch to make this check easy in this branch https://github.com/stacks-network/sbtc/tree/add-function-smart-contract-withdrawal-status. The function that you want is `StacksInteract::is_withdrawal_completed`. It makes testing a little uglier though, since we need to mock out stacks-core.

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1056,98 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+            }
+        }
+
+        let threshold = req_ctx.signatures_required;
+        let num_signers = ctx.state().current_signer_set().get_signers().len();
+        let reject_threshold = num_signers - threshold as usize + 1;
+
+        if blocks_observed < WITHDRAWAL_BLOCKS_EXPIRY && rejected_count < reject_threshold {
+            return Err(WithdrawalRejectErrorMsg::RequestNotRejected.into_error(req_ctx, self));
+        }
```
djordon, line 1149: If we go with what is in https://github.com/stacks-network/sbtc/issues/1363, we can simplify this to just, and get rid of the other stuff around vote counting.
```rust
if blocks_observed < WITHDRAWAL_BLOCKS_EXPIRY {
    return Err(WithdrawalRejectErrorMsg::RequestNotFinal.into_error(req_ctx, self));
}
```

signer/src/stacks/contracts.rs
```diff
@@ -987,11 +1056,98 @@ impl AsContractCall for RejectWithdrawalV1 {
...
+        //    the bitcoin anchor block associated with the Stacks block confirming the withdrawal
+        //    request transaction. Fail the withdrawal request if we’ve observed less than 6
+        //    bitcoin blocks since the anchor block.
+
+        let request_block_height = report.bitcoin_block_height;
+        let blocks_observed = req_ctx.chain_tip.block_height - request_block_height;
+
+        if blocks_observed < 6 {
+            return Err(WithdrawalRejectErrorMsg::RequestNotFinal.into_error(req_ctx, self));
+        }
```
djordon, line 1118: In https://github.com/stacks-network/sbtc/issues/1363 we only care about `WITHDRAWAL_BLOCKS_EXPIRY` being observed so we can remove this as well.

signer/tests/integration/withdrawal_reject.rs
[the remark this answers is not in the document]
djordon: > What about submit/confirm to stacks but submit/notconfirm to bitcoin? It's not really fulfilled, but we can't exactly reject it can we?

I think we should do this check in a separate PR. That work can depend on the work in https://github.com/stacks-network/sbtc/tree/add-function-smart-contract-withdrawal-status, or maybe we go a different route and write some queries. But yeah, let's not do this here.

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -0,0 +1,677 @@
...
+
+    // Normal: the request and how the signers voted needs to be added to
+    // the database. Here the bitmap in the withdrawal request object
+    // corresponds to how the signers voted.
+    setup.reject_withdrawal_request();
+    setup.store_withdrawal_request(&db).await;
+    setup.store_withdrawal_decisions(&db).await;
+
+    // Generate an insufficient number of blocks then backfill the DB
+    let mut hashes = faucet.generate_blocks(5);
```
djordon, line 572: For this test, after we update the validation logic to reflect the first proposed change in section 3.1 of https://github.com/stacks-network/sbtc/issues/1363, we just need to change the `5` here to `WITHDRAWAL_BLOCKS_EXPIRY - 1`.

djordon (commented):
Okay, I think we can simplify the implementation here a bit by adopting the proposed changes in https://github.com/stacks-network/sbtc/issues/1363. Specifically, we don't care about counting votes, we only care about observing `WITHDRAWAL_BLOCKS_EXPIRY` blocks after observing the bitcoin block height of the withdrawal request, we still care about the deployer and bitmaps matching.

The other remaining changes are:
1. Check that there are no withdrawals matching this withdrawal request (the one that we are about to reject) in the mempool.
2. Check that we haven't responded to this withdrawal request already on Stacks with either an accept or reject.

We should do these two other checks in another PR.

signer/src/stacks/contracts.rs
xoloki: Done

signer/src/stacks/contracts.rs
xoloki: Done.

signer/tests/integration/withdrawal_reject.rs
xoloki: Done.

signer/src/stacks/contracts.rs
xoloki: Done.

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -0,0 +1,459 @@
...
+    // Generate the transaction and corresponding request context.
+    let (reject_withdrawal_tx, req_ctx) = make_withdrawal_reject2(&setup, &db).await;
+
+    let validate_future = reject_withdrawal_tx.validate(&ctx, &req_ctx);
+    match validate_future.await.unwrap_err() {
+        Error::WithdrawalRejectValidation(ref err) => {
+            assert_eq!(err.error, WithdrawalRejectErrorMsg::RequestNotFinal)
+        }
+        err => panic!("unexpected error during validation {err}"),
+    }
```
djordon, line 253: Although this is covered by the other test, couldn't hurt to generate one more block and see that everything now passes validation.

djordon (commented):
Looks good, just one change to one of the tests

signer/tests/integration/withdrawal_reject.rs
xoloki: We had that before, but it got removed with the churn.

xoloki: Done

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -0,0 +1,500 @@
...
+    setup.store_dkg_shares(&db).await;
+
+    // Normal: the request and how the signers voted needs to be added to
+    // the database. Here the bitmap in the withdrawal request object
+    // corresponds to how the signers voted.
+    setup.store_withdrawal_request(&db).await;
+    setup.store_withdrawal_decisions(&db).await;
+
+    // Generate more blocks then backfill the DB
+    let mut hashes = faucet.generate_blocks(WITHDRAWAL_BLOCKS_EXPIRY - 2);
```
djordon, line 255: Why not `WITHDRAWAL_BLOCKS_EXPIRY - 1` here?

djordon (approved):
Looks good 🚀

signer/tests/integration/withdrawal_reject.rs
xoloki: Some of the recent changes to `TestSweepSetup2` probably mined an extra block.

Issue #1326, opened 2025-02-10
[Feature]: Implement query that fetches swept withdrawals


djordon:
# Feature - Implement query that fetches swept withdrawals

## 1. Description

Once we have swept out funds after receiving a withdrawal request, we need burn the sBTC on stacks. Doing that requires us to retrieve all such requests that have been swept but require an `accept-withdrawal-request` contract call.

### 1.1 Context & Purpose

The happy path for a withdrawal requests goes like this:
1. A user submits a transaction that makes a `initiate-withdrawal-request` contract call and that transaction gets confirmed.
2. The signers receive work of the transaction through the grapevine, and sweep out the funds.
3. Once the sweep has been confirmed, the signers submit a `accept-withdrawal-request` for the swept funds.

This task is for implementing the code that fetches withdrawal request for step (3).

## 2. Technical Details:

This is probably best done by implementing the `DbRead::get_swept_withdrawal_requests` query. That query should read from the `bitcoin_withdrawals_outputs`, `bitcoin_transactions`, and `withdrawal_accept_events` tables, maybe a few of others too. The results should filter requests that do not have a row in the `withdrawal_accept_events` table that are associated with transactions on the canonical stacks blockchain.

### 2.1 Acceptance Criteria:
   - [ ] We have a query that fetches withdrawal requests that have been swept but don't have an associated `accept-withdrawal-request` contract call that has been confirmed on the canonical stacks blockchain.

PR #1339, opened 2025-02-11
[feat] query for swept withdrawals

MCJOHN974:
## Description

Implements query for swept withdrawals

Closes: #1326 

## Changes

Implemented `get_swept_withdrawal_requests` and added tests for it.

## Testing Information

Added 4 new tests which can be run as follows:

```
cargo test --package signer --test integration -- postgres::get_swept_withdrawal_requests_returns_swept_withdrawal_requests --exact --show-output
cargo test --package signer --test integration -- postgres::get_swept_withdrawal_requests_does_not_return_unswept_withdrawal_requests --exact --show-output
cargo test --package signer --test integration -- postgres::get_swept_withdrawal_requests_does_not_return_withdrawal_requests_with_responses --exact --show-output
cargo test --package signer --test integration -- postgres::get_swept_withdrawal_requests_response_tx_reorged --exact --show-output
```

## Checklist:

### DIFF
--- a/signer/tests/integration/postgres.rs
+++ b/signer/tests/integration/postgres.rs
@@ -2145,13 +2145,14 @@ async fn get_swept_withdrawal_requests_returns_swept_withdrawal_requests() {
     let test_data = TestData::generate(&mut rng, &signer_set, &test_params);
     test_data.write_to(&db).await;
 
-    let bitcoin_tip = db.get_bitcoin_canonical_chain_tip().await.unwrap().unwrap();
-    let bitcoin_tip_height = db
-        .get_bitcoin_block(&bitcoin_tip)
+    let bitcoin_tip_ref = db
+        .get_bitcoin_canonical_chain_tip_ref()
         .await
         .unwrap()
-        .unwrap()
-        .block_height;
+        .unwrap();
+    let bitcoin_tip = bitcoin_tip_ref.block_hash;
+    let bitcoin_tip_height = bitcoin_tip_ref.block_height;
+
     let stacks_tip = db
         .get_stacks_chain_tip(&bitcoin_tip)
         .await
@@ -2220,7 +2221,6 @@ async fn get_swept_withdrawal_requests_returns_swept_withdrawal_requests() {
 
     // There should only be one request in the database and it has a sweep
     // trasnaction so the length should be 1.
-    let context_window = 20;
     let mut requests = db
         .get_swept_withdrawal_requests(&bitcoin_block.block_hash, context_window)
         .await


signer/src/storage/postgres.rs
```diff
@@ -1921,13 +1952,94 @@ impl super::DbRead for PgStore {
...
     ) -> Result<Vec<model::SweptWithdrawalRequest>, Error> {
-        // TODO: This can use a similar query to
-        // `get_swept_deposit_requests()`, but using withdrawal tables instead
-        // of deposit.
-        unimplemented!()
+        // The following tests define the criteria for this query:
+        // - [X] get_swept_deposit_requests_returns_swept_deposit_requests
+        // - [X] get_swept_deposit_requests_does_not_return_unswept_deposit_requests
+        // - [X] get_swept_deposit_requests_does_not_return_deposit_requests_with_responses
+        // - [X] get_swept_deposit_requests_response_tx_reorged
```
djordon, line 1962: You'll want withdrawal analogs for all of these integration tests.

signer/src/storage/postgres.rs
```diff
@@ -1921,13 +1952,94 @@ impl super::DbRead for PgStore {
...
+
+        let Some(stacks_chain_tip) = self.get_stacks_chain_tip(chain_tip).await? else {
+            return Ok(Vec::new());
+        };
```
djordon, line 1966: You should probably take the stacks chain tip as an input into this function.

signer/tests/integration/postgres.rs
```diff
@@ -2121,6 +2123,110 @@ async fn get_swept_deposit_requests_returns_swept_deposit_requests() {
...
+        .await
+        .unwrap();
+
+
+    
+    let context_window = 20;
+    let mut requests = db
+        .get_swept_withdrawal_requests(&bitcoin_block.block_hash, context_window)
+        .await
+        .unwrap();
```
djordon, line 2195: We probably want to call this before everything is inserted and show that it doesn't retrieve the withdrawal request unless we have the sweep information.

signer/src/storage/postgres.rs
djordon: We can keep it as is for now, it's an optimization. We can update them both some time later.

djordon:
> Sorry for such a big PR, but it is mostly tests which can be reviewed 1 by 1 so I think it is okish

Yeah it's fine, we're good. This is mostly tests.

djordon (commented):
Largely looks good. But I think we should have more of a blockchain in these tests.

Also, I might have missed something, but it's not clear where the reorg occurs in the `get_swept_withdrawal_requests_response_tx_reorged` test. Are we sure that there is a reorg there?

signer/tests/integration/postgres.rs
```diff
@@ -2122,6 +2123,213 @@ async fn get_swept_deposit_requests_returns_swept_deposit_requests() {
...
+        num_withdraw_requests_per_block: 0,
+        num_signers_per_request: num_signers,
+        consecutive_blocks: false,
+    };
+
+    let signer_set = testing::wsts::generate_signer_set_public_keys(&mut rng, num_signers);
+    let test_data = TestData::generate(&mut rng, &signer_set, &test_params);
+    test_data.write_to(&db).await;
+
+    let bitcoin_tip = db.get_bitcoin_canonical_chain_tip().await.unwrap().unwrap();
```
djordon, line 2148: If you use `get_bitcoin_canonical_chain_tip_ref` you get an object containing both the block hash and the block height.

signer/tests/integration/postgres.rs
```diff
@@ -2122,6 +2123,159 @@ async fn get_swept_deposit_requests_returns_swept_deposit_requests() {
...
+        block_hash: fake::Faker.fake_with_rng(&mut rng),
+        block_height: 1,
+        parent_hash: fake::Faker.fake_with_rng(&mut rng),
+    };
+    let stacks_block = model::StacksBlock {
+        block_hash: fake::Faker.fake_with_rng(&mut rng),
+        block_height: 1,
+        parent_hash: fake::Faker.fake_with_rng(&mut rng),
+        bitcoin_anchor: bitcoin_block.block_hash,
+    };
```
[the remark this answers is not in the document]
djordon: > I mean, are we ok just submitting some random blocks to the db

Yes we're okay with that. We do so in a lot of our tests. Moreover, we create a new database for each of our tests so that there is no interference.

> or we want to switch from this manual writing to db to some helper functions?

What I had in mind was generating a blockchain, placing the withdrawal request somewhere on the chain, and placing maybe any response response transactions (either accept or reject) somewhere later on the chain. Then run running the query. This way you can test that we are properly walking the chain.

The query looks fine, but maybe we change it later in some way that breaks things and we want our tests to catch it. That's why I want the chain to have more than one block in them. And this is something we need to do better in many of our tests. Lot's of them create "small" or "simple" block chains and test against that. It's something we'll need to fix eventually.

signer/tests/integration/postgres.rs
```diff
@@ -2471,6 +2713,117 @@ async fn get_swept_deposit_requests_response_tx_reorged() {
...
+
+    db.write_withdrawal_accept_event(&event).await.unwrap();
+
+    // since this withdrawal was accepted get_swept_withdrawal_requests should return nothing
+    let context_window = 20;
+    let requests = db
+        .get_swept_withdrawal_requests(&new_block.block_hash, context_window)
+        .await
+        .unwrap();
+    assert_eq!(requests.len(), 0);
```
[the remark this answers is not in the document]
djordon: Ahhh, I missed that the withdrawal accept event was on the new chain tip. And it was right there 😞.

signer/tests/integration/postgres.rs
```diff
@@ -2122,6 +2123,213 @@ async fn get_swept_deposit_requests_returns_swept_deposit_requests() {
...
+    let mut rng = rand::rngs::StdRng::seed_from_u64(16);
+
+    let num_signers = 3;
+    let test_params = testing::storage::model::Params {
+        num_bitcoin_blocks: 10,
+        num_stacks_blocks_per_bitcoin_block: 1,
+        num_deposit_requests_per_block: 0,
+        num_withdraw_requests_per_block: 0,
+        num_signers_per_request: num_signers,
+        consecutive_blocks: false,
```
djordon, line 2141: To make testing with these blocks easier, we might want to change this to `true` and then use the blocks generated instead of building on top of them.

signer/tests/integration/postgres.rs
```diff
@@ -2122,6 +2123,213 @@ async fn get_swept_deposit_requests_returns_swept_deposit_requests() {
...
+        .unwrap();
+    db.write_transaction(&sweep_tx_model).await.unwrap();
+    db.write_bitcoin_transaction(&sweep_tx_ref).await.unwrap();
+    db.write_bitcoin_withdrawals_outputs(&[swept_output.clone()])
+        .await
+        .unwrap();
+
+    // There should only be one request in the database and it has a sweep
+    // trasnaction so the length should be 1.
+    let context_window = 20;
```
djordon, line 2223: nit: this was already defined above.

djordon (approved):
I think this is basically done. The tests right now do minimal walking of the blockchain. This is okay, a lot of our tests do that for one reason or another. If you have the time we should try to make the test setup a little more challenging. But feel free to skip if you are confident that the tests capture everything already.

Issue #1329, opened 2025-02-10
[Feature]: Read withdrawal requests from the right place


djordon:
# Feature - Read withdrawal requests from the right place

## 1. Description

We write the withdrawal events to the `withdrawal_create_events` table, but many of our queries read from the `withdrawal_requests`. This ticket is about either writing the withdrawal requests to the `withdrawal_requests` table or reading withdrawal request data from the `withdrawal_create_events` table.

### 1.1 Context & Purpose

When the `withdrawal_requests` table was created, we did not know what the events were going to look like. When we know what the tables would look like, we just created a new table instead of modifying the old one, with the intention of eventually sorting this mess out.

## 2. Technical Details:

If write to the `withdrawal_requests` table, we should remove the foreign key constraint on the `stacks_blocks` table and add columns from the `withdrawal_create_events` table. While if we read from the `withdrawal_create_events` we will want to add a unique constraint on the `(request_id, block_hash)` columns.

### 2.1 Acceptance Criteria:
   - [ ] Our queries read from the place that `initiate-withdrawal-request` events get written to.

PR #1341, opened 2025-02-11
[chore] consolidate withdrawals tables

Jiloc:
## Description

Closes: #1329

## Changes
- drop `sbtc_signer.withdrawal_create_events` 
- drop FK to stacks_blocks from `withdrawal_requests`
- add missing `block_height` column to `withdrawal_requests`  ( received from the event)
- new withdrawal events received in the `new-block` handler are not saved into `withdrawal_requests`
- adapted queries
- fix `get_withdrawal_requests` in_memory implementation to include the withdrawals in the anchored to the block at window boundary.

## Testing Information
- updated unit and integration tests
- migrations results are tested by `writing_withdrawal_requests_postgres`
- needed to change random seed for `assert_should_store_decisions_received_from_other_signers` - with the current changes, the previous seed was generating 25 valid deposits, instead of the 30 expected, causing the test to fail
- 

## Checklist:

### DIFF
--- a/signer/migrations/0003__create_tables.sql
+++ b/signer/migrations/0003__create_tables.sql
@@ -152,6 +152,8 @@ CREATE TABLE sbtc_signer.completed_deposit_events (
     created_at          TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL
 );
 
+-- DEPRECATED: dropped in migration 0013__consolidate_withdrawal_tables.
+-- Use `sbtc_signer.withdrawal_requests` instead.
 CREATE TABLE sbtc_signer.withdrawal_create_events (
     id           BIGSERIAL PRIMARY KEY,
     txid         BYTEA   NOT NULL,
--- a/signer/migrations/0003__create_tables.sql
+++ b/signer/migrations/0003__create_tables.sql
@@ -152,8 +152,7 @@ CREATE TABLE sbtc_signer.completed_deposit_events (
     created_at          TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL
 );
 
--- DEPRECATED: dropped in migration 0013__consolidate_withdrawal_tables.
--- Use `sbtc_signer.withdrawal_requests` instead.
+-- DEPRECATED: Use `sbtc_signer.withdrawal_requests` instead.
 CREATE TABLE sbtc_signer.withdrawal_create_events (
     id           BIGSERIAL PRIMARY KEY,
     txid         BYTEA   NOT NULL,
--- a/signer/migrations/0013__consolidate_withdrawal_tables.sql
+++ b/signer/migrations/0013__consolidate_withdrawal_tables.sql
@@ -1,10 +1,11 @@
 
-DROP TABLE sbtc_signer.withdrawal_create_events;
-
--- Remove the FK `block_hash` REFERENCES `stacks_blocks(block_hash)` constraint.
+-- We cannot guarantee that we have the stacks block in the database by the time we
+-- have received a withdrawal request and are ready to write it into the database.
+-- So we need to drop the constraint.
 ALTER TABLE sbtc_signer.withdrawal_requests
     DROP CONSTRAINT withdrawal_requests_block_hash_fkey;
 
--- Add the new column to the `withdrawal_requests` table.
+-- The block height of the bitcoin blockchain when the stacks
+-- transaction that emitted this event was executed.
 ALTER TABLE sbtc_signer.withdrawal_requests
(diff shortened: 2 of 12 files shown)


signer/migrations/0013__consolidate_withdrawal_tables.sql
```diff
@@ -0,0 +1,10 @@
+
+DROP TABLE sbtc_signer.withdrawal_create_events;
```
djordon, line 2: We probably shouldn't drop this quite yet. There is a risk that someone has constructed a withdrawal request and has their funds locked up and we have a row for it in our database. It's very low probability, but we can check to make sure and remove it in any follow-up release.

signer/src/testing/transaction_coordinator.rs
```diff
@@ -185,6 +186,7 @@ where
             &bitcoin_chain_tip,
             vec![(model::TransactionType::SbtcTransaction, tx_1.clone())],
         );
+        test_data.remove(original_test_data);
```
djordon, line 189: Oh any idea why this was necessary?

signer/migrations/0013__consolidate_withdrawal_tables.sql
```diff
@@ -0,0 +1,10 @@
...
+
+-- Remove the FK `block_hash` REFERENCES `stacks_blocks(block_hash)` constraint.
+ALTER TABLE sbtc_signer.withdrawal_requests
+    DROP CONSTRAINT withdrawal_requests_block_hash_fkey;
+
+-- Add the new column to the `withdrawal_requests` table.
```
djordon, line 8: I would change the comment to be something like the one that you added in the rust field.

signer/migrations/0013__consolidate_withdrawal_tables.sql
```diff
@@ -0,0 +1,10 @@
+
+DROP TABLE sbtc_signer.withdrawal_create_events;
+
+-- Remove the FK `block_hash` REFERENCES `stacks_blocks(block_hash)` constraint.
```
djordon, line 4: I would change this to something like:
> We cannot guarantee that we have the stacks block in the database by the time we have received a withdrawal request and are ready to write it into the database. So we need to drop the constraint.

signer/migrations/0003__create_tables.sql
```diff
@@ -152,6 +152,8 @@ CREATE TABLE sbtc_signer.completed_deposit_events (
     created_at          TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL
 );
 
+-- DEPRECATED: dropped in migration 0013__consolidate_withdrawal_tables.
+-- Use `sbtc_signer.withdrawal_requests` instead.
```
djordon, line 156: We can keep this (and maybe even do something like this more often), but it's sometimes best not to modify old migration files. Some migration tools use a checksum to know if it should reapply a migration. We don't do that, so modifying a file is fine I just thought that it couldn't hurt for me to point out.

signer/src/testing/transaction_coordinator.rs
Jiloc: This test was failing due to a bug in `get_withdrawal_requests` within the in-memory storage. Previously, it passed because of the random seed. While debugging, I discovered that the in-memory database was duplicating withdrawal requests. They are added once in  `prepare_database_and_run_dkg` and a second time when we call `self.write_test_data(&test_data).await;` in the test. For most of the data it's not a problem because they are stored in a HashMap, but the mapping between stacks_blocks and withdrawal requests is stored with in vector that caused duplicates.

https://github.com/stacks-network/sbtc/blob/4a4dee23990ac91d4af96f95940a1fd90422683b/signer/src/storage/in_memory.rs#L81-L83

It wasn't strictly necessary, but removing the original data maked it more "correct" so the withdrawals are returned only once from the function.

djordon (approved):
Looks good 🚀

signer/src/testing/request_decider.rs
```diff
@@ -303,7 +303,7 @@ where
     /// Assert that the transaction signer will make and store decisions
     /// received from other signers.
     pub async fn assert_should_store_decisions_received_from_other_signers(self) {
-        let mut rng = rand::rngs::StdRng::seed_from_u64(46);
+        let mut rng = rand::rngs::StdRng::seed_from_u64(42);
```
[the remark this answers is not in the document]
Jiloc: After the changes in this PR, the call to `let test_data = self.generate_test_data(&mut rng, signer_set);` with the previous seed (46) started generating only 25 valid deposits, instead of the 30 expected. The test started to fail

Issue #1331, opened 2025-02-10
[Feature] Demo the withdrawals using our demo-cli


djordon:
# Feature - Demo the withdrawals using our demo-cli

## 1. Description

We to demo the happy path of the withdrawal flow. This requires a command that submits an `initiate-withdrawal-request` contract call to the stacks node using the given command line inputs.

### 1.1 Context & Purpose

Once we have this, we can demo things by first minting using out deposit command, and then withdraw back to the L1 using the new withdraw command.

## 2. Technical Details:

We have the code for creating a withdrawal request transaction: https://github.com/stacks-network/sbtc/blob/cb8070727192cade55c13387083ee987f96d20a4/signer/src/testing/wallet.rs#L112-L157 We need to take the cli input arguments and map them to the transaction and submit that transaction to the stacks node.

### 2.1 Acceptance Criteria:
   - [ ] We have a command that takes valid inputs, constructs a `initiate-withdrawal-request` contract call transaction, and submits it to the stacks node.

PR #1347, opened 2025-02-12
feat: add withdraw to demo-cli

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/1331

## Changes

Add `withdraw` to `demo-cli` to issue a `initiate-withdrawal-request` to the sBTC contract. Also cleanup a bit the demo-cli and added a bunch of defaults (for devenv) so that only the amount is required.
Now the `amount` specified in the cli (for both deposits and withdrawals) includes the `max_fee`.

## Testing Information

Tested in devenv:
```
./signers.sh demo # usual donation + deposit

cargo run -p signer --bin demo-cli deposit --amount 100000
cargo run -p signer --bin demo-cli withdraw --amount 100000
```

The contract call is accepted (and funds locked), then the events are then picked up by the signers event observer, and a new request is created on Emily (we should probably move this to the sidecar).

## Checklist:

### DIFF
--- a/docker/stacks/stacks-regtest-miner.toml
+++ b/docker/stacks/stacks-regtest-miner.toml
@@ -195,3 +195,7 @@ amount = 10000000000000000
 # below was created using the SignerWallet struct.
 address = "SN3R84XZYA63QS28932XQF3G1J8R9PC3W76P9CSQS"
 amount = 10000000000000000
+
+[[ustx_balance]]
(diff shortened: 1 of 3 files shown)


djordon (approved):
Looks good ✅

I left one suggestion about changing the withdrawal amount argument. What is here is fine.

PR #1349, opened 2025-02-12
fix: enable TLS in the blocklist client

djordon:
## Description

We accidentally disabled TLS support in reqwests and none of our tests caught it. That has been fixed.

I have branched off of tag `0.0.9-rc7` here.

## Changes

* Enable TLS using `rustls` for the reqwests library
* Add some simple tests that fail if we disable it in reqwests.

## Checklist:

### DIFF
--- a/blocklist-client/tests/tls_checking.rs
+++ b/blocklist-client/tests/tls_checking.rs
@@ -1,0 +1,15 @@
+use std::time::Duration;
+
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+
+#[tokio::test]
+async fn check_tls_support() {
+    let resp = reqwest::Client::new()
+        .get("https://google.com")
+        .timeout(REQUEST_TIMEOUT)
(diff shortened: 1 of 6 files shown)


blocklist-client/tests/tls_checking.rs
```diff
@@ -0,0 +1,15 @@
+use std::time::Duration;
+
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+
+#[tokio::test]
+async fn check_tls_support() {
+    let resp = reqwest::Client::new()
+        .get("https://google.com")
```
matteojug, line 8: We will probably regret this (eg, CI failing because some random runner ban or something).

matteojug (approved):

blocklist-client/tests/tls_checking.rs
djordon: Yeah I agree.

Issue #564, opened 2024-09-24
[Feature]: Create `accept-withdrawal` transactions


djordon:
# Feature - Create `accept-withdrawal` transactions

## 1. Description

The signer coordinator needs to put together stacks transactions for finalizing a withdrawal by creating an `accept-withdrawal` transaction. This ticket is for generating the stacks transaction (or in this case the `AcceptWithdrawalV1` struct) that can pass validation by other signers.

## 2. Technical Details:

Write a function that is analogous to the `TxCoordinatorEventLoop::construct_deposit_stacks_sign_request` function, but for `withdrawal-accept` transactions.

### 2.1 Acceptance Criteria:
   - [ ] We have the code that creates a `accept-withdrawal-request` smart contract call object for withdrawals that have been swept out.

## 3. Related Issues and Pull Requests (optional):

* https://github.com/stacks-network/sbtc/pull/555
* https://github.com/stacks-network/sbtc/issues/563

PR #1357, opened 2025-02-13
feat: create accept withdrawal transactions

Jiloc:
## Description

Closes: #564 

## Changes
- add `construct_withdrawal_accept_stacks_sign_request` method to the coordinator

## Testing Information
- add a unit test 

## Checklist:

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -971,7 +971,7 @@ where
         let outpoint = req.withdrawal_outpoint();
         let assessed_bitcoin_fee = tx_info
             .assess_output_fee(outpoint.vout as usize)
-            .ok_or_else(|| Error::VoutMissing(outpoint.txid.into(), outpoint.vout))?;
+            .ok_or_else(|| Error::VoutMissing(outpoint.txid, outpoint.vout))?;
 
         // TODO: Add the signer_bitmap to SweptWithdrawalRequest
         let req_id = QualifiedRequestId {
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -16,6 +16,7 @@ use futures::StreamExt as _;
 use sha2::Digest;
 
 use crate::bitcoin::utxo;
+use crate::bitcoin::utxo::BitcoinInputsOutputs as _;
 use crate::bitcoin::utxo::Fees;
 use crate::bitcoin::utxo::UnsignedMockTransaction;
 use crate::bitcoin::BitcoinInteract;
@@ -59,6 +60,7 @@ use crate::stacks::contracts::SMART_CONTRACTS;
 use crate::stacks::wallet::MultisigTx;
 use crate::stacks::wallet::SignerWallet;
 use crate::storage::model;
+use crate::storage::model::QualifiedRequestId;
 use crate::storage::model::StacksTxId;
 use crate::storage::DbRead as _;
 use crate::wsts_state_machine::FireCoordinator;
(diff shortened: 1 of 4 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -945,6 +946,47 @@ where
...
+            signer_bitmap: withdrawal_accept.signer_bitmap,
+            deployer: self.context.config().signer.deployer,
+            sweep_block_hash: withdrawal_accept.sweep_block_hash,
+            sweep_block_height: withdrawal_accept.sweep_block_height,
+        });
+
+        let tx_fee = self
+            .context
+            .get_stacks_client()
+            .estimate_fees(wallet, &contract_call, FeePriority::Medium)
```
Jiloc, line 973: should we use a `FeePriority::High`?

signer/src/transaction_coordinator.rs
```diff
@@ -945,6 +946,47 @@ where
...
+    ///
+    /// This function uses stacks-core for fee estimation of the transaction.
+    #[tracing::instrument(skip_all)]
+    pub async fn construct_withdrawal_accept_stacks_sign_request(
+        &self,
+        withdrawal_accept: model::WithdrawalAcceptEvent,
+        bitcoin_aggregate_key: &PublicKey,
+        wallet: &SignerWallet,
+    ) -> Result<(StacksTransactionSignRequest, MultisigTx), Error> {
+        let contract_call = ContractCall::AcceptWithdrawalV1(AcceptWithdrawalV1 {
```
Jiloc, line 960: In each function where we create a transaction, we have a `TODO: we should validate the contract call before asking others to sign it.`. But is it the case? In this one, for example, I suppose we'll be trusting the data we get from our own DB?

djordon: I'm not entirely sure. It cuts down on network traffic, at the cost of potentially slowing down fulfillment of accepted requests. If I would reject a request but suspect that other will accept it, me running validation will slow down approval since it would fail for me but succeed when someone else is the coordinator.

signer/src/transaction_coordinator.rs
djordon: Na, `Medium` seems just right. The user does not have access to the locked funds so no rush.

signer/src/transaction_coordinator.rs
```diff
@@ -945,6 +946,47 @@ where
...
     }
 
+    /// Transform the withdrawal accept event into a Stacks sign request
+    /// object.
+    ///
+    /// This function uses stacks-core for fee estimation of the transaction.
+    #[tracing::instrument(skip_all)]
+    pub async fn construct_withdrawal_accept_stacks_sign_request(
+        &self,
+        withdrawal_accept: model::WithdrawalAcceptEvent,
```
djordon, line 956: Oh we won't have access to this when the coordinator tries to make the call here. This type is supposed to be sent from the stacks node when the `AcceptWithdrawalV1` contract call successfully executes. Maybe have it take a `SweptWithdrawalRequest` instead.

djordon (commented):
We should probably change one of the input types for `construct_withdrawal_accept_stacks_sign_request`.

signer/src/transaction_coordinator.rs
Jiloc: That was a 🤦‍♂️! I updated it to use the  `SweptWithdrawalRequest`

signer/src/transaction_coordinator.rs
```diff
@@ -945,6 +948,88 @@ where
...
+            request_id: req.request_id,
+            txid: req.txid,
+            block_hash: req.block_hash,
+        };
+
+        let votes = self
+            .context
+            .get_storage()
+            .get_withdrawal_request_signer_votes(&req_id, bitcoin_aggregate_key)
+            .await?;
```
Jiloc, line 1000: ideally, we should have this information in the `SweptWithdrawalRequest`. I am not updating it in this PR because it would mess up the PR @MCJOHN974 is working on. But we should follow up with the change

signer/src/transaction_coordinator.rs
```diff
@@ -953,20 +955,61 @@ where
...
+                Error::BitcoinNoInputsInTransaction(
+                    req.sweep_txid.into(),
+                    Some(req.sweep_block_hash.into()),
+                )
+            })?
+            .previous_output;
+
+        let assessed_bitcoin_fee = tx_info
+            .assess_input_fee(&outpoint)
+            .ok_or_else(|| Error::OutPointMissing(outpoint))?;
```
djordon, line 988: The assessed fee here is for the output associated with the withdrawal request. But, to get that fee you need to know which output index is associated with this withdrawal. That value is not in the `SweptWithdrawalRequest` right now but it should be. So we could add it to `SweptWithdrawalRequest` and update the query when https://github.com/stacks-network/sbtc/issues/1326 gets done.

signer/src/transaction_coordinator.rs
```diff
@@ -953,20 +955,61 @@ where
...
+            .context
+            .get_bitcoin_client()
+            .get_tx_info(&req.sweep_txid, &req.sweep_block_hash)
+            .await?
+            .ok_or_else(|| {
+                Error::BitcoinTxMissing(req.sweep_txid.into(), Some(req.sweep_block_hash.into()))
+            })?;
+        // Retrieve the first input, which is the signer's UTXO.
+        let outpoint = tx_info
+            .inputs()
```
djordon, line 974: This should be the outpoint of the withdrawal.

Jiloc: fixed in https://github.com/stacks-network/sbtc/pull/1357/commits/d422c8b36c751722ce396e68b422ca06c9835759

signer/src/transaction_coordinator.rs
Jiloc: I hardcoded it to two (to pass the `assess_output_fee` call) until #1326 is done.

signer/src/transaction_coordinator.rs
Jiloc: fixed in https://github.com/stacks-network/sbtc/pull/1357/commits/d422c8b36c751722ce396e68b422ca06c9835759

signer/src/transaction_coordinator.rs
```diff
@@ -945,6 +947,75 @@ where
         Ok((sign_request, multi_tx))
     }
 
+    /// Transform the withdrawal accept event into a Stacks sign request
+    /// object.
+    ///
```
djordon, line 952: nit: Update to reference the `SweptWithdrawalRequest` instead?

djordon (approved):
Looks good ✅

signer/src/transaction_coordinator.rs
djordon: It might also be easier to do it here like this, we'll find out soon enough.

signer/src/transaction_coordinator.rs
Jiloc: fixed in https://github.com/stacks-network/sbtc/pull/1357/commits/60e71733e2dd4bea9272f1d132e7710a3a83c5ef

PR #1360, opened 2025-02-15
fix: correctly generate stacks chain in TestData

matteojug:
## Description

Closes: #?

## Changes

There was a bug in how we generate the stacks chain via `TestData`, constraining the stacks chain to a single bitcoin blocks. This fixes it and a couple of tests that were affected by the change.

## Testing Information

Add a test to ensure we generate stacks chain spanning multiple blocks.

## Checklist:

### DIFF
--- a/signer/src/testing/storage/model.rs
+++ b/signer/src/testing/storage/model.rs
@@ -105,6 +105,7 @@ impl TestData {
         let withdraw_data = WithdrawData::generate(
             rng,
             signer_keys,
+            &block,
             &stacks_blocks,
             &self.withdraw_requests,
             params.num_withdraw_requests_per_block,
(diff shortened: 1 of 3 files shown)


signer/src/testing/storage/model.rs
```diff
@@ -564,3 +567,57 @@ fn vec_diff<T: std::cmp::Eq + std::hash::Hash>(subtrahend: &mut Vec<T>, minuend:
...
+        let mut walk = vec![tip];
+        while let Some(current) = store
+            .get_stacks_block(&walk.last().unwrap().parent_hash)
+            .await
+            .unwrap()
+        {
+            walk.push(current);
+        }
+
+        assert_ge!(walk.len(), 10);
```
djordon, line 621: Can we also check that the heights increment in order. Actually, I'm not sure that the stacks block heights are used so maybe we don't need it.

djordon (approved):
🙏🏽 Looks good

Issue #1365, opened 2025-02-17
[Feature]: Expire signature hashes with bitcoin blocks


djordon:
# Feature – Expire Signature Hashes with Bitcoin Blocks

## 1. Description

Each signer decides which signature digest (sighash) they are willing to sign. Currently, there is no expiration for these sighashes, signers will sign them at any time. While this poses almost no risk for deposits, it can create issues for withdrawals. Consequently, we need to ensure sighashes expire, with the most natural expiration aligning with the lifetime of the current chain tip of the Bitcoin blockchain.

## 2. Technical Details

This could be done by making the following changes:
1. Create a SQL migration that makes the combination of `sighash` and `bitcoin_chain_tip` the compound primary key on the `bitcoin_tx_sighashes` table.
2. Update our queries to read from the `bitcoin_tx_sighashes` table using the bitcoin chain tip.
3. Change any lookups for sighashes in the table to use the specific Bitcoin block hash (`bitcoin_chain_tip`).

### 2.1 Acceptance Criteria
- [ ] Signers will only sign a particular sighash for the duration of one Bitcoin block.

## 4. Appendix

When I originally opened this ticket I had in mind some attack vector. The plan was that I'd fill in the details later. After coming back to this ticket to detail out the attack, I no longer think that the one that I had in mind was possible. For that reason, it might be a good idea to close this ticket and the associated PR.

djordon:
I don't think that this solves anything.

PR #1366, opened 2025-02-17
feat: expire signature hashes with bitcoin blocks

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1365.

## Changes

* Change the primary key on the `bitcoin_tx_sighashes` table to be a compound primary key with the bitcoin chain tip block hash.
* Require the chain tip when querying whether a signer will sign for a given sighash.

## Testing Information

This PR adds a small check that a given sighash does not return the same result with a different sighash.

## Checklist:


djordon:
After thinking about this some more, I'm no longer confident that this solves an actual issue. I think I was seeing ghost attack vectors 😅

PR #1380, opened 2025-02-18
feat: add Emily private mainnet stage

fabergat:
## Description

Add Emily private mainnet stage

## Changes

- Add private mainnet stage
- Fix deployment script

## Checklist:


scripts/deploy-emily.sh
```diff
@@ -36,6 +36,6 @@ fi
         --profile "${AWS_DEPLOYMENT_PROFILE}"
     echo "Done bootstrapping CDK."
     npx aws-cdk deploy \
-        --profile "${AWS_DEPLOYMENT_PROFILE}"
+        --profile "${AWS_DEPLOYMENT_PROFILE}" \
```
Jiloc, line 39: How was it working previously? O:

emily/cdk/lib/emily-stack-utils.ts
```diff
@@ -148,7 +148,8 @@ export class EmilyStackUtils {
         return this.getStageName() === Constants.DEV_STAGE_NAME
             || this.getStageName() === Constants.LOCAL_STAGE_NAME
             || this.getStageName() === Constants.UNIT_TEST_STAGE_NAME
-            || this.getStageName() === Constants.TEMP_STAGE_NAME;
+            || this.getStageName() === Constants.TEMP_STAGE_NAME
+            || this.getStageName() === Constants.PRIVATE_MAINNET_STAGE_NAME;
```
Jiloc, line 152: are we sure we want to consider private mainnet a developmentStack? 

For example, this setting will cause persistent resources to be deleted when the stack is removed. I understand that the total $$ spent on transactions in the private mainnet is likely lower than the additional cost of deploying a production stack, but I just wanted to double-check!

Jiloc (commented):
We also need to add the new stage to:

```
https://github.com/stacks-network/sbtc/blob/1024f47de057029271ef51d5497608244ff3afb0/emily/cdk/lib/emily-stack.ts#L346
```

This variable is used to validate the addresses Emily receives (as they may differ between mainnet and testnet)

scripts/deploy-emily.sh
fabergat: ignoring the last parameter...

emily/cdk/lib/emily-stack-utils.ts
fabergat: Technically, it is a test environment running on mainnet so I would consider like a dev environment...

fabergat:
> ```
> https://github.com/stacks-network/sbtc/blob/1024f47de057029271ef51d5497608244ff3afb0/emily/cdk/lib/emily-stack.ts#L346
> ```

thanks! Updated

Jiloc (approved):
lgtm

Issue #1384, opened 2025-02-18
Problems running dev env


vini-btc:
Hi everyone,

First of all, thanks for putting the work to bring sbtc to life. The whole community will benefit from it for years to come.

I'm trying to run a local devenv, but I'm facing the following issue when trying to start the containers on the docker-compose:

```shell
% DOCKER_BUILDKIT=0 make devenv-up
docker compose -f docker/docker-compose.yml --profile default --profile bitcoin-mempool --profile sbtc-signer up -d
[+] Running 1/1
 ! electrs Warning                                                                                                                                                                                                                       1.5s
ERRO[0006] Can't add file /Users/vini/Dev/sbtc/.generated-sources/emily/client/rust/private/docs/CorsApi.md to tar: io: read/write on closed pipe
ERRO[0006] Can't close tar writer: io: read/write on closed pipe
ERRO[0006] Can't add file /Users/vini/Dev/sbtc/.generated-sources/emily/client/rust/private/docs/CorsApi.md to tar: io: read/write on closed pipe
ERRO[0006] Can't close tar writer: io: read/write on closed pipe
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /Users/vini/Dev/sbtc/docker/bitcoin/Dockerfile: no such file or directory
make: *** [devenv-up] Error 1
```
I'm not entirely sure I missed some previous step (I should be able to just docker compose up a dev env, right?) or if that's some issue with my own environments (I've tried on Ubuntu 24.04 and Mac OS - M2 - Sonoma 14.6.1). If yes, please let me know.

vini-btc:
At `0.0.9-rc7.1` the command works fine.

djordon:
Hi @vini-btc, yeah that error seems pretty weird to me. The "missing" file (`sbtc/docker/bitcoin/Dockerfile`) was removed in a recent PR and I don't see it referenced anymore. I just pulled from `main` and ran `make devenv-up` from the command line without any issues. I'm currently on an M3 macOS Sonoma 14.5.

If you're still having issues let us know. We all want folks to be able to develop with/on sBTC easily. If it will help I can try on Ubuntu 24.04 to see if I encounter the same issue.

djordon:
[...]

PR #1387, opened 2025-02-19
chore: fix bitcoin-miner image in devenv

matteojug:
## Description

Address https://github.com/stacks-network/sbtc/issues/1384
Follow up for https://github.com/stacks-network/sbtc/pull/1307

We removed the `bitcoin` Dockerfile, but the miner was still referencing it. We didn't catch this because we didn't try to rebuild the whole stack and devenv doesn't run in CI.

A possible downside of the current approach is that we require a network connection for the container to start up (this applies to `stacks-node`, `stacks-signer-*` and `bitcoin-miner`), so it's not possible to run devenv offline.

## Testing Information

Tested with `docker compose -f docker/docker-compose.yml --profile default --profile bitcoin-mempool --profile sbtc-signer build --no-cache --pull` (+ usual demo run).

## Checklist:

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -268,7 +268,7 @@ services:
       - bitcoin-mempool
 
   bitcoin-miner:
-    build: bitcoin
+    image: bitcoin/bitcoin:25.2
     depends_on:
       bitcoin:
         condition: service_healthy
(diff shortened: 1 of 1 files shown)


djordon (approved):
Oh good man! I searched this file for Dockerfile, and forgot that it gets referenced automatically this way.

Issue #1375, opened 2025-02-18
[Feature]: Change uniqueness pre-validation check to use the `request_id`


djordon:
# Feature - Change uniqueness pre-validation to be unique on the `request_id`

## 1. Description

There is a pre-validation check that tests whether the `BitcoinPreSignRequest` is valid on it's face. One of the checks is to see whether the IDs in the object are unique. Here the IDs are the `QualifiedRequestId` and the `OutPoint`. However, we can be even more aggressive and require uniqueness of the `request_id` in the `QualifiedRequestId`.

### 1.1 Context & Purpose

The request ID in a withdrawal request is created by a contract call. As such, it is not globally unique but is unique per Stacks blockchain. Since we are only interested in one Stacks blockchain (the canonical one), we know that the request IDs that that need to validate must be unique. So we can quickly end a validation check if we see that one of the proposed sweep transactions will be bogus and include withdrawal that does not exist on the canonical Stacks blockchain.

## 2. Technical Details:

The uniqueness check happens here:
https://github.com/stacks-network/sbtc/blob/0941183261fc716fec5ab60959c06193b93d2d72/signer/src/bitcoin/validation.rs#L93-L102

Update it to use the request ID instead of the qualified request ID. Several test cases in `test_pre_validation` test fail, update them so that they pass and add a new one where the only thing wrong with the `BitcoinPreSignRequest` object in the test case is the fact that there are duplicate request IDs.

### 2.1 Acceptance Criteria:
   - [ ] Our uniqueness check during pre-validation reject duplicate request IDs in withdrawals.

PR #1389, opened 2025-02-19
[feat] Update Withdrawal Request Validation

setzeus:
## Description
This PR updates the 'is_unique' validation inside the 'pre_validation' checker. Specifically, this update swaps out the logic of checking the uniqueness of a withdrawal 'QualifiedRequestID' with checking the uniqueness of a withdrawal on-chain 'RequestID' (a property found in the withdrawal QualifiedRequestID).

Closes: #1375 

## Changes
Changes were made to the 'is_unique' helper found inside the 'pre_validation' function found inside the 'validation.rs' file within 'signer'.

## Testing Information
A test was updated to match the fail-case scenario here (the same request_id within the same tx). Additionally, a handful of other tests were updated to fail correctly (aka adhere to the name given).

## Checklist:

### DIFF
--- a/signer/src/bitcoin/validation.rs
+++ b/signer/src/bitcoin/validation.rs
@@ -94,11 +94,14 @@ impl From<&Requests<'_>> for TxRequestIds {
 pub fn is_unique(package: &[TxRequestIds]) -> bool {
     let mut deposits_set = HashSet::new();
     let mut withdrawal_request_id_set = HashSet::new();
-    
+
     package.iter().all(|reqs| {
         let deposits = reqs.deposits.iter().all(|out| deposits_set.insert(out));
-        let withdrawal_requests = reqs.withdrawals.iter().all(|id| withdrawal_request_id_set.insert(id.request_id));
-            
+        let withdrawal_requests = reqs
+            .withdrawals
+            .iter()
+            .all(|id| withdrawal_request_id_set.insert(id.request_id));
+
         deposits && withdrawal_requests
     })
 }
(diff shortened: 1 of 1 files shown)


signer/src/bitcoin/validation.rs
```diff
@@ -1872,36 +1877,7 @@ mod tests {
...
-                            txid: StacksTxId::from([1; 32]),
-                            block_hash: StacksBlockHash::from([2; 32]),
-                        },
-                    ],
-                },
-                TxRequestIds {
-                    deposits: Vec::new(),
-                    withdrawals: Vec::new(),
-                },
-            ],
```
djordon, line 1904: We don't want to remove all of these, since it's a separate test case than the one that you're turning it into. With this change, this becomes identical to the `empty-package_requests` test case.

signer/src/bitcoin/validation.rs
```diff
@@ -1841,7 +1846,7 @@ mod tests {
                 TxRequestIds {
                     deposits: vec![OutPoint {
                         txid: Txid::from_byte_array([1; 32]),
-                        vout: 1,
+                        vout: 0,
```
djordon, line 1849: This doesn't need to change, though I suppose it doesn't hurt.

djordon (commented):
We have a test case for duplicate qualified request ids in `duplicate-withdrawals-in-same-tx`, but can we add another one for duplicate `request_ids` only, where the block hash changes.

signer/src/bitcoin/validation.rs
setzeus: Updated in [53a503292064443d64cf37b6ceb771e1e87c9340](https://github.com/stacks-network/sbtc/pull/1389/commits/53a503292064443d64cf37b6ceb771e1e87c9340)

djordon (approved):
Looks good.

Issue #1321, opened 2025-02-07
[Feature]: Coordinator filters withdrawal requests using consensus rules


djordon:
# Feature - Coordinator filters withdrawal requests using consensus rules

## 1. Description

The consensus rules for withdrawal requests are detailed in https://github.com/stacks-network/sbtc/issues/741. The coordinator must filter the proposed withdrawal requests to be in line with these rules or else the sweep transaction will be rejected when going through validation with peer signers.

### 1.1 Context & Purpose

The coordinator selects the withdrawal requests to include in a sweep. If it includes a one “bad” request, the signers will reject the sweep transaction, affecting all requests that would be included in the sweep. So the coordinator needs to follow consensus rules when picking requests.

## 2. Technical Details:

An individual withdrawal request will validate through sbtc consensus rules if it’s `WithdrawalRequestReport` passes validation. The work for building the report is associated with https://github.com/stacks-network/sbtc/issues/1308, while this ticket can be accomplished by using the report in the coordinator code path.

### 2.1 Acceptance Criteria:

- [ ]  The coordinator uses consensus rules for filtering withdrawal requests

## 3. Related Issues and Pull Requests (optional):

- https://github.com/stacks-network/sbtc/issues/741
- https://github.com/stacks-network/sbtc/issues/1308

## 4. Appendix

There are collective cap rules around withdrawals, where we may want to throttle the system so that not too much is withdrawn per bitcoin block. We haven't formalized what those rules should be, so that part of consensus rules doesn't need to be done with this ticket.

Issue #741, opened 2024-10-30
[Feature]: ☂ Validate withdrawal requests as part of bitcoin validation


djordon:
# Feature - Validate withdrawal requests as part of bitcoin validation

## 1. Description

This is a subtask of https://github.com/stacks-network/sbtc/issues/286 and aligns with https://github.com/stacks-network/sbtc/issues/782.

There are a bunch of properties that need to be checked when validating withdrawal requests as part of bitcoin validation. Finish the functions that check these properties during validation.

### 1.1 Context & Purpose

The flow for validating withdrawals is to create a report struct containing the key pieces of information about a withdrawal request. They include:
1. **The double spend check**. Check whether the qualified request ID is in the `bitcoin_withdrawals_outputs` table, and that any of the associated `txid`s are confirmed on the canonical bitcoin blockchain. Fail the withdrawal request if such a transaction was found.
2. **The request exists**. Check whether the associated withdrawal request transaction is confirmed on the canonical stacks blockchain. Fail the withdrawal request if it is not on the canonical stacks blockchain.
3. **The request is final**. Check that we’ve seen six new bitcoin blocks since observing the bitcoin anchor block associated with the Stacks block confirming the withdrawal request transaction. Fail the withdrawal request if we’ve observed less than 6 bitcoin blocks since the anchor block.
4. **The request is collectively approved.** Check if the request has enough votes. Fail the withdrawal request if we do not have enough votes.
5. **The request is approved locally**. Check if you have approved the request. Fail the withdrawal request if we did not approve the withdrawal.
6. **Within fee bounds**. The assessed fees will be within the constraints of the max-fee. Fail the withdrawal request if the assessed fee was greater than the max-fee.
7. **The request has not expired**. Check if less than X bitcoin blocks have been confirmed since the anchor block confirming the withdrawal request transaction. Fail the withdrawal request if we’ve observed X or more bitcoin blocks since the observing the associated anchor block.
8. **The withdrawal amount is valid**. Check that the withdrawal amount is greater than some dust amount. This is checked in the smart contract as well, but we should check it too. Fail the withdrawal request if the withdrawal amount would validate some bitcoin enforced minimum dust amount.
9. **The withdrawal amount is within the cap**. Check that the withdrawal amount is less than or equal to the withdrawal cap. Perhaps there is a per-bitcoin withdrawal amount cap as well. Fail the withdrawal request if the withdrawal amount if greater than the withdrawal cap.
[...]

PR #1397, opened 2025-02-19
feat: concretize pending accepted withdrawals query and implement tests

cylewitruk:
## Description

Related to: #1321

~85% of this is test code.

Updates the existing query and implements (a lot of) tests for various scenarios to ensure its correctness according to #741, for the portions which it helps to verify:

- **[1]** The request must not have been swept within the current canonical Bitcoin chain.
- **[2]** The request must be confirmed in a canonical Stacks block.
- **[4]** (partial) The required _number_ of signers have approved the request. The check in the query acts as a sanity-pre-filter and does not verify that the signers are part of the current signer set. This validation is done in code to not increase query complexity as we need to pull the votes for the votes bitmap anyway.
- **[7]** The request is not expired based on its confirmed-at bitcoin height. This is done using the query's Bitcoin context window.
- Additionally, that there are no canonically confirmed withdrawal rejections.

Code-side verifications in next PR.

## Changes

- Updates the `get_eligible_pending_withdrawal_requests` query to include missing consensus rules which the query can effectively cover (we're in the hot path).
- Adds a suite of tests covering the above scenarios, including orphaning (I think I covered them all? Lmk if I missed something).
- Adds some "block" test helpers in `blocks.rs` (this is a bit of a WIP but works for my purposes in this and the next PR).
- The query now accepts the `stacks_chain_tip` instead of implicitly running the query internally (as done in the withdrawal reports -- I think we should opt for this in general so we're not running the same query countless times and implicitly).
    - Refactors in consuming code to support this.

## Checklist:


signer/src/storage/postgres.rs
```diff
@@ -1534,21 +1521,51 @@ impl super::DbRead for PgStore {
...
             JOIN sbtc_signer.withdrawal_signers signers ON
-                wr.txid = signers.txid AND
-                wr.request_id = signers.request_id AND
-                wr.block_hash = signers.block_hash
-            WHERE
-                signers.is_accepted
-            GROUP BY wr.request_id, wr.block_hash, wr.txid
-            HAVING COUNT(wr.request_id) >= $4
+                wr.request_id = signers.request_id
+                AND wr.txid = signers.txid
```
djordon, line 1533: We can remove this `txid` condition here as well.

signer/tests/integration/postgres.rs
```diff
@@ -4919,3 +4919,938 @@ async fn pending_rejected_withdrawal_already_accepted() {
...
+    /// This test creates blockchains with the following structure:
+    ///
+    /// ```text
+    /// Age:         3           2           1     
+    ///          ┌────────┐  ┌────────┐  ┌────────┐
+    /// Bitcoin: │   B1   ├──►   B2   ├──►   B3   │ The request is confirmed (✔)
+    ///          └─▲──────┘  └─▲──────┘  └─▲──────┘ in S2 and we test different
+    ///          ┌─┴──────┐  ┌─┴──────┐  ┌─┴──────┐ context windows.
+    /// Stacks:  │   S1 ✔ ├──►   S2   ├──►   S3   │
+    ///          └────────┘  └────────┘  └────────┘
```
djordon, line 5069: Very nice! How did you make it?

signer/tests/integration/postgres.rs
```diff
@@ -4919,3 +4919,938 @@ async fn pending_rejected_withdrawal_already_accepted() {
...
+                context_window,
+                signature_threshold,
+            )
+            .await
+            .expect("failed to query db");
+        assert!(requests.is_empty());
+    }
+
+    /// Asserts that a withdrawal request is returned when it has been confirmed
+    /// and has a rejection event, but the rejection event has been orphaned.
```
djordon, line 5199: Hopefully we never have to deal with this case. If it happens then we did something wrong. Glad we have the test though.

signer/tests/integration/postgres.rs
```diff
@@ -4919,3 +4919,938 @@ async fn pending_rejected_withdrawal_already_accepted() {
...
+                context_window,
+                signature_threshold,
+            )
+            .await
+            .expect("failed to query db");
+        assert!(requests.is_empty());
+    }
+
+    #[tokio::test]
+    async fn expired_request_not_returned() {}
```
djordon, line 5358: This can be removed, I think this was done in `expired_requests_not_returned`.

signer/tests/integration/postgres.rs
```diff
@@ -4919,3 +4919,938 @@ async fn pending_rejected_withdrawal_already_accepted() {
 
     signer::testing::storage::drop_db(db).await;
 }
+
+/// Module containing a test suite and helpers specific to
+/// `DbRead::get_pending_accepted_withdrawal_requests`.
+mod get_pending_accepted_withdrawal_requests {
```
djordon, line 4925: This is a fantastic suite of tests

djordon (approved):
Looks good.

Great job on the test suite here.

signer/tests/integration/postgres.rs
cylewitruk: Thanks -- I drew the general outline of one of the more complicated ones below on https://asciiflow.com/ and from there just a little copy-paste and small tweaking. 

It started out as a mental model for myself, but I found it helpful to follow when writing the tests and it was easy enough, so I decided to put one on all of them :)

signer/tests/integration/postgres.rs
cylewitruk: Whoops, this was intended to be a placeholder for that test 😅 Removed in [a99ac28](https://github.com/stacks-network/sbtc/pull/1397/commits/a99ac28015198f3bd0b1fdcab8870a7cbfb5d9be)

signer/src/storage/mod.rs
```diff
@@ -166,13 +166,39 @@ pub trait DbRead {
...
-    /// Get pending withdrawal requests that have been accepted by at least
-    /// `threshold` signers and has no responses
+    /// This function returns withdrawal requests filtered by a portion of the
+    /// consensus critera defined in #741.
+    ///
+    /// ## Filter Criteria
+    ///
+    /// 1. The withdrawal request transaction (`create-withdrawal-request`
+    ///    contract call) is confirmed in a stacks block anchored to the
+    ///    canonical bitcoin blockchain.
```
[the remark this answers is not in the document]
cylewitruk: Updated in [27f51e6](https://github.com/stacks-network/sbtc/pull/1397/commits/27f51e61cc0bb28036bae1fb5bfbc8debdf21167). But yeah, I'm always trying to find a more efficient way to write "in a block in the canonical stacks blockchain which is anchored in the canonical bitcoin blockchain" 🙈

signer/src/storage/postgres.rs
```diff
@@ -1534,21 +1521,51 @@ impl super::DbRead for PgStore {
...
+                ON bt.block_hash = canonical_sweep.block_hash
+
+            -- Are there any confirmed reject transactions?
+            LEFT JOIN sbtc_signer.withdrawal_reject_events AS wre
+                ON wre.request_id = wr.request_id
+            LEFT JOIN stacks_blockchain AS canonical_reject
+                ON wre.block_hash = canonical_reject.block_hash
+
+            GROUP BY 
+                wr.request_id, wr.block_hash, wr.txid
```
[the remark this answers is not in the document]
cylewitruk: Yup, [4efd8cf](https://github.com/stacks-network/sbtc/pull/1397/commits/4efd8cf681b56de9d57abcf0134c8720919a05f9)

signer/tests/integration/postgres.rs
cylewitruk: Yeah, it's less about testing though and more about asserting the behavior of the query given the case.

signer/src/storage/postgres.rs
```diff
@@ -1477,34 +1477,21 @@ impl super::DbRead for PgStore {
 
     async fn get_pending_accepted_withdrawal_requests(
```
matteojug, line 1478: Shouldn't this also filter out expired requests?

signer/src/testing/blocks.rs
matteojug, line 1: This is cool, but I'm wondering if it could have been added to `TestData` instead of adding a new thingy: we already have quite some duplication of code for different setups, and this seems another instance of it.

signer/src/storage/mod.rs
```diff
@@ -166,13 +166,39 @@ pub trait DbRead {
...
+    ///    (`reject-withdrawal-request` contract call) for the request.
+    ///
+    /// ## Notes
+    ///
+    /// -  This does does not filter `signature_threshold` on that the approving
+    ///    signers are part of the current signer set and this parameter is only
+    ///    used as a pre-filter as the votes themselves are generally also
+    ///    needed separately. Use
+    ///    [`DbRead::get_withdrawal_request_signer_votes`] to fetch the votes
+    ///    and perform this verification separately.
```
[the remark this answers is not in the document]
cylewitruk: I'm not sure which queries you're referring to? `get_deposit_request_signer_votes` and `get_withdrawal_request_signer_votes` are the only queries I see that filter based on the _current_ signer set? This comment simply documents the behavior that was already there.

Otherwise the other queries are still either only checking the approved count, regardless of whether or not the approving signer is part of the current signing set or not, or doing some filtering based on the current signer's pubkey.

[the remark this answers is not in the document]
matteojug: I was thinking we were always checking the votes only for the current signing set, but apparently that's not the case in general. But I guess we can revisit this once we start messing with the signer set to have more filters upstream.

signer/src/storage/postgres.rs
matteojug: I'm not sure about what that could return, nor about the possible strange configurations that we may end up with, nor about off by one errors, and that's kinda why I was thinking doing an explicit check may put us in a better situation. It also ensure that this is aligned with the other checks we do related to expiration around (eg in other queries, or in bitcoin validation).

signer/src/testing/blocks.rs
cylewitruk: Should have/could have, maybe? I'm all for trying to consolidate things, but `TestData` was a bit heavyweight and opinionated for what I was trying to do atm, and given it's used all over the place I didn't want to mess around with it in a hurry.

Ultimately I think it should be `TestData` which uses something like this under the hood, so that our test helpers  build upon each other in an additive way instead of duplicating each other in their own opinionated way.

signer/src/storage/postgres.rs
cylewitruk: Are you referring to this check from your rejection query?
```sql
                JOIN sbtc_signer.bitcoin_blocks block
                  ON block.block_hash = parent.bitcoin_anchor
                WHERE block.block_height >= (SELECT MIN(block_height) FROM bitcoin_blockchain)
```

matteojug: Yeah, as well as the check in `WithdrawalRequestReport` and `RejectWithdrawalV1` that are based on the height diff instead of the anchored block.

signer/src/storage/postgres.rs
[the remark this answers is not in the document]
djordon: > OR even gone so far as to remove this completely and done the check in code. If signers aren't voting on withdrawal requests for some reason (or we're not getting them), the only logs we'll see from them is that they're expiring and we won't know the reason for it.

Yeah, I would prefer this route since it's simpler overall. We should probably change the method name though. It should probably be done in a separate PR though, but I don't feel strongly about it.

signer/src/storage/postgres.rs
djordon: Good catch, yeah we need to change this.

cylewitruk: Alright, made some changes in [fdcd265](https://github.com/stacks-network/sbtc/pull/1397/commits/fdcd2654b5f51e69920377f30203e7ae9f391ba4).

The approach here is to:
- Retrieve all withdrawal requests, regardless of chain, whose bitcoin height is >= the provided min height.
- Fetch the canonical bitcoin chain until the lowest height of the retrieved withdrawal requests
- Fetch the canonical stacks chain until the same anchor height
- The rest is pretty much the same, just grouped some of the joins together with the withdrawal requests CTE.

Note that this is used together with `WITHDRAWAL_BLOCKS_EXPIRY`; the `WITHDRAWAL_EXPIRY_BUFFER` check is done in code (where we log that it's on its way to expiring) -- that's in the other PR.

signer/src/storage/mod.rs
```diff
@@ -166,13 +166,38 @@ pub trait DbRead {
...
     fn get_pending_accepted_withdrawal_requests(
         &self,
-        chain_tip: &model::BitcoinBlockHash,
-        context_window: u16,
-        threshold: u16,
+        bitcoin_chain_tip: &model::BitcoinBlockHash,
+        stacks_chain_tip: &model::StacksBlockHash,
+        min_bitcoin_height: u64,
+        signature_threshold: u16,
     ) -> impl Future<Output = Result<Vec<model::WithdrawalRequest>, Error>> + Send;
```
matteojug, line 201: maybe nit: a possible alternative is to get a bitcoin block ref here and compute `min_bitcoin_height` inside to ensure a consistent usage of it (instead of relying on the caller). WDYT?

signer/tests/integration/postgres.rs
```diff
@@ -4934,7 +4934,7 @@ async fn pending_rejected_withdrawal_no_events() {
 
     // Append some blocks to ensure we have expired requests; we expire the
     // requests in the first 5 canonical blocks, while keeping the others valid.
-    for _ in chain_depth..WITHDRAWAL_BLOCKS_EXPIRY + 5 {
+    for _ in chain_depth..WITHDRAWAL_BLOCKS_EXPIRY as u64 + 5 {
```
matteojug, line 4937: nit: isn't this cast redundant?

signer/src/storage/postgres.rs
```diff
@@ -1477,34 +1477,60 @@ impl super::DbRead for PgStore {
...
+                LEFT JOIN sbtc_signer.bitcoin_transactions bt
+                    ON bt.txid = bwo.bitcoin_txid
 
-                UNION ALL
+                -- Join in any rejection events we know about.
+                LEFT JOIN sbtc_signer.withdrawal_reject_events AS wre
+                    ON wre.request_id = wr.request_id
+                
+                -- Only requests where the bitcoin height is >= than the minimum.
+                WHERE wr.bitcoin_block_height >= $3
```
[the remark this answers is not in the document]
cylewitruk: That's filtered in code and logged

signer/src/storage/mod.rs
cylewitruk: I saw you had done that; I personally prefer that the caller provides it, which atm is only used in one place (the coordinator). Pulling in the constant here reduces flexibility when it comes to testing, and accepting the `expiry window` and doing the calculation here still depends on the caller to provide the correct value. I like to keep the storage layer "dumb" :)

signer/tests/integration/postgres.rs
cylewitruk: It probably came in when i was cherry-picking stuff from the other PR (where i made these changes originally).

signer/src/testing/blocks.rs
```diff
@@ -0,0 +1,220 @@
...
+        B: AsRef<BitcoinBlockHash>,
+    {
+        if self.0.is_empty() {
+            self.0.push(StacksBlock::new_genesis().anchored_to(anchor));
+        } else {
+            let new_block = self.0.last().unwrap().new_child().anchored_to(anchor);
+            self.0.push(new_block);
+        }
+        self.0.last().unwrap()
+    }
```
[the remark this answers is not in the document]
cylewitruk: [a8e19dc](https://github.com/stacks-network/sbtc/pull/1397/commits/a8e19dcc55877c1767a3da373f56eef140099bb0)

signer/src/testing/blocks.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    /// Gets the nth block in the chain, if it exists.
+    pub fn nth_block_checked(&self, height: usize) -> Option<&BitcoinBlock> {
+        self.0.get(height)
+    }
+
+    /// Gets the nth block in the chain, panicking if it does not exist.
+    pub fn nth_block(&self, height: usize) -> &BitcoinBlock {
+        self.0
+            .get(height)
```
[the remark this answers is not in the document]
cylewitruk: [a8e19dc](https://github.com/stacks-network/sbtc/pull/1397/commits/a8e19dcc55877c1767a3da373f56eef140099bb0)

signer/src/storage/postgres.rs
```diff
@@ -1477,34 +1477,60 @@ impl super::DbRead for PgStore {
...
                 SELECT
-                    parent.block_hash
-                  , parent.parent_hash
-                  , last.depth + 1
-                FROM sbtc_signer.bitcoin_blocks parent
-                JOIN extended_context_window last ON parent.block_hash = last.parent_hash
-                WHERE last.depth <= $3
+                    block_hash
+                  , block_height
+                FROM bitcoin_blockchain_until($1, (SELECT MIN(bitcoin_block_height) FROM requests))
```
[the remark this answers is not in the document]
cylewitruk: Tests added in [8a12fd3](https://github.com/stacks-network/sbtc/pull/1397/commits/8a12fd3e3f0d5f93501964342aecd3a56bb10fb7) showing current behavior.

signer/src/testing/blocks.rs
```diff
@@ -0,0 +1,220 @@
...
+            .expect("no nth bitcoin block (index out of range)")
+    }
+}
+
+impl BitcoinBlock {
+    /// Create a new bitcoin block with the following properties:
+    /// - block_hash: random
+    /// - block_height: 0
+    /// - parent_hash: all zeroes
+    pub fn new_genesis() -> Self {
```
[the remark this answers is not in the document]
cylewitruk: Yeah, probably, or at least a separate method that lets you specify one. He who needs it gets to add it :)

signer/tests/integration/postgres.rs
cylewitruk: Removed in [f1e0eed](https://github.com/stacks-network/sbtc/pull/1397/commits/f1e0eed82bc30653280131c2d4b48a2fe256ad40)

signer/src/storage/postgres.rs
```diff
@@ -1549,6 +1549,7 @@ impl super::DbRead for PgStore {
                     ON parent.block_hash = last.parent_hash
                 JOIN bitcoin_blockchain anchor
                     ON anchor.block_hash = parent.bitcoin_anchor
+                    AND anchor.block_height >= $3
```
matteojug, line 1552: Isn't this redundant, having `bitcoin_blockchain` already filtering for this (in `bitcoin_blockchain_until`)?

signer/src/testing/blocks.rs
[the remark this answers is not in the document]
matteojug: also to `let new_block = self.0.last().unwrap().new_child().anchored_to(anchor);`?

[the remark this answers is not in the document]
cylewitruk: sure: [71a0d19](https://github.com/stacks-network/sbtc/pull/1397/commits/71a0d1976ecd3a6869d4478bc2258e26b6f6f95d)

signer/src/storage/postgres.rs
cylewitruk: Yup, removed in [79a2e95](https://github.com/stacks-network/sbtc/pull/1397/commits/79a2e95507fe3912717f7ae3de8c464a55b5bdb1)

signer/src/storage/postgres.rs
cylewitruk: This now handles `N-1` per the Slack answers we got: [d49a0c6](https://github.com/stacks-network/sbtc/pull/1397/commits/d49a0c603f42980fb1745e3d198bd9f1851b9a52)

signer/tests/integration/postgres.rs
```diff
@@ -5399,3 +5399,1200 @@ async fn pending_rejected_withdrawal_already_accepted() {
...
+            .await
+            .expect("failed to write transaction");
+        db.write_bitcoin_transaction(&bitcoin_sweep_tx)
+            .await
+            .expect("failed to write bitcoin transaction");
+
+        // Write a fully validated withdrawal output for the request.
+        db.write_bitcoin_withdrawals_outputs(&[model::BitcoinWithdrawalOutput {
+            bitcoin_txid: bitcoin_sweep_tx.txid,
+            bitcoin_chain_tip: *at_bitcoin_block,
```
matteojug, line 5686: nit: `bitcoin_chain_tip` cannot be the same as `bitcoin_sweep_tx` and `transaction` `block_hash` because of causality, but I guess we don't compare them anywhere; anyway, nothing to do, just flagging it.

signer/tests/integration/postgres.rs
```diff
@@ -5580,3 +5580,1200 @@ async fn is_withdrawal_inflight_catches_withdrawals_in_package() {
...
+        let requests = db
+            .get_pending_accepted_withdrawal_requests(
+                bitcoin_chain_tip.as_ref(),
+                &stacks_chain_tip,
+                3,
+                signature_threshold,
+            )
+            .await
+            .expect("failed to query db");
+        assert!(requests.is_empty(), "min height: 1");
```
matteojug, line 5814:
```suggestion
        assert!(requests.is_empty(), "min height: 3");
```

signer/tests/integration/postgres.rs
```diff
@@ -5580,3 +5580,1200 @@ async fn is_withdrawal_inflight_catches_withdrawals_in_package() {
...
+    /// Asserts that requests which are confirmed in stacks blocks whose anchor
+    /// block is, while in the canonical bitcoin chain, too old to be considered
+    /// accepted and are not returned.
+    ///
+    /// The signature threshold/context window we are testing is _inclusive_.
+    ///
+    /// This test creates blockchains with the following structure:
+    ///
+    /// ```text
+    /// Age:         3           2           1
```
matteojug, line 5724: nit: age it's confusing since we deal with heights below

signer/tests/integration/postgres.rs
```diff
@@ -5580,3 +5580,1200 @@ async fn is_withdrawal_inflight_catches_withdrawals_in_package() {
...
+        assert!(requests.is_empty());
+
+        signer::testing::storage::drop_db(db).await;
+    }
+
+    /// Asserts that requests which are confirmed in stacks blocks whose anchor
+    /// block is, while in the canonical bitcoin chain, too old to be considered
+    /// accepted and are not returned.
+    ///
+    /// The signature threshold/context window we are testing is _inclusive_.
```
matteojug, line 5719:
```suggestion
    /// The signature threshold/min height we are testing is _inclusive_.
```

signer/tests/integration/postgres.rs
```diff
@@ -5580,3 +5580,1200 @@ async fn is_withdrawal_inflight_catches_withdrawals_in_package() {
...
+    ///
+    /// This test creates blockchains with the following structure:
+    ///
+    /// ```text
+    /// Age:         3           2           1     
+    ///          ┌────────┐  ┌────────┐  ┌────────┐
+    /// Bitcoin: │   B1   ├──►   B2   ├──►   B3   │ The request is confirmed (✔)
+    ///          └─▲──────┘  └─▲──────┘  └─▲──────┘ in S2 and we test different
+    ///          ┌─┴──────┐  ┌─┴──────┐  ┌─┴──────┐ context windows.
```
matteojug, line 5728:
```suggestion
    ///          ┌─┴──────┐  ┌─┴──────┐  ┌─┴──────┐ min heights.
```

signer/tests/integration/postgres.rs
```diff
@@ -5580,3 +5580,1200 @@ async fn is_withdrawal_inflight_catches_withdrawals_in_package() {
...
+                &stacks_chain_tip,
+                min_block_height,
+                signature_threshold,
+            )
+            .await
+            .expect("failed to query db");
+        assert_eq!(requests.len(), 1);
+
+        // Reject the withdrawal request in B2a/S2a (canonical).
+        reject_withdrawal_request(&db, &request, &stacks_2a).await;
```
matteojug, line 6023: nit: can we check that `requests.is_empty()` also right after this? To be sure that rejected because of this and not some strange interaction of the two rejections

signer/tests/integration/postgres.rs
```diff
@@ -5580,3 +5580,1200 @@ async fn is_withdrawal_inflight_catches_withdrawals_in_package() {
...
+                &bitcoin_block.block_hash,
+                &stacks_block.block_hash,
+                min_block_height,
+                signature_threshold,
+            )
+            .await
+            .expect("failed to query db");
+
+        assert_eq!(requests.len(), 1);
+        assert_eq!(requests.pop().unwrap(), withdrawal_request_2);
```
matteojug, line 6149: nit: there's a very cool `.single()` we can use in those cases 😄

signer/tests/integration/postgres.rs
```diff
@@ -5580,3 +5580,1200 @@ async fn is_withdrawal_inflight_catches_withdrawals_in_package() {
...
+            .await
+            .expect("failed to query db");
+
+        // We expect to receive requests `1`, `2a`, and `3a`. `2b` should be
+        // orphaned.
+        assert_eq!(requests.len(), 3);
+        // We assert them in reverse order since we use `pop()` to remove them.
+        assert_eq!(requests.pop().unwrap(), request_3a);
+        assert_eq!(requests.pop().unwrap(), request_2a);
+        assert_eq!(requests.pop().unwrap(), request_1);
```
matteojug, line 6242: nit?
```suggestion
        assert_eq!(requests, [request_1, request_2a, request_3a]);
```

signer/tests/integration/postgres.rs
```diff
@@ -5580,3 +5580,1200 @@ async fn is_withdrawal_inflight_catches_withdrawals_in_package() {
...
+            .get_pending_accepted_withdrawal_requests(
+                bitcoin_chain_tip.as_ref(),
+                &stacks_chain_tip,
+                min_block_height,
+                signature_threshold,
+            )
+            .await
+            .expect("failed to query db");
+        assert_eq!(requests.len(), 1);
+        assert_eq!(requests.pop().unwrap(), request);
```
matteojug, line 6396: nit: `.single()`?

signer/tests/integration/postgres.rs
```diff
@@ -5580,3 +5580,1200 @@ async fn is_withdrawal_inflight_catches_withdrawals_in_package() {
...
+            .get_pending_accepted_withdrawal_requests(
+                bitcoin_chain_tip.as_ref(),
+                &stacks_chain_tip,
+                min_block_height,
+                signature_threshold,
+            )
+            .await
+            .expect("failed to query db");
+        assert_eq!(requests.len(), 1);
+        assert_eq!(requests.pop().unwrap(), expected);
```
matteojug, line 6541: nit: `.single()`?

signer/tests/integration/postgres.rs
```diff
@@ -5580,3 +5580,1200 @@ async fn is_withdrawal_inflight_catches_withdrawals_in_package() {
...
+            [&stacks_1, &stacks_2, &stacks_3],
+        )
+        .await;
+
+        // Get our chain tips and assert they're what we expect.
+        let (bitcoin_chain_tip, stacks_chain_tip) = db.get_chain_tips().await;
+        assert_eq!(bitcoin_chain_tip.as_ref(), &bitcoin_3.block_hash);
+        assert_eq!(&stacks_chain_tip, &stacks_3.block_hash);
+
+        // Create a withdrawal request that has a bitcoin block height of `2`,
```
matteojug, line 6723:
```suggestion
        // Create a withdrawal request that has a bitcoin block height of `0`,
```

matteojug (approved):
Nice tests!

signer/tests/integration/postgres.rs
cylewitruk: Yeah I think I was just putting values that satisfied FK's and were valid for the canonical chain without needing to create extra stuff since it wasn't needed for the validations these tests are doing; but that was days ago, don't remember exactly the thought process 😅

signer/tests/integration/postgres.rs
cylewitruk: That case was handled in `request_with_confirmed_rejection_not_returned` so I didn't feel the need to re-test it here. Maybe a comment pointing that out?

signer/tests/integration/postgres.rs
cylewitruk: Oh, right 😅

signer/tests/integration/postgres.rs
cylewitruk: Oh, right, leftover from the previous version

signer/tests/integration/postgres.rs
matteojug: It was to be extra pedantic in building up the scenario, but it's fine as is; I don't think a comment it's adding value (since it's not code and we need to trust what it says, nothing really relevant for this test itself). Anyway, resolving.

signer/tests/integration/postgres.rs
cylewitruk: [eceeb7f](https://github.com/stacks-network/sbtc/pull/1397/commits/eceeb7f6db1f6392f9363e0135f883301a7241d4)

signer/tests/integration/postgres.rs
cylewitruk: [eceeb7f](https://github.com/stacks-network/sbtc/pull/1397/commits/eceeb7f6db1f6392f9363e0135f883301a7241d4)

signer/tests/integration/postgres.rs
cylewitruk: [eceeb7f](https://github.com/stacks-network/sbtc/pull/1397/commits/eceeb7f6db1f6392f9363e0135f883301a7241d4)

signer/tests/integration/postgres.rs
cylewitruk: [eceeb7f](https://github.com/stacks-network/sbtc/pull/1397/commits/eceeb7f6db1f6392f9363e0135f883301a7241d4)

Issue #568, opened 2024-09-25
[Feature]: Create `reject-withdrawal-request` transactions


djordon:
# Feature - Create `reject-withdrawal-request` transactions

## 1. Description

The signer coordinator needs to put together stacks transactions for finalizing a withdrawal by creating an `reject-withdrawal` transaction. This ticket is for generating the stacks transaction that can pass validation by other signers.

### 1.1 Context & Purpose

We need to actually generate the stacks transaction before we could submit it for other signers to sign. Unfortunately, we do not have the code for validating the `reject-withdrawal-request` contract call.

### 2.1 Acceptance Criteria:
   - [ ] We have the code that creates a `reject-withdrawal-request` smart contract call object (the `RejectWithdrawalV1` struct) for withdrawals that won't be swept out.

## 3. Related Issues and Pull Requests (optional):

* https://github.com/stacks-network/sbtc/issues/478
* https://github.com/stacks-network/sbtc/issues/563
* https://github.com/stacks-network/sbtc/issues/564

## 4. Appendix

We should also consider the policy changes proposed in https://github.com/stacks-network/sbtc/issues/620.

PR #1398, opened 2025-02-19
feat: add withdrawal rejection to the coordinator loop

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/568

## Changes

Add the withdrawal rejection to the coordinator event loop. Also add a query to fetch all the valid withdrawal outputs, required to check if any is in mempool and we want to skip rejection.

## Testing Information

Add a test checking the _off-query_ conditions for skipping a rejection (already accepted in the contract or output tx in mempool).

## Checklist:

### DIFF
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -2268,6 +2268,35 @@ impl super::DbRead for PgStore {
         .await
         .map_err(Error::SqlxQuery)
     }
+
+    async fn get_valid_withdrawal_outputs(
+        &self,
+        id: &model::QualifiedRequestId,
+    ) -> Result<Vec<model::BitcoinWithdrawalOutput>, Error> {
+        sqlx::query_as::<_, model::BitcoinWithdrawalOutput>(
+            r#"
+            SELECT
+                bwo.bitcoin_txid
+              , bwo.bitcoin_chain_tip
+              , bwo.output_index
+              , bwo.request_id
+              , bwo.stacks_txid
+              , bwo.stacks_block_hash
+              , bwo.validation_result
+              , bwo.is_valid_tx
+            FROM sbtc_signer.bitcoin_withdrawals_outputs AS bwo
+            WHERE bwo.request_id = $1
+              AND bwo.stacks_block_hash = $2
+              AND bwo.validation_result = 'ok'
+              AND bwo.is_valid_tx = true
+            "#,
+        )
+        .bind(i64::try_from(id.request_id).map_err(Error::ConversionDatabaseInt)?)
+        .bind(id.block_hash)
+        .fetch_all(&self.0)
+        .await
+        .map_err(Error::SqlxQuery)
+    }
 }
 
 impl super::DbWrite for PgStore {
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -2269,7 +2269,7 @@ impl super::DbRead for PgStore {
         .map_err(Error::SqlxQuery)
     }
 
-    async fn get_valid_withdrawal_outputs(
+    async fn get_withdrawal_outputs(
         &self,
         id: &model::QualifiedRequestId,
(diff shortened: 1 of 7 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -732,9 +737,102 @@ where
...
+
+        for output in valid_outputs {
+            let mempool_entry = bitcoin
+                .get_mempool_entry(&output.bitcoin_txid.into())
+                .await?;
+            if mempool_entry.is_some() {
+                // There's an output for this request in the mempool, let's wait
+                return Ok(());
+            }
+        }
```
matteojug, line 793: Should we also check if the tx is confirmed? eg in case it's not in the mempool because it was just accepted but we are still missing the notification from zmq.

signer/src/storage/postgres.rs
```diff
@@ -2268,6 +2268,35 @@ impl super::DbRead for PgStore {
...
+              , bwo.request_id
+              , bwo.stacks_txid
+              , bwo.stacks_block_hash
+              , bwo.validation_result
+              , bwo.is_valid_tx
+            FROM sbtc_signer.bitcoin_withdrawals_outputs AS bwo
+            WHERE bwo.request_id = $1
+              AND bwo.stacks_block_hash = $2
+              AND bwo.validation_result = 'ok'
+              AND bwo.is_valid_tx = true
```
djordon, line 2291: Hmmm, I'm not sure if we want to filter on these values. It's possible that the current signer disapproves of the transaction for some reason but the consensus majority think that everything is fine. We'd want this signer to scan the mempool for that transaction as well.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3413,3 +3420,268 @@ async fn test_conservative_initial_sbtc_limits() {
         testing::storage::drop_db(db).await;
     }
 }
+
+#[test_case(false, false; "rejectable")]
+#[test_case(true, false; "completed")]
+#[test_case(false, true; "in mempool")]
+#[tokio::test]
+async fn process_rejected_withdrawal(is_completed: bool, is_in_mempool: bool) {
```
djordon, line 3428: Nice, awesome.

djordon (commented):
Largely looks good, the only thing is that I don't think that we should filter for "valid" transactions that match the withdrawal request.

signer/src/transaction_coordinator.rs
djordon: Yeah, couldn't hurt right. And it's an edge case that we need to consider. I think `BitcoinInteract::get_tx` checks for the mempool or a block.

signer/src/storage/postgres.rs
matteojug: I was thinking it was fine since worst case the coordinator will go ahead with the signing round but the signers will prevent it (as they will need to do the same check, but they would have validated the outputs), but since it's cheap to check for the tx and we know the amount of them is kinda bounded under reasonable conditions (~18 + the number of consecutive blocks the signers RBFs without a confirmation) I agree that erring on checking more of them is better (and removed that check in [bc3ea8f](https://github.com/stacks-network/sbtc/pull/1398/commits/bc3ea8f216f14087f02b6c731ad7348e57e9d1bf)).

signer/src/transaction_coordinator.rs
matteojug: Done in [bc3ea8f](https://github.com/stacks-network/sbtc/pull/1398/commits/bc3ea8f216f14087f02b6c731ad7348e57e9d1bf).

signer/src/storage/postgres.rs
djordon: Ahhh I see. Yeah it probably doesn't matter since the other signers will probably catch this. But yeah it doesn't hurt much to do.

djordon (approved):
Looks good ✅

PR #1399, opened 2025-02-19
feat: filtering of withdrawal requests according to consensus rules in the tx-coordinator

cylewitruk:
## Description

Closes: #1321

## Changes

Implements validations, together with #1397, for withdrawal requests against the consensus rules defined in #741, as well as related tests.

I spent a while attempting to use the new validation reports, but it was a bit of a "square peg for this round hole" as those are tailored for individual requests and a "signer" context, and we're already doing a lot of the heavy query-lifting in the `get_pending_accepted_withdrawal_requests` query. I'm all for re-working this in the future, but we should convene on a design/strategy around centralizing validations first; ideally something where the individual rules are unit testable and the rule-sets can be conditionally constructed based the current use-case.

## Testing Information

Added additional tests for the coordinator-side of things.

## Checklist:


signer/src/testing/blocks.rs
cylewitruk, line 1: This file is in #1397, review there.

signer/src/transaction_coordinator.rs
```diff
@@ -1577,11 +1600,217 @@ where
...
+        let mut eligible_withdrawals = Vec::new();
+
+        // We set the context window for the pending-accepted query below to the
+        // number of blocks that the withdrawal request is considered
+        // valid (not expired). This limits the number of blocks the query will
+        // consider when fetching pending withdrawal requests.
+        let context_window = withdrawal_blocks_expiry;
+
+        // Fetch pending withdrawal requests from storage. This method performs
+        // the following filtering according to consensus rules:
```
djordon, line 1647: What I am suggesting needs some work, but it is closer to being true. What is written here suggests that the function makes sure that the returned requests are not expired. Maybe try something like this?
```suggestion
        // Fetch pending withdrawal requests from storage. This method, with
        // the given inputs, performs the following filtering according to
        // consensus rules:
```

signer/src/transaction_coordinator.rs
```diff
@@ -1577,11 +1600,217 @@ where
...
+        const SKIP_REASON_INSUFFICIENT_CONFIRMATIONS: &str = "insufficient_confirmations";
+        const SKIP_REASON_INSUFFICIENT_VOTES: &str = "insufficient_votes";
+
+        let mut eligible_withdrawals = Vec::new();
+
+        // We set the context window for the pending-accepted query below to the
+        // number of blocks that the withdrawal request is considered
+        // valid (not expired). This limits the number of blocks the query will
+        // consider when fetching pending withdrawal requests.
+        let context_window = withdrawal_blocks_expiry;
```
djordon, line 1644: I find the assignment of the `context_window` like this less clear than forgoing the creation of a `context_window` variable and using `withdrawal_blocks_expiry` directly.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3413,3 +3413,239 @@ async fn test_conservative_initial_sbtc_limits() {
...
+        votes: &[bool],
+    ) {
+        let signer_votes = signer_set.signer_keys().iter().zip(votes.iter());
+        for vote in signer_votes {
+            let signer = model::WithdrawalSigner {
+                request_id: request.request_id,
+                block_hash: request.block_hash,
+                txid: request.txid,
+                signer_pub_key: *vote.0,
+                is_accepted: *vote.1,
```
djordon, line 3465: suggestion: destructure the tuple like
```rust
for (signer_pub_key, is_accepted) in signer_votes {
    ...
```

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3413,3 +3413,239 @@ async fn test_conservative_initial_sbtc_limits() {
...
+        signature_threshold: 2,
+        num_approves: 1,
+        amount: 1_000,
+        sbtc_limits: SbtcLimits::unlimited(),
+        chain_length: 8,
+        at_block_height: 2,
+        num_expected_results: 0,
+    }; "insufficient_votes")]
+    #[test_log::test(tokio::test)]
+    async fn test_validations(params: TestParams) {
```
djordon, line 3606: Nice, it's not easy getting to the place where the tests are easy two follow but require so much setup. Good job.

signer/src/testing/transaction_coordinator.rs
```diff
@@ -255,7 +256,10 @@ where
             .expect("Empty pending requests");
         let withdrawals = pending_requests.withdrawals;
 
-        // Get pending withdrawals from storage
+        // Get the chain tips from storage.
+        let (bitcoin_chain_tip_ref, stacks_chain_tip) = storage.get_chain_tips().await;
```
djordon, line 260: This is already defined a few lines above.

djordon (commented):
This looks good. The comments are helpful and things are clean. I think I have only one substantive comment. Also, the tests were easy to follow after looking over the other two PRs. Nice work.

> I'm all for re-working this in the future, but we should convene on a design/strategy around centralizing validations first.

Yeah let's chat about it and tackle it if we think that there is something to do. I would really like to consolidate the validation logic so that it can all be found or changed in "one spot".

signer/src/transaction_coordinator.rs
cylewitruk: Updated a bit in [97a9f1a](https://github.com/stacks-network/sbtc/pull/1399/commits/97a9f1ab42e59919a0aabaf7c628e2c5d3c96136), is it clearer now?

signer/src/transaction_coordinator.rs
cylewitruk: Tweaked it a bit in [97a9f1a](https://github.com/stacks-network/sbtc/pull/1399/commits/97a9f1ab42e59919a0aabaf7c628e2c5d3c96136), thoughts?

signer/src/transaction_coordinator.rs
```diff
@@ -1577,11 +1600,217 @@ where
...
+    ///    Bitcoin chain.
+    /// 2. [x] The request must be confirmed in a canonical Stacks block.
+    /// 3. [x] The request must have reached the required number of Bitcoin
+    /// 4. [x] The request must be approved:
+    ///     - [x] By the required number of signers (this is implemented as a
+    ///       pre-filter in the query, any signer),
+    ///     - [x] And by the required number of signers _in the current signer
+    ///       set_.
+    /// 5. [ ] The request has been approved by this signer (this shouldn't be
+    ///    applicable to the coordinator).
```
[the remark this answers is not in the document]
cylewitruk: I tried to make it explicit in [97a9f1a](https://github.com/stacks-network/sbtc/pull/1399/commits/97a9f1ab42e59919a0aabaf7c628e2c5d3c96136), thoughts?

signer/src/transaction_coordinator.rs
```diff
@@ -1597,50 +1826,41 @@ where
...
+            // Fetch eligible deposit requests.
+            let deposits =
+                Self::get_eligible_pending_deposit_requests(&storage, self.context_window, &params)
+                    .await?;
 
-        let mut withdrawals = Vec::new();
+            // Fetch eligible withdrawal requests.
+            let withdrawal_expiry_window = WITHDRAWAL_BLOCKS_EXPIRY
+                .try_into()
+                .map_err(|_| Error::TypeConversion)?;
```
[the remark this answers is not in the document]
cylewitruk: Yeah I remember the discussion, wasn't sure if it was decided or not, but I like it. Should we increase the expiry period to 30 blocks then to not shorten the "effective" window?

Added in: [e03724c](https://github.com/stacks-network/sbtc/pull/1399/commits/e03724c49068a59204ce7eea5dc64426040f4608)

Tbh if we don't add it to consensus then we're, as you say, not solving a (potential) problem in the case where a coordinator doesn't apply this rule. But we can do that later if more thought/discussion is needed as well.

signer/tests/integration/transaction_coordinator.rs
cylewitruk: Yeah I'm still not in love with it; I would prefer to test each validation rule separately and have a better assertion, but time got to me so I manually verified the log output for each case instead for now (to make sure it was the correct skip reason).

signer/src/testing/transaction_coordinator.rs
cylewitruk: Indeed, merge gone wrong 😅 Removed in [8001b27](https://github.com/stacks-network/sbtc/pull/1399/commits/8001b27cce212a2586e986101622c163f3988a71) and also fixed a little with the new buffer value.

signer/src/transaction_coordinator.rs
```diff
@@ -1597,50 +1833,48 @@ where
...
+            // Calculate the withdrawal expiry window, which is effectively how
+            // many canonical bitcoin blocks back in time we will look for
+            // eligible requests. We subtract the withdrawal expiry buffer from
+            // the withdrawal expiry period to ensure that we don't include
+            // requests that are about to expire. See the constant descriptions
+            // for more details.
+            let withdrawal_expiry_window = WITHDRAWAL_BLOCKS_EXPIRY
+                .saturating_sub(WITHDRAWAL_EXPIRY_BUFFER)
+                .try_into()
+                .map_err(|_| Error::TypeConversion)?;
```
djordon, line 1862: We can make the difference here it's own constant.

signer/src/transaction_coordinator.rs
djordon: Yeah, this is better 👍🏽.

signer/src/transaction_coordinator.rs
djordon: 👍🏽

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
djordon: It's better than my suggestion and better than what's already here, so I'm okay with it. Let me try to rework it a bit to make it more concise.

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
djordon: I should have put my comment here: but we can make the difference here it's own constant of the right type, so that there is no conversion here either.

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
djordon: Well, how about this:
```
The request has been approved by this signer.

**Note:** This rule does not apply within the coordinator
module, where decisions are made collectively based on
consensus rules rather than an individual signer's approval.
However, the coordinator's signer module still processes the
request according to these same rules.
```
It's not terribly different. I'm okay with whichever you decide.

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
cylewitruk: I think I had tried making them `u16`'s but then the conversion was just pushed elsewhere in some other usage... I'll try it again

signer/src/transaction_coordinator.rs
djordon: Yeah, that works. Perfect.

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
cylewitruk: Sure, I'm not that picky ;) Updated to your suggestion in [2949b5e](https://github.com/stacks-network/sbtc/pull/1399/commits/2949b5e4b672a482da2ed5be760440677a06014e)

signer/src/transaction_coordinator.rs
cylewitruk: Added the const in [4d8eb68](https://github.com/stacks-network/sbtc/pull/1399/commits/4d8eb687e5c6e8b474f9c537c3b5991e1a247027) but it ended up being used in only one place 🙈

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
cylewitruk: This was done in the bigger [81b7e59](https://github.com/stacks-network/sbtc/pull/1399/commits/81b7e591bd681f654457b88effcab01a1d752acf) -- tbh I'm not sure it's worth it with all of the `into()`s and casts it requires (we compare to `u64` block height quite often) 🤔

signer/src/transaction_coordinator.rs
```diff
@@ -1674,76 +1697,295 @@ where
...
+                .block_height
+                .saturating_sub(req.bitcoin_block_height);
+
+            // [3] Ensure that we have the required number of confirmations for
+            // the withdrawal request.
+            if num_confirmations < min_confirmations.into() {
+                tracing::debug!(
+                    request_id = req.request_id,
+                    num_confirmations,
+                    required_confirmations = min_confirmations,
```
djordon, line 1838: I think this was changed by accident. Shouldn't it be:
```suggestion
            if num_confirmations < WITHDRAWAL_MIN_CONFIRMATIONS.into() {
                tracing::debug!(
                    request_id = req.request_id,
                    num_confirmations,
                    required_confirmations = WITHDRAWAL_MIN_CONFIRMATIONS,
```

cylewitruk: no it was on purpose, the value (`WITHDRAWAL_MIN_CONFIRMATIONS`) is passed in from `get_pending_requests()` to keep constants out of this method -- I prefer to have the constants "higher up" and passed down to the "worker methods"; it makes the method more flexible which is especially nice in tests.

signer/src/stacks/contracts.rs
```diff
@@ -1173,7 +1173,8 @@ impl AsContractCall for RejectWithdrawalV1 {
             .block_height
             .saturating_sub(report.bitcoin_block_height);
 
-        if blocks_observed <= WITHDRAWAL_BLOCKS_EXPIRY {
+        // 4. The request is expired.
+        if blocks_observed < WITHDRAWAL_BLOCKS_EXPIRY.into() {
```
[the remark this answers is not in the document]
cylewitruk: > @djordon: Yeah, we need to wait more than WITHDRAWAL_BLOCKS_EXPIRY blocks in order for a withdrawal request to be considered expired

signer/src/transaction_coordinator.rs
cylewitruk: which reminds me, there's still `WITHDRAWAL_DUST_LIMIT`. Do we need both this one and `DEPOSIT_DUST_LIMIT` which are the same value? Could they be consolidated into just `SBTC_DUST_LIMIT` and then it would fit perfectly into `GetPendingRequestsParams`.

signer/src/transaction_coordinator.rs
```diff
@@ -1672,76 +1695,295 @@ where
...
+        tracing::info!("preparing pending requests for processing");
+
+        let storage = self.context.get_storage();
+        let config = self.context.config();
+
+        // Get the current sBTC limits (caps).
+        let sbtc_limits = self.context.state().get_current_limits();
+
+        // Fetch pending deposit and withdrawal requests from storage.
+        let (deposits, withdrawals) = {
```
matteojug, line 1955: The body seems short enough to get rid of this and just compute the values, is there a reason to prefer the added inner scope here?

signer/src/lib.rs
```diff
@@ -125,6 +125,19 @@ pub const WITHDRAWAL_MIN_CONFIRMATIONS: u64 = 6;
...
+/// proposals.
+///
+/// This is the value in https://github.com/stacks-network/sbtc/issues/1363's
+/// proposed "buffer for expiring requests" section.
+pub const WITHDRAWAL_EXPIRY_BUFFER: u64 = 6;
+
+/// This is the number of bitcoin blocks that a withdrawal request will remain
+/// live before it is considered "soft expired". This is the point at which the
+/// coordinator will cease to include a withdrawal request in sweep proposals.
+pub const WITHDRAWAL_BLOCKS_SOFT_EXPIRY: u64 = WITHDRAWAL_BLOCKS_EXPIRY - WITHDRAWAL_EXPIRY_BUFFER;
```
matteojug, line 139: It seems we are only using this in a test (`assert_processes_withdrawals`), should we ditch the const and compute the difference there?

signer/src/transaction_coordinator.rs
```diff
@@ -1672,76 +1695,295 @@ where
...
-
-        let storage = self.context.get_storage();
-        let config = self.context.config();
-
-        // Get the current sBTC limits (caps).
-        let sbtc_limits = self.context.state().get_current_limits();
+    pub async fn get_eligible_pending_withdrawal_requests<DB>(
+        storage: &DB,
+        expiry_window: u64,
+        soft_expiry_window: u64,
```
matteojug, line 1724: nit: replace `soft_expiry_window` with `expiry_buffer`?

signer/src/transaction_coordinator.rs
```diff
@@ -1672,76 +1695,295 @@ where
...
+            .await?;
+
+        // If there are no pending deposit requests, we can exit early.
+        if pending_deposit_requests.is_empty() {
+            tracing::debug!("no pending deposit requests eligible for consideration found");
+            return Ok(eligible_deposits);
+        }
+
+        // Iterate through each deposit request, fetch its _actual_ votes from
+        // storage and
```
matteojug, line 1922: nit: cliffhanger

signer/src/transaction_coordinator.rs
```diff
@@ -1672,76 +1695,295 @@ where
...
             let votes = storage
-                .get_withdrawal_request_signer_votes(&req.qualified_id(), aggregate_key)
+                .get_deposit_request_signer_votes(&req.txid, req.output_index, params.aggregate_key)
                 .await?;
 
-            let withdrawal = utxo::WithdrawalRequest::from_model(req, votes);
-            withdrawals.push(withdrawal);
+            let deposit = utxo::DepositRequest::from_model(req, votes);
+            eligible_deposits.push(deposit);
         }
```
matteojug, line 1930: This is interesting: we are not checking how many votes we have for the signers in the signing set identified by `aggregate_key` -- I guess we are relying on them being filtered out in `compute_optimal_packages`, but we could prefilter them already.
But even if we did, we are probably using the wrong `aggregate_key`, since we should check the signing set identified by `req.signers_public_key` instead.
We should probably address those points, unless I'm missing something. WDYT?

signer/src/transaction_coordinator.rs
```diff
@@ -1672,76 +1695,295 @@ where
...
+        // Fetch pending withdrawal requests from storage. This method, with
+        // the given inputs, performs the following filtering according to
+        // consensus rules:
+        //
+        // - [1]  The request has not been swept in the canonical bitcoin chain,
+        // - [2]  Is confirmed in a canonical stacks block,
+        // - [4a] Is accepted by >= `threshold` signers (pre-filter),
+        // - [7]  Is not older than the provided withdrawal expiry threshold.
+        //
+        // We set the context window for the pending-accepted query below to the
```
matteojug, line 1763: nit: we no longer deal with a context window below, but it could be intended as the one induced by `min_bitcoin_height`, so it could stay as is. Feel free to resolve in that case.

signer/src/transaction_coordinator.rs
```diff
@@ -1672,76 +1695,295 @@ where
...
+                    min_soft_bitcoin_height,
+                    reason = SKIP_REASON_SOFT_EXPIRY,
+                    message = REQUEST_SKIPPED_MESSAGE
+                );
+                continue;
+            }
+
+            // [8] Ensure that the withdrawal request amount is above the dust
+            // limit.
+            if req.amount <= WITHDRAWAL_DUST_LIMIT {
```
matteojug, line 1798: Bitcoin validation consider valid an amount equal to `WITHDRAWAL_DUST_LIMIT`, we should align the two.

signer/src/transaction_coordinator.rs
```diff
@@ -1672,76 +1695,295 @@ where
...
+            if req.amount > per_withdrawal_cap {
+                tracing::debug!(
+                    request_id = req.request_id,
+                    amount = req.amount,
+                    per_withdrawal_cap = per_withdrawal_cap,
+                    reason = SKIP_REASON_PER_WITHDRAWAL_CAP_EXCEEDED,
+                    message = REQUEST_SKIPPED_MESSAGE
+                );
+                continue;
+            }
```
matteojug, line 1820: For deposits the cap check is done in `SbtcRequests`, should we do the same for withdrawals for consistency?

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3549,7 +3549,7 @@ async fn process_rejected_withdrawal(is_completed: bool, is_in_mempool: bool) {
         .is_empty());
 
     let new_tip = faucet
-        .generate_blocks(WITHDRAWAL_BLOCKS_EXPIRY + 1)
+        .generate_blocks(WITHDRAWAL_BLOCKS_EXPIRY as u64 + 1)
```
matteojug, line 3552: nit: leftover cast

signer/src/transaction_coordinator.rs
djordon: I misread this before, it looked like `if num_confirmations < num_confirmations.into()`, which I thought was the same variable being used accidentally 😅. I hastily drafted a comment instead of pausing to take a closer look, which is what I would normally do.

signer/src/lib.rs
cylewitruk: Yeah I was actually thinking the same; added it on daniel's request, but we assumed it would be used more widely. I'll remove it for now.

signer/src/transaction_coordinator.rs
cylewitruk: I actually think, when reading the method "from an outside perspective", that `soft_expiry_window` gives more immediate context as to its intent, and that maybe `expiry_window` should be renamed to `hard_expiry_window` or something. Thoughts?

signer/src/transaction_coordinator.rs
cylewitruk: Hm, interesting. I was following the text in #741, assumed the other places were too.

signer/src/transaction_coordinator.rs
cylewitruk: It was due to the borrow of `sbtc_limits` which is then moved when creating the `SbtcRequests` below, so the immutable borrow needs to drop first.

But thinking after, I could make that owned and then move it from `params` to `SbtcRequests` when done.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3727,3 +3727,307 @@ async fn process_rejected_withdrawal(is_completed: bool, is_in_mempool: bool) {
...
+        at_block_height: 4,
+        num_expected_results: 0,
+        ..Default::default()
+    }; "hard_expiry_one_block_too_old")]
+    #[test_log::test(tokio::test)]
+    async fn test_validations(params: TestParams) {
+        let db = testing::storage::new_test_database().await;
+
+        let (bitcoin_chain, stacks_chain, signer_set, signer_keys) =
+            test_setup(&db, params.chain_length + 1).await;
```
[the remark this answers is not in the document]
cylewitruk: Iirc it was to be able to translate to 1-indexed as the `nth_block(params.at_block_height)`'s below would panic if you specified `chain_length = 10` and tried to also set `at_block_height = 10` (= 11th element).

signer/src/testing/transaction_coordinator.rs
```diff
@@ -271,9 +272,39 @@ where
...
+            .block_height
+            .saturating_sub(crate::WITHDRAWAL_MIN_CONFIRMATIONS);
+        let min_processable_height = bitcoin_chain_tip
+            .block_height
+            .saturating_sub(crate::WITHDRAWAL_BLOCKS_SOFT_EXPIRY);
+
         // Assert that there are some withdrawals in storage while get_pending_requests return 0 withdrawals
         assert!(!withdrawals_in_storage.is_empty());
         for withdrawal in withdrawals_in_storage {
+            if withdrawal.bitcoin_block_height > max_processable_height {
```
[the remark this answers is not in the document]
cylewitruk: We already test that pretty good in several other integration tests; these filters were just to make it stop panicking (and for debugging why it was) so that I could move on 😅 

This test needs a revamp in any case as now it's duplicating some of what the other tests are doing, just worse. It should just create a couple of requests that will pass all validations and assert that it gets them back in the `SbtcRequests`; I just didn't take the time to do it yet.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3727,3 +3727,307 @@ async fn process_rejected_withdrawal(is_completed: bool, is_in_mempool: bool) {
...
+    }; "below_dust_limit")]
+    #[test_case(TestParams {
+        amount: 1_000,
+        sbtc_limits: SbtcLimits::zero(),
+        num_expected_results: 0,
+        ..Default::default()
+    }; "amount_over_per_withdrawal_limit")]
+    #[test_case(TestParams {
+        // This case will calculate the confirmations as:
+        // chain_length (10) - min_confirmations (4) = 4 (maximum block height),
```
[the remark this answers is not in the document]
cylewitruk: [6ff4487](https://github.com/stacks-network/sbtc/pull/1399/commits/6ff4487930b3a796612ec7f08045ed1def810a2b)

signer/tests/integration/transaction_coordinator.rs
cylewitruk: [6ff4487](https://github.com/stacks-network/sbtc/pull/1399/commits/6ff4487930b3a796612ec7f08045ed1def810a2b)

signer/src/transaction_coordinator.rs
cylewitruk: Nah it's better it's correct, reworded in [6ff4487](https://github.com/stacks-network/sbtc/pull/1399/commits/6ff4487930b3a796612ec7f08045ed1def810a2b)

signer/src/transaction_coordinator.rs
cylewitruk: /giphy stalone

[6ff4487](https://github.com/stacks-network/sbtc/pull/1399/commits/6ff4487930b3a796612ec7f08045ed1def810a2b)

signer/src/transaction_coordinator.rs
djordon: Hmmm, yeah the ticket states that it must be greater than some dust amount, which is correct. And I think we should align and use the bitcoin the validation logic. The "problem" is probably around the definitions of our constants and what they represent.

Here is some stuff that you guys probably know, but hopefully makes the "problem" clearer for others who are potentially following along.

---
Basically, transaction amounts spending below a certain amount is "dust" and bitcoin core will reject transactions that create outputs with "dust" amounts. The calculation of what is dust is configurable in bitcoin core -- it's not part of "consensus". And there is no constant value for what constitutes a dust amount for all outputs types, it depends on the script locking the output. We use 546 since it is non-dust for all standard output types.

So I think the issue is around the meaning and naming of `WITHDRAWAL_DUST_LIMIT` and `DEPOSIT_DUST_LIMIT`. They give the impression that the amount stated is a dust amount when it isn't. It's the minimum amount that is not dust for standard outputs in the eyes of bitcoin core using the default configuration values; 😅.

The smart contract is actually wrong here and we should change it; 546 is a valid amount (for bitcoin core using the default configuration values).

---
So, I think we should change this to `<`. We should should fix up the constants and their descriptions too but that doesn't need to happen in this PR. And eventually we'll update the smart contracts.

signer/tests/integration/transaction_coordinator.rs
cylewitruk: [b88da7c](https://github.com/stacks-network/sbtc/pull/1399/commits/b88da7c88b71db22f3cde725a505c4cc564d5d55)

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -125,7 +121,7 @@ async fn reject_withdrawal_validation_happy_path() {
     // Normal: We do not reject a withdrawal requests until more than
     // WITHDRAWAL_BLOCKS_EXPIRY blocks have been observed since the smart
     // contract that created the withdrawal request has bene observed.
-    faucet.generate_blocks(WITHDRAWAL_BLOCKS_EXPIRY + 1);
+    faucet.generate_blocks(WITHDRAWAL_BLOCKS_EXPIRY as u64 + 1);
```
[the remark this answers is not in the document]
cylewitruk: [b88da7c](https://github.com/stacks-network/sbtc/pull/1399/commits/b88da7c88b71db22f3cde725a505c4cc564d5d55)

signer/tests/integration/postgres.rs
```diff
@@ -4936,7 +4936,7 @@ async fn pending_rejected_withdrawal_no_events() {
 
     // Append some blocks to ensure we have expired requests; we expire the
     // requests in the first 5 canonical blocks, while keeping the others valid.
-    for _ in chain_depth..WITHDRAWAL_BLOCKS_EXPIRY + 5 {
+    for _ in chain_depth..WITHDRAWAL_BLOCKS_EXPIRY as u64 + 5 {
```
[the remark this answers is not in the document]
cylewitruk: [b88da7c](https://github.com/stacks-network/sbtc/pull/1399/commits/b88da7c88b71db22f3cde725a505c4cc564d5d55)

signer/src/bitcoin/validation.rs
```diff
@@ -1540,7 +1540,7 @@ mod tests {
             recipient: ScriptBuf::new(),
             bitcoin_block_height: 0,
         },
-        chain_tip_height: WITHDRAWAL_BLOCKS_EXPIRY + 1,
+        chain_tip_height: WITHDRAWAL_BLOCKS_EXPIRY as u64 + 1,
```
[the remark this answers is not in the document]
cylewitruk: [b88da7c](https://github.com/stacks-network/sbtc/pull/1399/commits/b88da7c88b71db22f3cde725a505c4cc564d5d55)

signer/src/bitcoin/validation.rs
```diff
@@ -1431,7 +1431,7 @@ mod tests {
             bitcoin_block_height: 0,
         },
         // This is part of sBTC consensus.
-        chain_tip_height: WITHDRAWAL_MIN_CONFIRMATIONS,
+        chain_tip_height: WITHDRAWAL_MIN_CONFIRMATIONS.into(),
```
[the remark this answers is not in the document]
cylewitruk: [b88da7c](https://github.com/stacks-network/sbtc/pull/1399/commits/b88da7c88b71db22f3cde725a505c4cc564d5d55)

signer/src/lib.rs
cylewitruk: Removed it in [b88da7c](https://github.com/stacks-network/sbtc/pull/1399/commits/b88da7c88b71db22f3cde725a505c4cc564d5d55)

signer/src/transaction_coordinator.rs
cylewitruk: It was due to some other borrowed field I had and removed after-the-fact I guess, because it worked fine now :) Changed in [55214dd](https://github.com/stacks-network/sbtc/pull/1399/commits/55214dd1d869dfc1bf2158d41b643be04c2a30bb)

signer/src/transaction_coordinator.rs
cylewitruk: Alright, yeah just reading the const description, issue text and the contract, I didn't land in the same conclusion ;)

Updated to `<` and adjusted comment/tests in [65329cf](https://github.com/stacks-network/sbtc/pull/1399/commits/65329cfd2f6b1565d8f1df40c2c9104e3b120bd8)

I mentioned it elsewhere, but I think we should just define one dust amount const, and call it something very clear, something along the lines of `BITCOIN_DUST_AMOUNT = 545`, and maybe a convenience const `BITCOIN_MIN_AMOUNT = BITCOIN_DUST_AMOUNT + 1`. 🤷

signer/src/transaction_coordinator.rs
cylewitruk: That const is now removed as it was only used in one test

signer/src/testing/transaction_coordinator.rs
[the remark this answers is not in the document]
matteojug: I would still suggest to add the checks: this is testing a more end to end thingy, including the correct usage of the constants in the coordinator at this point, something that the unit tests do not cover (unless I'm missing where).

signer/src/transaction_coordinator.rs
matteojug: I mainly flagged this because of the constant we passed in, now that we removed `WITHDRAWAL_BLOCKS_SOFT_EXPIRY` it could be fine leaving as is.
I find it still a bit confusing since _window_ refer to something slightly different in the two cases so it could be a bit confusing at first -- ie, `soft_expiry_window` is a window relative to `expiry_window` and not the request height -- and in the const (and related discussion) we call it a buffer.
But anyway, no strong opinions about it, do as you prefer.

signer/src/transaction_coordinator.rs
matteojug: mmm I don't know, but yeah agreed that this is something we can tackle at another time, for now it's enough to ensure we have all the rules in place 😅

signer/tests/integration/transaction_coordinator.rs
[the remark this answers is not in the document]
matteojug: Ah that makes sense, thanks! Do you mind tossing a comment in there about this? Just to make it clear that the `TestParams` in the test cases above are actually correct and this is just an artefact of the first block having a height of 0.

signer/src/testing/transaction_coordinator.rs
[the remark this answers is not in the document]
cylewitruk: No, you're right, there aren't currently any tests testing explicitly the usage of the constants; I added the asserts in [2fa00f0](https://github.com/stacks-network/sbtc/pull/1399/commits/2fa00f00df07160ce7464480cd85b628f4fc3b94).

What I mean with that this test is doing a poor job is that it uses `TestData` and parameters from `transaction_coordinator_test_environment()`, so the checks that I've added here handle what the seeded rng spits out for the moment, but changing either the seed or the `TestData` params would likely cause it to fail in other ways (like on amount). So we really should have separate, more controlled tests testing the in/out.

signer/src/transaction_coordinator.rs
matteojug: mmm but the current signing set may not be able to (fully) sign for an old deposit if that was spent to a different aggregate key (and signing set); on the compliance level I think it would still be the same, as if you don't want to sign for an input you would just not sign for it (and we record the bitmap, so no foul play there).

signer/src/transaction_coordinator.rs
cylewitruk: Oh I see what you mean, okay yeah I agree with you, what's passed here isn't a window.

I think it would be clearer if the calculation for the "soft expiry height" were done in `get_pending_requests()` instead and that both were passed in as absolute block heights, instead of mixing the two concepts. But to avoid changing that right now (and tests) I'll just rename this to buffer 👍

signer/tests/integration/transaction_coordinator.rs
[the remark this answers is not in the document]
cylewitruk: Sure, added [ad64214](https://github.com/stacks-network/sbtc/pull/1399/commits/ad642145a221d24f53b6a0d7c6c6778e265b3190)

signer/src/transaction_coordinator.rs
matteojug: Opened https://github.com/stacks-network/sbtc/issues/1412 for the second point, something for another day.

cylewitruk: Alright, let's tackle that in a separate PR so we don't mess around with deposit validation logic here; it'll probably break some test somewhere too.

I guess you're creating a ticket for the 2nd point :)

signer/src/transaction_coordinator.rs
cylewitruk: Tweaked in [5c4edf1](https://github.com/stacks-network/sbtc/pull/1399/commits/5c4edf14c0e5780d2a8dda6d86839e89fce235cd)

matteojug (approved):

signer/src/transaction_coordinator.rs
djordon: Yeah that makes sense. I think we can name these `BITCOIN_DUST_THRESHOLD` or `BITCOIN_MIN_AMOUNT`.

I think I added both of those constants because the smart contracts could have different limits -- they don't but they actually should. The deposit limit should be 1, so basically no dust limit, while the withdrawal limit should be whatever bitcoin enforces, so it should depend on the recipient. But adding them both was premature and we should simplify things if we can.

djordon (approved):
Looks good 🚀

PR #1404, opened 2025-02-20
fix: swept withdrawal in case of forked accepts

matteojug:
## Description

Closes: #?

## Changes

Fix the swept withdrawal query to cover for the case of forked events: if we have a canonical accept event and a forked one, we should still consider the request as accepted (and not return it).

## Testing Information

Add a check in an existing test.

## Checklist:


djordon (approved):
Looks good!

Issue #1374, opened 2025-02-18
[Feature]: Check mempool for withdrawal outputs before rejection


djordon:
# Feature - Check mempool for withdrawal outputs before rejection

## 1. Description

It's possible for a sweep transaction to be servicing a withdrawal request at the time that it expires. In this case, we must not reject the withdrawal request since that transaction could be confirmed, fulfilling the withdrawal request.

### 1.1 Context & Purpose

We must not fulfill a withdrawal request and reject it at the same time since that would de-peg sBTC and BTC. So we must do out best to check for withdrawal requests in the mempool.

## 2. Technical Details:

We should use bitcoin core RPCs and our database to figure out whether withdrawal request outputs are in sweep transactions in the mempool. There are three approaches that we can take:
1. If there is a transaction in `bitcoin_withdrawals_outputs` that is associated with our withdrawal request and is spending the signers' UTXO, assume that the transaction it is in the mempool,
2. Use transactions in the `bitcoin_withdrawals_outputs` table to find candidate transaction IDs and scan the mempool for any of them.
3. Look for transactions in the mempool that match the recipient and filter for sBTC sweep transactions.

In all three cases, if there is a match then the `reject-withdrawal-request` contract call must fail validation. We should almost certainly do (1).

### 2.1 Acceptance Criteria:
   - [ ] `reject-withdrawal-request` contract calls fail validation if we suspect that there is a transaction in the mempool servicing the request.

PR #1406, opened 2025-02-21
feat: check mempool for withdrawal outputs before rejection

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1374

## Changes

* Add a query that checks whether the signer thinks it's possible for an output in the mempool to fulfill a withdrawal request.
* Fail `withdrawal-reject` validation if the signer suspects that there is a sweep transaction in the mempool that is fulfilling a withdrawal request.
* Change the `withdrawal-reject` tests to remove unnecessary actions so that it's easier to see that they fail only for the difference condition that they are testing.

## Testing Information

This PR adds one new integration test for the new condition in the `RejectWithdrawalV1::validation` function, and two new integration tests for the new query.

## Checklist:

### DIFF
--- a/signer/src/storage/mod.rs
+++ b/signer/src/storage/mod.rs
@@ -324,9 +324,12 @@ pub trait DbRead {
     ///
     /// # Notes
     ///
-    /// The tables that would be able to answer whether a withdrawal is
-    /// "live" in the mempool are populated during validation of pre-sign
-    /// requests.
+    /// At this time, we cannot use the bitcoin mempool for this
+    /// information since we cannot match withdrawals with transaction
+    /// outputs unless a lot of computational effort. Instead we use our
+    /// database, where the query is straightforward. The tables that are
+    /// be able to answer whether a withdrawal is potentially in the
+    /// mempool are populated during validation of pre-sign requests.
     fn is_withdrawal_inflight(
         &self,
         id: &model::QualifiedRequestId,
--- a/signer/src/storage/mod.rs
+++ b/signer/src/storage/mod.rs
@@ -320,10 +320,10 @@ pub trait DbRead {
     ) -> impl Future<Output = Result<bool, Error>> + Send;
 
     /// Check whether it is possible that the withdrawal identified by the
-    /// given request identifier is currently in the mempool, assuming that
-    /// we have processed all recent pre-sign requests and stored the
+    /// given request identifier live in the bitcoin mempool. This assumes
+    /// that we have processed all recent pre-sign requests and stored the
     /// results into our database.
-    fn is_withdrawal_in_mempool(
+    fn is_withdrawal_live(
         &self,
         id: &model::QualifiedRequestId,
         bitcoin_chain_tip: &model::BitcoinBlockHash,
--- a/signer/src/storage/mod.rs
+++ b/signer/src/storage/mod.rs
@@ -319,11 +319,15 @@ pub trait DbRead {
         script: &model::ScriptPubKey,
     ) -> impl Future<Output = Result<bool, Error>> + Send;
 
-    /// Check whether it is possible that the withdrawal identified by the
-    /// given request identifier live in the bitcoin mempool. This assumes
-    /// that we have processed all recent pre-sign requests and stored the
-    /// results into our database.
-    fn is_withdrawal_live(
+    /// Returns whether the identified withdrawal may be included in a
+    /// sweep transaction that is in the bitcoin mempool.
+    ///
+    /// # Notes
+    ///
+    /// The tables that would be able to answer whether a withdrawal is
+    /// "live" in the mempool are populated during validation of pre-sign
+    /// requests.
+    fn is_withdrawal_inflight(
         &self,
         id: &model::QualifiedRequestId,
         bitcoin_chain_tip: &model::BitcoinBlockHash,
--- a/signer/src/storage/mod.rs
+++ b/signer/src/storage/mod.rs
@@ -319,6 +319,16 @@ pub trait DbRead {
         script: &model::ScriptPubKey,
     ) -> impl Future<Output = Result<bool, Error>> + Send;
 
+    /// Check whether it is possible that the withdrawal identified by the
+    /// given request identifier is currently in the mempool, assuming that
+    /// we have processed all recent pre-sign requests and stored the
+    /// results into our database.
+    fn is_withdrawal_in_mempool(
+        &self,
+        id: &model::QualifiedRequestId,
+        bitcoin_chain_tip: &model::BitcoinBlockHash,
+    ) -> impl Future<Output = Result<bool, Error>> + Send;
+
     /// Fetch the bitcoin transaction that is included in the block
     /// identified by the block hash.
     fn get_bitcoin_tx(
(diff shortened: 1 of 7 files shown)


djordon:
> Do you think it also will be useful to ask bitcoin client about mempool, something like `ctx.get_bitcoin_client().find_mempool_transactions_spending_output(outpoint)`? Or if this function return some txid then your query also will return it?

I don't think that it would be so useful in this case. There are two issues:
1. If we don't get a response from our bitcoin node when making the RPC, that doesn't mean that there aren't any sweep transactions fulfilling the withdrawal in some other node's mempool. Our bitcoin node might not have received word of the transaction over the p2p network.
2. If we do get a response from our bitcoin node for the RPC, we would need some way of correlating the results to a withdrawal. The only way to do that is to use the `bitcoin_withdrawal_outputs` table.

In either case, we need to do something else to be "more sure" that a withdrawal isn't being fulfilled.

signer/src/storage/postgres.rs
```diff
@@ -2070,6 +2070,57 @@ impl super::DbRead for PgStore {
         .map_err(Error::SqlxQuery)
     }
 
+    async fn is_withdrawal_inflight(
```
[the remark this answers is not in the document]
djordon: Yeah, probably, since it is the same check.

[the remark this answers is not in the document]
djordon: Updated to do the same "mempool" check in the coordinator in https://github.com/stacks-network/sbtc/pull/1406/commits/1f42040c6fd6896d502d23ec7fe15136800aa4f4.

signer/src/storage/mod.rs
```diff
@@ -319,6 +319,23 @@ pub trait DbRead {
...
     ) -> impl Future<Output = Result<bool, Error>> + Send;
 
+    /// Returns whether the identified withdrawal may be included in a
+    /// sweep transaction that is in the bitcoin mempool.
+    ///
+    /// # Notes
+    ///
+    /// At this time, we cannot use the bitcoin mempool for this
+    /// information since we cannot match withdrawals with transaction
+    /// outputs unless a lot of computational effort. Instead, we use our
```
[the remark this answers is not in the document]
djordon: Corrected in https://github.com/stacks-network/sbtc/pull/1406/commits/dcb3f435bd05724c7770044bec6210c81779226d.

signer/tests/integration/setup.rs
```diff
@@ -440,6 +415,27 @@ pub async fn backfill_bitcoin_blocks(db: &PgStore, rpc: &Client, chain_tip: &bit
...
+        let parent_header_hash = block_header.previous_block_hash.unwrap();
+        let bitcoin_block = BitcoinBlock {
+            block_hash: block_header.hash.into(),
+            block_height: block_header.height as u64,
+            parent_hash: parent_header_hash.into(),
+        };
+
+        db.write_bitcoin_block(&bitcoin_block).await.unwrap();
+        block_header = rpc.get_block_header_info(&parent_header_hash).unwrap();
+    }
```
matteojug, line 436: nit: is this snippet doing something different or we can just call `backfill_bitcoin_blocks`?
```suggestion
    backfill_bitcoin_blocks(db, rpc, &chain_tip_info.best_block_hash).await
```

signer/tests/integration/postgres.rs
```diff
@@ -5399,3 +5401,182 @@ async fn pending_rejected_withdrawal_already_accepted() {
...
+    fetch_canonical_bitcoin_blockchain(&db, rpc).await;
+
+    let chain_tip = db.get_bitcoin_canonical_chain_tip().await.unwrap().unwrap();
+
+    // This is needed for the part of the query that fetches the signers'
+    // UTXO.
+    setup.store_dkg_shares(&db).await;
+
+    // This donation is currently the signers' UTXO, which is needed in the
+    // `is_withdrawal_live` implementation.
```
matteojug, line 5431: nit: also other occurrences
```suggestion
    // `is_withdrawal_inflight ` implementation.
```

signer/tests/integration/setup.rs
```diff
@@ -440,6 +415,27 @@ pub async fn backfill_bitcoin_blocks(db: &PgStore, rpc: &Client, chain_tip: &bit
     }
 }
 
+/// Fetch all block headers from bitcoin-core and store it in the database.
+pub async fn fetch_canonical_bitcoin_blockchain(db: &PgStore, rpc: &Client) {
```
matteojug, line 419: maybe nit: given the usual usage we do of this we could also return `db.get_bitcoin_canonical_chain_tip().await.unwrap().unwrap()` for ergonomics

matteojug (approved):
LGTM (modulo the failing test)

signer/tests/integration/setup.rs
djordon: Oh right, of course. Will update.

djordon: Done in https://github.com/stacks-network/sbtc/pull/1406/commits/c7d18638e9733d402d3c47adfeb3c68351f49001

signer/tests/integration/postgres.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/1406/commits/4ace37b48ada4b4dd56f0b634a3e347e89546b9f

signer/tests/integration/setup.rs
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1406/commits/c7d18638e9733d402d3c47adfeb3c68351f49001.

matteojug: `chain_tip_info.best_block_hash.into()` may not be the db chain tip, returning it may be confusing; or maybe not, feel free to resolve in that case

matteojug (approved):

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3559,6 +3588,20 @@ async fn process_rejected_withdrawal(is_completed: bool, is_in_mempool: bool, is
...
+            validation_result: signer::bitcoin::validation::InputValidationResult::Ok,
+            aggregate_key: aggregate_key.into(),
+            is_valid_tx: false,
+            will_sign: false,
+            chain_tip: bitcoin_chain_tip.block_hash,
+            sighash: bitcoin::TapSighash::from_byte_array([1; 32]).into(),
+        };
+        db.write_bitcoin_txs_sighashes(&[sighash]).await.unwrap();
+
         if is_submitted {
```
matteojug, line 3605: I guess this is no longer useful right now, but we can remove the `is_submitted` case later. Or just keep it.

signer/tests/integration/setup.rs
djordon: Maybe I should add an assert

signer/tests/integration/transaction_coordinator.rs
djordon: Removed in https://github.com/stacks-network/sbtc/pull/1406/commits/ceea0fba418c81220f59c4086723c9d850cfb91d.

signer/tests/integration/setup.rs
djordon: Couldn't hurt, so I added it in https://github.com/stacks-network/sbtc/pull/1406/commits/efc0622d2e19c8dd33d1c25e5bb53fe3c7d5cf3a.

Issue #1067, opened 2024-12-09
[Chore]: Migrate Deposit and Withdrawal Updates to the Emily Sidecar


Jiloc:
# Chore - Migrate Deposit and Withdrawal Updates to the Emily Sidecar

## 1. Description

Now that we have a dedicated sidecar for Emily, we can simplify the architecture by moving all deposit and withdrawal update responsibilities from the 15 signers to the sidecar. The signers will only be responsible for sending accepted deposit updates to Emily.

## 2. Why It's Necessary
- The sidecar leverages the retry mechanism provided by the stacks-node. If an event update to Emily fails, the sidecar will handle retries, ensuring all events will be forwarded to Emily. Currently, the signers rely on redundant messaging to ensure Emily receives updates, which does not guarantee all messages are successfully delivered.
- This change reduces dependencies between the signers and Emily.

## 3. Execution Checklist

- [ ] Remove update calls from `new_block_handler` in the signer.
- [ ] Implement event parsing and communication with Emily in the sidecar.

PR #1408, opened 2025-02-21
[feat] Emily client python

Jiloc:
## Description

Part of #1067  - build the OpenAPI Emily client used by the sidecar. 

Previously, the sidecar was doing a raw POST request because it was using only the chainstate endpoint. Now we need the client for the other requests.

## Changes
- modified `emily/openapi-gen/package.json` to add the command for generating the python client
- modified `Makefile` 
- everything else is just autogenerated code


Jiloc:
We decided to go with a different approach https://github.com/stacks-network/sbtc/pull/1416

PR #1409, opened 2025-02-21
[chore] sidecar send new block updates

Jiloc:
## Description

Part of https://github.com/stacks-network/sbtc/issues/1067 

## Changes
- Update from Flask to FastAPI to reduce boilerplate for parsing jsons and error handling (discussed in https://github.com/stacks-network/sbtc/pull/1050)
- Sends new-block events instead of chainstate
  
## Testing Information
- updated unit tests to work with new block events
- added integration tests - currently not integrated into Makefile `make integration-test`
 
## Checklist:


signer/tests/fixtures/withdrawal-reject-event.json
```diff
@@ -30,7 +30,7 @@
                     "Tuple": {
                         "data_map": {
                             "request-id": {
-                                "UInt": 2
+                                "UInt": 1
```
Jiloc, line 33: This was needed to make the `test_withdrawal_reject` integration tests. Where a withdrawal can be rejected only if a connected withdrawal already exists

djordon: Changing the test is fine, but what do we do if Emily does not have a create-withdrawal event in the database and it receives either an accept or a reject? Shouldn't it just store the event?

docker/sbtc/emily-sidecar/Dockerfile
```diff
@@ -1,13 +1,11 @@
...
 # Install dependencies
 RUN pip install --no-cache-dir -r requirements.txt
-# Set environment variables
-ENV FLASK_APP=app.py
-# Expose the Flask port
+# Expose the FastAPI port
 EXPOSE 20540
 ENV EMILY_API_KEY=testApiKey
-ENV EMILY_CHAINSTATE_URL=http://emily-server:3031/chainstate
-ENV DEPLOYER_ADDRESS=SN3R84XZYA63QS28932XQF3G1J8R9PC3W76P9CSQS
```
djordon, line 12: Ahh we don't need the deployer in the side-car since filtering is getting done in Emily itself.

emily_sidecar/main.py
```diff
@@ -0,0 +1,63 @@
...
+logger = logging.getLogger(__name__)
+
+logger.info("Using Emily endpoint: %s", settings.EMILY_ENDPOINT)
+
+headers = {"Content-Type": "application/json", "x-api-key": settings.API_KEY}
+
+
+class NewBlockEventModel(BaseModel, extra="allow"):
+    block_height: int
+    index_block_hash: str
```
djordon, line 27: Oh nice, that `extra` field is critical because we expect more fields to be sent in event from the stacks node but pydantic ignores them by default. We should probably add a comment above here describing why it's necessary.

emily_sidecar/main.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@app.post("/attachments/new")
+async def handle_attachments() -> dict:
+    return {}
+
+
+@app.get("/")
+def read_root() -> dict:
+    return {"message": "Hello, World!"}
```
djordon, line 63: We should probably just return `200 Ok` here.

djordon (commented):
Nice, this largely looks good.

signer/tests/fixtures/withdrawal-reject-event.json
Jiloc: At the moment, Emily returns an error when the update_withdrawal is called, and the sidecar retries indefinitely. Emily cannot store the event because withdrawals are stored as objects within a list in the corresponding Withdrawal entry in DynamoDB. If the Withdrawal entry doesn’t exist, we have no place to store its related update.

By design, this situation shouldn't happen—an update should never arrive before the withdrawal itself unless the entry was forcefully removed from the DB. Ideally, we could create the withdrawal entry when receiving an update, but the event doesn't have enough information.

I was considering a follow-up PR to introduce a new query for withdrawal lookups by [req_id, txid] (which isn't currently there). I would modify both [`handle_withdrawal_accept`](https://github.com/stacks-network/sbtc/blob/2c8b215d35d722479dcc1f72eebde304a1bc8858/emily/handler/src/api/handlers/new_block.rs#L405)  and [`handle_withdrawal_reject`](https://github.com/stacks-network/sbtc/blob/2c8b215d35d722479dcc1f72eebde304a1bc8858/emily/handler/src/api/handlers/new_block.rs#L331) to check if the corresponding withdrawal exists before proceeding. If it doesn’t, they would return an error to the `new_block` handler, which would log the issue and skip processing the request, preventing unnecessary retries and potential stalls. (Same behavior as in [`handle_completed_deposits`](https://github.com/stacks-network/sbtc/blob/2c8b215d35d722479dcc1f72eebde304a1bc8858/emily/handler/src/api/handlers/new_block.rs#L284-L293) )

docker/sbtc/emily-sidecar/Dockerfile
Jiloc: exactly, now the sidecar does the bare minimum

emily_sidecar/main.py
Jiloc: changed in https://github.com/stacks-network/sbtc/pull/1409/commits/658aa451475a5b17ee66b6b6fd874368ce8e99b5

emily_sidecar/main.py
Jiloc: addressed in https://github.com/stacks-network/sbtc/pull/1409/commits/658aa451475a5b17ee66b6b6fd874368ce8e99b5

signer/tests/fixtures/withdrawal-reject-event.json
djordon: Hmmm, I think that we can prevent it from happening. The three situations that came to mind are:
1. Some bug causes the code to bail before writing to the database.
2. Some error in AWS prevents writing to the database.
3. The side-car being down and missing the event `initiate-withdrawal-request` event.
 
They all seem unlikely for one reason or another, but I was wondering if there is some way that we can handle this more gracefully. One potential alternative to the current design is to write all events in some sort of history table. We would then potentially recreate the current state by looking at the relevant events at request time. I'm not sure how to handle the fact that the update event doesn't have enough data, although it seems better than missing even more data by dropping the event.

This kind of design would work for both deposits and withdrawals, and gets us out of having to delete or miss data, which is would save us from some headaches. The migration from our current design to the new one might be annoying but we could have both systems in place while we iron out some potential issues. What do you think?

emily_sidecar/main.py
```diff
@@ -0,0 +1,66 @@
...
+            e,
+        )
+        # lets return an error so that the node will retry
+        raise HTTPException(status_code=500, detail="Failed to send new_block event")
+    logger.info(
+        "Successfully processed new block event: block_height=%s, index_block_hash=%s",
+        new_block_event.block_height,
+        new_block_event.index_block_hash,
+    )
+    return {}
```
matteojug, line 56: nit: why returning a dict? vs an empty response

emily_sidecar/requirements.txt
matteojug, line 1: not necessarily for this PR: can we switch to a dependency management that supports locking? eg poetry or uv (please not pipenv).

emily_sidecar/settings.py
matteojug, line 1: nit/not for this PR: pydantic has a nice thingy to handle settings (https://docs.pydantic.dev/latest/concepts/pydantic_settings/), in case we want something more robust.

emily_sidecar/test_integration.py
```diff
@@ -0,0 +1,221 @@
...
+import logging_config
+import settings
+from main import app, headers
+
+logging_config.silence_logging()
+
+# Set logging level to CRITICAL to suppress other log messages during tests
+logging.basicConfig(level=logging.CRITICAL)
+for logger_name in logging.root.manager.loggerDict:
+    logging.getLogger(logger_name).setLevel(logging.CRITICAL)
```
matteojug, line 18: Do we need this even with the `logging_config.silence_logging()` above?

emily_sidecar/main.py
```diff
@@ -0,0 +1,66 @@
...
+
+# Set up logging when the app starts
+logging_config.setup_logging()
+
+# Get a logger instance
+logger = logging.getLogger(__name__)
+
+logger.info("Using Emily endpoint: %s", settings.EMILY_ENDPOINT)
+
+headers = {"Content-Type": "application/json", "x-api-key": settings.API_KEY}
```
matteojug, line 21: nit: if we use `json` in request we shouldn't need to set `Content-Type`

emily_sidecar/main.py
Jiloc: from what I remembered, the stacks node is expecting a json response, and FastAPI automatically converts the dictionary to a JSON response with a status code of 200. But I am open to a `return JSONResponse(content={}, status_code=200)` instead, it's definitely more explicit

emily_sidecar/main.py
djordon: Oh I think what I was thinking was to return nothing, something like this:
```python
from fastapi import Response, status
...
@app.get("/")
def read_root() -> Response:
    return Response(status_code=status.HTTP_200_OK)
```

emily_sidecar/main.py
djordon: I would think that it's still good to keep it in though, since it is explicit.

emily_sidecar/test_integration.py
Jiloc: nope, was there before the silence_logging

emily_sidecar/requirements.txt
Jiloc: I forgot about this! Btw, I am very curious to know about the hate behind pipenv :D

emily_sidecar/settings.py
Jiloc: nice!

emily_sidecar/requirements.txt
djordon: I haven't used `pipenv`, but people seem to dislike it and I don't really know why. Why not pipenv?

signer/tests/fixtures/withdrawal-reject-event.json
Jiloc: I see the reasoning behind saving updates in a separate table, but I’m unsure if it would actually improve the situation. When thinking through the three cases you mentioned, I believe they are already handled by the design of how the stacks-node, the sidecar and Emily interact:

- If the sidecar is down, the stacks-node will stall and keep retrying.
- If, for any reasons, Emily is down or fails to execute a deposit or withdrawal creation, it returns an error to the sidecar, which propagates it back to the stacks-node—causing it to stall and keep  retrying.

I mean, there are still case I can think of where data truly goes missing. Like human error, if someone manually removes a withdrawal entry from the database. But even then, we would still have the same problems:

- For deposits, deletion is irreversible because we'd lose things like the reclaim and deposit scripts that can't be reconstructed.
- For withdrawals, even if we tried to rebuild the original events from the updates, we’d still miss some required data. In such a case, we’d need to either replay stacks-node events or query the smart contract / Hiro api, to recover the missing information.

The main advantage I see with the new approach is that it wouldn’t stall the stacks-node. But even taht wouldn’t be a "critical" issue because Emily would still accept new deposit creations. It would only temporarily display outdated data downstream to users, but the signers will work normally. And this can be "easily" solved by detecting and logging cases where a withdrawal update arrives without a corresponding entry. (In any case I think we should connect those logs to an AWS alert, so that they don't go unnoticed)

I mean, I still feel this would be a more elegant design overall, but I don’t see enough advantages over the current implementation to justify the effort. That said, I might be missing some aspects you’ve already thought through…?

emily_sidecar/test_integration.py
Jiloc: fixed in https://github.com/stacks-network/sbtc/pull/1409/commits/20ce7f422a76a93257b5aa128b4fb7adf1d738ed

emily_sidecar/main.py
Jiloc: yep misunderstood, changed https://github.com/stacks-network/sbtc/pull/1409/commits/4a331686dd2ba655ab6c2b9e2ae47481190cce6c

emily_sidecar/main.py
Jiloc: @djordon , sorry, I read the comment too late after I alredy changed it in D: https://github.com/stacks-network/sbtc/pull/1409/commits/6b2023e89530b79a7907eb79f86c55f4a1185518

matteojug: Resolving the convo, but I'd not oppose to have the content type explicitly in the headers; although I'm not sure if it's the type of explicit thing that it's informative rather than redundant, but again nothing against it.

djordon: Okay, no worries, I'm fine with whichever.

signer/tests/fixtures/withdrawal-reject-event.json
djordon: > I mean, I still feel this would be a more elegant design overall, but I don’t see enough advantages over the current implementation to justify the effort. That said, I might be missing some aspects you’ve already thought through…?

I don't think that I've thought through anything that you haven't thought through. And I'm not entirely convinced that the overall effort is worth it quite yet either. But I'm not entirely satisfied with how Emily handles missing data (or reorgs, or whatever). So if we need to keep it around, we may need to do something about it. I'm not wedded to my proposed idea at all, but I feel like we'll eventually need to do something about the current state.

Anyhow, so long as we don't return the error to the node (like you said) I think we're probably going to be okay for launching withdrawals.

Jiloc: Yeah, I definitely agree that Emily needs a restructuring, but I think we're discussing two separate issues here:

- Handlers missing data – This is what we’ve been focusing on, like receiving an update for a deposit/withdrawal that doesn’t exist in the DB. This shouldn't ever happen, but if it does is most likely caused by human error. So I am fine having a more "manual" process handling this problem. Right now, Emily returns an error, which causes the sidecar and stacks-node to stall and retry forever. If we add a check for missing entries, we can log it and recover later—without stalling the node.

- Emily deleting data during reorgs – This only affects updates related to an entry, not the entry itself (e.g., deposits always stay, but updates like accepted → confirmed get deleted). We know this will happen frequently by design, but I also don’t love that we’re outright removing data.

We could approach this in two ways:

- Make Emily act more like a node – Store all updates for all forks and modify the API so it expects a chain tip, and response will be relative to the given chain. (e.g. give the the pending deposits associated with this chaintip).
- Keep it as-is but flag reorged updates instead of deleting them – That way, we still have a reference in the DB instead of fully removing them. The data shouldn't be returned downstream, but can be used for debugging.

djordon: Yeah great points. You're right that these are different issues.

I think that we're on the same page. Things aren't great, doing either soft deletes or storing all updates are good ideas, and neither of which are relevant to this PR 😄.

emily_sidecar/settings.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+# Environment variables and configuration
+API_KEY = os.getenv("EMILY_API_KEY", "testApiKey")
+EMILY_ENDPOINT = os.getenv(
+    "EMILY_ENDPOINT", "http://@host.docker.internal:3031"
```
matteojug, line 6: nit
```suggestion
    "EMILY_ENDPOINT", "http://host.docker.internal:3031"
```

matteojug (approved):

Issue #1407, opened 2025-02-21
[Feature]: Wait before rejecting a withdrawal request


djordon:
# Feature - Wait before rejecting a withdrawal request

## 1. Description

This ticket is to wait even more before rejecting a withdrawal request. In this case, we need to wait for 6 blocks after a sweep has been confirmed where the relevant sweep is the last one since we have considered fulfilling the withdrawal.

This is stop gap until we have updated the smart contracts.

### 2.1 Acceptance Criteria:
   - [ ] We do not submit reject withdrawal contract calls if we have not seen at least 6 blocks since confirmation of a sweep after attempting to fulfill a withdrawal request.

PR #1411, opened 2025-02-21
feat: feature wait before rejecting a withdrawal request

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1407

## Changes

* Add in a validation check that looks for "active" withdrawals. These are withdrawals where it's possible that a fork could make the transaction valid in the eyes of the bitcoin network.
* Made sure that the coordinator did the same "is active" check when selecting requests to reject.

## Testing Information

So far this PR has added one test. The comments in it are likely wrong and stuff, and it needs a little polish. I will fix that all up and add more tests.

## Checklist:

### DIFF
--- a/signer/src/stacks/contracts.rs
+++ b/signer/src/stacks/contracts.rs
@@ -1115,8 +1115,8 @@ impl AsContractCall for RejectWithdrawalV1 {
     /// 7. Whether the withdrawal request is being serviced by a sweep
     ///    transaction that is in the mempool.
     /// 8. Whether we need to worry about forks causing the withdrawal to
-    ///    be confirmed by a sweep that was broadcst changing the status of
-    ///    the request from rejected to accepted.
+    ///    be confirmed by a sweep that was broadcast changing the status
+    ///    of the request from rejected to accepted.
     async fn validate<C>(&self, ctx: &C, req_ctx: &ReqContext) -> Result<(), Error>
     where
         C: Context + Send + Sync,
--- a/signer/src/stacks/contracts.rs
+++ b/signer/src/stacks/contracts.rs
@@ -59,6 +59,7 @@ use crate::storage::model::ToLittleEndianOrder as _;
 use crate::storage::DbRead;
 use crate::DEPOSIT_DUST_LIMIT;
 use crate::WITHDRAWAL_BLOCKS_EXPIRY;
+use crate::WITHDRAWAL_MIN_CONFIRMATIONS;
 
 use super::api::StacksInteract;
 
@@ -1029,6 +1030,10 @@ pub enum WithdrawalRejectErrorMsg {
     /// been either accepted or rejected already.
     #[error("the smart contract has been updated to indicate that the request has been completed")]
     RequestCompleted,
+    /// A withdrawal request is still active if it's reasonably possible
+    /// for the request to be fulfilled by a sweep transaction on bitcoin.
+    #[error("the withdrawal request is still active, we cannot reject it yet")]
+    RequestStillActive,
     /// Withdrawal request fulfilled
     #[error("Withdrawal request fulfilled")]
     RequestFulfilled,
@@ -1109,6 +1114,9 @@ impl AsContractCall for RejectWithdrawalV1 {
     /// 6. Whether the withdrawal request has expired. Fail if it hasn't.
     /// 7. Whether the withdrawal request is being serviced by a sweep
     ///    transaction that is in the mempool.
+    /// 8. Whether we need to worry about forks causing the withdrawal to
+    ///    be confirmed by a sweep that was broadcst changing the status of
+    ///    the request from rejected to accepted.
     async fn validate<C>(&self, ctx: &C, req_ctx: &ReqContext) -> Result<(), Error>
     where
         C: Context + Send + Sync,
@@ -1189,6 +1197,16 @@ impl AsContractCall for RejectWithdrawalV1 {
             .await?;
         if withdrawal_is_inflight {
             return Err(WithdrawalRejectErrorMsg::RequestBeingFulfilled.into_error(req_ctx, self));
+        }
+
+        // 8. Check whether the withdrawal request is still active, as in
+        //    it could still be fulfilled.
+        let withdrawal_is_active = db
+            .is_withdrawal_active(&self.id, &req_ctx.chain_tip, WITHDRAWAL_MIN_CONFIRMATIONS)
+            .await?;
+
+        if withdrawal_is_active {
+            return Err(WithdrawalRejectErrorMsg::RequestStillActive.into_error(req_ctx, self));
         }
 
         Ok(())
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -448,7 +448,7 @@ impl PgStore {
     /// This function returns the bitcoin block height of the first
     /// confirmed sweep that happened on or after the given minimum block
     /// height.
-    pub async fn get_least_txo_height(
+    async fn get_least_txo_height(
         &self,
         chain_tip: &model::BitcoinBlockHash,
         min_block_height: i64,
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -2158,9 +2158,17 @@ impl super::DbRead for PgStore {
         bitcoin_chain_tip: &model::BitcoinBlockRef,
         min_confirmations: u64,
     ) -> Result<bool, Error> {
-        let min_block_height = sqlx::query_scalar::<_, Option<i64>>(
+        // We want to find the first sweep transaction that happened
+        // *after* the last time the signers considered sweeping the
+        // withdrawal request. We do this in two stages:
+        // 1. Find the block height of the last time that the signers
+        //    considered the withdrawal request (the query right below this
+        //    comment).
+        // 2. Find the least height of any sweep transaction confirmed in a
+        //    block with height greater than the height returned from (1).
+        let last_considered_height = sqlx::query_scalar::<_, Option<i64>>(
             r#"
-            SELECT MAX(bb.block_height) + 1
+            SELECT MAX(bb.block_height)
             FROM sbtc_signer.bitcoin_withdrawals_outputs AS bwo
             JOIN sbtc_signer.bitcoin_blocks AS bb
               ON bb.block_hash = bwo.bitcoin_chain_tip
@@ -2174,12 +2182,17 @@ impl super::DbRead for PgStore {
         .await
         .map_err(Error::SqlxQuery)?;
 
-        // This means that there are no rows associated with the ID in the
-        // `bitcoin_withdrawals_outputs` table.
-        let Some(min_block_height) = min_block_height else {
+        // We add one because we are interested in sweeps that were
+        // confirmed after the signers last considered the withdrawal.
+        let Some(min_block_height) = last_considered_height.map(|x| x + 1) else {
+            // This means that there are no rows associated with the ID in the
+            // `bitcoin_withdrawals_outputs` table.
             return Ok(false);
         };
 
+        // We now know that the signers considered this withdrawal request
+        // at least once. We now try to find the height of the oldest
+        // signer TXO whose height is greater than the height from above.
         let chain_tip_hash = &bitcoin_chain_tip.block_hash;
         let least_txo_height = self
             .get_least_txo_height(chain_tip_hash, min_block_height)
@@ -2192,8 +2205,14 @@ impl super::DbRead for PgStore {
         let Some(least_txo_height) = least_txo_height else {
             return Ok(true);
         };
-
-        Ok(least_txo_height + min_confirmations >= bitcoin_chain_tip.block_height)
+        // We test whether the TXO height has a minimum number of
+        // confirmations. If it doesn't have enough confirmations, then it
+        // is considered active since a fork could put it into the mempool
+        // again.
+        let txo_confirmations = bitcoin_chain_tip
+            .block_height
+            .saturating_sub(least_txo_height);
+        Ok(txo_confirmations <= min_confirmations)
     }
 
     async fn get_bitcoin_tx(
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -2150,8 +2150,10 @@ impl super::DbRead for PgStore {
         let output_type = model::TxOutputType::SignersOutput;
         let chain_tip_hash = &bitcoin_chain_tip.block_hash;
         let signer_utxo_fut = self.get_utxo(chain_tip_hash, output_type, min_block_height);
+        // If this returns None, then the sweep itself could be in the
+        // mempool. If that's the case then this is definitely active.
         let Some(signer_utxo) = signer_utxo_fut.await? else {
-            return Ok(false);
+            return Ok(true);
         };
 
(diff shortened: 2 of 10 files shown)


signer/src/storage/postgres.rs
```diff
@@ -2121,6 +2125,46 @@ impl super::DbRead for PgStore {
...
+        let output_type = model::TxOutputType::SignersOutput;
+        let chain_tip_hash = &bitcoin_chain_tip.block_hash;
+        let signer_utxo_fut = self.get_utxo(chain_tip_hash, output_type, min_block_height);
+        // If this returns None, then the sweep itself could be in the
+        // mempool. If that's the case then this is definitely active.
+        let Some(signer_utxo) = signer_utxo_fut.await? else {
+            return Ok(true);
+        };
+
+        Ok(signer_utxo.block_height + min_confirmations >= bitcoin_chain_tip.block_height)
```
matteojug, line 2165: mmm `get_utxo` always return the latest utxo, so if we have a sweep every couple of blocks every withdrawal will always be active, as even for one way in the past we would still get a recent utxo (and `bitcoin_chain_tip.block_height` will always be less than `min_confirmations` ahead in the example above).

djordon: Bah! What was I thinking, of course it returns an unspent transaction output 🤦🏽‍♂️. But an easy fix.

signer/src/stacks/contracts.rs
```diff
@@ -1109,6 +1114,9 @@ impl AsContractCall for RejectWithdrawalV1 {
     /// 6. Whether the withdrawal request has expired. Fail if it hasn't.
     /// 7. Whether the withdrawal request is being serviced by a sweep
     ///    transaction that is in the mempool.
+    /// 8. Whether we need to worry about forks causing the withdrawal to
+    ///    be confirmed by a sweep that was broadcst changing the status of
```
cylewitruk, line 1118:
```suggestion
    ///    be confirmed by a sweep that was broadcast changing the status of
```

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -603,3 +604,135 @@ async fn reject_withdrawal_validation_request_being_fulfilled() {
...
+    // transaction is spending the current signer UTXO, then it could
+    // possibly be in the mempool. Since the signers' UTXO is a donation,
+    // we're saying that the coordinator may have tried to fulfill the
+    // withdrawal.
+    setup.store_bitcoin_withdrawals_outputs(&db).await;
+
+    // Generate the transaction and corresponding request context.
+    let (reject_withdrawal_tx, req_ctx) = make_withdrawal_reject(&setup, &db).await;
+
+    let validation_result = reject_withdrawal_tx.validate(&ctx, &req_ctx).await;
```
cylewitruk, line 701: Suggestion: maybe a comment on each of these validation attempts about where in relation to the `WITHDRAWAL_MIN_CONFIRMATIONS` things are expected to be so it's easier to follow along at a glance?

signer/tests/integration/setup.rs
```diff
@@ -847,13 +860,31 @@ impl TestSweepSetup2 {
...
+    /// network.
+    pub fn broadcast_sweep_tx(&mut self, rpc: &Client) {
         // Okay now we try to peg-in the deposit by making a transaction.
         // Let's start by getting the signer's sole UTXO.
         let aggregated_signer = &self.signers.signer;
         let signer_utxo = aggregated_signer.get_utxos(rpc, None).pop().unwrap();
 
+        // Well we want a BitcoinCoreClient, so we create one using the
+        // settings. Not, the best thing to do, sorry. TODO: pass in a
+        // bitcoin core client object.
```
cylewitruk, line 873: This will all die "soon" with some integration test improvements anyway, right? ;)

signer/src/storage/postgres.rs
```diff
@@ -2121,6 +2152,50 @@ impl super::DbRead for PgStore {
...
+        .await
+        .map_err(Error::SqlxQuery)?;
+
+        // This means that there are no rows associated with the ID in the
+        // `bitcoin_withdrawals_outputs` table.
+        let Some(min_block_height) = min_block_height else {
+            return Ok(false);
+        };
+
+        let chain_tip_hash = &bitcoin_chain_tip.block_hash;
```
cylewitruk, line 2183: Comment suggestion:

```suggestion
        // We now know that there's at least one `bitcoin_withdrawals_outputs` for this request,
        // and we have the bitcoin block height of the _newest_ one. We proceed to attempt to find
        // the height of the _oldest_ signer UTXO whose height is >= the withdrawal output height
        // from above.
        let chain_tip_hash = &bitcoin_chain_tip.block_hash;
```

signer/src/storage/postgres.rs
```diff
@@ -2121,6 +2152,50 @@ impl super::DbRead for PgStore {
...
+            .map(u64::try_from)
+            .transpose()
+            .map_err(|_| Error::TypeConversion)?;
+        // If this returns None, then the sweep itself could be in the
+        // mempool. If that's the case then this is definitely active.
+        let Some(least_txo_height) = least_txo_height else {
+            return Ok(true);
+        };
+
+        Ok(least_txo_height + min_confirmations >= bitcoin_chain_tip.block_height)
```
cylewitruk, line 2196: I found this a little easier to reason with 🤷 

```suggestion
        // We test whether the oldest UTXO height has fallen outside of the minimum
        // confirmation window.
        let max_allowed_height = bitcoin_chain_tip.block_height - min_confirmations;
        OK(least_txo_height <= max_allowed_height)
```

signer/src/storage/postgres.rs
```diff
@@ -2121,6 +2152,50 @@ impl super::DbRead for PgStore {
         .map_err(Error::SqlxQuery)
     }
 
+    async fn is_withdrawal_active(
+        &self,
+        id: &model::QualifiedRequestId,
+        bitcoin_chain_tip: &model::BitcoinBlockRef,
+        min_confirmations: u64,
+    ) -> Result<bool, Error> {
+        let min_block_height = sqlx::query_scalar::<_, Option<i64>>(
```
cylewitruk, line 2161: Comment suggestion:

```suggestion
        // We look for all `bitcoin_withdrawal_outputs` for the given request id for which
        // we have a matching bitcoin block and return the highest block height we find (if any).
        let min_block_height = sqlx::query_scalar::<_, Option<i64>>(
```

signer/src/stacks/contracts.rs
```diff
@@ -1029,6 +1030,10 @@ pub enum WithdrawalRejectErrorMsg {
     /// been either accepted or rejected already.
     #[error("the smart contract has been updated to indicate that the request has been completed")]
     RequestCompleted,
+    /// A withdrawal request is still active if it's reasonably possible
```
cylewitruk, line 1033: Suggestion: maybe tweak this description or the `RequestBeingFulfilled` one to just make it more clear the differences; "active" and "being fulfilled" felt a bit like the same thing to me when I first glanced this over.

djordon: Yeah, I thought the same thing and I couldn't think of a good name. Hopefully we can kill this soon by updating the smart contract.

signer/tests/integration/setup.rs
```diff
@@ -898,7 +929,22 @@ impl TestSweepSetup2 {
...
+            block_hash: block_header.best_block_hash,
+            txid,
+        });
+    }
+
+    /// This function generates a sweep transaction that sweeps in the
+    /// deposited funds and sweeps out the withdrawal funds in a proper
+    /// sweep transaction, that is also confirmed on bitcoin.
+    pub fn submit_sweep_tx(&mut self, rpc: &Client, faucet: &Faucet) {
+        self.broadcast_sweep_tx(rpc);
```
matteojug, line 944: should we skip `broadcast_sweep_tx` if `self.broadcast_info.is_some()`?

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -603,3 +604,135 @@ async fn reject_withdrawal_validation_request_being_fulfilled() {
...
+    // Normal: the signer follows the bitcoin blockchain and event observer
+    // should be getting new block events from bitcoin-core. We haven't
+    // hooked up our block observer, so we need to manually update the
+    // database with new bitcoin block headers.
+    fetch_canonical_bitcoin_blockchain(&db, rpc).await;
+
+    // Different: we need to store a row in the dkg_shares table so that we
+    // have a record of the scriptPubKey that the signers control. We need
+    // this so that the donation gets picked up correctly below.
+    setup.store_dkg_shares(&db).await;
```
matteojug, line 659: Why is this _Different_?

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -603,3 +604,135 @@ async fn reject_withdrawal_validation_request_being_fulfilled() {
...
+    // hooked up our block observer, so we need to manually update the
+    // database with new bitcoin block headers.
+    fetch_canonical_bitcoin_blockchain(&db, rpc).await;
+
+    // Different: We submit a sweep transaction into the mempool so that
+    // the TestSweepSetup2 struct has the sweep_tx_info set. We also need
+    // to submit the transaction in order for
+    // `TestSweepSetup2::store_bitcoin_withdrawals_outputs` to work as
+    // expected.
+    setup.broadcast_sweep_tx(rpc);
```
matteojug, line 688: nit: it seems it's no longer the case
```suggestion
    // Different: We submit a sweep transaction into the mempool so that
    // the TestSweepSetup2 struct has the sweep_tx_info set.
    setup.broadcast_sweep_tx(rpc);
```

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -603,3 +604,135 @@ async fn reject_withdrawal_validation_request_being_fulfilled() {
...
+
+    let validation_result = reject_withdrawal_tx.validate(&ctx, &req_ctx).await;
+    match validation_result.unwrap_err() {
+        Error::WithdrawalRejectValidation(ref err) => {
+            assert_eq!(err.error, WithdrawalRejectErrorMsg::RequestStillActive)
+        }
+        err => panic!("unexpected error during validation {err}"),
+    }
+
+    let withdrawals = setup.withdrawals.drain(..).collect::<Vec<_>>();
```
matteojug, line 709: Why do we need to drain here?

signer/tests/integration/withdrawal_reject.rs
```diff
@@ -603,3 +604,135 @@ async fn reject_withdrawal_validation_request_being_fulfilled() {
...
+    setup.submit_sweep_tx(rpc, faucet);
+
+    faucet.generate_blocks(WITHDRAWAL_MIN_CONFIRMATIONS);
+
+    setup.store_sweep_tx(&db).await;
```
matteojug, line 714: nit: can we move this `store_sweep_tx` right after `submit_sweep_tx`? Otherwise it may seems we need to have the blocks in the middle

signer/tests/integration/setup.rs
djordon: Oh, yeah we should.

signer/tests/integration/withdrawal_reject.rs
djordon: It isn't. I copied all of the comments from another test, and didn't update anything yet. Also, it isn't exactly "Different" in that other test either.

signer/tests/integration/withdrawal_reject.rs
djordon: Yeah not obvious. If we didn't drain, `TestSweepSetup2::submit_sweep_tx` would create a sweep transaction which includes the withdrawal. We don't want that for this test because then `RejectWithdrawalV1` would fail validation for the wrong reason when the sweep gets confirmed (the withdrawal would be fulfilled). I'll add a comment here saying basically the same thing.

signer/tests/integration/setup.rs
djordon: I'll make the suggested change here, but later I think we should break this out as `broadcast_sweep_tx` and `confirm_sweep_tx`, where `confirm_sweep_tx` does not broadcast anything.

djordon: Done in https://github.com/stacks-network/sbtc/pull/1411/commits/73f04b6ccaff811d15ca6ca82b7499f457b53797.

signer/src/storage/postgres.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/1411/commits/3b8bfbaaee84886e4a2433d635a55b6b8714f2ad

signer/src/storage/postgres.rs
djordon: Done in https://github.com/stacks-network/sbtc/pull/1411/commits/3b8bfbaaee84886e4a2433d635a55b6b8714f2ad

signer/src/stacks/contracts.rs
djordon: Fixed in https://github.com/stacks-network/sbtc/pull/1411/commits/d54581faabf3dfb0e5acdd15056a6e4654eb25cb

signer/tests/integration/withdrawal_reject.rs
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1411/commits/856e346f565eb336631484851d2785208fe2b9b0, or maybe one of the other commmits

signer/src/storage/postgres.rs
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1411/commits/3b8bfbaaee84886e4a2433d635a55b6b8714f2ad

signer/tests/integration/withdrawal_reject.rs
djordon: I added a comment in https://github.com/stacks-network/sbtc/pull/1411/commits/55a0aae4a01e50369815c6558035524368cc6b30

signer/tests/integration/withdrawal_reject.rs
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1411/commits/856e346f565eb336631484851d2785208fe2b9b0

signer/tests/integration/withdrawal_reject.rs
djordon: I think this was updated in https://github.com/stacks-network/sbtc/pull/1411/commits/55a0aae4a01e50369815c6558035524368cc6b30.

signer/tests/integration/withdrawal_reject.rs
djordon: I updated this in https://github.com/stacks-network/sbtc/pull/1411/commits/55a0aae4a01e50369815c6558035524368cc6b30

signer/tests/integration/withdrawal_reject.rs
cylewitruk: Much easier to follow :)

signer/src/storage/postgres.rs
cylewitruk: Much better :)

cylewitruk (approved):

matteojug (approved):

PR #1410, opened 2025-02-21
[feat] Sidecar update Emily for withdrawal and deposits

Jiloc:
## Description

Part of https://github.com/stacks-network/sbtc/issues/1067 - Add parsing of sBTC events and sends the withdrawal-create, withdrawal-accept, withdrawal-reject and deposit-complete updates to Emily. 
The filtering logic and creation of the updates is very similar to the implementation in new_block.rs

## Changes
- add parsing of Clarity values (from the smart contracts events)
    - the parser was created to be simple by design, and only cover the 4 sbtc contract events we care about. Errors in parsing unsupported types will cause the event to be skipped. 
- add handling of sbtc contracts - transforming them into requests expected by Emily 
- sends updates to Emily
     
## Testing Information
- tests are not currently integrated in the `make test` suite, but can be launched with (after installing the dependencies from requirements.txt):
    - `cd emily_sidecar`
    - `python -m test`

## Checklist:

- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

Issue #1168, opened 2024-12-19
[Design]: Deposit into sBTC directly from Bitcoin


djordon:
# Design - Deposit into sBTC directly from Bitcoin

This ticket holds the design of depositing into sBTC directly from the bitcoin layer, with the need of using Emily or a bridge.

# 1. Summary

Include the necessary peg-in data in an `OP_RETURN` output, and have the signers search for `OP_RETURN` outputs that have the right format.

# 2. Context & Purpose

The proposed design is to have deposit transactions that have enough data on chain to make deposit UTXOs discoverable without the need for the transaction to be revealed to an external API like Emily. This will make it easier for others to build their own bridges, or forgo using a bridge entirely if they desire.

Pegging in directly from Bitcoin L1 places the burden on wallets

***Relevant Research Discussions***

Other bridges support direct from the L1 peg-ins, link to them from https://github.com/stacks-network/sbtc/issues/153.

***External Resources***

- https://github.com/stacksgov/sips/pull/113, for the original discussion that went from a `OP_RETURN` style deposit flow to commit-reveal style deposit flow.

# 3. Design

## 3.1 Proposed Component Design

The design is to have an `OP_RETURN` that is discoverable, and contains enough data for the signers to spend the deposit UTXO, committing the full deposit data on chain. 

**Process Overview:**

1. **Deposit Construction:** Users create a deposit UTXO and an `OP_RETURN` output containing data correlating with their deposit UTXO specifics.
2. **Signer Validation:** Signers scan all transactions for `OP_RETURN` outputs matching the HASH160 of a signer’s aggregate key. Valid transactions are considered for sweep operations.
3. **Transaction Processing:** Validated deposits are added to Emily.

Note that the signers already scan all outputs in all transactions in every block that they process, so matching on an `OP_RETURN` output is not much additional effort.

The proposed layout for the `OP_RETURN` output is this:

```
0         1                    21     22        23         24                    44          65
|---------|--------------------|------|---------|----------|---------------------|-----------|
  version   signer pubkey hash   vout   max fee   locktime   reclaim pubkey hash   recipient
```

Meaning:

1. `version` (1 byte) is just a version byte for the layout.
2. `signer pubkey hash` (20 bytes) is the HASH160 of one of the signers’ x-only aggregate key.
3. `vout` (1 byte) is the output index of the deposit UTXO.
[...]

PR #1416, opened 2025-02-23
[feat] emily new block handler

Jiloc:
## Description

Part of https://github.com/stacks-network/sbtc/issues/1067. Alternative approach to #1410.

This PR introduces a new authenticated new_block endpoint in Emily, leveraging most of the existing signer code (excluding the types from the stacks-core crate).

It will be followed by two additional PRs:
- Removing new_block updates from the signers.
- Updating the sidecar to send raw new_block events to Emily.

The Integration tests are still in progress, but this should be already good to start the review.

## Changes
- New `DEPLOYER_ADDRESS` env variable required in CDK for event filtering. @fabergat this affects your deployment scripts.
- New private `new_block` endpoint in Emily, adapted from the signer's implementation. A future PR will be needed to refactor shared data structures for reuse between the signer and Emily.
- Fix for stack overflow and build-time recursion limit caused by deeply nested route compositions in Emily. ([CI error reference](https://github.com/stacks-network/sbtc/actions/runs/13474985974/job/37653039961#step:6:63)).

## Testing
- Added unit tests.
- Integration tests are still in progress.

## Checklist:


emily/handler/src/api/routes/chainstate.rs
```diff
@@ -12,8 +12,11 @@ pub fn routes(
 ) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
     get_chainstate_at_height(context.clone())
         .or(set_chainstate(context.clone()))
+        .boxed()
         .or(update_chainstate(context.clone()))
+        .boxed()
         .or(get_chain_tip(context))
+        .boxed()
```
djordon, line 19: Why all of the boxing?

Jiloc: Adding the new_block route caused the build to hit [recursion limits](https://github.com/stacks-network/sbtc/actions/runs/13474985974/job/37653039961#step:6:63). I initially tried increasing the recursion limit for Emily [here](https://github.com/stacks-network/sbtc/pull/1416/commits/66015a099d159c235aa2a57e420be7a305919d4b), but that only resolved the build issue. At runtime, it was still causing a stack overflow.

The boxed() calls prevent both of these issues. Without boxed, warp’s type system tries to resolve all nested or() routes into a single, deeply nested type, but it exceeds rust’s recursion depth. boxing erases these types

Jiloc: I left a comment at the root of the routes for documenting it. https://github.com/stacks-network/sbtc/blob/f079713e2f7ccba9c69ed377916d4a0407088321/emily/handler/src/api/routes/mod.rs#L45-L46

It should also be enough just to leave the boxed calls in the mod.rs, and not do it for each "sub-route" in the specific modules

djordon: 🙏🏽 okay, thank you.

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,580 @@
...
+            }
+        };
+        // Set the global limits.
+        let registry_address = SBTC_REGISTRY_IDENTIFIER.get_or_init(|| {
+            // Although the following line can panic, our unit tests hit this
+            // code path so if tests pass then this will work in production.
+            let contract_name = ContractName::from(SBTC_REGISTRY_CONTRACT_NAME);
+            let issuer = StandardPrincipalData::from(context.settings.deployer_address);
+            QualifiedContractIdentifier::new(issuer, contract_name)
+        });
```
djordon, line 100: I don't think that we need to remove this, but I'm not sure that this buys us anything in production since it gets executed in a lambda.

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,580 @@
...
+                context.clone(),
+                context.settings.trusted_reorg_api_key.clone(),
+                Chainstate {
+                    stacks_block_height: stacks_chaintip.block_height,
+                    stacks_block_hash: stacks_chaintip.block_hash.clone(),
+                },
+            ),
+            "failed to update chainstate in Emily",
+        )
+        .await?;
```
djordon, line 137: Does this bail get logged somewhere?

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,580 @@
...
+
+/// Helper function to handle internal API calls with error handling.
+async fn handle_internal_call<F, R>(api_call: F, error_msg: &str) -> Result<(), Error>
+where
+    F: Future<Output = R>,
+    R: Reply,
+{
+    let response = api_call.await.into_response();
+    if !response.status().is_success() {
+        tracing::error!("{}", error_msg);
```
djordon, line 429: nit: this can be simplified  to `tracing::error!("{error_msg}")`.

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,580 @@
...
+            handle_internal_call(
+                update_withdrawals(
+                    context.clone(),
+                    context.settings.trusted_reorg_api_key.clone(),
+                    UpdateWithdrawalsRequestBody {
+                        withdrawals: updated_withdrawals,
+                    },
+                ),
+                "failed to update withdrawals in Emily",
+            )
```
djordon, line 249: Just pointing out that we have logs for each of the individual events by this point. So if we encounter an error when processing withdrawals, we'd know which withdrawals were dropped. Perhaps we can be noisier here at launch and include more event details? It isn't necessary because we log the Stacks transaction ID and we can use the explorer to figure out the details of the event that was emitted, but it might be easier to do this via logs or something.

djordon (commented):
This looks good, just had a few comments.

The real thing that is bugging me is the warp boxed thing. I'm slightly worried that this indicative of some other warp issue that we'll get hit by in production. We should spend a little bit of time to get to the bottom of this so that we're confident that it's not an issue. Worst case we can migrate over to axum and enjoy some quality of life improvements with the move.

emily/handler/tests/integration/new_block.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    // Check if registry_event matches the expected variant
+    is_expected_variant(&registry_event)
+        .cloned() // Return a copy of the variant if it matches
+        .expect("Expected specified RegistryEvent variant")
+}
+
+// /// Test that the handler can handle a new block event with a valid payload
+// /// that contains a CompletedDeposit event.
+// /// The handler should update the chain state in Emily and mark the deposit as confirmed.
```
matteojug, line 57: nit: extra `//`

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,580 @@
...
+        new_block_event: NewBlockEventRaw,
+    ) -> Result<impl warp::reply::Reply, Error> {
+        let new_block_event = match new_block_event.deserialize() {
+            Ok(event) => event,
+            Err(error) => {
+                tracing::error!(%error, "failed to deserialize new block event");
+                return Err(error.into());
+            }
+        };
+        // Set the global limits.
```
matteojug, line 93: nit: leftover comment?

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,580 @@
...
+        let registry_address = SBTC_REGISTRY_IDENTIFIER.get_or_init(|| {
+            // Although the following line can panic, our unit tests hit this
+            // code path so if tests pass then this will work in production.
+            let contract_name = ContractName::from(SBTC_REGISTRY_CONTRACT_NAME);
+            let issuer = StandardPrincipalData::from(context.settings.deployer_address);
```
matteojug, line 98: Should we change the type of `deployer_address` in `StandardPrincipalData` directly? It doesn't seem we are using the fact that is a `StacksAddress`

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,580 @@
...
+        // the contract call.
+        let events = new_block_event
+            .events
+            .into_iter()
+            .filter(|x| x.committed)
+            .filter_map(|x| x.contract_event.map(|ev| (ev, x.txid)))
+            .filter(|(ev, _)| &ev.contract_identifier == registry_address && ev.topic == "print")
+            .collect::<Vec<_>>();
+
+        tracing::debug!(block = %events.len(), "received events from stacks-core");
```
matteojug, line 123: nit
```suggestion
        tracing::debug!(events = %events.len(), "received events from stacks-core");
```

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,580 @@
...
+        )
+        .await?;
+
+        if events.is_empty() {
+            // If there are no events to process, we return early with a 200 OK
+            // status code so that the node does not retry the webhook.
+            return Ok(warp::reply());
+        }
+
+        tracing::debug!(count = %events.len(), "processing events for new stacks block");
```
matteojug, line 145: nit: isn't this the same log as :123?

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,580 @@
...
+) -> WithdrawalUpdate {
+    tracing::debug!(topic = "withdrawal-accept", "handled stacks event");
+
+    WithdrawalUpdate {
+        request_id: event.request_id,
+        status: Status::Confirmed,
+        fulfillment: Some(Fulfillment {
+            bitcoin_block_hash: event.sweep_block_hash.to_string(),
+            bitcoin_block_height: event.sweep_block_height,
+            bitcoin_tx_index: event.outpoint.vout, // TODO: We don't have this information in the event
```
matteojug, line 343: nit: I think we can remove the comment -- once the signer start to correctly populate `event.outpoint.vout` this should start working without any change

matteojug (approved):
LGTM, the `boxed` thing is a bit suspicious though and I'm +1 the suggestion to dig a bit more into it before merging.

emily/handler/src/api/handlers/new_block.rs
Jiloc: Eventually, I always end up in doubt here..
 
I thought that the `bitcoin_tx_index` and `bitcoin_txid` of the Fulfillment were referring to the sweep tx. i.e. the sweet txid and the vout relative to this withdrawal. 

In the event we have the `sweep_txid`, and an `outpoint`. The `output` refers to `/// This is the outpoint for the bitcoin transaction that serviced the request.` that I thought was a different value?

djordon: > I thought that the `bitcoin_tx_index` and `bitcoin_txid` of the Fulfillment were referring to the sweep tx. i.e. the sweet txid and the vout relative to this withdrawal.

Yup, that is correct. Moreover, the `outpoint.txid` should always equal the `sweep_txid`. They're actually redundant, one of them doesn't need to be in the event or the smart contract. And the `outpoint.vout` should be correctly set right now, so I think we can remove the comment.

Jiloc: Great, thank you! About the vout, maybe we are still hardcoding it to 2 unless there is another PR fixing it? https://github.com/stacks-network/sbtc/blob/a9db98c6f4c3c97feec74b62a7dfb12d76efd78f/signer/src/storage/model.rs#L473

emily/handler/src/api/handlers/new_block.rs
Jiloc: Yes, Emily logs all responses to the routes https://github.com/stacks-network/sbtc/blob/2fe76e38df8513cdcc25aa7b2f15947e075b785f/emily/handler/src/api/routes/mod.rs#L69-L86

emily/handler/src/api/handlers/new_block.rs
djordon: 🤦🏽 nope, I was wrong about the vout. This should be an easy fix though.

djordon: I opened https://github.com/stacks-network/sbtc/issues/1454 for this.

emily/handler/src/api/handlers/new_block.rs
Jiloc: This is a great point! Often, we are logging a lot of unnecessary info but not the one that matter. In https://github.com/stacks-network/sbtc/pull/1416/commits/02a1bd09f6e443a9b2f151a72d140f69e5f0268a, I added better control over error handling for updating withdrawals and deposits.

I also introduced a new ValidationError, which I plan to use gradually to improve the errors returned by Emily. For example, we were previously returning an InternalError when entries failed validation—now, we have more specific errors to make debugging easier.

djordon (approved):
Looks good. Left a suggestion to store the `scriptPubKey` instead of the address.

Jiloc:
@matteojug @djordon Regarding boxed(): unfortunately seems like it's a "common" Warp issue.. as seen in [this discussion](https://github.com/seanmonstar/warp/issues/811).

In [this commit](https://github.com/stacks-network/sbtc/pull/1416/commits/90c2e08ca5c8394f0c6c81082b1152fea53665b8), I reduced the number of boxed() calls by keeping them only at the top-level route composition, making the types "short" enough to avoid compiler and runtime issues.

I looked into alternative solutions, such as increasing `RUST_MIN_STACK` and `#![recursion_limit = "256"]`, but I didn't like relying on those as a fix..

That said, I’d be happy to move to Axum :P

emily/handler/src/api/handlers/new_block.rs
djordon: 🙏🏽

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -467,9 +467,39 @@ pub async fn update_deposits(
...
+                    .await
+                    .map_err(|error| {
+                        tracing::error!(
+                            %bitcoin_txid,
+                            bitcoin_tx_output_index,
+                            %error,
+                            "failed to update deposit"
+                        );
+                        error
+                    })?;
```
djordon, line 490: For these cases you can use `.inspect_err`

djordon:
> @matteojug @djordon Regarding boxed(): unfortunately seems like it's a "common" Warp issue.. as seen in [this discussion](https://github.com/seanmonstar/warp/issues/811).
> 
> In [this commit](https://github.com/stacks-network/sbtc/pull/1416/commits/90c2e08ca5c8394f0c6c81082b1152fea53665b8), I reduced the number of boxed() calls by keeping them only at the top-level route composition, making the types "short" enough to avoid compiler and runtime issues.
> 
> I looked into alternative solutions, such as increasing `RUST_MIN_STACK` and `#![recursion_limit = "256"]`, but I didn't like relying on those as a fix..
> 
> That said, I’d be happy to move to Axum :P

I see. Well, thanks for taking a look.

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,568 @@
...
+) -> Result<CreateWithdrawalRequestBody, Error> {
+    tracing::debug!(topic = "withdrawal-create", "handled stacks event");
+    let params = if is_mainnet {
+        &params::MAINNET
+    } else {
+        &params::REGTEST
+    };
+    let recipient = bitcoin::Address::from_script(&event.recipient, params)
+        // This should never happen since the recipient address is already validated
+        .map_err(|e| Error::Debug(format!("Failed to create recipient address: {}", e)))?;
```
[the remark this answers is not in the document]
Jiloc: Yeah, that makes a lot of sense! Thanks for the suggestion. I was influenced by the comments referring to it as "The recipient Bitcoin address," but storing the  scriptPubKey is definitely the better approach

emily/handler/src/api/handlers/deposit.rs
Jiloc: changed in https://github.com/stacks-network/sbtc/pull/1416/commits/2bb9c9c75cd2175f01da14e7dd736bbbd39a8f75

emily/handler/src/api/handlers/new_block.rs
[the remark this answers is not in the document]
Jiloc: changed in https://github.com/stacks-network/sbtc/pull/1416/commits/a2e5ff2dae7097318c6adb50666f33e8bde4e6c7

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,514 @@
...
+        for withdrawal in created_withdrawals {
+            handle_internal_call(
+                create_withdrawal(context.clone(), withdrawal),
+                "failed to create withdrawal in Emily",
+            )
+            .await?;
+        }
+
+        if !updated_withdrawals.is_empty() {
+            handle_internal_call(
```
djordon, line 210: Sorry if I asked before, but for these `handle_internal_call` calls, we want the node to try again on an error here right?

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -0,0 +1,514 @@
...
+        if !completed_deposits.is_empty() {
+            handle_internal_call(
+                update_deposits(
+                    context.clone(),
+                    context.settings.trusted_reorg_api_key.clone(),
+                    UpdateDepositsRequestBody { deposits: completed_deposits },
+                ),
+                "failed to update deposits in Emily",
+            )
+            .await?;
```
djordon, line 195: Oh, I forgot! We may want to just log the error here too because we may want to support direct from L1 deposit requests (https://github.com/stacks-network/sbtc/issues/1168). These requests could be processed by the signers without informing Emily. In this case, Emily need not know about the deposit at all when the event comes in. Well, unless the signers tell Emily, which is also a possibility.

Then again, we can update this when we implement that feature so no need to change anything here.

djordon (approved):

emily/handler/src/api/handlers/new_block.rs
Jiloc: It's always good to double check :) Yes, we want to try again here. If this calls go wrong, is most certainly for an AWS/Dynamo problem. So we want to bail and have the sidecar/stacks-node try again. 

The case where we don't want the sidecar/stacks-node to try again is a bit up, where we parse the event. In this PR we retry, in a follow up PR we'll check for the missing withdrawals like we are doing for the deposits:

https://github.com/stacks-network/sbtc/blob/54ec10d7bbc5d707603351c85d6ea0ed6ff64805/emily/handler/src/api/handlers/new_block.rs#L143-L152

emily/handler/src/api/handlers/new_block.rs
Jiloc: Roger, good to know! Fortunately this case is already covered as described in the other comment👍

emily/handler/src/database/entries/withdrawal.rs
```diff
@@ -411,7 +411,7 @@ pub struct WithdrawalInfoByRecipientEntrySearchToken {
 #[derive(Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
 #[serde(rename_all = "PascalCase")]
 pub struct WithdrawalInfoByRecipientEntryKey {
-    /// The recipient of the withdrawal.
+    /// The The recipient Bitcoin hex-encoded scriptPubKey.
```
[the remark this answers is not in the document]
Jiloc: fixed in multiple places in https://github.com/stacks-network/sbtc/pull/1416/commits/fb75a78bcca09f5642c03b8d2d9823233aaeb11e

emily/handler/src/api/handlers/new_block.rs
djordon: Okay cool

PR #1421, opened 2025-02-23
[chore] remove emily calls from signer's new_block handler

Jiloc:
## Description

Part of https://github.com/stacks-network/sbtc/issues/1067

Following #1409, the deposits and withdrawal updates will be sent to Emily only from the sidecar. In this PR we remove all calls to Emily from the new-block handler, that will now be only used to store the events in the DB.

Currently, the only update that the signers send to Emily is the Deposit Accepted. We should probably consider whether we want to remove it as well.

 
## Changes
- Remove Emily calls from signer's new_block 
- Adapt sbtc event handlers to only store the data and log the event.

## Testing Information
- Simplify new block unit tests
- Remove the stacks_events_observer integration tests

## Checklist:


OfficialFedar (approved):

djordon (approved):
Looks good 🚀

djordon:
I think we can merge this because it doesn't resolve a ticket labeled as `immunefi-scope`. I think that we'd want to update the immunefi branch pretty soon (but not too soon) if this PR did resolve the ticket.

Issue #518, opened 2024-09-12
[Feature]: Explicit duplicate deposit `txid` check is missing


evonide:
# (Low) Explicit duplicate deposit `txid` check is missing

## 1. Description

When creating a deposit via Emily API you can specify a `txid` and `vout` to point to the UTXO that’s supposed to be used for depositing.

https://github.com/stacks-network/sbtc/blob/033e2cbd2def829a3e30df77bc274b325d68cdbe/emily/handler/src/api/handlers/deposit.rs#L194-L244

Potential duplicates or reuse of an existing `txid` shouldn't be a problem due to the database's design which has a primary (and hence unique) key over `(txid, output_index)`. However, there should be explicit checks inside the code to verify that a TX has not been seen before in addition to that to avoid potential long-term regressions.

aldur:
@Jiloc please comment and close (if appropriate).

Jiloc:
At the time this issue was opened, duplicate deposits were indeed prevented by the DynamoDB constraints on (txid, output_index), but the concern was still valid since the logic allowed overriding and replacing the original deposit data.

Since then, we have introduced validation against the raw transaction hex (https://github.com/stacks-network/sbtc/pull/1197), which means that even though the code still technically "[allows](https://github.com/stacks-network/sbtc/blob/84948f2b7ec6cc5d7ac502ef68a6b496bb8e66a2/emily/handler/src/api/handlers/deposit.rs#L348-L356)" an override for pending and reprocessing deposits (except for stacks_block_* fields), it is now practically impossible to modify an existing deposit with different data. This was originally designed to allow correcting deposit and reclaim scripts when needed, but with the current validation mechanism, a conflicting deposit request simply wouldn't pass.

So this is not a security concern as of now, but adding an explicit duplicate error would make this behavior clearer and more explicit

PR #1425, opened 2025-02-23
[chore] duplicate check on create_deposit

Jiloc:
## Description

Closes: #518 

## Changes
- return 200 if a copy of the deposit already exists.

## Testing Information
- added an integration test

## Checklist:

### DIFF
--- a/emily/handler/src/api/handlers/deposit.rs
+++ b/emily/handler/src/api/handlers/deposit.rs
@@ -162,7 +162,7 @@ pub async fn get_deposits(
     context: EmilyContext,
     query: GetDepositsQuery,
 ) -> impl warp::reply::Reply {
-    debug!("In get deposits");
+    debug!("in get deposits");
     // Internal handler so `?` can be used correctly while still returning a reply.
     async fn handler(
         context: EmilyContext,
@@ -331,8 +331,8 @@ pub async fn create_deposit(
         api_state.error_if_reorganizing()?;
 
         let chaintip = api_state.chaintip();
-        let mut stacks_block_hash: String = chaintip.key.hash;
-        let mut stacks_block_height: u64 = chaintip.key.height;
+        let stacks_block_hash = chaintip.key.hash;
+        let stacks_block_height = chaintip.key.height;
 
         // Check if deposit with such txid and outindex already exists.
         let entry = accessors::get_deposit_entry(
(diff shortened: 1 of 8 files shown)


emily/handler/src/api/handlers/deposit.rs
```diff
@@ -343,17 +343,9 @@ pub async fn create_deposit(
...
-            Ok(deposit) => {
-                if deposit.status != Status::Pending && deposit.status != Status::Reprocessing {
-                    return Err(Error::Conflict);
-                } else {
-                    // If the deposit is pending or reprocessing, we should keep height and hash same as in the old deposit
-                    stacks_block_hash = deposit.last_update_block_hash;
-                    stacks_block_height = deposit.last_update_height;
-                }
-            }
+            Ok(_) => return Err(Error::Conflict),
```
djordon, line 348: After thinking about this some more, I think that we should return `200 OK` here instead of an error. Well, we haven't validated the request yet so maybe we should do that before reaching out to the database. If the body passes validation then then we can reach out the database and return `200 OK` is it exists.

emily/handler/tests/integration/deposit.rs
```diff
@@ -972,11 +972,14 @@ async fn overwrite_deposit(status: Status, should_reject: bool) {
     assert_eq!(response.bitcoin_txid, bitcoin_txid);
     assert_eq!(response.status, status);
 
+    let duplicate_deposit =
+        apis::deposit_api::create_deposit(&configuration, create_deposit_body).await;
+
+    assert_eq!(duplicate_deposit.is_ok(), true);
```
matteojug, line 978: nit
```suggestion
    assert!(duplicate_deposit.is_ok());
```

emily/handler/tests/integration/deposit.rs
```diff
@@ -972,11 +972,14 @@ async fn overwrite_deposit(status: Status, should_reject: bool) {
...
+
     assert_eq!(
-        apis::deposit_api::create_deposit(&configuration, create_deposit_body)
-            .await
-            .is_err(),
-        should_reject
+        *expected_deposit.deposits.first().unwrap(),
```
matteojug, line 981: nit
```suggestion
        response,
```

matteojug (approved):

emily/handler/tests/integration/deposit.rs
Jiloc: changed in https://github.com/stacks-network/sbtc/pull/1425/commits/94d31e7344bf9f663ee8ec462d09b289562bf9d9

djordon (approved):
Looks good ✅

Issue #1192, opened 2025-01-09
[Bug]: Validate the coordinator proposed bitcoin/stacks fees


matteojug:
# Bug - Validate the coordinator proposed bitcoin/stacks fees

## 1. Description

The fee used for bitcoin/stacks transaction is decided by the coordinator, and the signers blindly accept it. Instead, they should validate that the proposed one is "reasonable" given their own view of bitcoin/stacks.

## 2. Why It's Necessary

Without validation, a coordinator can force the users to spend more than required in fees (still, within their specified max_fee) or waste the STX funds allocated for the stacks transactions fees.

AshtonStephens:
I'm going to pick this one up. The goal is to have a procedure that signers can follow to prevent themselves from approving transaction that uses far too high of a fee, but is also resilient against bad stacks fees being reported from the estimator.

Here's the procedure I'm planning to implement:

- the config will specify both a maximum and minimum sats / vbyte and stx fee.
- a constant will specify the % more than the signer's estimate that the coordinator's estimate can be for the signer to approve the transaction

1. If our fee estimate is outside of specified bounds then use the coordinator's fee suggestion unless coordinator suggestion is also out of bounds, in which case reject. _This will make the signer resilient to its own STX or BTC fee estimate being bad_
2. If coordinator suggests a fee more than constant x% above our estimate then reject
3. If coordinator suggests a fee that exceeds our maximum then reject

djordon:
So we think that for the check we do an exactness check for the claimed mempool transaction fees for any existing sweep and then have some liberal bounds on the fee rate.

PR #1431, opened 2025-02-24
feat: validate coordinator proposed bitcoin fees

AshtonStephens:
## Description

Closes the Bitcoin half of: #1192

This PR is from another PR (#1392) that got split up to separate Bitcoin and Stacks behavior.

This part was separated from the Stacks half because there's an issue with the way the transaction will be represented in the Signer database when a transaction is marked as invalid based on the coordinator's proposed fee:

> @djordon 
> The only trouble with this is that it's not clear why the the transaction is invalid when looking at the data in the signers' database. I wonder if we should do something else here instead. Maybe we could change the result for the signers' UTXO to be something other than Ok when the fee is too high?  

## Changes

- Adds logic to bitcoin transaction validation that considers a transaction to be invalid if the fee on it is more than 5 times plus 5 satoshis greater than the Signer estimates it should cost.
- Adds unit tests and integration tests

## Testing Information

unit and integration tests

## Checklist:

### DIFF
--- a/signer/src/bitcoin/rpc.rs
+++ b/signer/src/bitcoin/rpc.rs
@@ -23,12 +23,15 @@ use bitcoincore_rpc_json::GetNetworkInfoResult;
 use bitcoincore_rpc_json::GetRawTransactionResultVin;
 use bitcoincore_rpc_json::GetRawTransactionResultVout as BitcoinTxInfoVout;
 use bitcoincore_rpc_json::GetTxOutResult;
+use futures::future::try_join_all;
 use serde::Deserialize;
 use url::Url;
 
+use crate::bitcoin::utxo::Fees;
 use crate::bitcoin::BitcoinInteract;
 use crate::error::Error;
 
+use super::utxo;
 use super::GetTransactionFeeResult;
 use super::TransactionLookupHint;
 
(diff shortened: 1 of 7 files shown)


signer/src/bitcoin/rpc.rs
```diff
@@ -731,3 +734,107 @@ impl BitcoinInteract for BitcoinCoreClient {
         self.get_network_info()
     }
 }
+
+/// Assesses the total fees paid for any outstanding sweep transactions in
+/// the mempool which may need to be RBF'd. If there are no sweep
+/// transactions which are spending the signer's UTXO, then this function
+/// will return [`None`].
+///
+/// TODO: This method currently blindly assumes that the mempool transactions
```
cylewitruk, line 743: Could you create an issue and add it to the TODO? I think it sounds reasonable to have a sanity check..

signer/src/bitcoin/validation.rs
```diff
@@ -32,13 +32,26 @@ use crate::WITHDRAWAL_BLOCKS_EXPIRY;
...
 use super::utxo::RequestRef;
 use super::utxo::Requests;
 use super::utxo::SignatureHash;
 use super::utxo::UnsignedTransaction;
 use super::utxo::WithdrawalRequest;
 
+/// The maximum acceptable multiplier fee difference between the signer and
+/// coordinator estimates when the coordinator's estimate is greater than
+/// the signer's.
+const SIGNER_MAX_BITCOIN_FEE_MULTIPLIER: f64 = 5.0;
```
cylewitruk, line 48: So I remember there being a discussion about config vs. not and landing in that these shouldn't be exposed as config. What's the thinking around having config for Stacks max transaction fee but not these parameters?

signer/src/bitcoin/validation.rs
```diff
@@ -286,6 +316,7 @@ impl BitcoinPreSignRequest {
         btc_ctx: &BitcoinTxContext,
         requests: &'a TxRequestIds,
         signer_state: SignerBtcState,
+        signers_view_of_last_fees: &mut Option<Fees>,
```
cylewitruk, line 319: Hm, I'm not really following why this is `&mut` and then you're clearing it at the end? Is it that you only want this to be provided for the first transaction in the event of chained utxo's in mempool? If so then I would probably change the `for` loop above to use `.enumerate()` and only provide this on the first iteration instead of doing this dance.

Issue #1334, opened 2025-02-10
[Feature]: Use stacks-core in CI


djordon:
# Feature – Use stacks-core in CI

## 1. Description

All of our enabled integration tests currently mock `stacks-core`. This is far from ideal because our integration tests do not reveal regressions. As we've changed parts of the `sbtc` signers to rely on `stacks-core` directly, the absence of a real `stacks-core` environment has become more problematic.

### 1.1 Context & Purpose

We aim to get as close as possible to a testnet environment during our integration tests. Integrating with `stacks-core` in CI has been a longstanding goal, and it's about time we tackle it. However, some known challenges exist, the largest being that we require `stacks-core` to be on the latest hard-fork, Epoch 3.x. The real issue is that to use `stacks-core` in Epoch 3.0, the node must progress through the earlier epochs, which can take some time do. It likely needs to go through at least one "Cycle" in Epoch 2.5 before advancing to Epoch 3.0. This can take ~1-3 minutes, depending on how aggressively we can set the block production parameter in CI.

## 2. Technical Details

Although we cannot start `stacks-node` directly in Epoch 3.0, we don't need to wait for the chain to progress to Epoch 3.0. We can boot the node using "existing" chain state data so it essentially starts in the right epoch. Here's the general approach:

1. Modify our local `devenv` to produce Bitcoin blocks and Stacks blocks only until shortly after the start of Epoch 3. Doing so requires modification to the `bitcoin-miner` docker service to stop bitcoin block production and the `tx-broadcaster` docker services as stacks blocks get generated when there is a transaction in the mempool.
2. Remove the sbtc signers event observers entries from the stacks node config. Basically, remove this stuff https://github.com/stacks-network/sbtc/blob/1c0b804241ad279b346434f7242257ddd8c041d1/docker/stacks/stacks-regtest-miner.toml#L65-L88
3. Start a modified version of our local `devenv` and wait for the node to reach Epoch 3.0, which should be around Bitcoin block 233.
4. Extract the Bitcoin blockchain data directory from the Bitcoin container, as well as the Stacks blockchain and API data for its directory.
5. Figure out how to make the blockchain data accessible in CI. Ideally, we would avoid committing it to the repo and instead download it from S3 or some GitHub storage service.

Note that in order to the run a stacks node you need to run a stacks signer. Also, 

### 2.1 Acceptance Criteria

- [ ] At least one of our integration tests runs against a Stacks node without needing to wait a long time for the node to sync.
[...]

PR #1438, opened 2025-02-25
Feature – Use stacks-core in CI

tonyjzhou:
## Description

Closes: #1334

### At least one of our integration tests (**contracts::estimate_tx_fees**) runs against a Stacks node within 3 seconds
```
make integration-env-down-ci && make integration-env-up-ci && sleep 3 && cargo test -p signer --test integration -- contracts::estimate_tx_fees --nocapture
```

### **contracts::estimate_tx_fees** consistently runs among other Integration tests within 3 seconds
```
make integration-env-down-ci && make integration-env-up-ci && sleep 3 && make integration-test
```

### Access the bitcoin in an instant
```
make integration-env-down-ci && make integration-env-up-ci && docker exec docker-bitcoind-1 bitcoin-cli getblockchaininfo
```

### Access stacks blockchain in an instant
```
make integration-env-down-ci && make integration-env-up-ci && curl -s http://localhost:20443/v2/info
```

## Checklist:

- [v] I have performed a self-review of my code
- [v] My changes generate no new warnings
- [v] New and existing unit tests pass locally with my changes
- [v] Any dependent changes have been merged and published in downstream modules


cylewitruk:
Nice! I'll check this out when I have some time to dig my heels in. 

I'm wondering if we shouldn't use git LFS for the database/chainstate files like we touched on last week, so that the repo deltas don't explode with these in the commit history -- i.e. I'm a little worried about the overhead they can add to regular git operations.

docker/stacks/stacks-regtest-miner.toml
```diff
@@ -70,7 +70,7 @@ events_keys = [
 
 # Add sbtc-signer as an event observers
 [[events_observer]]
-endpoint = "host.docker.internal:8801"
+endpoint = "sbtc-signer-1:8801"
```
matteojug, line 73: With this change we cannot run devenv with local signers (running them via `./signers run 3`)

docker/testnet/stacks/Config.toml
```diff
@@ -37,7 +37,7 @@ private_neighbors = false
 # delete the file `event_observers.sqlite` in your chainstate folder
 # to prune old events not yet delivered.
 [[events_observer]]
-endpoint = "sbtc-signer:8801"
+endpoint = "sbtc-signer-1:8801"
```
matteojug, line 40: This is testnet related, why do we need the change?

docker/docker-compose.test.yml
matteojug, line 1: This looks like an old version of `docker-compose.yml`, can't we use that one directly?

docker/docker-compose.test.yml
```diff
@@ -74,22 +123,531 @@ services:
...
+      DATABASE_HOST: "mariadb"
+      DATABASE_DATABASE: "mempool"
+      DATABASE_USERNAME: "mempool"
+      DATABASE_PASSWORD: "mempool"
+      STATISTICS_ENABLED: "true"
+    command: "./wait-for-it.sh mariadb:3306 --timeout=720 --strict -- ./start.sh"
+    profiles:
+      - bitcoin-mempool
+
+  bitcoind:
```
matteojug, line 615: There's already a bitcoin container at the beginning, why do we need a separate one?

docker/docker-compose.test.yml
```diff
@@ -74,22 +123,531 @@ services:
...
+      - 18444:18444
+      - 28332:28332
+      - 28333:28333
+    healthcheck:
+      test: ["CMD-SHELL", "bitcoin-cli getblockcount"]
+      interval: 3s
+      timeout: 1s
+      retries: 3
+
+  postgres:
```
matteojug, line 640: what is this used for?

docker/docker-compose.ci.yml
matteojug, line 1: Similar comment to the other: why can't we reuse `docker/docker-compose.yml` directly?

docker/docker-compose.ci.yml
```diff
@@ -1,14 +1,92 @@
...
+    - |
+      set -e
+      envsubst < config.toml.in > config.toml
+      exec stacks-signer run --config config.toml
+  profiles:
+    - default
+
+# Services
+services:
+  # Existing services from docker-compose.ci.yml
```
matteojug, line 84: It seems `bitcoind` and `postgres` here are redundant (we already have a bitcoin node in here, and we could reuse the db of signer-1 already exposed on 5432), why do we need those?

Makefile
```diff
@@ -72,8 +72,8 @@ nextest-archive-clean:
 # INTEGRATION TESTS
 # ##############################################################################
 
-integration-env-up: emily-cdk-synth
-	docker compose --file docker/docker-compose.test.yml up -d
+integration-env-up: emily-cdk-synth clean-nakamoto-headers
+	docker compose -f docker/docker-compose.test.yml --profile default --profile bitcoin-mempool --profile sbtc-signer up -d
```
matteojug, line 76: For integration tests I don't think we want `bitcoin-mempool` (it spawns also mempool api & co), nor the `sbtc-signer`s. We also probably don't want the bitcoin miner,  tx-broadcaster nor some other container in there. Also applies to the CI variant.

If the sbtc-signers are up beacuse of the event observer, it's probably better to handle it mounting different configurations (it should be fine wrt the state, assuming a clean event-observer db).

Makefile
```diff
@@ -82,21 +82,21 @@ integration-test-build:
 	cargo $(CARGO_FLAGS) test build $(CARGO_EXCLUDES) --test integration --no-run --locked
 
 integration-env-down:
-	docker compose --file docker/docker-compose.test.yml down -t 0 -v
+	docker compose -f docker/docker-compose.test.yml --profile default --profile bitcoin-mempool --profile sbtc-signer down -t 0
```
matteojug, line 85: What volumes do we need to keep (the missing `-v`)?

Makefile
```diff
@@ -82,21 +82,21 @@ integration-test-build:
...
+	docker compose -f docker/docker-compose.test.yml build
 
 integration-test-full: integration-env-down integration-env-up integration-test integration-env-down
 
-integration-env-up-ci: emily-cdk-synth
-	docker compose --file docker/docker-compose.ci.yml up --detach --quiet-pull
-	DYNAMODB_ENDPOINT=http://localhost:8000 \
+integration-env-up-ci: emily-cdk-synth clean-nakamoto-headers
+	docker compose --file docker/docker-compose.ci.yml --profile default --profile bitcoin-mempool --profile sbtc-signer up -d
+	env DYNAMODB_ENDPOINT=http://localhost:8000 \
```
matteojug, line 94: Is this because of some OS incompatibility?

matteojug (commented):
Nice! Do you mind expanding a bit (in the PR description maybe):
 - What's the idea behind this and on what does it rely to work? IIRC you mentioned clearing out the headers or something
  - How is the committed chainstate generated and how to regenerate it (should one requires it)

tonyjzhou:
Sorry for the late reply. @djordon 

The chainstate data was generated by running `make integration-env-up` and waited until the Stack node reached Epoch 3.x.

It can be **re-generated** with `make integration-env-up` or `make integration-env-up-ci` again _on a blank slate_.


It works because these build targets make use of the updated **stacks-node** volume configuration. When it's running _for the first time_ on a blank slate, it will generate the chainstate (and other related data) and persist them on disk. When it runs next time, it will reuse the data.

Hope that helps.

djordon:
> It can be **re-generated** with `make integration-env-up` or `make integration-env-up-ci` again _on a blank slate_.

Oh, so simple! Thanks for taking the time comment. So, I think we can basically finish this after all ❤️

PR #1445, opened 2025-02-27
1444 chorebug restrict wipe databases to testing api

### LINKED ISSUE #1444: [Chore/Bug]: Restrict `wipe_databases` to Testing API
# Chore/Bug - Restrict `wipe_databases` to Testing API  

## 1. Description  

The `wipe_databases` testing endpoint is currently exposed in the Emily public API, even though access is restricted by the signers' API keys. This means it is available in all deployments, including testnet.  

This task will disable the endpoint in the public API, ensuring it is only accessible through the testing API.  

NOTE: The endpoint verbs are manually deleted in production via the API Gateway.

## 2. Why It's Necessary  

- The endpoint is meant for testing purposes and should not be exposed in public-facing APIs, even with API key protection.  

## 3. Execution Checklist  

- [ ] Disable the `wipe_databases` endpoint in the Emily public API.

Jiloc:
## Description

Closes #1444 

The endpoint was already disabled in the private API (used by the sidecar) and will now only be accessible through the testing API and client.

I'm not very happy with the `as_testing()` method and am open to suggestions. Ideally, all test calls to Emily should go through the test-emily-client, but this would require a larger refactor since the structs used in the two crates are different.

## Changes
- disable the wipe_databases endoint.
- add the test-emily-client crate to the dev configuration of the signer.
- use the test-emily-client crate to call the wipe-databases.  

### DIFF
--- a/emily/openapi-gen/generators/public.rs
+++ b/emily/openapi-gen/generators/public.rs
@@ -33,7 +33,7 @@ use super::CorsSupport;
         api::handlers::chainstate::set_chainstate,
         // api::handlers::chainstate::update_chainstate, // signers may not set the chainstate to cause a reorg.
         // Testing endpoints.
-        api::handlers::testing::wipe_databases,
+        // api::handlers::testing::wipe_databases, // The public api cannot perform testing operations.
         // Limits endpoints.
         api::handlers::limits::get_limits,
         // api::handlers::limits::set_limits, // signers / users may not set limits.
--- a/signer/tests/integration/setup.rs
+++ b/signer/tests/integration/setup.rs
@@ -13,6 +13,7 @@ use blockstack_lib::types::chainstate::StacksAddress;
 use clarity::types::chainstate::StacksBlockId;
 use clarity::vm::types::PrincipalData;
 
+use emily_client::apis::configuration::Configuration as EmilyApiConfiguration;
 use fake::Fake;
 use fake::Faker;
 use rand::rngs::OsRng;
@@ -57,6 +58,8 @@ use signer::testing::context::TestContext;
 use signer::testing::context::*;
 use signer::testing::dummy::Unit;
 use signer::DEFAULT_MAX_DEPOSITS_PER_BITCOIN_TX;
+use testing_emily_client::apis::configuration::ApiKey as TestingEmilyApiKey;
+use testing_emily_client::apis::configuration::Configuration as TestingEmilyApiConfiguration;
 
 use crate::utxo_construction::generate_withdrawal;
 use crate::utxo_construction::make_deposit_request;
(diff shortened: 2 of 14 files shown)


signer/tests/integration/setup.rs
```diff
@@ -75,6 +78,27 @@ impl AsBlockRef for bitcoincore_rpc_json::GetBlockHeaderResult {
     }
 }
 
+pub trait IntoEmilyTestingConfig {
+    fn as_testing(&self) -> TestingEmilyApiConfiguration;
+}
+
+impl IntoEmilyTestingConfig for EmilyApiConfiguration {
+    fn as_testing(&self) -> TestingEmilyApiConfiguration {
```
djordon, line 86: I don't think we need the new trait, we could implement `From<&EmilyApiConfiguration>` for `TestingEmilyApiConfiguration` instead. Then we could use `.into()` all over the place.

emily/openapi-gen/generators/public.rs
```diff
@@ -33,7 +33,7 @@ use super::CorsSupport;
         api::handlers::chainstate::set_chainstate,
         // api::handlers::chainstate::update_chainstate, // signers may not set the chainstate to cause a reorg.
         // Testing endpoints.
-        api::handlers::testing::wipe_databases,
+        // api::handlers::testing::wipe_databases, // The public api cannot perform testing operations.
```
djordon, line 36: Should we add a test that this endpoint doesn't exist, in case someone accidentally tries to uncomment this line?

signer/tests/integration/setup.rs
Jiloc: I initially tried that way, but I ran into the orphan rule which doesn't allow me to implement the From trait because both types are from external crates.. Is there a workaround?

djordon: Oh right. What you did here is one workaround, the others feel a little off. Is there an easy way for us to add the `From` impl in the `testing-emily-client` crate?

Jiloc: I looked into that, and it seems like the two common ways to do that are:
- to manually add the From implementations after generation—this would mean modifying the testing-emily-client crate each time, adding the public-client dependency, before building. But since the changes would be wiped on every regeneration, this isn’t really a viable option. 
- Use a custom openapi client template. This would allow to inject the From implementation automatically (I still have to look into how to add the dependency on the public-client crate).
The first option is a definite no. As for the second, I worry it might make the client generation process even more of a black box than it already is.

@AshtonStephens, do you have any suggestions on this? Maybe you know a simpler way?

emily/openapi-gen/generators/public.rs
Jiloc: About the testing feature- we also found out at the same time that we are compiling all of the binaries with the testing feature on by default. That's why the endpoint was teachable. Daniel opened a PR to fix it on the signer #1446 . We still have to do the same for emily

djordon: Hmm, good points; basically, make sure the binary gets compiled without the "testing" feature. Right now this all gets compiled by [default](https://github.com/stacks-network/sbtc/blob/f80107143cbcb111a049ac95e56a997d4164468c/emily/handler/Cargo.toml#L10), since it is part of the default features and we don't pass in the `--no-default-features` flag during compilation.

I'm cool with either the endpoint from Emily, like what we're going here, or disabling compilation of the code path, or both. I'd vote for both but maybe that's more annoying to do for one reason or another, so I'm okay with this. And we can always tackle removing the default feature change in follow-up PRs.

Jiloc: I would also vote for having both, as we just demonstrated that one wasn't enough to prevent it..

djordon: > Regarding the test to check if the endpoint is not exposed, it's not straightforward.. On the integration env, we run all the routes available **with** the `testing` feature, which includes also the `testing/wipe`. And we need it for the integration tests.

Yeah, makes sense that it's not straightforward to test. Let's go without it then.

signer/tests/integration/setup.rs
djordon: Yeah no easier way than doing what we're doing here. This is fine then 👍🏽.

PR #1446, opened 2025-02-27
build: do not compile signer crate with testing feature

djordon:
## Description

For folks using VS Code, consider adding the following to you `settings.json`:
```json
"rust-analyzer.cargo.features": [
    "testing"
],
```
This way rust-analyzer with continue to work as usual, and won't complain about phantom issues.

## Changes

* Remove the "testing" feature from the list of default features in the signer binary
* Move the `in_memory` storage module to be conditionally compiled in tests and the "testing" feature.
* Minor clean-up after the "testing" feature is no longer enabled by default.

## Testing Information

This is only a build change, so if the tests pass thing things should work. I also tested this change on local devenv and things worked fine.


Makefile
```diff
@@ -20,14 +20,14 @@ install:
 	pnpm --recursive install
 
 build: blocklist-client-codegen emily-client-codegen contracts
-	cargo $(CARGO_FLAGS) build --all-targets $(CARGO_EXCLUDES) ${CARGO_BUILD_ARGS}
+	cargo $(CARGO_FLAGS) build --features "testing" --all-targets $(CARGO_EXCLUDES) ${CARGO_BUILD_ARGS}
```
Jiloc, line 23: should we have `--features "testing"` even in `make build`?

djordon: Hmmm, great question. No we shouldn't. I added it by mistake, all we need is `make test-build`. I will remove this.

djordon: Fixed in https://github.com/stacks-network/sbtc/pull/1446/commits/99480c11ff60c89c226f1b04e8d11faf82aa8712

Jiloc (approved):
lgtm!

signers.sh
```diff
@@ -75,12 +75,12 @@ exec_run() {
 }
 
 exec_demo() {
-  cargo run -p signer --bin demo-cli donation --amount 10000
-  cargo run -p signer --bin demo-cli deposit --amount 10000000 # 0.1 BTC
+  cargo run -p signer --bin demo-cli --features "testing" donation --amount 10000
+  cargo run -p signer --bin demo-cli --features "testing" deposit --amount 10000000 # 0.1 BTC
```
matteojug, line 79: qq: do we need this? Given that we have `required-features = ["testing"]` in `demo-cli` stanza.

djordon: Yes, unfortunately I think that we do. I believe the `required-features = ["testing"]` line just says that we need to specify the features for that binary. I think that I'll update the `demo_cli` so that it isn't necessary, because it is a pain to have to specify the feature.

matteojug: Or we can add an option to `signers.sh` to just run `cargo run -p signer --bin demo-cli --features "testing" $@`

djordon: I changed the demo-cli so that it no longer requires the testing feature in https://github.com/stacks-network/sbtc/pull/1446/commits/56937fa8d94d257c65e84b49b4c35483860423d7

Issue #1405, opened 2025-02-20
[Chore]: Use wrapper structs for bitcoin and stacks block height


MCJOHN974:
# Chore - Use wrapper structs for bitcoin and stacks block height

## 1. Description

We want to create structs like `struct BitcoinBlockHeight(u64)` and `struct StacksBlockHeight(u64)` and switch from using raw integers to using this structs

## 2. Why It's Necessary

It shows semantics of variable more explicitly. It can prevent some confusions and misunderstandings.

## 3. Execution Checklist

- [ ] structs for bitcoin and stacks heights are implemented with all necessary functions
- [ ] all raw integers representing stacks/bitcoin block height replaced by this structs

MCJOHN974:
It was actually closed by #1448

PR #1448, opened 2025-02-28
[chore]: use wrapper structs for heights

MCJOHN974:
## Description

Switch from raw `u64` to `struct BlockchainNameBlockHeight(u64)` in `signer`

Closes: #1405

## Changes

Scope: refactoring was only in `signer`
I will not promise that I changed 100% of heights to wrappers, but I think so. 

## Testing Information

Since it is refactoring we do not need any new tests.

## Checklist:


signer/src/config/mod.rs
```diff
@@ -339,7 +339,7 @@ pub struct SignerConfig {
     /// already been run, the coordinator will attempt to re-run DKG after this
     /// block height is met if `dkg_target_rounds` has not been reached. If DKG
     /// has never been run, this configuration has no effect.
-    pub dkg_min_bitcoin_block_height: Option<NonZeroU64>,
```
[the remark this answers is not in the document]
djordon: Yeah, it doesn't need to be nonzero, because there is nothing particularly bad here with setting it to zero. So removing that constraint is fine by me.

signer/src/storage/model.rs
```diff
@@ -1337,6 +1335,178 @@ pub struct WithdrawalRejectEvent {
...
+            }
+        }
+    };
+}
+
+/// Implements various traits and functions for integer wrapper types.
+/// Most of this impls are arithmetics and conversions.
+macro_rules! implement_int {
+    ($type:ident, $inner:ty) => {
+        $crate::implement_from!($type, $inner, u8, u16, u32, u64, usize, i32);
```
setzeus, line 1430: Just confirming my understanding here but this specific line is what guarantees that a u64.into() into type 'BitcoinBlockHeight'? Like so: block_height: block.header.chain_length.into(),

signer/src/bitcoin/validation.rs
```diff
@@ -834,7 +835,7 @@ impl DepositRequestReport {
         match self.lock_time {
             LockTime::Blocks(height) => {
                 let max_age = height.value().saturating_sub(DEPOSIT_LOCKTIME_BLOCK_BUFFER) as u64;
-                if deposit_age >= max_age {
+                if deposit_age >= max_age.into() {
```
cylewitruk, line 838: It feels a little like we should be working with integers and not heights here as we're dealing with "age". Maybe do the `.into()` on the `.saturating_sub()` above?

signer/src/bitcoin/validation.rs
```diff
@@ -989,11 +990,11 @@ impl WithdrawalRequestReport {
         }
 
         let block_wait = bitcoin_chain_tip_height.saturating_sub(self.bitcoin_block_height);
-        if block_wait < WITHDRAWAL_MIN_CONFIRMATIONS {
+        if block_wait < WITHDRAWAL_MIN_CONFIRMATIONS.into() {
```
cylewitruk, line 993: Same here and below, I'd move the `.into()` since we're not dealing with heights anymore and rather a diff.

signer/src/bitcoin/validation.rs
```diff
@@ -1465,10 +1467,10 @@ mod tests {
             recipient: TEST_RECIPIENT.clone(),
             // This needs to be WITHDRAWAL_MIN_CONFIRMATIONS less than the
             // chain_tip_height.
-            bitcoin_block_height: 0,
+            bitcoin_block_height: 0.into(),
         },
-        // This is part of sBTC consensus.
-        chain_tip_height: WITHDRAWAL_MIN_CONFIRMATIONS,
+        // This is partScriptBuf::new() of sBTC consensus.
```
cylewitruk, line 1472: Intellisense mixup? Don't think this was intended to be changed.

signer/src/config/mod.rs
```diff
@@ -799,7 +800,7 @@ mod tests {
         let settings = Settings::new_from_default_config().unwrap();
         assert_eq!(
             settings.signer.dkg_min_bitcoin_block_height,
-            Some(NonZeroU64::new(42).unwrap())
+            Some(BitcoinBlockHeight::from(42))
```
cylewitruk, line 803: nit: `42.into()` for consistency?

signer/src/context/signer_state.rs
```diff
@@ -116,16 +117,16 @@ impl SignerState {
...
     }
 
     /// Set the sbtc start height
-    pub fn set_sbtc_bitcoin_start_height(&self, height: u64) {
+    pub fn set_sbtc_bitcoin_start_height(&self, height: BitcoinBlockHeight) {
         self.is_sbtc_bitcoin_start_height_set
             .store(true, Ordering::SeqCst);
         self.sbtc_bitcoin_start_height
-            .store(height, Ordering::SeqCst);
+            .store(height.into(), Ordering::SeqCst);
```
cylewitruk, line 129: nit: I think elsewhere you deref (`*`) when going from struct -> int?

signer/src/stacks/contracts.rs
```diff
@@ -338,7 +339,7 @@ impl AsContractCall for CompleteDepositV1 {
             ClarityValue::UInt(self.amount as u128),
             ClarityValue::Principal(self.recipient.clone()),
             ClarityValue::Sequence(SequenceData::Buffer(burn_hash_buff)),
-            ClarityValue::UInt(self.sweep_block_height as u128),
+            ClarityValue::UInt(u64::from(self.sweep_block_height) as u128),
```
cylewitruk, line 342: nit: I might have implemented `From<XxxBlockHeight> for u128` to be able to skip the cast (and since it would always be valid).

signer/src/storage/model.rs
```diff
@@ -310,9 +312,8 @@ pub struct WithdrawalRequest {
     pub sender_address: StacksPrincipal,
     /// The block height of the bitcoin blockchain when the stacks
     /// transaction that emitted this event was executed.
-    #[sqlx(try_from = "i64")]
```
cylewitruk, line 313: QQ: Couldn't we have implemented `TryFrom<i64>` for the newtypes and kept this, and skipped the custom impls in `sqlx.rs`?

cylewitruk (commented):
I've still got a bit to go through, but leaving my comments so far

signer/src/storage/model.rs
```diff
@@ -1352,6 +1349,178 @@ pub struct WithdrawalRejectEvent {
     pub signer_bitmap: BitArray<[u8; 16]>,
 }
 
+/// Helper for [`implement_trait!`] macro. Implements arithmetic traits for integer wrapper types.
+#[macro_export]
```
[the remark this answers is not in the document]
cylewitruk: IMHO we should do a bigger lift and move all of our not-specifically-storage-related types out of this module, in which case it would make more sense for these macros to be co-located with their types and thus a bit more encapsulated.

But as that's a bit out-of-scope for this PR, I think that skipping the macros and only implementing the necessary conversion traits here would probably be better to remain consistent with how things are currently done in this file.

signer/src/api/info.rs
```diff
@@ -655,10 +652,11 @@ mod tests {
                 settings.signer.signer_round_max_duration = Duration::from_secs(2);
                 settings.signer.bitcoin_presign_request_max_duration = Duration::from_secs(3);
                 settings.signer.dkg_max_duration = Duration::from_secs(4);
-                settings.signer.sbtc_bitcoin_start_height = Some(101);
+                settings.signer.sbtc_bitcoin_start_height = Some(BitcoinBlockHeight::from(101u64));
```
matteojug, line 655: It seems we mostly use `.into()` for the conversions, is there a reason to prefer the explicit version here (and below for `dkg_min_bitcoin_block_height`)?

signer/src/storage/model.rs
```diff
@@ -145,10 +148,9 @@ impl From<bitcoin::Block> for BitcoinBlock {
 pub struct StacksBlock {
     /// Block hash.
     pub block_hash: StacksBlockHash,
-    /// Block height.
-    #[sqlx(try_from = "i64")]
+    /// Block height.    
     #[cfg_attr(feature = "testing", dummy(faker = "0..u32::MAX as u64"))]
-    pub block_height: u64,
+    pub block_height: StacksBlockHeight,
```
matteojug, line 153: qq: why do we still need the `cfg_attr` for this (and other) while we removed it in `BitcoinBlock::block_height`?

signer/src/storage/model.rs
```diff
@@ -1352,6 +1349,260 @@ pub struct WithdrawalRejectEvent {
...
+}
+impl From<usize> for BitcoinBlockHeight {
+    fn from(value: usize) -> Self {
+        Self(value as u64)
+    }
+}
+
+// Conversion u64 => StacksBlockHeight is not implemented intentionally.
+// Use deref instead.
+
```
matteojug, line 1380: qq: why? Also s/StacksBlockHeight/BitcoinBlockHeight/

signer/src/storage/model.rs
```diff
@@ -1352,6 +1349,260 @@ pub struct WithdrawalRejectEvent {
...
+    type Error = TryFromIntError;
+    fn try_from(value: BitcoinBlockHeight) -> Result<Self, Self::Error> {
+        i64::try_from(value.0)
+    }
+}
+
+impl TryFrom<i64> for BitcoinBlockHeight {
+    type Error = TryFromIntError;
+    fn try_from(value: i64) -> Result<Self, Self::Error> {
+        Ok(Self(<u64>::try_from(value)?))
```
matteojug, line 1408: nit? also for stacks one
```suggestion
        u64::try_from(value).map(Self)
```

signer/src/storage/model.rs
```diff
@@ -1352,6 +1349,260 @@ pub struct WithdrawalRejectEvent {
...
+    Hash,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Serialize,
+    Deserialize,
+    sqlx::FromRow,
+)]
+#[sqlx(transparent)]
```
matteojug, line 1586: nit: it doesn't seem we are using `sqlx(transparent)` (since we implemented sqlx encode/decode), do we need this?

signer/src/testing/dummy.rs
```diff
@@ -510,6 +512,30 @@ impl fake::Dummy<fake::Faker> for SigHash {
...
+    fn dummy_with_rng<R: rand::Rng + ?Sized>(_config: &std::ops::Range<u64>, rng: &mut R) -> Self {
+        rng.next_u32().into()
+    }
+}
+
+impl fake::Dummy<std::ops::Range<u64>> for StacksBlockHeight {
+    fn dummy_with_rng<R: rand::Rng + ?Sized>(_config: &std::ops::Range<u64>, rng: &mut R) -> Self {
+        rng.next_u32().into()
+    }
+}
```
matteojug, line 525: Shouldn't those use `_config`? If not, do we need those two (vs the general version below)

signer/src/block_observer.rs
```diff
@@ -791,7 +791,12 @@ mod tests {
             .with_stacks_client(test_harness.clone())
             .with_emily_client(test_harness.clone())
             .with_bitcoin_client(test_harness.clone())
-            .modify_settings(|settings| settings.signer.sbtc_bitcoin_start_height = min_height)
+            .modify_settings(|settings| {
+                settings.signer.sbtc_bitcoin_start_height = match min_height {
```
cylewitruk, line 795: You could probably use a simple `.map()` instead of a whole `match` (here and below).

signer/tests/integration/transaction_signer.rs
```diff
@@ -325,7 +325,7 @@ pub async fn presign_requests_with_dkg_shares_status(status: DkgSharesStatus, is
         .unwrap();
     let chain_tip = BitcoinBlockRef {
         block_hash: block_header.hash.into(),
-        block_height: block_header.height as u64,
+        block_height: (block_header.height as u64).into(),
```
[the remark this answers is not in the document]
cylewitruk: But otoh `block_header.height.into()` should work given all the conversion impls you have?

signer/src/api/info.rs
MCJOHN974: I used explicit `BitcoinBlockHeight::from` instead of `into()` in cases where rust is not smart enough to understand into which type it should cast. However, we even bumped rust version since I did most of my changes, so maybe in some cases it can be rewrited to `.into()`

signer/src/storage/model.rs
```diff
@@ -98,9 +102,7 @@ pub struct BitcoinBlock {
     /// Block hash.
     pub block_hash: BitcoinBlockHash,
     /// Block height.
-    #[sqlx(try_from = "i64")]
-    #[cfg_attr(feature = "testing", dummy(faker = "0..i64::MAX as u64"))]
```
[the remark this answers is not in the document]
MCJOHN974: Yeah, we don't need this in other places as well I think

signer/src/storage/model.rs
MCJOHN974: WoW, nice!

signer/src/storage/model.rs
MCJOHN974: In some cases where rust is not smart enough to understand how to interpret `.into()` it was much nicer looking to use deref:
`*height` vs `u64::from(height)`.

And I disabled this to make everything more consistent

signer/src/storage/model.rs
MCJOHN974: Yea, need to remove this, thanks

signer/src/storage/model.rs
```diff
@@ -1352,6 +1349,260 @@ pub struct WithdrawalRejectEvent {
...
+        Self(value)
+    }
+}
+impl From<usize> for StacksBlockHeight {
+    fn from(value: usize) -> Self {
+        Self(value as u64)
+    }
+}
+
+// Conversion u64 => StacksBlockHeight is not implemented intentionally.
```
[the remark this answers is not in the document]
MCJOHN974: Yeah, thanks!

signer/src/storage/model.rs
```diff
@@ -1352,6 +1349,260 @@ pub struct WithdrawalRejectEvent {
...
+    }
+}
+impl Add for BitcoinBlockHeight {
+    type Output = BitcoinBlockHeight;
+    fn add(self, other: BitcoinBlockHeight) -> Self::Output {
+        Self(self.0.add(other.0))
+    }
+}
+
+// I don't see any case where int - height can make sense.
```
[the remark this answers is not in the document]
MCJOHN974: Maaaybe it can still be useful to have a not why we don't have int - height implementation but it worth rephrasing this, at least "I" probably don't sound good in comments

signer/src/testing/storage/model.rs
```diff
@@ -536,15 +537,15 @@ impl BitcoinBlockRef {
     fn hallucinate_parent(block: &model::BitcoinBlock) -> Self {
         Self {
             block_hash: block.parent_hash,
-            block_height: 1337, // Arbitrary number
+            block_height: 1337u64.into(), // Arbitrary number
```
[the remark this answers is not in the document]
MCJOHN974: It will take sooooo lot of manual changes =(

signer/src/testing/dummy.rs
matteojug: If tests were using a specific range previously, not using `_config` may be changing the test in some unintended way (eg, potentially generating only "smaller" numbers). Can you check if we are using the ranged version for anything meaningful, and if not switch to the other one?

signer/src/storage/model.rs
MCJOHN974: This conversion was actually requested by @cylewitruk here: https://github.com/stacks-sbtc/sbtc/pull/1448#discussion_r2026921374

For me personally I don't think it making me having weird feelngs, but I also agree that usage of u128 is not so wide, so I'm fine with both. Thus leaving this discussion to you two =)

signer/tests/integration/postgres.rs
```diff
@@ -1850,7 +1852,8 @@ async fn is_signer_script_pub_key_checks_dkg_shares_for_script_pubkeys() {
         signature_share_threshold: 1,
         dkg_shares_status: Faker.fake_with_rng(&mut rng),
         started_at_bitcoin_block_hash: fake::Faker.fake_with_rng(&mut rng),
-        started_at_bitcoin_block_height: fake::Faker.fake_with_rng::<u32, _>(&mut rng) as u64,
+        started_at_bitcoin_block_height: (fake::Faker.fake_with_rng::<u32, _>(&mut rng) as u64)
```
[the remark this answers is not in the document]
MCJOHN974: [8641af3](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/8641af3f4fb29cae00ebb5202d488e05cde5172b)

signer/src/storage/model.rs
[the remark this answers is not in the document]
MCJOHN974: [1d085ef](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/1d085effad4b5dfdcf4c4c0554d25fdba241c226)

signer/src/storage/model.rs
MCJOHN974: [4f2167b](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/4f2167bfe8773652452d5b37a9e27fe406b9e964)

signer/src/storage/model.rs
MCJOHN974: [5e704e6](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/5e704e6bc9abd974bac92c104945db7f487ac2f4)

signer/src/storage/model.rs
MCJOHN974: [e34aa64](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/e34aa640c7011ca3e285c4a5a27c8d053ea71e46)

signer/src/storage/postgres.rs
```diff
@@ -469,10 +469,17 @@ impl PgStore {
...
         .fetch_optional(&self.0)
         .await
         .map_err(Error::SqlxQuery)
+        .map(|height| {
+            height.map(|height| {
+                BitcoinBlockHeight::try_from(height).map_err(Error::ConversionDatabaseInt)
+            })
+        })
+        .map(|opt| opt.transpose())
+        .and_then(|inner| inner)
```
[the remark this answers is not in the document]
MCJOHN974: [e96ba34](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/e96ba34c1af2027a2ba78ccc88326bd20efdc909)

signer/src/storage/postgres.rs
```diff
@@ -653,7 +669,9 @@ impl PgStore {
         .bind(min_block_height_candidate)
         .fetch_optional(&self.0)
         .await
-        .map_err(Error::SqlxQuery)?;
+        .map_err(Error::SqlxQuery)?
+        .map(|height| BitcoinBlockHeight::try_from(height).map_err(Error::ConversionDatabaseInt))
```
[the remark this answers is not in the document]
MCJOHN974: [661692c](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/661692cbbee6aa147109ea61af2cd54846308246)

signer/src/storage/postgres.rs
```diff
@@ -3085,8 +3101,14 @@ impl super::DbWrite for PgStore {
         &self,
```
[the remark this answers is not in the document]
MCJOHN974: [b730bfc](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/b730bfcc134f596bc84717810aaae4ad868bbada)

signer/src/storage/sqlx.rs
```diff
@@ -199,6 +201,62 @@ impl sqlx::postgres::PgHasArrayType for PublicKeyXOnly {
     }
 }
 
+// For the [`StacksBlockHeight`]
+
+impl<'r> sqlx::Decode<'r, sqlx::Postgres> for StacksBlockHeight {
+    fn decode(value: sqlx::postgres::PgValueRef<'r>) -> Result<Self, BoxDynError> {
+        let height = <i64 as sqlx::Decode<sqlx::Postgres>>::decode(value)?;
+        Ok(StacksBlockHeight::try_from(height)?)
```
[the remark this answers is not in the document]
MCJOHN974: Actually it have different error types and rust compiler suggested this wrapping into Ok(..?)

signer/src/testing/storage/model.rs
```diff
@@ -625,7 +626,10 @@ mod tests {
             .unwrap()
         {
             // Check the stacks heights increment as expected
-            assert_eq!(current.block_height, walk.last().unwrap().block_height - 1);
+            assert_eq!(
+                current.block_height,
+                walk.last().unwrap().block_height - 1u64
```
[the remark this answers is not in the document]
MCJOHN974: [29d0dce](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/29d0dce5d8e7b8eb96478a091598d4bd05dff68c)

signer/src/testing/blocks.rs
```diff
@@ -78,12 +80,12 @@ impl BitcoinChain {
     }
 
     /// Gets the nth block in the chain, if it exists.
-    pub fn nth_block_checked(&self, height: usize) -> Option<&BitcoinBlock> {
-        self.0.get(height)
+    pub fn nth_block_checked(&self, height: BitcoinBlockHeight) -> Option<&BitcoinBlock> {
```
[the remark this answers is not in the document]
MCJOHN974: Hmm, I'd say this wrappers at all don't bring much _usually_. But sometimes they can really help and even here probably they can help once...

signer/src/testing/request_decider.rs
```diff
@@ -231,7 +231,7 @@ where
     /// Assert that the transaction signer will make and store decisions
     /// for pending withdraw requests.
     pub async fn assert_should_store_decisions_for_pending_withdrawal_requests(self) {
-        let mut rng = rand::rngs::StdRng::seed_from_u64(46);
+        let mut rng = rand::rngs::StdRng::seed_from_u64(44);
```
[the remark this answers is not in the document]
MCJOHN974: It seems this test wasn't fixed in another pr so I'll just adjust this seed once will be merging upstram =(

signer/src/testing/transaction_coordinator.rs
```diff
@@ -803,7 +803,7 @@ where
                     Value::Sequence(SequenceData::Buffer(BuffData {
                         data: withdrawal_req.sweep_block_hash.to_le_bytes().to_vec()
                     })),
-                    Value::UInt(withdrawal_req.sweep_block_height as u128),
+                    Value::UInt(*withdrawal_req.sweep_block_height as u128),
```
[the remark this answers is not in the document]
MCJOHN974: @matteojug do you think we want this `impl` ?

signer/src/block_observer.rs
MCJOHN974: [0808b24](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/0808b247642e64f0e2a191460ed1ad94ecfde6bd)

signer/src/transaction_signer.rs
```diff
@@ -1699,12 +1699,16 @@ mod tests {
         chain_tip_height: u64,
         should_allow: bool,
     ) {
+        let chain_tip_height = chain_tip_height.into();
+        let dkg_min_bitcoin_block_height = match dkg_min_bitcoin_block_height {
```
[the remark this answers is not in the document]
MCJOHN974: [d209c12](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/d209c12d74b24588cf7ecb3ab60effe542d47930)

signer/src/transaction_coordinator.rs
```diff
@@ -2742,12 +2743,16 @@ mod tests {
         chain_tip_height: u64,
         should_allow: bool,
     ) {
+        let chain_tip_height = chain_tip_height.into();
+        let dkg_min_bitcoin_block_height = match dkg_min_bitcoin_block_height {
```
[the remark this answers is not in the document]
MCJOHN974: [d209c12](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/d209c12d74b24588cf7ecb3ab60effe542d47930)

signer/tests/integration/postgres.rs
```diff
@@ -1189,6 +1190,7 @@ async fn writing_withdrawal_accept_requests_postgres() {
     let event: WithdrawalAcceptEvent = fake::Faker.fake_with_rng(&mut rng);
 
     // Let's see if we can write these rows to the database.
+    eprintln!("{:#?}", event.clone());
```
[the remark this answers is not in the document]
MCJOHN974: [e975866](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/e97586609e347c43572f73c3270e0f582d3c79f0)

MCJOHN974:
@matteojug @cylewitruk thanks for all your comments!

I think I pushed changes which resolve _most_ of them, please close ones where you agree that it is closed.

It is still some major unadressed issues:

- which conversions we want to implement? It seems you two have different opinions and I'm ok with both
- usage of `_config` in `Dummy` implementation. I think, we actually don't need to care about it and just return random number between 0..u32max with uniform distribution (which next_u32 does) is fine for all our usecases, and I just need to implement this functions to satisfy traits. Am I wrong?

signer/src/storage/sqlx.rs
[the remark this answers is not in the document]
cylewitruk: Weird, thought that would work. In that case I don't really have a preference for which style you want to use :)

signer/src/testing/transaction_coordinator.rs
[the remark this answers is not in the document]
cylewitruk: Converting from a `*BlockHeight` (`u64`)  to `u128` will always succeed, so the difference is almost purely syntactic. But in this case you need to both deref (`*`) to `u64` and then cast (`as`) to `u128`, which isn't as clear as an explicit conversion, imo.

signer/src/testing/dummy.rs
```diff
@@ -510,6 +512,30 @@ impl fake::Dummy<fake::Faker> for SigHash {
...
+
+impl fake::Dummy<fake::Faker> for BitcoinBlockHeight {
+    fn dummy_with_rng<R: rand::Rng + ?Sized>(_config: &fake::Faker, rng: &mut R) -> Self {
+        rng.next_u32().into()
```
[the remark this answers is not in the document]
MCJOHN974: Oh, yeas, I dumb, sorry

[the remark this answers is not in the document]
MCJOHN974: Btw do you @cylewitruk think we should somehow use this `_config`? I believe we can just ignore it since we never use some "configured" faker

signer/src/storage/model.rs
cylewitruk: The deref to me is natural as it's deref'ing to its inner type, whereas for `u128` it's an actual conversion. I have nothing against also including `From<> for u64`'s, though. My argument for the `u128`conversion, despite its narrow use, are down [here](https://github.com/stacks-sbtc/sbtc/pull/1448#discussion_r2037175489).

signer/src/testing/storage/model.rs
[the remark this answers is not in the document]
MCJOHN974: @matteojug wdyt? 

I prefer slightly prefer the form I initially write (`1u64`), mostly because it is 6 symbols smaller. But it is also not a strong opinion so I'm ok to change if majority will ask for it

signer/src/testing/dummy.rs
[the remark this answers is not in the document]
MCJOHN974: [b20069a](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/b20069a0b59d32acd4abdc5339cc8ec244aee4d5) and [62dac02](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/62dac029ff467df29f47f427e0124eb76e90521f) 

I think this two address this problem. They are a bit different from what you suggested, though

signer/src/testing/dummy.rs
MCJOHN974: [b20069a](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/b20069a0b59d32acd4abdc5339cc8ec244aee4d5) and [62dac02](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/62dac029ff467df29f47f427e0124eb76e90521f)

signer/src/testing/dummy.rs
```diff
@@ -510,6 +512,48 @@ impl fake::Dummy<fake::Faker> for SigHash {
...
 
+impl fake::Dummy<std::ops::Range<u64>> for BitcoinBlockHeight {
+    fn dummy_with_rng<R: rand::Rng + ?Sized>(config: &std::ops::Range<u64>, rng: &mut R) -> Self {
+        // Overwriting config because values greater then i64::MAX will cause db conversion errors.
+        let mut config = config.clone();
+        if config.end > i64::MAX as u64 || config.start >= config.end {
+            config.start = 0;
+            config.end = i64::MAX as u64;
+        }
+        rng.gen_range(config).into()
```
djordon, line 523: If `end > i64::MAX`, why should the start then be fixed to zero? Why not just do:
```rust
rng.gen_range(config).min(i64::MAX as u64).into()
```
Well, that panics if the range is empty but that is probably the correct thing to do. Same for `StacksBlockHeight`

signer/src/testing/dummy.rs
```diff
@@ -510,6 +512,48 @@ impl fake::Dummy<fake::Faker> for SigHash {
...
+    }
+}
+
+impl fake::Dummy<fake::Faker> for BitcoinBlockHeight {
+    fn dummy_with_rng<R: rand::Rng + ?Sized>(config: &fake::Faker, rng: &mut R) -> Self {
+        let x: u64 = config.fake_with_rng(rng);
+        let y: u64 = config.fake_with_rng(rng);
+        let range = std::cmp::min(x, y)..std::cmp::max(x, y);
+        range.fake_with_rng(rng)
```
djordon, line 544: We should probably simplify this to `rng.gen_range(0..i64::MAX as u64).into()`. Same for `StacksBlockHeight`.

signer/src/storage/postgres.rs
```diff
@@ -469,10 +469,13 @@ impl PgStore {
...
         .bind(chain_tip)
-        .bind(min_block_height)
+        .bind(i64::try_from(min_block_height).map_err(Error::ConversionDatabaseInt)?)
         .fetch_optional(&self.0)
         .await
-        .map_err(Error::SqlxQuery)
+        .map_err(Error::SqlxQuery)?
+        .map(BitcoinBlockHeight::try_from)
+        .transpose()
+        .map_err(Error::ConversionDatabaseInt)
```
djordon, line 478: This error variant is supposed to be for storing into the database. Maybe add a new one, or use `Error::TypeConversion`, when reading from the database since we'll get a better error message that way.

signer/src/storage/model.rs
```diff
@@ -536,9 +533,8 @@ pub struct EncryptedDkgShares {
     pub started_at_bitcoin_block_hash: BitcoinBlockHash,
     /// The block height of the chain tip of the canonical bitcoin blockchain
     /// when the DKG round associated with these shares started.
-    #[sqlx(try_from = "i64")]
     #[cfg_attr(feature = "testing", dummy(faker = "0..i64::MAX as u64"))]
-    pub started_at_bitcoin_block_height: u64,
+    pub started_at_bitcoin_block_height: BitcoinBlockHeight,
```
djordon, line 537: We should no longer need the `cfg_attr` here.

signer/src/storage/model.rs
```diff
@@ -1352,6 +1346,256 @@ pub struct WithdrawalRejectEvent {
...
+    Hash,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Serialize,
+    Deserialize,
+    sqlx::FromRow,
+)]
+pub struct BitcoinBlockHeight(u64);
```
djordon, line 1581: I don't think `sqlx::FromRow` is needed for these, but maybe I'm mistaken. Also, do we need `Deserialize` or `Default`?

djordon: And we probably want `#[serde(transparent)]` on these as well.

signer/src/storage/model.rs
```diff
@@ -1337,6 +1335,178 @@ pub struct WithdrawalRejectEvent {
     pub signer_bitmap: BitArray<[u8; 16]>,
 }
 
+/// Helper for [`implement_trait!`] macro. Implements arithmetic traits for integer wrapper types.
+#[macro_export]
+macro_rules! implement_trait {
```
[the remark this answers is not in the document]
djordon: Forgot to remove this after some of the more recent changes.

signer/src/storage/sqlx.rs
```diff
@@ -199,6 +201,76 @@ impl sqlx::postgres::PgHasArrayType for PublicKeyXOnly {
...
+impl sqlx::Type<sqlx::Postgres> for BitcoinBlockHeight {
+    fn type_info() -> sqlx::postgres::PgTypeInfo {
+        <i64 as sqlx::Type<sqlx::Postgres>>::type_info()
+    }
+}
+
+impl<'r> sqlx::Encode<'r, sqlx::Postgres> for BitcoinBlockHeight {
+    fn encode_by_ref(&self, buf: &mut PgArgumentBuffer) -> Result<IsNull, BoxDynError> {
+        eprintln!("Encoding BitcoinBlockHeight: {}", self);
+        eprintln!("res {:#?}", i64::try_from(*self));
```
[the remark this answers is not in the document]
djordon: Oh, I think these were here for debugging purposes and have since been removed.

signer/src/testing/dummy.rs
MCJOHN974: My logic was following:

- Let's fallback to uniform distribution 0..i64max in case we have invalid config
- Why don't return err/panic on bad config? Mostly because my impl of Dummy<Faker> will generate invalid config in 50% cases
- Why I'm ok with all of that??? 1) In most cases I want height to be just random number (and fallback to unform on 0..i64::max is perfect here). 2) If in some cases I want some non-default random then this range based dummy should work well for me. I don't see anything else I want from Dummy impl of heights

signer/src/testing/dummy.rs
MCJOHN974: This impl will not use `config`, and @cylewitruk and @matteojug highlighted that we want use this config... However, imo, we ok to just have uniform on 0..i64max for all random heights

signer/src/storage/postgres.rs
MCJOHN974: Sooo, for extracting from db we don't want to use this error?..

signer/src/storage/model.rs
MCJOHN974: Thanks!

signer/src/storage/postgres.rs
djordon: The error message talks about converting to a value when writing to the database. That's not really this case, so we should probably use something else.

signer/src/testing/dummy.rs
djordon: For this `impl` the config is the `fake::Faker` type, which is not really a config. It's a unit type, so it's okay to not use it in this cases.

signer/src/testing/dummy.rs
djordon: I see. Well, I don't think that the `Dummy<Faker>` impl should fail. And even for this one, we probably shouldn't take the `min` as I was suggesting.

signer/src/storage/postgres.rs
MCJOHN974: Maybe still use this type? Like this conversion is in some sense "while writing (or reading) to db". 

If the issue is that here we don't directly write but preparing to write (do post-read actions) I don't think it is really matter, logically it is same -- we got a number from db which conflicts with our type system (or we try to write some data to db which conflicts with db type system).

If the issue is that we want separate reads and writes then 1) From name of the error I don't think it is about only reads or only writes. 2) Maaaybe we can benefit from such separation but probably just nice error message is enough

signer/src/api/info.rs
matteojug: Resolving since even if the `into` would work, the explicit one does seem better being this a test

signer/src/storage/model.rs
matteojug: I'm happy with the comment specifying that there's nothing fishy going on, resolving this convo.

signer/src/testing/transaction_coordinator.rs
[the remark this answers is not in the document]
matteojug: Resolved the other convo, feel free to add conversions to your heart's content

signer/src/testing/dummy.rs
matteojug: Since this is a version using an explicit range, if the provided range is wrong it wouldn't be too unexpected to panic for invalid ones. But we could silently cap the range for ergonomics (given we use `Range<u64>`, so maybe we end up with some automatic `u64::MAX` that would be annoying to explicitly type).
But regardless of that choice, I also find it strange to fallback to `0..i64max` for *any* invalid config.

cylewitruk: Yeah `min(i64::max)` would be bad because you'll end up with a whole bunch of `i64::MAX`.

Why not just use the u64 range directly (for this explicit impl)? There's no invariant on the block height itself, it's postgres that forces i64 on us. This is how it worked before in the dummy attributes, and yes i've caused it to panic by forgetting to limit it to i64::max... but that was my fault 😊

I'm not sure there's any usages left after the attribute removals in models.rs anyway, so we may be able to just drop these impls too?

djordon: > I'm not sure there's any usages left after the attribute removals in models.rs anyway, so we may be able to just drop these impls too?

Good point, I think that we should drop these impls if we aren't using them right now. If we need them later for something specific then we can add them back.

signer/src/storage/postgres.rs
djordon: Well, the question that I'm asking is how tough would our lives be if we needed to debug some issue here. By reusing the same error here it's not clear whether the error happened when reading or writing to the database. And we'd want to know because they'd have different implications. That's why I think we want a different error variant here, just to make our lives easier. I suggested using the `Error::TypeConversion` variant but we probably don't want to use that one either. That variant is mainly used in cases where we think that the conversion will succeed but are not confident enough to use `expect()`.

signer/src/testing/storage/model.rs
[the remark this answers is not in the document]
MCJOHN974: Well, if nobody have strong preference here I will keep it as is and close this thread, feel free to reopen

signer/src/storage/model.rs
MCJOHN974: [fa39d1e](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/fa39d1e500d7df870afaa533ae8dee34cc19dbdf)

signer/src/storage/model.rs
MCJOHN974: Added transparent here: [2c70343](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/2c703435c93dba2f20d67bd79f290213b4a6b3c3)

MCJOHN974: About `Deserialize` and `Default` I'd say even if we don't need them now there is no problem to have them. They don't make code more difficult to understand since this are the most basic `derive's` rust have and this list of `derive's` will be still huge even if we remove them. Also they do not slow down code execution. The only drawback they can bring -- they proooobably will make binary slightly bigger in some builds (but I think compiler will eliminate dead code in release builds), but even for debug builds, binary size is not really important for us I think

MCJOHN974: About `FromRow`: removed it here: [f5b6030](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/f5b60304edb7a923b103eadb35b0648b802cde71). Compared to `Deserialize` and `Default` it is much less obvious what this `derive` does imo, so it make sense to remove it

signer/src/storage/postgres.rs
MCJOHN974: Oh, I see, this makes sense. However, we already use `Error::ConversionDatabaseInt` for db writes for example here: https://github.com/stacks-sbtc/sbtc/blob/1719efefef588167398ac098f42dfd22d465da20/signer/src/storage/postgres.rs#L2697

But I agree with your suggestion that it can be useful to have different error types for read and write. And, I'd suggest to refactor this errors a bit, using some very explicit error names: `ConverstionDatabaseIntRead` and `ConversionDatabaseIntWrite`, but do it in a separate PR.

This will give us more explicit error naming and will not bring false feeling "`ConverstionDatabaseInt` occurs only on readings".

Do you think it can be a better choice to open an issue for such refactoring and do it in separate PR?

signer/src/testing/dummy.rs
MCJOHN974: Yeah, agree, I like decision to remove range-config impl and just do range(0..i64max) on dummy config impl. Implemented here: [8b76743](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/8b76743e5e1b1b3a0dd12c2cc8d327cd815fe74e)

signer/src/testing/dummy.rs
MCJOHN974: I think this [8b76743](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/8b76743e5e1b1b3a0dd12c2cc8d327cd815fe74e) resolves this thread

signer/src/testing/transaction_coordinator.rs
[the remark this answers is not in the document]
MCJOHN974: Whooops, we actually already had this impl, and it seems I used deref here kind of for consistency with the rest of the file. Removed deref: [6990de1](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/6990de1b7aea776a3ee7cfe2a4ab59f82b065eb2)

signer/src/storage/postgres.rs
```diff
@@ -1742,7 +1759,7 @@ impl super::DbRead for PgStore {
         .bind(chain_tip.block_hash)
         .bind(i32::from(context_window))
         .bind(stacks_chain_tip.block_hash)
-        .bind(i64::try_from(expiration_height).map_err(Error::ConversionDatabaseInt)?)
```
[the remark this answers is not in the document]
MCJOHN974: Whoops, we already had this impl, removed unnecessary derefs: [4f5a8b1](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/4f5a8b1acc160d2ceca8d38c84f3d0591d31f9c6)

signer/tests/integration/transaction_signer.rs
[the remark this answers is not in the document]
MCJOHN974: Yep, this `as u64` is not needed. Removed: [43a847a](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/43a847a9ab9f10fb152df7f7fab09e640b21bed2)

MCJOHN974:
Cool! Now I think we have just a few threads that need some attention:
- Error type for dbWrites: https://github.com/stacks-sbtc/sbtc/pull/1448#discussion_r2038748484
- Questionable usage of wrappers in some cases: https://github.com/stacks-sbtc/sbtc/pull/1448#discussion_r2035437732

If I'm not missing anything all other threads could be just closed.

MCJOHN974:
> Cool! Now I think we have just a few threads that need some attention:
> 
> * Error type for dbWrites: [[chore]: use wrapper structs for heights #1448 (comment)](https://github.com/stacks-sbtc/sbtc/pull/1448#discussion_r2038748484)
> * Questionable usage of wrappers in some cases: [[chore]: use wrapper structs for heights #1448 (comment)](https://github.com/stacks-sbtc/sbtc/pull/1448#discussion_r2035437732)
> 
> If I'm not missing anything all other threads could be just closed.

cc @djordon @cylewitruk @matteojug

signer/src/storage/postgres.rs
djordon: > However, we already use `Error::ConversionDatabaseInt` for db writes for example here:

Yeah, it's supposed to be used only when converting to signed integers before writing to the database. Typically, we don't need one for reading because sqlx will fail when converting to the desired struct and give us an error there. Which reminds me that there may be is a better way to handle this.

signer/src/storage/postgres.rs
```diff
@@ -450,8 +450,8 @@ impl PgStore {
     async fn get_least_txo_height(
         &self,
         chain_tip: &model::BitcoinBlockHash,
-        min_block_height: i64,
-    ) -> Result<Option<i64>, Error> {
+        min_block_height: BitcoinBlockHeight,
+    ) -> Result<Option<BitcoinBlockHeight>, Error> {
         sqlx::query_scalar::<_, i64>(
```
djordon, line 455: I think we can do this:
```suggestion
        sqlx::query_scalar::<_, BitcoinBlockHeight>(
```
Then there is no need to do the conversion afterwards.

signer/src/storage/model.rs
djordon: I usually just derive what is needed, because I cannot think of a good reason to derive something if it isn't needed. Same thing for code that is unused. Well, at least that's my rule of thumb. You can probably find some derived impls that are unused or were never used. We should probably remove them.

For `Default`, I don't think that we should have one, except maybe in tests. My thinking is that it can be a footgun for production code. For `Deserialize`, we don't implement it for most types in this crate, including the majority in this module (and I wonder if the ones in this module can be removed). It's small code for each one but it would add up to a lot if we did it for all of our types. Also, if it's unused, why add it.

MCJOHN974: Whoops, actually, `Deserialize` is required by `SignerConfig` and default by 
```
impl<const N: usize> ReorgDescription<N> {
    fn num_blocks(&self) -> u64 {
        *self.sweep_heights.into_iter().max().unwrap_or_default()
    }
}
```

signer/src/storage/postgres.rs
MCJOHN974: So you are ok to keep it as is for now?

signer/src/storage/postgres.rs
MCJOHN974: Whow nice catch! [776797d](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/776797daa99c421255b29467ac9e57e8322bff18)

signer/src/storage/model.rs
djordon: Fair enough!

signer/src/storage/postgres.rs
```diff
@@ -584,8 +584,8 @@ impl PgStore {
 
         // Given the utxo candidate above, this is our best guess of the
         // minimum UTXO height. It might be wrong, we'll find out shortly.
-        let min_block_height_candidate = utxo_candidate
-            .block_height
+        let min_block_height_candidate = BitcoinBlockHeight::try_from(utxo_candidate.block_height)
+            .map_err(Error::ConversionDatabaseInt)?
```
djordon, line 588: You can change the type of `PgCandidateUtxo.block_height` to be a bitcoin block height. No need to make the change for this PR though.

djordon (commented):
Looks good. Just holding off on approval until we get the next release cut.

signer/src/storage/postgres.rs
djordon: Oh wait, we aren't waiting on me, so I suppose there is no harm in making the change here, and it's preferable to do so now.

signer/src/storage/postgres.rs
djordon: With https://github.com/stacks-sbtc/sbtc/pull/1448/commits/776797daa99c421255b29467ac9e57e8322bff18 I think we're good.

signer/src/storage/postgres.rs
MCJOHN974: [f5efcf6](https://github.com/stacks-sbtc/sbtc/pull/1448/commits/f5efcf6eaa58292ebcec1ccd753783908e362ca6)

MCJOHN974:
@djordon I think I addressed your last comment, we did release, aaand I even merged upstream. Maybe good time to approve it?)

djordon (commented):
Looks good, just holding off on an approval until we cut the release.

djordon (approved):

signer/src/storage/model.rs
cylewitruk: Closing this as we removed the macros

cylewitruk (approved):
LGTM!

MCJOHN974:
Closing this since https://github.com/stacks-sbtc/sbtc/pull/1614 was merged (it is just a wrapper around this PR to merge unsigned commits)

PR #1449, opened 2025-02-28
[chore] rebuild protos

MCJOHN974:
## Description

Closes: No issue.

## Changes

Now our build will rebuild protobufs with `REBUILD_PROTOS` env setted.

## Testing Information

I just tried it few times on my machine

## Checklist:


cylewitruk:
Not sure with others think, but one suggestion could be to add a `bin` to `Cargo.toml` for the build script specifically. I haven't tried this, but I've seen it somewhere:

```
[[bin]]
name = "signer-build-script"
path = "build.rs"
```

And update the `Makefile` with a new rule:
```diff
# ##############################################################################
# MAIN TARGETS
# ##############################################################################
...
contracts:
	pnpm --prefix contracts run build

+protos:
+       REBUILD_PROTOS=true cargo $(CARGO_FLAGS) run --bin signer-build-script
+.PHONY: install build test test-build lint format contracts clean protos
-.PHONY: install build test test-build lint format contracts clean
...
```
```

**Comment by cylewitruk** [comment]
@MCJOHN974 looks like you'll need to make sure protoc is installed for the relevant workflow jobs (eg `Build Test Artifacts` and anything else that wants to build `signer`):
```
thread 'main' panicked at signer/build.rs:66:10:
  Unable to compile protocol buffers: Custom { kind: NotFound, error: "Could not find `protoc`.
```

**Comment by djordon** [comment]
Oh right, we removed it from CI because the action was quite annoying and would fail randomly sometimes with some rate limit error. The action was this one: https://github.com/marketplace/actions/setup-protoc. If we're going to do this, let's not use that one.

File: .github/workflows/on-push.yaml [hunk 1]
```diff
@@ -45,6 +45,7 @@ jobs:
         with:
           components: clippy, rustfmt
           cache-key: "rust-tests"
+      - uses: arduino/setup-protoc@v3
```
cylewitruk, line 48: Let's make sure we use the commit hash instead of `@v3` here and for the others

cylewitruk (approved):
Lgtm!

cylewitruk:
@MCJOHN974 could we try using `apt install -y protobuf-compiler` instead of the github action? Apparently it is documented in the current action that there can be issues with rate-limiting, and this would avoid that. But I'm not sure which versions are available via apt...

cylewitruk (dismissed):

matteojug (commented):
Should we pin `protobuf-compiler` version? We don't for most of the apt installs around (eg, in the dockerfiles), but maybe in CI it would be better to do it

cylewitruk:
@MCJOHN974 do you know what `protoc` version this actually installs? The APT versioning doesn't seem to match the "normal" versioning of the GH releases. When I `protoc --version` I get:
```
% protoc --version
libprotoc 29.3
```

PR #1471, opened 2025-03-04
[fix] sidecar merge 

Jiloc:
## Description

- When merging #1409 , the base branch was changed from `feat/emily-new-block-handler` to `main`, and this caused the change that was removing the `boxed` to be lost. 

- Upgrade the sidecar integration tests to the latest changes of the `new-block-handler`. 
    - It now returns the hex encoded script pub key and not the bitcoin address.
    - The fulfillment of the confirmed withdrawal is now using the `bitcoin_txid` of the event. In the real events, the parameter is the same as `sweep-txid`, but the fixture was actually using two different values. 
             
## Checklist:


djordon (approved):
Looks good ✅

matteojug (approved):

PR #1473, opened 2025-03-05
chore: remove signer's emily-client implementations to call private endpoints

Jiloc:
## Description

Remove the implementation of two private endpoints from the signer's emily-client. Those endpoints will be remove from the Public API and the signers will not have access to the autogenerated client code that are currently importing.

## Changes
- remove `create_withdrawals` - The withdrawals will be created from the `new-block`  calls from the private API.
- remove `set_chainstate` - Only the sidecar can set a chainstate from the private API.

## Testing Information
- remove mocking of the endpoints above in a few tests.

## Checklist:


djordon (approved):
Looks good ✅

PR #1474, opened 2025-03-05
chore: emily remove private endpoints

Jiloc:
## Description

Remove the declaration of a couple of endpoints that are not meant to be called by the signers from the public API.

## Changes
- remove `create_withdrawal` from the public API. The withdrawals are expected to be created from the new-block endpoint. We keep it in the **private** API in order to have a way to recreate missing withdrawals in an easy manner in case of a problem.

- remove `set_chainstate` from the public API. Only the sidecar is supposed the endpoint throught the private API.

## Checklist:


djordon (approved):
Looks good ✅

Issue #1310, opened 2025-02-06
[Feature]: Have our production Docker builds verify their attestation at runtime, before running


aldur:
`gh attestation verify` can be skipped accidentally or because of sloppiness. 

Can the environment by default execute this check and only run the code if it passes? 

We should probably provide signers an opt-out flag for example for testing or other purposes.

PR #1476, opened 2025-03-05
feat: Add attestation entry point to Docker Images

fabergat:
## Description

This PR adds the support to the attestation verification in offline mode and give a suggestion on how to override the Docker entry point for Docker and Docker Compose.

Closes: https://github.com/stacks-network/sbtc/issues/1310

Note: 

_The key material in trusted_root.jsonl does not have a built-in expiration date, so anything signed before you generate the trusted root file will continue to successfully verify. Anything signed after the file is generated will verify until that Sigstore instance rotates its key material, which typically happens a few times per year. You will not know if key material has been revoked since you last generated the trusted root file._

Open questions:

- Is this thing adding a lot more security? 
- Is it worth it complicate the config?

## Testing Information

This is a build workflow that generates the attestation files: https://github.com/stacks-network/sbtc/actions/runs/13830677511/job/38694062940

## Checklist:


docker/mainnet/gh-attestation/entry_point.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
fdefelici, line 1: Rename this file `entry_point.sh` in order to match the `entrypoint.sh` file referred in the README.md

docker/mainnet/gh-attestation/README.md
```diff
@@ -0,0 +1,106 @@
...
+    volumes:
+      - /path/to/your/bundle.jsonl:/path/to/your/bundle.jsonl
+      - /path/to/your/trusted_root.jsonl:/path/to/your/trusted_root.jsonl
+    entrypoint: ["/entrypoint.sh"]
+    command: ["/usr/local/bin/blocklist-client"]
+```
+
+This will:
+1. **Set up the Docker container** with the required environment variables for attestation.
+2. **Use `/entrypoint.sh`**: The entry point script checks the attestation and proceeds if verified.
```
fdefelici, line 94: Add hyperlink to the entry point file

docker/mainnet/gh-attestation/README.md
```diff
@@ -0,0 +1,106 @@
...
+  /usr/local/bin/blocklist-client
+```
+
+This command will:
+1. **Set the environment variables**:
+    - `TAG="signer"`: This sets the image tag used to identify the Docker image for verification.
+    - `BUNDLE_PATH="/path/to/your/bundle.jsonl"`: Specifies the local path to the attestation bundle file.
+    - `TRUSTED_ROOT_PATH="/path/to/your/trusted_root.jsonl"`: Specifies the local path to the trusted root file for the attestation.
+  
+2. **Use `/entrypoint.sh`**: The entrypoint of the Docker image is overridden to run the `entrypoint.sh` script, which performs the attestation verification before running the application.
```
fdefelici, line 54: Add hyperlink to the entry point file

docker/mainnet/gh-attestation/README.md
```diff
@@ -0,0 +1,106 @@
...
+  -e BUNDLE_PATH="/path/to/your/bundle.jsonl" \
+  -e TRUSTED_ROOT_PATH="/path/to/your/trusted_root.jsonl" \
+  -v /path/to/your/bundle.jsonl:/path/to/your/bundle.jsonl \
+  -v /path/to/your/trusted_root.jsonl:/path/to/your/trusted_root.jsonl \
+  --entrypoint /entrypoint.sh \
+  image_name \
+  /usr/local/bin/blocklist-client
+```
+
+This command will:
```
fdefelici, line 48: Considering the section "Enviroment Variables",  The description could be simplified in something like this:

```
The commands above:
1. **Set the environment variables**: `TAG`, `BUNDLE_PATH` and `TRUSTED_ROOT_PATH`
2. .... 
```

By this the message is still the same, but get rid off duplication respect of the "Environment Variables" section

fdefelici (changes requested):
Added few remarks.

I need also a clarification, to check if my undestanding is right about this need.
We are offering this attestation verification and related samples, to allow a user to create its own docker image with blockstack-cli and signer binary?

docker/mainnet/gh-attestation/entry_point.sh
fabergat: done

docker/mainnet/gh-attestation/README.md
fabergat: done

docker/mainnet/gh-attestation/README.md
fabergat: done

aldur (commented):
Can you also modify the files under https://github.com/stacks-network/sbtc/blob/main/docker/mainnet/docker-compose.yml and the `testnet` corresponding version?

fdefelici (approved):

fabergat:
> https://github.com/stacks-network/sbtc/blob/main/docker/mainnet/docker-compose.yml

done

docker/mainnet/docker-compose.yml
```diff
@@ -26,6 +26,12 @@ x-common-vars:
   - &STREAM_NAME ${STREAM_NAME} # Required
   - &AWS_ACCESS_KEY_ID ${AWS_ACCESS_KEY_ID} # Required
   - &AWS_SECRET_ACCESS_KEY ${AWS_SECRET_ACCESS_KEY} # Required
+  # Attestation
+  - &SIGNER_TAG ${SIGNER_TAG} # Required
```
fdefelici, line 30: We could express the concept of mandatory variables by using "Required value" interpolation, and specifically the one that check if a VAR is set and not empty: `:?[OPTIONAL MESSAGE]`

This way, in case the variable is missing, the docker process is stopped exiting with error (otherwise the default behavior is to set the variable to an empty string and log just a warning message)

Example:
```
&MY_VAR ${MY_VAR:?}

&MY_VAR ${MY_VAR:?My Error message}
```

it's probably more than enough without the optional message.

fdefelici (changes requested):

docker/mainnet/docker-compose.yml
fabergat: actually it is a good idea! added

fdefelici (approved):

.github/workflows/image-build-and-draft-release.yaml
```diff
@@ -108,6 +108,29 @@ jobs:
           subject-name: index.docker.io/${{ env.docker-org }}/sbtc
           subject-digest: ${{ steps.docker_build.outputs.digest }}
           push-to-registry: true
+
+      - name: Download artifact attestation 
+        run: |
+          gh attestation download oci://index.docker.io/${{ env.docker-org }}/sbtc:${{ matrix.docker_target }}-${{ github.ref_name }} -R stacks-network/sbtc
```
matteojug, line 114: Shouldn't this include also `-${{ matrix.dist }}`? Right now we only have debian, but if we add another one we will always use the debian attestation (IIUC)

docker/mainnet/docker-compose.yml
```diff
@@ -75,9 +82,15 @@ services:
...
       POSTGRES_HOST: *POSTGRES_HOST
       POSTGRES_PORT: *POSTGRES_PORT
+      TAG: *SIGNER_TAG  
+      BUNDLE_PATH: *SIGNER_BUNDLE_PATH
+      TRUSTED_ROOT_PATH: *TRUSTED_ROOT_PATH
     volumes:
       - ./sbtc-signer/signer-config.toml.in:/signer-config.toml.in:ro
       - ./logs:/logs
+      - ${SIGNER_BUNDLE_PATH}:${SIGNER_BUNDLE_PATH}:ro
+      - ${TRUSTED_ROOT_PATH}:${TRUSTED_ROOT_PATH}:ro
```
matteojug, line 92: If we use `SIGNER_BUNDLE_PATH`/`TRUSTED_ROOT_PATH` also on the container side we may end up with funny paths (or maybe not working ones on windows?), why don't we use some constant paths for the mounted volumes instead?

fabergat: in fact I was intending as "file name" but you are right, I will fix the path from the docker side

fabergat: done

.github/workflows/image-build-and-draft-release.yaml
fabergat: let me also add the dist

fabergat: done

docker/mainnet/docker-compose.yml
matteojug: this also applies to `docker/testnet/docker-compose.testnet.yml` I think

fabergat: yes I am just stupid, done!

docker/mainnet/gh-attestation/README.md
```diff
@@ -0,0 +1,120 @@
...
+  -v /path/to/your/bundle.jsonl:/path/to/your/bundle.jsonl \
+  -v /path/to/your/trusted_root.jsonl:/path/to/your/trusted_root.jsonl \
+  --entrypoint /entrypoint.sh \
+  image_name \
+  /usr/local/bin/signer --config /signer-config.toml --migrate-db
+```
+
+```bash
+docker run --rm \
+  -e TAG="blocklist-cli" \
```
matteojug, line 38: nit: `blocklist-client`? multiple occurrences

docker/mainnet/gh-attestation/README.md
```diff
@@ -0,0 +1,120 @@
...
+  image_name \
+  /usr/local/bin/blocklist-client
+```
+
+This command will:
+1. **Set the environment variables**: `TAG`, `BUNDLE_PATH` and `TRUSTED_ROOT_PATH`
+  
+2. **Use [/entrypoint.sh](/docker/mainnet/gh-attestation/entrypoint.sh)**: The entrypoint of the Docker image is overridden to run the `entrypoint.sh` script, which performs the attestation verification before running the application.
+   
+3. **Run the Signer Application**: The signer application is started with the provided configuration file (`/signer-config.toml`) and the database will be migrated using the `--migrate-db` flag.
```
matteojug, line 53: nit: we list both the signer and blocklist client commands, while this only applies to the signer binary (first command)

docker/mainnet/gh-attestation/entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+# Define the image and repo (since they are fixed)
+IMAGE="index.docker.io/blockstack/sbtc:$TAG"  # You can pass $TAG as environment variable
+REPO="stacks-network/sbtc"
+
+# Verifying attestation
+echo "✅ Verifying attestation for image: $IMAGE..."
+apt-get update
+apt-get install -y curl
```
matteojug, line 17: Should we pin `curl`? Although I got bit by pinning it (and distro removing the package) in the past. Or move the install to the image so that it's baked into it

docker/mainnet/gh-attestation/entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
+&& chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
+&& echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
+&& apt update \
+&& apt install gh -y
```
matteojug, line 22: As above: we are not pinning anything (and running this at every restart), are we sure about it?

docker/mainnet/gh-attestation/entrypoint.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -e  # Exit on error
+
+# Ensure required environment variables are set
+if [[ -z "$BUNDLE_PATH" || -z "$TRUSTED_ROOT_PATH" ]]; then
```
matteojug, line 5: nit: is `$TAG` also required?

docker/mainnet/gh-attestation/entrypoint.sh
```diff
@@ -0,0 +1,30 @@
...
+IMAGE="index.docker.io/blockstack/sbtc:$TAG"  # You can pass $TAG as environment variable
+REPO="stacks-network/sbtc"
+
+# Verifying attestation
+echo "✅ Verifying attestation for image: $IMAGE..."
+curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
+&& chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
+&& echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
+&& apt update \
+&& apt install gh=2.67.0 -y
```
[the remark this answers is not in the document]
fabergat: It was there but before was only apt install apt install gh but because I needed the latest version I had to install the key and so on so I thought was better to do it outside without touching the original image and takes no more than 2 seconds. If we think it is ok and safe I can move it back

.github/workflows/image-build-and-draft-release.yaml
```diff
@@ -108,7 +108,30 @@ jobs:
...
+          # Generate trusted root
+          gh attestation trusted-root > trusted_root.jsonl
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload Attestation Files as Artifacts
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 #v4.6.1
+        with:
+          name: attestation-files
+          overwrite: true
```
djordon, line 130: How does attestation work for old releases? Does this `overwrite: true` mean that we cannot attest for images of prior releases?

docker/mainnet/gh-attestation/README.md
```diff
@@ -0,0 +1,120 @@
...
+To start the service with Docker Compose, use:
+
+```bash
+docker-compose up
+```
+---
+
+## **Additional Information**
+
+The entrypoint install the last version the GitHub CLI (`gh`) to verify the attestation.
```
djordon, line 105: This is no longer true.

docker/mainnet/gh-attestation/entrypoint.sh
djordon, line 1: Is there an advantage of having this file outside of the image itself. As in, would it be bad for some reason -- bad for security of whatever -- to have this included in our images? If this file were included, it could simplify running our images. I'm not so sure that we want to change the default docker entrypoint to go through this file, I'm just suggesting that it already be included in the image.

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
```diff
@@ -16,6 +17,12 @@ RUN apt-get install -y --no-install-recommends \
     g++ && \
         apt-get clean && rm -rf /var/lib/apt/lists/*
 
+RUN curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg && \
+    chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg && \
+    echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null && \
+    apt-get update && \
+    apt-get install -y gh=2.67.0
```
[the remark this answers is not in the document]
fabergat: done

.github/workflows/image-build-and-draft-release.yaml
fabergat: No no the true values it means if we have a file with the same name will be overwritten, in this case is valid only for the trusted root, so we have always the last version. This is also valid for the file export the job is doing not the attestation in docker hub

docker/mainnet/gh-attestation/README.md
fabergat: Removed

docker/mainnet/gh-attestation/entrypoint.sh
fabergat: Nothing bad, I was thinking can give more freedom to the partners if they wanna run it via this script or plain

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
```diff
@@ -7,6 +7,7 @@ RUN apt-get update
 RUN apt-get install -y --no-install-recommends \
     libclang-dev \
     git \
+    curl \
```
matteojug, line 10: Is `curl` required (in the builder) or a leftover?

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
```diff
@@ -31,6 +32,13 @@ COPY --from=build /code/sbtc/target/release/blocklist-client /usr/local/bin/bloc
 # ca-certificates is required to establish TLS connections.
 # gettext provides envsubst
```
matteojug, line 33: nit: those comments probably shouldn't be separate from the apt install they refer to

docker/mainnet/gh-attestation/entrypoint.sh
fabergat: Yeah I mean as you said the best effort

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
```diff
@@ -31,6 +32,13 @@ COPY --from=build /code/sbtc/target/release/blocklist-client /usr/local/bin/bloc
...
 
-RUN apt-get update && apt-get install -y ca-certificates gettext --no-install-recommends && \
-        apt-get clean && rm -rf /var/lib/apt/lists/*
+RUN curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg && \
```
matteojug, line 35: It seems curl is not included by default in `debian:bookworm-slim`

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
fabergat: reworked

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
fabergat: reworked

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
fabergat: reworked

matteojug (approved):

djordon (approved):

PR #1479, opened 2025-03-05
[feature] `get_deposits` returns accepted and pending

Jiloc:
## Description

Closes: #1463

## Changes
- renamed the previous `get_deposits` -> `get_deposits_with_status`. 
- add a new `get_deposits` implementation that returns both `Pending` and `Accepted` deposits. If only one of the two calls fails, we log an error but still return the deposits fetched by the other call.

## Testing Information
- added an integration test

## Checklist:

### DIFF
--- a/signer/src/emily_client.rs
+++ b/signer/src/emily_client.rs
@@ -89,9 +89,15 @@ pub trait EmilyInteract: Sync + Send {
         output_index: u32,
     ) -> impl std::future::Future<Output = Result<Option<CreateDepositRequest>, Error>> + Send;
 
-    /// Get pending deposits from Emily.
+    /// Get pending and accepted deposits to process from Emily.
     fn get_deposits(
         &self,
+    ) -> impl std::future::Future<Output = Result<Vec<CreateDepositRequest>, Error>> + Send;
+
+    /// Get pending deposits from Emily.
+    fn get_deposits_with_status(
+        &self,
+        status: Status,
     ) -> impl std::future::Future<Output = Result<Vec<CreateDepositRequest>, Error>> + Send;
 
     /// Update accepted deposits after their sweep bitcoin transaction has been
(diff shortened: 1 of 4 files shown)


signer/src/emily_client.rs
```diff
@@ -89,11 +89,17 @@ pub trait EmilyInteract: Sync + Send {
         output_index: u32,
     ) -> impl std::future::Future<Output = Result<Option<CreateDepositRequest>, Error>> + Send;
 
-    /// Get pending deposits from Emily.
+    /// Get pending and accepted deposits to process from Emily.
     fn get_deposits(
         &self,
     ) -> impl std::future::Future<Output = Result<Vec<CreateDepositRequest>, Error>> + Send;
 
+    /// Get pending deposits from Emily.
```
matteojug, line 97: nit: outdated comment

signer/tests/integration/emily.rs
```diff
@@ -670,6 +673,84 @@ async fn test_get_deposits_request_paging(
...
+    deposit_api::update_deposits(
+        emily_client.config(),
+        UpdateDepositsRequestBody { deposits: deposits },
+    )
+    .await
+    .expect("cannot update deposits");
+
+    // Check that we get all deposits
+    let deposits = emily_client.get_deposits().await.unwrap();
+    assert_eq!(deposits.len(), num_deposits);
```
matteojug, line 755: nit: can we also check that we got `num_accepted` deposits with an `Accepted` state?

matteojug (approved):
Do we need to halve/change `pagination_timeout` default now that potentially do twice the amount of fetching?

signer/tests/integration/emily.rs
Jiloc: added in https://github.com/stacks-network/sbtc/pull/1479/commits/9e22935eb9d88408fbacc6e39d34b07870179d7b

PR #1482, opened 2025-03-05
chore: add comments for `get_swept_*` queries near context boundary

matteojug:
## Description

Closes: #?

## Changes

The `get_swept_deposit_requests` and `get_swept_withdrawal_requests` queries may return completed requests as swept and pending if the relative completed events are anchored to a bitcoin block that precede the sweep bitcoin block.
This is fine as we have other layers taking care of this:
 - The signers validating the contract call will ask the stacks node it the request is still pending
 - The contracts rejecting replay

This PR add comments about it.

## Testing Information

Add a test to show this situation.

## Checklist:


djordon (approved):
Looks good 🚀

One caveat is that the coordinator does not currently validate the contract call before asking others to sign it. I just opened a ticket for it in https://github.com/stacks-network/sbtc/issues/1483.

Issue #1452, opened 2025-03-03
[Bug]: Rebroadcast signer withdrawal votes


djordon:
# Bug - Rebroadcast signer request votes

## 1. Description

The underlying issue was described in https://github.com/stacks-network/sbtc/issues/1237. This ticket is to implement the same solution for withdrawals.

### 1.1 Context & Purpose

After a signer has processed a bitcoin block and gotten their Stacks blockchain data up-to-date with their Stacks node, they begin to vote on outstanding withdrawal and deposit requests and broadcast their vote to their peers. When one signer doesn't know about the withdrawal request when they receive the vote, that vote gets dropped. For deposits we try to rebroadcast the vote for some number of blocks, and we should do the same for withdrawals.

## 2. Technical Details:

We should adopt the approach implemented in https://github.com/stacks-network/sbtc/pull/1243 for withdrawals.

### 2.1 Acceptance Criteria:
   - [ ] Each signer rebroadcasts withdrawal votes for some number of bitcoin blocks after they vote themselves.

PR #1485, opened 2025-03-07
[feat] rebroadcast withdrawals

MCJOHN974:
## Description

Closes: #1452

## Changes

- add new withdrawal_decisions_retry_window param to the signers' settings (defaults to 3)
- add query to get the all the WithdrawalSigner entries from the DB from the last N blocks
- add resend mechanism in the request_decider to resend all the Requests of the last 3 blocks when a new bitcoin block arrives

## Testing Information

- integration test for query

## Checklist:


signer/src/storage/postgres.rs
```diff
@@ -2527,6 +2527,41 @@ impl super::DbRead for PgStore {
...
+                JOIN sbtc_signer.bitcoin_blocks blocks USING (block_hash)
+                ORDER BY chain.block_height ASC
+                LIMIT 1
+            )
+            SELECT
+                ws.request_id,
+                ws.txid,
+                ws.block_hash,
+                ws.signer_pub_key,
+                ws.is_accepted
```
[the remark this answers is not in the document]
djordon: Oh the query for `get_deposit_signer_decisions` has the same style of using trailing commas. Whoops, that one slipped through. We should probably update that one as well.

djordon (commented):
Largely looks good. One requested change that is semi-related to this PR, and another style nit or two.

djordon (approved):
Looks good. Just make sure to add that comment to the in memory implementation of the query.

signer/src/request_decider.rs
```diff
@@ -277,6 +295,24 @@ where
         Ok(())
     }
 
+    /// Send the given deposit decisions to the other signers for redundancy.
```
matteojug, line 298: nit: 
```suggestion
    /// Send the given withdrawal decisions to the other signers for redundancy.
```

signer/src/storage/mod.rs
```diff
@@ -132,6 +132,15 @@ pub trait DbRead {
         signer_public_key: &PublicKey,
     ) -> impl Future<Output = Result<Vec<model::DepositSigner>, Error>> + Send;
 
+    /// Get all the withdrewal decisions for the given signer in the given window
```
matteojug, line 135:
```suggestion
    /// Get all the withdrawal decisions for the given signer in the given window
```

matteojug (approved):

signer/src/storage/in_memory.rs
```diff
@@ -927,6 +927,69 @@ impl super::DbRead for SharedStore {
             .map(|s| (s.will_sign, s.aggregate_key)))
     }
 
+    // This might look like it do not have same behaviour as the implementation from postgres.rs
+    // However, this is correct, e use the stacks block anchor block as a proxy for the timestamp of
+    // when the decision was made.
```
djordon, line 932: nit: I would use something like this
```suggestion
    // The postgres implementation uses a timestamp to figure out when a
    // decision was inserted into the database. The in memory database
    // does not have such a timestamp, so we use the stacks block anchor
    // block as a proxy for when the decision was made.
```

djordon (approved):

Issue #1483, opened 2025-03-05
[Feature]: Coordinator to validate contract calls before asking others to sign


djordon:
# Feature - Coordinator to validate contract calls before asking others to sign

## 1. Description

The coordinator does not validate the contract call before asking others to sign it. That's not nice, it should do it.

### 1.1 Context & Purpose

This will just cut down on logging noise and reduce network activity. All good things.

The thing that we need to be careful of is making sure that the signers don't bring in any "personal" feelings about a Stacks transaction. For example, how they voted on a request probably shouldn't matter on the Stacks side; but even that isn't obviously true.

## 2. Technical Details:

I think we just need to call `ContractCall::validate`?

### 2.1 Acceptance Criteria:
   - [ ] The coordinator validates all contract call transactions before asking others to sign it.

## 4. Appendix

The coordinator does validate the transaction when they receive the sign request internally. This ticket is for them to validate the transaction an additional time, once before deciding to ask others to sign the transaction.

PR #1488, opened 2025-03-07
feature: coordinator checks deposit status before mint signing round

matteojug:
## Description

Closes: https://github.com/stacks-network/sbtc/issues/1483

## Changes

Add a `is_deposit_completed` call in the coordinator before calling `construct_deposit_stacks_sign_request` (and starting the signing round). Also nit a couple of logs for consistency.

## Testing Information

Add a test to ensure the coordinator do/don't try a signing round based on the deposit status.

## Checklist:

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -613,7 +613,7 @@ where
         // If `get_pending_requests()` returns `Ok(None)` then there are no
         // eligible requests to service; we can exit early.
         let Some(pending_requests) = pending_requests_fut.await? else {
-            tracing::debug!("no bitcoin requests to handle");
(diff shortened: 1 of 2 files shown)


matteojug:
This is not validating the whole transaction but only checking for the obvious case that we found (a deposit already confirmed but lacking the event). Are there other cases that would make us want to validate the actual transaction (in the coordinator)?

djordon:
> This is not validating the whole transaction but only checking for the obvious case that we found (a deposit already confirmed but lacking the event). Are there other cases that would make us want to validate the actual transaction (in the coordinator)?

I don't think so. I think that this is the only case where an honest coordinator might make a mistake when compared to the validation that it does as a signer.

djordon (approved):
Looks good!

Issue #1489, opened 2025-03-07
[Feature]: Update Emily about accepted withdrawal requests


djordon:
# Feature - Update Emily about accepted withdrawal requests 

## 1. Description

After the coordinator has successfully submitted a bitcoin transaction fulfilling a withdrawal request, the coordinator should update Emily about the new state. We do this for deposits here:

https://github.com/stacks-network/sbtc/blob/ba744ebf893e97b137fa44f21f3443a03dc3e174/signer/src/transaction_coordinator.rs#L650-L653

And should do the same thing for withdrawals. Note that we bail on an error in the above code, when we shouldn't.

### 1.1 Context & Purpose

Updating Emily of the status leads to a better user experience. And Emily errors here are not bail worthy, so we shouldn't.

## 2. Technical Details:

We'll need to add a new method to the `EmilyInteract` trait for a `accept_withdrawals` method that takes the same inputs as the `accept_deposits` method and returns the analogous object for withdrawals. Also, when we get errors when updating Emily, we should just log them and carry on with life.

### 2.1 Acceptance Criteria:
   - [ ] Update Emily about accepted withdrawal requests
   - [ ] Don't bail when updating Emily

PR #1492, opened 2025-03-10
[feat]: Update Emily about accepted withdrawal requests

MCJOHN974:
## Description

WIP

Closes: #1489

## Checklist:

### DIFF
--- a/signer/tests/integration/transaction_coordinator.rs
+++ b/signer/tests/integration/transaction_coordinator.rs
@@ -3768,6 +3768,12 @@ async fn sign_bitcoin_transaction_withdrawals() {
 
     // Now we should manually put withdrawal request to Emily, pretending that
     // sidecar did it.
+    let stacks_tip_height = db
+        .get_stacks_block(&stacks_chain_tip)
+        .await
+        .unwrap()
+        .unwrap()
+        .block_height;
     let request_body = testing_emily_client::models::CreateWithdrawalRequestBody {
         amount: withdrawal_request.amount,
         parameters: Box::new(testing_emily_client::models::WithdrawalParameters {
@@ -3777,12 +3783,7 @@ async fn sign_bitcoin_transaction_withdrawals() {
         request_id: withdrawal_request.request_id,
         sender: withdrawal_request.sender_address.to_string(),
         stacks_block_hash: withdrawal_request.block_hash.to_string(),
-        stacks_block_height: db
-            .get_stacks_block(&stacks_chain_tip)
-            .await
-            .unwrap()
-            .unwrap()
-            .block_height,
+        stacks_block_height: stacks_tip_height,
     };
     let response =
         withdrawal_api::create_withdrawal(&emily_client.config().as_testing(), request_body).await;


signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3763,6 +3780,47 @@ async fn sign_bitcoin_transaction_withdrawals() {
...
+                request_id: withdrawal_request.request_id,
+                sender: withdrawal_request.sender_address.to_string(),
+                stacks_block_hash: withdrawal_request.block_hash.to_string(),
+                stacks_block_height: db
+                    .get_stacks_block(&stacks_chain_tip)
+                    .await
+                    .unwrap()
+                    .unwrap()
+                    .block_height,
+            }
```
matteojug, line 3804: nit: we can probably pass a const for `stacks_block_height` if we don't need the real height

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3763,6 +3780,47 @@ async fn sign_bitcoin_transaction_withdrawals() {
...
+            &test_emily_config,
+            testing_emily_client::models::Status::Accepted,
+            None,
+            None,
+        )
+        .await
+        .unwrap()
+        .withdrawals
+        .len(),
+        0
```
matteojug, line 3821: nit: `is_empty`?

matteojug (commented):
LGTM, just some nits

signer/tests/integration/transaction_coordinator.rs
matteojug: Makes sense. Can we move the `db.get_stacks_block...block_height` outside, for readability?

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3763,6 +3763,44 @@ async fn sign_bitcoin_transaction_withdrawals() {
...
         sender_address: PrincipalData::from(StandardPrincipalData::transient()).into(),
     };
+
+    // Now we should manually put withdrawal request to Emily, pretending that
+    // sidecar did it.
+    assert!(
+        testing_emily_client::apis::withdrawal_api::create_withdrawal(
+            &emily_client.config().as_testing(),
+            testing_emily_client::models::CreateWithdrawalRequestBody {
+                amount: withdrawal_request.amount,
```
djordon, line 3773: Can we break this all out of the assert for readability? I'm thinking something like
```rust
use testing_emily_client::apis::withdrawal_api;
use testing_emily_client::models::CreateWithdrawalRequestBody;
...

let emily_config = emily_client.config().as_testing();
let request = CreateWithdrawalRequestBody {
    ...
};

let response = withdrawal_api::create_withdrawal(&emily_config, request).await;
assert!(response.is_ok());
```

djordon: Also, just checking, does an `Ok` response from the `create_withdrawal` request mean that it was successfully updated? Would a 400 response also mean be `Ok`?

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3763,6 +3763,44 @@ async fn sign_bitcoin_transaction_withdrawals() {
...
+    assert!(testing_emily_client::apis::withdrawal_api::get_withdrawals(
+        &emily_client.config().as_testing(),
+        testing_emily_client::models::Status::Accepted,
+        None,
+        None,
+    )
+    .await
+    .unwrap()
+    .withdrawals
+    .is_empty());
```
djordon, line 3802: same thing here, can we break this out into a two-step thing:
```rust
let response = withdrawal_api::get_withdrawals(config, ...).await.unwrap();
assert!(response.withdrawals.is_empty());
```

signer/src/transaction_coordinator.rs
```diff
@@ -647,10 +647,24 @@ where
...
+                    tracing::error!(%error, "could not accept deposits on Emily");
+                });
+
+            let _ = self
+                .context
+                .get_emily_client()
+                .accept_withdrawals(&transaction, &stacks_chain_tip)
+                .await
+                .inspect_err(|error| {
+                    tracing::error!(%error, "could not accept withdrawals on Emily");
```
djordon, line 666: I would change these to warning level logs. Updating Emily is important, but I feel it is on a different level when compared to, say, getting an error from our database or something like that.

djordon (commented):
Largely looks good. Had some relatively minor comments.

signer/tests/integration/transaction_coordinator.rs
djordon: > About usings -- I'm afraid they can lead to name conflicts between testing api and public api, so it would be better to alias them in this import aaand, probably won't make code much more beautiful, so I'd leave it as is, but if it is a strong preference from your side I don't mind to change.

You call on the imports here. Just want to point out that I don't think that we'll have an import conflict when importing in our integration tests because for the testing API we need to use the testing config and the testing request body types.


> This what I see in `create_withdrawal` method in Emily code, and I don't think 400 will create withdrawal. Also, if 400 create withdrawal somehow test will false fail, not false pass, which is ok

True, but it may make thinks a little difficult to figure out and track down when/if it fails later. Better to check for the thing that we want, which is that we got a success status.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3766,40 +3768,36 @@ async fn sign_bitcoin_transaction_withdrawals() {
...
+        recipient: withdrawal_request.recipient.to_string(),
+        request_id: withdrawal_request.request_id,
+        sender: withdrawal_request.sender_address.to_string(),
+        stacks_block_hash: withdrawal_request.block_hash.to_string(),
+        stacks_block_height: db
+            .get_stacks_block(&stacks_chain_tip)
+            .await
+            .unwrap()
+            .unwrap()
+            .block_height,
```
djordon, line 3785: Can we break this out from here as well

matteojug (approved):

signer/tests/integration/transaction_coordinator.rs
djordon: Just looked through the code and we actually return an `Err` when status code is 400 or greater (but less than 600). So what is here is the right check for what we want.

djordon (approved):
Looks good

PR #1494, opened 2025-03-10
chore: sidecar use uv for dependency management

Jiloc:
## Description

Given the recent discussions and events, along with the addition of new dependencies in the sidecar to track deposit statuses (RBFs, Locktime, etc.), it became more important to improve dependency management for the sidecar. Using [UV](https://github.com/astral-sh/uv) for locked dependencies provides better security and reproducibility compared to the simpler requirements.txt approach.

## Changes
-Migrate Python dependency management to UV
- Update project structure to be compatible with UV
- Upgrade Dockerfile to use Python 3.13 and install UV
- Update integration logic to include the new recipient field

## Checklist:


docker/sbtc/emily-sidecar/Dockerfile
```diff
@@ -1,11 +1,17 @@
-FROM python:3.12-slim AS emily-sidecar
+FROM python:3.13-slim AS emily-sidecar
 
-WORKDIR /app
+# Install uv.
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
```
matteojug, line 4: we should probably pin uv version (via a sha hash of the image I guess, given how we are installing it)

docker/sbtc/emily-sidecar/Dockerfile
```diff
@@ -1,11 +1,17 @@
...
+# Install the application dependencies.
+WORKDIR /app
+RUN uv sync --frozen --no-cache
+
 # Expose the FastAPI port
 EXPOSE 20540
 ENV EMILY_API_KEY=testApiKey
 ENV EMILY_ENDPOINT=http://emily-server:3031
-CMD ["uvicorn", "--host", "0.0.0.0", "--port", "20540", "--workers", "4", "--timeout-keep-alive", "30", "--log-level" , "info", "main:app"]
+CMD ["/app/.venv/bin/uvicorn", "--host", "0.0.0.0", "--port", "20540", "--workers", "4", "--timeout-keep-alive", "30", "--log-level" , "info", "app.main:app"]
```
matteojug, line 17: nit: we could use `uv run` instead of the hardcoded path

emily_sidecar/pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[project]
+name = "emily-sidecar"
+version = "0.1.0"
+description = "Add your description here"
+requires-python = ">=3.13"
+dependencies = [
+    "apscheduler>=3.11.0",
+    "btclib>=2023.7.12",
```
matteojug, line 8: qq: are those two new dependencies for something new (not in here)?

emily_sidecar/pyproject.toml
matteojug, line 1: Can we pin exactly (instead of `>=`)? I guess it's generally fine since we have the lock file, but with `>=` I guess we may end up trying to update the lock more frequently

emily_sidecar/pyproject.toml
Jiloc: yeah, I decided to integrate uv just when I was adding those extra modules. One is for handling cron jobs and the second in for parsing the scripts and extract the locktime. But I should have added them in the appropriate PR

matteojug: if it's not too much a hassle can we remove those from this one? Just from a process perspective, since we are paying more attention to dependencies

matteojug (approved):

PR #1496, opened 2025-03-10
fix: random update_withdrawals_is_forbidden fails

Jiloc:
## Description

Fixes the test that were occasionally failing in CI. As described in [here ](https://github.com/frondeus/test-case/pull/48), that's a known problem in the test_case library. It requires the tokio::test to be defined after the test cases

Note that I wasn't able to reproduce the error locally.

## Changes
- move tokio::test definition after the test cases

## Checklist:


matteojug (commented):
Does this also apply to `create_deposit_handles_duplicates`?

matteojug (approved):

PR #1497, opened 2025-03-11
[chore]: info about tests to readme

MCJOHN974:
## Description

Closes: no issue

I often use this command and sometimes loose it. Extract it from readme will be easier then extracting from my shell history. Is this readme okish place for this command or maybe `signer/README.md` will fit better?

## Changes

Added few lines to readme

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,9 @@ Once running, the following services are available:
 - Stacks explorer at [localhost:3020](http://localhost:3020)
 - Mempool.space Bitcoin explorer at [localhost:8083](http://localhost:8083)
 
+Sometimes this docker-compose command can be useful for running integration tests
+`docker compose --file docker/docker-compose.test.yml up`
+
 #### Update local docker builds
 
 To rebuild the containers from your current branch you can use:


matteojug:
Isn't this already covered by:
https://github.com/stacks-network/sbtc/blob/394dc72e5bc31db2e51d88f3f7445a724072fa56/README.md?plain=1#L82-L84
?

djordon:
> @matteojug I trying to use our make commands and they sometimes do not work well. This manual docker compose works fine every time I use it
> 
> Maybe it is a call to add some changes to Makefile, but I personally ok to just call this docker compose manually =)

@MCJOHN974 what specific issues have you run into and how often are you running into them? What you added to the README is the `make integration-env-up` command but without basically running `pnpm --filter run synth`.

MCJOHN974:
I can't reproduce any particular issue, however it still can be useful for running integration tests and look at emily logs

djordon:
Closing, since we cannot reproduce the underlying issue. And if we could, we should probably address that directly by updating the `make` command or solving the underlying problem.

PR #1502, opened 2025-03-11
feat: sidecar deposit management

Jiloc:
## Description

Closes: #?

## Checklist:


djordon (commented):
Why do this in python? I can understand why the sidecar was done in python; we were in a time crunch where we needed to ship the changes in under a week or something like that, we felt that we couldn't get it done in Rust in time, and we were going to remove it eventually when the stacks node can reach out to Emily directly. None of these conditions are present in this case, so I feel like the presumption is that we should ship this in Rust.


Here is why. With Rust, we can rely on all of the shared code. There would be no need to reimplement any of our existing types, and no need for any of us to double check that it is done correctly. Same thing for any future changes. It's less cognitive load. Plus, as a team we have a shared knowledge of Rust and it's tooling, and that same shared knowledge isn't same with other languages, including Python. Overall I feel it's simpler. Same tooling, code style/formatting, reusable testing setups, and so on. 

So in general, unless there is a compelling case to use something else, I think it would be better to stick to using Rust here. Thoughts?

emily_cron/app/services/deposit_processor.py
```diff
@@ -0,0 +1,155 @@
...
+        # Find transactions with expired locktime
+        locktime_expired_txs = [
+            tx
+            for tx in enriched_deposits
+            if tx.confirmed_height > 0  # Only process confirmed transactions
+            and bitcoin_chaintip.height
+            >= tx.confirmed_height
+            + tx.lock_time
+            + settings.MIN_BLOCK_CONFIRMATIONS  # Check if locktime has expired
+        ]
```
matteojug, line 49: This may not be enough, if we have a sweep tx from the signers in the mempool and no signers activity after that; the correct way to handle this case is a bit annoying (see how we check for withdrawals _safe to reject_ case).
If we erroneously mark a deposit as failed but the signers go ahead and (correctly) process the request, will the sidecar be able to update this request as confirmed (once minted)?

emily_cron/app/services/deposit_processor.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # Find transactions with expired locktime
+        locktime_expired_txs = [
+            tx
+            for tx in enriched_deposits
+            if tx.confirmed_height > 0  # Only process confirmed transactions
+            and bitcoin_chaintip.height
+            >= tx.confirmed_height
+            + tx.lock_time
+            + settings.MIN_BLOCK_CONFIRMATIONS  # Check if locktime has expired
```
matteojug, line 48: nit: for readability, can we move this to a separate _is expired_ (or similar) function?

emily_cron/app/services/deposit_processor.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+
+        for tx in locktime_expired_txs:
+            logger.info(f"Marking transaction {tx.bitcoin_txid} with expired locktime as FAILED")
+            updates.append(
+                DepositUpdate(
+                    bitcoin_txid=tx.bitcoin_txid,
+                    bitcoin_tx_output_index=tx.bitcoin_tx_output_index,
+                    last_update_height=stacks_chaintip.height,
+                    last_update_block_hash=stacks_chaintip.hash,
```
matteojug, line 65: nit: `last_update_height` and block hash no longer needed

emily_cron/app/services/deposit_processor.py
```diff
@@ -0,0 +1,155 @@
...
+        transaction_details = []
+        for deposit in deposits:
+            tx_data = MempoolAPI.get_bitcoin_transaction(deposit.bitcoin_txid)
+
+            if not tx_data:
+                transaction_details.append(EnrichedDepositInfo.from_missing(deposit))
+                continue
+
+            spending_outputs = {
+                vout.get("scriptpubkey_address", "Unknown"): vout.get("value", 0)
```
matteojug, line 135: Both here and below: I'm not sure if default some of those is safe. What about logging and ignoring the ones missing the required ones?

matteojug (commented):
Some preliminary comments, but my main question is why not integrating the cron in the same "service" as the sidecar, so that we may be able to reuse something (at least the config) and have less moving parts.

docker/docker-compose.yml
```diff
@@ -150,6 +150,26 @@ services:
     depends_on:
       - emily-server
 
+  emily-cron:
```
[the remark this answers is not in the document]
fabergat: I guess we can do that, we have to work a bit with the docker file but nothing crazy, the only concern is about the logs, probably they will be multiplexed together but I don't see as a big no

emily_cron/app/clients/emily.py
```diff
@@ -0,0 +1,49 @@
...
+        self.headers = {"x-api-key": api_key}
+
+    @classmethod
+    def fetch_deposits(cls, status: RequestStatus) -> list[DepositInfo]:
+        """Fetch deposits based on status."""
+        data = cls.get(f"/deposit?status={status.value}")
+        return [DepositInfo.from_json(deposit) for deposit in data.get("deposits", [])]
+
+
+class PrivateEmilyAPI(APIClient):
```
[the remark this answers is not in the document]
matteojug: I'm not sure doing that it's for the best, as it would mean our services would need to go to the public endpoint for some stuff and (eg) share the rate limiting with the external users. Keeping them separated (while for a kinda cursed reason) may have some benefits, so I would keep in the private all the ones we may need even if already present in the public one.

emily_cron/app/clients/base.py
```diff
@@ -0,0 +1,61 @@
...
+    def _make_request(
+        cls,
+        method: str,
+        endpoint: str,
+        params: Optional[dict[str, Any]] = None,
+        json_data: Optional[dict[str, Any]] = None,
+        headers: Optional[dict[str, str]] = None,
+    ) -> dict[str, Any]:
+        """Make an HTTP request and return JSON response."""
+        url = f"{cls.BASE_URL}{endpoint}"
```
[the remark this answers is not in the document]
matteojug: Doing `rstrip("/")` of the whole string does something different though: I was worried about mistakingly setting `BASE_URL` to (eg) `http://emily.whatever/` and messing up the endpoints (as we would have `//`). Since the endpoints are hardcoded I'm don't think we need rstrip that one.

emily_cron/app/clients/base.py
```diff
@@ -18,21 +18,47 @@ def _make_request(
...
             response.raise_for_status()
-            return response.json()
+
+            try:
+                return response.json()
+            except Exception as e:
+                logger.error(f"Invalid JSON response from {method} {url}: {e}")
+                if raise_for_error:
+                    raise
+                return {}
```
matteojug, line 56: nit: instead of the nested try we could differentiate by exception type

emily_cron/app/clients/hiro.py
```diff
@@ -0,0 +1,73 @@
...
+            txid: The Bitcoin transaction ID
+
+        Returns:
+            bool: True if the deposit was completed, False otherwise
+        """
+        params = {
+            "sender": settings.DEPLOYER_ADDRESS,
+            "arguments": [
+                f"0x0200000020{txid}",
+                "0x0100000000000000000000000000000000",
```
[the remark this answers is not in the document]
matteojug: mmm `f"{hex(vout)}100000000000000000000000000000000"` seems wrong, are you sure about it?

emily_cron/app/clients/base.py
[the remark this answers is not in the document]
matteojug: Oh nvm, I didn't notice we already have `removesuffix` from the config. Then I guess we can just restore this as it was (without the rstrip)

docker/docker-compose.yml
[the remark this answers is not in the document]
matteojug: Resolving since a single container is not that bad

emily_cron/app/services/deposit_processor.py
```diff
@@ -0,0 +1,191 @@
...
+        # Get current blockchain state
+        bitcoin_chaintip_height = MempoolAPI.get_tip_height()
+        stacks_chaintip = HiroAPI.get_stacks_block()
+
+        logger.info(f"Bitcoin chain tip: {bitcoin_chaintip_height}")
+        logger.info(f"Stacks chain tip: {stacks_chaintip}")
+
+        # Fetch pending and accepted deposits
+        pending_deposits = PrivateEmilyAPI.fetch_deposits(RequestStatus.PENDING)
+        accepted_deposits = PrivateEmilyAPI.fetch_deposits(RequestStatus.ACCEPTED)
```
matteojug, line 139: nit: can we add a comment explaining why we are also checking the accepted ones? IIUC it's because emily cannot detect bitcoin only forks, so we could have accepted deposits forked (and invalidated) but still in an accepted state

emily_cron/app/services/deposit_processor.py
```diff
@@ -0,0 +1,191 @@
...
+                )
+                continue  # Cannot determine reclaim, assume it's a signer sweep
+
+            logger.debug(
+                f"Deposit {tx.bitcoin_txid} UTXO spent by {spending_txid}, checking for reclaim..."
+            )
+            spending_tx_details = MempoolAPI.get_transaction(spending_txid)
+            is_reclaim_check = False
+            if spending_tx_details:
+                for vin in spending_tx_details.get("vin", []):
```
matteojug, line 78: nit: the `outspend` API returns `vin` (index), we could use it to avoid finding it again

emily_cron/app/services/deposit_processor.py
```diff
@@ -0,0 +1,191 @@
...
+                    if (
+                        vin.get("txid") == tx.bitcoin_txid
+                        and vin.get("vout") == tx.bitcoin_tx_output_index
+                    ):
+                        witness_data = vin.get("witness", [])
+                        # Simple check: does witness contain the reclaim script?
+                        if any(tx.reclaim_script in item for item in witness_data):
+                            is_reclaim_check = True
+                            break
```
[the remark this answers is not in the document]
matteojug: I don't think we can avoid checking the reclaim (or deposit) scripts, we could look for who is spending it (if it's the signers aggregate key) but that changes over time so the script check seems the more robust way. Also better to check the reclaim than the spending script since the latter format is fixed, while the former is not.

sanity check: is it true though that the whole reclaim script is always contained in a single witness entry?

matteojug (dismissed):
LGTM, some nits/qq

emily_cron/app/services/deposit_processor.py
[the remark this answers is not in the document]
djordon: Just want to make I understand the question so let me rephrase: are we certain that every transaction input used to reclaim a user's funds after a failed deposit attempt includes the reclaim script in its witness data? I believe the answer is yes. All reclaims must spend from the spending path of a taproot output. In order to spend from a spending path of a taproot output, the witness data must include:
1. The spending script in the merkle tree of the taproot output.
2. A proof that the spending script is in the merkle tree of the taproot output.
3. The data that satisfies the conditions of the spending script.

The above is true of all inputs attempting to spend from a spending path of a taproot output. Did I answer the right question?

emily_cron/app/models/deposit.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @property
+    def max_fee(self) -> int:
+        """Extracts the max fee from deposit script."""
+        script = Script.parse(self.deposit_script)
+        if script.redeemscript:
+            max_fee_bytes = script.redeemscript
+        else:
+            max_fee_bytes = bytes.fromhex(script.view().split()[0])
+        return int.from_bytes(max_fee_bytes[:8], byteorder="big")
```
djordon, line 65: Are both of these `int.from_bytes` correct? Python integers are signed, but the `max_fee` and the `locktime` integers are both signed. For `locktime`s I suspect that with our current code it doesn't matter in our case, since python represents integers using the two's complement layout. With the expected layout of the `locktime` this shouldn't be an issue. But pointing it out to double check.

emily_cron/app/models/deposit.py
```diff
@@ -48,16 +48,21 @@ def from_json(cls, j: dict[str, Any]) -> Self:
     @property
     def lock_time(self) -> int:
         """Extracts lock time from reclaim script."""
-        lock_time = script.parse(self.reclaim_script)[0]
-        if lock_time.startswith("OP_"):
-            return int(lock_time[len("OP_") :])
-        return int.from_bytes(bytes.fromhex(lock_time), byteorder="little")
+        script = Script.parse(self.reclaim_script)
+        op_code_maybe = script.view().split()[0]
```
matteojug, line 52: nit: using `as_list=True` in `view` we can avoid join+split (and be sure there's nothing messing up the splitting); also below

emily_cron/app/models/deposit.py
matteojug: From the docs (https://docs.python.org/3/library/stdtypes.html#int.from_bytes), by default `signed=False`, so this is producing only unsigned values -- which could be wrong for the lock time IIUC. 
While I guess the second one is correct since that comes from "our" encoding, and we want it unsigned.

emily_cron/app/models/deposit.py
```diff
@@ -48,16 +48,21 @@ def from_json(cls, j: dict[str, Any]) -> Self:
...
     @property
     def max_fee(self) -> int:
         """Extracts the max fee from deposit script."""
-        data = script.parse(self.deposit_script)[0]
-        return int.from_bytes(bytes.fromhex(data)[:8])
+        script = Script.parse(self.deposit_script)
+        if script.redeemscript:
+            max_fee_bytes = script.redeemscript
+        else:
+            max_fee_bytes = bytes.fromhex(script.view().split()[0])
```
matteojug, line 64: qq: how is `redeemscript` computed, and why we don't get it always (and we need the `else`)?

emily_cron/app/models/deposit.py
matteojug: mmm apparently bitcoin uses a different encoding. Looking at https://bitcoin.stackexchange.com/questions/122939/understanding-scripts-number-encoding-cscriptnum, taking the first example (`[0x32; 0x79; 0x86]`), the bitcoin code does produce the claimed value (https://onecompiler.com/cpp/43e89drzc, taken from https://github.com/bitcoin/bitcoin/blob/cfe025ff0e1dcf984179a5a9b01dc8b865d95409/src/script/script.h#L383), while python produces something different:
```
>>> int.from_bytes(bytes([0x32, 0x79, 0x86]), byteorder="little", signed=True)
-7964366
```

djordon: And shouldn't the max fee also be `signed=False`?

matteojug: Nice! Can you also add a comment with the source of that code in `decode_cscript_int`?

matteojug (approved):

djordon (commented):
Not a lot of substantive comments so far. Most of my comments are about potential future changes to responses that Emily gives.

emily_cron/app/services/deposit_processor.py
```diff
@@ -0,0 +1,197 @@
...
+logger = logging.getLogger(__name__)
+
+
+class DepositProcessor:
+    """Service for processing deposits."""
+
+    def process_expired_locktime(
+        self,
+        enriched_deposits: list[EnrichedDepositInfo],
+        bitcoin_chaintip_height: int,
```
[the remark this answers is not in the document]
djordon: Ahhh right, I forgot that she does that. No need to change anything.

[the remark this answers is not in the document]
djordon: Oh I brought it up because in the signer codebase we use `chain_tip` all over the place, but we should keep Emily naming conventions consistent within the Emily's sphere. Maybe later we can rename stuff if someone wants, but I'm okay so long as we're consistent.

emily_cron/app/clients/base.py
```diff
@@ -0,0 +1,88 @@
...
+                headers=headers,
+                timeout=30,
+            )
+            response.raise_for_status()
+            return response.json()
+
+        except (JSONDecodeError, RequestException) as e:
+            logger.error(f"Error during {method} {url}: {e}")
+            if not ignore_errors:
+                raise
```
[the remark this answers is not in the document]
djordon: Oh, that makes sense. Maybe we can update the documentation/comments in cases where we don't ignore errors to state that the responses in those cases are critical to further processing, hence we raise on 400-599 response status codes.

emily_cron/app/models/deposit.py
```diff
@@ -0,0 +1,162 @@
...
+    def lock_time(self) -> int:
+        """Extracts lock time from reclaim script."""
+        script = Script.parse(self.reclaim_script)
+        op_code_maybe = script.view(as_list=True)[0]
+        if op_code_maybe.startswith("OP_"):
+            return int(op_code_maybe[len("OP_") :])
+        return decode_cscript_int(script.commands[0])
+
+    @property
+    def max_fee(self) -> int:
```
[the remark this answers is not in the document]
djordon: 👍🏽 🙏🏽

matteojug (approved):

emily_cron/app/services/deposit_processor.py
```diff
@@ -0,0 +1,197 @@
...
+        # a previously accepted deposit, or a malicious signer may have modified
+        # a pending or invalid deposit to Accepted.
+        pending_deposits = PrivateEmilyAPI.fetch_deposits(RequestStatus.PENDING)
+        accepted_deposits = PrivateEmilyAPI.fetch_deposits(RequestStatus.ACCEPTED)
+
+        # Enrich deposits with additional transaction data
+        enriched_deposits = self._enrich_deposits(chain(pending_deposits, accepted_deposits))
+
+        # Process deposits and collect updates
+        updates = []
```
setzeus, line 151: NIT - since 'locktime_updates' isn't used anywhere else, is there any reason to not initialize & set 'updates' in one line? 

`updates = self.process_expired_locktime(enriched_deposits, bitcoin_chaintip_height)`

emily_cron/app/clients/hiro.py
```diff
@@ -0,0 +1,99 @@
...
+        if not result.get("okay") or result.get("result") == "0x09":
+            return False  # Not minted yet
+        # 0x0a represents 'Some' in Clarity's option type encoding
+        # 0x0c represents a Tuple type in Clarity
+        # If we get 'Some(Tuple(...))', the deposit is completed
+        return result.get("result", "").startswith("0x0a0c")
+
+    @classmethod
+    @functools.lru_cache(maxsize=256)
+    def _get_stacks_block_cached(cls, height_or_hash: int | str) -> dict[str, Any]:
```
setzeus, line 55: Curious more than anything else but any reason why this function name starts w/ an underline?

matteojug (approved):

PR #1506, opened 2025-03-12
feat: sidecar RBF management

### LINKED ISSUE #1268: [Feature]: RBF Deposits Management
## 1. Description

Currently, Emily does not track RBF transactions. When an RBF transaction occurs, the Bridge sends a new `create_deposit` request while the previous deposit request remains indefinitely in a `PENDING` state. 

This feature aims to address the issue by enabling Emily to handle RBF transactions more effectively, minimizing system bloat and mitigating potential security risks.

### 1.1 Context & Purpose

Without proper RBF handling, Emily accumulates duplicate or stale deposit requests, leading to inefficiencies and potential data integrity issues. Properly managing RBF transactions will ensure accurate state tracking while preventing malicious exploits.

### Proposed Solution:
- Add a parameter to the `create_deposit` endpoint to link the new deposit to an existing deposit in Emily.
  - The linked deposit must be in the `PENDING` state when the request is made.
  - If the parameter is provided, update the existing deposit rather than creating a new one.
- Use the full `TX_HEX` of the new transaction to determine if it is a RBF of an existing deposit?

### Challenges:
- **Data Authenticity:** Emily cannot independently verify the authenticity of the data in `create_deposit` requests. This introduces the risk of malicious actors linking legitimate user requests to fake RBF transactions, invalidating them.
  - Investigate ways to validate the authenticity of the `TX_HEX` data to mitigate this risk.

### 2.1 Acceptance Criteria:
   - [ ] A parameter is added to the `create_deposit` endpoint to link a new deposit to an existing one in `PENDING` status.
   - [ ] Duplicate deposits are avoided, and the original deposit state is updated instead.
   - [ ] Validation mechanisms are implemented to minimize the risk of malicious linking of deposits.
   - [ ] Ensure proper logging and monitoring for all RBF-related actions for future auditing.

Jiloc:
## Description

Closes: #1268 

This PR implements the Locktime Request Management feature to address the issue of pending requests remaining in the system indefinitely. The implementation focuses on automatically marking deposits as FAILED when they exceed their locktime without being fulfilled.

## Changes
-  Added RBF transaction detection and processing to the DepositProcessor service:
    - Enhanced the deposit enrichment process to check for RBF replacements for unconfirmed transactions
    - Implemented process_rbf_transactions method to identify and handle RBF'd transactions
    - Added logic to group related RBF transactions into chains using the _group_rbf_transactions helper method
    - When a transaction in an RBF chain gets confirmed, all other unconfirmed transactions in the chain are marked as FAILED
- Extended the EnrichedDepositInfo model to include rbf_txids field to track transaction replacements
- Integrated RBF processing into the main deposit update workflow

## Testing Information
- Added unittest - integrated with CI

## Checklist:


emily_cron/app/services/deposit_processor.py
```diff
@@ -160,6 +214,49 @@ def update_deposits(self) -> None:
...
+
+        # First, build groups of related transactions (original + replacements)
+        for tx in rbf_txs:
+            # Create a set of all txids in this RBF chain
+            chain_txids = set(tx.rbf_txids)
+            chain_txids.add(tx.bitcoin_txid)
+
+            # Find all groups that overlap with this chain
+            overlapping_groups = []
+            for group_id, group_txids in list(rbf_groups.items()):
```
djordon, line 234: I think the `list` is unnecessary here, since `items()` should return a generator.

djordon (commented):
Largely looks good. I need to go back over the tests and the `_group_rbf_transactions` function.

emily_cron/app/clients/mempool.py
```diff
@@ -7,11 +7,48 @@
...
+    return txids
+
+
 class MempoolAPI(APIClient):
     """Client for interacting with the Mempool API."""
 
     BASE_URL = settings.MEMPOOL_API_URL
 
+    @classmethod
+    def check_for_rbf(cls, txid: str) -> set[str]:
```
MCJOHN974, line 40: Is `cls` an alias for `self` ??

djordon: Kinda, it's shorthand for class. This is a class method, so it could be called without an instance of the type. When calling this method, it does not have access to the properties that get set during `__init__` or other methods, it can only access class fields.

emily_cron/test/test_deposit_processor.py
```diff
@@ -32,21 +32,38 @@
...
+    """Base class for DepositProcessor tests with common setup."""
 
     def setUp(self):
         self.processor = DepositProcessor()
         self.current_time = int(datetime.now().timestamp())
 
         # Mock blockchain state
+        self.stacks_chaintip = BlockInfo(height=500, hash="stx_hash", time=self.current_time)
         self.bitcoin_chaintip_height = 1000
+        # settings.MIN_BLOCK_CONFIRMATIONS = 6
```
djordon, line 45: This seems like it can be removed.

emily_cron/app/services/deposit_processor.py
```diff
@@ -16,6 +16,56 @@
...
+                if tx.confirmed_height is None:
+                    logger.info(
+                        f"Marking RBF'd transaction {tx.bitcoin_txid} as FAILED (replaced by confirmed tx)"
+                    )
+                    updates.append(
+                        DepositUpdate(
+                            bitcoin_txid=tx.bitcoin_txid,
+                            bitcoin_tx_output_index=tx.bitcoin_tx_output_index,
+                            status=RequestStatus.FAILED.value,
+                            status_message=f"Replaced by confirmed tx {confirmed_txs[0].bitcoin_txid}",
```
djordon, line 64: We should consider adding the replacement transaction ID to the deposit's event object instead of putting it within the status message. That way it will be easier to work with programmatically.

djordon (dismissed):
Looks good

emily_cron/app/services/deposit_processor.py
```diff
@@ -160,6 +227,65 @@ def update_deposits(self) -> None:
...
+        # First, build groups of related transactions (original + replacements)
+        for tx in rbf_txs:
+            # Create a set of all txids in this RBF chain
+            chain_txids = set(tx.rbf_txids)
+            chain_txids.add(tx.bitcoin_txid)
+
+            # Find all groups that overlap with this chain
+            overlapping_groups = []
+            for group_id, group_txids in rbf_groups.items():
+                if chain_txids.intersection(group_txids):
```
aldur, line 264: trivia: It looks like Mempool returns the _full_ RBF chain when queried for any transaction along the chain. So here the intersection should in practice always be either full or empty.

emily_cron/app/clients/mempool.py
```diff
@@ -7,11 +7,48 @@
 logger = logging.getLogger(__name__)
 
 
+def _collect_rbf_txids(data: dict[str, Any] | None) -> set[str]:
```
aldur, line 10: nit: Since the RBF transactions form a chain, a list or an ordered set might be more appropriate.

emily_cron/app/services/deposit_processor.py
```diff
@@ -16,6 +16,69 @@
...
+        if not rbf_txs:
+            return updates
+
+        logger.info(f"Found {len(rbf_txs)} transactions with RBF replacements")
+
+        # Group by replacement chain
+        rbf_groups = self._group_rbf_transactions(rbf_txs)
+
+        # Process each group
+        for group_txids in rbf_groups.values():
```
aldur, line 55: nit: You could use `itertools.groupby` if you first sorted the `rbf_txs` so that all those in the same chain are adjacent to each other. This one could be easily doable as well, for instance you could do `''.join(tx.rbf_txid)` and then use lexicographic sorting. This would also remove the need for `_group_rbf_transactions`.

Possibly, it would make things less legible though.

emily_cron/app/services/deposit_processor.py
```diff
@@ -16,6 +16,56 @@
...
+            for tx in group_txs:
+                if tx.confirmed_height is None:
+                    logger.info(
+                        f"Marking RBF'd transaction {tx.bitcoin_txid} as FAILED (replaced by confirmed tx)"
+                    )
+                    updates.append(
+                        DepositUpdate(
+                            bitcoin_txid=tx.bitcoin_txid,
+                            bitcoin_tx_output_index=tx.bitcoin_tx_output_index,
+                            status=RequestStatus.FAILED.value,
```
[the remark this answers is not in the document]
aldur: +1 on this

emily_cron/app/services/deposit_processor.py
aldur: nit: Based on this information you could actually reduce the complexity of this from N**2 to NlogN by:

1. Storing the RBF chain sorted.
2. Sorting `rbf_txs` based on the chains.

(See my other comment about `itertools.groupby`).

aldur (commented):
I haven't reviewed the tests but assume they pass (and we run them in CI, please let me know if we don't).

This looks good! Since for once I found myself in very familiar code, I just went ahead and left a few nits/code golf comments. No need to address them, just resolve them when you want this merged.

aldur (approved):

emily_cron/app/clients/mempool.py
```diff
@@ -7,11 +7,48 @@
...
+
+    Returns:
+        List of transaction IDs that replaced the original transaction
+    """
+    txids = []
+
+    if data is None:
+        return txids
+
+    if tx := data.get("tx", {}):
```
matteojug, line 24: nit: since we are checking for falsy we don't need the default. Or we could keep it and just remove the `if` (since it would be caught by the `if txid := tx.get("txid")` anyway). Or even chain the two in a single if.
```suggestion
    if tx := data.get("tx"):
```

aldur (approved):

emily_cron/app/clients/mempool.py
```diff
@@ -7,11 +7,48 @@
...
 
+    @classmethod
+    def check_for_rbf(cls, txid: str) -> list[str]:
+        """Check if a Bitcoin transaction was replaced by RBF.
+
+        Args:
+            txid: The transaction ID to check
+
+        Returns:
+            List of transaction IDs that replaced the original transaction
```
matteojug, line 47: nit: _[txid] that replaced the original transaction_ here is a bit confusing, IIUC `_collect_rbf_txids` returns all the txs in the replacement "tree", where the `txid` argument passed in here could be in any point of that tree, so we get both the txs replaced by `txid` as well as the ones replacing `txid` (and even eventual siblings that are not either of those). A similar nit does applies to the doc of `_collect_rbf_txids` and `rbf_txids` comment.

emily_cron/app/services/deposit_processor.py
```diff
@@ -16,6 +16,81 @@
...
+
+        # Find transactions with RBF replacements
+        rbf_txs = [tx for tx in enriched_deposits if tx.rbf_txids]
+        if not rbf_txs:
+            return updates
+
+        logger.info(f"Found {len(rbf_txs)} transactions with RBF replacements")
+
+        # Group txs by RBF chain. All txs in a group have the same RBF chain.
+        for rbf_key, group_iter in groupby(rbf_txs, key=lambda tx: "".join(sorted(tx.rbf_txids))):
```
matteojug, line 52: nit: since we log `rbf_key`, maybe a separator can be helpful
```suggestion
        for rbf_key, group_iter in groupby(rbf_txs, key=lambda tx: ",".join(sorted(tx.rbf_txids))):
```

emily_cron/app/services/deposit_processor.py
```diff
@@ -151,6 +226,10 @@ def update_deposits(self) -> None:
         locktime_updates = self.process_expired_locktime(enriched_deposits, bitcoin_chaintip_height)
         updates.extend(locktime_updates)
 
+        # Process RBF transactions
+        rbf_updates = self.process_rbf_transactions(enriched_deposits)
+        updates.extend(rbf_updates)
```
matteojug, line 231: Should we wait some blocks before marking the RBFd as failed? In case of reorgs, we would may endup with failed deposits that shouldn't be.

emily_cron/test/test_deposit_processor.py
```diff
@@ -365,43 +366,191 @@ def tearDownModule():
...
+    @patch("app.clients.MempoolAPI.get_transaction")
+    @patch("app.clients.MempoolAPI.check_for_rbf")
+    def test_enrich_deposits_with_rbf(self, mock_check_rbf, mock_get_tx):
+        """Test the deposit enrichment process with RBF."""
+        # Create test deposits
+        deposit1 = self._create_mock_deposit(
+            txid="replacement1",
+            confirmed_height=None,
+            lock_time=10,
+            rbf_txids=["replacement1", "replacement2"],
```
matteojug, line 477: qq: why do we need `rbf_txids` in `deposit1` and `deposit2`? I would expect them to be populated by `mock_check_rbf`

emily_cron/test/test_deposit_processor.py
```diff
@@ -365,43 +366,191 @@ def tearDownModule():
...
+
+            # Verify from_deposit_info was called with the correct parameters
+            self.assertEqual(mock_from_info.call_count, 2)
+
+            # Assertions on the configured mock results
+            self.assertEqual(result[0].bitcoin_txid, deposit1.bitcoin_txid)
+            self.assertEqual(result[0].confirmed_height, deposit1.confirmed_height)
+            self.assertEqual(result[0].fee, 100000)
+            self.assertEqual(result[0].in_mempool, True)
+            self.assertEqual(result[0].rbf_txids, {"replacement1", "replacement2"})
```
matteojug, line 540: nit: can we add also the same check for `result[1]`, checking that `rbf_txids` is empty?

matteojug (approved):

djordon (approved):

PR #1507, opened 2025-03-12
feat: sidecar long pending deposits management

Jiloc:
## Description

This PR adds support for handling long-pending transactions. It addresses the issue of transactions that remain in a pending state for extended periods without being confirmed in the mempool, which can lead to system bloat.

## Changes

- Added long-pending transaction detection and processing to the DepositProcessor service:
- Implemented process_long_pending method to identify transactions that have been pending for too long
- Added logic to mark transactions as FAILED if they:
       - Are in PENDING status
       - Cannot be found in the mempool (likely dropped)
       - Have been pending for longer than the configured maximum time (default: 24 hours)
       - Integrated long-pending transaction processing into the main deposit update workflow
- Added MAX_UNCONFIRMED_TIME configuration setting (default: 86400 seconds / 24 hours) to control how long a transaction can remain unconfirmed before being marked as failed

## Testing Information
- WIP

## Checklist:

### DIFF
--- a/emily_cron/app/services/deposit_processor.py
+++ b/emily_cron/app/services/deposit_processor.py
@@ -87,16 +87,13 @@ class DepositProcessor:
             list[DepositUpdate]: List of deposit updates
         """
         updates = []
-
         # Find transactions with expired locktime
         locktime_expired_txs = [
             tx
             for tx in enriched_deposits
             if tx.confirmed_height > 0  # Only process confirmed transactions
-            and bitcoin_chaintip.height
-            >= tx.confirmed_height
-            + tx.lock_time
-            + settings.MIN_BLOCK_CONFIRMATIONS  # Check if locktime has expired
+            # Check if locktime has expired
+            and bitcoin_chaintip.height >= tx.confirmed_height + tx.lock_time + settings.MIN_BLOCK_CONFIRMATIONS
         ]
 
         if not locktime_expired_txs:
@@ -121,6 +118,50 @@ class DepositProcessor:
 
         return updates
 
+    def process_long_pending(
+        self,
+        enriched_deposits: list[EnrichedDepositInfo],
+        stacks_chaintip: BlockInfo,
+    ) -> list[DepositUpdate]:
+        """Process long-pending transactions.
+        Args:
+            enriched_deposits: List of enriched deposit information
+            stacks_chaintip: Current Stacks block info
+        Returns:
+            list[DepositUpdate]: List of deposit updates
+        """
+        updates = []
+
+        # Get the current time
+        current_time = int(datetime.now().timestamp())
+
+        long_pending_txs = [
+            tx
+            for tx in enriched_deposits
+            if tx.status == RequestStatus.PENDING.value  # Only check pending transactions
+            and not tx.in_mempool  # that we can't find via the mempool API (it might have been dropped)
+            and current_time - tx.deposit_time >= settings.MAX_UNCONFIRMED_TIME  # and has been pending for too long
+        ]
+
+        if not long_pending_txs:
+            return updates
+
+        logger.info(f"Found {len(long_pending_txs)} long-pending transactions to mark as FAILED")
+
+        for tx in long_pending_txs:
+            logger.info(f"Marking long-pending transaction {tx.bitcoin_txid} as FAILED")
+            updates.append(
+                DepositUpdate(
+                    bitcoin_txid=tx.bitcoin_txid,
+                    bitcoin_tx_output_index=tx.bitcoin_tx_output_index,
+                    last_update_height=stacks_chaintip.height,
+                    last_update_block_hash=stacks_chaintip.hash,
+                    status=RequestStatus.FAILED.value,
+                    status_message=f"Pending for too long ({settings.MAX_UNCONFIRMED_TIME} seconds)",
+                )
+            )
+
+        return updates
 
     def update_deposits(self) -> None:
         """Update deposit statuses.
@@ -161,6 +202,13 @@ class DepositProcessor:
             stacks_chaintip,
         )
         updates.extend(rbf_updates)
+
+        # Process long-pending transactions
+        pending_updates = self.process_long_pending(
+            enriched_deposits,
+            stacks_chaintip,
+        )
+        updates.extend(pending_updates)
 
         # Apply updates
         if updates:
--- a/emily_cron/app/settings.py
+++ b/emily_cron/app/settings.py
@@ -11,3 +11,6 @@ HIRO_API_URL = os.getenv("HIRO_API_URL", "https://api.hiro.so/extended").removes
 
 # The number of confirmations required for a deposit update to be considered final
 MIN_BLOCK_CONFIRMATIONS = int(os.getenv("MIN_BLOCK_CONFIRMATIONS", 6))
+
+# Maximum time (in seconds) a transaction can remain unconfirmed before being marked as FAILED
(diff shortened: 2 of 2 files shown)


emily_cron/app/services/deposit_processor.py
```diff
@@ -203,6 +204,49 @@ def process_expired_locktime(
...
         return updates
 
+    def process_long_pending(
+        self,
+        enriched_deposits: list[EnrichedDepositInfo],
+    ) -> list[DepositUpdate]:
+        """Process long-pending transactions.
+        Args:
+            enriched_deposits: List of enriched deposit information
+            stacks_chaintip: Current Stacks block info
```
matteojug, line 214: nit: `stacks_chaintip` leftover

emily_cron/app/services/deposit_processor.py
```diff
@@ -203,6 +204,49 @@ def process_expired_locktime(
...
+        Returns:
+            list[DepositUpdate]: List of deposit updates
+        """
+        updates = []
+
+        # Get the current time
+        current_time = int(datetime.now().timestamp())
```
matteojug, line 221: Should we compare the time wrt the latest stacks block instead of system time? For no overwhelmingly good reason, just to avoid tz/clock skew issues.

emily_cron/app/services/deposit_processor.py
```diff
@@ -203,6 +204,49 @@ def process_expired_locktime(
...
+
+        long_pending_txs = [
+            tx
+            for tx in enriched_deposits
+            if tx.status == RequestStatus.PENDING.value  # Only check pending transactions
+            and not tx.in_mempool  # that we can't find via the mempool API (it might have been dropped)
+            and current_time - tx.deposit_time
+            >= settings.MAX_UNCONFIRMED_TIME  # and has been pending for too long
+        ]
```
matteojug, line 230: maybe nit: moving the last comment on its own line before avoiding breaking the last condition may improve the list comprehension readability

emily_cron/app/models/deposit.py
```diff
@@ -66,7 +66,7 @@ def max_fee(self) -> int:
     @functools.cached_property
     def deposit_time(self) -> int:
```
matteojug, line 67: maybe nit: `deposit_time` is a bit misleading as it could suggest (given the usage below) it's the creation time, maybe `deposit_last_update` or similar could be a more precise name

emily_cron/app/services/deposit_processor.py
```diff
@@ -203,6 +204,49 @@ def process_expired_locktime(
...
+        # Get the current time
+        current_time = int(datetime.now().timestamp())
+
+        long_pending_txs = [
+            tx
+            for tx in enriched_deposits
+            if tx.status == RequestStatus.PENDING.value  # Only check pending transactions
+            and not tx.in_mempool  # that we can't find via the mempool API (it might have been dropped)
+            and current_time - tx.deposit_time
+            >= settings.MAX_UNCONFIRMED_TIME  # and has been pending for too long
```
matteojug, line 229: very nit: since `MAX_UNCONFIRMED_TIME` is _Maximum time (in seconds) a transaction can remain unconfirmed_ I would expect a `>` here.

matteojug (dismissed):

emily_cron/app/services/deposit_processor.py
matteojug: Approving since as long as we are not too aggressive with `MAX_UNCONFIRMED_TIME` it should be fine even if mess up the clock. My main concern was just to ensure we were comparing apples to apples, in some extreme sense. But I don't think there are strange rules constraining the stacks blocks timestamp that could make them differ too much from the "real" unix time.

I would leave the convo open until the next approval just in case someone has something else to say about it, otherwise please feel free to resolve this as soon as it's a merge blocker.

emily_cron/app/services/deposit_processor.py
```diff
@@ -203,6 +207,55 @@ def process_expired_locktime(
...
+        current_time = int(datetime.now().timestamp())
+
+        long_pending_txs = []
+        for tx in enriched_deposits:
+            # Only check pending transactions
+            if tx.status != RequestStatus.PENDING.value:
+                continue
+            # that we can't find via the mempool API (it might have been dropped)
+            if tx.in_mempool:
+                continue
```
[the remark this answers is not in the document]
djordon: I think that deposits that are in the mempool for too long should also have their own status besides FAILED (or PENDING), just like deposits that have been RBF'ed should be marked differently as well. We can do all of that in a follow-up PR though.

Does the above make sense? I might have missed something that makes skipping transactions in the mempool okay.

emily_cron/test/test_deposit_processor.py
```diff
@@ -542,12 +572,172 @@ def mock_from_info_side_effect(deposit, additional_info):
...
+
+        deposits = [edge_case]
+
+        updates = self.processor.process_long_pending(deposits)
+
+        # With MAX_UNCONFIRMED_TIME=1 hour, this should be marked as long pending
+        self.assertEqual(
+            len(updates), 1, "Transaction should be marked as failed with reduced timeout"
+        )
+        self.assertEqual(updates[0].bitcoin_txid, "edge_case")
```
djordon, line 665: Can we check the status here as well.

emily_cron/app/services/deposit_processor.py
```diff
@@ -203,6 +207,55 @@ def process_expired_locktime(
...
+
+            try:
+                # and have been pending for too long
+                if current_time - tx.deposit_last_update > settings.MAX_UNCONFIRMED_TIME:
+                    long_pending_txs.append(tx)
+            except (RequestException, ValueError, JSONDecodeError) as e:
+                logger.warning(
+                    f"Could not check pending status for deposit {tx.bitcoin_txid} due to "
+                    f"API error fetching block {tx.last_update_block_hash}: {e}. Skipping."
+                )
```
djordon, line 243: Why is this in a `try-except` block?

emily_cron/app/services/deposit_processor.py
djordon: I'm okay with using the local time. I think ubuntu installs and runs NTP (or something newer) that keeps the systems time in sync.

But, I'm not sure if we are using ubuntu, or if our OS is configured to do this.

fabergat: we are using ubuntu and the machines are in sync, ofc we can have a skew of seconds but I don't except more than this

emily_cron/app/services/deposit_processor.py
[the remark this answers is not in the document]
djordon: Ahh I see. I think I misread the opening comment. I read
> ... without being confirmed in the mempool ...

as
> ... without being confirmed from the mempool ...

which is not what you were saying. Okay, yeah we can handle that case in a separate PR.

emily_cron/app/services/deposit_processor.py
djordon: Ahhh, I see. Would it be better to put LRU caching on the `HiroAPI.get_stacks_block` call instead of on an instance of the `DepositInfo` class?

matteojug (dismissed):

emily_cron/app/services/deposit_processor.py
```diff
@@ -203,6 +207,55 @@ def process_expired_locktime(
...
+
+        Args:
+            enriched_deposits: List of enriched deposit information
+
+        Returns:
+            list[DepositUpdate]: List of deposit updates
+        """
+        updates = []
+
+        current_time = int(datetime.now().timestamp())
```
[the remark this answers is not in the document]
djordon: Oh, good to know. Thank you.

matteojug (approved):

Issue #1493, opened 2025-03-10
[Chore]: Update recommended configurations to ensure signer endpoints are locked down


djordon:
# Chore - Update recommended configurations to ensure signer endpoints are locked down

## 1. Description

All signer endpoints need to be locked down, such that only a trusted source can call them. This is particularly true for the `/new_block` endpoint, which receives events `withdrawal-create` events from the stacks node.

## 2. Why It's Necessary

Allowing an untrusted source to access a signer's endpoint could compromise the system by taking up resources or putting invalid data into the signers database.

## 3. Execution Checklist

- [ ] Make sure our documentation for running a signer has instructions for how to lock down a signer.

fabergat:
@djordon @aldur are we talking about on how to block the ports on the hosting machines?

aldur:
No, this is about removing lines like this one: https://github.com/stacks-network/sbtc/blob/394dc72e5bc31db2e51d88f3f7445a724072fa56/docker/mainnet/docker-compose.yml#L83

Which, on Docker, bind that port to public (since Docker also opens the firewall).

PR #1510, opened 2025-03-13
feat: Update recommended configurations to ensure signer endpoints are locked down

fabergat:
## Description

Update recommended configurations to ensure signer endpoints are locked down

Closes: https://github.com/stacks-network/sbtc/issues/1493

## Changes

- Remove event observer ports from docker compose config

## Checklist:


docker/mainnet/docker-compose.yml
```diff
@@ -80,7 +80,6 @@ services:
       - ./logs:/logs
     ports:
       - "4122:4122" # P2P communication between sBTC signers
-      - "8801:8801" # So that Stacks event-observer can reach the sBTC signer
```
aldur, line 83: https://docs.docker.com/engine/network/#published-ports

Maybe something like this? 127.0.0.1:8080:80 

And an explanation of why we are doing that?

fabergat: so you are saying instead of removing the ports, just expose it on the local network? I am ok, do we need this stuff outside docker?

aldur: It depends on how the network is configured, depends on every sBTC signer config

fabergat: something like this?

docker/mainnet/docker-compose.yml
```diff
@@ -81,6 +81,7 @@ services:
     ports:
       - "4122:4122" # P2P communication between sBTC signers
       - "9184:9184" # Prometheus metrics
+      - "127.0.0.1:8801:8801" # this export is only expose in the local machine for security reason
```
aldur:
```suggestion
      - "127.0.0.1:8801:8801" # this export is only exposed to the Docker host for security reason, NEVER expose it to the internet!
```

docker/testnet/docker-compose.testnet.yml
```diff
@@ -21,6 +21,7 @@ services:
       - ./logs:/logs
     ports:
       - "4122:4122"
+      - "127.0.0.1:8801:8801" # this export is only expose in the local machine for security reason
```
aldur: Same as above

docker/mainnet/docker-compose.yml
fabergat: done

docker/testnet/docker-compose.testnet.yml
fabergat: done

docker/mainnet/docker-compose.yml
```diff
@@ -80,8 +80,8 @@ services:
       - ./logs:/logs
     ports:
       - "4122:4122" # P2P communication between sBTC signers
-      - "8801:8801" # So that Stacks event-observer can reach the sBTC signer
       - "9184:9184" # Prometheus metrics
+      - "127.0.0.1:8801:8801" # this export is only exposed to the Docker host for security reason, NEVER expose it to the internet!
```
djordon, line 84: small nit: typo
```suggestion
      - "127.0.0.1:8801:8801" # This export is only exposed to the Docker host for security reasons. NEVER expose it to the internet!
```
same thing for the comment in the other file.

fabergat: done!

aldur (approved):

djordon (approved):
Looks good ✅

Issue #1504, opened 2025-03-12
[Feature]: Add rolling withdrawal limit amounts to Emily


djordon:
# Feature - Add rolling withdrawal limit amounts to Emily

## 1. Description

Emily needs to inform the signers of the maximum amount that can be withdrawn over some number of blocks. In this case, Emily will inform them of both the amount and the number of blocks.

## 2. Technical Details:

There are two new fields: `rolling_withdrawal_cap` which is an amount in sats, and `rolling_withdrawal_blocks` which is a number of blocks. I think this ticket boils down to the following:
1. Updating the limits table to store another two attributes,
2. Updating the `POST /limits` endpoint to accept the two new fields,
3. Updating the `GET /limits` endpoint to return the new fields,
4. Updating the signers so that they take note of the new fields.

### 2.1 Acceptance Criteria:
   - [ ] We can update the new attributes in Emily.
   - [ ] The signers get the new fields when they ask for the limits.

## 3. Related Issues and Pull Requests (optional):

The original umbrella ticket for adding the caps: https://github.com/stacks-network/sbtc/issues/854.

PR #1513, opened 2025-03-13
feat: signer get rolling withdrawals limits from Emily

Jiloc:
## Description

Closes: #1504

Modified the signer's `Limits` struct to integrate the new withdrawal rolling cap parameters.

## Changes
- Add `rolling_withdrawal_blocks` to Limits
- Add `rolling_withdrawal_cap` to Limits

## Testing Information
- updated the tests 

## Checklist:

### DIFF
--- a/signer/src/context/signer_state.rs
+++ b/signer/src/context/signer_state.rs
@@ -137,6 +137,10 @@ pub struct SbtcLimits {
     per_withdrawal_cap: Option<Amount>,
     /// Represents the maximum amount of sBTC that can currently be minted.
     max_mintable_cap: Option<Amount>,
+    /// Represents the number of blocks over which the rolling withdrawal cap is applied.
+    rolling_withdrawal_blocks: Option<u64>,
+    /// Represents the maximum amount of sBTC that can be withdrawn in the rolling withdrawal blocks window.
+    rolling_withdrawal_cap: Option<Amount>,
 }
 
 impl std::fmt::Display for SbtcLimits {
@@ -156,6 +160,8 @@ impl SbtcLimits {
         per_deposit_minimum: Option<Amount>,
         per_deposit_cap: Option<Amount>,
         per_withdrawal_cap: Option<Amount>,
+        rolling_withdrawal_blocks: Option<u64>,
+        rolling_withdrawal_cap: Option<Amount>,
         max_mintable_cap: Option<Amount>,
     ) -> Self {
         Self {
@@ -163,6 +169,8 @@ impl SbtcLimits {
             per_deposit_minimum,
             per_deposit_cap,
             per_withdrawal_cap,
+            rolling_withdrawal_blocks,
+            rolling_withdrawal_cap,
             max_mintable_cap,
         }
     }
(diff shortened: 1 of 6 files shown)


djordon (dismissed):
Looks good ✅

signer/src/context/signer_state.rs
```diff
@@ -135,6 +135,10 @@ pub struct SbtcLimits {
     per_deposit_cap: Option<Amount>,
     /// Represents the maximum amount of sBTC allowed to be pegged-out per transaction.
     per_withdrawal_cap: Option<Amount>,
+    /// Represents the number of blocks over which the rolling withdrawal cap is applied.
+    rolling_withdrawal_blocks: Option<u64>,
+    /// Represents the maximum amount of sBTC that can be withdrawn in the rolling withdrawal blocks window.
+    rolling_withdrawal_cap: Option<Amount>,
```
matteojug, line 141: nit: update comments with upstream version? Also below in getters

signer/src/context/signer_state.rs
```diff
@@ -174,6 +182,8 @@ impl SbtcLimits {
             Some(Amount::MAX_MONEY),
             Some(Amount::ZERO),
             Some(Amount::ZERO),
+            Some(u64::MAX),
```
matteojug, line 185: The `u64::MAX` scares me a bit (assuming a certain usage within queries), what about using `1` instead? It should have the same effect

signer/src/context/signer_state.rs
```diff
@@ -207,6 +217,16 @@ impl SbtcLimits {
     pub fn max_mintable_cap(&self) -> Amount {
         self.max_mintable_cap.unwrap_or(Amount::MAX_MONEY)
     }
+
+    /// Get the number of blocks over which the rolling withdrawal cap is applied.
+    pub fn rolling_withdrawal_blocks(&self) -> u64 {
+        self.rolling_withdrawal_blocks.unwrap_or(0)
```
matteojug, line 223: Maybe 1 would be a safer default? A 0-width rolling window may be ill defined. Also in `unlimited` below

signer/src/context/signer_state.rs
matteojug: oh right, if we don't have withdrawals then it would reset. Let's keep it as is, we can re-asses later it this cause issues with the query.

signer/src/context/signer_state.rs
matteojug: Talked offline, the rationale is that we want consistent inputs, and if don't have them we want to stop processing withdrawals; so:
 - Emily should validate that you either set both or none of them when updating the limits, and fail if only one of them is set;
 - The signers will stop processing withdrawals (and log it) if they only have one of them (eg, set with manual intervention on db);

So I guess for this PR we want to remove the unwrap for both of them and return the "raw" value.
Then in https://github.com/stacks-network/sbtc/pull/1515 we will add the signer side logic. And I guess in a different PR add the Emily validation.

matteojug (approved):

djordon (approved):

PR #1514, opened 2025-03-13
fix: stop logging the trusted api key on deployment

Jiloc:
## Description

On deployment, the we are logging the settings of the lambda.

https://github.com/stacks-network/sbtc/blob/394dc72e5bc31db2e51d88f3f7445a724072fa56/emily/handler/src/bin/emily-lambda.rs#L21

This PR modifies the Debug implementation of the EmilyContext to not display the api key.

## Changes
- modify Debug implementation of EmilyContext to write each field one by one, but explicitly redacting the trusted api key.

## Checklist:

### DIFF
--- a/emily/handler/src/context.rs
+++ b/emily/handler/src/context.rs
@@ -56,12 +56,27 @@ pub struct EmilyContext {
 /// Implement debug print for the context struct.
 impl fmt::Debug for EmilyContext {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "{}",
-            serde_json::to_string_pretty(self)
-                .expect("Failed to serialize Emily Context in debug print.")
-        )
+        f.debug_struct("Settings")
+            .field("is_local", &self.settings.is_local)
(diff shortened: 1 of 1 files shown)


emily/handler/src/context.rs
```diff
@@ -56,12 +56,27 @@ pub struct EmilyContext {
...
+            )
+            .field("limit_table_name", &self.settings.limit_table_name)
+            .field("default_limits", &self.settings.default_limits)
+            .field("is_mainnet", &self.settings.is_mainnet)
+            .field("version", &self.settings.version)
+            .field(
+                "deployer_address",
+                &self.settings.deployer_address.to_string(),
+            )
+            .field("api_key", &"[REDACTED]")
```
matteojug, line 78: nit: should we just omit the `api_key` line?

matteojug (dismissed):

emily/handler/src/context.rs
Jiloc: I wanted to keep it explicit to leave this property out, in case we add other settings in the future and extend this function. Good point about using the full name though

fabergat (approved):
LGTM

matteojug (approved):

Issue #1516, opened 2025-03-15
[Bug]: Stop coordinator activities when tenure finishes


djordon:
# Bug - Stop coordinator activities when tenure finishes

## 1. Description

When the coordinator's tenure is over, it should stop trying to do its duties. If the signer does not stop it's duties, the signer network could get into a bad state during a periods where there is lots of work to do. In such situations, all coordinator processes will continue to try and complete their remaining work, but doing so is futile since their efforts will be ignored by other signers.

### 1.1 Context & Purpose

If a signer is no longer the coordinator, other signers will likely ignore its messages, so there no use in continuing. Moreover, the process that does the coordinator's duties is single threaded, so tying it up with fruitless activities stops the signer from fulfilling it's duties during when they are the coordinator again.

The specific problem arises in a situation where there are many requests to fulfill, but for simplicity let's assume no new requests coming in. It all goes like this:
1. The coordinator starts their tenure and begins fulfilling deposit and withdrawal requests. 
2. The coordinator can only complete 5 percent of the outstanding work before the arrival of the next bitcoin block. After the end of their tenure, they continue to try to sign bitcoin and stacks transactions. These efforts fail, but each attempt to sign a stacks transaction takes 30 seconds to fail, because that is the timeout.
3. The next coordinator picks up where the previous coordinator left off but can only get through 5 percent of the total work before their tenure ends. It also continues working after their tenure ends.
4. The process continues where a new signer is the coordinator but they cannot finish the outstanding work before their tenure ends. They continue working after their tenure ends.

Eventually, all signers are still trying to fulfill requests, but their efforts are all in vain. Other signers ignore messages if they think the signer's view of the chain tip is incorrect. And each coordinator process is still trying to complete their outstanding work using an old chain tip. The coordinator process only updates their view of chain tip after they end their tenure.

## 2. Technical Details:

The coordinator should occasionally check whether the chain tip that was known at the start of it's tenure is still the canonical bitcoin chain tip and stop if they differ.

### 2.1 Acceptance Criteria:
   - [ ] The signer, when acting as coordinator, stops it's coordinator duties if a new bitcoin block is observed.

PR #1518, opened 2025-03-16
fix: stop coordinator activities when tenure finishes

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1516.

I thought about adding just the bitcoin chain tip block height to the signer state, since that is all that is needed for this PR; plus an atomic integer is nicer. But a `BitcoinBlockRef` is probably be more useful, generally speaking, so here we are.

## Changes

* Add the bitcoin chain tip into the signer state object. Set it in the block observer when a new block is observed.
* End the coordinator's tenure when creating deposit and withdrawal stacks transactions if the current chain tip doesn't match the one that was used at the start of the coordinator's tenure.

## Testing Information

I tested that this solves the problem described in https://github.com/stacks-network/sbtc/issues/1516 in devenv, and it works as it's supposed to.

## Checklist:

### DIFF
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -618,6 +618,18 @@ impl<C: Context, B> BlockObserver<C, B> {
         Ok(())
     }
 
+    /// Update the `SignerState` object with current bitcoin chain tip.
+    async fn update_bitcoin_chain_tip(&self) -> Result<(), Error> {
+        let db = self.context.get_storage();
+        let chain_tip = db
+            .get_bitcoin_canonical_chain_tip_ref()
+            .await?
+            .ok_or(Error::NoChainTip)?;
+
+        self.context.state().set_bitcoin_chain_tip(chain_tip);
+        Ok(())
+    }
+
     /// Update the `SignerState` object with data that is unlikely to
     /// change until the arrival of the next bitcoin block.
     ///
@@ -627,12 +639,16 @@ impl<C: Context, B> BlockObserver<C, B> {
     /// * sBTC limits from Emily.
     /// * The current signer set.
     /// * The current aggregate key.
+    /// * The current bitcoin chain tip.
     async fn update_signer_state(&self, chain_tip: BlockHash) -> Result<(), Error> {
         tracing::info!("loading sbtc limits from Emily");
         self.update_sbtc_limits().await?;
 
         tracing::info!("updating the signer state with the current signer set");
-        self.set_signer_set_and_aggregate_key(chain_tip).await
+        self.set_signer_set_and_aggregate_key(chain_tip).await?;
+
+        tracing::info!("updating the signer state with the current bitcoin chain tip");
+        self.update_bitcoin_chain_tip().await
     }
 
     /// Checks if the latest dkg share is pending and is no longer valid
--- a/signer/src/context/signer_state.rs
+++ b/signer/src/context/signer_state.rs
@@ -11,6 +11,8 @@ use hashbrown::HashSet;
 use libp2p::PeerId;
 
 use crate::keys::PublicKey;
+use crate::storage::model::BitcoinBlockHash;
+use crate::storage::model::BitcoinBlockRef;
 
 /// A struct for holding internal signer state. This struct is served by
 /// the [`SignerContext`] and can be used to cache global state instead of
@@ -23,6 +25,9 @@ pub struct SignerState {
     sbtc_contracts_deployed: AtomicBool,
     sbtc_bitcoin_start_height: AtomicU64,
     is_sbtc_bitcoin_start_height_set: AtomicBool,
+    // The current bitcoin chain tip. This gets updated at the end of the
+    // block observer's duties when it observes a new bitcoin block.
+    bitcoin_chain_tip: RwLock<BitcoinBlockRef>,
 }
 
 impl SignerState {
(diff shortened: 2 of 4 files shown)


signer/src/context/signer_state.rs
```diff
@@ -63,6 +68,24 @@ impl SignerState {
             .replace(aggregate_key);
     }
 
+    /// Get the current bitcoin chain tip.
+    pub fn bitcoin_chain_tip(&self) -> BitcoinBlockRef {
```
setzeus, line 72: NIT: any reason not to add 'get' in the name function here? The setter below has 'set' prepended & I noticed other getters in the this file also prepend with 'get' (such as 'get_current_limits')

signer/src/transaction_coordinator.rs
```diff
@@ -691,6 +691,10 @@ where
             tracing::error!(%error, "could not process deposit response transactions on stacks");
         }
 
+        if &self.context.state().bitcoin_chain_tip() != chain_tip {
```
setzeus, line 694: NIT: I noticed that this function appears four times in this commit but this is the only instance w/o the 'tracing::info!("new bitcoin chain tip, stoping coordinator activities");' statement. Just checking that's intentional? Also not sure where we land on DRY principals in this repo but could this be written once & called during the four times it shows up?

signer/src/context/signer_state.rs
```diff
@@ -120,6 +143,10 @@ impl Default for SignerState {
             sbtc_contracts_deployed: Default::default(),
             sbtc_bitcoin_start_height: Default::default(),
             is_sbtc_bitcoin_start_height_set: Default::default(),
+            bitcoin_chain_tip: RwLock::new(BitcoinBlockRef {
+                block_height: 0,
```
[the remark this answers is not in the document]
djordon: This came about because I did not want to have the `bitcoin_chain_tip` be an `Option<BitcoinBlockRef>`, but that's probably the better thing to do so I'll do that.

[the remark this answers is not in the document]
djordon: Sorry, I didn't answer the question fully above. Zero is the wrong value, but I need to write something here and zero is the least incorrect value. It end's up being okay because we always set it in the block observer, and that always runs before the transaction coordinator. However, we may want to use this in other parts of the codebase, such as the request-decider or the tx-signer and those do not have the same kinds of guarantees.

signer/src/transaction_coordinator.rs
```diff
@@ -799,6 +794,11 @@ where
                 "kind" => "complete-deposit"
             )
             .increment(1);
+
+            if &self.context.state().bitcoin_chain_tip() != chain_tip {
+                tracing::info!("new bitcoin chain tip, stoping coordinator activities");
```
setzeus, line 799: NIT - misspelling 'stopping' here & below.

signer/src/context/signer_state.rs
djordon: Yeah, some do and some don't add the `get_` prefix. We should probably do one and not the other, in this case I prefer it not prefixed because it's fewer words.

signer/src/transaction_coordinator.rs
djordon: Updated in https://github.com/stacks-network/sbtc/pull/1518/commits/6b575547db5f4200dde37769e233097f33513dc6.

signer/src/context/signer_state.rs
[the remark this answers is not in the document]
djordon: Oh, well I updated it to be an `Option` in https://github.com/stacks-network/sbtc/pull/1518/commits/6b575547db5f4200dde37769e233097f33513dc6

signer/src/transaction_coordinator.rs
matteojug: +1 on adding the log here as well: we can deduce if this happened or not given the presence/lack of `no withdrawal stacks transactions to create` since there are no possible errors in between (currently), but an explicit log may be easier also to get stats about how frequently we overlap tenures.

signer/src/transaction_coordinator.rs
```diff
@@ -799,6 +794,11 @@ where
                 "kind" => "complete-deposit"
             )
             .increment(1);
+
+            if &self.context.state().bitcoin_chain_tip() != chain_tip {
+                tracing::info!("new bitcoin chain tip, stopping coordinator activities");
+                return Ok(());
+            }
```
matteojug, line 801: I'm wondering if we should move the check at the beginning of the loop instead. Doing so we can remove the check from `construct_and_sign_stacks_response_transactions`.
If we want to keep them as is, we should probably add another check at the beginning of `construct_and_sign_stacks_response_transactions` to guard also the first deposit request.

djordon: Moved in https://github.com/stacks-network/sbtc/pull/1518/commits/76605d37af16f8a0e1829173c6d47a83bb8174c3

matteojug (approved):

matteojug (approved):

signer/src/transaction_coordinator.rs
djordon: I ended up removing this one and keeping the others in https://github.com/stacks-network/sbtc/pull/1518/commits/76605d37af16f8a0e1829173c6d47a83bb8174c3.

signer/src/block_observer.rs
```diff
@@ -618,6 +618,18 @@ impl<C: Context, B> BlockObserver<C, B> {
...
 
+    /// Update the `SignerState` object with current bitcoin chain tip.
+    async fn update_bitcoin_chain_tip(&self) -> Result<(), Error> {
+        let db = self.context.get_storage();
+        let chain_tip = db
+            .get_bitcoin_canonical_chain_tip_ref()
+            .await?
+            .ok_or(Error::NoChainTip)?;
+
+        self.context.state().set_bitcoin_chain_tip(chain_tip);
```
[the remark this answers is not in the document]
djordon: Yeah, none of the other functions take a context, so I feel that it is better here to follow that behavior.

PR #1523, opened 2025-03-18
feat: add leb128 variable length integer implementation to `sbtc` crate

cylewitruk:
## Description

Break-out from withdrawal id encoding. LEB128 encoding is used to efficiently encode integer values.

## Changes

- Adds a new type `Leb128` and related tests to the `sbtc` crate which can be used to encode/decode and calculate the size of LEB-128 encoded values.

## Testing Information

- Adds a bunch of new tests testing the encoding logic.
- Introduces the `proptest` crate for ensuring maximum coverage.

## Checklist:


sbtc/src/leb128.rs
```diff
@@ -0,0 +1,556 @@
...
+
+    /// Tests error handling for malformed inputs
+    ///
+    /// Ensures robustness against truncated byte sequences and proper error
+    /// response for invalid inputs.
+    #[test_case(&[0x80] => Err(Error::IncompleteSequence); "incomplete byte sequence")]
+    #[test_case(&[0x80, 0x80] => Err(Error::IncompleteSequence); "truncated multi-byte")]
+    #[test_case(&[0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0xFF] => Err(Error::InvalidContinuation) ; "too many continuation bytes")]
+    #[test_case(&[0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x01] => Err(Error::InvalidContinuation) ; "exceeds u64 (11 bytes)")]
+    #[test_case(&[0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x01] => Err(Error::InvalidContinuation) ; "exceeds u64 (12 bytes)")]
```
matteojug, line 271: Can we also add a test for _exceeds u64 (10 bytes)_? eg, `&[255, 255, 255, 255, 255, 255, 255, 255, 255, 1+1]` and `ValueOutOfBounds`

sbtc/src/leb128.rs
```diff
@@ -0,0 +1,556 @@
...
+    /// Tests position advancement in cursor reading
+    #[test]
+    fn test_cursor_position_tracking() {
+        // Create buffer with multiple LEB128 values
+        let buffer = [
+            // Value 1: small (1 byte)
+            0x01, // Value 2: medium (2 bytes)
+            0x80, 0x01, // Value 3: slightly larger (2 bytes)
+            0xFF, 0x01,
+        ];
```
matteojug, line 285: maybe nit: misaligned comments? unless we want them "inline"

sbtc/src/leb128.rs
```diff
@@ -0,0 +1,556 @@
...
+    #[test_case(&[0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0x80, 0xFF] => Err(Error::InvalidContinuation) ; "too many continuation bytes")]
+    #[test_case(&[0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x01] => Err(Error::InvalidContinuation) ; "exceeds u64 (11 bytes)")]
+    #[test_case(&[0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x01] => Err(Error::InvalidContinuation) ; "exceeds u64 (12 bytes)")]
+    fn test_leb128_decode_invalid(bytes: &[u8]) -> Result<(u64, usize), Error> {
+        Leb128::try_decode(bytes)
+    }
+
+    /// Tests position advancement in cursor reading
+    #[test]
+    fn test_cursor_position_tracking() {
```
matteojug, line 278: nit: can we add also a testcase for this where we have some garbage after the three valid ints? eg a continuation byte an nothing else

sbtc/src/leb128.rs
```diff
@@ -0,0 +1,556 @@
...
+    const VALUE_MAX_4_BYTES: u64 = (1 << 28) - 1; // 268,435,455
+    const VALUE_MAX_5_BYTES: u64 = (1 << 35) - 1; // 34,359,738,367
+    const VALUE_MAX_6_BYTES: u64 = (1 << 42) - 1; // 4,398,046,511,103
+    const VALUE_MAX_7_BYTES: u64 = (1 << 49) - 1; // 562,949,953,421,311
+    const VALUE_MAX_8_BYTES: u64 = (1 << 56) - 1; // 72,057,594,037,927,935
+    const VALUE_MAX_9_BYTES: u64 = (1 << 63) - 1; // 9,223,372,036,854,775,807
+    const VALUE_MAX_10_BYTES: u64 = u64::MAX; // 18,446,744,073,709,551,615
+
+    /// Number of payload bits per LEB128 byte (excluding continuation bit)
+    const BITS_PER_BYTE: u32 = 7;
```
matteojug, line 368: nit: why not reusing the super one?

sbtc/src/leb128.rs
```diff
@@ -0,0 +1,556 @@
...
+        ///
+        /// Verifies correct byte count at each size boundary:
+        /// - Maximum values for each byte count
+        /// - Minimum values requiring additional bytes
+        /// - Consistent encoding across all boundaries
+        #[test]
+        fn test_all_compression_boundaries(bytes in 1u32..MAX_BYTES as u32) {
+            // Get boundary values
+            let max_value = max_value_for_bytes(bytes);
+            let min_next_value = if bytes < 10 { min_value_for_next_bytes(bytes) } else { u64::MAX };
```
matteojug, line 541: mmm this doesn't include `bytes = MAX_BYTES`, so the `min_next_value` `if` should be redundant. If instead we want to include it, `max_value_for_bytes` should be fixed to account for that (or be guarded by the same condition).

sbtc/src/leb128.rs
```diff
@@ -0,0 +1,556 @@
...
+            bytes.len(),
+            expected_len,
+            "Value {} should encode to {} bytes",
+            value,
+            expected_len
+        );
+        prop_assert_eq!(Leb128::calculate_size(value), expected_len);
+
+        let (decoded, _) = Leb128::try_decode(&bytes).unwrap();
+        prop_assert_eq!(decoded, value);
```
matteojug, line 410: nit: we could also check that bytes read matches `expected_len`

sbtc/src/leb128.rs
```diff
@@ -0,0 +1,556 @@
...
+        /// Tests that 10-byte values round-trip correctly
+        #[test]
+        fn test_10_byte_values(value in VALUE_MAX_9_BYTES+1..=VALUE_MAX_10_BYTES) {
+            test_value(value, 10).unwrap();
+        }
+
+        /// Tests bit-level encoding structure
+        ///
+        /// Verifies the internal structure of encoded values:
+        /// - Correct continuation bit usage
```
matteojug, line 481: nit: the _Correct continuation bit usage_ doesn't seem to be tested

sbtc/src/leb128.rs
```diff
@@ -0,0 +1,556 @@
...
+            let mut cursor = Cursor::new(encoded.as_slice());
+            let mut decoded = Vec::new();
+
+            while cursor.position() < encoded.len() as u64 {
+                let value = cursor.read_leb128().unwrap();
+                decoded.push(value);
+            }
+
+            // Verify all values were decoded correctly
+            prop_assert_eq!(values, decoded);
```
matteojug, line 528: nit: can we also check that the cursor reached the end of the buffer?

sbtc/src/leb128.rs
matteojug, line 541: nit: can't we replace `min_value_for_next_bytes` with `max_value + 1`? (except for 10 bytes)

sbtc/src/leb128.rs
```diff
@@ -0,0 +1,556 @@
...
+    /// without actually performing the encoding.
+    ///
+    /// ## Parameters
+    /// * `value` - The integer value to measure
+    ///
+    /// ## Returns
+    /// The number of bytes required to encode the value
+    pub fn calculate_size(mut value: u64) -> usize {
+        let mut size = 0;
+        while value >= 0x80 {
```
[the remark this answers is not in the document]
cylewitruk: Yeah I didn't use the mask constant here just because it doesn't make a lot of sense reading it. I'll have a think on it and see if I come up with some good name for a constant or something.

sbtc/src/leb128.rs
```diff
@@ -0,0 +1,556 @@
...
+    /// This encoding achieves optimal compression by using fewer bytes for
+    /// smaller values, which are common in delta encodings and sparse sequences.
+    ///
+    /// ## Parameters
+    /// * `value` - The integer to encode
+    /// * `bytes` - The buffer to append the encoded bytes to
+    pub fn encode_into(mut value: u64, bytes: &mut Vec<u8>) {
+        loop {
+            let mut byte = (value & LOWER_BITS_MASK as u64) as u8;
+            value = value.checked_shr(BITS_PER_BYTE).unwrap_or(0);
```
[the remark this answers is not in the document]
cylewitruk: Right, it can't fail unless `rhs` is >= 64. It was just `value >>= BITS_PER_BYTE;` before making an effort to use checked operations.

sbtc/src/leb128.rs
cylewitruk: Good call, [9e11140](https://github.com/stacks-network/sbtc/pull/1523/commits/9e111402f9cc91a0c87cb6bd58d9699950fba1b0)

sbtc/src/leb128.rs
cylewitruk: Yup: [d6897bc](https://github.com/stacks-network/sbtc/pull/1523/commits/d6897bceb357cd069fac23215d362e71d366ac35)

sbtc/src/leb128.rs
cylewitruk: Good catch, [37d671e](https://github.com/stacks-network/sbtc/pull/1523/commits/37d671ef40e675887b1235a96a88700316027373)

sbtc/src/leb128.rs
cylewitruk: Yup, added in the lower part of [993731d](https://github.com/stacks-network/sbtc/pull/1523/commits/993731d5d51c979c800378a2b1fbddc90211757f) as well

sbtc/src/leb128.rs
cylewitruk: Well, I didn't really trust it 🤔 Nah, probably had this one first, removed in [9e11140](https://github.com/stacks-network/sbtc/pull/1523/commits/9e111402f9cc91a0c87cb6bd58d9699950fba1b0)

Cargo.toml
```diff
@@ -54,6 +54,7 @@ metrics = { version = "0.24.1", default-features = false }
 metrics-exporter-prometheus = { version = "0.16.1", default-features = false, features = ["http-listener"] }
 openssl = { version = "0.10.68", default-features = false, features = ["vendored"] }
 p256k1 = { version = "7.2.2", default-features = false }
+proptest = "1.6.0"
```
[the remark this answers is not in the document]
cylewitruk: Absolutely, [76894a7](https://github.com/stacks-network/sbtc/pull/1523/commits/76894a74c26a6ecd3f60d86d9003db49e39ca22f)

sbtc/src/leb128.rs
cylewitruk: Yeah, so you actually uncovered an interesting thing about `checked_shl(rhs)` -- it doesn't return `None` when the shift itself (would) overflow because it truncates excess bits (apparently I'm not the only one who has [misread this](https://github.com/rust-num/num-traits/issues/57)). It returns `None` when `rhs` itself is >= the target bit width, which will never happen with `u64` and our `BITS_PER_BYTE * (MAX_BYTES - 1)`; when we exceed the bit width, we get `Some(0)` and thus the `ValueOutOfBounds` is never actually triggered.

So, we never overflow, but we also won't get the expected error and we implicitly allow "semantically invalid" 10th bytes. Added explicit checks for the "last" bit in the last byte in [993731d](https://github.com/stacks-network/sbtc/pull/1523/commits/993731d5d51c979c800378a2b1fbddc90211757f) as well as some additional tests around the edge.

sbtc/src/leb128.rs
```diff
@@ -488,7 +516,23 @@ mod proptests {
 
             // Encoding size should never exceed 10 bytes (for u64 maximum)
             prop_assert!(bytes.len() <= 10,
-                "Encoding should be at most 10 bytes for u64");
+                "encoding should be at most 10 bytes for u64");
+
+            // Explicitly verify continuation bits
+            if bytes.len() > 1 {
```
matteojug, line 522: nit: since bytes is non empty (maybe we should add that check) we can remove the if guard here, so that we also check the case of a single byte not having the continuation flag (ie, when `bytes.len() == 1`)

cylewitruk: It's covered by other tests but yeah, no point in explicitly _not_ testing it here [4e637a4](https://github.com/stacks-network/sbtc/pull/1523/commits/4e637a45ef835b0d3ca871e44380875a0918b8ad)

sbtc/src/leb128.rs
cylewitruk: Yeah, that method was a leftover [4e637a4](https://github.com/stacks-network/sbtc/pull/1523/commits/4e637a45ef835b0d3ca871e44380875a0918b8ad)

sbtc/src/leb128.rs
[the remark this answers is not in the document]
cylewitruk: Went with `MULTI_BYTE_THRESHOLD` in [7a3d243](https://github.com/stacks-network/sbtc/pull/1523/commits/7a3d2433fa0ea45bb22b002d94e130968370ed21)

sbtc/src/leb128.rs
```diff
@@ -26,13 +26,35 @@
...
+/// Each byte uses bits 0-6 for data and bit 7 for continuation.
 const LOWER_BITS_MASK: u8 = 0x7F;
+
+/// Flag bit indicating that more bytes follow in the LEB128 sequence.
+/// When this bit is set in a byte, it means the value continues in the next byte.
+/// When clear, it indicates the final byte of the encoded sequence.
 const CONTINUATION_FLAG: u8 = 0x80;
 
+/// Represents the threshold where a value needs more than one byte in
+/// Values from 0-127 (0x7F) fit in a single byte, while values ≥128 (0x80)
```
matteojug, line 53: qq: what's the "policy" on non ascii chars (`≥`) in the codebase? We do have emojis for logs and ascii art diagrams around, but other than that (and a random `¡`) we don't usually use them.

sbtc/src/leb128.rs
matteojug: Interesting, I wasn't expecting that behaviour for a _checked_ operation but I guess it kinda makes sense.

sbtc/src/leb128.rs
cylewitruk: Don't think there's any explicit guidelines/policy; I have `≤≥` on my keyboard so I used them 😅

sbtc/src/leb128.rs
```diff
@@ -0,0 +1,618 @@
...
+const MAX_BYTES: usize = 10;
+
+/// Number of value bits stored in each LEB128 byte.
+/// LEB128 encoding uses 7 bits per byte for actual data, with the 8th bit
+/// serving as a continuation flag.
+const BITS_PER_BYTE: u32 = 7;
+
+/// Maximum number of bits in a u64 value.
+/// Used for bounds checking to prevent overflow during encoding/decoding.
+const MAX_BITS: usize = 64;
```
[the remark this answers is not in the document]
cylewitruk: I wanted to avoid "magic numbers" in the display strings is all

sbtc/src/leb128.rs
cylewitruk: Yeah, neither was I, but yeah it does make sense... really should be made more obvious/clear in the docs, though.

sbtc/src/leb128.rs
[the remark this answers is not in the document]
matteojug: That makes sense, but since this one isn't actually used for bounds checking it seems a bit strange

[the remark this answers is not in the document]
cylewitruk: I remembered that we had `u64::BITS`, so changed to use that in [3308ed0](https://github.com/stacks-network/sbtc/pull/1523/commits/3308ed0776b3c31faf5a171c0bb188024068495a)

matteojug (approved):

cylewitruk:
> FYI, I benchmarked your impl against https://crates.io/crates/leb128 and your impl is ~6% faster in release mode, nice!

Well that's cool, and here I was mostly going for safety and not performance 😅

matteojug (approved):

sbtc/src/leb128.rs
cylewitruk: Resolving this, if anyone has an issue with ≤≥ they're welcome to bring it up

PR #1524, opened 2025-03-18
feat: Improve CI security

fabergat:
## Description

This PR improves the general CI security following security suggestions from the Zizmor tool. 

## Changes

- Add checkout action protection
- Add input sanitize 

## Checklist:

### DIFF
--- a/.github/workflows/cargo-vet.yaml
+++ b/.github/workflows/cargo-vet.yaml
@@ -15,6 +15,8 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
+        with:
+          persist-credentials: false
 
       - name: Install Cargo Vet
         run: cargo install --locked cargo-vet
--- a/.github/workflows/image-build-and-draft-release.yaml
+++ b/.github/workflows/image-build-and-draft-release.yaml
@@ -123,6 +123,12 @@ jobs:
         run: |
           set -euo pipefail
           TAG_NAME=$(printf "${TAG}" | sed 's/[^a-zA-Z0-9_.-]//g')
+
+          if [ "$TAG" != "$TAG_NAME" ]; then
+            echo "Error: TAG_NAME '$TAG_NAME' does not match the original tag '$TAG'."
+            exit 1
+          fi
+
           printf "TAG_NAME=$TAG_NAME" >> "$GITHUB_ENV"
         env:
           TAG: ${{ github.ref_name }}
--- a/.github/workflows/image-build-and-draft-release.yaml
+++ b/.github/workflows/image-build-and-draft-release.yaml
@@ -11,7 +11,7 @@ permissions:
   id-token: write
   contents: write
   attestations: write
-  packages: write
+  packages: none
 
 ## Define which docker arch to build for
 env:
@@ -59,6 +59,8 @@ jobs:
       ## This requires that a release branch exists for the tag.
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
 
       ## if the repo owner is not `stacks-network`, default to a docker-org of the repo owner (i.e. github user id)
       ## this allows forks to run the docker push workflows without having to hardcode a dockerhub org (but it does require docker hub user to match github username)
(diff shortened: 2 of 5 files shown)


.github/workflows/cargo-vet.yaml
```diff
@@ -15,6 +15,8 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
+        with:
```
fdefelici, line 18: Just sharing something on top of my head.

Might it be worth creating a composite action (that configures the checkout with proper persistent credential), to centralize the management and make it reusable over time?

Something like this:


```
//custom-checkout.yml

name: My Custom Checkout
description: ...
inputs:
  repository:
    description: 'Repository to checkout'
    required: false
    default: ${{ github.repository }}

runs:
  using: "composite"
  steps:
    - name: Checkout Repository
      uses: actions/checkout@....
      with:
        persist-credentials: false
        repository: ${{ inputs.repository }}
```

And the used like this:
```
jobs:
  my-job:
    runs-on: ...
    steps:
      - name: My checkout
        uses:  /path/to/custom-checkout
```

What do you think?

fdefelici (changes requested):

.github/workflows/cargo-vet.yaml
fabergat: I am not understanding your point, what we earn doing this? Instead of 
`
      - name: Checkout
        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
        with:
          persist-credentials: false
`

fdefelici: Just centralising in one point the checkout action configuration (like version, persist-credentials) instead of repeat it over and over in current files and future reuse in a new files (so we don't have to remember to apply the persist-credentials to false for instance )

Anyhow this was just for sharing. If we don't think it is valuable we can leave just like that.

fabergat: nothing against it! Just thinking if adding it makes a value, for sure we can enforce the persist-credentials but sometime could be also wrong do it, eg. doing git operations

fdefelici (approved):

.github/workflows/image-build-and-draft-release.yaml
```diff
@@ -117,11 +120,17 @@ jobs:
     steps:
 
       - name: Set TAG_NAME
-        run: echo "TAG_NAME=${{ github.ref_name }}" >> "$GITHUB_ENV"
+        run: |
+          set -euo pipefail
+          TAG_NAME=$(printf "${TAG}" | sed 's/[^a-zA-Z0-9_.-]//g')
```
djordon, line 125: Can we have a check to make sure that only valid tags trigger this workflow to begin with?

fabergat: It is not possible with our plan

fabergat: I added the mod

djordon: I see. The other possibility may be to limit it at the workflow level. Something like putting the expected pattern in tag matching; https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet. This is fine too.

fabergat: I totally agree! Did we define our tag schema? Because if not and we change version structure we have to update it every time

.github/workflows/cargo-vet.yaml
aldur: For the record: with this new org that's exactly what I want, and also what https://github.com/stacks-sbtc/sbtc/issues/1563 requests.

.github/workflows/image-build-and-draft-release.yaml
aldur: This will now be possible given we are on an enterprise plan in the new org 🥳

aldur: We now enforce this: https://docs.github.com/en/enterprise-cloud@latest/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/creating-rulesets-for-a-repository#ensure-tag-names-use-semantic-versioning

fabergat: Really good!!

djordon:
This is included in other streams of work, some already merged, some elsewhere.

Issue #1495, opened 2025-03-10
[Feature]: Implement rolling withdrawal window in Emily


djordon:
# Feature - Implement rolling withdrawal window in Emily

## 1. Description

We probably want to stop users from bridging out if we know that the withdrawal would exceed any of our caps. For that, the bridge needs to know how much has been withdrawn in the last X number of bitcoin blocks. The bridge will need to communicate with Emily to figure this out.

### 1.1 Context & Purpose

We will have at least two withdrawal caps at launch:
1. The maximum amount that a single withdrawal can have.
2. The maximum amount that all withdrawals can have over some number of bitcoin blocks.

Users should at least be notified that their withdrawal request will go unanswered if they submit the request through the bridge.

## 2. Technical Details:

Right now, Emily keeps track of time using Stacks blocks instead of bitcoin blocks, making it a little more difficult to track rolling withdrawn amounts using the existing tables. There are a couple of approaches that we can take here but the one that comes to mind is to store the current bitcoin chain tip block height and use that to help with compuuting the withdrawal requests rolling sum when the bridge asks for the remaining amount that can be withdrawn.

The first tricky question is how to handle withdrawal requests that have not been fulfilled or rejected. This ticket proposes that we treat all created withdrawal requests as accepted until they have been rejected, so that created and accepted withdrawal requests both count toward the rolling sum.

### 2.1 Acceptance Criteria:
   - [ ] Emily can give the bridge information necessary to warn the user about a withdrawal being ignored.

PR #1528, opened 2025-03-19
[feat]: Implement rolling withdrawal window in Emily

MCJOHN974:
## Description

Closes: #1495

## Changes

WIP

I added (at least I think so) bitcoin chain tip and mapping `bitcoin height => first stacks block ancored to this bitcoin block height` to db. Also, wrote a function which returns total withdrawn amount using this info.

TODO:
- [x] add everything needed to keep this new tables in db up to date
- [x] add endpoints to provide results of this function to external consumers
- [ ] fix various bugs and style issues
- [ ] add tests
- [ ] fix some .env files and other things to make this new tables in db work

## Checklist:

### DIFF
--- a/emily/handler/src/database/accessors.rs
+++ b/emily/handler/src/database/accessors.rs
@@ -913,7 +913,20 @@ pub async fn wipe_all_tables(context: &EmilyContext) -> Result<(), Error> {
     wipe_withdrawal_table(context).await?;
     wipe_chainstate_table(context).await?;
     wipe_limit_table(context).await?;
+    wipe_heights_mapping_table(context).await?;
+    wipe_bitcoin_chainstate_table(context).await?;
     Ok(())
+}
+
+#[cfg(feature = "testing")]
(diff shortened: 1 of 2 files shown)


djordon (commented):
Hmmmm, I think we should move in a slightly different direction.

I'm thinking that we keep track of the bitcoin blockchain (the block height and block hash) as part of the `/new_block` handler. Although the data there is primarily about Stacks blocks, the `NewBlockEvent` data sent from the Stacks node also contains bitcoin blockchain data. Doing things in the `/new_block` handler simplifies things quite a bit, since that source is already trusted. So, no need for new endpoints for this first PR, even though we may want it later.

We might be able to store this bitcoin blockchain data in the same table that Emily stores the Stacks blockchain data, but as additional attributes. The only thing that we'd need to look into is whether Emily deletes rows there, and if so, whether we should store bitcoin blockchain data in some other table.

Issue #1491, opened 2025-03-09
[Feature]: Ensure signers cannot update Emily's chainstate


djordon:
# Feature - Ensure signers cannot update Emily's chainstate

## 1. Description

It looks like any signer can add entries into Emily's chainstate, so long as it doesn't lead to a reorg. However, the signers probably shouldn't be able to this. Emily probably won't handle the update when it finally gets valid data from the sidecar.

### 1.1 Context & Purpose

This ticket is just to make sure that things will go smoothly in the event of a malicious signer.

### 2.1 Acceptance Criteria:
   - [ ] The signers cannot update Emily's chainstate/

PR #1530, opened 2025-03-19
feat: feature limit signer ability to update emilys database

Jiloc:
## Description

Closes: #1491

## Changes
- Removed stacks_chaintip from deposit/withdrawal updates. Emily now infers the chaintip from her current view of the chain.
    - Accepted updates: Emily’s chaintip view is irrelevant since these updates are not tied to a specific Stacks block.
    - Rejected withdrawals & confirmed deposits/withdrawals: The chaintip view is critical, as reorgs can remove updates tied to specific block heights. However, since updates now only come from the new_block event from the sidecar, Emily will always process updates with the correct chainstate.
- Exception: CreateWithdrawal updates still send the stacks chain tip because the print event itself contains block_height. If we can ensure block_height always matches the corresponding new_block event, we may consider removing it.
- Status message enum: The issue suggests limiting status_message values via an enum, but this PR does not yet implement that restriction.
     - We lack a comprehensive list of possible status messages, especially for failed transactions where specifying a rejection reason is useful.
     - However, this is no longer a security issue. Previously, a malicious signer could infinitely append chainstates with arbitrary block hashes. Now, they can only update pending requests once, with a max 400KB string (limited by DynamoDB) per entry.

## Testing Information
- Updated existing integration tests
- Removed  `update_withdrawals_updates_chainstate` and `update_deposits_updates_chainstate` tests

## Checklist:

### DIFF
--- a/emily/handler/src/api/models/deposit/requests.rs
+++ b/emily/handler/src/api/models/deposit/requests.rs
@@ -213,13 +213,12 @@ impl UpdateDepositsRequestBody {
             }
         }
 
-        // If there are failed conversion, return an error.
+        // If there are failed conversions, return an error.
         if !failed_txs.is_empty() {
             return Err(ValidationError::DepositsMissingFulfillment(failed_txs).into());
         }
 
-        // Order the updates by order of when they occur so that it's as though we got them in
-        // chronological order.
+        // Sort updates by stacks_block_height to process them in chronological order.
         deposits.sort_by_key(|(_, update)| update.event.stacks_block_height);
 
         Ok(ValidatedUpdateDepositsRequest { deposits })
--- a/emily/handler/src/api/models/withdrawal/requests.rs
+++ b/emily/handler/src/api/models/withdrawal/requests.rs
@@ -135,13 +135,12 @@ impl UpdateWithdrawalsRequestBody {
             }
         }
 
-        // If there are failed conversion, return an error.
+        // If there are failed conversions, return an error.
         if !failed_ids.is_empty() {
             return Err(ValidationError::WithdrawalsMissingFulfillment(failed_ids).into());
         }
 
-        // Order the updates by order of when they occur so that it's as though we got them in
-        // chronological order.
+        // Sort updates by stacks_block_height to process them in chronological order.
         withdrawals.sort_by_key(|(_, update)| update.event.stacks_block_height);
 
         Ok(ValidatedUpdateWithdrawalRequest { withdrawals })


signer/src/transaction_coordinator.rs
```diff
@@ -649,7 +649,7 @@ where
             // but emily deposit still marked as pending.
             self.context
                 .get_emily_client()
-                .accept_deposits(&transaction, &stacks_chain_tip)
+                .accept_deposits(&transaction)
```
djordon, line 652: This has since changed on main, where now we update withdrawals as well.

djordon (commented):
Awesome, this looks good. We just need to update it to match the changes that have landed on main.

djordon:
Just ran the failing test locally and it passes if I comment out the part of the test that checks whether Emily has been updated correctly. So, this probably means that she isn't updated like she was. Not sure where the bug is though.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -3493,9 +3495,9 @@ async fn sign_bitcoin_transaction_withdrawals() {
     )
     .unwrap();
 
-    testing_api::wipe_databases(&emily_client.config().as_testing())
-        .await
-        .unwrap();
+    let testing_config = emily_client.config().as_testing();
```
djordon, line 3498: I would call this `emily_config`, since `testing_config` is kinda generic.

djordon (dismissed):
Looks good ✅ 🚀

djordon (dismissed):

matteojug (approved):
Before merging can you double check that Emily wouldn't complain if signers were to still send the stacks chain tip data?

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -143,6 +183,49 @@ pub struct UpdateDepositsRequestBody {
...
+                    failed_txs.push(format!(
+                        "{}:{}",
+                        update.bitcoin_txid.clone(),
+                        update.bitcoin_tx_output_index
+                    ));
+                }
+            }
+        }
+
+        // If there are failed conversion, return an error.
```
setzeus, line 216: NIT: incorrect grammar here. Can be either 'If there are failed conversion**s**' or 'If there is a failed conversion.'

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -143,6 +183,49 @@ pub struct UpdateDepositsRequestBody {
...
+        if !failed_txs.is_empty() {
+            return Err(ValidationError::DepositsMissingFulfillment(failed_txs).into());
+        }
+
+        // Order the updates by order of when they occur so that it's as though we got them in
```
setzeus, line 221: NIT: 'Order...by order' sounds a bit confusing here? Maybe 'Sort updates by update timestamp' works better?

emily/handler/src/api/models/withdrawal/requests.rs
```diff
@@ -65,10 +63,87 @@ pub struct WithdrawalUpdate {
...
+            match update
+                .clone()
+                .try_into_validated_withdrawal_update(chainstate.clone())
+            {
+                Ok(validated_update) => withdrawals.push((index, validated_update)),
+                Err(_) => failed_ids.push(update.request_id),
+            }
+        }
+
+        // If there are failed conversion, return an error.
```
setzeus, line 138: NIT: same comments as above for this & the 'Order the updates by order' below.

matteojug (approved):

PR #1539, opened 2025-03-24
[feat]: Implement rolling withdrawal window on Emily

MCJOHN974:
## Description

Closes: #1495

## Changes

- added bitcoin height to Emily chainstate db and made it secondary index
- updated `limits/` endpoint, adding new field for available_for_withdrawals
- added functions to populate this field useing new db field

## Testing Information

- added 2 more integration tests
- some old tests also test this

## Checklist:

### DIFF
--- a/emily/handler/tests/integration/limits.rs
+++ b/emily/handler/tests/integration/limits.rs
@@ -451,7 +451,7 @@ async fn test_available_to_withdraw_no_chainstate_in_db_at_target_height() {
         per_deposit_cap: Some(None),
         per_withdrawal_cap: Some(None),
         rolling_withdrawal_blocks: Some(Some(100)),
-        rolling_withdrawal_cap: Some(Some(10000)),
+        rolling_withdrawal_cap: Some(Some(10_000)),
         account_caps: HashMap::new(),
     };
     // Set some chainstates to make set_limits work
@@ -508,7 +508,7 @@ async fn test_available_to_withdraw_success() {
         per_deposit_cap: Some(None),
         per_withdrawal_cap: Some(None),
         rolling_withdrawal_blocks: Some(Some(10)),
-        rolling_withdrawal_cap: Some(Some(10000)),
+        rolling_withdrawal_cap: Some(Some(10_000)),
         account_caps: HashMap::new(),
     };
     // Set some chainstates to make set_limits work
@@ -521,10 +521,10 @@ async fn test_available_to_withdraw_success() {
     assert!(result.is_ok());
 
     // Create chainstates
-    let min_bitcoin_height = 1000000;
-    let max_bitcoin_height = 1000020;
+    let min_bitcoin_height = 1_000_000;
+    let max_bitcoin_height = 1_000_020;
     let stacks_block_per_bitcoin_block = 5;
-    let mut stacks_height = 2000000;
+    let mut stacks_height = 2_000_000;
     let mut chainstates: Vec<_> = Default::default();
 
     for bitcoin_height in min_bitcoin_height..max_bitcoin_height {
@@ -539,16 +539,16 @@ async fn test_available_to_withdraw_success() {
 
     // Create withdrawals
 
-    // bitcoin heights in window: [1000010;1000019] (both sides including)
-    // stacks heights in window: [2000050;2000099] (both sides including)
+    // bitcoin heights in window: [1_000_010;1_000_019] (both sides including)
+    // stacks heights in window: [2_000_050;2_000_099] (both sides including)
 
     // Here we put different amount to withdrawals that should be included in window and to ones that shouldn't.
     // Thus, if total sum is correct, then only correct withdrawals was counted
     for (stacks_height, amount) in [
-        (2000050, 1000),
-        (2000049, 999),
-        (2000099, 1000),
-        (2000070, 1000),
+        (2_000_050, 1000),
+        (2_000_049, 999),
+        (2_000_099, 1000),
+        (2_000_070, 1000),
     ] {
         let request = CreateWithdrawalRequestBody {
             amount,


emily/handler/src/api/handlers/new_block.rs
```diff
@@ -90,6 +96,11 @@ pub async fn new_block(
             block_height: new_block_event.block_height,
         };
 
+        let bitcoin_chaintip = BitcoinBlock {
+            block_hash: new_block_event.burn_block_hash.to_hex(),
```
cylewitruk, line 100: Sanity check: in the signer there's some magic around reversing the byte order of bitcoin block hashes, is this the correct ordering for emily?

emily/handler/src/database/entries/deposit.rs
```diff
@@ -694,6 +694,8 @@ impl ValidatedUpdateDepositsRequest {
             .map(|(_, update)| Chainstate {
                 stacks_block_hash: update.event.stacks_block_hash,
                 stacks_block_height: update.event.stacks_block_height,
+                bitcoin_block_hash: Default::default(),
```
cylewitruk, line 697: qq: what implications does setting this to `default()` have? i would have expected this to be the bitcoin chain tip at the time of event creation, but i'm less familiar with the exact constraints of emily's chainstate management so i'm not really sure here.

Same Q for withdrawals.

emily/handler/src/database/accessors.rs
```diff
@@ -696,6 +696,66 @@ pub async fn set_api_state(context: &EmilyContext, api_state: &ApiStateEntry) ->
...
+    for status in all_statuses_except_failed {
+        let mut withdrawals_for_status =
+            get_all_withdrawal_entries_modified_from_height_with_status(
+                context,
+                status,
+                minimum_stacks_height_in_window,
+                None,
+            )
+            .await?;
+        withdrawals.append(&mut withdrawals_for_status);
```
djordon, line 751: Since we only care about the sum, why not just keep track of the sum of the amounts?

emily/handler/src/database/accessors.rs
```diff
@@ -696,6 +696,66 @@ pub async fn set_api_state(context: &EmilyContext, api_state: &ApiStateEntry) ->
...
+    }
+
+    let total_withdrawn: u64 = withdrawals.iter().map(|withdrawal| withdrawal.amount).sum();
+    let amount_left = rolling_withdrawal_cap.map(|cap| cap - total_withdrawn);
+    amount_left.ok_or(Error::NotFound)
```
djordon, line 756: Looks like we can bail early if `rolling_withdrawal_cap` is None.

djordon (commented):
Not, done with the review yet, just submitting what I have now.

emily/handler/src/database/accessors.rs
```diff
@@ -696,6 +696,66 @@ pub async fn set_api_state(context: &EmilyContext, api_state: &ApiStateEntry) ->
...
+    .iter()
+    .min_by_key(|entry| entry.key.height)
+    .map(|entry| entry.key.height)
+    .ok_or(Error::NotFound)
+}
+
+async fn calculate_sbtc_left_for_withdrawals(
+    context: &EmilyContext,
+    rolling_withdrawal_blocks: Option<u64>,
+    rolling_withdrawal_cap: Option<u64>,
```
djordon, line 721: Actually, looks like these shouldn't be `Option`s at all, since we bail if they are `None`.

emily/handler/src/database/accessors.rs
```diff
@@ -753,8 +814,27 @@ pub async fn get_limits(context: &EmilyContext) -> Result<Limits, Error> {
...
         .collect();
+
+    // Calculate total withdrawn amount.
+    let sbtc_left_for_withdrawals = match calculate_sbtc_left_for_withdrawals(
+        context,
+        global_cap.rolling_withdrawal_blocks,
+        global_cap.rolling_withdrawal_cap,
+    )
+    .await
+    {
```
djordon, line 824: nit: can we format this a little differently. Like
```rust
let fut = calculate_sbtc_left_for_withdrawals(
    context,
    global_cap.rolling_withdrawal_blocks,
    global_cap.rolling_withdrawal_cap,
);
let sbtc_left_for_withdrawals = match fut.await {
    ...
```

emily/handler/src/database/accessors.rs
djordon: Well, we should probably follow whatever is implemented in the signers since it will be less mental overhead. And I believe they treat these variables as a pair, where if they are both `None` then it means no limit. If one is `None` and not the other then that is an error state and it assumes the limit is zero.

emily/handler/src/database/accessors.rs
```diff
@@ -696,6 +696,69 @@ pub async fn set_api_state(context: &EmilyContext, api_state: &ApiStateEntry) ->
...
+}
+
+async fn calculate_sbtc_left_for_withdrawals(
+    context: &EmilyContext,
+    rolling_withdrawal_blocks: Option<u64>,
+    rolling_withdrawal_cap: Option<u64>,
+) -> Result<u64, Error> {
+    if rolling_withdrawal_blocks.is_none() || rolling_withdrawal_cap.is_none() {
+        return Err(Error::NotFound);
+    }
```
djordon, line 725: Let's essentially unwrap these here by matching on `(Some(_), Some(_))`. That way we don't need to unwrap later. Unwrapping is okay because we have this check, but better to not have to do it at all, since seeing unwrap should set off alarm bells.

emily/handler/src/database/accessors.rs
```diff
@@ -696,6 +696,69 @@ pub async fn set_api_state(context: &EmilyContext, api_state: &ApiStateEntry) ->
...
+            context,
+            status,
+            minimum_stacks_height_in_window,
+            None,
+        )
+        .await?;
+        total_withdrawn += withdrawals
+            .iter()
+            .map(|withdrawal| withdrawal.amount)
+            .sum::<u64>();
```
djordon, line 755: You probably want this to be a saturating add. So something like this
```suggestion
        for withdrawal in withdrawals {
            total_withdrawn = total_withdrawn.saturating_add(withdrawal.amount);
        }
```

emily/handler/src/database/accessors.rs
djordon: Also, `Error::NotFound` seems a bit strange as the error variant here, especially since we use the same one for a missing bitcoin chain tip.

emily/cdk/lib/emily-stack.ts
```diff
@@ -337,6 +337,23 @@ export class EmilyStack extends cdk.Stack {
             billingMode: dynamodb.BillingMode.PAY_PER_REQUEST, // On-demand provisioning
             pointInTimeRecovery: pointInTimeRecovery,
         });
+        const byBitcoinHeight: string = "BitcoinHeight";
+        table.addGlobalSecondaryIndex({
+            indexName: byBitcoinHeight,
+            partitionKey: {
+                name: 'BitcoinHeight',
+                type: dynamodb.AttributeType.NUMBER
```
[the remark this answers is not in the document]
fabergat: Not present yes, null no

emily/handler/src/database/entries/chainstate.rs
```diff
@@ -233,3 +240,88 @@ impl PrimaryIndexTrait for SpecialApiStateIndexInner {
...
+
+/// Chainstate table entry key. This is the secondary index key.
+#[derive(Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct ChainstateByBitcoinHeightEntryKey {
+    /// Bitcoin height
+    pub bitcoin_height: u64,
+    /// Stacks height
+    #[serde(rename = "Height")]
+    pub stacks_height: u64,
```
djordon, line 254: nit: just realized that elsewhere these are called `bitcoin_block_height` and `stacks_block_height` in other structs. We should stick to that and use the `rename` thingy to make sure that the names are correct when reading and writing to the database.

emily/handler/tests/integration/limits.rs
```diff
@@ -414,3 +436,135 @@ async fn test_complete_rolling_withdrawal_limit_config_works(
...
+    };
+    // Set some chainstates to make set_limits work
+    let chainstates: Vec<Chainstate> = (0..110)
+        .into_iter()
+        .map(|height| new_test_chainstate(height, height, 0))
+        .collect();
+    let _ = batch_set_chainstates(&configuration, chainstates).await;
+
+    let result = apis::limits_api::set_limits(&configuration, limits.clone()).await;
+    assert!(result.is_ok());
```
djordon, line 463: You probably want to `expect` or `unwrap` here, so that you can see the specific error in the panic if it didn't succeed.

emily/handler/tests/integration/limits.rs
```diff
@@ -414,3 +436,135 @@ async fn test_complete_rolling_withdrawal_limit_config_works(
...
+    };
+
+    apis::withdrawal_api::create_withdrawal(&configuration, request.clone())
+        .await
+        .expect("Received an error after making a valid create withdrawal request api call.");
+
+    // Get limits and perform assertions
+    let _limits = apis::limits_api::get_limits(&configuration)
+        .await
+        .expect_err("Should fail because necessary chainstates don't present in the db");
```
djordon, line 493: I'm missing something, and don't quite understand why this panics. Didn't we set the chainstate above?

emily/handler/src/database/entries/chainstate.rs
```diff
@@ -233,3 +240,87 @@ impl PrimaryIndexTrait for SpecialApiStateIndexInner {
...
+// Chainstate by bitcoin height --------------------------------------------------------------
+
+/// Chainstate table entry key. This is the secondary index key.
+#[derive(Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct ChainstateByBitcoinHeightEntryKey {
+    /// Bitcoin height
+    pub bitcoin_height: u64,
+    /// Stacks height
+    pub height: u64,
```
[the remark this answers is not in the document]
djordon: Oh right, the secondary index.

emily/handler/src/database/entries/chainstate.rs
djordon: > You mean, we want to use everywhere in the emily only `stacks_block_height` and `bitcoin_block_height` or `stacks_height` and `bitcoin_height` but not a mix?

Well, ideally yes. But for this PR, I mean the new structs that we are adding should call the fields referring to the stacks block height `stacks_block_height` and similarly for the bitcoin block height field.

> If yes I'd say I agree it would be a bit nicier, but we unfortunately don't follow this rule already ... So I think we can keep it as is

I don't quite follow. Doesn't this example show that we should use `stacks_block_height`? Also, assuming that this struct uses some other random name for the field, I don't think that fact supports the conclusion that we should be adding yet another name to refer to the same thing.

emily/handler/src/database/accessors.rs
```diff
@@ -696,6 +696,64 @@ pub async fn set_api_state(context: &EmilyContext, api_state: &ApiStateEntry) ->
...
+    else {
+        // Note that we validate in set_limits that these values are both Some or both None
+        return Ok(None);
+    };
+    let chaintip = get_api_state(context).await?.chaintip();
+    let bitcoin_tip = chaintip.bitcoin_height.ok_or(Error::NotFound)?;
+    let bitcoin_end_block = bitcoin_tip.saturating_sub(rolling_withdrawal_blocks.saturating_sub(1));
+
+    let minimum_stacks_height_in_window =
+        get_oldest_stacks_block_for_bitcoin_block(context, bitcoin_end_block).await?;
```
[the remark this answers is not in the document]
djordon: Well, maybe instead of panicking we can just write a different query. One that is similar to the one already used that looks for the entry with the minimum bitcoin block height that is greater than or equal to the target height (sort by stack block height ascending limit 1).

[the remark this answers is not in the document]
djordon: I'd go custom query. It should be relatively straightforward, the one I posted in https://github.com/stacks-network/sbtc/pull/1539#discussion_r2012933319 should work by using `>=` instead of `=` in the `key_condition_expression`. We might be able to simplify it further with
```rust
let dynamodb_client = &context.dynamodb_client;
let query_output = dynamodb_client
    .query()
    .table_name(&context.settings.chainstate_table_name)
    .index_name("BitcoinBlockHeightIndex")
    .key_condition_expression("BitcoinHeight >= :v")
    .expression_attribute_values(":v", serde_dynamo::to_attribute_value(&bitcoin_height)?)
    .scan_index_forward(true)
    .limit(1)
    .send()
    .await?;
```
I changed the index name, but fingers crossed that it's not busted for some other reason.

emily/handler/tests/integration/limits.rs
djordon: Oh, which logs are you referring to?

emily/handler/src/api/handlers/new_block.rs
djordon: > @djordon, do you have any additional context on this?

I do. A few points:
1. I believe the code at this line is using data generated from the stacks node, which returns bitcoin block hashes in Reverse Byte Order -- also called RPC Byte Order, Network Byte Order, and, I think, Big-Endian Byte Order.
2. I believe the stacks node stores it's block hashes in Reverse Byte Order, which is the same ordering that we see when on mempool.space or through bitcoin core's RPC responses.
3. Bitcoin core stores it's block and transaction hashes in Natural Byte Order -- also called Native Byte Order, Internal Byte Order, and, I think, Little-Endian Byte Order. This is the same format used over bitcoin-cores p2p interface, and it is what we use in the sbtc signers as well.
4. Because the sbtc signers store things in little-endian byte order, and stacks core uses big-endian byte order, the signers reverse the bytes when making their stacks contract calls so that things match up during execution.
5. I sometimes confuse little-endian with big-endian. So in our codebase, I think I named the [`ToLittleEndianOrder`](https://github.com/stacks-network/sbtc/blob/23c310b96847071faeb77b39bd1d253a6edea11d/signer/src/storage/model.rs#L743) trait incorrectly... maybe? Stacks-core has a comment that seems to confuse little-endian with big-endian https://github.com/stacks-network/stacks-core/blob/47a0d911ccfea964b5b7407ee95cafde56ba42d2/stacks-common/src/types/chainstate.rs#L489-L491. I'm using https://learnmeabitcoin.com/technical/general/byte-order/ to help clear this up (and this too https://developer.bitcoin.org/reference/p2p_networking.html).
6. The decision to use one ordering over the other only becomes a little more difficult if Emily ever has bitcoin core as a datasource.

emily/cdk/lib/emily-stack.ts
```diff
@@ -337,6 +337,23 @@ export class EmilyStack extends cdk.Stack {
             billingMode: dynamodb.BillingMode.PAY_PER_REQUEST, // On-demand provisioning
             pointInTimeRecovery: pointInTimeRecovery,
         });
+        const byBitcoinHeight: string = "BitcoinHeight";
+        table.addGlobalSecondaryIndex({
```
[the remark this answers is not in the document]
djordon: I mean the name of the index itself. This name should probably be distinct from other names used within that table.

emily/handler/tests/integration/limits.rs
cylewitruk: I guess I'm a bit confused as well -- I wasn't under the impression that `get_limits()` should ever fail based on chainstate data? But maybe I've missed something.. I thought the limits were 100% statically set values (via API or direct DB setting or otherwise)?

emily/handler/src/database/accessors.rs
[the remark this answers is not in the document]
cylewitruk: My naive opinion based on what's written here is that if the only real risk is cost (and potentially minor performance reductions), and that that risk is relatively low, that it's worth shipping it as-is and addressing this if and when it presents itself as an actual issue.

emily/handler/tests/integration/limits.rs
cylewitruk: > At this point, there are two known cases where get_limits may fail:

I would also include scenarios such as the sidecar itself not functioning properly (a call to some new event observer endpoint we don't handle?), etc.

> The main question is whether the signers should operate under such conditions.

I think the question is two-fold:
1) Should the **bridge** operate under such conditions --> _probably not_,
2) Should the **signers** operate under such conditions --> the signers calculate this value themselves and thus aren't dependent on this new field, they only want the limits configuration.

So it's more the addition of a new failure mode for `get_limits()` for something which the signers don't need that I'm reacting on, which led me to the suggestion of putting this in a separate endpoint intended for use primarily by the bridge. I'm not certain that the entire sBTC protocol should halt due to this parameter not being able to be calculated is all.

cylewitruk: Despite my comments above, this isn't a blocker for me for this PR, we can iterate on it if needed.

cylewitruk (approved):
Largely looks good, just had my comments re: the failure scenarios, but as I said we can iterate on that if it would be needed.

PR #1541, opened 2025-03-24
[feat]: add remaining sBTC to withdraw to limits Emily endpoint

MCJOHN974:
## Description

Closes: #1495

## Changes

- added new field to limits/ endpoint
- added accesor function to populate this field

## Testing Information
WIP. this needs a lot of testing
## Checklist:

### DIFF
--- a/emily/handler/src/api/models/chainstate.rs
+++ b/emily/handler/src/api/models/chainstate.rs
@@ -4,7 +4,20 @@ use serde::{Deserialize, Serialize};
 use utoipa::{ToResponse, ToSchema};
 
 /// Chainstate.
-#[derive(Clone, Default, Debug, Hash, Serialize, Deserialize, ToSchema, ToResponse)]
+#[derive(
+    Clone,
+    Default,
+    Debug,
+    Eq,
+    PartialEq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Serialize,
+    Deserialize,
+    ToSchema,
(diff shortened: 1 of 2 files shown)


emily/handler/src/database/accessors.rs
```diff
@@ -696,13 +696,173 @@ pub async fn set_api_state(context: &EmilyContext, api_state: &ApiStateEntry) ->
...
+            actual_location = mid;
+            break;
+        }
+        if mid_height < height {
+            left_border = mid + 1;
+            continue;
+        }
+        right_border = mid - 1;
+    }
+    // Now we have in actual_location _some_ stacks block ancored to given bitcoin block,
```
setzeus, line 735: NIT: same misspelling as above

emily/handler/src/database/accessors.rs
```diff
@@ -696,13 +696,175 @@ pub async fn set_api_state(context: &EmilyContext, api_state: &ApiStateEntry) ->
 
 // Limits ----------------------------------------------------------------------
 
+/// Returns height of first stacks block ancored to bitcoin block with provided height
+async fn calculate_associated_stacks_block_fast(
```
djordon, line 700: I think that we should implement this using one simple query to DynamoDB where we just get the least stacks height with a given bitcoin anchor height. That query will just be easier to understand that what is here, even if it's a little more expensive.

PR #1542, opened 2025-03-25
set `stacks_block_height` to Withdrawals

Jiloc:
## Description

Previously, the `stacks_block_height` field was incorrectly defaulting to 0 in withdrawal records rather than using the actual block height from the withdrawal event. 

## Changes
- Added stacks_block_height field to withdrawal creation in the handler
Expanded test assertions to verify all withdrawal fields

## Testing Information
- Updated tests to verify that the block height is correctly set in withdrawal records

## Checklist:

### DIFF
--- a/emily/handler/tests/integration/new_block.rs
+++ b/emily/handler/tests/integration/new_block.rs
@@ -120,13 +120,22 @@ async fn test_new_blocks_sends_create_withdrawal_request() {
     // Check that the withdrawal is confirmed
     assert_eq!(withdrawal.amount, withdrawal_event.amount);
     assert!(withdrawal.fulfillment.is_none());
-    assert_eq!(withdrawal.last_update_block_hash, withdrawal_event.block_id.to_hex());
+    assert_eq!(
+        withdrawal.last_update_block_hash,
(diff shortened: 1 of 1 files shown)


emily/handler/tests/integration/new_block.rs
```diff
@@ -118,8 +118,26 @@ async fn test_new_blocks_sends_create_withdrawal_request() {
...
         .expect("Failed to get withdrawal request");
     // Check that the withdrawal is confirmed
-    assert_eq!(withdrawal.status, Status::Pending);
+    assert_eq!(withdrawal.amount, withdrawal_event.amount);
     assert!(withdrawal.fulfillment.is_none());
+    assert_eq!(
+        withdrawal.last_update_block_hash,
+        withdrawal_event.block_id.to_hex()
+    );
+    assert_eq!(withdrawal.last_update_height, 253);
```
setzeus, line 127: Maybe a dumb Q but curious where the '253' magic figure comes from? Is this the default first block when tests work?

matteojug (approved):

cylewitruk (approved):

djordon (approved):

Issue #1529, opened 2025-03-19
[Chore]: add initiate withdrawal tx to get withdrawal by id


hozzjss:
# Chore - initiate withdrawal tx id

## 1. Description

For ux purposes and @aldur purposes i need the tx id for the initiate withdrwal stacks tx

## 2. Why It's Necessary

UX purposes and possibly metrics

@djordon help me fill this out pls

PR #1548, opened 2025-03-26
[feat] add txid to Emily withdrawals

MCJOHN974:
## Description

Closes: #1529

## Changes

- added txid field for emily withdrawals table
- added txid field for emily withdrawal api schemas

## Testing Information

This PR adds pretty stupid thing -- just 1 more field to input jsons, output json, and to db. We already have integration tests checking that withdrawal we put to Emily equals to withdrawal we extract from Emily. Thus just adding new field to structs in this tests should test everything we need

## Checklist:

### DIFF
--- a/.generated-sources/emily/client/rust/private/docs/CreateWithdrawalRequestBody.md
+++ b/.generated-sources/emily/client/rust/private/docs/CreateWithdrawalRequestBody.md
@@ -11,7 +11,7 @@ Name | Type | Description | Notes
 **sender** | **String** | The sender's Stacks principal. | 
 **stacks_block_hash** | **String** | The stacks block hash in which this request id was initiated. | 
 **stacks_block_height** | **u64** | The stacks block hash in which this request id was initiated. | 
-**txid** | **String** | The txid of the stacks transaction that generated this event. | 
+**txid** | **String** | The hex encoded txid of the stacks transaction that generated this event. | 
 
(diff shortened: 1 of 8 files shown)


hozzjss:
Thanks @MCJOHN974 but no withdrawals are not posted to emily from the bridge (THANK GOD)

emily/handler/src/api/models/withdrawal/mod.rs
```diff
@@ -56,6 +56,8 @@ pub struct Withdrawal {
     /// Details about the on chain artifacts that fulfilled the withdrawal.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub fulfillment: Option<Fulfillment>,
+    /// The txid of the stacks transaction that generated this event.
```
djordon, line 59: I would change this (and the others) to be a little more specific.
```suggestion
    /// The hex encoded txid of the stacks transaction that generated this event.
```

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -336,6 +336,7 @@ fn handle_withdrawal_create(
         request_id: event.request_id,
         stacks_block_hash: event.block_id.to_hex(),
         stacks_block_height,
+        txid: event.txid.to_string(),
```
djordon, line 339: nit: We should prefer something like `.to_hex()` since it is clear what is being stored by looking at the code. It doesn't change what actually happens under the hood, since the `std::fmt::Display` implementation uses `.to_hex()`, it's just clearer.

djordon (approved):
Looks good ✅

Issue #1466, opened 2025-03-04
[Bug]: Fix should_store_decisions_for_pending_withdrawal_requests test


MCJOHN974:
# Bug - Fix should_store_decisions_for_pending_withdrawal_requests test

## 1. Description

As #1465 shows, `should_store_decisions_for_pending_withdrawal_requests` breaks if we change seed. We need to investigate why it happens, and fix it

## 2. Why It's Necessary

There is no explanation why this test needs fixed seed and we ok if it fails with another seed

## 3. Execution Checklist

- [ ] investigate why it happens
- [ ] fix it

PR #1559, opened 2025-03-28
feat: signers no longer store transactions in the database

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1557 and closes https://github.com/stacks-network/sbtc/issues/1466.

## Changes

* Drop the `transactions.tx` column in the signer database.
* Remove the `tx` field from the `model::Transaction` struct.
* Remove the `DbRead::get_bitcoin_tx` method, since we cannot implement it with the `transactions.tx` column removed. Also, it was unused.
* Remove the `BitcoinTx` struct.
* Ignore the in-memory version of the signer UTXO tests. This is because the `TestData` struct does not correctly populate the database with `TxPrevouts` or `TxOutputs`. Might be worth fixing at some later point.

## Testing Information

This is kind of a refactor. No new tests, but if the existing tests pass then we should be good. Granted, this PR ignores some tests...

## Checklist:


signer/src/storage/in_memory.rs
```diff
@@ -153,7 +148,32 @@ impl Store {
...
+                                })
+                                .collect(),
+                            output: outputs
+                                .iter()
+                                .map(|outpout| bitcoin::TxOut {
+                                    value: bitcoin::Amount::from_sat(outpout.amount),
+                                    script_pubkey: outpout.script_pubkey.clone().into(),
+                                })
+                                .collect(),
+                        })
```
matteojug, line 176: nit: move to separate fn and reuse it in both places?

signer/src/storage/in_memory.rs
```diff
@@ -237,7 +257,7 @@ impl Store {
         let context_window_end_block = std::iter::successors(first_block, |block| {
             Some(self.bitcoin_blocks.get(&block.parent_hash).unwrap_or(block))
         })
-        .nth(context_window as usize);
+        .nth(context_window as usize - 1);
```
matteojug, line 260: I'm having troubles understanding why this is correct: in `get_pending_withdrawal_requests` given a `context_window` value we end up with that many + 1 blocks in `extended_context_window`, so it seems here we would consider one less block.

signer/src/testing/storage/model.rs
```diff
@@ -323,7 +318,7 @@ impl TestData {
         num_stacks_blocks: usize,
     ) -> Vec<model::StacksBlock> {
         let mut stacks_block: model::StacksBlock = fake::Faker.fake_with_rng(rng);
-        stacks_block.bitcoin_anchor = new_bitcoin_block.parent_hash;
+        stacks_block.bitcoin_anchor = new_bitcoin_block.block_hash;
```
matteojug, line 321: Originally this was the parent hash as the stacks block that will be "included" in `new_bitcoin_block` are actually anchored to its parent. Now, anchoring to the current block, we are in some sense offsetting the stacks chain wrt the bitcoin one. Why do we want this?

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -2132,20 +2126,20 @@ async fn sign_bitcoin_transaction_multiple_locking_keys() {
...
-            FROM sbtc_signer.transactions
+            SELECT script_pubkey
+            FROM sbtc_signer.bitcoin_tx_outputs
             WHERE txid = $1
+              AND output_type = 'signers_output'
             "#,
         )
         .bind(txid.to_byte_array())
-        .fetch_one(db.pool())
+        .fetch_one(ctx.storage.pool())
```
matteojug, line 2138: qq: why not using `db`?

signer/tests/integration/postgres.rs
```diff
@@ -3176,6 +3116,7 @@ async fn should_process_withdrawals() {
     testing::storage::drop_db(store).await;
 }
 
+#[ignore = "some test fixtures need some work before this can be enabled"]
```
matteojug, line 3119: mmm mixed feeling about ignoring the pg tests as well, do we need to?

signer/tests/integration/transaction_coordinator.rs
djordon: Lots of sloppiness here, I copy pasted this and forgot to change some things back.

signer/tests/integration/postgres.rs
djordon: Ahhh! This is me being sloppy. It's not supposed to be here

signer/src/storage/in_memory.rs
djordon: Oh, I was trying stuff out when one of the tests was failing and I forgot to remove this.

signer/src/testing/storage/model.rs
djordon: The nice thing about doing it this way is that each Stacks block has an anchor block that is also in the database. The current code had the issue where the first Stacks block(s) had no known anchor block. So I think this fixes one of our tests (https://github.com/stacks-network/sbtc/issues/1466).

Besides that small benefit, I couldn't think of any downsides.

signer/src/storage/in_memory.rs
djordon: https://github.com/stacks-network/sbtc/pull/1559/commits/1fb10cfe6e912d828f3efacb51b9d99184a67d3a

signer/src/storage/in_memory.rs
djordon: https://github.com/stacks-network/sbtc/pull/1559/commits/5ee090603fc9afaf23011c9ccdc9e851c86da74f

signer/tests/integration/transaction_coordinator.rs
djordon: https://github.com/stacks-network/sbtc/pull/1559/commits/746a980b1c12269d97740d518ae077321703a83f

signer/tests/integration/postgres.rs
djordon: https://github.com/stacks-network/sbtc/pull/1559/commits/a408682acea83df37b859bc68179a01cf19a666a

signer/tests/integration/postgres.rs
```diff
@@ -5142,6 +5070,39 @@ async fn verification_status_one_way_street() {
...
+            )
+            .await
+            .unwrap();
+
+        if in_canonical_stacks_blockchain {
+            return block.block_hash;
+        }
+    }
+
+    stacks_chain_tip
```
matteojug, line 5103: qq: I'm wondering if returning `stacks_chain_tip` if no block is found could be an "unexpected" result, vs just panicking. If your test setup was correct you shouldn't (I think?) get here.

matteojug (approved):

signer/src/testing/request_decider.rs
```diff
@@ -231,7 +231,7 @@ where
     /// Assert that the transaction signer will make and store decisions
     /// for pending withdraw requests.
     pub async fn assert_should_store_decisions_for_pending_withdrawal_requests(self) {
-        let mut rng = rand::rngs::StdRng::seed_from_u64(46);
+        let mut rng = rand::rngs::StdRng::seed_from_u64(6);
```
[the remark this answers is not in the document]
cylewitruk: Nice catch -- I think we should investigate this before merging so we at least get a ticket/bug report if not a fix.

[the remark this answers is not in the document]
djordon: Ahh, this worked when I originally opened the PR and I tried it against a bunch of different seeds. Specifically was probably partially fixed by f4a455860070e28ff715100280e9bf8f82fa9258 and then broken again with https://github.com/stacks-sbtc/sbtc/commit/5ee090603fc9afaf23011c9ccdc9e851c86da74f (undoing the minus 1). Reverting https://github.com/stacks-sbtc/sbtc/commit/5ee090603fc9afaf23011c9ccdc9e851c86da74f will fix it. Now I need to check if reverting is desirable.

signer/tests/integration/postgres.rs
djordon: Yeah, I wasn't sure if panicking was the right thing to do, but having thought about this after a brief hiatus, it's better to panic since you probably expect there to be a Stacks block for the given bitcoin block. And if not you may need something other than the chain tip anyway. Updated in https://github.com/stacks-sbtc/sbtc/pull/1559/commits/1c849153462c792b20b2d0c1dea4ec7e62412baf.

signer/src/testing/request_decider.rs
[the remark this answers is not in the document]
djordon: Okay, https://github.com/stacks-sbtc/sbtc/commit/5ee090603fc9afaf23011c9ccdc9e851c86da74f is correct no need to revert it. I finally fixed the test in https://github.com/stacks-sbtc/sbtc/pull/1559/commits/b3e51577227508afb3438ab8beab44acd9fbcd55, which needs to look back the correct number of blocks.

signer/src/testing/request_decider.rs
```diff
@@ -231,7 +232,7 @@ where
     /// Assert that the transaction signer will make and store decisions
     /// for pending withdraw requests.
     pub async fn assert_should_store_decisions_for_pending_withdrawal_requests(self) {
-        let mut rng = rand::rngs::StdRng::seed_from_u64(46);
+        let mut rng = OsRng;
```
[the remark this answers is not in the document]
djordon: Yeah I think I had the print there at some point, but I removed it since I figured it will all get changed soon with https://github.com/stacks-sbtc/sbtc/pull/1578.

Issue #1546, opened 2025-03-26
[Feature]: Simplify deposit request processing in block observer


djordon:
# Feature - Simplify deposit request processing in block observer

## 1. Description

Right now we store consensus encoded bytes for all sBTC related transactions in the database. This includes signer sweep transaction, donations, and deposit requests, on bitcoin and contract calls into the sbtc smart contracts on stacks. The actual encoded bytes are never used, and they just take up space. Better to drop the column and not actually store them.

If it turns out that we need the raw bitcoin or stacks transactions for some reason in the future, then we can fetch it from the respective node.

Moreover, when processing deposit requests, we do so in a way that can lead to memory exhaustion in the signers under certain scenarios. That whole flow can be simplified in a way that avoids that issue.

## 2. Technical Details:

* Drop the `transactions.tx` column from the database.
* Remove the `tx` field from the `model::Transaction`.
* Change the logic of deposit request processing in `block_observer.rs` to be more streamlined.  Currently we go from `&[CreateDepositRequest]` to `Vec<Deposit>` to `Vec<(model::DepositRequest, model::Transaction)>`, which is less than ideal. Instead we should yield a `model::DepositRequest` and `model::Transaction` for each `CreateDepositRequest`. That way we bound the amount of memory that is used when processing a collection of deposit requests.

### 2.1 Acceptance Criteria:
   - [x] #1558
   - [x] https://github.com/stacks-network/sbtc/issues/1557

PR #1560, opened 2025-03-28
refactor: simplify deposit request processing in block the observer

djordon:
## Description

Closes https://github.com/stacks-network/sbtc/issues/1558 and closes https://github.com/stacks-network/sbtc/issues/1546.

## Changes

* Simplify deposit request processing in the block observer so that it is very unlikely to OOM.
* Add a helper struct for metrics. This is mainly a cosmetic change so that the logic of the function is easier to see.

## Testing Information

This is a refactor, so if tests pass then we should be good.

## Checklist:


matteojug (approved):

signer/src/block_observer.rs
```diff
@@ -228,24 +229,25 @@ impl<C: Context, B> BlockObserver<C, B> {
...
-                Ok(None) => "unconfirmed",
-                Err(_) => "failed",
+            Metrics::increment_deposit_total(&deposit);
+            let Ok(Some(deposit)) = deposit else { continue };
+
+            self.process_bitcoin_blocks_until(deposit.tx_info.block_hash)
+                .await?;
+
+            let tx = model::Transaction {
+                txid: deposit.tx_info.txid.to_byte_array(),
```
setzeus, line 239: Just double-checking but this doesn't need to be reversed / in LE format correct?

signer/src/metrics.rs
```diff
@@ -52,6 +55,24 @@ impl From<Metrics> for metrics::KeyName {
     }
 }
 
+impl Metrics {
+    /// Increment the deposit request counter for incoming deposit requests
+    pub fn increment_deposit_total(deposit: &Result<Option<Deposit>, Error>) {
+        let deposit_status = match deposit {
+            Ok(Some(_)) => "success",
+            Ok(None) => "unconfirmed",
```
setzeus, line 63: Curiosity here but if marked as 'unconfirmed' is there a queue setup to revisit the deposit / update the metric?

signer/src/block_observer.rs
djordon: Good question. This doesn't need to be reversed. Transaction IDs and bitcoin block hashes are written to the database in the same way that bitcoin core does. Also, this is the same code that was there before, I just moved it to a different spot.

signer/src/metrics.rs
djordon: We will attempt to reprocess the deposit request if it is unconfirmed for as long as Emily has it in a Pending state. The signers just reach out to Emily for all Pending and Accepted deposit requests every bitcoin block.

As for the metric, there is nothing to update it when a deposit request goes from unconfirmed to confirmed. That makes it less useful than it could be but it's still better than nothing. Maybe in the future we can add some LRU cache that tries to prevent double counting in this metric.

MCJOHN974 (dismissed):

matteojug (approved):

MCJOHN974 (approved):

PR #1564, opened 2025-03-31
fix: Allow re-starting devenv while offline

aldur:
## Description

While offline, I found out that `devenv` wouldn't start because we need connectivity to `apt-install` a few packages.
This PR proposes a poor-man's version of those packages (`jq` and `devenv`) that relies on binaries already including within the container image.

It's probably poor (if the image updates, this will break). But then, even doing `apt-get install` at container startup was poor :)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -72,10 +72,9 @@ x-stacks-signer: &stacks-signer
     - /bin/bash
     - -c
     - |
-      apt-get update && apt-get install -y gettext-base
       cd /data/
       set -e
-      envsubst < config.toml.in > config.toml
+      perl -pe 's/\$\{?([A-Za-z_][A-Za-z0-9_]*)\}?/$$ENV{$1}/ge' < config.toml.in > config.toml
       exec stacks-signer run --config config.toml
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -303,21 +302,20 @@ services:
...
       - |
-        apt-get update && apt-get install -y jq
         set -e
         trap "exit" INT TERM
         trap "kill 0" EXIT
         bitcoin-cli -rpcconnect=bitcoin -rpcwait getmininginfo
-        bitcoin-cli -rpcconnect=bitcoin -named createwallet wallet_name=main descriptors=false
-        bitcoin-cli -rpcconnect=bitcoin -named createwallet wallet_name=depositor descriptors=true
+        bitcoin-cli -rpcconnect=bitcoin -named createwallet wallet_name=main descriptors=false || true
+        bitcoin-cli -rpcconnect=bitcoin -named createwallet wallet_name=depositor descriptors=true || true
```
matteojug, line 310: qq: why this change?

cylewitruk (approved):

PR #1565, opened 2025-03-31
fix: No need to bind devenv to all interfaces

aldur:
## Description

Docker has the poor security habit of opening firewalls ports when port-forwarding to `0.0.0.0` (the default if not specifying an interface). 

In our case, we don't need that since the whole dev environment runs locally.


docker/docker-compose.yml
```diff
@@ -574,8 +574,8 @@ services:
     image: mariadb:10.5.21
     container_name: mariadb
     stop_grace_period: 5s
-    ports:
-      - 3306:3306
+    # ports:
+    #   - "127.0.0.1:3306:3306"
```
aldur, line 578: There are a few cases where we expose ports even though they shouldn't be required (because the "client" can access them directly through Docker container networks).

I am not sure if someone is using these ports anyway -- eg to inspect the DB content from the host.

This serves `mempool` and I doubt anyone used it, so I commented it out.

signer/src/config/default.toml
```diff
@@ -58,7 +58,7 @@ endpoints = [
 # Environment: SIGNER_BITCOIN__RPC_ENDPOINTS
 # Environment Example: http://user:pass@seed-1:4122,http://foo:bar@seed-2:4122
 rpc_endpoints = [
-    "http://devnet:devnet@localhost:18443",
+    "http://devnet:devnet@127.0.0.1:18443",
```
aldur, line 61: I had to make this change otherwise on macOS `localhost` would be looking at IPv6 and failing.

We could decide to just stick to IPv4 on the default config.

docker/docker-compose.yml
matteojug: Personally I've never used this one, I would be for removing even the comment.

matteojug (commented):
- qq: was `emily-server` left untouched for a specific reason?
 - Can you also edit `envs/signer-*.env`? Otherwise devenv running the binaries is broken; replacing all the `localhost` works, it may require only a subset of them.

aldur:
> * qq: was `emily-server` left untouched for a specific reason?

I just missed it, fixed it now!

> * Can you also edit `envs/signer-*.env`? Otherwise devenv running the binaries is broken; replacing all the `localhost` works, it may require only a subset of them.

I think I am ok with having `s/localhost/127.0.0.1/g` on the default config and the signer envs. What do you think?

matteojug:
> I think I am ok with having `s/localhost/127.0.0.1/g` on the default config and the signer envs. What do you think?

Sounds good to me, given it's dockerized we are not relying on non-network `localhost` anyway; and I'm fine with ipv4 for everything.

matteojug (dismissed):
Tested devenv (both docker and binaries) and everything worked.

matteojug (approved):

cylewitruk (approved):
Looks good 👍

Issue #1547, opened 2025-03-26
[Feature]: Limit the number of peer initiated connections to a signers


djordon:
# Feature - Limit the number of peer initiated connections to a signers

## 1. Description

It seems like we allow for an unlimited number of inbound connections to the signers. This is less than ideal since it could lead to resource exhaustion. Instead we should use libp2p provided behaviors to limit the number of attempted connections.

Our current flow of limiting unwanted peers happens only after a connection has been established. We should cap the number of concurrent connection attempts too.

## 2. Technical Details:

In libp2p they have the [`ConnectionLimits`](https://docs.rs/libp2p/latest/libp2p/connection_limits/struct.ConnectionLimits.html) struct to limit the number of connected peers and peers trying to establish a connection. We should use it.

### 2.1 Acceptance Criteria:
   - [ ] The signers limit the number of incoming and established connections.

PR #1568, opened 2025-04-01
feat: introduce p2p connection limiting (`#1547`)

cylewitruk:
## Description

Closes: #1547

## Changes
- Makes use of the built-in libp2p `ConnectionLimits` behavior to implement some (hopefully) sane limits on connections.
- Adds a few utility functions used in tests
- Adds enable/disable for both kademlia and autonat (these both affect connections, so needed to disable them for testing).
- Adds a context event for the p2p event loop having started (also used in test)

## Testing Information

This proved to be a bit trickier to test than originally thought. I added a test for the "big one" which is total established connections, but we could probably add a few more to test the other limits (I have seen some of the other errors while fiddling with tests, though). Otherwise, this is tested in `devenv` and things seem to work just fine there, but it would be nice to test on `devnet` to ensure that the limits are indeed sane.

## Checklist:


signer/src/network/libp2p/swarm.rs
```diff
@@ -70,22 +85,66 @@ impl SignerBehavior {
...
+        let limits = connection_limits::ConnectionLimits::default()
+            .with_max_established(Some(max_established))
+            .with_max_established_incoming(Some(max_established_incoming))
+            .with_max_established_per_peer(Some(max_established_per_peer))
+            .with_max_pending_incoming(Some(max_pending_incoming))
+            .with_max_pending_outgoing(Some(max_pending_outgoing));
+
+        connection_limits::Behaviour::new(limits)
+    }
+
```
cylewitruk, line 147: To reviewers: please put extra focus on this block, we need this calculations to be sane with a volatile signing set, but note that these limits are currently still counting on the fact that the signing set is a closed network.

signer/src/network/libp2p/swarm.rs
```diff
@@ -70,22 +85,66 @@ impl SignerBehavior {
...
+        // Allow for `num_signers` _incoming_ connections to be pending at once.
+        // This is conservative, but realistic as we wouldn't expect each signer
+        // to be attempting to open multiple connections to us simultaneously,
+        // even during bootstrapping.
+        let max_pending_incoming = num_signers;
+
+        // Allow for `num_signers` _outgoing_ connections to be pending at once.
+        // This is conservative, but realistic as we wouldn't expect to be dialing
+        // more than `num_signers` signers at once, even during bootstrapping.
+        let max_pending_outgoing = num_signers;
```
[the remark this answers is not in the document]
cylewitruk: Oh, this is only a restriction on libp2p connections for the p2p network, not general network connections, so Emily, database, etc. are unaffected.

signer/src/main.rs
```diff
@@ -227,19 +227,36 @@ async fn run_shutdown_signal_watcher(ctx: impl Context) -> Result<(), Error> {
...
+        .bootstrap_signing_set()
+        .len()
+        .try_into()
+        .unwrap_or(signer::MAX_KEYS);
+
+    // This is how many seconds the swarm will wait before attempting to
+    // bootstrap (i.e. connect to other peers). Three seconds is a sane default
+    // value, giving the swarm a few seconds to start up and bind listener(s)
+    // before proceeding.
+    const INITIAL_BOOTSTRAP_DELAY_SECS: u64 = 3;
```
matteojug, line 249: nit/qq: should we move this const in a different place for visibility?

signer/src/network/libp2p/swarm.rs
```diff
@@ -228,19 +314,31 @@ impl<'a> SignerSwarmBuilder<'a> {
...
         };
-        let behavior = SignerBehavior::new(keypair.clone(), swarm_config)?;
+        let behavior = SignerBehavior::new(keypair.clone(), behavior_config)?;
+
+        // Protocol configuration
         let noise =
             noise::Config::new(&keypair).map_err(|e| SignerSwarmError::LibP2P(Box::new(e)))?;
         let yamux = yamux::Config::default();
-        let swarm_config = libp2p::swarm::Config::with_tokio_executor()
-            .with_idle_connection_timeout(Duration::from_secs(60));
```
matteojug, line 241: qq: is the 10s default enough for us?

signer/src/network/libp2p/swarm.rs
```diff
@@ -228,19 +314,31 @@ impl<'a> SignerSwarmBuilder<'a> {
...
         let tcp_config = tcp::Config::default().nodelay(true);
 
+        // This is hard-coded to a non-zero u8 value and will never fail.
+        #[allow(clippy::unwrap_in_result)]
```
matteojug, line 334: nit: is this `allow` needed?

signer/src/network/libp2p/swarm.rs
```diff
@@ -228,19 +314,31 @@ impl<'a> SignerSwarmBuilder<'a> {
...
+        let dial_concurrency_factor = NonZeroU8::new(4)
+            .ok_or_else(|| SignerSwarmError::Generic("failed to create dial concurrency factor"))?;
```
matteojug, line 336: qq: why do we want to use 4 instead of the default 8?

signer/src/network/libp2p/swarm.rs
```diff
@@ -70,22 +85,66 @@ impl SignerBehavior {
...
+        // for the local signer. This is used to calculate the connection limits.
+        // Note: We use `max(2)` here to ensure that we always have limits
+        // that support at least two signers.
+        let num_signers = config.num_signers.saturating_sub(1).max(2) as u32;
+
+        // Allow `num_signers` * 3 connections to be established at once. This
+        // allows room for one incoming/outgoing connection to each signer, plus
+        // one additional connection per signer as a buffer for protocols such
+        // as autonat.
+        let max_established = num_signers.saturating_mul(3);
```
matteojug, line 115: I'm wondering if we should drop this limit in favour of just using `max_established_per_peer`. Are we gaining something by having this explicit one as well as the per peer one? Maybe also drop `max_established_incoming` for the same reason

signer/src/network/libp2p/swarm.rs
```diff
@@ -70,22 +85,66 @@ impl SignerBehavior {
...
+
+        // Allow `num_signers` * 2 _incoming_ connections to be established at
+        // once. This allows room for one incoming connection from each signer,
+        // plus one additional connection per signer as a buffer for protocols
+        // such as autonat.
+        let max_established_incoming = num_signers.saturating_mul(2);
+
+        // Allow for one incoming, one outgoing and one additional connection per
+        // peer.
+        let max_established_per_peer = 3;
```
matteojug, line 125: This seems the most "interesting" one. What about bumping it up (eg, 10?) to allow for some slack? For no specific reason, it's just that exact limits seems error prone.

signer/src/network/libp2p/swarm.rs
```diff
@@ -70,22 +85,66 @@ impl SignerBehavior {
...
+
+        // Allow for `num_signers` _incoming_ connections to be pending at once.
+        // This is conservative, but realistic as we wouldn't expect each signer
+        // to be attempting to open multiple connections to us simultaneously,
+        // even during bootstrapping.
+        let max_pending_incoming = num_signers;
```
matteojug, line 131: Since this is before checking peed id a low value could be more detrimental ("external" attempts would count against it). Should we instead use some (high) const?

signer/src/network/libp2p/swarm.rs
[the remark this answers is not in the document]
matteojug: I'm wondering why would we want an outgoing limit at all. What are we gaining with it?

signer/src/main.rs
```diff
@@ -227,19 +227,36 @@ async fn run_shutdown_signal_watcher(ctx: impl Context) -> Result<(), Error> {
...
 
+    // Limit the number of signers to the maximum number of signer pubkeys we
+    // can support. Note that this value is used as a base value for swarm
+    // connection limit calculations.
+    let num_signers = config
+        .signer
+        .bootstrap_signing_set()
+        .len()
+        .try_into()
+        .unwrap_or(signer::MAX_KEYS);
```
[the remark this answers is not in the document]
cylewitruk: This should actually probably use the current signer set in the signer state. I'm not a fan of arbitrarily allowing 128*n connections tbh, especially since we know for a fact that it's unrealistic in the near future.

This will ofc need to be updated when we support dynamically changing signer sets, but that's not in the scope of this PR imo.

signer/src/network/libp2p/swarm.rs
cylewitruk: The documentation states that the defaults should be sufficient for most use-cases, and I didn't observe any weirdness in devenv, so the built-in keepalives together with the ping behavior seem to be fine for us.

signer/src/network/libp2p/swarm.rs
cylewitruk: Ah, no not since I changed it to a proper error 👍

signer/src/network/libp2p/swarm.rs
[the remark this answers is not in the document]
cylewitruk: We have the `Autonat` behaviors which can request peers to test an endpoint to help determine NAT status, this ensures that that can't be exploited and used in a DoS attack.

signer/src/network/libp2p/swarm.rs
matteojug: Well, 1 extra connection with an already possible usage (autonat) doesn't leave much of a slack. I'm mostly worried about cases where the connection itself is left "established" (eg by some intermediate proxy) with the peer trying to establish a new one being prevent from by this limit. `10` was definitely a made up number, but maybe a 5 (2 normal + 2 new connection buffer + 1 autonat and co) wouldn't be too bad.

cylewitruk: Mmm but I don't think we should account for that tbh, this doesn't represent a socket (which can be left in some indeterminate state), it's the actual libp2p fully-established connections. A proxy wouldn't be handling the keepalives and so the connection would be dropped after (the now) 10s default..

signer/src/network/libp2p/swarm.rs
cylewitruk: It's largely just being conservative, allowing a signer to simultaneously attempt to open connections to enough peers to become well-connected quickly (remember, they only effectively need 1 peer connection to function) while also trying to limit the effects of any potential connection amplification attacks.

signer/src/network/libp2p/swarm.rs
cylewitruk: My thinking here is that since these limits are working on accepted sockets, transitioning into the "Upgrade" (negotiation) phase where a malicious peer could purposely hold the connection open for the full timeout duration (now 5s), this helps to prevent possible negotiation-phase resource exhaustion attacks (each negotiation consumes CPU and memory).

I'm not against upping this a bit, but otoh it's just as easy for an attacker to open, say, 65k connections as `num_signers` * 5 or 50.

Note that this value doesn't have any affect on the signers' ability to dial _out_, so as long as their seed peer(s) are reachable during bootstrapping then they will still be able to become well-connected.

signer/src/network/libp2p/swarm.rs
cylewitruk: I think that having a general upper-bound on our established connections when we know roughly how many connections we should be expecting (at a maximum) is generally a good practice. If this isn't limited, then resource exhaustion attacks are theoretically possible as we're letting the connections "go further" than they otherwise would.

Why I'm skeptical to remove this (and `max_established_incoming`) is because the documentation explicitly mentions these two parameters for controlling eclipse attacks, where `max_established_per_peer` is `inbound+outbound`, this lets us explicitly control our inbound connection ratio, making me think we should probably even reduce `max_established_incoming`.

signer/src/main.rs
cylewitruk: sure, moved in [ae2e340](https://github.com/stacks-sbtc/sbtc/pull/1568/commits/ae2e340a7f6687ad9992599902df47ef8c96de79)

signer/src/network/libp2p/swarm.rs
cylewitruk: Oh we do do this already in `main.rs`, though I just updated it to use the `SignerState` view instead of the config specifically (as that should be more future-proof) in [00ebd4d](https://github.com/stacks-sbtc/sbtc/pull/1568/commits/00ebd4d55ed8acf882456efc9f770241abf0a4c0), but it is capped at 128 via `MAX_KEYS`.

signer/src/main.rs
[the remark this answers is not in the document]
cylewitruk: I changed this to use `SignerState` current signer set btw, just to be a little more future-proof-ish, [00ebd4d](https://github.com/stacks-sbtc/sbtc/pull/1568/commits/00ebd4d55ed8acf882456efc9f770241abf0a4c0)

signer/src/network/libp2p/event_loop.rs
```diff
@@ -155,19 +161,22 @@ pub async fn run(ctx: &impl Context, swarm: Arc<Mutex<Swarm<SignerBehavior>>>) {
...
-                                    tracing::debug!(%peer_id, %address, "added peer address to kademlia");
-                                }
-                                RoutingUpdate::Failed => {
-                                    tracing::warn!(%peer_id, %address, "failed to add peer address to kademlia");
+                                .as_mut() // Returns `None` if disabled.
+                                .map(|kademlia| kademlia.add_address(&peer_id, address.clone()));
+
+                            if let Some(result) = result {
+                                match result {
+                                    RoutingUpdate::Success => {
```
djordon, line 169: This is fine, but we can also do:
```rust
match result {
    Some(RoutingUpdate::Success) => {...}
    ...
    None => {}
}   
```

signer/src/network/libp2p/swarm.rs
```diff
@@ -22,23 +24,46 @@ use tokio::sync::Mutex;
...
+    pub autonat_client: Toggle<autonat::v2::client::Behaviour<StdRng>>,
+    pub autonat_server: Toggle<autonat::v2::server::Behaviour<StdRng>>,
     pub bootstrap: bootstrap::Behavior,
+    pub connection_limits: connection_limits::Behaviour,
 }
 
 pub struct SignerSwarmConfig {
     pub enable_mdns: bool,
+    pub enable_kademlia: bool,
+    pub enable_autonat: bool,
```
djordon, line 63: Why should we allow autonat to be disabled?

signer/src/network/libp2p/swarm.rs
```diff
@@ -228,25 +333,49 @@ impl<'a> SignerSwarmBuilder<'a> {
...
+        // TCP transport configuration.
         let tcp_config = tcp::Config::default().nodelay(true);
 
+        // Set the dial concurrency factor to limit the number of concurrent dialing
+        // attempts. The default is 8, but we conservatively set it to 4, allowing
+        // a signer to simultaneously attempt to connect to enough peers to become
+        // well-connected while also reducing the impact of any potential connection
+        // amplification attacks.
+        let dial_concurrency_factor = NonZeroU8::new(4)
+            .ok_or_else(|| SignerSwarmError::Generic("failed to create dial concurrency factor"))?;
```
djordon, line 363: I would probably leave this at the defaults, mainly because I don't have a good reason to change it yet, but I suppose 4 is fine. Still, we should make this a constant and put it with the rest of our constants.

Also, I think we can do `NonZeroU8::new(4).unwrap()` in `const` contexts now.

signer/src/main.rs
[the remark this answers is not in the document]
djordon: This comment is so old that the change was already made.

signer/src/context/signer_state.rs
```diff
@@ -527,6 +527,16 @@ impl SignerSet {
             .expect("BUG: Failed to acquire read lock")
             .contains(peer_id)
     }
+
+    /// Gets the number of signers in the current signer set.
+    #[allow(clippy::len_without_is_empty)]
+    pub fn len(&self) -> usize {
```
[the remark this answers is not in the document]
cylewitruk: Yeah I like that better, [e5364c0](https://github.com/stacks-sbtc/sbtc/pull/1568/commits/e5364c051513850d04c9de5086e119b0ca74ca70)

signer/src/main.rs
```diff
@@ -227,19 +233,30 @@ async fn run_shutdown_signal_watcher(ctx: impl Context) -> Result<(), Error> {
...
 
+    // Limit the number of signers to the maximum number of signer pubkeys we
+    // can support. Note that this value is used as a base value for swarm
+    // connection limit calculations.
+    let num_signers = ctx
+        .state()
+        .current_signer_set()
+        .len()
+        .try_into()
+        .unwrap_or(signer::MAX_KEYS);
```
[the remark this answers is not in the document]
cylewitruk: I had thought about it but ulimately landed in that we're only using this for connection limiting and that the responsibility for bailing if it's actually invalid for the signer itself should be elsewhere 🤷

signer/src/network/libp2p/event_loop.rs
```diff
@@ -107,13 +110,16 @@ pub async fn run(ctx: &impl Context, swarm: Arc<Mutex<Swarm<SignerBehavior>>>) {
                             let _ = swarm.disconnect_peer_id(peer_id);
                         } else {
                             tracing::debug!(%peer_id, ?endpoint, "connected to peer");
-                            if endpoint.is_dialer() {
+                            if endpoint.is_dialer() && swarm.behaviour().kademlia.is_enabled() {
```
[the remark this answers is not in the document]
cylewitruk: The change is largely to enable more control when testing as kademlia is a behavior that can use a dynamic number of substreams based on what it's doing, and what it's doing is difficult to control as it has its periodic discovery tasks.

For our purposes, kademlia is always enabled outside of tests right now. But it wouldn't be unrealistic for a signer to want to disable automatic peer discovery and connect to explicit seeds/peers, if they have a very strict network security policy. Then if we wanted to use the DHT, we would do that in a separate NetworkBehavior wrapping Kademlia and with auto-discovery disabled. But neither of those things are currently implemented.

signer/src/network/libp2p/event_loop.rs
cylewitruk: Yeah, I was on the fence. I just found:
```
Some(RoutingUpdate::Success) => ...
Some(RoutingUpdate::Failed) => ...
Some(RoutingUpdate::Pending) => ...
None => {}
```
more annoying to the eye than:
```
RoutingUpdate::Success => ...
RoutingUpdate::Failed => ...
RoutingUpdate::Pending => ...
```
RoutingUpdate::Success

signer/src/main.rs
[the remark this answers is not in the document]
djordon: Ahhh good point. Probably when loading the configuration file or when adding peers into the signer set. Okay, we can change this later after validating things upstream of this.

signer/src/network/libp2p/swarm.rs
djordon: 👍🏽 okay

signer/src/network/libp2p/event_loop.rs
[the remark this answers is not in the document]
djordon: Okay, makes sense. This and the autonat setting is setting us up for the case where an operator wants fine-grained control over network settings. Okay cool.

signer/src/network/libp2p/swarm.rs
```diff
@@ -22,23 +23,46 @@ use tokio::sync::Mutex;
...
+/// * Kademlia: 1-3 streams during active lookups, each query can use its own stream
+/// * AutoNAT: 2 streams (one for client, one for server operations)
+/// * Identify: 1 stream for peer identification
+/// * Ping: 1 stream for keepalive pings
+const MAX_SUBSTREAMS_PER_CONNECTION: usize = 20;
+
+/// The maximum time to wait for a connection negotiation to complete. This is
+/// used to prevent potentially malicious peers from being able to hold a
+/// connection in a pending state for all too long, preventing legitimate peers
+/// from connecting. The timeout is set to 5 seconds, which should be sufficient
```
djordon, line 39:
```suggestion
/// from connecting. The timeout is set to 10 seconds, which should be sufficient
```

djordon (dismissed):
Looks good

signer/src/network/libp2p/swarm.rs
cylewitruk: [2912eca](https://github.com/stacks-sbtc/sbtc/pull/1568/commits/2912eca651a99be70242b43a3933d7abdbb736ed)

MCJOHN974 (approved):
~~upset that can't say "you can use `get_rng()` now!"~~ LGTM!

matteojug (approved):

PR #1572, opened 2025-04-03
feat: use `testcontainers` for integration tests

cylewitruk:
## Description

Supersedes: #1223

## Changes

- Introduces the `testcontainers` crate and implements images for bitcoin core and emily & co. I stuffed these into a separate crate `sbtc-docker-testing` at the moment, I'm on the fence if we should bake it into the `sbtc` crate or not...
- Refactors the `sbtc::testing::regtest` module a bit to get rid of `static` references (it made sense when there was only a single `bitcoind`...).
- Refactors [almost] all integration tests.
- Includes some refactoring of the ZMQ tests due to trying to figure out what was going on with them. They ended up in a "better/more robust" state than they were in to begin with, so I left them the way they are.
- Adds a `nextest.toml` and updates the `Makefile` commands to use it. This lets us specify specific test filters that nextest will use to group tests, run certain tests serially, using more/less threads, etc.
  - Configuration file documentation [here](https://nexte.st/docs/configuration/)
  - Test groups and mutual exclusion docs [here](https://nexte.st/docs/configuration/test-groups/)
- Updated CI a little

## Testing Information

- Make sure you're running the latest `nextest` (`cargo install nextest --locked`) as some of the config file options are only available in newer versions.
- Otherwise, just use the normal `make` testing commands as usual.

P.S. still working on getting CI to work properly

## Checklist:


docker-testing/src/images/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+            match tokio::net::TcpStream::connect(&endpoint).await {
+                Ok(_) => break,
+                Err(_) => {
+                    tokio::time::sleep(Duration::from_millis(10)).await;
+                }
+            }
+        }
+    })
+    .await
+    .expect("ZMQ endpooint not accessible within allotted timeout");
```
moodmosaic, line 61:
```suggestion
    .expect("ZMQ endpoint not accessible within allotted timeout");
```

docker-testing/src/images/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+///
+/// # Parameters
+/// * `host` - Host address to connect to (IP or hostname)
+/// * `port` - Port number to connect to
+/// * `timeout` - Maximum duration to keep attempting connections
+///
+/// # Returns
+/// * `Ok(())` if connection was established successfully
+/// * `Err(Error)` if timeout occurred before successful connection
+pub async fn wait_for_tcp_connectivity(host: &str, port: u16, timeout: Duration) {
```
moodmosaic, line 47: Why not use a circuit breaker here (like failsafe-rs or resilient) with exponential back-offs instead of manually handling timeouts?

docker-testing/src/images/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+/// ## Returns
+/// A string formatted as "sbtc-test-{name}-{random_suffix}"
+fn container_name(name: &str) -> String {
+    let suffix: String = rand::thread_rng()
+        .sample_iter(&Alphanumeric)
+        .take(12)
+        .map(char::from)
+        .collect();
+
+    format!("sbtc-test-{name}-{suffix}")
```
moodmosaic, line 30: Could we include the test name in the container name for easier debugging?  

- Current: `sbtc-test-bitcoind-abc123`  
- Maybe: `sbtc-test-bitcoind-test_get_block_works-abc123`?

In this example, `test_get_block_works` is the test name (can probably be derived).

docker-testing/src/images/mod.rs
cylewitruk: We want the tests to be able to continue ASAP, so exponential back-off isn't really applicable, at least in the current use case. This is almost instantaneous in the current usages as well, so bringing in another dependency for something `tokio` already gives us isn't very appealing.

docker-testing/src/images/mod.rs
cylewitruk: Using a derive macro, sure, but otherwise we're stuck either providing the name as a string, or using a (questionably reliable) hack like `any::type_name<fn()>()` and stripping module prefix/suffixes...

PR #1573, opened 2025-04-03
fix: re-enable `--help` on CLI binaries

aldur:


cylewitruk (approved):
LGTM, this was likely a default feature that got missed since it didn't fail to compile or tests.

cylewitruk:
> Btw some point from me: we are auditing (at least I hope we do so) new versions of crates, but I think we don't do so with new features of crates. However, new features sound much less dangerous

@MCJOHN974 at least in this case we did have this feature previously, but it got disabled when I had missed to add it to the feature list after disabling default features. So it didn't end up pulling in any additional dependencies for this.

Issue #1562, opened 2025-03-28
[Chore]: Disable use of all caches during release builds


cylewitruk:
# Chore - Disable use of all caches during release builds

## 1. Description

As a general security measure, release builds should always be built in isolation, not making use of any caches. We currently use several caching mechanisms to speed up builds, these should be explicitly disabled.

## 2. Why It's Necessary

It's possible for caches to become poisoned as part of an attack, so packages should be freshly pulled and built for release builds.

PR #1576, opened 2025-04-03
chore(ci): disable docker image build cache and update release note template

cylewitruk:
## Description

Closes: #1562 and updates the "default" release note template to be more in-line with the latest notes

## Changes

- Adds the `no-cache: true` option to the `Build and Push` job, which should effectively be the same as using `--no-cache` on the command line. I'm not sure this has any practical effect as the local cache on the VM should be empty, but good to be explicit.
- Updates the default release notes template.

### DIFF
--- a/.github/workflows/image-build-and-draft-release.yaml
+++ b/.github/workflows/image-build-and-draft-release.yaml
@@ -94,6 +94,7 @@ jobs:
           tags: ${{ steps.docker_metadata.outputs.tags }}
           labels: ${{ steps.docker_metadata.outputs.labels }}
           target: ${{ matrix.docker_target }}
+          no-cache: true
           push: true
           build-args: |
             GIT_COMMIT=${{ github.ref_name }}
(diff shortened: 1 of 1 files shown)


cylewitruk:
We may want to revert the `no-cache: true` option as it may slow things down even more since we run several builds...

aldur:
> We may want to revert the `no-cache: true` option as it may slow things down even more since we run several builds...

Right so assuming an empty VM, there's no other caching involved other than the Docker caching. Is that correct? No cargo artifacts being reused from previous builds, etc?

cylewitruk:
> > We may want to revert the `no-cache: true` option as it may slow things down even more since we run several builds...
> 
> Right so assuming an empty VM, there's no other caching involved other than the Docker caching. Is that correct? No cargo artifacts being reused from previous builds, etc?

Yeah, after looking at it. The docker image build is an isolated build in itself, using the code from the workflow checkout.

But it may not matter if this is set or not, actually... since the job build matrix will run each job variation in its own (clean) VM anyway.

aldur:
I think even if there was caching at the "workflow" level it would be acceptable. As long as there's no caching _across_ workflows. This is good news, we might _not_ need to re-release!

cylewitruk:
> I think even if there was caching at the "workflow" level it would be acceptable. As long as there's no caching _across_ workflows. This is good news, we might _not_ need to re-release!

Yeah what I was getting at is that there isn't any caching at the workflow level, so that's why that parameter won't actually have any effect either way. Agree, workflow-level is fine since it's the same "run".

Issue #1567, opened 2025-03-31
[Chore]: Get rid of fixed random seeds in tests


MCJOHN974:
# Chore - Get rid of fixed random seeds in tests

## 1. Description

Currently in many tests we use random with fixed seed. It means, that we always test for few cases, and this cases are same in all runs. This already hit us, we found green CI which turns into red if change the seed at least twice. We should use some more robust testing to avoid such situations in future.

## 2. Why It's Necessary

More robust testing reduces probability of bug merged into main branch.

## 3. Execution Checklist

New test infrastructure should run each test with different outcomes of `random` function. It can be even better if this test cases will be different from run to run, once they are reproducible.

PR #1578, opened 2025-04-03
[chore]: get rid of fixed random in tests

MCJOHN974:
## Description

Closes: #1567

## Changes

- Added function `get_rng` to `signer::testing` which return _seedable_ rng, based on some _unseedable_ one, and prints seed in case of test failure.
- Migrated all usings of seedable rng in tests to using this function
- Disabled now failing tests (and opened issues for them)

## Testing Information

It is refactoring, no new tests needed

## Checklist:

## Note:

Now, with all random in tests not fixed we may have test green first when we merge it, and then red in some other PR or in main. My suggestion to mitigate it -- just disable test and open issue whenever we see red test where we shouldn't see it. (I don't think running tests dozens of times in each CI run worth it. However, it can be okish option for merging queue)


cylewitruk:
So in general I've also considered adding a new proc-macro crate for "things", so I'm not against it in general. I might change the name to `sbtc-macros` or similar instead though so that it could be home to additional utility macros, and put this kind of thing within a `testing` module (with the usual `testing` feature gate).

And just for anyone else reviewing who's unaware, proc macros have to be placed in a separate crate.

As for the concrete code proposal in this PR (as it stands right now), I'm not sure that this benefits us more than just using e.g. `let seed = RngCore::next_u64(&mut OsRng);` (probably wrapped in a helper fn)? I don't think that just using a non-hardcoded seed is the real issue, it's rather a framework around it to run the same test with multiple seeds `n` number of times. That's where `proptest` really excels, as it also performs shrinking and tracks regressions that are found in a file and makes sure to always include those seeds in future runs of the test (to ensure it doesn't regress again).

Maybe take a look at https://pub.npry.dev/proptest_async/tree/src/lib.rs (this is the `proptest_async` crate) as well?

djordon:
I'm usually against adding macros. They are often hard to understand, hard to debug, and thus hard to maintain. This makes onboarding folks who are newer to rust a little more difficult. So my default preference is to try to make due without them. This one is not so difficult, but maybe it's not so bad. Of course, I tend to downplay how well I understand something when making these calls, so I'm still against adding this in this case.

As an alternative, we can just change our tests to either use the `OsRng` directly, or print out the seed in each test. The number of line changes necessary for manually printing out the seed for each test is probably the same or lower than changing each test to take a seed as an input and calling the macro.

signer/src/block_observer.rs
```diff
@@ -778,11 +779,16 @@ mod tests {
...
 
+    use rand::rngs::OsRng;
+
     use super::*;
 
     #[test(tokio::test)]
     async fn should_be_able_to_extract_bitcoin_blocks_given_a_block_header_stream() {
-        let mut rng = rand::rngs::StdRng::seed_from_u64(46);
+        let mut rng = OsRng::default();
+        let seed = OsRng::next_u64(&mut rng);
```
djordon, line 789: I believe these can all be simplified to 
```suggestion
        let seed = OsRng.next_u64();
```

djordon: Alternatively, you can create a test helper function for creating random number generators and printing the seed.

MCJOHN974:
For this PR I think we can just do this switch to different seeds and disable failing tests.

MCJOHN974:
Aaaand, here we are, one more failing test!

djordon (commented):
This looks good, the only change is to remove the new comment above the dummy private keys variable.

signer/src/testing/mod.rs
```diff
@@ -112,3 +116,13 @@ where
...
 
 impl<I, T> IterTestExt<T> for I where I: IntoIterator<Item = T> + Sized {}
+
+/// Returns a seedable rng with random seed. Prints the seed to
+/// stderr so that it can be used to reproduce the test
+pub fn get_rng() -> StdRng {
+    let seed = OsRng.next_u64();
+
+    // Nextest prints stdout and stderr only for failing tests
+    eprintln!("Failed with seed: {seed}");
```
cylewitruk, line 126: I might change this to the following, just because `eprintln`'s do show up if you run the test in other ways (e.g. with `--nocapture` etc.).

```suggestion
    // Nextest prints stderr only for failing tests, so this message
    // will only appear if the test fails (by default).
    eprintln!("Test executed with seed: {seed}");
```

signer/src/testing/mod.rs
```diff
@@ -112,3 +116,13 @@ where
 }
 
 impl<I, T> IterTestExt<T> for I where I: IntoIterator<Item = T> + Sized {}
+
+/// Returns a seedable rng with random seed. Prints the seed to
+/// stderr so that it can be used to reproduce the test
+pub fn get_rng() -> StdRng {
```
cylewitruk, line 122: Just a thought, maybe this should return `impl RngCore + Clone + Send` instead of a concrete type? I'm pretty sure we only need `RngCore` functionality in the tests themselves and it just helps keep things more encapsulated.

signer/src/ecdsa.rs
cylewitruk, line 1: Just a side note, the kinds of tests here are perfect `proptest` candidates.

signer/src/testing/mod.rs
```diff
@@ -112,3 +116,13 @@ where
...
+    let seed = OsRng::default().next_u64();
```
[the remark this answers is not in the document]
MCJOHN974: [49601ad](https://github.com/stacks-sbtc/sbtc/pull/1578/commits/49601ad6342064d68c50c7b64d83fab39bdc5285)

cylewitruk (commented):
This looks largely fine to me, save for a couple of minor suggestions.

I'm a bit worried about ignoring the failing tests and burying them in tickets, though... I'd almost prefer to keep the "working" seeded RNG for those so that they're at least being _run_ by coming updates, and include changing them to `get_rng()` in the tracking tickets instead.

signer/src/testing/mod.rs
MCJOHN974: Actually, we need `impl CryptoRng + RngCore + Clone + Send` here.

`StdRng` implements also `SeedableRng` `Eq` and `PartialEq`. I'm not 100% sure if it will help us, even if we will try to switch to another rng once, but let's do it

signer/src/ecdsa.rs
MCJOHN974: We should open an issue for this!

signer/src/testing/mod.rs
cylewitruk: I wouldn't include any constraints that we don't explicitly use in tests, anyway. But yeah, this method then becomes living documentation of what traits we're actually using as well.

MCJOHN974:
> I'm a bit worried about ignoring the failing tests and burying them in tickets, though... I'd almost prefer to keep the "working" seeded RNG for those so that they're at least being run by coming updates, and include changing them to get_rng() in the tracking tickets instead.


I'm strongly _for_ explicitly ignoring tests. If we keep seed with which they don't fail we will build illusion for ourselves that they are working, while they are actually not. I think we want to highlight both in code and in tickets that some of our tests are failing

MCJOHN974:
> > I'm a bit worried about ignoring the failing tests and burying them in tickets, though... I'd almost prefer to keep the "working" seeded RNG for those so that they're at least being run by coming updates, and include changing them to get_rng() in the tracking tickets instead.
> 
> I'm strongly _for_ explicitly ignoring tests. If we keep seed with which they don't fail we will build illusion for ourselves that they are working, while they are actually not. I think we want to highlight both in code and in tickets that some of our tests are failing

Also, leaving fixed seed can make illusion that aim of this issue to finish migration to new test approach, while actually this issue is about fixing bug in code we have in our main branch. And second feels like much more priority thing

signer/src/testing/mod.rs
MCJOHN974: [fe63eb3](https://github.com/stacks-sbtc/sbtc/pull/1578/commits/fe63eb3f3b1ffae6eb094f312fcb80f6c35aa877)

cylewitruk:
> > > I'm a bit worried about ignoring the failing tests and burying them in tickets, though... I'd almost prefer to keep the "working" seeded RNG for those so that they're at least being run by coming updates, and include changing them to get_rng() in the tracking tickets instead.
> > 
> > 
> > I'm strongly _for_ explicitly ignoring tests. If we keep seed with which they don't fail we will build illusion for ourselves that they are working, while they are actually not. I think we want to highlight both in code and in tickets that some of our tests are failing
> 
> Also, leaving fixed seed can make illusion that aim of this issue to finish migration to new test approach, while actually this issue is about fixing bug in code we have in our main branch. And second feels like much more priority thing

I agree in principle, but in reality tickets have a tendency to hang around for a while and get forgotten (just as well as a handful of our currently ignored tests). Keeping them "alive" with the known-working fixed seed, together with highlighting them with `// TODO(1234): This test is broken with random seeds` etc. is _at least_ still testing an entire sunny-day code-path, which could otherwise get broken by some other changes and go unnoticed.

That said, I'm not sure what the nature of the failures in the newly ignored tests are -- if it's so that the test proves absolutely nothing if it doesn't work with all possible valid inputs then I'd agree that ignoring them is better, but if it's some small, non-unit-tested component which is failing as part of a larger test, then I'd opt for keeping the test "working" and fixing and adding unit tests for the failing component instead.

Not sure what others think 🤷

signer/src/testing/mod.rs
cylewitruk: Nice, looks like you just missed a `cargo fmt` or something based on CI though

MCJOHN974:
> I agree in principle, but in reality tickets have a tendency to hang around for a while and get forgotten (just as well as a handful of our currently ignored tests)

Yea.... Probably it is a call to add some better priority system in our tickets then we have now and put hight priority for such tests.


Argument for testing at least something is better then testing nothing sounds valid...

djordon:
> I'm a bit worried about ignoring the failing tests and burying them in tickets ...
> ...
> I'm strongly for explicitly ignoring tests ...

If you merge main, the failing test might actually pass now. If so then everyone wins.

---

My sense here is that if the test is generally broken then try to fix it pretty quickly. If for some reason fixing it right away is not possible or not urgent then ignoring it and fixing it soon is fine. My success rate with this strategy is probably 40-60%, there is often something way more important than fixing the test and the behavior is often captured by some other test anyway.

If the test is broken for some seeds and not others then there is some utility in leaving it enabled with a working seed. We want to test enabled to catch changes that break things even more. We would still open the ticket and try to get to it though.

In summary, I agree.

djordon:
I lied, the test is still broken 😞;

djordon (dismissed):
Looks good

MCJOHN974:
@cylewitruk @djordon I unignored all this tests

signer/src/request_decider.rs
```diff
@@ -592,7 +592,6 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore = "We track fix of this test in #1466"]
```
cylewitruk: I would still leave a:
```rust
// TODO(#1466): This test is currently using a known-working fixed seed, but is flaky with other seeds.
```

MCJOHN974: [1ca50a5](https://github.com/stacks-sbtc/sbtc/pull/1578/commits/1ca50a5c314a6a0e2e13f62c17301de9e0dce22e)

cylewitruk (approved):
LGTM! ✅

djordon (approved):
Looks good

PR #1579, opened 2025-04-03
feat: build `signer` and `blocklist-client` with `nix`

aldur:
## Description

This PR:

1. Updates the `nix` Rust toolchain runtime to 1.85.1.
2. Makes it possible to build both the sBTC signer and the blocklist client through `nix`.
3. Paves the way, if we choose to, to build both binaries and Docker images that are reproducible.
4. Paves the way, if we choose to, to use `nix` to setup the CI environment (so that we don't need to depend on a bunch of actions).

It also allows running `clippy` on the workspace and a few more things.


cylewitruk:
@aldur I just opened #1582 so you might want to wait with this one and use 1.86.0 once that's merged

cylewitruk:
@aldur it's merging now so you could update this

aldur:
Thanks, rebased on `main`.

MCJOHN974 (approved):
Well, I'm not nix expert so can't do really good review, but at least I don't see how this changes can be harmful, so approving

djordon:
I don't think that we're any closer to getting a review on this than we were when it was opened. Right now, I'm thinking that we close this PR until the team can properly review and maintain it. What do you think.

PR #1581, opened 2025-04-04
Push release Docker images to GHCR instead of DockerHub 

fabergat:
## Description

Push release Docker images to GHCR instead of DockerHub

Closes: https://github.com/stacks-sbtc/sbtc/issues/1571

## Changes

- Remove DockerHub connection
- Add GHCR login
- Fix the release links

## Checklist:

- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #1584, opened 2025-04-07
chore: 4-eyes to create a tag 

fabergat:
## Description

This PR add a GiHub action for the tag creation. This is under manual approval configured in the environment Tags.
The PR requires https://github.com/stacks-sbtc/sbtc/pull/1581 to be approved before.

Manual invocation: 
Screenshot 2025-04-07 at 10 05 47

Closes: https://github.com/stacks-sbtc/sbtc/issues/1570

## Changes

- Add tag creation workflow
- Change image build invocation method

## Checklist:


.github/workflows/create-tags.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+  trigger_release:
+    name: Trigger Release
+    runs-on: ubuntu-latest
+    needs: create_tag
+    if: inputs.release == 'yes'
+    steps:
+      - name: Trigger Release Workflow
+        run: |
+          curl -X POST -H "Accept: application/vnd.github.everest-preview+json" \
```
aldur, line 46: Curios, is this the only way to call a workflow?

.github/workflows/create-tags.yaml
```diff
@@ -0,0 +1,50 @@
...
+        description: "Tag Name (e.g., v1.0.0)"
+        required: true
+      release:
+        description: "Trigger Release? (yes/no)"
+        required: true
+        default: "no"
+
+permissions:
+  contents: write
+  actions: write
```
aldur, line 16: Just to make our lives easier later on, can you add a comment on why we need these permissions?

.github/workflows/create-tags.yaml
```diff
@@ -0,0 +1,50 @@
...
+    inputs:
+      tag_name:
+        description: "Tag Name (e.g., v1.0.0)"
+        required: true
+      release:
+        description: "Trigger Release? (yes/no)"
+        required: true
+        default: "no"
+
+permissions:
```
aldur, line 14: Have you run `zizmor` against this?

.github/workflows/create-tags.yaml
cylewitruk: Hm, yeah I would have thought we could use the gh cli here?

.github/workflows/create-tags.yaml
```diff
@@ -0,0 +1,50 @@
...
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Create Tag
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          script: |
+            github.rest.git.createRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'refs/tags/${{ inputs.tag_name }}',
```
cylewitruk, line 34: Does this validate that the tag is semver-compatible?

.github/workflows/create-tags.yaml
fabergat: yes, the only problem is the TAG that can be expanded as command but if we limit the tag space we are ok

.github/workflows/create-tags.yaml
```diff
@@ -0,0 +1,50 @@
+name: Manual Tag and Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: "Tag Name (e.g., v1.0.0)"
+        required: true
+      release:
```
cylewitruk, line 9: it looks like we could use a `choice` here so that it's a dropdown? Something like:

```
release:
        type: choice
        description: "Trigger Release?"
        default: "No"
        options:
        - No
        - Yes
```

.github/workflows/create-tags.yaml
fabergat: There is another way with the PAT token and using other public actions but not from GH

fabergat: but let me double check

.github/workflows/create-tags.yaml
fabergat: Good idea! 
Screenshot 2025-04-07 at 10 02 05

.github/workflows/create-tags.yaml
aldur: Yes there's a ruleset enforcing semver.

.github/workflows/create-tags.yaml
fabergat: Ok this is the best thing we can do... we can't use workflow dispatch otherwise the release can be triggered manually (also we have a check can be noisy) and here we are not passing secrets

.github/workflows/create-tags.yaml
cylewitruk: I'm just curious how this will "fail" if i type in "asdf" there?

.github/workflows/create-tags.yaml
cylewitruk: But the GH CLI you install earlier should be doing pretty much the same thing under the hood, shouldn't it? Just thinking if that's usable if it's potentially cleaner? But maybe that requires you to run `login` or something..

.github/workflows/create-tags.yaml
fabergat: I think you can't create a tag without following the pattern so probably the action will fail but we must try it

.github/workflows/create-tags.yaml
fabergat: There is only this one: https://cli.github.com/manual/gh_workflow_run

But this a workflow dispatch event not a repository dispatch, if I configure a workflow dispatch means that the build process can be run directly from the UI, can be block by an approval but means two approvals, one for the tag one for the build process... don't know if we wanna make the release optional is not optimal...

cylewitruk: Looks like you updated this a bit in [d28973f](https://github.com/stacks-sbtc/sbtc/pull/1584/commits/d28973f1393fe3c6fc8abd29f7ba504dc25d6a24) ?

.github/workflows/create-tags.yaml
cylewitruk: Yeah let's make sure to try it so we know. We might be able to put some input validation on it otherwise? I'm not sure if it's natively supported or if you'd need to do something like:
```
jobs:
  validate:
    runs-on: ubuntu-latest
    steps:
      - name: Validate Input
        uses: BelmoMusta/inputs-validation-action@v0
        with:
          input-b: ${{ github.event.inputs.input-b }}
          validation-script: |
            input-b:
              type: string
              regex: '^foo\\.bar.*$'
```
(or via just some regex cmd line tool)

Then make the remaining jobs dependent on this one, etc.

.github/workflows/create-tags.yaml
fabergat: yeah I improved if from the curl version

.github/workflows/create-tags.yaml
fabergat: Yeah for now only @aldur can try it I guess (creating a tag) but I can also add a check in the action!

fdefelici: I agree. Considering that it is a user input, It is better to check and sanitize it instead of just relying on the enterprise feature. Possibly we could stop workflow execution earlier and provide also a better error message in case.

.github/workflows/image-build-and-draft-release.yaml
```diff
@@ -150,23 +147,33 @@ jobs:
   
           [Highlights](#highlights) • [Images](#images) • [Upgrade Instructions](#upgrade-instructions)
   
-          ## ✨ Highlights <a id="highlights">
+                    ## ✨ Highlights <a id="highlights">
```
fdefelici, line 150: nit: same remark about extra space done in PR #1581 here https://github.com/stacks-sbtc/sbtc/pull/1581#discussion_r2033231373.

Just reporting because don't know who will win after the merge, or eventually conflict :)

.github/workflows/create-tags.yaml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          script: |
+            const response = await github.request('POST /repos/{owner}/{repo}/dispatches', {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              event_type: 'image_build_and_draft_release',
+              client_payload: {
+                tag_name: '${{ inputs.tag_name }}' // Pass the input value
+              }
+            });
```
fdefelici, line 58: Just exposing a possible alternative:
- set the approvals (`environment` tag) on the `trigger_release` 
- remove approvals from `image_build_and_draft_release.yml` and configure it as "on: workflow_call"
- Like this, we can point to the workflow yaml file path (image_build_and_draft_release.yml) with the directive "uses" passing the required input and eventually the GITHUB_TOKEN

So basically converting the workflow dispatch to a workflow call. 
If it is feasable we could gain:
- the step who launch `image_build_and_draw_release` could become syncronous (currently once triggered, the current workflow ends)
- avoid doing explicit POST call still keeping 4-eyes over the release.
- Bonus: eventually image push or other action done in the `image_build_and_draft_release` can be more testable if not protected by the approvals)

Just an idea, I wanted to share. Let's understand if could worth it

.github/workflows/create-tags.yaml
cylewitruk: I'll close this then as adriano said we could; I think it looks a lot cleaner 👍

.github/workflows/image-build-and-draft-release.yaml
fabergat: fixed

.github/workflows/create-tags.yaml
fabergat: added a check

.github/workflows/create-tags.yaml
fdefelici: > Oh interesting. What do we gain if `.github/workflows/image-build-and-draft-release.yaml` becomes synchronous? 

Well, if it is synchrounous, basically the triggering workflow remanin "active" until the release is finished. Eventually this could be helpfull in terms of action graph, but also if we wanted to do something else after the release is completed (think on eventual check or verification, or whatever). Maybe not something needed now, but open possibilities.

> Also, removing approvals from `image_build_and_draft_release.yml` sounds risky. Perhaps it isn't; I'd like to understand why it isn't though.

Basically, threating that workflow like an utility, and moving the approval upwards in the chain in the release trigger.

> Which has me wondering about a few things about how GitHub actions currently gate things. What happens if I open a PR calling into or modifying `.github/workflows/image-build-and-draft-release.yaml` to run on push to my branch. What happens?

Maybe the same happen with the current implementation, if someone make a PR removing the approval (the environment tag) from `.github/workflows/image-build-and-draft-release.yaml`. Or even creating a brand new workflow that just does fancy things.

If the PR come from a fork, we should be protected by github security mechanism, where on pull_request (from fork) secres.GITHUB_TOKEN is not exposed for instance. So here, the the real gate is the review process.

djordon: 🙏🏽 thank you for the explanation.

Okay, that makes sense. Seems worth doing, especially if we want to build off the work here. I'd vote for making that change in a follow-up PR then.

djordon (approved):
Looks good. I don't see the tag filter/check but I assume that's at the environment level.

.github/workflows/image-build-and-draft-release.yaml
fabergat: don't know why but last commit was not here

.github/workflows/create-tags.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Checkout Repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Create Tag
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          script: |
+            const tag = `${{ inputs.tag_name }}`.replace(/[^a-zA-Z0-9_.-]/g, '');
+            if (tag !== `${{ inputs.tag_name }}`) {
+              core.setFailed("Error: The tag contains invalid characters.");
```
fdefelici, line 36: Do we want just checking for valid characters or enforce the semver tag pattern like this:

```javascript
  //strict
  const semverRegex = /^v?(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-[0-9A-Za-z.-]+)?(?:\+[0-9A-Za-z.-]+)?$/;
  //simplified
  const semverRegex = /^v?\d+\.\d+\.\d+(-[\w.-]+)?(\+[\w.-]+)?$/;
  
  if (!semverRegex.test(tag)) {
    core.setFailed(`Invalid tag format: "${tag}". Expected SemVer like X.Y.Z, vX.Y.Z-note, or X.Y.Z+note`);
```

fdefelici: well, if you think to a layred software, these would be respectively UI checks and Business check, so in this case it wouldn't be a bad duplication :) 

That would be my preference, but I'm opened to do as you propose in the current implementation

fabergat: I can also implement like this, not a problem, I was thinking that in this way is more general and we don't have to update it if want to change the schema, I can use the simplied that give us more freedom and be more restrictive in case we need in the environment

fdefelici (approved):

.github/workflows/create-tags.yaml
cylewitruk: I guess my personal take would be that it doesn't make much of a difference here since the validation failure will happen within the same action, and I get that this is just a sanity-check sanitation prior to making the API request. But I also agree with not relying on an enterprise feature for enforcing it, especially since most of us can't see that setting it becomes very implicit.

Sanity check: have you tested that `github.rest.git.createRef()` returns errors for invalid semvers?

fabergat: Screenshot 2025-04-11 at 14 52 10

I tried in my personal repo

.github/workflows/create-tags.yaml
```diff
@@ -0,0 +1,63 @@
...
+        with:
+          script: |
+            const response = await github.request('POST /repos/{owner}/{repo}/dispatches', {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              event_type: 'image_build_and_draft_release',
+              client_payload: {
+                tag_name: '${{ inputs.tag_name }}' // Pass the input value
+              }
+            });
```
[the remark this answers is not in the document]
fabergat: done

cylewitruk (dismissed):
This looks fine to me as it is. Better to get it out and start trying to use it IMO, as it's (conceptually) better than we have now.

djordon (approved):

.github/workflows/create-tags.yaml
cylewitruk: I will close this as @fdefelici is okay either way (myself too, since it's verified that the GH API will fail this as well).

.github/workflows/create-tags.yaml
cylewitruk: Closing this as it was hashed out in [this comment](https://github.com/stacks-sbtc/sbtc/pull/1584#discussion_r2038995415)

.github/workflows/image-build-and-draft-release.yaml
cylewitruk: Closing this as it seems to be gone when looking at latest

Issue #1304, opened 2025-02-04
[Feature]: Add bitcoin reorg integration tests


djordon:
# Feature - Add bitcoin reorg integration tests

## 1. Description

All of our logic is supposed to be reorg aware. In particular, if there is a reorg on bitcoin, we are supposed to be able to recover pretty seamlessly; picking up the right signer UTXO and re-minting any deposits that have been orphaned due to the reorg. Although we have some tests for this, most of them are not end-to-end tests. This ticket is to add them

### 1.1 Context & Purpose

We want to add more confidence that things will go smoothly in the event of a bitcoin reorg that affects the signers’ UTXO or a deposit.

## 2. Technical Details:

Add integration tests that use the `invalidateblock` and `reconsiderblock` RPCs on bitcoin core to test end-to-end handling of reorgs.

### 2.1 Acceptance Criteria:

- [ ]  We have more end-to-end tests that check that we handle things correctly in the event of a reorg on bitcoin.

## 4. Appendix

A great test would also incorporate how we handle the Stacks side of things in addition to the bitcoin side of things. Alas, this would take quite a bit of work to do because of the difficulty of running a stacks node in a regtest environment.

matteojug-stacks:
The current status is that we have some tests for Bitcoin RPCs during forks, and some devenv tests (https://github.com/stacks-sbtc/sbtc/pull/1585) but they are not really stable, in particular Stacks in devenv doesn't seem to reliably recover from Bitcoin forks.

PR #1585, opened 2025-04-07
test: add devenv forking test for deposit

matteojug:
## Description

Some work for [#1304](https://github.com/stacks-sbtc/sbtc/issues/1304), [384](https://github.com/stacks-sbtc/sbtc/issues/384)

## Changes

Add a test for a forking scenario orphaning a deposit in devenv. The annoying part is that it requires devenv running and with a bit of setup before it can be run. And also it fails most of the times, but it's ignored anyway.


signer/src/bin/demo_cli.rs
```diff
@@ -248,6 +255,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         CliCommand::Info => exec_info(&ctx).await?,
         CliCommand::FundBtc(args) => exec_fund_btc(&ctx, args).await?,
         CliCommand::FundStx(args) => exec_fund_stx(&ctx, args).await?,
+        CliCommand::Forward(args) => exec_forward(&ctx, args).await?,
```
cylewitruk, line 258: What is meant by `Forward`? Wondering if there's a more descriptive/obvious name..

sbtc/src/testing/regtest.rs
```diff
@@ -99,6 +99,36 @@ pub fn initialize_blockchain() -> (&'static Client, &'static Faucet) {
...
+        .unwrap()
+    });
+
+    let faucet = FAUCET.get_or_init(|| {
+        get_or_create_wallet(rpc, BITCOIN_CORE_WALLET_NAME);
+        let faucet = Faucet::new(FAUCET_SECRET_KEY, AddressType::P2wpkh, rpc);
+        faucet.track_address(FAUCET_LABEL);
+
+        // We cannot create 100 blocks here, as it may break the Stacks signers.
+        // So we just assume someone already funded the faucet a bit to survive
```
cylewitruk, line 123: I wonder if you could instead add a wait-loop here that checks for 100 blocks with an optional timeout as a function arg? Just thinking it might help potential users of the function and/or help ensure consistency in tests.

signer/tests/integration/main.rs
```diff
@@ -7,6 +7,7 @@ mod block_observer;
 mod communication;
 mod complete_deposit;
 mod contracts;
+mod devenv;
```
cylewitruk, line 10: What would you think of creating a separate `devenv` folder next to `integration` and putting devenv-targeted tests there? That way you don't need to ignore them and you'd just use `--test devenv` instead of `--test integration` when you want to run them? I'm just thinking that while yeah they're integration tests, they're quite special and will never run in CI etc. (well, at least until we have a working solution for CI).

cylewitruk (commented):
Still need to read through the test more thoroughly, just tossing out a couple of comments in the meantime.

sbtc/src/testing/regtest.rs
matteojug: The issue is that the test requires to stop the bitcoin miner to avoid interferences, so based on when it was stopped we may never reach enough blocks so we would still need some "outside" intervention. That sucks, but expecting some funding happening beforehand as test setup seemed the less painful approach.

signer/src/bin/demo_cli.rs
matteojug: Renamed in [68f2d61](https://github.com/stacks-sbtc/sbtc/pull/1585/commits/68f2d61ae661cd29a87eae9c1ec7c1cbd9b9e6c2)

signer/tests/integration/main.rs
matteojug: Yeah that makes sense, moved in [635bbf7](https://github.com/stacks-sbtc/sbtc/pull/1585/commits/635bbf7b55ebc739abb16ae381f2809a7dfd5baf)

sbtc/src/testing/regtest.rs
cylewitruk: Ah okay, could we make the "limitations" clear on the function docs so that it's visible externally to this file?

cylewitruk (approved):
LGTM, just had one doc request otherwise

sbtc/src/testing/regtest.rs
matteojug: Oh right, makes sense, added in [118bedf](https://github.com/stacks-sbtc/sbtc/pull/1585/commits/118bedfaeb5a80bee969e750a721639e6f7c63f6)

cylewitruk (approved):

signer/tests/devenv/devenv.rs
```diff
@@ -0,0 +1,487 @@
...
+    let (deposit_tx, deposit_request, _) = make_deposit_request(
+        &depositor,
+        deposit_amount,
+        depositor_utxo.clone(),
+        max_fee,
+        aggregate_key.into(),
+        recipient.clone().into(),
+    );
+    rpc.send_raw_transaction(&deposit_tx).unwrap();
+    let block_deposit_tx = faucet.generate_block();
```
djordon, line 399: Shouldn't we generate the block after submitting to Emily? I would think that it could fail if the signers were sufficiently fast (even though that's unlikely).

signer/tests/devenv/devenv.rs
```diff
@@ -0,0 +1,487 @@
...
+    let invlidating_deposit_txid = rpc.send_raw_transaction(&invlidating_deposit_tx).unwrap();
+
+    rpc.call::<GenerateBlockJson>(
+        "generateblock",
+        &[
+            faucet.address.to_string().into(),
+            to_value(&[invlidating_deposit_txid.to_string()]).unwrap(),
+        ],
+    )
+    .unwrap();
```
djordon, line 462: Ahhh, this ensures that the invalidating transaction gets confirmed. I'd add a comment to that effect here.

signer/tests/devenv/devenv.rs
```diff
@@ -0,0 +1,487 @@
...
+    assert_eq!(deposit_tx_status, "success");
+
+    let deposit_txid = deposit_tx["tx_id"].as_str().unwrap();
+
+    // Now we fork:
+    //  - we invalidate the deposit tx block via rpc
+    rpc.invalidate_block(&block_deposit_tx).unwrap();
+
+    //  - we invalidate the deposit tx by spending its vin
+    let invlidating_deposit_tx = create_nop_transaction(
```
djordon, line 445: Looks like `invalidating` is misspelled here.

signer/tests/devenv/devenv.rs
```diff
@@ -0,0 +1,487 @@
...
+    .unwrap();
+
+    // Now we wait for a bit, generating some blocks to enable the stacks miner
+    // to catch up
+    for _ in 0..5 {
+        tokio::time::sleep(Duration::from_secs(10)).await;
+        faucet.generate_block();
+    }
+
+    // Check that the recipient did get the expected sBTC
```
djordon, line 471:
```suggestion
    // Check that the recipient no longer has sBTC after the fork
```

signer/tests/devenv/devenv.rs
```diff
@@ -0,0 +1,487 @@
...
+    let balance = get_sbtc_balance(&stacks, &deployer, &recipient.into())
+        .await
+        .unwrap();
+    dbg!(&balance);
+    assert_eq!(balance.to_sat(), 0);
```
djordon, line 476: Nice

djordon (commented):
Largely looks good 🙏🏽

signer/tests/devenv/devenv.rs
matteojug: Good point, I don't think it would fail too badly, but just to be sure I moved it after the creation on Emily.

signer/tests/devenv/devenv.rs
```diff
@@ -0,0 +1,626 @@
...
+            regtest::p2tr_sign_transaction(&mut tx, input_index, &[utxo.clone()], keypair)
+        }
+        _ => unimplemented!(),
+    };
+    rpc.send_raw_transaction(&tx).unwrap();
+    let txid = &tx.compute_txid();
+
+    // The tx should be in mempool
+    let tx = bitcoin.get_tx(txid).await.unwrap().unwrap();
+    assert!(tx.confirmations.is_none());
```
matteojug, line 236: I'm still not sure why here `confirmations` is `None`, while after confirming and forking it then we get `Some(0)`.

matteojug: Added more checks after offline chat, IIUC:
 - if the tx is in mempool (ie, is still a valid tx in the current mempool), regardless of its history, then we get null confirmations and no block hash
 - if the tx is not in the mempool (eg, invalidated by another one), then if it was included in some forked block we get that block and the number of confirmations (0, assuming nobody built on top of that).

signer/tests/devenv/devenv.rs
```diff
@@ -0,0 +1,739 @@
...
+        .unwrap();
+    assert!(!original_in_fork);
+
+    block_observer_handle.abort();
+    testing::storage::drop_db(db).await;
+}
+
+/// Test checking `getrawtransaction` behaviour after a fork
+#[test_log::test(tokio::test)]
+async fn getrawtransaction_simple_fork() {
```
djordon, line 183: Can't these `getrawtransaction` tests run in CI? If so, we can move them as an integration test in the `sbtc` crate where we have some tests that are actually just bitcoin core behavior tests. We could also move them to the signer crate.

djordon (approved):
Looks good!

I was wondering if we can move the getrawtransaction tests to run in CI. If so we should move them so that they do.

signer/src/bin/demo_cli.rs
```diff
@@ -631,3 +639,15 @@ async fn exec_fund_stx(ctx: &Context, args: FundStxArgs) -> Result<(), Error> {
...
+
+async fn exec_generate_block(ctx: &Context, args: GenerateBlockArgs) -> Result<(), Error> {
+    // Integration tests faucet
+    let recipient = Address::from_str("bcrt1qw508d6qejxtdg4y5r3zarvary0c5xw7kygt080")?
+        .require_network(ctx.network)?;
+    dbg!(
+        ctx.bitcoin_client
+            .generate_to_address(args.count, &recipient)
+            .expect("failed generate blocks")
+    );
```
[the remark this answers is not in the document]
matteojug: Nah, testing leftover, removed in [a404e83](https://github.com/stacks-sbtc/sbtc/pull/1585/commits/a404e8332d8de7770f4b07018cf729703ce42793)

signer/src/testing/btc.rs
```diff
@@ -89,3 +92,13 @@ pub async fn new_zmq_block_hash_stream(endpoint: &str) -> ReceiverStream<Result<
...
+
+/// Return the canonical (tallest) chain tip from `get_chain_tips`
+pub fn get_canonical_chain_tip(rpc: &Client) -> GetChainTipsResultTip {
+    rpc.get_chain_tips()
+        .unwrap()
+        .iter()
+        .max_by_key(|t| t.height)
+        .unwrap()
+        .clone()
+}
```
matteojug, line 104: Now we can have multiple tips on bitcoin, so popping from `get_chain_tips` was returning the wrong one. Looking at the docs it doesn't seem `getchaintips` rpc guarantee a specific ordering on the results. We could switch this in favor of searching for the `active` tip instead of the tallest.

djordon: Yeah, we should go with `active` tip, since that is the canonical one.

matteojug: Done in [a101cd8](https://github.com/stacks-sbtc/sbtc/pull/1585/commits/a101cd86bcf0c094b231f376fb3d331658edd644)

djordon (approved):
Looks good ✅

Issue #1520, opened 2025-03-17
[Chore]: Remove `TRUSTED_REORG_API_KEY` from Emily


Jiloc:
# Chore - Remove `TRUSTED_REORG_API_KEY` from Emily

## 1. Description

Currently, the Emily lambda is deployed with an environment variable called `TRUSTED_REORG_API_KEY`, which is used to differentiate permissions levels in certain endpoints. This is necessary because we rely on security(("ApiGatewayKey" = [])) for API authentication, which due to OpenAPI and AWS API Gateway limited support, only allows permission enforcement at the API Gateway level, not per endpoint.

This chore aims to remove `TRUSTED_REORG_API_KEY` and restructure the API definitions so that public and private endpoints are explicitly defined in the API Gateway, removing the need for the lambda to access API keys.

Currently, 3 endpoints use the variable:
- `update_deposits`/`update_withdarwals` both use the variable to identify the sidecar vs the signers. The signers can only update requests from `Pending` to `Accepted`.
- `set_chainstate` - this code is only used via the `new_block` handler via the private API. It's not exposed anymore on the public API. 

To achieve this, we will take the following approach:

- Short-term solution: Remove `TRUSTED_REORG_API_KEY` and any reference to it from the environment variables and CDK. Create specialized versions of `update_deposits` and `update_withdrawals`, one for the public API and one for the private API. Modify the public and private OpenAPI definitions so that each API exposes the correct endpoints. This ensures the API Gateway can enforce access control at the source, eliminating the need for lambda-level API key verification.

- Long-term solution: Modify the infrastructure to support per-endpoint permissions, removing the dependency on OpenAPI for API Gateway configuration. This could be implemented in CDK, but ideally, we should migrate to Terraform for consistency across infrastructure. This approach would also eliminate the distinction between public and private APIs.

For this issue, we want to address the short-term solution.

## 2. Why It's Necessary

Using TRUSTED_REORG_API_KEY introduces security and operational issues:

- The Lambda should not require access to API keys.
- API keys are stored in environment variables, which is a security risk.
- A previous issue caused the key to be logged by mistake, highlighting the potential for sensitive data exposure.

## 3. Execution Checklist

- [ ] Remove `TRUSTED_REORG_API_KEY` from environment variables.
- [ ] Update CDK and Emily to remove references to `TRUSTED_REORG_API_KEY` and `trusted_reorg_api_key`.
[...]

PR #1586, opened 2025-04-08
[feat] remove trusted api key from Emily

MCJOHN974:
## Description

Closes: #1520

## Changes

- removed `TRUSTED_API_KEY` everwhere
- separated `update_withdraws/deposits` to two versions -- for signer and for sidecar. Version for signer is in public api and version for sidecar is in private one

## Testing Information

Removed usage of `trusted_api_key` from tests. Made test using correct functions from corresponding API now.

## Checklist:


.generated-sources/emily/client/rust/testing/src/apis/withdrawal_api.rs
```diff
@@ -317,10 +329,56 @@ pub async fn get_withdrawals_for_sender(
...
+            local_var_req_builder.header(reqwest::header::USER_AGENT, local_var_user_agent.clone());
+    }
+    if let Some(ref local_var_apikey) = local_var_configuration.api_key {
+        let local_var_key = local_var_apikey.key.clone();
+        let local_var_value = match local_var_apikey.prefix {
+            Some(ref local_var_prefix) => format!("{} {}", local_var_prefix, local_var_key),
+            None => local_var_key,
+        };
+        local_var_req_builder = local_var_req_builder.header("x-api-key", local_var_value);
+    };
```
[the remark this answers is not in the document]
cylewitruk: They haven't added support for tagging the copilot bot, yet, so it doesn't hear you :( But it'll only comment on generated code when it's been changed in a PR, so it shouldn't be too annoying.

emily/handler/src/api/handlers/chainstate.rs
```diff
@@ -97,28 +97,23 @@ pub async fn get_chainstate_at_height(
...
     // Internal handler so `?` can be used correctly while still returning a reply.
     async fn handler(
         context: EmilyContext,
-        api_key: String,
         body: Chainstate,
     ) -> Result<impl warp::reply::Reply, Error> {
         // Convert body to the correct type.
         let chainstate: Chainstate = body;
-        let can_reorg = context.settings.trusted_reorg_api_key == api_key;
+        let can_reorg = true;
```
cylewitruk, line 110: QQ: Is the `can_reorg` parameter of `add_chainstate_entry_or_reorg()` still relevant after these changes?

emily/handler/src/api/handlers/withdrawal.rs
```diff
@@ -336,17 +334,14 @@ pub async fn update_withdrawals(
...
         api_state.error_if_reorganizing()?;
 
-        let is_trusted_key = context.settings.trusted_reorg_api_key == api_key;
         // Signers are only allowed to update withdrawals to the accepted state.
-        if !is_trusted_key {
-            let is_unauthorized = body
-                .withdrawals
-                .iter()
-                .any(|withdrawal| withdrawal.status != Status::Accepted);
+        let is_unauthorized = body
```
cylewitruk, line 338: I might rename this variable (here and elsewhere) to `is_allowed` as it fits better with `403 Forbidden`, where `is_unauthorized` would fit better with `401 Unauthorized` (but this error is about resource permissions and not authentication so `Forbidden` is correct here imo).

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -474,7 +469,100 @@ pub async fn update_deposits(
...
+        }
+
+        updated_deposits.sort_by_key(|(index, _)| *index);
+        let deposits = updated_deposits
+            .into_iter()
+            .map(|(_, deposit)| deposit)
+            .collect();
+        let response = UpdateDepositsResponse { deposits };
+        Ok(with_status(json(&response), StatusCode::CREATED))
+    }
```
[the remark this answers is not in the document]
MCJOHN974: Mmm, from one side yeas, from other side they are a different endpoints now... I'd lean to refactoring it  some day and making signer one `accept_withdrawal`, and make them more different then trying to link them one to each other

emily/handler/src/api/handlers/withdrawal.rs
```diff
@@ -362,21 +357,103 @@ pub async fn update_withdrawals(
...
 
+            updated_withdrawals.push((index, withdrawal));
+        }
+        updated_withdrawals.sort_by_key(|(index, _)| *index);
+        let withdrawals = updated_withdrawals
+            .into_iter()
+            .map(|(_, withdrawal)| withdrawal)
+            .collect();
+        let response = UpdateWithdrawalsResponse { withdrawals };
+        Ok(with_status(json(&response), StatusCode::CREATED))
```
cylewitruk, line 389: QQ: Are we actually creating something here or just updating? If only updating then `200 OK` would probably be more appropriate? (I know this is probably just copied from the other code, so it might not be a question for this PR really but asking anyway)

emily/handler/tests/integration/deposit.rs
```diff
@@ -1013,29 +1008,58 @@ async fn update_deposits_is_forbidden(
...
-                fulfillment,
-                status: new_status,
-                status_message: "foo".into(),
-            }],
-        },
-    )
-    .await;
+    #[derive(Debug)]
+    enum DepositUpdateError {
+        Signer(apis::Error<apis::deposit_api::UpdateDepositsSignerError>),
```
cylewitruk, line 1013: QQ: Do these errors actually have different content, or could it have used the same error type for both (since these endpoints are the same in spirit)?

emily/handler/tests/integration/withdrawal.rs
```diff
@@ -19,6 +19,12 @@ const BLOCK_HASH: &'static str = "TEST_BLOCK_HASH";
 const BLOCK_HEIGHT: u64 = 0;
 const INITIAL_WITHDRAWAL_STATUS_MESSAGE: &'static str = "Just received withdrawal";
 
+#[derive(Debug)]
+enum WithdrawalError {
```
cylewitruk, line 23: Same Q as above

emily/handler/tests/integration/deposit.rs
MCJOHN974: Yeah, generated code produce different types of errors (actually identical inside), and it don't allow to compile them without such a hack. Probably it is not the best hack to do, but some hack is needed. Or, we can divide this test in two, as Francesco suggested

emily/handler/src/api/handlers/withdrawal.rs
MCJOHN974: I'm not an expert in choosing status codes but following common sense I agree  that 200 OK sounds better here. Btw some LLM agree with you as well. But probably it will cause a lot of changes, for example at bridge, and @hz will not really like it

MCJOHN974: Do we ok with changing this, @Jiloc ? Probably it is used in some other places as well

emily/handler/src/api/handlers/withdrawal.rs
MCJOHN974: [9c21d75](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/9c21d7536926a7dfbc6b8b34250f62a2e05d6393)

MCJOHN974:
hmm, @Jiloc where is the code describing how Emily updates deposits and withdrawals? `emily_sidecar/` looks pretty small and I didn't find anything about deposit/withdrawal updates there...

emily/handler/src/api/handlers/deposit.rs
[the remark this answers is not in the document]
MCJOHN974: Yeah, I see, we really have lot of todos already... 

Here is deduplication [eb3fe2b](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/eb3fe2b2112f16525965084e0b31f8796b8dcf02) and generated diff is so ugly unfortunately..

emily/handler/tests/integration/withdrawal.rs
MCJOHN974: [89d24c4](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/89d24c4e09852299089583ff891fbd22e9e1f283)

emily/handler/tests/integration/deposit.rs
MCJOHN974: I agreed with Francesco and deleted this enums at all: [89d24c4](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/89d24c4e09852299089583ff891fbd22e9e1f283)

emily/handler/src/api/handlers/withdrawal.rs
cylewitruk: Yeah nothing to do here, was mostly curious.

emily/handler/src/api/handlers/withdrawal.rs
```diff
@@ -335,12 +335,12 @@ pub async fn update_withdrawals_signer(
         api_state.error_if_reorganizing()?;
 
         // Signers are only allowed to update withdrawals to the accepted state.
-        let is_unauthorized = body
+        let is_allowed = !body
```
cylewitruk, line 339: I think there was another instance down below in another function as well

MCJOHN974: Whoops, [eb3fe2b](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/eb3fe2b2112f16525965084e0b31f8796b8dcf02) actually

emily/handler/src/api/handlers/deposit.rs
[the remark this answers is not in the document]
MCJOHN974: And here some fix to deduplication: [5a68086](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/5a680865070b587f7b013ebb2ad496be1a2c17bc)

emily/handler/src/api/handlers/chainstate.rs
MCJOHN974: At least from what I see this function is called only from handlers from update/set chainstate, and in both of them `can_reorg` is always true now.

However, from naming, `can_reorg` sounds like it can be a wider amount of reasons, so maybe we should keep it. @Jiloc wdyt

emily/handler/src/api/routes/deposit.rs
```diff
@@ -89,17 +90,28 @@ fn create_deposit(
...
+        .then(handlers::deposit::update_deposits_signer)
+}
+
+/// Update deposits from sidecar endpoint.
+fn update_deposits_sidecar(
+    context: EmilyContext,
+) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
+    warp::any()
+        .map(move || context.clone())
+        .and(warp::path!("deposit_private"))
```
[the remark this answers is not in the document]
MCJOHN974: Well, lets leave as is then

emily/handler/tests/integration/withdrawal.rs
MCJOHN974: Hmm, don't think so, just did search for WithdrawalError and DepositError and nothing interesting. Maybe I updated it in some other commit....

emily/handler/src/api/handlers/chainstate.rs
MCJOHN974: Nice! [cd2440b](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/cd2440b7cd4373ad914319129481b9d52de6f5dc)

emily/handler/tests/integration/withdrawal.rs
```diff
@@ -628,13 +617,148 @@ async fn update_withdrawals_is_forbidden(
...
+#[test_case(Status::Pending, Status::Pending, false; "pending_to_pending")]
+#[test_case(Status::Pending, Status::Reprocessing, false; "pending_to_reprocessing")]
+#[test_case(Status::Pending, Status::Confirmed, false; "pending_to_confirmed")]
+#[test_case(Status::Pending, Status::Failed, false; "pending_to_failed")]
+#[test_case(Status::Confirmed, Status::Pending, false; "confirmed_to_pending")]
+#[tokio::test]
+async fn update_withdrawals_is_forbidden_for_sidecar(
+    previous_status: Status,
+    new_status: Status,
+    is_forbidden: bool,
```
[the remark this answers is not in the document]
MCJOHN974: Thanks! [450ff80](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/450ff80c17d9cbf5fd8af60ea8ea575cebadb273) (For signer versions of tests there are some cases where it is true and some where it is false so for signer versions this parameter is still necessary)

emily/handler/tests/integration/withdrawal.rs
MCJOHN974: Just double-checked, I don't see any of this structs in diff between this branch and main. Do you see them?

MCJOHN974:
@Jiloc , @cylewitruk I think all notes are addressed and can be resolved. Thanks for your reviews!

djordon (commented):
Part way done here, looking good so far.

It would be nice if we could differentiate the endpoints with a `/private/deposit` and `/signer/deposit` or something, but that seems more complicated than what is here so what's here is fine.

emily/handler/src/api/handlers/chainstate.rs
```diff
@@ -176,28 +167,22 @@ pub async fn update_chainstate(
...
-            }
+            info!("Inconsistent chainstate found; attempting reorg for {entry:?}");
+            let execute_reorg_request = ExecuteReorgRequest {
+                canonical_tip: chainstate.clone(),
+                conflicting_chainstates,
+            };
+            // Execute the reorg.
+            execute_reorg_handler(context, execute_reorg_request)
+                .await
+                .inspect_err(|e| warn!("Failed executing reorg with error {}", e))?;
```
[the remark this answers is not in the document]
cylewitruk: Yeah, we should avoid inline interpolation in tracing messages and prefer tracing fields. The message is then static and and fields stored separately, which integrates well with tracing and log management tooling, for example you can aggregate `count()` over the static message, grouped by some field(s).

emily/handler/src/api/handlers/withdrawal.rs
```diff
@@ -336,71 +335,117 @@ pub async fn update_withdrawals(
...
+    handler(context, body)
         .await
         .map_or_else(Reply::into_response, Reply::into_response)
 }
 
+async fn update_withdrawals(
+    api_state: ApiStateEntry,
+    context: EmilyContext,
+    body: UpdateWithdrawalsRequestBody,
+    from_sidecar: bool,
```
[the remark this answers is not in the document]
cylewitruk: If you're gonna change it then I'd recommend prefixing `bool`s with `is_` as well :)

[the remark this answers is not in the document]
MCJOHN974: [fae54c1](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/fae54c11a461080a9ed0091a5b64f7f89aa08a4f)

emily/handler/src/api/handlers/chainstate.rs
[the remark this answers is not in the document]
MCJOHN974: [2dd7515](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/2dd75150a8840c0689105c6e55cb6e9ef56d5a00)

cylewitruk (dismissed):
Looks good 👍

emily/handler/src/api/handlers/withdrawal.rs
```diff
@@ -336,71 +335,121 @@ pub async fn update_withdrawals(
...
         // Signers are only allowed to update withdrawals to the accepted state.
-        if !is_trusted_key {
-            let is_unauthorized = body
-                .withdrawals
-                .iter()
-                .any(|withdrawal| withdrawal.status != Status::Accepted);
+        let is_allowed = !body
+            .withdrawals
+            .iter()
+            .any(|withdrawal| withdrawal.status != Status::Accepted);
```
djordon, line 342: I think this would be slightly easier to read:
```rust
let is_allowed = body
    .withdrawals
    .iter()
    .all(|withdrawal| withdrawal.status == Status::Accepted);
```
Also, an empty withdrawals object wouldn't return a `401 Forbidden` with the proposed change. I'm also in favor of returning early and logging if we get an empty withdrawals object.

emily/handler/tests/integration/withdrawal.rs
```diff
@@ -628,13 +617,14 @@ async fn update_withdrawals_is_forbidden(
...
         assert!(response.is_err());
+
         match response.unwrap_err() {
             testing_emily_client::apis::Error::ResponseError(ResponseContent {
                 status, ..
             }) => {
                 assert_eq!(status, 403);
             }
-            e => panic!("Expected a 403 error, got {e}"),
+            e => panic!("Expected a 403 error, got {:#?}", e),
```
djordon, line 627: If you want to use the `Debug` implementation, then you can do:
```suggestion
            e => panic!("Expected a 403 error, got {e:#?}"),
```

djordon (commented):
Generally looks good, had two small comments.

emily/handler/src/api/handlers/withdrawal.rs
MCJOHN974: [5f335df](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/5f335dfc033b2fb2ee1b1e40d91010a462bde2d7) simplified reading here

MCJOHN974: About changing return codes I'd say it is out of scope for this pr

djordon: Do you mean returning early is out of scope? I'm not pushing for the early return, it's likely not an issue either way. And I didn't want to change the return code here, just avoid sending the wrong one when the withdrawals object is empty. With https://github.com/stacks-sbtc/sbtc/pull/1586/commits/5f335dfc033b2fb2ee1b1e40d91010a462bde2d7 we'll avoid that so we're good.

djordon (commented):
Basically, looks good. I think we need to make one more change in https://github.com/stacks-sbtc/sbtc/blob/86c2cb5f67e39e6cee8eee0f437ec83a107c6bd6/emily_cron/app/clients/emily.py#L31-L48 specifically line 44. Or maybe not and @Jiloc will get to it in https://github.com/stacks-sbtc/sbtc/pull/1506?

djordon:
Oh I should have reread the comments above, @Jiloc beat me to it in https://github.com/stacks-sbtc/sbtc/pull/1586#issuecomment-2815495466

MCJOHN974:
@djordon @Jiloc I think this commit does necessary sidecar changes [6b70b24](https://github.com/stacks-sbtc/sbtc/pull/1586/commits/6b70b244516f529dfc38cab15d283d85713da2e7)

PR #1587, opened 2025-04-08
chore(ci): fix double CI-runs between PR and push

cylewitruk:
## Description

Turns out we ran into the issue of concurrently running (same) workflows for push + PR, this attempts to resolve it.

## Changes

Changes the concurrency grouping to use `${{ github.workflow }}-${{ github.head_ref || github.ref }}` instead of only `github.ref`. `github.head_ref` is apparently used for PRs, and `github.ref` for pushes. This should hopefully result in them being the same for both...

## Checklist:

### DIFF
--- a/.github/workflows/on-push.yaml
+++ b/.github/workflows/on-push.yaml
@@ -10,7 +10,7 @@ concurrency:
   # Ensure that this workflow only runs once at a time for each PR or push,
   # cancelling any in-progress runs for the same HEAD (PR) or branch (Push).
   group: push-pr-${{ github.head_ref || github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: true 
 
 env:
   NODE_VERSION: 22.1.0
--- a/.github/workflows/on-push.yaml
+++ b/.github/workflows/on-push.yaml
@@ -1,6 +1,5 @@
 name: Push & PR
 on:
-  push:
   pull_request:
     types: [opened, synchronize, reopened]
 
--- a/.github/workflows/on-push.yaml
+++ b/.github/workflows/on-push.yaml
@@ -9,7 +9,7 @@ permissions: read-all
 concurrency:
   # Ensure that this workflow only runs once at a time for each PR or push,
   # cancelling any in-progress runs for the same HEAD (PR) or branch (Push).
-  group: on-pr-or-push-${{ github.head_ref || github.ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
 
 env:
--- a/.github/workflows/on-push.yaml
+++ b/.github/workflows/on-push.yaml
@@ -10,7 +10,7 @@ concurrency:
   # Ensure that this workflow only runs once at a time for each PR or push,
   # cancelling any in-progress runs for the same HEAD (PR) or branch (Push).
   group: push-pr-${{ github.head_ref || github.ref }}
-  cancel-in-progress: true 
+  cancel-in-progress: true
 
 env:
   NODE_VERSION: 22.1.0
--- a/.github/workflows/on-push.yaml
+++ b/.github/workflows/on-push.yaml
@@ -11,7 +11,7 @@ permissions: read-all
 concurrency:
   # Ensure that this workflow only runs once at a time for each PR or push,
   # cancelling any in-progress runs for the same HEAD (PR) or branch (Push).
-  group: push-pr-${{ github.head_ref || github.ref }}
+  group: push-pr-${{ github.head_ref || github.ref_name || github.run_id}}
   cancel-in-progress: true
 
 env:
--- a/.github/workflows/on-push.yaml
+++ b/.github/workflows/on-push.yaml
@@ -1,4 +1,4 @@
-name: On Push
+name: On PR/Push
 on:
   push:
   pull_request:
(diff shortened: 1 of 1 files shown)


cylewitruk:
Hm, it seems that the concurrency control isn't doing anything regardless...

cylewitruk:
So it looks like we can drop `push` and use only `synchronize (pull_request)`, this means that it won't automatically run on any push to random branches which aren't associated with a PR though. But maybe that's okay? You can still run it manually I think?

Thoughts?

.github/workflows/on-push.yaml
```diff
@@ -1,14 +1,17 @@
-name: On Push
+name: Push & PR
 on:
   push:
+    branches:
+      - main
```
fabergat, line 5: I knew we discuss it in the call, but when will run?

fdefelici: This should run when merging on `main` branch.

If it is the desired behaviour, it is the right configuration.

.github/workflows/on-push.yaml
```diff
@@ -1,14 +1,17 @@
...
     types: [opened, synchronize, reopened]
 
 permissions: read-all
 
 concurrency:
-  # limit concurrency of entire workflow runs for a specific branch
-  group: ${{ github.workflow }}-${{ github.ref }}
+  # Ensure that this workflow only runs once at a time for each PR or push,
+  # cancelling any in-progress runs for the same HEAD (PR) or branch (Push).
+  group: push-pr-${{ github.head_ref || github.ref }}
```
fdefelici, line 14: This should work properly.

Eventually (but not required), you could also use the `github.ref_name` (instead of `github.ref`) that gives you just the branch name without "refs/heads" prefix  (similar to `github.head_ref` does for PRs branch)

Futhermore if it is possible to run this workflow also manually (e.g. from Github UI, or maybe "ready_for_review" event), where the other two attributes don't exist, we could improve the expression using `github.run_id` (represents the workflow unique id).

So the group expression become something like this:
`group: push-pr-${{ github.head_ref || github.ref || github.run_id}}`
Or:
`group: push-pr-${{ github.head_ref || github.ref_name || github.run_id}}`

.github/workflows/on-push.yaml
```diff
@@ -1,14 +1,17 @@
-name: On Push
+name: Push & PR
 on:
   push:
+    branches:
+      - main
   pull_request:
     types: [opened, synchronize, reopened]
```
fdefelici, line 7: just checking: Could be useful to handle also the event `ready_for_review`?

.github/workflows/on-push.yaml
fabergat: yes sure, I am thinking the utility if the code is already merged

fdefelici: I lost you maybe. Does the purpose of this workflow is to run on PRs and then when merging (the PRs) on the `main` branch (that is protected)?

fabergat: I am saying:

- We run the tests during the lifecycle of the PR
- later we merged the PR
- the tests will run when the code is pushed to Main

or am I wrong?

fdefelici: Yes it is correct. 

The action will run twice on different events:
- on pull request: on the PR branch
- on merge: on the main branch

Basically making sure the main branch stay in a good shape after the merge.

I think it is good, because you can also have a PR-branch that is not up-to date with main branch HEAD.

fabergat: You still need a fix but yes can makes sense. Now there is also this new feature that is blocking the merge if you are not in sync but yeah, probably only more traffic

fdefelici: Let's put the other way, based on your consideration. If we are in this scenario:
- main is protected
- the workflow is exactly the same for both `push on main` and `pull request` 
- "blocking merge if not in sync" feature enabled

then yes, probably we could avoid the `push` event on main, because it should be redundant

.github/workflows/on-push.yaml
cylewitruk: Hm, but CI should have been run already on all pushes to a draft review due to `synchronize`, so `ready_for_review` wouldn't be any different from the last `synchronize`, would it?

.github/workflows/on-push.yaml
cylewitruk: Ah, great points 👍 Will adopt them

.github/workflows/on-push.yaml
fdefelici: Yes, you're right. If we don't have any specific action that we want to run from state change "draft to ready", we are all set with the current configuration.

.github/workflows/on-push.yaml
cylewitruk: Changed in [a174061](https://github.com/stacks-sbtc/sbtc/pull/1587/commits/a1740616e5dac462632db5bc1d3ead5527534643)

.github/workflows/on-push.yaml
cylewitruk: Yeah we don't have anything special actions for that transition atm 👍

.github/workflows/on-push.yaml
fdefelici: Basically this should strengthen the scenario described above, where we could avoid the `push` event on main branch (just to save time/resources)

Futhermore, we could also limit this workflow to run only on PRs against `main` branch (if it is desired behavior) like this:

```
on:
  pull_request:
    branches:
      - main
    types: [opened, synchronize, reopened]
```

fabergat (approved):

djordon (dismissed):
Looks good.

Conceptually, the push to main change appears to be redundant, but I'm okay keeping it in for now. Also, I'm not sure if main will show "green" if we don't have it in. Moreover, keeping it in is not completely useless, since some of our tests are not deterministic.

matteojug (dismissed):

fdefelici (approved):

MCJOHN974 (approved):

PR #1592, opened 2025-04-10
fix: ignore (and log) coinbase tx

matteojug:
## Description

Closes: https://github.com/stacks-sbtc/sbtc/issues/1196

## Changes

Avoid calling `get_tx_info` if the transaction is a coinbase one, and log it when extracting sbtc txs (deposit ones are already logged with something like `WARN block-observer:load_latest_deposit_requests:load_requests: signer::block_observer: could not validate deposit request error=transaction is coinbase, txid: ..., block hash Some(...)`).

Also removed some random trailing spaces in the first commit.

## Testing Information

Added a test for coinbase donation and deposit.

### DIFF
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -441,10 +441,16 @@ impl<C: Context, B> BlockObserver<C, B> {
             // through all of the transactions within that block. This
             // means the `get_tx_info` call below should not fail.
             let txid = tx.compute_txid();
-            let tx_info = btc_rpc
-                .get_tx_info(&txid, &block_hash)
-                .await?
-                .ok_or(Error::BitcoinTxMissing(txid, None))?;
+
+            let tx_info_result = btc_rpc.get_tx_info(&txid, &block_hash).await;
+
+            // If the tx is coinbase, we just ignore it.
+            if let Err(err @ Error::BitcoinTxCoinbase(_, _)) = tx_info_result {
+                tracing::warn!(%err, "ignoring coinbase tx when extracting sbtc transaction");
+                continue;
+            }
+
+            let tx_info = tx_info_result?.ok_or(Error::BitcoinTxMissing(txid, None))?;
 
             // sBTC transactions have as first txin a signers spendable output
             let tx_type = if tx_info.is_signer_created(&signer_script_pubkeys) {
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -90,6 +90,10 @@ impl DepositRequestValidator for CreateDepositRequest {
             return Ok(None);
         };
 
+        if response.tx.is_coinbase() {
+            return Err(Error::BitcoinTxCoinbase(self.outpoint.txid));
+        }
+
         // The `get_tx_info` call here should not return None, we know that
         // it has been included in a block.
         let Some(tx_info) = client.get_tx_info(&self.outpoint.txid, &block_hash).await? else {
@@ -436,21 +440,20 @@ impl<C: Context, B> BlockObserver<C, B> {
                 continue;
             }
 
+            let txid = tx.compute_txid();
+            if tx.is_coinbase() {
+                tracing::warn!(%txid, "ignoring coinbase tx when extracting sbtc transaction");
+                continue;
+            }
+
             // This function is called after we have received a
             // notification of a bitcoin block, and we are iterating
             // through all of the transactions within that block. This
             // means the `get_tx_info` call below should not fail.
-            let txid = tx.compute_txid();
-
-            let tx_info_result = btc_rpc.get_tx_info(&txid, &block_hash).await;
-
-            // If the tx is coinbase, we just ignore it.
-            if let Err(err @ Error::BitcoinTxCoinbase(_, _)) = tx_info_result {
-                tracing::warn!(%err, "ignoring coinbase tx when extracting sbtc transaction");
-                continue;
-            }
-
-            let tx_info = tx_info_result?.ok_or(Error::BitcoinTxMissing(txid, None))?;
+            let tx_info = btc_rpc
+                .get_tx_info(&txid, &block_hash)
+                .await?
+                .ok_or(Error::BitcoinTxMissing(txid, None))?;
 
             // sBTC transactions have as first txin a signers spendable output
             let tx_type = if tx_info.is_signer_created(&signer_script_pubkeys) {
--- a/signer/tests/integration/block_observer.rs
+++ b/signer/tests/integration/block_observer.rs
@@ -1624,7 +1624,7 @@ async fn block_observer_ignores_coinbase() {
     let validate_result =
         signer::block_observer::DepositRequestValidator::validate(&request, &bitcoin_client, false);
     match validate_result.await {
-        Err(Error::BitcoinTxCoinbase(tx, _)) if tx == deposit_request.outpoint.txid => {}
+        Err(Error::BitcoinTxCoinbase(tx)) if tx == deposit_request.outpoint.txid => {}
         _ => panic!("Expected a err, got something else"),
     }
 
--- a/signer/tests/integration/block_observer.rs
+++ b/signer/tests/integration/block_observer.rs
@@ -15,11 +15,16 @@ use bitcoincore_rpc::RpcApi as _;
 use blockstack_lib::chainstate::nakamoto::NakamotoBlock;
 use blockstack_lib::chainstate::nakamoto::NakamotoBlockHeader;
 use blockstack_lib::net::api::getpoxinfo::RPCPoxInfoData;
+use clarity::types::chainstate::StacksAddress;
+use clarity::vm::types::PrincipalData;
 use emily_client::apis::deposit_api;
 use emily_client::models::CreateDepositRequestBody;
 use fake::Fake as _;
 use fake::Faker;
 use rand::SeedableRng as _;
+use sbtc::deposits::CreateDepositRequest;
+use sbtc::deposits::DepositScriptInputs;
+use sbtc::deposits::ReclaimScriptInputs;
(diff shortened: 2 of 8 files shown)


djordon (commented):
I think we can use `Transaction::is_coinbase` instead here.

signer/src/block_observer.rs
```diff
@@ -441,10 +441,16 @@ impl<C: Context, B> BlockObserver<C, B> {
...
-                .await?
-                .ok_or(Error::BitcoinTxMissing(txid, None))?;
+
+            let tx_info_result = btc_rpc.get_tx_info(&txid, &block_hash).await;
+
+            // If the tx is coinbase, we just ignore it.
+            if let Err(err @ Error::BitcoinTxCoinbase(_, _)) = tx_info_result {
+                tracing::warn!(%err, "ignoring coinbase tx when extracting sbtc transaction");
+                continue;
+            }
```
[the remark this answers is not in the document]
djordon: Well we'd want logging, but yeah.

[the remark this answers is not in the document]
matteojug: Oh nice, I didn't know about that, changed in [8af0982](https://github.com/stacks-sbtc/sbtc/pull/1592/commits/8af0982a6a8f9eb00181c90b55e4e63a5990d764)

signer/src/block_observer.rs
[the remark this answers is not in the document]
matteojug: Returning the custom error in [8af0982](https://github.com/stacks-sbtc/sbtc/pull/1592/commits/8af0982a6a8f9eb00181c90b55e4e63a5990d764) (so we get the same error as the one mentioned in the PR description)

djordon (approved):
Looks good 🙏🏽

signer/tests/integration/block_observer.rs
```diff
@@ -1415,3 +1420,251 @@ async fn block_observer_updates_dkg_shares_after_observing_bitcoin_block() {
 
     testing::storage::drop_db(db).await;
 }
+
+#[test_log::test(tokio::test)]
+async fn block_observer_ignores_coinbase() {
+    let mut rng = rand::rngs::StdRng::seed_from_u64(51);
```
MCJOHN974, line 1426: We now can use `signer::testing::get_rng()` here!

signer/tests/integration/block_observer.rs
```diff
@@ -1415,3 +1420,251 @@ async fn block_observer_updates_dkg_shares_after_observing_bitcoin_block() {
...
+    shares.dkg_shares_status = model::DkgSharesStatus::Verified;
+    db.write_encrypted_dkg_shares(&shares).await.unwrap();
+
+    // Okay, now to make the actual coinbase donation. We send some funds to
+    // their address.
+    let script_pub_key = shares.script_pubkey.deref();
+    let address = Address::from_script(script_pub_key, bitcoin::Network::Regtest).unwrap();
+
+    // We include also another donation to ensure we correctly process the block
+    let donation_amount = 123_456;
```
MCJOHN974, line 1527: nit: generating this amount with rng will make test a little bit better

signer/tests/integration/block_observer.rs
```diff
@@ -1415,3 +1420,251 @@ async fn block_observer_updates_dkg_shares_after_observing_bitcoin_block() {
...
+
+    assert_eq!(amount, donation_amount);
+    assert_eq!(output_index, donation_outpoint.vout);
+    assert_eq!(txid.deref(), &donation_outpoint.txid);
+
+    // ** Step 4 **
+    //
+    // Create a deposit from coinbase.
+
+    // First, we also send a donation (that will endup in the same block as the
```
MCJOHN974, line 1553:
```suggestion
    // First, we also send a donation (that will end up in the same block as the
```

signer/tests/integration/block_observer.rs
```diff
@@ -1415,3 +1420,251 @@ async fn block_observer_updates_dkg_shares_after_observing_bitcoin_block() {
...
+    // deposit) to ensure we process the block just fine.
+    let donation_amount_2 = 123_456 + 1;
```
MCJOHN974, line 1555: Why we want it to be `donation_amount + 1`? Can this be generated via rng as well?

signer/tests/integration/postgres.rs
```diff
@@ -7142,7 +7142,7 @@ mod get_pending_accepted_withdrawal_requests {
     ///
     /// ```text
     ///
-    /// Height:       0           1           2
```
MCJOHN974, line 7145: What tool you used to find this blankspaces? Maybe create an issue to add this tool to our CI?

MCJOHN974: Oh, it is probably fmt/clippy but applied to all targets, maybe will be convered by Aldur's PR

signer/tests/integration/block_observer.rs
matteojug: mmm I don't think in this case: we are not testing the parsing itself, we just need an amount to ensure we parsed the expected donation. Having a random amount here could add more noise.

signer/tests/integration/block_observer.rs
matteojug: To have a different amount from the first donation, as above imho having this random only adds possible noise

signer/tests/integration/postgres.rs
matteojug: I searched manually with `[ \t]+\n` in vscode 😅

signer/tests/integration/block_observer.rs
matteojug: Nice! Changed in [77a428b](https://github.com/stacks-sbtc/sbtc/pull/1592/commits/77a428bb29c1e1e90b12708b1b9ff27cfe890183)

signer/tests/integration/postgres.rs
MCJOHN974: WoW, nice!

MCJOHN974 (dismissed):

signer/tests/integration/block_observer.rs
```diff
@@ -1415,3 +1420,249 @@ async fn block_observer_updates_dkg_shares_after_observing_bitcoin_block() {
...
+    let script_pub_key =
+        sbtc::deposits::to_script_pubkey(deposit_script.clone(), reclaim_script.clone());
+    let address = Address::from_script(&script_pub_key, bitcoin::Network::Regtest).unwrap();
+
+    let block_hash = rpc.generate_to_address(1, &address).unwrap()[0];
+    let transactions = rpc.get_block(&block_hash).unwrap();
+
+    let deposit_tx = transactions.txdata[0].clone();
+    // Sanity check
+    assert!(deposit_tx.input[0].previous_output.is_null());
```
cylewitruk, line 1656: Maybe use the `.is_coinbase()` here since that's the check done in the observer?

signer/src/block_observer.rs
```diff
@@ -436,11 +440,16 @@ impl<C: Context, B> BlockObserver<C, B> {
                 continue;
             }
 
+            let txid = tx.compute_txid();
```
cylewitruk, line 443: QQ: What's the actual downside/risk of having coinbase transactions extracted/stored when the deposit requests themselves will fail validation? I'm just pondering over what we gain by giving them special treatment vs. other failed deposits (your test would have been much simpler too).

signer/tests/integration/block_observer.rs
matteojug: Oh right, updated in [c4ea8af](https://github.com/stacks-sbtc/sbtc/pull/1592/commits/c4ea8af172192f4b09216d56fba5e1f44e3673d9)

signer/src/block_observer.rs
matteojug: We wouldn't extract deposits txs regardless (since they are not spent "to the signers" directly), so this is only relevant for donations. In that case, I guess the main issue is that they cannot be spent right away -- this is not so relevant now that the system is up and running, but for the same reason it doesn't seems we are gaining anything by supporting them. OTOH, having them in the mix could mess up some assumptions on "any tx we care about have prevouts" (eg, we check that to distinguish between donations and sbtc tx)

djordon: Yeah, the diff is about as small, I'm checking whether it will work now.

cylewitruk: Ah right, was thinking of the coinbase as a deposit but of course that's not the case 🤦‍♂️

I'm fine with the check, it's tiny (unlike the test to verify it...), my question was primarily out of a trying-to-minimize-special-case-handling perspective.

cylewitruk: I'll leave this open since apparently daniel is checking something, but feel free to close it.

cylewitruk (approved):
Looks fine to me 👍

signer/src/block_observer.rs
```diff
@@ -436,11 +440,16 @@ impl<C: Context, B> BlockObserver<C, B> {
...
+            if tx.is_coinbase() {
+                tracing::warn!(%txid, "ignoring coinbase tx when extracting sbtc transaction");
+                continue;
+            }
```
MCJOHN974, line 447: nit: computing `txid` after this check can save us some cpu cycles.

matteojug: We would still need it for the logging though, so it should be equivalent

signer/src/block_observer.rs
djordon: I just realized now that I was probably responding to the wrong thing above. Whoops!

signer/src/block_observer.rs
MCJOHN974: of, I see, we use txid for logging, sorry

PR #1596, opened 2025-04-11
feat: do not require the `vin.prevout` field in `getrawtransaction` responses

djordon:
## Description

We currently require the `vin.prevout` field in `getrawtransaction` responses. This is okay, because the only time that we don't have a prevout is for coinbase transactions and we don't support them (https://github.com/stacks-sbtc/sbtc/pull/1592). This change does not allow us to support coinbase transactions, but does make things a little more correct.

Plus I did the work already when testing out the coinbase PR.

## Changes

* Make the `BitcoinTxInfo.vin.prevout` field optional.

## Testing Information

There is a test that an RPC to get a coinbase transaction will succeed.

## Checklist:


signer/src/bitcoin/rpc.rs
```diff
@@ -197,7 +197,7 @@ pub struct BitcoinTxVin {
     /// This field is omitted if block undo data is not available, so it is
     /// missing whenever the `fee` field is missing in the
     /// [`BitcoinTxInfo`].
-    pub prevout: BitcoinTxVinPrevout,
+    pub prevout: Option<BitcoinTxVinPrevout>,
```
matteojug, line 200: nit: should we add to the comment that this is none for coinbase?

signer/src/storage/model.rs
```diff
@@ -214,7 +214,7 @@ impl From<Deposit> for DepositRequest {
         // It's most likely the case that each of the inputs "came" from
         // the same Address, so we filter out duplicates.
         let sender_script_pub_keys: BTreeSet<ScriptPubKey> = tx_input_iter
-            .map(|tx_in| tx_in.prevout.script_pub_key.script.into())
+            .filter_map(|x| Some(x.prevout?.script_pub_key.script.into()))
```
matteojug, line 217: For coinbase, we end up with an empty `sender_script_pub_keys`. I guess that's correct, and since we use `sender_script_pub_keys` only for the blocklist check, this is fine. I just wanted to double check this is what we want -- i.e., if we lift the deposit coinbase check, we would (correctly) create the request with empty `sender_script_pub_keys`.

signer/tests/integration/bitcoin_rpc.rs
```diff
@@ -133,6 +133,31 @@ fn btc_client_gets_transaction_info() {
...
+fn btc_client_gets_transaction_info_coinbase() {
+    let client = BitcoinCoreClient::new(
+        "http://localhost:18443",
+        regtest::BITCOIN_CORE_RPC_USERNAME.to_string(),
+        regtest::BITCOIN_CORE_RPC_PASSWORD.to_string(),
+    )
+    .unwrap();
+    let (rpc, _) = regtest::initialize_blockchain();
+
+    // Now let's confirm it and try again
```
matteojug, line 146: nit: outdated comment?

djordon: ohh, copy paste on the comments :(

signer/src/storage/model.rs
djordon: I think this is what we want. We need input addresses but don't have any, so we should be fine. I'm not sure what else we could do.

signer/src/bitcoin/rpc.rs
djordon: Good point.

djordon: Ohhh wait, maybe we do not want to merge this actually 🤔.

djordon:
I don't think that we should do this. We could be missing the `prevouts` because of the missing undo data, in addition to the coinbase thing. This change hides the fact that we could be missing prevout data for other reasons, so let's fix the prevout thing first and then maybe do this.

PR #1601, opened 2025-04-14
feat: drop stacks transactions and transactions tables

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1598.

## Changes

* Drop the `stacks_transactions` and `transactions` tables.
* Remove unnecessary functions that write to the dropped tables.
* Remove unnecessary structs.

## Testing Information

No new tests, but this is closer to a refactor than anything, so if tests pass then we should be okay here.

## Checklist:


signer/tests/integration/contracts.rs
```diff
@@ -205,40 +201,7 @@ async fn complete_deposit_wrapper_tx_accepted<T: AsContractCall>(contract: Contr
...
-    assert!(!transactions.is_empty());
-
-    let txids = transactions
-        .iter()
-        .map(|stx| blockstack_lib::burnchains::Txid::from_bytes(&stx.txid))
-        .collect::<Option<HashSet<_>>>()
-        .unwrap();
-
-    assert!(txids.contains(&txid));
+    println!("{txid}");
```
MCJOHN974, line 204: Do we need this print?

MCJOHN974: As I see this test does not have nor `assert`'s nor `unwrap`'s so we can delete it (or add some assertions)

cylewitruk (approved):
Well I've tried to come up with reasons as to why we would need to retain this but haven't managed. So I think this is just fine 👍  We have our nodes if we would need to rebuild either this or part of it in the future.

signer/tests/integration/contracts.rs
djordon: Well, I suppose we could delete it, since it is ignored already. However, I actually get some use out of these tests every now and again, and the print is useful when I use it.

This test is supposed to check that the stacks node will accept our contract call transactions and that they get confirmed. The call to `sign_and_submit` will panic if the transaction gets rejected, so it tests acceptance. The part that is deleted is supposed to check that the transaction gets confirmed. It never did that "correctly", it needs to wait for some arbitrary amount of time for the confirmation, so we'd need to change it when we finally run a stacks node in CI. Oh, and we should delete the print when the test is enabled.

MCJOHN974: Oookay, makes sense

MCJOHN974 (approved):
lgtm

MCJOHN974 (dismissed):

signer/migrations/0017__remove_stacks_transactions_table.sql
```diff
@@ -0,0 +1,3 @@
+DROP TABLE sbtc_signer.stacks_transactions CASCADE;
```
cylewitruk, line 1: On 2nd thought, could we get rid of `CASCADE`? Cascading deletes/drops are generally scary... Better to be explicit in removing their dependencies IMO, and it makes it clearer for our signer partners that manually review/apply their migrations.

djordon: Done in https://github.com/stacks-sbtc/sbtc/pull/1601/commits/32285c2c8626cfc69ef8c84deca8123d4ed0a977

MCJOHN974 (approved):

Issue #1588, opened 2025-04-09
[Feature]: Run DKG if current signer set changes


djordon:
# Feature - Run DKG if current signer set changes

## 1. Description

We want the signers to run distributed key generation (DKG) and rotate keys on Stacks whenever the current signer set changes. Right now, DKG is only triggered if the signer is configured to allow it based on the Bitcoin blockchain’s chain tip block height. This update adds a comparison between the current signer set and the set recognized on Stacks where DKG is permitted if they differ.

## 2. Technical Details:

Running DKG requires the `TxCoordinator` to initiate DKG and the `TxSigner` to allow it. The `TxCoordinator` decides to run DKG by calling the `transaction_coordinator::should_coordinate_dkg` function, while the `TxSigner` uses `transaction_signer::assert_allow_dkg_begin` function. Both of these should have the same underlying logic so one should probably be modified to use the other. Also, the function should add a new check for whether `config.signer.bootstrap_signing_set` and `config.state().current_signer_public_keys()` differ and allow DKG to run if they do.

We may also want to rename the `bootstrap_signing_set` field in the signer config to something else since it will mean the "current signing set".

### 2.1 Acceptance Criteria:
   - [ ] Update the logic of `transaction_coordinator::should_coordinate_dkg` to allow DKG to run if the current signer set differs from the one in the smart contract. 
   - [ ] Update the DKG gating logic used in the `TxSigner` to mirror the one used by the coordinator.
   - [ ] We have less duplicated code when deciding whether to run DKG.

PR #1602, opened 2025-04-14
[feat] trigger dkg on changed signer set

MCJOHN974:
## Description

This PR will make coordinator initiate dkg, and signer allow dkg if signer set changed. More precisely, dkg will be initiated if signer set in signer config and signer set in last rotate keys transaction differ.

Closes: #1588, and closes https://github.com/stacks-sbtc/sbtc/issues/1617

## Changes
- added new parameter to signer config
- changed `should_coordinate_dkg` and `assert_allow_dkg_begin` to start dkg if signer set changed.
- fixed mistake in random comment

## Testing Information
- added two integration (however they are more unit) tests

## Checklist:


signer/tests/integration/transaction_coordinator.rs
```diff
@@ -1057,6 +1062,95 @@ async fn run_dkg_from_scratch() {
     }
 }
 
+/// Tests that dkg will not be triggered if signer set didn't change
```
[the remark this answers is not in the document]
MCJOHN974: Rust do not allow this, unfortunately (or, imo, thankfully).

signer/src/config/mod.rs
```diff
@@ -354,6 +354,8 @@ pub struct SignerConfig {
     pub dkg_verification_window: u16,
     /// The maximum stacks fee in microSTX that the signer will accept for any stacks transaction.
     pub stacks_fees_max_ustx: NonZeroU64,
+    /// Indicates if the signer will approve starting dkg on changed signer set.
+    pub start_dkg_if_signer_set_changed: bool,
```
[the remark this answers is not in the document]
djordon: I don't think that it is good practice to use configuration variables only for integration tests. The production configuration and any internal integration testing related stuff should be separate.

> Hmm, do you think one more production config variable is _that_ bad ? Like, I imagine, for signers who don't read whole the code, they just read name of the config parameter and it make sense for them. For those who read the code it will be relatively fast to see that this parameter don't have any side effects and only do what is described, as well.

Yeah, nothing is really end of the world here, but it would be bad practice. Better to do the work to make testing easier.

> About making it configurable in tests only -- yeah maybe it can make some sense. But also it will add some compexity to production code, and I think it will look weird a bit: if we_are_in_tests {...} will bring first thoghts like "are we really testing it in good way? behavior in tests and prod differs!"

Yeah it would look funny, but we wouldn't have the code compile in non-testing code by using something like `#[cfg(test)]` above the testing related code. But you've convinced me that this is a bad idea too. So let's do the work for making testing easier.

signer/src/transaction_coordinator.rs
```diff
@@ -2540,6 +2540,21 @@ pub async fn should_coordinate_dkg(
...
 
+    // Trigger dkg if signer set changes
+    let last_dkg_signer_set: BTreeSet<_> = context
+        .state()
+        .current_signer_set()
+        .get_signers()
+        .iter()
+        .map(|signer| *signer.public_key())
+        .collect();
+    let config_signer_set = config.signer.bootstrap_signing_set();
```
cylewitruk, line 2551: Maybe we could consolidate this logic into the state since it's done in multiple places? e.g. something like:
```rust
let changed = !context
    .state()
    .current_signer_set()
    .is_same(&config.signer.bootstrap_signing_set());
```

MCJOHN974:
Note: `coordinator_skip_onchain_completed_deposits::deposit_not_completed` was flaky on prev commit, but I runned it ~5 times on last commit and it was okish all 5 times

MCJOHN974:
Well, it is actually flaky =(

signer/src/testing/context.rs
```diff
@@ -206,6 +206,43 @@ impl<Storage, Bitcoin, Stacks>
     }
 }
 
+/// Addition to Context trait, used in tests.
+pub trait TestingContext: Clone + Sync + Send {
```
cylewitruk, line 210: Why was this needed? If you need to modify settings on-the-fly with the `TestContext` below you can use `.modify_settings(|config| { ... })`.

MCJOHN974: `modify_settings` works for `TestContext::builder()`, while I want sometimes change settings for already built context

signer/src/context/signer_context.rs
```diff
@@ -166,6 +169,19 @@ where
     }
 }
 
+#[cfg(any(test, feature = "testing"))]
+impl<S, BC, ST, EM> TestingContext for SignerContext<S, BC, ST, EM>
```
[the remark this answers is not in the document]
MCJOHN974: `TestContext` is a wrapper over `SignerContext` and it can't access `config` field of inner `SignerContext` without any extra code

MCJOHN974:
Well it is still flaky

MCJOHN974:
`coordinator_skip_onchain_completed_deposits::deposit_not_completed` still flaky =(

It seems that regtest::initialize blockchain sometimes generate rotate key transactions and we don't have control over coordinator process in this test (and afaik we want to have more or less solid and full coordinator process here). So, I think, this test should be rewrited to something more controllable then regtest

signer/src/transaction_coordinator.rs
MCJOHN974: [cd9d6cd](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/cd9d6cd531acee663b5cb3a3a387d632ddc68f0e)

signer/src/testing/context.rs
MCJOHN974: [634b84b](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/634b84beeb03c4dfd8a64efec19597c1e80a8eb8)

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4368,6 +4471,25 @@ async fn coordinator_skip_onchain_completed_deposits(deposit_completed: bool) {
...
+        .push(new_pubkey);
+    fake_ctx.config_mut().signer.bootstrap_signing_set =
+        ctx.config().signer.bootstrap_signing_set.clone();
+
+    ctx.state()
+        .update_current_signer_set(ctx.config().signer.bootstrap_signing_set());
+    fake_ctx
+        .state()
+        .update_current_signer_set(ctx.config().signer.bootstrap_signing_set());
+
```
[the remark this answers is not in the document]
MCJOHN974: This test was fixed, resolving this.

signer/src/transaction_coordinator.rs
cylewitruk: `have` is the wrong form here, could we name it `has_same_pubkeys` (or `cmp_pubkeys`) instead?

MCJOHN974: Whooops, English......

About `cmp_pubkeys` imo it is less obvious how to interpret it's results.

MCJOHN974: [e9bdafb](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/e9bdafb256bc8649d96d08a6285911bdd4ebc6f4)

signer/src/context/signer_state.rs
```diff
@@ -537,6 +537,18 @@ impl SignerSet {
             .expect("BUG: Failed to acquire read lock")
             .len()
     }
+
+    /// Returns true if the two signer sets have the same public keys.
+    pub fn has_same_pubkeys<I, P>(&self, other: I) -> bool
+    where
+        I: IntoIterator<Item = P>,
+        P: Into<PublicKey>,
```
djordon, line 545: No need to make this generic, since we only use `BTreeSet<PublicKey>` for the type of `other`.

signer/src/keys.rs
```diff
@@ -51,6 +51,12 @@ impl Deref for PublicKey {
     }
 }
 
+impl From<&PublicKey> for PublicKey {
+    fn from(pk: &PublicKey) -> Self {
+        *pk
+    }
+}
+
```
djordon, line 59: We can remove this, if we use a concrete type for `other` in `has_same_pubkeys`.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4258,6 +4360,21 @@ async fn process_rejected_withdrawal(is_completed: bool, is_in_mempool: bool) {
...
+    // probability of chosing correct coordinator is 50%.
+    loop {
+        let need_break = coordinator_skip_onchain_completed_deposits_inner(deposit_completed).await;
+        if need_break {
+            break;
+        }
+    }
+}
+
+async fn coordinator_skip_onchain_completed_deposits_inner(deposit_completed: bool) -> bool {
```
djordon, line 4377: Instead of doing all of this, I think we can change the test so that the current signer set is just one signer (themselves) with the bootstrap signer set is identical to it. It's still unsatisfactory since it's a complex setup, but then we do not need to loop through the test multiple times.

signer/src/context/signer_state.rs
MCJOHN974: Are you sure we will never want to use it for HashSet or Vec ? And also I think this generic don't look very complex.

djordon: Yeah, it's not that complex. It's just that it adds unnecessary complexity; we need to add even more code to satisfy the trait bounds (we needed the `From<&PublicKey>` implementation). Instead we should skip all of that and just do what we need to. Most of our functions are not generic unless they reduce complexity (well, that's the goal at least).

Maybe we never need to use this with a `HashSet` or whatever, or we will. We can always change it in the future.

signer/src/testing/context.rs
```diff
@@ -206,6 +206,43 @@ impl<Storage, Bitcoin, Stacks>
...
+    fn config_mut(&mut self) -> &mut Settings {
+        self.inner.config_mut()
+    }
+}
+
+/// DKG can be triggered if last dkg signer set differ from one in config.
+/// However, we don't want to test this functionality in some tests, so
+/// this function makes sure that dkg won't be triggered because of changes in signer set.
+/// Warn: in some cases dkg still can be triggered on changed signer set even if you used
+/// this function.
```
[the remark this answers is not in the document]
MCJOHN974: Hmm, actually I think there are no such cases, removed this warn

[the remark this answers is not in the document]
MCJOHN974: [196bacf](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/196bacf8cf7416a90ca661ad74ad0b0e95b07fca)

signer/src/keys.rs
MCJOHN974: [ef75378](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/ef7537847ffe636587abe7e733399f34bf95adc7)

signer/src/context/signer_state.rs
MCJOHN974: [ef75378](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/ef7537847ffe636587abe7e733399f34bf95adc7)

signer/src/testing/context.rs
```diff
@@ -206,6 +211,28 @@ impl<Storage, Bitcoin, Stacks>
     }
 }
 
+/// DKG can be triggered if the last DKG signer set differs from the one in config.
+/// However, we don't want to test this functionality in some of our tests, so
+/// this function makes sure that DKG won't be triggered because of changes in signer set.
+pub fn prevent_dkg_on_changed_signer_set<Storage, Bitcoin, Stacks, Emily>(
```
cylewitruk, line 217: Could this not have been a method on `TestContext` directly (i.e. in the block below)?

cylewitruk:
This looks generally good to me now, happy to approve once all comments are resolved (and you're done pushing) 👍

signer/src/transaction_coordinator.rs
```diff
@@ -2791,6 +2802,8 @@ mod tests {
             .await
             .unwrap();
 
+        prevent_dkg_on_changed_signer_set(&mut context);
```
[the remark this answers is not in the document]
MCJOHN974: For example in signer/tests/integration/block_observer.rs we use it in the loop and we really _need_ to update config on each iteration

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4282,6 +4386,12 @@ async fn coordinator_skip_onchain_completed_deposits(deposit_completed: bool) {
     // state gets updated in the block observer. We're not running a
     // block observer in this test, nor are we going through main, so
     // we manually update the state here.
+    // When the signer binary starts up in main(), it sets the current
+    // signer set public keys in the context state using the values in
+    // the bootstrap_signing_set configuration parameter. Later, the
+    // state gets updated in the block observer. We're not running a
+    // block observer in this test, nor are we going through main, so
+    // we manually update the state here.
```
matteojug, line 4394: nit: dup?

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4345,6 +4457,10 @@ async fn coordinator_skip_onchain_completed_deposits(deposit_completed: bool) {
     let start_flag = Arc::new(AtomicBool::new(false));
     let flag = start_flag.clone();
 
+    // We will use network messages to detect the coordinator attempt, so we
+    // need to connect to the network
+    let fake_ctx = ctx.clone();
+
```
matteojug, line 4463: nit: is there a reason for this move? Its previous location seems more fitting

matteojug (commented):
With this (and https://github.com/stacks-sbtc/sbtc/pull/1612), we should still work fine on config change, but we shouldn't be able to complete the new DKG if we are adding a peer because of `SignerSet::is_allowed_peer`. I guess we are missing a specific issue for that?

signer/src/transaction_coordinator.rs
```diff
@@ -2548,6 +2548,17 @@ pub async fn should_coordinate_dkg(
     let storage = context.get_storage();
     let config = context.config();
 
+    // Trigger dkg if signer set changes
+    let signer_set_changed = !context
+        .state()
+        .current_signer_set()
+        .has_same_pubkeys(&config.signer.bootstrap_signing_set);
+
+    if signer_set_changed {
```
matteojug, line 2557: The ticket doesn't mention it, but should we trigger this also if `bootstrap_signatures_required` change?

MCJOHN974: Hmm, why we want to run dkg if signatures required changed? Because if signatures required changed we want to rebuild signers aggregated address changing it's spending condition and we do such changes via dkg?

signer/tests/integration/transaction_coordinator.rs
MCJOHN974: [52c4f73](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/52c4f73f473e673eca89b107b190a50527f42e81)

signer/src/testing/context.rs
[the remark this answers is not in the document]
MCJOHN974: [3a6df5d](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/3a6df5d919368e0572e6766250a7ae7ed765f2fa)

signer/tests/integration/transaction_coordinator.rs
MCJOHN974: [5f46e77](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/5f46e77f77738b6801ce260a4ae93274976e1687)

signer/src/transaction_coordinator.rs
djordon: Yes, if the signatures required changed then that implies a different set of secret shares and stuff.

The next question that I have, not for this PR, is whether we want to remove the flexibility of the signer threshold altogether and have it derived from the number of signers in the signing set.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -1049,6 +1052,102 @@ async fn run_dkg_from_scratch() {
...
+        .get_bitcoin_canonical_chain_tip_ref()
+        .await
+        .expect("failed to get chain tip")
+        .expect("no chain tip");
+
+    ctx.inner
+        .state()
+        .update_current_signer_set(config_signer_set.iter().cloned().collect());
+    assert!(!should_coordinate_dkg(&ctx, &chaintip).await.unwrap());
+    assert!(assert_allow_dkg_begin(&ctx, &chaintip).await.is_err());
```
djordon, line 1098: We should also have these assertions above, although it's less clear where.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -1049,6 +1052,102 @@ async fn run_dkg_from_scratch() {
     }
 }
 
+/// Tests that dkg will not be triggered if signer set didn't change
+#[test(tokio::test)]
+async fn dont_run_dkg_if_signer_set_didnt_change() {
```
djordon, line 1057: Both of these tests need to drop the database at the end.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -1049,6 +1052,102 @@ async fn run_dkg_from_scratch() {
...
+        .get_bitcoin_canonical_chain_tip_ref()
+        .await
+        .expect("failed to get chain tip")
+        .expect("no chain tip");
+
+    ctx.inner
+        .state()
+        .update_current_signer_set(config_signer_set.iter().cloned().collect());
+    assert!(should_coordinate_dkg(&ctx, &chaintip).await.unwrap());
+    assert!(assert_allow_dkg_begin(&ctx, &chaintip).await.is_ok());
```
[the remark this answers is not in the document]
MCJOHN974: What about approach like in this [1ae0d51](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/1ae0d51765421fb96f8ae44a0b22d9999b1c74b4) commit? Here it is clearly seen that two testcases in this test differ just 1 line of code -- removing one signer, and clearly seen that in two testcases the opposite assertions are true.

signer/tests/integration/transaction_coordinator.rs
MCJOHN974: [4a3a399](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/4a3a399fbafeee758c35b2812f2d2c94d4459ac8)

signer/src/transaction_coordinator.rs
```diff
@@ -2548,6 +2548,30 @@ pub async fn should_coordinate_dkg(
     let storage = context.get_storage();
     let config = context.config();
 
+    // Trigger dkg if signatures_required has changed
+    if let Some(last_key_rotation_signatures_required) = context
+        .get_storage()
+        .get_last_key_rotation(&bitcoin_chain_tip.block_hash)
+        .await?
+        .map(|last_key_rotation| last_key_rotation.signatures_required)
```
djordon, line 2556: We probably do not want to make this database read because it takes too long. Instead we want to add a new `signatures_required` field to the context's state object and just read it. We should set this new field in the `BlockObserver::set_signer_set_and_aggregate_key` function.

MCJOHN974: @djordon two following commits should address it

signer/tests/integration/transaction_coordinator.rs
MCJOHN974: [5b34a19](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/5b34a19ef412f70b7494bff329363e1d48f82614) and one more commit add this checks

signer/tests/integration/transaction_coordinator.rs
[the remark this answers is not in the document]
MCJOHN974: added this as well

signer/src/block_observer.rs
```diff
@@ -583,8 +583,8 @@ impl<C: Context, B> BlockObserver<C, B> {
         Ok(())
     }
 
-    /// Update the `SignerState` object with current signer set and
-    /// aggregate key data.
+    /// Update the `SignerState` object with current signer set, signatures
+    /// required and aggregate key data.
```
djordon, line 587:
```suggestion
    /// Update the `SignerState` object with the current signer set, signatures
    /// required, and aggregate key data.
```

signer/src/block_observer.rs
```diff
@@ -693,6 +696,19 @@ impl<C: Context, B> BlockObserver<C, B> {
     }
 }
 
+/// Structure describing the extended info about signer set.
+pub struct SignerSetInfo {
+    /// The aggregate key used for signing, aggregate public key of all signers
+    /// which was produced during the DKG.
+    pub maybe_aggregate_key: Option<PublicKey>,
```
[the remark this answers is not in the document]
MCJOHN974: I see some usages of such prefix in signer and quite a lot in emily, so I think we have them a lot enough to make this usage not a deviation from repo standard, and for me I'd say this prefix is really nice and representative

signer/src/testing/wsts.rs
```diff
@@ -468,16 +468,19 @@ impl SignerSet {
     }
 
     /// Dump the current signer set as a dummy rotate-keys transaction to the given storage
+    /// `None` for `signatures_required` defaults to the number of signers
     pub async fn write_as_rotate_keys_tx<S, Rng>(
         &self,
         storage: &S,
         chain_tip: &model::BitcoinBlockHash,
         shares: &EncryptedDkgShares,
+        signatures_required: Option<u16>,
```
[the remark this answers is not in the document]
MCJOHN974: [42f88c6](https://github.com/stacks-sbtc/sbtc/pull/1602/commits/42f88c65f5f8cf7c498cbb1ca623a6d39e4e6fb2)

signer/src/block_observer.rs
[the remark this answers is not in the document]
cylewitruk: The `maybe_` prefix is pretty common/idiomatic (and we do use it scattered throughout the signer as well), so I'm positive to working towards this direction _for variable names_.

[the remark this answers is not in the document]
djordon: Yeah, for variable names, where you cannot see the type without IDE help, it makes sense. I'm okay with variable names having them if the author wants them. However, for field names it seems unnecessary, which is part of the reason why I was leaning against it.

signer/src/testing/wsts.rs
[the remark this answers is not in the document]
djordon: > You mean it is incorrect for this `signatures_required` to be not equal to `shares.signature_share_threshold` ?

Hmmm, so the previous code always set this to the number of signers and I have no idea why. The old code seems wrong. If changing it to `shares.signature_share_threshold` doesn't break anything then let's go with that.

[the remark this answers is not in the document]
MCJOHN974: I think it doesn't break anything and commit I mentioned above adds it

signer/src/testing/wsts.rs
```diff
@@ -468,6 +468,7 @@ impl SignerSet {
     }
 
     /// Dump the current signer set as a dummy rotate-keys transaction to the given storage
+    /// `None` for `signatures_required` defaults to the number of signers
```
matteojug, line 471: nit: this comment addition is a bit unclear: there is no `signatures_required` argument, and it seems the function always create a rotate key event from `shares.signature_share_threshold` that cannot be none. What is it referring to?

matteojug (approved):

MCJOHN974:
@matteojug @djordon please reapprove

matteojug (approved):

PR #1604, opened 2025-04-14
feat: fetch stacks block headers instead of the blocks

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1589.

This doesn't batch process stacks blocks, but it fetches block headers so the signer is unlikely to OOM. Still, it does so all at once, writing them to the database all at once as well.

After the conversion, headers take up 112 bytes each, assuming each slimmed stacks block header has a bitcoin header hash and an 8 byte block height. A signer can keep `17857142 = 2 * 1000 * 1000 * 1000 / 112` blocks with 2 GB of memory. Assuming 8000 blocks per day, that buys us 6 years worth of stacks blocks.

## Changes

* Only take the headers for the nakamoto blocks fetched for each tenure.

## Testing Information

I still need to test this on devenv, but it is ready for review.

## Checklist:


signer/src/stacks/api.rs
```diff
@@ -320,16 +321,87 @@ impl TenureBlocks {
...
-            .map(|block| StacksBlock::from_nakamoto_block(block, bitcoin_anchor))
+    pub fn as_stacks_blocks(self) -> impl Iterator<Item = StacksBlock> {
+        let bitcoin_anchor = self.anchor_block_hash;
+        self.headers.into_iter().map(move |header| StacksBlock {
+            block_hash: header.block_id.into(),
+            block_height: header.block_height,
+            parent_hash: header.parent_block_id.into(),
+            bitcoin_anchor,
+        })
     }
```
cylewitruk, line 405: I'd probably rename this to `into_iter()` or `into_xx_iter()` etc. while you're at it as it's more conventional for a method that consumes `self` and returns an `Iterator`.

signer/src/stacks/api.rs
```diff
@@ -320,16 +321,87 @@ impl TenureBlocks {
...
+            anchor_block_hash: value.anchor_block_hash,
+            anchor_block_height: value.anchor_block_height,
+        }
+    }
+}
+
+/// This struct represents a non-empty subset of the Stacks block headers
+/// that were created during a tenure.
+#[derive(Debug)]
+pub struct TenureHeaders {
```
cylewitruk, line 373: Maybe `TenureBlockHeaders`? I first read this as `a list of headers of all tenures`

signer/src/stacks/api.rs
```diff
@@ -1146,20 +1219,20 @@ where
...
-        if db.stacks_block_exists(block.header.parent_block_id).await? {
+        if db.stacks_block_exists(header.parent_block_id).await? {
             tracing::debug!("parent block known in the database");
             break;
         }
         // There are more blocks to fetch, so let's get them.
-        let tenure_blocks = stacks.get_tenure(block.header.parent_block_id).await?;
-        blocks.push(tenure_blocks);
+        let tenure_blocks = stacks.get_tenure(header.parent_block_id).await?;
+        headers.push(tenure_blocks.into());
```
cylewitruk, line 1232: Maybe we should update the fn docs to explicitly mention that the returned headers are in descending order, and also change the doc to not mention a datastore since this method has no knowledge of such.

djordon: Yeah, make sense. We should also return them in ascending order, it doesn't matter for the current code but maybe it does for some newer code.

As for the datastore mention, this method needs something that implements `DbRead` as an input, so we should probably mention why in the docs (or at least not remove it). Or am I misunderstanding what your asking.

signer/src/stacks/api.rs
djordon: https://github.com/stacks-sbtc/sbtc/pull/1604/commits/4986f8c64bc81e8568ded96e3451fad1e71d50c8

signer/src/stacks/api.rs
djordon: https://github.com/stacks-sbtc/sbtc/pull/1604/commits/98036af2577459ebc6f87ba1abc86cebf71cb39e

signer/src/stacks/api.rs
djordon: Done in https://github.com/stacks-sbtc/sbtc/pull/1604/commits/1c59451abb9014900e3c5f955ef3062f2c3bc129.

cylewitruk (dismissed):
Lgtm!

signer/src/block_observer.rs
```diff
@@ -380,17 +380,23 @@ impl<C: Context, B> BlockObserver<C, B> {
     async fn process_stacks_blocks(&self) -> Result<(), Error> {
         tracing::info!("processing stacks block");
         let stacks_client = self.context.get_stacks_client();
+        let db = self.context.get_storage_mut();
```
MCJOHN974, line 383: I suspect you don't need mut here

MCJOHN974 (approved):
lgtm but check comment above

signer/src/block_observer.rs
djordon: We do for the call to `write_stacks_block_headers` below.

signer/src/stacks/api.rs
```diff
@@ -320,16 +321,109 @@ impl TenureBlocks {
...
-        let bitcoin_anchor = &self.anchor_block_hash;
-        self.blocks
-            .iter()
-            .map(|block| StacksBlock::from_nakamoto_block(block, bitcoin_anchor))
+/// A slimmed down [`NakamotoBlockHeader`]
+#[derive(Debug, Clone, PartialEq)]
+pub struct StacksBlockHeader {
+    /// The total number of StacksBlocks and NakamotoBlocks preceding this
+    /// block in this block's history.
+    pub block_height: u64,
```
MCJOHN974, line 335: We can use `model::StacksBlockHeight` now!

signer/src/stacks/api.rs
```diff
@@ -320,16 +321,109 @@ impl TenureBlocks {
...
+#[derive(Debug)]
+pub struct TenureBlockHeaders {
+    /// The subset of Stacks block headers that of Nakamoto blocks that
+    /// were created during a tenure. This is always non-empty.
+    headers: Vec<StacksBlockHeader>,
+    /// The bitcoin block that this tenure builds off of.
+    pub anchor_block_hash: BitcoinBlockHash,
+    /// The height of the bitcoin block associated with the above block
+    /// hash.
+    pub anchor_block_height: u64,
```
MCJOHN974, line 381: This can be `model::BitcoinBlockHeight`

signer/src/stacks/api.rs
djordon: Good call. I'll get to this when I change the base branch of the PR and merge main.

MCJOHN974 (dismissed):
lgtm

signer/src/stacks/api.rs
djordon: Done in https://github.com/stacks-sbtc/sbtc/pull/1604/commits/d9657bf66b7d1a357d2bdc3bb9545d62878b7516

cylewitruk (approved):
lgtm!

MCJOHN974 (approved):

PR #1606, opened 2025-04-15
CI: Use Wrapped External Actions From `stacks-sbtc/actions` Repository

BowTiedDevOps:
This PR replaces the actions used in the workflows with the ones from `actions` repository. I've also added `name` and `id` to all steps of the jobs and applied formatting for consistency and readability.

This PR should only be merged after https://github.com/stacks-sbtc/actions/pull/2, as it is dependent on some of the modifications included there.


BowTiedDevOps:
Note: the vet dependencies are failing because the cache action doesn't exist in the actions repository yet

.github/workflows/on-push.yaml
```diff
@@ -37,44 +37,72 @@ env:
...
+        id: setup_buf
+        uses: stacks-sbtc/actions/setup-buf@main
         with:
           version: ${{ env.BUF_VERSION }}
           setup_only: true
-      - uses: actions-rust-lang/setup-rust-toolchain@11df97af8e8102fd60b60a77dfbf58d40cd843b8 #v1.10.1
+
+      - name: Setup Rust
+        id: setup_rust
+        uses: stacks-sbtc/actions/setup-rust-toolchain@main
```
MCJOHN974, line 76: It is better to use some fixed commit imo

MCJOHN974: Oh, it is main of our own repo, it is fine then

BowTiedDevOps: You're right, we could also have fixed commits even for our repo for security. This depends on how we want to allow commits to the main branch, if it's only PRs and must be approved by maintainers we could leave it as is. @aldur @fabergat could you please share your opinion on this?

fabergat: Let's stay with main otherwise if have to update an action we have to do in many places...

.github/workflows/image-build-and-draft-release.yaml
```diff
@@ -41,21 +41,22 @@ jobs:
...
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
       ## Checkout the branch of the release provided.
       ## This requires that a release branch exists for the tag.
-      - name: Checkout
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Checkout Repository
+        id: checkout_repository
```
fabergat, line 58: I think on this style thing we should vote, @cylewitruk @djordon @matteojug, what's your opinion?

fabergat:
@BowTiedDevOps you have also a check failing

.github/workflows/image-build-and-draft-release.yaml
MCJOHN974: If you mean '-' vs '_' I'm for '-'

fdefelici: Maybe better "snake_case", because when we need to use it in expressions can be accessed with dot notation
`${{ steps.my_step_id.whatever }}`

it is true that we could also access hyphen like this:
`
${{ steps.['mystep-id'].whatever }}
`
Anyhow "snake_case" can be used without issue everywhere letting us just enforcing one way to access the `steps` context

.github/workflows/image-build-and-draft-release.yaml
```diff
@@ -71,7 +72,7 @@ jobs:
       ## ex. debian will have this tag: `type=ref,event=tag,enable=${{ matrix.dist == 'debian' }}`
       - name: Docker Metadata ( ${{ matrix.dist }} )
         id: docker_metadata
-        uses: docker/metadata-action@8e5442c4ef9f78752691e2d8f8d19755c6f78e81 # v5.5.1
+        uses: stacks-sbtc/actions/docker/metadata-action@main
```
djordon, line 75: Can we pin the commit for `stacks-sbtc` actions as well?

.github/workflows/on-push.yaml
djordon: I actually think that we should fix the commit, with the main benefit here being a predictable and stable CI.

djordon (commented):
The main thing is for us to continue to pin the commit for our actions. I think doing so is a bit more secure and makes it so all updates to our CI process are explicitly done through an additional PR. It adds some overhead but I'm not a fan of CI surprises.

fdefelici (approved):
LGMT

About the fixed commit or not on `stacks-stbc/actions`, there isn't a correct answer:
- using `@main` gives less maintainance effort and the certainty that all the automations are always aligned with latest changes for free
- using `fixed commit` requires potentially more effort, but gives more predictability on the automation behaviour

I don't think security, in both cases, it is a concern because of how we manage development (main protected branch, PRs approvals and so on)

Just for sharing:
- In previous experiences, I usually opted for the first solution (@main)
- on `stacks-core` also, we are using `@main`

My preference would be for using `@main`, but I'll be fine if the majority opts for the `fixed commit`

aldur:
Thank you @fdefelici! After giving this some (collective) thought on a sync, we have decided we'll try using a pinned commit in sBTC CI. If that turns out to be too much work, we can revisit.

In theory, we could also have Dependabot open PRs when the actions repository gets updated.

.github/workflows/image-build-and-draft-release.yaml
fabergat: @BowTiedDevOps I think the majority of people prefer snake case

.github/workflows/image-build-and-draft-release.yaml
fabergat: The action is pinned but in the stack actions repo, we can also pin our repo but in case we need to update an action we have to update both the stack action repo and all the occurrences

fabergat:
> LGMT
> 
> About the fixed commit or not on `stacks-stbc/actions`, there isn't a correct answer:
> 
> * using `@main` gives less maintainance effort and the certainty that all the automations are always aligned with latest changes for free
> * using `fixed commit` requires potentially more effort, but gives more predictability on the automation behaviour
> 
> I don't think security, in both cases, it is a concern because of how we manage development (main protected branch, PRs approvals and so on)
> 
> Just for sharing:
> 
> * In previous experiences, I usually opted for the first solution (@main)
> * on `stacks-core` also, we are using `@main`
> 
> My preference would be for using `@main`, but I'll be fine if the majority opts for the `fixed commit`

I also agree with you

aldur (approved):

aldur:
I am approving it with a nit. It would be cool, if possible, to specify the commit we are pointing to in a variable at the top of the workflow, so we can change that to update the full thing.

MCJOHN974 (approved):
lgtm

cylewitruk:
> I am approving it with a nit. It would be cool, if possible, to specify the commit we are pointing to in a variable at the top of the workflow, so we can change that to update the full thing.

I thought that I had tried this previously but didn't get it to work.. would be very happy if it does though :)

BowTiedDevOps:
> I am approving it with a nit. It would be cool, if possible, to specify the commit we are pointing to in a variable at the top of the workflow, so we can change that to update the full thing.

I've tried it on a private repo, unfortunately it doesn't work - the `uses:` key takes a hardcoded string, as it's being evaluated before anything else (env, secrets or other ways to store the hash).

aldur (approved):

MCJOHN974 (approved):
new commits seems fine, reapproving

fdefelici (approved):
re-approving as well

aldur:
Need one more, then this can be merged -- @BowTiedDevOps if the corresponding PRs have been merged in all other repos, we can also switch to enforcing the policy at the org level.

PR #1608, opened 2025-04-15
fix: emily-cron hiro api usage

Jiloc:
## Description

When testing the devenv, I noticed that the dockerfiles where actually using a wrong URL for the Hiro API (with `/extended` ). Also, since when we removed the stacks chain tip information from the deposit updates, we don't need to fetch it in the cron-job anymore.

## Changes
- fixed Hiro base url in docker files
- removed `HiroAPI.get_stacks_block()` from `update_deposits` and relative tests

## Checklist:


djordon (approved):
Looks good

MCJOHN974 (approved):
lgtm

PR #1609, opened 2025-04-16
Fix GH CLI version and add tag name print

fabergat:
## Description

This PR fixes two problems:

- GH CLI version during the building phase
- Add the tag name print

## Checklist:


.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
```diff
@@ -32,9 +32,10 @@ RUN apt-get update && \
     apt-get install -y curl && \
     curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg && \
     chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg && \
-    echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null && \
+    echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" > /etc/apt/sources.list.d/github-cli.list && \
+    echo "Package: gh\nPin: origin cli.github.com\nPin-Priority: 1001" > /etc/apt/preferences.d/gh-cli && \
     apt-get update && \
-    apt-get install -y gh=2.69.0
+    apt-get install -y gh
```
cylewitruk, line 38: Do we not want a pinned version anymore? I guess we're ensuring (or at least trying to ensure) that the package is from the github repo and signed so maybe it isn't that big of a deal, but it could "randomly break" the workflow if it updates.

.github/workflows/create-tags.yaml
```diff
@@ -19,10 +19,27 @@ permissions:
...
 
 jobs:
+  log_tag:
+    name: Print Tag Info
+    runs-on: ubuntu-latest
+    outputs:
+      tag_name: ${{ steps.set-output.outputs.tag_name }}
+    steps:
+      - name: Show Tag Name
+        id: print-name
```
cylewitruk, line 29: I'm not sure you need to explicitly set `id` for these if you're not planning on using their output? Or do you want to be explicit?

BowTiedDevOps: That's true, you don't have to explicitly set ids unless you're using the outputs in future steps. However, for consistency and readability I've added names and ids to all steps in the repo in my pr, but we can proceed however you think is best.

cylewitruk: Fair enough, I don't have a preference, this was more a question on the convention we want to use moving forward

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
cylewitruk: Ah okay 🫡

cylewitruk (approved):
Lgtm

.github/workflows/create-tags.yaml
fabergat: Good point

cylewitruk: My 2 cents: I personally prefer the dash as it's what (I want to believe) is more common, but I don't really care either :)

cylewitruk (approved):

BowTiedDevOps (approved):
LGTM

.github/workflows/create-tags.yaml
MCJOHN974: also like '-' more but not strongly

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
```diff
@@ -32,9 +32,10 @@ RUN apt-get update && \
     apt-get install -y curl && \
     curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg && \
     chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg && \
-    echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null && \
+    echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" > /etc/apt/sources.list.d/github-cli.list && \
+    echo "Package: gh\nPin: origin cli.github.com\nPin-Priority: 1001" > /etc/apt/preferences.d/gh-cli && \
```
MCJOHN974, line 36: Why `Pin-Priority` is `1001`? Maybe I'm newbie in github actions but probably it worth to add comment explaining this?

MCJOHN974:
Largely lgtm but check my comment above

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
fabergat: Because it forces APT to use that repo for GH CLI and not the debian one

fabergat: The GH cli in the debian is pretty old before we were pinning the version and APT was able to get the right one from the right repo but in this case I had to instruct it

MCJOHN974 (approved):

PR #1612, opened 2025-04-16
feat: coordinator processes requests on DKG failure

setzeus:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1593

This PR updates so that when DKG coordination fails in the transaction coordinator, instead of failing the entire process (aka entirely halt), we now gracefully fall back to using the existing aggregate key if one is available. This improves system resilience by allowing operations to continue even when DKG coordination encounters issues.

## Changes
Modified `process_new_blocks` to handle DKG coordination failures by falling back to existing aggregate key. The system will now:

1. Attempt DKG coordination as normal
2. On failure, log the error and attempt to use existing aggregate key
3. Only fail if _**no**_ existing aggregate key is available

## Testing Information
There is a single test added that verifies fallback behavior works as expected.

## Checklist:

### DIFF
--- a/signer/tests/integration/transaction_coordinator.rs
+++ b/signer/tests/integration/transaction_coordinator.rs
@@ -4731,23 +4731,25 @@ async fn should_handle_dkg_coordination_failure() {
         aggregate_key: existing_aggregate_key,
         ..Faker.fake_with_rng(&mut rng)
     };
-    storage
-        .write_encrypted_dkg_shares(&dkg_shares)
-        .await
-        .unwrap();
+    storage.write_encrypted_dkg_shares(&dkg_shares).await.unwrap();
 
     // Create a bitcoin block to serve as chain tip
     let bitcoin_block: model::BitcoinBlock = Faker.fake_with_rng(&mut rng);
     storage.write_bitcoin_block(&bitcoin_block).await.unwrap();
 
-    // Update the context state with the existing aggregate key
-    context
-        .state()
-        .set_current_aggregate_key(existing_aggregate_key);
+    // Create a set of signer public keys and update the context state
+    let mut signer_keys = BTreeSet::new();
+    for _ in 0..3 {  // Create 3 signers
+        let private_key = PrivateKey::new(&mut rng);
+        let public_key = PublicKey::from_private_key(&private_key);
+        signer_keys.insert(public_key);
+    }
+    context.state().update_current_signer_set(signer_keys);
+    context.state().set_current_aggregate_key(existing_aggregate_key);
 
     // Create coordinator with test parameters using SignerNetwork::single
     let network = SignerNetwork::single(&context);
-    let coordinator = TxCoordinatorEventLoop {
+    let mut coordinator = TxCoordinatorEventLoop {
         context: context.clone(),
         network: network.spawn(),
         private_key: PrivateKey::new(&mut rng),
@@ -4759,12 +4761,9 @@ async fn should_handle_dkg_coordination_failure() {
         is_epoch3: true,
     };
 
-    // Run the coordinator - this will handle process_new_blocks internally
-    let run_result = coordinator.run().await;
-    assert!(
-        run_result.is_ok(),
-        "Coordinator run should complete successfully"
-    );
+    // Call process_new_blocks directly instead of run()
+    let result = coordinator.process_new_blocks().await;
+    assert!(result.is_ok(), "process_new_blocks should complete successfully");
 
     // Verify the existing aggregate key was used as fallback
     let current_key = context.state().current_aggregate_key();
--- a/signer/tests/integration/transaction_coordinator.rs
+++ b/signer/tests/integration/transaction_coordinator.rs
@@ -4726,7 +4726,7 @@ async fn should_handle_dkg_coordination_failure() {
     let bitcoin_block: model::BitcoinBlock = Faker.fake_with_rng(&mut rng);
     storage.write_bitcoin_block(&bitcoin_block).await.unwrap();
 
-    // Get chain tip reference 
+    // Get chain tip reference
     let chain_tip = storage
         .get_bitcoin_canonical_chain_tip_ref()
         .await
@@ -4735,29 +4735,36 @@ async fn should_handle_dkg_coordination_failure() {
 
     // Create a set of signer public keys and update the context state
     let mut signer_keys = BTreeSet::new();
-    for _ in 0..3 {  // Create 3 signers
+    for _ in 0..3 {
+        // Create 3 signers
         let private_key = PrivateKey::new(&mut rng);
         let public_key = PublicKey::from_private_key(&private_key);
         signer_keys.insert(public_key);
     }
-    context.state().update_current_signer_set(signer_keys.clone());
+    context
+        .state()
+        .update_current_signer_set(signer_keys.clone());
 
     // Mock the stacks client to handle contract source checks
-    context.with_stacks_client(|client| {
-        client.expect_get_contract_source().returning(|_, _| {
-            Box::pin(async {
-                Ok(ContractSrcResponse {
-                    source: String::new(),
-                    publish_height: 1,
-                    marf_proof: None,
+    context
+        .with_stacks_client(|client| {
+            client.expect_get_contract_source().returning(|_, _| {
+                Box::pin(async {
+                    Ok(ContractSrcResponse {
+                        source: String::new(),
+                        publish_height: 1,
+                        marf_proof: None,
+                    })
                 })
-            })
-        });
-    }).await;
+            });
+        })
+        .await;
 
     // Verify DKG should run
     assert!(
-        transaction_coordinator::should_coordinate_dkg(&context, &chain_tip).await.unwrap(),
+        transaction_coordinator::should_coordinate_dkg(&context, &chain_tip)
+            .await
+            .unwrap(),
         "DKG should be triggered since no shares exist yet"
     );
 
@@ -4778,9 +4785,15 @@ async fn should_handle_dkg_coordination_failure() {
 
     // This should trigger DKG coordination which will fail, falling back to existing key
     let result = coordinator.process_new_blocks().await;
-    assert!(result.is_ok(), "process_new_blocks should complete successfully even with DKG failure");
+    assert!(
+        result.is_ok(),
+        "process_new_blocks should complete successfully even with DKG failure"
+    );
 
     // Verify that we can still process blocks after DKG failure
     let result = coordinator.process_new_blocks().await;
-    assert!(result.is_ok(), "Should be able to continue processing blocks after DKG failure");
+    assert!(
+        result.is_ok(),
+        "Should be able to continue processing blocks after DKG failure"
+    );
 }
--- a/signer/tests/integration/transaction_coordinator.rs
+++ b/signer/tests/integration/transaction_coordinator.rs
@@ -4731,7 +4731,10 @@ async fn should_handle_dkg_coordination_failure() {
         aggregate_key: existing_aggregate_key,
         ..Faker.fake_with_rng(&mut rng)
     };
-    storage.write_encrypted_dkg_shares(&dkg_shares).await.unwrap();
+    storage
+        .write_encrypted_dkg_shares(&dkg_shares)
+        .await
+        .unwrap();
 
     // Create a bitcoin block to serve as chain tip
     let bitcoin_block: model::BitcoinBlock = Faker.fake_with_rng(&mut rng);
@@ -4739,17 +4742,20 @@ async fn should_handle_dkg_coordination_failure() {
 
     // Create a set of signer public keys and update the context state
     let mut signer_keys = BTreeSet::new();
-    for _ in 0..3 {  // Create 3 signers
+    for _ in 0..3 {
+        // Create 3 signers
         let private_key = PrivateKey::new(&mut rng);
         let public_key = PublicKey::from_private_key(&private_key);
         signer_keys.insert(public_key);
     }
     context.state().update_current_signer_set(signer_keys);
-    context.state().set_current_aggregate_key(existing_aggregate_key);
+    context
+        .state()
+        .set_current_aggregate_key(existing_aggregate_key);
 
     // Create coordinator with test parameters using SignerNetwork::single
     let network = SignerNetwork::single(&context);
-    let mut coordinator = TxCoordinatorEventLoop {
+    let coordinator = TxCoordinatorEventLoop {
         context: context.clone(),
         network: network.spawn(),
         private_key: PrivateKey::new(&mut rng),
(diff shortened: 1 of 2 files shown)


signer/tests/integration/transaction_coordinator.rs
```diff
@@ -1413,6 +1413,65 @@ async fn run_subsequent_dkg() {
...
+    ctx.inner
+        .state()
+        .update_current_signer_set(signer_set_without_signer.iter().cloned().collect());
+    assert!(!should_coordinate_dkg(&ctx, &chaintip).await.unwrap());
+
+    // test case 2: All peers connected - dkg should run
+    for peer in signer_set_without_signer.iter() {
+        ctx.inner.state().add_connected_peer(peer.clone().into());
+    }
+    assert!(should_coordinate_dkg(&ctx, &chaintip).await.unwrap());
```
MCJOHN974, line 1472: You probably wan't check that `assert_allow_dkg` have similar behaviour

MCJOHN974: And above as well

setzeus: Looked into this but the current implementation of 'assert_allow_dkg' returns ok() in both of these scenarios since it *just* checks that the signer set has changed: https://github.com/stacks-sbtc/sbtc/blob/main/signer/src/transaction_signer.rs#L1529

Want me to updated the assert function as well or leave as is?

signer/src/context/signer_state.rs
```diff
@@ -28,6 +28,8 @@ pub struct SignerState {
     // The current bitcoin chain tip. This gets updated at the end of the
     // block observer's duties when it observes a new bitcoin block.
     bitcoin_chain_tip: RwLock<BitcoinBlockRef>,
+    // All peers currently connected
+    connected_peers: RwLock<HashSet<PeerId>>,
```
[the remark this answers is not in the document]
setzeus: Updated in [e2d811](https://github.com/stacks-sbtc/sbtc/pull/1612/commits/e2d581171de58d13fe19b01c808913de98488732)

signer/tests/integration/transaction_coordinator.rs
MCJOHN974: I mean, check that `assert_allow_dkg` return `Ok` when we expect signer to approve dkg and `Err` otherwise

signer/src/context/signer_state.rs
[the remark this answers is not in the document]
MCJOHN974: Whoops, we are using `Hashbrown`'s hashset and it is better then btreeset, sorry for this

signer/tests/integration/transaction_coordinator.rs
MCJOHN974: Yeah, sorry, I mean't `assert_allow_dkg_begin`. It and `should_coordinate_dkg` should have similar behavior and probable we want occasionally create single function both for coordinator and signer

setzeus: Copy. For now will just updated 'assert_allow_dkg_begin' so that they have similar behavior & then open an issue for consolidating both of those functions into one.

signer/src/context/signer_state.rs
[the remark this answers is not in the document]
setzeus: Updated in [cc108e0](https://github.com/stacks-sbtc/sbtc/pull/1612/commits/cc108e05a5089482e4cc5b50f6b8a1c6abc18def)

signer/tests/integration/transaction_coordinator.rs
setzeus: Updated in [4d94618](https://github.com/stacks-sbtc/sbtc/pull/1612/commits/4d94618a9e4d65d1fd110b90445ae7733d53d375)

signer/src/network/libp2p/event_loop.rs
```diff
@@ -125,6 +126,7 @@ pub async fn run(ctx: &impl Context, swarm: Arc<Mutex<Swarm<SignerBehavior>>>) {
                     }
                     SwarmEvent::ConnectionClosed { peer_id, cause, endpoint, .. } => {
                         tracing::trace!(%peer_id, ?cause, ?endpoint, "connection closed");
+                        ctx.state().remove_connected_peer(&peer_id);
```
djordon, line 129: One thing that I'm not sure of is whether a closed connection necessarily means that we are no longer connected to a peer, because peers can open more than one connection at a time. So we probably need to track the number of open connections per peer as well.

djordon: I suppose we can keep the API, and hence this code the same, but update the implementation to track the number of connections.

setzeus: Hmmm okay. So basically need to refactor starting from the SignerState struct - instead of tracking connections w/ a HashSet, I should track with a HashMap<PeerId, u32>?

setzeus: Addressed in [29bcd59](https://github.com/stacks-sbtc/sbtc/pull/1612/commits/29bcd59a4d6a4870b18d6584cb4ecb5add45a763)

signer/src/network/libp2p/event_loop.rs
```diff
@@ -110,6 +110,7 @@ pub async fn run(ctx: &impl Context, swarm: Arc<Mutex<Swarm<SignerBehavior>>>) {
                             let _ = swarm.disconnect_peer_id(peer_id);
                         } else {
                             tracing::debug!(%peer_id, ?endpoint, "connected to peer");
+                            ctx.state().add_connected_peer(peer_id);
```
cylewitruk, line 113: While this information is interesting, it's not necessarily going to be correct for what you're trying to assert.

We're using `gossipsub`, which entails that peers _don't need to be directly connected to each other_ and may receive messages relayed via peers. This would actually be expected for a node operator which strictly controls which peers they connect to.

What I was (poorly) suggesting [here](https://github.com/stacks-sbtc/sbtc/issues/1593#issuecomment-2796680488) is to use a "last seen" timestamp on the existing `Signer` struct, but what I left out was details on how that should be achieved.

We can't rely on any heartbeats from direct connections (like `ping`), so we need some sort of heartbeat that's sent over gossipsub to determine signers' "alive" status.

setzeus: Hmm, okay. Don't really follow so I'll look into what you linked + the gossipsub module. So is essentially the entire current solution doa?

djordon: > Hmm, okay. Don't really follow so I'll look into what you linked + the gossipsub module. So is essentially the entire current solution doa?

Yeah, I do not think that we should tackle this using networking related changes. Let's start with a simple change that does not bail when we call `coordinate_dkg`, so delete the `?` in: https://github.com/stacks-sbtc/sbtc/blob/1413a5d2d0b27d203ab752bdc6a1fb15f4119995/signer/src/transaction_coordinator.rs#L345-L347 and log an error. You can open a new PR or repurpose this one for that, but it should be focused on not bailing.

For this connected peer stuff, this is my mistake. I had thought that this was the simplest way for us to reduce the cost of wanting to run DKG when we are not connected to all peers. But it probably isn't. Instead, I think that we should use more narrowly focused timeouts during the stages of DKG. And for that we'd want to first figure out how the `process_timeout` function works on the FIRE coordinator https://github.com/stacks-sbtc/wsts/blob/a03bfd874df5c8442aaab5fd79c37a57b3399e11/src/state_machine/coordinator/fire.rs#L67-L69, since maybe we can use that.

setzeus: Changed in [e214ffe](https://github.com/stacks-sbtc/sbtc/pull/1612/commits/e214ffe2bc7594600a978ab1be0abf526eaf0c36)

djordon (approved):
Looks good

MCJOHN974 (dismissed):
lgtm

signer/src/network/libp2p/event_loop.rs
cylewitruk: Looks good 👍

cylewitruk:
Also, could we update the PR description and title to reflect the change? This title and PR link end up in the changelog.

MCJOHN974 (dismissed):
lgtm

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4712,3 +4714,89 @@ mod get_eligible_pending_withdrawal_requests {
...
+    let mut coordinator = TxCoordinatorEventLoop {
+        context: context.clone(),
+        network: network.spawn(),
+        private_key: PrivateKey::new(&mut rng),
+        threshold: 3,
+        context_window: 5,
+        signing_round_max_duration: std::time::Duration::from_secs(5),
+        bitcoin_presign_request_max_duration: std::time::Duration::from_secs(5),
+        // extremely short duration so that DKG fails...
+        dkg_max_duration: Duration::from_millis(1),
```
cylewitruk, line 4785: I would probably set this to like 1ms or something to make absolutely certain that no combination of config/fast computer can cause this to succeed.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4712,3 +4714,89 @@ mod get_eligible_pending_withdrawal_requests {
...
+        is_epoch3: true,
+    };
+
+    // This should trigger DKG coordination which will fail, falling back to existing key
+    let result = coordinator.process_new_blocks().await;
```
cylewitruk, line 4790: To make this air-tight, you should probably listen to the `network` and make sure that you don't get 3 `DkgEnd { success == true }` messages, otherwise the test is just implicitly assuming that DKG is actually failing.

cylewitruk: Yeah that'd actually be quite a bit easier 👍 :)

setzeus: Copy. Though I should be checking that the return is none() correct? (aka DKG *did* run but it failed?)

cylewitruk: Nah you'd want to check for `is_some()`; it's the 'tx signer' side of things that manages these rows, and they only write a row (as `Unverified`) after the key-generation portion of DKG has completed iirc. So `Some(_)` would indicate that DKG did indeed run.

cylewitruk: Yeah, we jumped on a call and ultimately decided to skip adding this check for now given the complexity vs. getting this feature merged.

While the test _currently doesn't definitively assert that DKG has indeed run and **failed**_, current assumptions dictate that if `should_coordinate_dkg() == true` then the coordinator _should_ attempt DKG, and a timeout of `10ms` _should_ cause the coordinator to kill the DKG process prematurely, resulting in a failure.

cylewitruk (dismissed):
Looks OK. We're still not verifying in the test that DKG _did run and **fail**_, but the motivation is [here](https://github.com/stacks-sbtc/sbtc/pull/1612/files#r2073299269).

signer/tests/integration/transaction_coordinator.rs
djordon: Wait, why isn't checking that `db.get_latest_encrypted_dkg_shares().await` returns `None` sufficient? My thinking is this:
1. We know `should_coordinate_dkg() == true`, so DKG will be attempted.
2. With these changes, `TxCoordinatorEventLoop::process_new_blocks` should return `Ok(())` regardless of whether DKG succeeded or failed, assuming nothing else fails.
3. If DKG succeeds then we are guaranteed to get a row in the `dkg_shares` table by the time that `TxCoordinatorEventLoop::process_new_blocks` returns (this fact isn't obvious but I believe that it is true). So checking that `db.get_latest_encrypted_dkg_shares().await` is `None` and that `process_new_blocks` returned `Ok(())` tells us that DKG failed as expected and the other code paths in that function were executed, which is what we want.

Above I wrote that we want to check that `db.get_latest_encrypted_dkg_shares().await` is `Some(_)` but that may be misleading since we want it to be `None` since DKG shouldn't succeed.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4709,3 +4711,96 @@ mod get_eligible_pending_withdrawal_requests {
...
+    assert!(
+        result.is_ok(),
+        "process_new_blocks should complete successfully even with DKG failure"
+    );
+
+    // Check if DKG shares exist
+    let dkg_shares = storage.get_latest_encrypted_dkg_shares().await.unwrap();
```
djordon, line 4794: We should add a few more details here. Specifically, we should say something like:
>    // We know that when coordinator.process_new_blocks().await returns Ok,
>    // DKG will have run and either failed or succeeded. If it failed then
>    // we will not have a row in the dkg_shares table, so the following
>    // assertion completes the check that DKG has run and failed.

djordon (commented):
Nice, I would add some more comments to the test for what it is doing and why it works.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4709,3 +4711,107 @@ mod get_eligible_pending_withdrawal_requests {
         testing::storage::drop_db(db).await;
     }
 }
+
+// This test checks that blocks are still created
```
djordon, line 4715:
```suggestion
// This test checks that the coordinator attempts to fulfill its other duties
```

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4709,3 +4711,107 @@ mod get_eligible_pending_withdrawal_requests {
...
+        signing_round_max_duration: std::time::Duration::from_secs(5),
+        bitcoin_presign_request_max_duration: std::time::Duration::from_secs(5),
+        // short be short enough to broadcast, yet fail
+        dkg_max_duration: Duration::from_millis(10),
+        is_epoch3: true,
+    };
+
+    // We're verifying that the coordinator is currently
+    // processing blocks correctly. Since we previously checked
+    // that 'should_coordinate_dkg' will trigger & we set the
```
djordon, line 4791:
```suggestion
    // We're verifying that the coordinator is currently
    // processing requests correctly. Since we previously checked
    // that 'should_coordinate_dkg' will trigger and we set the
```

djordon (dismissed):
Looks good, left some nits on the comments.

MCJOHN974 (dismissed):
Reapproving, but pls don't forget to resolve nits from @djordon

signer/tests/integration/transaction_coordinator.rs
setzeus: Updated in [8787517](https://github.com/stacks-sbtc/sbtc/pull/1612/commits/8787517dfc69d0a9262929f68830c03fb0f1d3d0)

signer/tests/integration/transaction_coordinator.rs
setzeus: Updated in [8787517](https://github.com/stacks-sbtc/sbtc/pull/1612/commits/8787517dfc69d0a9262929f68830c03fb0f1d3d0)

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4709,3 +4711,59 @@ mod get_eligible_pending_withdrawal_requests {
         testing::storage::drop_db(db).await;
     }
 }
+
+#[test_log::test(tokio::test)]
+async fn should_handle_dkg_coordination_failure() {
```
MCJOHN974, line 4719: I don't really see why we will experience dkg failure in this test. Maybe add some comments about it

MCJOHN974: Ohhh, you rebased, never mind

djordon (approved):
Looks good ✅

PR #1614, opened 2025-04-17
[chore]: use wrapper structs for heights

MCJOHN974:
It is a wrapper over main PR: https://github.com/stacks-sbtc/sbtc/pull/1448

The intermediate PR: https://github.com/MCJOHN974/sbtc/pull/1

### DIFF
--- a/signer/src/api/info.rs
+++ b/signer/src/api/info.rs
@@ -11,7 +11,7 @@ use crate::{
     stacks::api::StacksInteract,
     storage::{
         DbRead,
-        model::{BitcoinBlockHash, StacksBlockHash},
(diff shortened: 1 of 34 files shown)


cylewitruk (approved):
Went through the list of commits and they seem to match the original PR, so shouldn't be anything strange. Approving here as the original PR is also approved.

djordon (approved):
I checked that the last commit hash in the https://github.com/MCJOHN974/sbtc/pull/1 PR matches the last one in https://github.com/stacks-sbtc/sbtc/pull/1448 PR, and then checked that the merged commit hash in the https://github.com/MCJOHN974/sbtc/tree/viktar/heights-wrappers branch matches the one here.

I would like to add that I checked the full commit hashes, not the shortened commit hashes.

PR #1618, opened 2025-04-22
chore: make bootstrap_signer_set containing self pubkey mandatory

MCJOHN974:
## Description

This PR makes mandatory for `bootstrap_signer_set` to contain public key of this signer.

Closes: no issue.

## Motivation

Currently, `SignerConfig` have both method `bootstrap_signing_set()` and similarly named field `bootstrap_signing_set`. They also have different types, `BTreeSet<Pubkey>` vs `Vec<Pubkey>`. This can give an impression, that the method just converts `Vec` to `BTreeSet`. But it is wrong, and method also adds one more pubkey to set, which is counterintuitive imo.

I did some lookup through the code, found zero places relying on `bootstrap_signing_set` does not contain pubkey of current signer, so I made `bootstrap_signing_set` always containing the pubkey of current signer, via settings validation.

## Changes

- Now `bootstrap_signing_set()` always contain same pubkeys as `bootstrap_signing_set` and only move them from `Vec` to `BTreeSet`.
- Now signer will panic during config validation if `bootstrap_signing_set` doesn't contain pubkey of this signer.

## Testing Information

- Added one test checking that we panic if `bootstrap_signing_set` doesn't contain pubkey of this signer.
- Added some changes to existing tests to make them complying new rule of config validation.

## Checklist:

### DIFF
--- a/signer/src/config/mod.rs
+++ b/signer/src/config/mod.rs
@@ -360,6 +360,12 @@ pub struct SignerConfig {
 impl Validatable for SignerConfig {
     fn validate(&self, cfg: &Settings) -> Result<(), ConfigError> {
         self.p2p.validate(cfg)?;
+
+        if !self.bootstrap_signing_set().contains(&self.public_key()) {
+            let err = SignerConfigError::MissingPubkeyInBootstrapSignerSet;
+            return Err(ConfigError::Message(err.to_string()));
+        }
+
         if self.deployer.is_mainnet() != self.network.is_mainnet() {
             let err = SignerConfigError::NetworkDeployerMismatch;
             return Err(ConfigError::Message(err.to_string()));
@@ -428,17 +434,10 @@ impl Validatable for SignerConfig {
 }
 
 impl SignerConfig {
-    /// Return the bootstrapped signing set from the config. This function
-    /// makes sure that the signing set includes the current signer.
+    /// Return the bootstrapped signing set from the config.
+    /// This function just convert [`Vec`] to [`BTreeSet`]
     pub fn bootstrap_signing_set(&self) -> BTreeSet<PublicKey> {
-        // We add in the current signer into the signing set from the
-        // config just in case it hasn't been included already.
-        let self_public_key = PublicKey::from_private_key(&self.private_key);
-        self.bootstrap_signing_set
-            .iter()
-            .copied()
-            .chain([self_public_key])
-            .collect()
+        self.bootstrap_signing_set.iter().copied().collect()
     }
 
     /// Return the public key of the signer.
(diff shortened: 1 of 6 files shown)


signer/src/config/mod.rs
```diff
@@ -428,17 +434,10 @@ impl Validatable for SignerConfig {
 }
 
 impl SignerConfig {
-    /// Return the bootstrapped signing set from the config. This function
-    /// makes sure that the signing set includes the current signer.
+    /// Return the bootstrapped signing set from the config.
+    /// This function just convert [`Vec`] to [`BTreeSet`]
     pub fn bootstrap_signing_set(&self) -> BTreeSet<PublicKey> {
```
djordon, line 439: We can deserialize to a `BTreeSet` if we want to. Then we could remove this function entirely. The only thing about that comes to mind about such a change is that we wouldn't be able to warn the operator if there were duplicates, but we don't do that now anyway.

signer/src/config/mod.rs
```diff
@@ -756,17 +755,39 @@ mod tests {
     fn default_config_toml_loads_signer_private_key_config_with_environment() {
         clear_env();
 
-        let new = "a1a6fcf2de80dcde3e0e4251eae8c69adf57b88613b2dcb79332cc325fa439bd";
+        let new = "9bfecf16c9c12792589dd2b843f850d5b89b81a04f8ab91c083bdf6709fbefee01";
```
djordon, line 758: Why add those two bytes at the end here?

djordon: Oh, you used the private key associated with one of the keys in the bootstrap signing set so that these tests pass. Still, we should lop off the byte, it is a nonstandard private key representation.

djordon (dismissed):
This looks fine, just floating the idea of deserializing to a `BTreeSet` here if we want to.

signer/src/config/mod.rs
MCJOHN974: Idk.. For me is fine to have both `Vec` from field and `BTreeSet` from function.

signer/src/config/mod.rs
MCJOHN974: I'm stripping two bytes programmatically. To advocate for what I have now, current key can be searched and found in some of our readme files, describing what this key is, but I don't mind to delete this bytes manually

signer/src/config/mod.rs
cylewitruk: I'd be for deserializing directly to `BTreeSet`. We really don't care if there are duplicates, i.e. it's not an "error" per-se that requires preventing startup.

MCJOHN974: Yeah, I tried to imagine why we may want to have `Vec` but failed. Here is migration to just btreeset: [27ab903](https://github.com/stacks-sbtc/sbtc/pull/1618/commits/27ab90374f0d308030bcde3b3dc2dfa62ea22ed3)

cylewitruk (approved):
LGTM

djordon (approved):
Looks good

Issue #1319, opened 2025-02-07
[Feature]: Add supporting for backing-up and restoring encrypted DKG shares


aldur:
# Feature - Add supporting for backing-up and restoring encrypted DKG shares

## 1. Description

DKG shares do not _currently_ depend on the private key specified in the config, but are generated through a PRNG. 

They are stored, encrypted, in the DB. 

A signer that loses the DB would mean losing a share of the corresponding aggregate key.

### 1.1 Context & Purpose

Long-term, we should consider deterministically deriving DKG shares based on some recoverable seed. See https://github.com/stacks-network/sbtc/issues/1303.

Short-term, we should make it easy for sBTC signers to:

1. Export their encrypted DKG shares.
2. Restore their encrypted DKG shares.

This would make backup operations easier.

### 2.1 Acceptance Criteria:

- As an sBTC signer, I can run the sBTC binary with a CLI flag that allows me to export the encrypted DKG shares.
- As an sBTC signer, I can run the sBTC binary with a CLI flag that allows me to restore the encrypted DKG shares (adding to existing ones or failing if the DB is already populated).

PR #1619, opened 2025-04-22
feat: add CLI support for backing up and restoring critical signer state

cylewitruk:
## Description

Closes: #1319

## Changes

Adds new CLI commands for backing-up and restoring critical signer state to and from files, which in its current form includes the encrypted DKG shares but leaves room for inclusion of other data.

This PR _includes_ #1610 (the protos PR), which introduces a generic backup file format and necessary conversions.

This needed some slight restructuring of `main.rs` to support multiple commands. I tried to "hide away" most of the CLI stuff in a separate `cli` mod, which also provides a relatively clean structure for adding additional commands in the future. I also took the liberty of making sure the remaining methods in the file are well-documented.

## Testing Information

I still need to do some more local "real" testing, but figured I'd get some input from the team before spending too much more time on it.

## Checklist:


protobufs/sbtc/signer/v1/backups.proto
djordon, line 1: We can move the other files to this directory (or a similarly named one) or move this one. This way this file lives close to it's peers. The current naming convention is `organization / application / version / file`. I think we should change the `application` to be `sbtc` rather than change or remove the `organization`.

signer/src/proto/convert.rs
```diff
@@ -585,6 +586,96 @@ impl TryFrom<proto::StacksTransactionSignRequest> for StacksTransactionSignReque
...
+
+        // Convert DKG status enum
+        let dkg_shares_status =
+            proto::sbtc::signer::v1::backups::DkgSharesStatus::try_from(value.dkg_shares_status)
+                .map_err(|_| Error::TypeConversion)?; // Handle invalid enum value
+
+        let db_dkg_shares_status = match dkg_shares_status {
+            proto::sbtc::signer::v1::backups::DkgSharesStatus::Unspecified => {
+                Err(Error::TypeConversion)
+            } // Or handle as default?
```
djordon, line 643: What is here is correct since it's a programming error if we see this.

signer/src/proto/convert.rs
```diff
@@ -585,6 +586,96 @@ impl TryFrom<proto::StacksTransactionSignRequest> for StacksTransactionSignReque
...
+        let signer_set_public_keys = value
+            .signer_set_public_keys
+            .into_iter()
+            .map(|pk| pk.try_into())
+            .collect::<Result<Vec<_>, _>>()?;
+
+        // Convert DKG status enum
+        let dkg_shares_status =
+            proto::sbtc::signer::v1::backups::DkgSharesStatus::try_from(value.dkg_shares_status)
+                .map_err(|_| Error::TypeConversion)?; // Handle invalid enum value
```
djordon, line 638: There should be a helper function implemented on `DkgShares` to do the conversion. I think it'll be called `value.dkg_shares_status()` or something.

protobufs/sbtc/signer/v1/backups.proto
```diff
@@ -0,0 +1,99 @@
...
+    crypto.PublicKey tweaked_aggregate_key = 2;
+
+    // Encrypted private shares of the aggregate key
+    bytes encrypted_private_shares = 3;
+
+    // Public shares used for verification
+    bytes public_shares = 4;
+
+    // The Bitcoin script_pubkey derived from the tweaked aggregate key
+    bytes script_pubkey = 5;
```
djordon, line 35: Might be worth adding a `script_pubkey` type just for this. But I'm okay either way, the benefit doesn't seem particularly high.

djordon (commented):
I haven't gotten to the meat of this yet, but my main question so far is why not have this CLI stuff in a separate binary?

protobufs/sbtc/signer/v1/backups.proto
cylewitruk: Hm, I was thinking more along the lines of the organization changing from `stacks` to `sbtc` (we have our own github org afterall), and the rest remaining the same...

protobufs/sbtc/signer/v1/backups.proto
cylewitruk: I had thought about it, but ended up in that it's just a bunch of bytes anyway...

signer/src/main.rs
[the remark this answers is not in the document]
cylewitruk: Yeah I wanted to avoid extra burden on the signers and the risk of some outdated CLI binary getting used against a newer signer db, since this isn't like `bitcoin-cli` which is just a glorified RPC client. This keeps things nicely packaged.

signer/src/cli/backups/restore.rs
```diff
@@ -0,0 +1,567 @@
...
+            println!(
+                "NOTE: Skipping import of existing shares for aggregate key '{}'",
+                shares.aggregate_key
+            );
+            continue;
+        }
+
+        // Write the non-existing shares to the db
+        println!(
+            "Importaing shares for aggregate key '{}'",
```
MCJOHN974, line 199:
```suggestion
            "Importing shares for aggregate key '{}'",
```

signer/src/lib.rs
```diff
@@ -10,6 +10,7 @@ pub mod api;
 pub mod bitcoin;
 pub mod block_observer;
 pub mod blocklist_client;
+pub mod cli;
```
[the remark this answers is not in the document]
djordon: I was originally thinking that this functionality would be added to a separate binary. But after speaking with you at stand up, it isn't so clear that this should be in it's own binary. The first thing is that this functionality doesn't interact with the signer, but instead interacts with the signer's database. The second is that there is value in having this all bundled into one binary. Maybe sometime in the future it will make more sense to feature flagging this and some other stuff and put it into a separate binary but for now I cannot make that case.

protobufs/sbtc/signer/v1/backups.proto
djordon: Yeah that makes sense. Let's use `sbtc` as the organization and keep the rest the same.

PR #1623, opened 2025-04-24
[fix]: Emily don't return Error on limits/ if rolling window calculation failed

MCJOHN974:
## Description

Closes: #?

## Checklist:

### DIFF
--- a/emily/handler/src/database/accessors.rs
+++ b/emily/handler/src/database/accessors.rs
@@ -838,7 +838,7 @@ pub async fn get_limits(context: &EmilyContext) -> Result<Limits, Error> {
         global_cap.rolling_withdrawal_blocks,
         global_cap.rolling_withdrawal_cap,
     )
-    .await?;
+    .await.ok().flatten();
 
     // Get the global limit for the whole thing.
     Ok(Limits {


emily/handler/src/database/accessors.rs
```diff
@@ -838,7 +844,9 @@ pub async fn get_limits(context: &EmilyContext) -> Result<Limits, Error> {
         global_cap.rolling_withdrawal_blocks,
         global_cap.rolling_withdrawal_cap,
     )
-    .await?;
+    .await
+    .ok()
+    .flatten();
```
MCJOHN974, line 849: This will convert any error to None and will allow to return other limits

cylewitruk: Could we add an `inspect_err` so that it's at least logged?

cylewitruk: Oh, I guess you do that above. But you could explicitly log here that you're returning `None` from get limits due to chainstate failure.

cylewitruk: We need to check how the bridge handles `None` as well.

MCJOHN974: Yeah, `inspect_err` worth adding. However, I suspect it will be always `Error::NotFound`

MCJOHN974: In general I'd suggest instead of treating `None` as no limit here and in other limits as well, treat `None` as error, and `Some(u64::MAX)` as no limit

cylewitruk: Nice, closing this now based on standup discussion 👍 The bridge will effectively fall-back to the per-withdrawal limits with this, and the signers will do their own thing (calculations).

cylewitruk (approved):
Looks good and in-line with the discussion at standup:
- Returns `None` upon chainstate-retrieval error, instead of `404 Not Found`
- Signers will continue receiving static limits and calculating their own rolling withdrawals.

emily/handler/src/database/accessors.rs
```diff
@@ -838,7 +844,12 @@ pub async fn get_limits(context: &EmilyContext) -> Result<Limits, Error> {
         global_cap.rolling_withdrawal_blocks,
         global_cap.rolling_withdrawal_cap,
     )
-    .await?;
+    .await
+    .inspect_err(|err| {
+        warn!(%err, "calculate_sbtc_left_for_withdrawals returned error");
```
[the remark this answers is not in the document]
djordon: We spoke about this during a meeting and it's okay to return `None` here.

djordon (approved):
Looks good.

Just a few non blocking nits.

djordon (approved):

cylewitruk (approved):
👍

PR #1633, opened 2025-04-28
Add composable scenario tests for signer using madhouse

moodmosaic:
Adds `madhouse`-based scenario testing to signer.

Provides reusable, DRY test steps that can be composed into scenarios.

First example:
```rust
scenario![
    test_context,
    CreateFailedDkgShares,
    WriteDkgShares,
    VerifyDkgVerificationFailed,
];
```

Based on property-based state machine testing (see [spec](https://github.com/moodmosaic/stacks-core/commit/1cb033c39947d8b6c999fcf68fca3009db2e3263)).


signer/Cargo.toml
```diff
@@ -62,6 +62,8 @@ wsts.workspace = true
 # Only for testing
 fake = { workspace = true, optional = true }
 mockall = { workspace = true, optional = true }
+madhouse = { git = "https://github.com/stacks-network/madhouse-rs.git", rev = "8623f2a4efc6a8b691f25bc95eb9801fec17b4c0" }
```
cylewitruk, line 65: Please add this to the workspace `Cargo.toml` instead and use `madhouse.workspace = true` here.

Does this need to be a full dependency, or will it work as a `[dev-dependency]`? If it needs to be a full dependency, then please also add `optional = true` and include it in the `testing` feature.

signer/Cargo.toml
```diff
@@ -62,6 +62,8 @@ wsts.workspace = true
...
+proptest = { git = "https://github.com/proptest-rs/proptest.git", rev = "c9bdf18c232665b2b740c667c81866b598d06dc7" }
```
cylewitruk, line 66: We already have `proptest` in the workspace, so `proptest.workspace = true` should be fine. This should be fine as a `[dev-dependency]` as well, no?

cylewitruk (commented):
Will look more closely later, just tossing out a couple of quick comments on the dependency handling.

This repo has the convention of defining all dependencies in the workspace `Cargo.toml`, where dependencies in the `[dependencies]` stanza (i.e. full dependencies) are always specified with `default-features = false` with explicit feature enablement.

You'll also need to `cargo vet` and handle the dependency diffs.

signer/Cargo.toml
BowTiedRadone: Addressed in a1b519229b4c58bcde9d8fc82f2ebf9b4e0ed29b.

signer/Cargo.toml
BowTiedRadone: Addressed in a1b519229b4c58bcde9d8fc82f2ebf9b4e0ed29b.

signer/tests/integration/commands/dkg.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    fn apply(&self, state: &mut TestState) {
+        let mut rng = StdRng::seed_from_u64(self.seed);
+        let aggregate_key: PublicKey = Keypair::new_global(&mut rng).public_key().into();
+        let shares = EncryptedDkgShares {
+            aggregate_key,
+            dkg_shares_status: DkgSharesStatus::Failed,
+            started_at_bitcoin_block_height: 0u64.into(),
+            ..Faker.fake()
```
moodmosaic, line 41: I think this won’t replay correctly? I’d have to check closer, or perhaps @cylewitruk knows, but I think this uses its own prng internally?

BowTiedRadone: Good point. Addressed in a5df54da2a76fd8acdbdef4f1dadf9e4ec9373b8.

BowTiedRadone:
@cylewitruk: It seems no vetting is needed after moving `madhouse-rs` to dev dependencies. `cargo vet` output:
```
$ cargo vet
Vetting Succeeded (147 fully audited, 4 partially audited, 559 exempted)
 WARN Your supply-chain has unnecessary exemptions which could be relaxed or pruned.
 WARN   Consider running `cargo vet prune` to prune unnecessary exemptions and imports.
 ```

MCJOHN974:
> @cylewitruk: It seems no vetting is needed after moving `madhouse-rs` to dev dependencies. `cargo vet` output:

Try to run `cargo vet prune` and it should fix problems

djordon-stacks:

PR #1640, opened 2025-04-28
fix: add compound primary key to `rotate_keys_transactions` table

djordon:
## Description

This was supposed to be done in https://github.com/stacks-sbtc/sbtc/pull/1599, but was missed. The bug would make possible for us to drop key rotation events when they get replayed following a fork.

## Changes

* Update the primary key of the `rotate_keys_transactions` table to be a compound primary key on the `txid` and `block_hash` table.

## Testing Information

This PR adds an integration test to check that we can insert two events with the same transaction ID into the database.

## Checklist:


MCJOHN974 (approved):
I didn't reallly got how this test and this sql changes are related... But separately both seems legitimate

djordon:
> I didn't reallly got how this test and this sql changes are related... But separately both seems legitimate

Oh, before the you could only have one row per transaction ID and in the test I create two rows with the same transaction ID (but with different block hashes). But it's not that obvious, so let me fix that.

MCJOHN974:
> > I didn't reallly got how this test and this sql changes are related... But separately both seems legitimate
> 
> Oh, before the you could only have one row per transaction ID and in the test I create two rows with the same transaction ID (but with different block hashes). But it's not that obvious, so let me fix that.

Oh, txid was primary key so it was impossible to create two similar txids, got it. Why we want to add two same txids in different rows ? It seems a bit strange..

djordon:
> Oh, txid was primary key so it was impossible to create two similar txids, got it. Why we want to add two same txids in different rows ? It seems a bit strange..

You can have the same transaction confirmed on two different blocks. This can happen if the first block confirming the transaction gets orphaned, which may happen when a bitcoin fork causes stacks to orphan some blocks. When the transaction gets reconfirmed, we will receive another event from the stacks node.

For key rotation events, we used to ignore the second event and just join on the `stacks_transactions` table to figure out if the event was associated with a transaction that was confirmed on the canonical stacks blockchain. This approach, just storing a mapping between transactions and the stacks blocks confirming them, is storage efficient for duplicates, since you don't need a duplicate of the event in the db. However, its not storage efficient overall since you usually don't have forks or duplicates. It also adds complexity to our queries and the codebase. So overall I think it's better to just store a duplicate.

MCJOHN974 (approved):
lgtm

cylewitruk (approved):
lgtm 👍

PR #1642, opened 2025-04-28
feat: make the `getblock` RPC return detailed transacation info

djordon:
## Description

This is part of https://github.com/stacks-sbtc/sbtc/issues/1605, but does not address it.

This PR makes some `BitcoinTxInfo` fields optional since coinbase transactions omit them; this way we can reuse the same struct for both `getblock` and `getrawtransaction` responses. We also dropped rarely used fields to reduce memory overhead when processing blocks.

## Changes

* Remove several fields from the `BitcoinTxInfo` struct. It now only contains what is necessary, except for the `fee` field which we keep for convenience.
* Make some fields in the `BitcoinTxInfo` optional. Those fields were never guaranteed to be present, so could have been omitted by bitcoin core.
* Add a function that validates that we have the optional, but required-for-us fields, in `BitcoinTxInfo`.
* Update the `getblock` command to return a detailed transaction objects for a block.

## Testing Information

This is largely a refactor, but we should do some testing on private mainnet before shipping this. According to this ticket (https://github.com/bitcoin/bitcoin/issues/30495) we'll take a ~30x performance slowdown compared to the status quo. And we want mainnet testing because we have full blocks there.

## Checklist:


signer/src/bitcoin/rpc.rs
```diff
@@ -60,77 +56,161 @@ pub struct GetTxResponse {
...
+}
+
+/// The previous output, omitted if block undo data is not available.
+///
+/// The `height` field is included in the actual response, but omitted
+/// here.
+#[derive(Clone, PartialEq, Eq, Debug, serde::Deserialize, serde::Serialize)]
+pub struct BitcoinTxVinPrevout {
+    /// Whether the previous output is from a coinbase transaction or not.
+    pub generated: bool,
```
MCJOHN974, line 141: Maybe `is_coinbase` will be better naming here ?

signer/src/testing/dummy.rs
```diff
@@ -108,6 +114,112 @@ pub fn block<R: rand::RngCore + ?Sized>(
...
+impl Dummy<Faker> for BitcoinBlockInfo {
+    fn dummy_with_rng<R: Rng + ?Sized>(_: &Faker, rng: &mut R) -> Self {
+        let block_height = (0..i64::MAX as u64).choose(rng).unwrap();
+        BitcoinBlockInfo::random_with_height(block_height, rng)
+    }
+}
+
+impl BitcoinBlockInfo {
+    /// Create a random bitcoin block with the given height
+    pub fn random_with_height<R: Rng + ?Sized>(block_height: u64, rng: &mut R) -> Self {
```
MCJOHN974, line 190: Maybe use `BitcoinBlockHeight` instead of u64 ?

MCJOHN974 (dismissed):

signer/src/testing/dummy.rs
djordon: Yeah, this "new" function replaces an existing one that does basically the same thing as the one that it replaces. The old function used `u64` as the `block_height`, so I basically just copied the type signature. But I'm okay with either changing it or leaving it.

signer/src/bitcoin/rpc.rs
djordon: I started to rename it, but then realized that it is unnecessary, so I removed the field instead.

djordon: https://github.com/stacks-sbtc/sbtc/pull/1642/commits/f1fbc5ba6628b80c04b72bee784ad895e6a3573a

signer/src/testing/dummy.rs
MCJOHN974: I think "use wrapper instead of raw int where possible" can be a good rule of thumb for us

signer/src/testing/dummy.rs
```diff
@@ -179,16 +179,17 @@ impl Dummy<bitcoin::TxIn> for BitcoinTxVin {
 
 impl Dummy<Faker> for BitcoinBlockInfo {
     fn dummy_with_rng<R: Rng + ?Sized>(_: &Faker, rng: &mut R) -> Self {
-        let block_height = (0..i64::MAX as u64).choose(rng).unwrap();
-        BitcoinBlockInfo::random_with_height(block_height, rng)
+        let height = (0..i64::MAX as u64).choose(rng).unwrap();
```
MCJOHN974, line 182: I think you can do something like `let height: BitcoinBlockHeight = faker.fake_with_rng(&mut rng);`

djordon: Oh right, will update.

MCJOHN974 (dismissed):

MCJOHN974 (approved):

matteojug (dismissed):

signer/src/bitcoin/rpc.rs
```diff
@@ -146,6 +224,38 @@ pub struct BitcoinTxFeeInfo {
...
+    /// The hash of the consensus encoded header of the block.
+    #[serde(rename = "hash")]
+    pub block_hash: BlockHash,
+    /// The number of blocks preceding this one on the blockchain that
+    /// includes this block.
+    pub height: BitcoinBlockHeight,
+    /// The Unix epoch time when the block was mined. It reflects the
+    /// timestamp as recorded by the miner of the block.
+    pub time: u64,
+    ///The median block time expressed in UNIX epoch time
```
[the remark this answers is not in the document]
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1642/commits/4738aec252628ae99a63a200d338aeda04e11b43

signer/src/block_observer.rs
```diff
@@ -443,26 +444,23 @@ impl<C: Context, B> BlockObserver<C, B> {
...
             sbtc_txs.push(model::BitcoinTxRef {
                 txid: txid.into(),
                 block_hash: block_hash.into(),
             });
+            // Bail if bitcoin-core doesn't return all the data that we
+            // care about for a non-coinbase transaction. This will happen
+            // if bitcoin core hasn't computed the undo data for the block
+            // with these transactions, of it there is a bug in bitcoin
+            // core.
+            tx_info.validate()?;
```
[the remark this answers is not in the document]
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1642/commits/1940a1e2a0b830a2d33bda465de0cdc47cd0badd

signer/src/testing/dummy.rs
```diff
@@ -108,6 +114,112 @@ pub fn block<R: rand::RngCore + ?Sized>(
...
+
+        let tx_in = bitcoin::TxIn {
+            previous_output: OutPoint {
+                txid: config.fake_with_rng::<BitcoinTxId, _>(rng).into(),
+                vout: (1..=50u32).choose(rng).unwrap(),
+            },
+            script_sig: ScriptBuf::new(),
+            sequence: bitcoin::Sequence(0),
+            witness: bitcoin::Witness::new(),
+        };
```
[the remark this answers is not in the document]
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1642/commits/eae595c9f01bf4b08af0e999a58e3cf276a27330

signer/src/bitcoin/rpc.rs
```diff
@@ -732,3 +805,105 @@ impl BitcoinInteract for BitcoinCoreClient {
...
+        // Let's add another input so that they can be disordered.
+        let tx_in = bitcoin::TxIn {
+            previous_output: OutPoint {
+                txid: Faker.fake_with_rng::<BitcoinTxId, _>(&mut rng).into(),
+                vout: Faker.fake_with_rng(&mut rng),
+            },
+            script_sig: ScriptBuf::new(),
+            sequence: bitcoin::Sequence::MAX,
+            witness: bitcoin::Witness::new(),
+        };
```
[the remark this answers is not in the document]
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1642/commits/eae595c9f01bf4b08af0e999a58e3cf276a27330

signer/src/block_observer.rs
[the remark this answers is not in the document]
djordon: I update the order again in https://github.com/stacks-sbtc/sbtc/pull/1669 to lead with the is coinbase check and validation before anything else.

matteojug (approved):

cylewitruk (approved):
lgtm

PR #1650, opened 2025-05-05
chore: add uv and python venvs installation

Jiloc:
This PR is the first step towards integrating Python unit and integration tests into the main `make` targets and the CI/CD pipeline.

It refactors the existing `make install` target by splitting it into two distinct targets: `install-py` for Python dependencies managed by `uv`, and `install-pnpm` for Node.js dependencies managed by `pnpm`. The main `install` target now simply calls `install-pnpm` and will integrate the `install-py` in a followup PR.

This refactoring is necessary because the `Dockerfile` currently executes `make install` after checking out the `main` branch. The full `make install` (including the new Python setup) would fail in the Docker build (with `make integration-env-build`) as it introduces a dependency on `uv`, which is not installed in the container and isn't required for the Docker build steps.

https://github.com/stacks-sbtc/sbtc/blob/f1b608f2a716ee6307204b6cc2fbb7506ca6b753/docker/sbtc/Dockerfile#L22-L31

By splitting the install process, a subsequent PR can safely modify the `Dockerfile` to use only `make install-pnpm`, ensuring the Docker build process only installs the necessary `pnpm` dependencies without requiring `uv` and without breaking the build.

## Changes

-   **Makefile:**
    -   Created a new `install-py` target responsible for creating Python virtual environments (`uv venv`) and syncing dependencies (`uv pip sync`) for `emily_cron` and `emily_sidecar`.
    -   Created a new `install-pnpm` target responsible for running `pnpm --recursive install`.
-   **README.md:**
    -   Added `uv` to the list of required developer tools.
    -   Updated the tool version check command and example output to include `uv`.

## Testing Information

-   Verified locally that `make install` successfully executes both `make install-py` and `make install-pnpm`.
-   Verified locally that `make install-py` correctly creates/updates Python environments and dependencies.
-   Verified locally that `make install-pnpm` correctly installs Node dependencies.
-   Confirmed that the Docker build process is not yet changed in this PR and will be addressed subsequently.

## Checklist:

-   [x] I have performed a self-review of my code
-   [x] My changes generate no new warnings
-   [x] New and existing unit tests pass locally with my changes
-   [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ The following are the developer tools that you should install on your local mach
 - **[Cargo-lambda](https://www.cargo-lambda.info/)** - [Installation Guide](https://www.cargo-lambda.info/guide/getting-started.html) - Compile the package for AWS Lambda.
 - **[pnpm](https://pnpm.io)** - [Installation guide](https://pnpm.io/installation) - Manages node packages.
 - **[Make](https://www.gnu.org/software/make/)** - Development task runner; natively present on nearly every system.
-- **[Docker](https://docs.docker.com/manuals/)** - [Installation Guide](https://docs.docker.com/desktop/). This is used for running integration tests.
-- **[protoc](https://github.com/protocolbuffers/protobuf)** - [Installation Guide](https://grpc.io/docs/protoc-installation/). Compiles protobuf files.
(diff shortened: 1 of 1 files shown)


djordon (approved):
✅ Looks good

MCJOHN974 (approved):
lgtm

PR #1652, opened 2025-05-05
chore: remove openssl

MCJOHN974:
## Description

Closes: no issue

## Changes

removing unused dependency from `Cargo.toml`

## Testing Information

no new tests needed

## Checklist:


cylewitruk (approved):
Looks fine, I notice there's no `Cargo.lock` or changes to cargo vet's files, so this likely means that we're not _actually_ getting rid of the dependency. But I'm happy to not have it in our list of direct dependencies anyway.

matteojug (approved):
> Looks fine, I notice there's no Cargo.lock or changes to cargo vet's files, so this likely means that we're not actually getting rid of the dependency. But I'm happy to not have it in our list of direct dependencies anyway.

I guess we didn't really have that specific version as dependency, therefore it wasn't present at all in the lock or the audits? 🤷 

It would be nice to have a process to check if we have other similar cases.

Issue #1648, opened 2025-05-01
[Feature]: Process all requests during deposit or withdrawal updates


djordon:
# Feature - Process all requests during deposit or withdrawal updates

## 1. Description

The Emily endpoints that process deposit or withdrawal updates do so in batches. She currently bails when processing an update and encounters an issue, abandoning any of the remaining updates. However, we shouldn’t bail when a signer attempts to update the status of a deposit or withdrawal request in Emily and attempts to change the state from `Accepted` to `Accepted`. This can happen from honest behavior from the signers, where a signer, acting as coordinator, attempts to RBF an existing sweep transaction in the mempool and does not know that Emily has been updated. Emily shouldn’t update her database when there is no state change, but she should continue processing the remaining request updates in the batch.

## 2. Technical Details:

In Emily, when we know that the signer is attempting to change the state from `Accepted` to `Accepted`, we should not bail with an error.

### 2.1 Acceptance Criteria:

- [ ]  Emily processes all request updates when a signer attempts to update requests where some of them are already known to having been `Accepted` in Emily.
- [ ]  Emily processes all request updates when a signer attempts to update deposit requests where some of them are completely unknown. Return a 404 Not Found if all of them are unknown, and a 200 if at least one was known.
- [ ]  Emily should return a 200 status code instead of a 201 on success for the update endpoints.

## 4. Appendix

I’m not sure if we want to continue processing requests if Emily gets a DB error or something more serious during processing. I’m inclined to think that she should.

cylewitruk:
I would probably consider returning a `207 Multi-Status` with a JSON body (please just die, WebDAV...) including the individual status of each update so that the signer _has the option to_ handle it accordingly.

djordon:
> I would probably consider returning a `207 Multi-Status` with a JSON body (please just die, WebDAV...) including the individual status of each update so that the signer _has the option to_ handle it accordingly.

Good point, lets do that.

djordon:
> I would probably consider returning a `207 Multi-Status` with a JSON body

Oops, I didn't fully realize what you meant here. When you return a 207 status code you're supposed to return XML. Hmmm, I'm no longer sure that we should do that.

cylewitruk:
> > I would probably consider returning a `207 Multi-Status` with a JSON body
> 
> Oops, I didn't fully realize what you meant here. When you return a 207 status code you're supposed to return XML. Hmmm, I'm no longer sure that we should do that.

Yeah, hence my comment:
[...]

PR #1653, opened 2025-05-07
feat: Emily process all update events.

MCJOHN974:
## Description

Closes: #1648

## Changes

- Now Emily process other update events in a batch if some of events in a batch are unknown instead of bailing
- Now Emily allows updates for statuses `Accepted` => `Accepted` (however it silently ignores them)
- Now successfull update requests return 200 instead of 201. `UpdateReqwestResponce` now is a `Vec` of wrapper structs which look like `Status + Deposit/Withdrawal`. Statuses are 200 if deposit/withdrawal updated and 404 if deposit/withdrawal unknown, or other status according to error occured

## Testing Information

- Added some integration tests to Emily

## Checklist:


MCJOHN974:
Note: I did necessary changes _only_ for deposits. But withdrawals should be exactly same so I'll add them after most review notes will be addressed

emily/handler/src/api/models/deposit/responses.rs
```diff
@@ -30,5 +30,15 @@ pub struct GetDepositsResponse {
...
+    pub deposits: Vec<DepositWithStatus>,
+}
+
+/// Wrapper for deposit with status code. Used for multi-status responses.
+#[derive(Clone, Default, Debug, PartialEq, Hash, Serialize, Deserialize, ToSchema, ToResponse)]
+#[serde(rename_all = "camelCase")]
+pub struct DepositWithStatus {
+    /// Deposit itself.
+    pub deposit: Deposit,
+    /// Status code.
```
djordon, line 42:
```suggestion
    /// The status code of update operation associated with the
    /// above deposit request.
```

emily/handler/src/api/models/deposit/responses.rs
```diff
@@ -30,5 +30,15 @@ pub struct GetDepositsResponse {
...
-    pub deposits: Vec<Deposit>,
+    pub deposits: Vec<DepositWithStatus>,
+}
+
+/// Wrapper for deposit with status code. Used for multi-status responses.
+#[derive(Clone, Default, Debug, PartialEq, Hash, Serialize, Deserialize, ToSchema, ToResponse)]
+#[serde(rename_all = "camelCase")]
+pub struct DepositWithStatus {
+    /// Deposit itself.
+    pub deposit: Deposit,
```
djordon, line 41:
```suggestion
    /// The parsed deposit request.
    pub deposit: Deposit,
```

emily/handler/src/database/accessors.rs
```diff
@@ -201,7 +202,10 @@ pub async fn pull_and_update_deposit_with_retry(
         if update.is_unnecessary(&deposit_entry) {
             return Ok(deposit_entry);
         }
-        if !is_trusted_key && deposit_entry.status != Status::Pending {
+        let is_valid_untrusted_status_update = deposit_entry.status == Status::Pending
+            || update.event.status == StatusEntry::Accepted
+                && deposit_entry.status == Status::Accepted;
```
djordon, line 207: I think this needs to be
```rust
let is_valid_untrusted_status_update = update.event.status == StatusEntry::Accepted &&
    (deposit_entry.status == Status::Pending || deposit_entry.status == Status::Accepted);
```

Well, maybe we need to allow an update if the current status is `Reprocessing` too, I forget if we actually use that state.

djordon (commented):
Started the review and had some comments.

emily/handler/src/api/models/deposit/responses.rs
MCJOHN974: Hmm, tbh I don't really like your wording but I agree we should annotate that it is http status, not sbtc (`Pending`, `Accepted`, etc) status

emily/handler/src/api/models/deposit/responses.rs
MCJOHN974: Why it is "parsed"? I'm not sure such comment bring a lot of value to reader and... Where we parse it? This struct is not related to some parsing functions or so, so I don't think it is parsed.... This `Deposit` is essentially same deposit signer sent to Emily, maybe slightly modified

emily/handler/src/database/accessors.rs
MCJOHN974: Hmm, yeas, such condition looks better. About `Reprocessing` I think we don't really use it but not sure.

MCJOHN974: [840d818](https://github.com/stacks-sbtc/sbtc/pull/1653/commits/840d8181e1a41903b044c4919c08ce3e60503272)

emily/handler/src/database/accessors.rs
```diff
@@ -201,7 +202,10 @@ pub async fn pull_and_update_deposit_with_retry(
         if update.is_unnecessary(&deposit_entry) {
             return Ok(deposit_entry);
         }
-        if !is_trusted_key && deposit_entry.status != Status::Pending {
+        let is_valid_untrusted_status_update = update.event.status == StatusEntry::Accepted
+            && (deposit_entry.status == Status::Pending
+                || deposit_entry.status == Status::Accepted);
+        if !is_trusted_key && !is_valid_untrusted_status_update {
```
matteojug, line 208: IIUC this enables overwriting an accepted entry with different accepted info (eg, the stacks block). Why do we want this? vs keeping that updating an accepted deposit with accepted is forbidden, and just not bail in the caller

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -552,7 +566,20 @@ async fn update_deposits(
...
+                status: StatusCode::OK.as_u16(),
+            },
+        ));
+    }
+
+    if updated_deposits
+        .iter()
+        .all(|(_, deposit)| deposit.status == StatusCode::NOT_FOUND.as_u16())
+    {
+        return Err(Error::NotFound);
```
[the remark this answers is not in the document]
MCJOHN974: Hmm, probably what you suggesting can reduce a bit complexity both at Emily and clients, so I'd agree to remove it. @djordon wdyt

emily/handler/src/database/accessors.rs
MCJOHN974: AFAIU requests are `Accepted` before the transaction is confirmed, and if this tx got RBFd we want to make update `Accepted` -> `Accepted`.

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -541,7 +541,21 @@ async fn update_deposits(
...
+                updated_deposits.push((
+                    index,
+                    DepositWithStatus {
+                        deposit: Deposit::default(),
+                        status: StatusCode::NOT_FOUND.as_u16(),
+                    },
+                ));
+                continue;
+            }
+            Err(e) => return Err(e),
```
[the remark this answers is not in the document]
MCJOHN974: Returning failed entry instead of bailing will increase reliability, yeah, but it probably will hide some problem from us, since it is much more likely that we will pay attention to failed http request rather then just failed entry.

So I'd say, for now, it might be better to explicitly bail if some unknown error happens, and silent only errors we know about and we know that they are okish.

emily/handler/src/database/accessors.rs
matteojug: oh you're right, in case of sweep rbf we want this to be updated.

emily/handler/src/api/handlers/deposit.rs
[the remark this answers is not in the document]
matteojug: mmm but then we risk losing the update from the coordinator just because of a single error that may be transient. We can add some logs in here and add alerts for it, killing the whole batch seems like the "worst" case as the signers will not retry the update.

emily/handler/src/api/handlers/deposit.rs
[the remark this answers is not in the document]
djordon: Hmmm. My default would be to return the same status code in cases where they all match. But that doesn't make much sense if we're using 207.

But honestly I'm not sure. The original ticket description didn't call for the multi-status responses, hence the conundrum. Also, using 207 seems "incorrect", since we are returning JSON, and not XML (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/207).

emily/handler/src/api/models/deposit/responses.rs
djordon: Well, my comment here is just trying to say that we should add more description to our comments. Saying that the `deposit` field with type `Deposit` is the "Deposit itself" doesn't give the reader any more context that can already be gleamed from the names.

Using the parse language here seems to add confusion, so we can scratch it. I'll still answer your questions though.

---

> Why it is "parsed"? I'm not sure such comment bring a lot of value to reader and... Where we parse it?

The stuff that gets stored in the database gets populated after transforming the `CreateDepositRequest` body into a `DepositInfo` type (I'll be referring to the `sbtc::deposits::DepositInfo` throughout). To do that Emily takes the deposit and reclaim script bytes and "extracts" the recipient, max fee, and lock time. So whenever I see `Deposit` I think `DepositInfo` -- even there are more fields in the `Deposit` than what is in `DepositInfo` -- and the `Deposit` needs you to "read" the scripts and interpret the bytes to set some of its fields. I call this interpreting "parsing".

> This struct is not related to some parsing functions or so, so I don't think it is parsed....

Yeah Emily's `Deposit` type is related to a `DepositEntry`, and that type is what gets created using the `DepositInfo`. And to create a `DepositInfo`, we use `ReclaimScriptInputs::parse` and `DepositScriptInputs::parse` functions to extract some fields.


---

Maybe something like "The fully extracted and validated deposit request" would be more clear?

emily/handler/src/database/accessors.rs
djordon: Okay, I just a quick read through and it doesn't look like we use it. Would you mind checking too? We can remove it in a separate PR.

emily/handler/src/database/accessors.rs
```diff
@@ -201,7 +202,10 @@ pub async fn pull_and_update_deposit_with_retry(
         if update.is_unnecessary(&deposit_entry) {
             return Ok(deposit_entry);
         }
-        if !is_trusted_key && deposit_entry.status != Status::Pending {
+        let is_valid_untrusted_status_update = update.event.status == StatusEntry::Accepted
+            && (deposit_entry.status == Status::Pending
+                || deposit_entry.status == Status::Accepted);
```
djordon, line 207: We probably want to exit early with `Ok(_)` when `update.event.status == StatusEntry::Accepted && deposit_entry.status == Status::Accepted`. Basically, we do not want to actually add a new entry to the History attribute in this case, but we don't want to error either. And we'll want to modify the docs with detailing this behavior and why.

emily/handler/src/api/handlers/deposit.rs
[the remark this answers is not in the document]
djordon: I'm leaning toward just returning a 200-202 if there was at least one success. We have no plans on having the signers act if they receive an error, so I'm okay with doing something simple and sensible, like returning the max of all status codes or something like that.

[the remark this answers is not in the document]
MCJOHN974: Well, I don't really see any harm we may get from returning more information than we need (performance wise I don't expect we will hit any limits in observable future and code wise process multi-response instead of 1 response is something like 2 extra loc), so I'd keep multiresponce and add more info

emily/handler/src/api/models/deposit/responses.rs
MCJOHN974: Well, now I agree that this deposit is really parsed, thanks for explanation!

Aaand, maybe yeah, it is the information you might be interested in while looking at this struct, okay, let's do it that way

emily/handler/src/database/accessors.rs
MCJOHN974: [a9b2df2](https://github.com/stacks-sbtc/sbtc/pull/1653/commits/a9b2df23db510a3e39a919ad2b21396a5f704602)

emily/handler/src/api/handlers/deposit.rs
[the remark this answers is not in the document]
cylewitruk: I would keep the multistatus _response body_ for all responses and return `200` if we want to avoid repurposing `207`. Then we have a single implementation for handling the response, instead of response-code-specific branches.

emily/handler/src/database/accessors.rs
MCJOHN974: I don't see any usages of this. Opened issue for it #1678

MCJOHN974:
REMINDER FOR ME: do same for withdrawals before merging once all review passes

emily/handler/src/api/handlers/deposit.rs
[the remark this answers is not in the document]
MCJOHN974: I decided to return 207 always fyi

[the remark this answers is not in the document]
matteojug: It doesn't seem there was a consensus for the original question, but given that now we can have different errors for failing deposits I would suggest to remove this custom response (that would happen only if all the errors are 404) and just always return the whole body for consistency.

MCJOHN974:
I think now I did it in such a way that we never bail early and always process all requests in a batch even if all of them are error ones. Also it always return 207 MULTISTATUS even if all requests have same error status. It would be easier if everything is unified, and if somebody need to check if all updates was under an error they can do this logic on their side. Yes, sometimes it _can_ be slightly performance degradation but I don't think we care about it that much

emily/handler/src/api/handlers/deposit.rs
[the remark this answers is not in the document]
MCJOHN974: Btw I think we return not multistatus only if we failed to extract `ApiState` from the db in the very beginning or reorg currently happening, so maybe we can close this thread

emily/handler/src/database/accessors.rs
MCJOHN974: I think then we can keep this early exist for both trusted and untrusted for now. If sidecar don't use it we can change it once it will use

MCJOHN974:
I think all open threads can be closed now

MCJOHN974:
NOTE: after merging #1688 make sure that batches with RBF and Failed statuses work well. (Add test for it or double-check if we already have one).

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -194,32 +194,50 @@ impl UpdateDepositsRequestBody {
...
+                                bitcoin_tx_output_index,
+                            )),
+                        ));
+                    }
+                    other => {
+                        tracing::error!(
+                            bitcoin_txid = update.bitcoin_txid,
+                            bitcoin_tx_output_index = update.bitcoin_tx_output_index,
+                            error = format!("{other:?}"),
+                            "encountered error unexpected error while validating deposit update. This error should never happen during a deposit update validation.",
```
matteojug, line 228: nit?
```suggestion
                            "unexpected error while validating deposit update: this error should never happen during a deposit update validation.",
```

emily/handler/tests/integration/deposit.rs
```diff
@@ -1146,7 +1151,245 @@ async fn update_deposits_is_not_forbidden_for_sidecar(previous_status: Status, n
...
+        ],
+    };
+    let update_responce = apis::deposit_api::update_deposits_signer(
+        &testing_configuration,
+        update_deposits_request_body,
+    )
+    .await
+    .expect("Received an error after making a valid update deposit request api call.");
+
+    // Check that multi
```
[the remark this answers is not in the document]
MCJOHN974: [aa50493](https://github.com/stacks-sbtc/sbtc/pull/1653/commits/aa504937d22ba2f6226456edc1ffe38150da9e40)

emily/handler/tests/integration/deposit.rs
```diff
@@ -1146,7 +1151,245 @@ async fn update_deposits_is_not_forbidden_for_sidecar(previous_status: Status, n
...
+                bitcoin_txid: create_deposit_body1.bitcoin_txid.clone(),
+                fulfillment: None,
+                status: Status::Accepted,
+                status_message: "Second update".into(),
+            },
+        ],
+    };
+    apis::deposit_api::update_deposits_signer(&testing_configuration, update_deposits_request_body)
+        .await
+        .expect("Received an error after making a valid update deposit request api call.");
```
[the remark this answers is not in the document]
MCJOHN974: [28ce4c9](https://github.com/stacks-sbtc/sbtc/pull/1653/commits/28ce4c90fd0c8fac36c6641276bcbe2b6cc0ed1c)

emily/handler/tests/integration/deposit.rs
```diff
@@ -1146,7 +1151,245 @@ async fn update_deposits_is_not_forbidden_for_sidecar(previous_status: Status, n
...
+        apis::deposit_api::get_deposits(&testing_configuration, Status::Pending, None, None)
+            .await
+            .expect("Received an error after making a valid get deposits api call.");
+    assert_eq!(deposits.deposits.len(), 2);
+
+    // Update first deposit to Accepted.
+    let update_deposits_request_body = UpdateDepositsRequestBody {
+        deposits: vec![DepositUpdate {
+            bitcoin_tx_output_index,
+            bitcoin_txid: bitcoin_txid.clone(),
```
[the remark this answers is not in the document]
MCJOHN974: [e5b7173](https://github.com/stacks-sbtc/sbtc/pull/1653/commits/e5b71735ede58ab5a13c86bcb631a94850b9b565)

MCJOHN974:
Thanks for your comments, @matteojug ! I think now they all are addressed

emily/handler/src/database/accessors.rs
```diff
@@ -434,7 +443,14 @@ pub async fn pull_and_update_withdrawal_with_retry(
             return Ok(entry);
         }
 
-        if !is_trusted_key && entry.status != Status::Pending {
+        // We don't want to add a new entry if the status is already accepted.
+        // Updates Accepted -> Accepted occurs usually due to RBF.
+        if update.event.status == StatusEntry::Accepted && entry.status == Status::Accepted {
+            return Ok(entry);
+        }
```
MCJOHN974, line 450: What about withdrawals switching `Accepted` -> `Accepted`? Rbf can't be here.

MCJOHN974:
> LGTM. Are you going to address the withdrawals in a different PR?

@matteojug nope, I just pushed changes to this PR.

emily/handler/src/database/accessors.rs
MCJOHN974: cc @djordon

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -194,32 +194,44 @@ impl UpdateDepositsRequestBody {
...
+                        bitcoin_tx_output_index,
+                        "failed to update deposit: request missing fulfillment for completed request."
+                    );
+                    deposits.push((index, Err(err.clone())));
+                }
+                Err(err) => {
+                    tracing::error!(
+                        bitcoin_txid = update.bitcoin_txid,
+                        bitcoin_tx_output_index = update.bitcoin_tx_output_index,
+                        ?err,
```
djordon, line 222: We should call this field `error` to be consistent with the other usage in this PR and elsewhere[^1]. Also, we should use the display implementation instead of the debug implementation, so we should do `%error` instead of `?err`.


[^1]: Well, technically there are some exceptions in `main.rs` in the signer repo, we need to fix that

emily/handler/src/api/models/withdrawal/requests.rs
```diff
@@ -122,28 +122,40 @@ impl UpdateWithdrawalsRequestBody {
...
+                    tracing::warn!(
+                        request_id,
+                        "failed to update withdrawal: request missing fulfillment for completed request."
+                    );
+                    withdrawals.push((index, Err(err.clone())));
+                }
+                Err(err) => {
+                    tracing::error!(
+                        request_id = update.request_id,
+                        ?err,
```
djordon, line 146: Same thing here, `?err` to `%error`

emily/handler/src/database/accessors.rs
djordon: These updates are for transactions where sweep in deposits and/or sweep out withdrawals, so we can have the status of a withdrawal request going from `Accepted` to `Accepted` when the signers try to RBF an outstanding sweep transaction.

emily/handler/src/database/entries/deposit.rs
```diff
@@ -647,7 +647,10 @@ pub struct ValidatedUpdateDepositsRequest {
     ///
     /// This allows the updates to be executed in chronological order but returned in the order
     /// that the client sent them.
-    pub deposits: Vec<(usize, ValidatedDepositUpdate)>,
+    pub deposits: Vec<(
+        usize,
+        Result<ValidatedDepositUpdate, crate::common::error::ValidationError>,
```
djordon, line 652: Usually we just import the type instead of using the full path. Any particular reason for doing it this way?

emily/handler/src/database/entries/withdrawal.rs
```diff
@@ -636,7 +636,10 @@ pub struct ValidatedUpdateWithdrawalRequest {
     ///
     /// This allows the updates to be executed in chronological order but returned in the order
     /// that the client sent them.
-    pub withdrawals: Vec<(usize, ValidatedWithdrawalUpdate)>,
+    pub withdrawals: Vec<(
+        usize,
+        Result<ValidatedWithdrawalUpdate, crate::common::error::ValidationError>,
```
djordon, line 641: Oh same question here

emily/handler/src/database/accessors.rs
MCJOHN974: Aaaand, we want to change an event in that case, yes?

emily/handler/src/database/entries/deposit.rs
MCJOHN974: [f36db4f](https://github.com/stacks-sbtc/sbtc/pull/1653/commits/f36db4fdbee7619b946abfa6a11be7ec4ee05c51)

emily/handler/src/api/models/withdrawal/requests.rs
MCJOHN974: [f36db4f](https://github.com/stacks-sbtc/sbtc/pull/1653/commits/f36db4fdbee7619b946abfa6a11be7ec4ee05c51)

emily/handler/src/api/models/deposit/requests.rs
MCJOHN974: [f36db4f](https://github.com/stacks-sbtc/sbtc/pull/1653/commits/f36db4fdbee7619b946abfa6a11be7ec4ee05c51)

emily/handler/src/database/accessors.rs
djordon: I think we want the same behavior that we have for deposits, where Emily ignores events when we're going from `Accepted` to `Accepted`.

matteojug (approved):

emily/handler/src/api/models/deposit/responses.rs
```diff
@@ -30,5 +30,15 @@ pub struct GetDepositsResponse {
...
+}
+
+/// Wrapper for deposit with status code. Used for multi-status responses.
+#[derive(Clone, Default, Debug, PartialEq, Hash, Serialize, Deserialize, ToSchema, ToResponse)]
+#[serde(rename_all = "camelCase")]
+pub struct DepositWithStatus {
+    /// The fully extracted and validated deposit request.
+    pub deposit: Deposit,
+    /// HTTP status code, returned as part of multi-status responses.
+    pub status: u16,
```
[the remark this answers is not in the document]
MCJOHN974: Yeah, also had such thoughts, let's create issue about it and try to keep prs smaller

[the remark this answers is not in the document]
MCJOHN974: I opened #1697 and will close this thread

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -513,11 +503,23 @@ async fn update_deposits(
...
+            updated_deposits.push((
+                index,
+                DepositWithStatus {
+                    deposit: Deposit::default(),
+                    status: StatusCode::BAD_REQUEST.as_u16(),
+                },
+            ));
+            continue;
+        }
+        let update = update.unwrap();
```
djordon, line 521:
```suggestion
        let Ok(update) = update else {
            updated_deposits.push((
                index,
                DepositWithStatus {
                    deposit: Deposit::default(),
                    status: StatusCode::BAD_REQUEST.as_u16(),
                },
            ));
            continue;
        }
```
Maybe for the future, this already has two approvals

emily/handler/src/api/handlers/withdrawal.rs
```diff
@@ -408,28 +399,78 @@ async fn update_withdrawals(
...
+            updated_withdrawals.push((
+                index,
+                WithdrawalWithStatus {
+                    withdrawal: Withdrawal::default(),
+                    status: StatusCode::BAD_REQUEST.as_u16(),
+                },
+            ));
+            continue;
+        }
+        let update = update.unwrap();
```
djordon, line 417: Same stylistic comment here with using the `let .. else` syntax. No need to change though

emily/handler/tests/integration/deposit.rs
```diff
@@ -1146,7 +1151,267 @@ async fn update_deposits_is_not_forbidden_for_sidecar(previous_status: Status, n
...
+    } = DepositTxnData::new(DEPOSIT_LOCK_TIME, DEPOSIT_MAX_FEE, &[DEPOSIT_AMOUNT_SATS]);
+    let reclaim_script = reclaim_scripts.first().unwrap().clone();
+    let deposit_script = deposit_scripts.first().unwrap().clone();
+
+    let create_deposit_body1 = CreateDepositRequestBody {
+        bitcoin_tx_output_index,
+        bitcoin_txid: bitcoin_txid.clone(),
+        deposit_script: deposit_script.clone(),
+        reclaim_script: reclaim_script.clone(),
+        transaction_hex: transaction_hex.clone(),
```
djordon, line 1187: No need for the clones here

djordon (approved):
Looks good. Left some nits, we already have approvals so feel free to do them or not.

PR #1655, opened 2025-05-07
chore: use get_rng in should_process_withdrawals

matteojug:
## Description

Address https://github.com/stacks-sbtc/sbtc/issues/1590

## Changes

We still have cases where the generated data does not have "valid" withdrawals in `should_process_withdrawals` (eg for seed `7437941319183672874`). In ~100 runs of the test, I only got such seed once, so I'm proposing to just drop the non-empty check. Not the best, since we may change stuff and end up with always having zero withdrawals in that test, but that will be probably catch in other places.

## Testing Information

Tested with a previously broken seed.

## Checklist:

### DIFF
--- a/signer/src/testing/transaction_coordinator.rs
+++ b/signer/src/testing/transaction_coordinator.rs
@@ -211,6 +211,36 @@ where
             prevouts: Vec::new(),
         };
         test_data.push_bitcoin_txs(&bitcoin_chain_tip, vec![tx_info], &signer_script_pubkeys);
+
+        // Also ensure one valid withdrawal exists for test consistency
+        let stacks_block = test_data
+            .stacks_blocks
+            .iter()
+            .find(|b| b.bitcoin_anchor == bitcoin_chain_tip.block_hash)
+            .unwrap();
+        let mut withdrawal = test_data
+            .withdraw_requests
+            .iter()
+            .find(|w| w.block_hash == stacks_block.block_hash)
+            .unwrap()
+            .clone();
+
+        let mut withdrawal_votes = test_data
+            .withdraw_signers
+            .iter()
+            .filter(|ws| ws.qualified_id() == withdrawal.qualified_id())
+            .cloned()
+            .collect::<Vec<_>>();
+
+        withdrawal.request_id *= 100;
+        withdrawal_votes.iter_mut().for_each(|ws| {
+            ws.is_accepted = true;
+            ws.request_id = withdrawal.request_id;
+        });
+
+        test_data.withdraw_requests.push(withdrawal);
+        test_data.withdraw_signers.append(&mut withdrawal_votes);
+
         test_data.remove(original_test_data);
         self.write_test_data(&test_data).await;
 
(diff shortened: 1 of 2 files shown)


signer/src/testing/transaction_coordinator.rs
```diff
@@ -287,8 +285,6 @@ where
             .saturating_sub(crate::WITHDRAWAL_BLOCKS_EXPIRY)
             .saturating_add(crate::WITHDRAWAL_EXPIRY_BUFFER);
 
-        // Assert that there are some withdrawals in storage while get_pending_requests return 0 withdrawals
-        assert!(!withdrawals_in_storage.is_empty());
```
djordon, line 291: Why do we want to remove this check?

matteojug: I'm not sure we want, but right now it's still possible for the testdata to generate no "valid" withdrawals (albeit rarely, 1 in ~100 tries). So either we remove this check, or we manually add a withdrawal -- I went for the former, nothing against the latter, I'm not sure if it's significantly better since it would be an "obvious" one.

djordon: Yeah, makes sense. Any idea for why it's possible to not have any valid withdrawals in the database? I thought that each stacks block had withdrawals in confirmed within them, even the ones not on the canonical chain. Maybe the change is to force consecutive blocks?

MCJOHN974: Hmm, I don't mind to merge adding withdrawals to storage manually or even break our "no flaky tests" rule and merge this. But, then we should open an issue to dig into our testing infra and understand why we don't have withdrawals in storage in this case (and I suspect it is because in our testing helpers we do something wrong)

MCJOHN974: Yeaah, I remember I already posted somewhere that our forking is too aggressive: https://en.wikipedia.org/wiki/Random_recursive_tree

MCJOHN974 (approved):
Lgtm but pls open an issue for revisiting our `TestData` structure before merging. At least we can change our forking aggressiveness, but I suspect there are some more serious problems

signer/src/testing/transaction_coordinator.rs
```diff
@@ -213,6 +211,42 @@ where
...
+            .clone();
+
+        let mut withdrawal_votes = test_data
+            .withdraw_signers
+            .iter()
+            .filter(|ws| ws.qualified_id() == withdrawal.qualified_id())
+            .cloned()
+            .collect::<Vec<_>>();
+
+        withdrawal.request_id *= 100;
```
djordon, line 241: Is zero a valid withdrawal `request_id` here? Does it matter? I think the answer is "no" for this test since we get the withdrawal request from a block anchored to the bitcoin chain tip so it is unlikely to be first withdrawal.

Also, can't wait for us to not have to do this after doing https://github.com/stacks-sbtc/sbtc/issues/1671.

djordon (approved):
Looks good, just left a non-question.

signer/src/testing/transaction_coordinator.rs
matteojug: It shouldn't be possible to get zero here since `TestData` uses incremental IDs starting from 1. But yeah, really looking forward to a more sane approach for `TestData` 😅

PR #1666, opened 2025-05-08
chore: switch more code to get_rng()

MCJOHN974:
## Description

This PR aims to eliminate all (or all I'll find xD) usages of raw `OsRng` and replace them with `get_rng()`. This is useful because it 1) Make flaky tests easy to reproduce. 2) Make our testing code more consistent across codebase.

Closes: #?

## Changes

Note: Scope of the changes is _only_ signer crate.

- Switched from raw `OsRng` to `get_rng()` usages.
- In `bitcoin::utxo.rs` added some workaround (`global_rng()`) to handle random usages in `test_case` headers.
- In `bitcoin::utxo.rs` some helper functions now expect `rand::Rng` input parameter instead of creating rng themselves

WIP

## Testing Information

No new tests needed.

## Checklist:

### DIFF
--- a/signer/src/bitcoin/utxo.rs
+++ b/signer/src/bitcoin/utxo.rs
@@ -1643,9 +1643,9 @@ mod tests {
     use fake::Fake as _;
     use model::SignerVote;
     use more_asserts::assert_ge;
+    use rand::SeedableRng;
     use rand::distributions::Distribution;
     use rand::distributions::Uniform;
-    use rand::rngs::OsRng;
     use sbtc::deposits::DepositScriptInputs;
     use secp256k1::Keypair;
     use secp256k1::SecretKey;
@@ -1657,6 +1657,8 @@ mod tests {
     use crate::context::RollingWithdrawalLimits;
     use crate::testing;
     use crate::testing::btc::base_signer_transaction;
+    use crate::testing::get_rng;
+    use rand::RngCore;
 
     /// The maximum virtual size of a transaction package in v-bytes.
     const MEMPOOL_MAX_PACKAGE_SIZE: u32 = 101000;
@@ -1666,25 +1668,40 @@ mod tests {
 
     static NEXT_REQUEST_ID: AtomicU64 = AtomicU64::new(0);
 
-    fn generate_x_only_public_key() -> XOnlyPublicKey {
-        let secret_key = SecretKey::new(&mut OsRng);
+    fn generate_x_only_public_key<R: rand::Rng>(rng: &mut R) -> XOnlyPublicKey {
+        let secret_key = SecretKey::new(rng);
         secret_key.x_only_public_key(SECP256K1).0
+    }
+
+    static GLOBAL_RNG: LazyLock<rand::rngs::StdRng> = LazyLock::new(|| {
+        let seed = get_rng().next_u64();
+        // Nextest prints stdout and stderr only for failed tests so this
+        // wouldn't be too much noise.
+        eprintln!("bitcoin::utxo.rs GLOBAL_RNG seed: {seed:}");
+        rand::rngs::StdRng::seed_from_u64(seed)
+    });
+
+    fn global_rng() -> rand::rngs::StdRng {
+        GLOBAL_RNG.clone()
     }
 
     // The is the least non dust amount for withdrawal outputs locked by
     // the generate_address() script, which generates P2WPKH outputs
-    static MINIMAL_NON_DUST_AMOUNT_P2WPKH: LazyLock<u64> =
-        LazyLock::new(|| generate_address().minimal_non_dust().to_sat());
+    static MINIMAL_NON_DUST_AMOUNT_P2WPKH: LazyLock<u64> = LazyLock::new(|| {
+        generate_address(&mut global_rng())
+            .minimal_non_dust()
+            .to_sat()
+    });
 
-    fn generate_address() -> ScriptPubKey {
-        let secret_key = SecretKey::new(&mut OsRng);
+    fn generate_address<R: rand::Rng>(rng: &mut R) -> ScriptPubKey {
+        let secret_key = SecretKey::new(rng);
         let pk = CompressedPublicKey(secret_key.public_key(SECP256K1));
 
         ScriptBuf::new_p2wpkh(&pk.wpubkey_hash()).into()
     }
 
-    fn generate_outpoint(amount: u64, vout: u32) -> OutPoint {
-        let sats: u64 = Uniform::new(1, 500_000_000).sample(&mut OsRng);
+    fn generate_outpoint<R: rand::Rng>(amount: u64, vout: u32, rng: &mut R) -> OutPoint {
+        let sats: u64 = Uniform::new(1, 500_000_000).sample(rng);
 
         let tx = Transaction {
             version: Version::TWO,
@@ -1723,8 +1740,13 @@ mod tests {
     }
 
     /// Create a new deposit request depositing from a random public key.
-    fn create_deposit(amount: u64, max_fee: u64, signer_bitmap: u128) -> DepositRequest {
-        let signers_public_key = generate_x_only_public_key();
+    fn create_deposit<R: rand::Rng>(
+        amount: u64,
+        max_fee: u64,
+        signer_bitmap: u128,
+        rng: &mut R,
+    ) -> DepositRequest {
+        let signers_public_key = generate_x_only_public_key(rng);
 
         let contract_name = std::iter::repeat_n('a', 128).collect::<String>();
         let principal_str = format!("{}.{contract_name}", StacksAddress::burn_address(false));
@@ -1736,7 +1758,7 @@ mod tests {
         };
 
         DepositRequest {
-            outpoint: generate_outpoint(amount, 1),
+            outpoint: generate_outpoint(amount, 1, rng),
             max_fee,
             signer_bitmap: BitArray::new(signer_bitmap.to_le_bytes()),
             amount,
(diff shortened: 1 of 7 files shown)


MCJOHN974:
Note for myself: search for `rand::thread_rng` as well.

signer/src/bitcoin/utxo.rs
```diff
@@ -2743,28 +2787,28 @@ mod tests {
         // Create deposit and withdrawal requests, some with too low of a
         // max fees and some with a good max fee.
         let deposit_low_fee = ((SOLO_DEPOSIT_TX_VSIZE - 1.0) * fee_rate) as u64;
-        let low_fee_deposits = std::iter::repeat_with(|| uniform.sample(&mut OsRng))
+        let low_fee_deposits = std::iter::repeat_with(|| uniform.sample(&mut rng.clone()))
```
cylewitruk, line 2790: Note that `.clone()` here (and everywhere else) may not have the behavior you might expect; it doesn't behave like a `Rc<Rng>` (i.e. shared ownership), every `clone()` is giving you an exact clone of `rng`:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d0538e45a3c396db1ba714e5af6c9105

MCJOHN974: I don't know exactly but I believe probability of any problems because of such clones are pretty small

signer/src/testing/request_decider.rs
```diff
@@ -146,14 +144,18 @@ where
...
     /// Assert that the transaction signer will make and store decisions
     /// for pending deposit requests.
-    pub async fn assert_should_store_decisions_for_pending_deposit_requests(self) {
-        let mut rng = get_rng();
+    pub async fn assert_should_store_decisions_for_pending_deposit_requests<
+        R: rand::Rng + rand::CryptoRng + Send + Clone + 'static,
+    >(
+        self,
+        rng: &mut R,
+    ) {
```
djordon, line 152: For large function signatures like this, can we format it like this here?
```suggestion
    pub async fn assert_should_store_decisions_for_pending_deposit_requests<R>(
        self,
        rng: &mut R,
    )
    where
        R: rand::Rng + rand::CryptoRng + Send + Clone + 'static 
    {
```

signer/src/bitcoin/utxo.rs
MCJOHN974: Whoops, actually such clones _are_ a problem, will remove them

djordon: Had a conversation about this online and this can be an issue because in some of our tests the randomness is used during key generation and such.

signer/src/network/in_memory2.rs
```diff
@@ -237,15 +239,22 @@ mod tests {
             .expect("client 2 did not receive all messages in time")
         });
 
+        let mut rng_for_corutine = rng.clone();
```
djordon, line 242: Cloning a random number generator is a tricky thing, and I'm not sure that this is equivalent to what we had before. Perhaps we should create a new RNG per spawned task so that there is no risk of duplication of randomness for each task?

Also, spelling nit: `corutine -> coroutine`

signer/src/testing/network.rs
```diff
@@ -25,20 +23,23 @@ impl RandomMemoryMultiaddr for Multiaddr {
...
+pub async fn assert_clients_can_exchange_messages<
+    C: network::MessageTransfer + Send + 'static,
+    R: rand::Rng + rand::CryptoRng,
+>(
     client_1: C,
     client_2: C,
     private_key_1: PrivateKey,
     private_key_2: PrivateKey,
+    rng: &mut R,
 ) {
```
djordon, line 35: I find this easier to read:
```suggestion
pub async fn assert_clients_can_exchange_messages<C, R>(
    client_1: C,
    client_2: C,
    private_key_1: PrivateKey,
    private_key_2: PrivateKey,
    rng: &mut R,
) 
where
    C: network::MessageTransfer + Send + 'static,
    R: rand::Rng + rand::CryptoRng,
{
```

signer/src/testing/request_decider.rs
```diff
@@ -231,15 +233,18 @@ where
 
     /// Assert that the transaction signer will make and store decisions
     /// for pending withdraw requests.
-    pub async fn assert_should_store_decisions_for_pending_withdrawal_requests(self) {
-        // TODO(#1466): fix this test for other seeds and use `get_rng()`
-        let mut rng = rand::rngs::StdRng::seed_from_u64(42);
+    pub async fn assert_should_store_decisions_for_pending_withdrawal_requests<
+        R: rand::Rng + rand::CryptoRng + Send + Clone + 'static,
```
djordon, line 237: Similarly, I'd move this trait bounds to a where clause for readability. Same for the ones in:
* `assert_should_store_decisions_received_from_other_signers`
* `assert_should_be_able_to_coordinate_signing_rounds`
* `assert_get_signer_utxo_simple`
* `assert_get_signer_utxo_fork`
* `assert_get_signer_utxo_unspent`
* `assert_get_signer_utxo_donations`
* `assert_should_be_able_to_participate_in_dkg`
* `run_until_dkg_end`
* `run_until_signature_share_response`
* `participate_in_signing_rounds_forever`
* `participate_in_signing_round`
* `store_dkg_shares`

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -951,7 +950,7 @@ async fn run_dkg_from_scratch() {
     });
 
     let tx_signer_processes = signers.iter().map(|(context, _, _, net)| {
-        TxSignerEventLoop::new(context.clone(), net.spawn(), OsRng)
+        TxSignerEventLoop::new(context.clone(), net.spawn(), rng.clone())
```
djordon, line 953: Oh this clone looks dangerous but it isn't, because the `TxSignerEventLoop` no longer uses the supplied random number generator. I'll open a PR removing the field entirely.

signer/tests/integration/transaction_signer.rs
```diff
@@ -742,7 +741,10 @@ mod validate_dkg_verification_message {
 
     impl Default for TestParams {
         fn default() -> Self {
-            let new_aggregate_key = Keypair::new_global(&mut OsRng).x_only_public_key().into();
+            // Here we use explicitly seeded RNG to ensure that [`default()`] is deterministic.
+            // (and default don't allow to pass any parameters)
+            let mut rng = rand::rngs::StdRng::seed_from_u64(51);
+            let new_aggregate_key = Keypair::new_global(&mut rng).x_only_public_key().into();
```
djordon, line 747: It makes sense to have default deterministic, however the tests benefit from the randomness that was here. Better to move this implementation away from `Default::default` and into a new `TestParams::new_with_rng` function (or similar) so that it can take a random number generate as an input parameter.

signer/src/bitcoin/utxo.rs
```diff
@@ -1666,25 +1668,40 @@ mod tests {
...
         secret_key.x_only_public_key(SECP256K1).0
     }
 
+    static GLOBAL_RNG: LazyLock<rand::rngs::StdRng> = LazyLock::new(|| {
+        let seed = rand::rngs::OsRng.next_u64();
+        // Nextest prints stdout and stderr only for failed tests so this
+        // wouldn't be too much noise.
+        eprintln!("bitcoin::utxo.rs GLOBAL_RNG seed: {seed:}");
+        rand::rngs::StdRng::seed_from_u64(seed)
+    });
```
djordon, line 1682: Why not use `get_rng` here, like with:
```suggestion
    static GLOBAL_RNG: LazyLock<rand::rngs::StdRng> = LazyLock::new(get_rng);
```
?

signer/src/bitcoin/utxo.rs
```diff
@@ -3041,17 +3087,20 @@ mod tests {
         create_deposit(
             DEPOSIT_DUST_LIMIT + SOLO_DEPOSIT_TX_VSIZE as u64,
             10_000,
-            0
+            0,
+            &mut global_rng(),
```
djordon, line 3091: I think we can use `get_rng` here directly, no need for a global one.

signer/src/bitcoin/utxo.rs
djordon: Just rereading this now. Some of my comments are repeating the point made here.

signer/src/bitcoin/utxo.rs
djordon: Actually, I don't think the `GLOBAL_RNG` is necessary at all.

signer/src/bitcoin/utxo.rs
```diff
@@ -1666,25 +1668,40 @@ mod tests {
...
+
     // The is the least non dust amount for withdrawal outputs locked by
     // the generate_address() script, which generates P2WPKH outputs
-    static MINIMAL_NON_DUST_AMOUNT_P2WPKH: LazyLock<u64> =
-        LazyLock::new(|| generate_address().minimal_non_dust().to_sat());
-
-    fn generate_address() -> ScriptPubKey {
-        let secret_key = SecretKey::new(&mut OsRng);
+    static MINIMAL_NON_DUST_AMOUNT_P2WPKH: LazyLock<u64> = LazyLock::new(|| {
+        generate_address(&mut global_rng())
```
djordon, line 1691: We should be able to use `get_rng` here without any issues. Any random number generate should do the trick here, since the output is not dependent on the randomness. We could even make it deterministic.

MCJOHN974:
It is not urgent, and also it is in such a condition that it would be easier to do it from scratch, so closing for now

Issue #1657, opened 2025-05-07
[Feature]: Do not reorg in Emily past ten bitcoin blocks


djordon:
# Feature - Do not reorg in Emily past ten bitcoin blocks

## 1. Description

This ticket is to make sure that Emily doesn't delete data when it detects a reorg is occurring that affects more than some reasonable number of bitcoin blocks. If Emily detects that such an event is occurring, it's more likely that Emily is hooked up to a new stacks node or something like that. In this case, we do not want Emily do modify those old entries.

## 2. Technical Details:

Now that we have bitcoin block information in the chainstate table, we can add an additional check that the reorg depth does not exceed a certain threshold. This depth check would use the bitcoin block chain tip height to get the reorg depth to prevent false flag reorgs.

### 2.1 Acceptance Criteria:
   - [ ] Emily does not reorg bitcoin blocks that are more than too deep.

MCJOHN974:
@djordon why you think we want 10 blocks here? Usually 6 blocks considered safe for finality

PR #1667, opened 2025-05-08
[feat]: Emily don't reorg for very old chaintips

MCJOHN974:
## Description

Emily don't reorg for old chaintips

Closes: #1657

## Changes

- Now if Emily receives chainstate update request with tip older then current Emily chaintip - 6 Emily will not do a reorg but still return 200 to not make stacks node sending same chaintip again.

## Testing Information

- Added integration test checking that Emily don't do reorg if chaintip in new chainstate is old enough and updated old tests to make sure reorg happens when we expect it

## Checklist:

### DIFF
--- a/emily/handler/src/api/handlers/chainstate.rs
+++ b/emily/handler/src/api/handlers/chainstate.rs
@@ -175,16 +175,25 @@ pub async fn add_chainstate_entry_or_reorg(
     // likely means that we reconnected to new stacks node which is not fully synced,
     // rather then actual reorg happen.
     let new_bitcoin_tip_height = chainstate.bitcoin_block_height;
-    let current_bitcoin_tip_height = accessors::get_api_state(&context).await?.chaintip().bitcoin_height;
+    let current_bitcoin_tip_height = accessors::get_api_state(context)
+        .await?
+        .chaintip()
+        .bitcoin_height;
+
+    // 6 block confirmations are considered as industry standard for considering that this block
+    // will not be reorged. See https://en.bitcoin.it/wiki/Confirmation
+    let no_reorg_depth = 6;
 
     if let Some(current_bitcoin_tip_height) = current_bitcoin_tip_height {
         if let Some(new_bitcoin_tip_height) = new_bitcoin_tip_height {
-            if new_bitcoin_tip_height < current_bitcoin_tip_height - 2 {
-                return Err(Error::TooOldChaintipToReorg(new_bitcoin_tip_height, current_bitcoin_tip_height));
+            if new_bitcoin_tip_height < current_bitcoin_tip_height - no_reorg_depth {
+                return Err(Error::TooOldChaintipToReorg(
+                    new_bitcoin_tip_height,
+                    current_bitcoin_tip_height,
+                ));
             }
         }
     }
-
 
     // Get chainstate as entry.
     let entry: ChainstateEntry = chainstate.clone().into();
--- a/emily/handler/src/api/handlers/chainstate.rs
+++ b/emily/handler/src/api/handlers/chainstate.rs
@@ -186,7 +186,7 @@ pub async fn add_chainstate_entry_or_reorg(
 
     if let Some(current_bitcoin_tip_height) = current_bitcoin_tip_height {
         if let Some(new_bitcoin_tip_height) = new_bitcoin_tip_height {
-            if new_bitcoin_tip_height < current_bitcoin_tip_height - no_reorg_depth {
+            if new_bitcoin_tip_height < current_bitcoin_tip_height.saturating_sub(no_reorg_depth) {
                 return Err(Error::TooOldChaintipToReorg(
                     new_bitcoin_tip_height,
                     current_bitcoin_tip_height,
--- a/emily/handler/src/api/handlers/chainstate.rs
+++ b/emily/handler/src/api/handlers/chainstate.rs
@@ -93,6 +93,7 @@ pub async fn get_chainstate_at_height(
         (status = 400, description = "Invalid request body", body = ErrorResponse),
         (status = 404, description = "Address not found", body = ErrorResponse),
         (status = 405, description = "Method not allowed", body = ErrorResponse),
+        (status = 422, description = "Unprocessable entity", body = ErrorResponse),
         (status = 500, description = "Internal server error", body = ErrorResponse)
     ),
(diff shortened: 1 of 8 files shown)


MCJOHN974:
We agreed that depth 6 should be here

djordon (commented):
I wonder if there are some not-so-unlikely edge cases on stacks where things look like an invalid reorg to Emily, but it's actually something else. Thinking out loud, I wonder what happens when there is a stall and an actual reorg that we find out about late on the stacks block chain. Actually, things are "okay" regardless, but it's useful thinking through.

Also, I'm leaning toward us merging this work after we add a history table. The idea is to have something in place so that we can easily go and look at what happened or is happening. Then when we merge this and we observe some deep reorg, we can better observe what happened.

emily/handler/src/api/handlers/chainstate.rs
```diff
@@ -169,6 +171,30 @@ pub async fn add_chainstate_entry_or_reorg(
...
+    // rather then actual reorg happen.
+    let new_bitcoin_tip_height = chainstate.bitcoin_block_height;
+    let current_bitcoin_tip_height = accessors::get_api_state(context)
+        .await?
+        .chaintip()
+        .bitcoin_height;
+
+    // 6 block confirmations are considered as industry standard for considering that this block
+    // will not be reorged. See https://en.bitcoin.it/wiki/Confirmation
+    let no_reorg_depth = 6;
```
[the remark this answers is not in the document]
djordon: Values like this are usually made into constants and placed someplace central, mainly because they are a kind of application parameter and they are special. Whether their value is arbitrarily chosen or revealed by the gods, we treat such magic numbers differently. In the signer, we usually place values like this in the same file or in `lib.rs` where we can examine all of these application parameters. There are exceptions though, sometimes we have something like `let x = variable - CONSTANT + 1` and `1` isn't made a constant. But generally, we try to make values a constant, document their meaning and how the value was chosen.

emily/handler/src/api/handlers/chainstate.rs
```diff
@@ -169,6 +171,30 @@ pub async fn add_chainstate_entry_or_reorg(
...
+    if let Some(current_bitcoin_tip_height) = current_bitcoin_tip_height {
+        if let Some(new_bitcoin_tip_height) = new_bitcoin_tip_height {
+            if new_bitcoin_tip_height < current_bitcoin_tip_height.saturating_sub(no_reorg_depth) {
+                return Err(Error::TooOldChaintipToReorg(
+                    new_bitcoin_tip_height,
+                    current_bitcoin_tip_height,
+                ));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
MCJOHN974: [42908ed](https://github.com/stacks-sbtc/sbtc/pull/1667/commits/42908ede76c7146d4d99d69500ccd607d5ccb678)

emily/handler/src/api/handlers/chainstate.rs
```diff
@@ -169,6 +174,27 @@ pub async fn add_chainstate_entry_or_reorg(
...
+    // We don't want to do reorg for too old chainstates, because it highly
+    // likely means that we reconnected to new stacks node which is not fully synced,
+    // rather then actual reorg happen.
+    let new_bitcoin_tip_height = chainstate.bitcoin_block_height;
+    let current_bitcoin_tip_height = accessors::get_api_state(context)
+        .await?
+        .chaintip()
+        .bitcoin_height;
+
+    if let Some(current_bitcoin_tip_height) = current_bitcoin_tip_height {
```
cylewitruk, line 186: You should be able to reduce nesting with `if let (Some(current_height), Some(new_height)) = (current_bitcoin_tip_height, new_bitcoin_tip_height) { ... }`

MCJOHN974: @cylewitruk

cylewitruk:
Looks largely good to me 👍 Pending approval on @djordon's comments re: first adding a history table or not.

djordon (approved):

djordon (approved):

emily/handler/src/api/handlers/chainstate.rs
```diff
@@ -133,6 +137,7 @@ pub async fn set_chainstate(context: EmilyContext, body: Chainstate) -> impl war
         (status = 400, description = "Invalid request body", body = ErrorResponse),
         (status = 404, description = "Address not found", body = ErrorResponse),
         (status = 405, description = "Method not allowed", body = ErrorResponse),
+        (status = 422, description = "Unprocessable entity", body = ErrorResponse),
```
djordon: Wait, do we actually return this?

MCJOHN974: Noo, we decided to use silent 200 here https://github.com/stacks-sbtc/sbtc/pull/1667#discussion_r2085919408. However since we want to create alarm around it maybe we want to modify sidecar such that it will consume this 422 and do not propagate it to stacks node

@Jiloc wdyt?

emily/handler/src/api/handlers/chainstate.rs
```diff
@@ -169,6 +172,29 @@ pub async fn add_chainstate_entry_or_reorg(
...
+        .bitcoin_height;
+
+    if let (Some(current_bitcoin_tip_height), Some(new_bitcoin_tip_height)) =
+        (current_bitcoin_tip_height, new_bitcoin_tip_height)
+    {
+        if new_bitcoin_tip_height < current_bitcoin_tip_height.saturating_sub(NO_REORG_DEPTH) {
+            tracing::warn!(
+                "Attempting to add chainstate with bitcoin tip height {new_bitcoin_tip_height} \
+                    which is too old to reorg from current bitcoin tip height {current_bitcoin_tip_height}"
+            );
```
MCJOHN974: @fabergat for you, this is a log line we print when we skip reorg because it is too old.

djordon: Normally we'd do something like: 
```suggestion
            tracing::warn!(
                %new_bitcoin_tip_height,
                %current_bitcoin_tip_height,
                "Will not add chainstate with bitcoin tip height that is too old from the current bitcoin tip height"
            );
```

MCJOHN974:
@djordon , yeas, thanks for note about logs, reaaprove pls

matteojug (approved):

PR #1670, opened 2025-05-09
fix: use transactions in database migrations

joncode5:
Fixed a critical bug in the migration system where migration records were being inserted outside the transaction that executed the migration scripts.

Bug Details
When applying migrations, each migration script was executed within a transaction, but the record of that migration was inserted using the main connection pool instead of the active transaction. If a migration script failed after previous scripts had executed successfully, the transaction would roll back all schema changes, but the records of "successful" migrations would remain in the __sbtc_migrations table. On subsequent runs, these migrations would be skipped despite their schema changes never having been applied, leading to database schema inconsistency.

Fix
Modified insert_migration to accept an executor parameter Updated call in apply_migrations to pass the active transaction Ensures migration scripts and their metadata records are committed or rolled back atomically within the same transaction

## Description

Closes: #?

## Checklist:

### DIFF
--- a/signer/src/storage/postgres.rs
+++ b/signer/src/storage/postgres.rs
@@ -264,7 +264,7 @@ impl PgStore {
         .execute(executor)
         .await
         .map_err(Error::SqlxQuery)?;
-    
+
         Ok(())
     }
 


MCJOHN974 (approved):
Thanks @joncode5 for your contribution! lgtm.

djordon (approved):
Looks good, and thank you ✅

MCJOHN974 (approved):

aldur:
Thank you @joncode5!

We require [signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) before merging, would you be able to enable that?

I _think_ you can enable sign commits, sign these changes, rebase the branch and force push.

PR #1672, opened 2025-05-13
fix: install uv in docker

MCJOHN974:
## Description

Installs uv in docker.

## Issue

When I'm on `feature` branch which don't have uv update yet and try to do `make integration-env-build` I see this docker error:

```shell
 => ERROR [emily-aws-setup builder  9/13] RUN make install && make build                                                                              0.2s
 => CACHED [emily-aws-setup emily-aws-setup 2/5] WORKDIR /code                                                                                        0.0s
------
 > [emily-aws-setup builder  9/13] RUN make install && make build:
0.166 uv --directory emily_cron venv && uv --directory emily_cron pip sync pyproject.toml
0.167 /bin/sh: 1: uv: not found
0.167 make: *** [Makefile:20: install-py] Error 127
```

## Changes

- Added few lines to dockerfile for installing uv and setting it's path

## Testing Information

- No new tests needed

## Checklist:

### DIFF
--- a/docker/sbtc/Dockerfile
+++ b/docker/sbtc/Dockerfile
@@ -23,7 +23,7 @@ ARG GIT_BRANCH=main
 RUN git checkout $GIT_BRANCH
 
 # Install uv from Astral's official script
-RUN curl -Ls https://astral.sh/uv/install.sh | sh
+RUN curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/uv/releases/download/0.7.3/uv-installer.sh | sh
 
 # Add installed uv to PATH
 ENV PATH="/root/.cargo/bin:/root/.local/bin:$PATH"


cylewitruk (approved):
Looks fine. My only question would be if we should use release-specific installs as outlined on their [release page](https://github.com/astral-sh/uv/releases), e.g.:
```
curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/uv/releases/download/0.7.3/uv-installer.sh | sh
```

MCJOHN974:
Yeah, we definetely should, thank you @cylewitruk

matteojug (commented):
I'm not sure we need this: the error (` > [emily-aws-setup builder  9/13] RUN make install && make build:`) seems to reference the old version of the dockerfile (running `install` instead of `install-pnpm`).
Do we need this in main? If you update your branch to a recent main, do you still get the error?

If we need this, can you install `uv` similarly to how it is done in the emily cron/sidecar dockerfiles?

MCJOHN974:
Well, it seems it actually builds succesfully if I merge upstream, so we probably really don't need it

PR #1673, opened 2025-05-13
feat: integral support for storage transactions + bitcoin transaction processing in transaction

cylewitruk:
## Description

Closes: #1194

## Changes

> [!NOTE]
Like 95% of this PR's line count is due to moving/splitting up our stores (`in_memory.rs` and `postgres.rs`). I _did not_ alter any SQL, so the fact that current integration tests also remain unchanged (short of some minor tweaks in `use` statements etc.) and are passing should give a fairly solid indication that the refactor generally does not _break_ anything. As a reviewer I would probably put most of my focus on ensuring that no malicious code has snuck into the diffs.

- Restructured the `storage` module a bit:
  - Two new folders: `memory` and `pgsql`.
  _(I had originally just named these `in_memory` and `postgres` but ran into issues with the Emily container builds... 🤷 )_.
  - Split `DbRead` and `DbWrite` implementations into `read.rs` and `write.rs` for both memory/pgsql.
  - Core "store" logic is in `store.rs` for both memory/pgsql.
- New traits:
  - `Transactable`: Implemented for stores which can create transactions (i.e. our `SharedStore` and `PgStore`).
  - `TransactionHandle`: Implemented for transaction handle types
- New types:
  - `InMemoryTransaction`: Represents a transaction for the in-memory database and implements `TransactionHandle`. It uses a simple, naive optimistic concurrency check for concurrent transactions.
  - `PgTransaction`: Represents an sqlx postgres transaction and implements `TransactionHandle`.
  - For postgres:
    - New `PgRead` and `PgWrite` types which implement the "core" queries using postgres `Executor`s, which the `DbRead` and `DbWrite` implementations now delegate to. 
    - This adds a bit more boilerplate when adding new `Db*` methods, but keeps the actual SQL logic centralized. 
    - I did try to find ways to minimize this, but so far haven't succeeded to find another solution that works across-the-board like this, so I'm ending my time-box and PR'ing the current solution.
- **Wraps bitcoin block+tx handling in a transaction** in `block_observer.rs`.
- Updated tests & test helpers due to the above changes.

## Checklist:

- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #1674, opened 2025-05-13
feat: implement core storage transaction support + in-memory implementation

cylewitruk:
## Description

Breakout from the larger PR #1673

## Changes

**Note:** This PR _only_ affects test-related code. The new transactions are _not used_ in any actual tests here (see the other PR for that), with the exception of the handful of new tests explicitly testing the transactional functionality for the in-mem store.

Highlights:
- `SharedStore` now implements `Transactable` which lets you call `begin_transaction()` to get an `InMemoryTransaction` handle, which you can `commit()` or `rollback()`.
- `DbRead` is implemented for both `SharedStore` and `InMemoryTransaction` in `read.rs`.
- `DbWrite` is implemented for both `SharedStore` and `InMemoryTransaction` in `write.rs`.
- The implementations for `InMemoryTransaction` are just delegating calls to the `SharedStore` implementations, i.e. super boilerplate.

More details:
- Restructured the `storage` module a bit:
  - `in_memory.rs` migrated to `memory` folder.
  _(I had originally just named this `in_memory` but ran into issues with the Emily container builds complaining about duplicate modules, despite that not being true... 🤷 )_.
  - Split `DbRead` and `DbWrite` implementations into `read.rs` and `write.rs` to try to manage file size and separate the logical implementations.
  - Core "store" logic is in `store.rs` (i.e. `Store` and `SharedStore`).
- New traits:
  - `Transactable`: Implemented for stores which can create transactions (i.e. our `SharedStore` and `PgStore`).
  - `TransactionHandle`: Implemented for transaction handle types
- New types:
  - `InMemoryTransaction`: Represents a transaction for the in-memory database and implements `TransactionHandle`. It uses a simple, naive optimistic concurrency check for concurrent transactions. Also implements `DbRead` and `DbWrite`.
- `SharedStore` now implements `Transactable`, giving it `begin_transaction()` capabilities.

## Checklist:


signer/src/storage/memory/store.rs
```diff
@@ -0,0 +1,415 @@
...
+    /// Reference to the original store's mutex to commit changes back.
+    pub original_store_mutex: SharedStore,
+    /// Tracks if commit/rollback has been called.
+    pub completed: AtomicBool,
+}
+
+impl TransactionHandle for InMemoryTransaction {
+    async fn commit(self) -> Result<(), Error> {
+        if self.completed.load(std::sync::atomic::Ordering::SeqCst) {
+            panic!("Transaction already completed");
```
matteojug, line 360: Why panicking, vs returning an error? also below

signer/src/storage/memory/store.rs
```diff
@@ -0,0 +1,415 @@
...
+
+        Ok(())
+    }
+
+    async fn rollback(self) -> Result<(), Error> {
+        if self.completed.load(std::sync::atomic::Ordering::SeqCst) {
+            panic!("Transaction already completed");
+        }
+
+        // Rollback is a no-op for in-memory transactions.
```
matteojug, line 396: Don't we still need to acquire a lock (maybe on `self.store`)? Otherwise we could concurrently execute `commit` and `rollback`; it should have no real consequence (as we don't do anything in here, and at the end in `commit` we would still mark completed as true), but it feels wrong.

signer/src/storage/memory/store.rs
```diff
@@ -0,0 +1,415 @@
...
+            });
+        }
+
+        // Commit the changes from the transactional store to the original store.
+        // This copies all fields from transactional_store_content.
+        *original_store = store.clone();
+
+        // Explicitly set the original store's version to be one greater than the version
+        // this transaction started with. This marks that the store has been updated.
+        original_store.version = self.version + 1;
```
matteojug, line 382: IIUC if I start a transaction, do something on the underlying `DbWrite` store, and then `commit`, I will undo everything that happened between the transaction begin and the commit. Am I missing something? Should we increase the `version` on each `DbWrite` operation on the underlying store?

signer/src/storage/memory/store.rs
cylewitruk: I was just lazy and didn't want to add a bunch of additional error variants for the in-mem db. Maybe I should just add a single `InMemoryDb(&'static str)` variant and replace the one I added for optimistic concurrency...

signer/src/storage/memory/store.rs
cylewitruk: Good catch, yeah you're completely right, this is based on the assumption that all writes go through a an `InMemoryTransaction`, but that of course isn't the case.

signer/src/storage/memory/store.rs
matteojug: It's test code so no biggie, OTOH it does seem one we could avoid (given below we use an error). I would prefer removing it, but no strong opinions either way, up to you (feel free to resolve this).

cylewitruk: I agree, I'm going to do like I suggested above with just a single error variant that takes a string, so that we're at least returning an error. It kinda sucks panicking because it makes it more difficult to see what call actually caused the panic in tests, so better to unwrap in the test and get the actual callsite.

signer/src/storage/memory/store.rs
cylewitruk: To keep the interfaces consistent (since `Context::get_storage()` and co. return `DbWrite + DbRead + (new) Transactable`), yeah we need some form of "transaction support". The naive optimistic concurrency approach was the simplest I could think of.

matteojug: Got it, then yeah I guess incrementing the version in the writing methods is the "easiest" way.

signer/src/storage/memory/store.rs
cylewitruk: [82ef200](https://github.com/stacks-sbtc/sbtc/pull/1674/commits/82ef2004c9c8ecba7910bc4bf888a276694a2fa2) + [1288a72](https://github.com/stacks-sbtc/sbtc/pull/1674/commits/1288a72ea5bc52f8e9b46a8383cbbaaf983cbe5d)

signer/src/storage/memory/store.rs
cylewitruk: [d22566f](https://github.com/stacks-sbtc/sbtc/pull/1674/commits/d22566feef969c3b28fe36b7f19e74833b73be08)

signer/src/storage/memory/store.rs
cylewitruk: [0d1332a](https://github.com/stacks-sbtc/sbtc/pull/1674/commits/0d1332a647231579b1a691aa0ba08069941800be)

signer/src/storage/memory/store.rs
```diff
@@ -356,21 +356,24 @@ pub struct InMemoryTransaction {
 
 impl TransactionHandle for InMemoryTransaction {
     async fn commit(self) -> Result<(), Error> {
+        // Lock the transaction's clone of the store and get a guard
+        let store = self.store.lock().await.clone();
```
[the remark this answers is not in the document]
cylewitruk: Yeah, it's our guard to ensure that you can't enter `commit` or `rollback` while another instance is executing, to make sure that the atomic ops are "transactional", i.e. the last:
```rust
self.completed
  .store(true, std::sync::atomic::Ordering::SeqCst);
```
Has completed before another `commit` or `rollback` attempt can check the flag and return its already-completed error.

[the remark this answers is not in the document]
matteojug: Sorry, wrong wording on my side: I thought so, but isn't `let store = self.store.lock().await.clone();` dropping the guard right away?

signer/src/storage/memory/store.rs
matteojug: Yeah good point, one less thing to remember when writing tests using the in memory db.

signer/src/storage/memory/store.rs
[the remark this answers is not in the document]
cylewitruk: Oh, yes definitely, missed to remove that when I was changing the ordering around I guess. Good eye! [93c0102](https://github.com/stacks-sbtc/sbtc/pull/1674/commits/93c0102f93ede04cd5ef21708fbb86939869ee76)

matteojug (approved):

cylewitruk:
> Why we want .rollback() be no-op in in-memory db? It will introduce difference in behaviour in-memory vs postgres and will harm tests, isn't it ?

@MCJOHN974 no-op here just means that the changes applied within the transaction are discarded, same that would happen in any transactional DB

MCJOHN974 (approved):

signer/src/storage/mod.rs
```diff
@@ -29,6 +29,28 @@ use crate::storage::model::CompletedDepositEvent;
 use crate::storage::model::WithdrawalAcceptEvent;
 use crate::storage::model::WithdrawalRejectEvent;
 
+/// Represents a handle to an ongoing database transaction.
+pub trait TransactionHandle: DbRead + DbWrite + Send {
+    /// Commits the transaction.
+    fn commit(self) -> impl Future<Output = Result<(), Error>> + Send;
+    /// Rolls back the transaction.
+    fn rollback(self) -> impl Future<Output = Result<(), Error>> + Send;
```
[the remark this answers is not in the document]
djordon: > it's surprisingly easy to accidentally forgot a `commit()` when writing transactional code, and an explicit `rollback()` makes it very clear that a rollback is intentional.

Yeah me too.

But I think my point was whether we only use this in tests, and if so, we should communicate that this is the case by feature gating it. I'd probably say this about all code. But I think you're saying that we're going to use the `rollback()` in production code at some point, is that right? If so, then I don't think that it makes sense to feature gate it.

signer/src/storage/memory/store.rs
```diff
@@ -0,0 +1,427 @@
...
+        // Lock the original store and get a guard
+        let mut original_store = self.original_store_mutex.lock().await;
+
+        // Naive optimistic concurrency check
+        if self.version != original_store.version {
+            return Err(Error::InMemoryDatabase(format!(
+                "Optimistic concurrency check failed: expected version {}, found version {}",
+                self.version, original_store.version
+            )));
+        }
```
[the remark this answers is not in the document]
djordon: > We are basically doing CoW for the _entire database_ here, so allowing the concurrent modification of the db's contents by multiple transactions can/would result in problematic scenarios as each `commit()` will overwrite the entire database with _that transaction's_ view, effectively discarding any potential changes committed by other transactions during its lifetime, which itself would be a significant departure from how pgsql behaves.
>
> We could definitely _improve_ this with other strategies (like finer-grained version control, change tracking+merging, etc.), but all of which significantly more involved to implement correctly.

Oh right, when there is a conflict caused by having two concurrent transactions attempt to commit, postgres will handle it in a nice way, where the behavior is handled by the queries in the transcation (with things like `ON CONFLICT` clauses). With my proposed change to this in-memory implementation, the last commit causes the other rows to be dropped, which is incorrect. We'd need to write something to merge two datastores in order to have the behavior match the postgres implementation.

Well, what I want to avoid is having this implementation raise an error when there isn't something that we actually need to worry about in our production code. But we don't have that issue right now, so if that happens we can figure out a way to fix it then.

signer/src/storage/memory/store.rs
```diff
@@ -0,0 +1,427 @@
...
+/// Represents an active in-memory transaction.
+pub struct InMemoryTransaction {
+    /// Records the version of the store at the time of transaction creation.
+    pub version: usize,
+    /// Holds a clone of the store's data for operations within this transaction.
+    pub store: SharedStore,
+    /// Reference to the original store's mutex to commit changes back.
+    pub original_store_mutex: SharedStore,
+    /// Tracks if commit/rollback has been called.
+    pub completed: AtomicBool,
```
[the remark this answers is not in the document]
djordon: Yeah, we're in agreement on the desire to make our intent more clear. Actually, we don't have a disagreement; we can keep the variable, it doesn't hurt.

And I'm not against an explicit rollback, I just haven't figured how to prevent myself from adding an explicit commit or rollback. My suggestion above about conditional compilation was more about whether we will use rollback in production code.

signer/src/storage/mod.rs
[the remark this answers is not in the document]
cylewitruk: I think it's fully possible that we may have scenarios where conditional rollback is desirable. But regardless, exposing this API un-gated from a "completeness" perspective is extremely cheap and in my opinion more correct.

matteojug (approved):

PR #1681, opened 2025-05-20
Add Workflow to Deploy Emily

BowTiedDevOps:
This PR deploys emily off specific branches (ones that match the `environment/*` pattern).

Requirements:
- [ ] Add environments in the `sbtc` repository each containing the required environment variables and secrets for the respective environment (the environment names have to match the branch names from which the deployment is made):
  - `environment/dev` or `environment/devnet`
  - `environment/staging`
  - `environment/private-mainnet`
  - `environment/prod` or `environment/mainnet`
  - `approve-emily-deployment` (used only to require a manual approval before the deployment takes place, no variables/secrets have to be added here)
  
Workflow jobs:
- `Build & Diff Emily` - This job prints a diff of the changes in GitHub Step Summary, which has to be reviewed before proceeding
- `Approve Emily Deployment` - This job waits for the manual approval of the workflow before proceeding with the deployment
- `Deploy Emily` - This job builds Emily then deploys it, only runs after the approval step is successful.

Test runs:
- Emily deployment to devnet through this workflow: https://github.com/BowTiedDevOps/sbtc/actions/runs/15139506011
- Colored diff output in step summary (performed after the devnet deployment, hence I've cancelled the workflow): https://github.com/BowTiedDevOps/sbtc/actions/runs/15140197457#summary-42561938508

**IMPORTANT NOTE:** This PR can be safely merged, but the mentioned environments have to be created before their respective deployment branches are published.

### DIFF
--- a/.github/workflows/deploy-emily.yaml
+++ b/.github/workflows/deploy-emily.yaml
@@ -1,0 +1,193 @@
+name: Deploy Emily
+
+on:
+  push:
+    branches:
+      - 'environment/*'
+
+concurrency:
+  group: docker-image-${{ github.head_ref || github.ref || github.run_id }}
+  cancel-in-progress: true
+
+env:
+  EMILY_HANDLER_PATH: "emily/handler"
+  EMILY_CDK_PATH: "emily/cdk"
+  NODE_VERSION: 20
+
+permissions:
+  id-token: write
+
+jobs:
+  diff:
+    name: Build & Diff Emily
+    runs-on: ubuntu-latest
+    environment: ${{ github.ref_name }}
+    env:
+      AWS_ACCOUNT: ${{ vars.AWS_ACCOUNT }}
+      AWS_DEPLOYMENT_PROFILE: ${{ vars.AWS_DEPLOYMENT_PROFILE }}
+      AWS_REGION: ${{ vars.AWS_REGION }}
+      AWS_STAGE: ${{ vars.AWS_STAGE }}
+      CUSTOM_ROOT_DOMAIN_NAME: ${{ vars.CUSTOM_ROOT_DOMAIN_NAME }}
+      DEPLOYER_ADDRESS: ${{ vars.DEPLOYER_ADDRESS }}
+      HOSTED_ZONE_ID: ${{ vars.HOSTED_ZONE_ID }}
+      NUM_SIGNER_API_KEYS: ${{ vars.NUM_SIGNER_API_KEYS }}
+    steps:
+      - name: Checkout Repository
+        id: checkout_repository
+        uses: stacks-sbtc/actions/checkout@181f8c67da2707c66b5e31f24e7418c47adefdd1
+
+      - name: Configure AWS Credentials via OIDC
+        id: configure_aws
+        uses: stacks-sbtc/actions/aws/configure-aws-credentials@181f8c67da2707c66b5e31f24e7418c47adefdd1
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Setup Rust
+        id: setup_rust
+        uses: stacks-sbtc/actions/setup-rust-toolchain@181f8c67da2707c66b5e31f24e7418c47adefdd1
+
+      - name: Install Cargo Lambda
+        id: install_cargo_lambda
+        run: |
+          cargo install cargo-lambda
+
+      - name: Setup Node
+        id: setup_node
+        uses: stacks-sbtc/actions/setup-node@181f8c67da2707c66b5e31f24e7418c47adefdd1
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      - name: Install Zig
+        id: install_zig
+        run: |
+          npm install -g @ziglang/cli -y
+
+      - name: Build Emily
+        id: build_emily
+        working-directory: ${{ env.EMILY_HANDLER_PATH }}
+        run: |
+          echo "Building emily-lambda..."
+          cargo lambda build \
+            --bin emily-lambda \
+            --release \
+            --output-format zip \
+            --x86-64
+          echo "Done building emily lambda."
+
+      - name: Install CDK Packages
+        id: install_cdk_packages
+        working-directory: ${{ env.EMILY_CDK_PATH }}
+        run: |
+          npm i
+
+      - name: CDK Bootstrap & Diff
+        id: cdk_bootstrap_diff
+        working-directory: ${{ env.EMILY_CDK_PATH }}
+        run: |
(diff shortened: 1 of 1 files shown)


fabergat:
A question: the name of the environment is "devnet" or "environment/devnet"?

aldur:
Thanks! Just a request for the env names to be named accurately so we don't need to guess/try/create them multiple times.

.github/workflows/deploy-emily.yaml
```diff
@@ -0,0 +1,193 @@
...
+        id: install_zig
+        run: |
+          npm install -g @ziglang/cli -y
+
+      - name: Build Emily
+        id: build_emily
+        working-directory: ${{ env.EMILY_HANDLER_PATH }}
+        run: |
+          echo "Building emily-lambda..."
+          cargo lambda build \
```
cylewitruk, line 71: It's good practice to use `cargo --locked` to force cargo to use exact versions from `Cargo.lock`

.github/workflows/deploy-emily.yaml
```diff
@@ -0,0 +1,193 @@
...
+      - name: Setup Node
+        id: setup_node
+        uses: stacks-sbtc/actions/setup-node@181f8c67da2707c66b5e31f24e7418c47adefdd1
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      - name: Install Zig
+        id: install_zig
+        run: |
+          npm install -g @ziglang/cli -y
```
MCJOHN974, line 64: Wow, where we use zig ??

cylewitruk:
QQ: What's the plan for testing this? Can we start with only defining non-mainnet environments so this can't accidentally mess things up in prod the first times we try it?

.github/workflows/deploy-emily.yaml
BowTiedDevOps: https://github.com/stacks-sbtc/sbtc/pull/1681/commits/c3e7bc9cff97530539321d50de0cfdbfe81a8b92

BowTiedDevOps:
> QQ: What's the plan for testing this? Can we start with only defining non-mainnet environments so this can't accidentally mess things up in prod the first times we try it?

We're going to test it just as I've tested it on my fork (see the initial PR message) - devnet deployment, manual approval, deploy. We can then move to staging and private-mainnet, before deploying mainnet using this workflow

.github/workflows/deploy-emily.yaml
```diff
@@ -0,0 +1,195 @@
...
```
fdefelici, line 64: `-y` is not a valid flag for `npm install` and can be safely removed to avoid possible failure (probably current behaviour is  that it is ignored)

```suggestion
          npm install -g @ziglang/cli
```

.github/workflows/deploy-emily.yaml
```diff
@@ -0,0 +1,195 @@
...
```
fdefelici, line 164: `-y` is not a valid flag for `npm install` and can be safely removed to avoid possible failure (probably current behaviour is  that it is ignored)

```suggestion
          npm install -g @ziglang/cli
```

fdefelici (approved):

.github/workflows/deploy-emily.yaml
```diff
@@ -0,0 +1,195 @@
+name: Deploy Emily
+
+on:
+  push:
+    branches:
+      - 'environment/*'
+
+concurrency:
+  group: docker-image-${{ github.head_ref || github.ref || github.run_id }}
+  cancel-in-progress: true
```
matteojug, line 10: Can we safely cancel ongoing deployments on the same env, or will it leave the deployment in a strange state (and, eg, require manual intervention)?

also maybe nit, `head_ref` should always be empty according to the docs
>  This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target

BowTiedDevOps: Yes, the run can be safely cancelled - the deployment step is made right at the end, and it either runs or gets cancelled until then.

As per the head ref, you're right, I've removed it - also renamed the group as it was leftover from the docker image one.

matteojug: I couldn't find details about the `cancel-in-progress` behaviour, but If I'm reading correctly https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/canceling-a-workflow#steps-github-takes-to-cancel-a-workflow-run (the general cancel workflow cancel) it seems GH will try to also cancel the in-progress step in multiple ways, so could it still happen that the actual deployment (`npx aws-cdk deploy`) is interrupted?

fabergat: so technically when you start the deploy is aws handling all the steps so the pipeline is only checking the status BUT this could not be safe... can we do something like wait till the end?

cylewitruk: Maybe it would be best to _not_ cancel ongoing runs at all, and let them be queued. This isn't like our PR workflow where we _want_ to cancel things due to new pushes because we want the CI to always run on the latest version (because of the time it takes, and CI on outdated code doesn't help us). I don't think that kind of scenario will be relevant for this workflow?

matteojug (approved):

.github/workflows/deploy-emily.yaml
matteojug: But then having potentially two concurrent deployment may mess things up? Anyway, approved to unblock the PR, we could tackle this issue separately and for the time being just avoid concurrent runs 😅 
Feel free to resolve this convo when you want to merge the PR.

cylewitruk: I guess I was thinking that with the merge queue they actually get queued one-by-one? But maybe that's wrong?

cylewitruk: Oh, this is `on-push`, so yeah maybe they would run concurrently if the branch was pushed to multiple times... maybe there's a better trigger...

BowTiedDevOps: Hm, would it be sufficient triggering on `workflow_dispatch` with only a few people having access to dispatch actions?

matteojug: Even if we were to enable merge queue for the `environment/*` branches, if we add two different PRs to the queue it could still happen that we end up with two sequential pushes close enough to trigger the second deployment before the first one finished; so I don't think the merge queue can help us with the issue of triggering the deployment on the same env "almost" concurrently.

But just to recap the above, this could be an issue only if the first deployment is in the last step of the job (the actual deployment) and we kill it via the concurrency control, or (to be confirmed) we don't kill it but the second one also reaches the deployment step.

matteojug: Opened https://github.com/stacks-sbtc/sbtc/issues/1713 to (eventually) address this.

PR #1688, opened 2025-05-26
[feat]: send RBF status from Emily-cron to Emily

### LINKED ISSUE #1685: [Chore]: Introduce `Rbf` status for Emily deposits
# Chore -  Introduce `Rbf` status for Emily deposits

## 1. Description

We want to add `Rbf` status for Emily deposits. Now, all deposits updates should have optional field for previous transaction, and if it present, Emily set `Rbf` status for previous transaction

## 2. Why It's Necessary

It would allow bridge to handle rbf'd transactions.

## 3. Execution Checklist

- [ ] Emily now have `Rbf` status
- [ ] Deposit update now have optional field for previous transaction
- [ ] Emily return an error and don't do an update if previous transaction is not in Emily db
- [ ] Emily automatically set status of transaction to `Rbf` during deposit update if this tx was marked as previous tx
- [ ] Previous transaction field is allowed only for sidecar
- [ ] Previous transaction field is also stored in db and can be retrieved via GET deposits/

- [ ] (Optional) recursive query to get all hierarchy of rbf'ing

MCJOHN974:
## Description

Closes: #1685

## Changes

- now Emily cron send rbf status for rbf transactions and provide `replaced_by_tx` field

## Testing Information

- Updated test to check `replaced_by_tx` field and to expect new status.

## Checklist:

### DIFF
--- a/emily_cron/app/models/deposit.py
+++ b/emily_cron/app/models/deposit.py
@@ -14,6 +14,7 @@ class RequestStatus(Enum):
     CONFIRMED = "confirmed"
     FAILED = "failed"
     REPROCESSING = "reprocessing"
+    RBF = "rbf"
 
 
 @dataclass
@@ -154,3 +155,4 @@ class DepositUpdate:
     status: str
     status_message: str
     fulfillment: Optional[Fulfillment] = None
+    replaced_by_txid: Optional[str] = None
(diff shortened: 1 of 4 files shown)


Jiloc (commented):
Changes lgtm, but could you also add asserts in the tests that have rbf updates in https://github.com/stacks-sbtc/sbtc/blob/main/emily_cron/test/test_deposit_processor.py to check for the new status and field?

MCJOHN974:
@Jiloc do you think [2bee217](https://github.com/stacks-sbtc/sbtc/pull/1688/commits/2bee2171e66464b832f369c4e820af0f11c3a625) addresses your comment?

Jiloc (approved):
lgtm!

Issue #1136, opened 2024-12-16
[Feature]: Persist discovered P2P peers


cylewitruk:
# Feature - Persist discovered P2P peers

## 1. Description

We should persist our discovered P2P peers to increase resilience if our currently configured seed node goes offline. As we only provide one seed node today, this creates a potential single-point-of-failure.

## 2. Technical Details:
Peers should be stored together with their last-connected-at timestamp, and only peers last seen within _[a reasonable timespan]_ should be considered when re-starting the signer to avoid dynamic IPs which have changed causing unnecessary noise. The timestamp should be updated every time we connect to the peer on that address.

Only a peer's _confirmed external address_ should be persisted. Peers which we cannot reach externally should not be stored. This means that _incoming connections_ are not by themselves viable since they will appear with ephemeral ports.

If an existing peer is seen with a new external address, it should be updated with the new address.

PR #1689, opened 2025-05-26
feat: #1136 persist known, valid p2p peers & consider them during bootstrapping

cylewitruk:
## Description

Closes: #1136

Includes/based on PR #1682 for db timestamp support via sqlx

## Changes

- New db table `p2p_peers` which stores the _most recent_ successful dial (outbound) connection to a peer. We don't update peers based on incoming connections (at the moment) due to the uncertainty of the quality of the address we "see".
- The libp2p event loop upserts to this table upon successful dial, after the signer key has been verified, etc.
- "Known peers" are now considered as part of bootstrapping. The known peers are loaded at startup and added into the bootstrapping behavior. Seed peers are _always_ dialed first, then the known peers are dialed afterwards.
  - Known peers considered are currently those which have been successfully dialed within the last 30 days, with a maximum of 6 (ordered by most recent successful dial). This is setup in `main.rs` to keep this filtering front-and-center, the bootstrap behavior itself does not contain any filtering logic and uses what it is given.
- New tests verifying this logic/behavior.

## Testing Information

I've also tested this quite a bit in devnet, by e.g.:
1) Starting devnet and letting the peers connect and update their `p2p_peers` tables.
2) Stopping the signers
3) Modifying the signer code to comment-out the "dial seeds" portion of the bootstrapping code
4) Starting the signers and verify that they use the known peers to connect via logs and that bootstrapping succeeds

## Checklist:


signer/migrations/0018__p2p_peers.sql
```diff
@@ -0,0 +1,34 @@
+-- Stores information about known libp2p peers.
+CREATE TABLE p2p_peers (
+    -- The libp2p PeerId of the peer (base58 encoded multihash).
+    peer_id TEXT,
```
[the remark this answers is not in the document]
cylewitruk: The `PRIMARY KEY` constraint below implicitly enforces that included columns are `NOT NULL`

signer/src/network/libp2p/swarm.rs
```diff
@@ -305,6 +309,19 @@ impl<'a> SignerSwarmBuilder<'a> {
         self
     }
 
+    /// Add known peers to the builder.
+    pub fn add_known_peers(mut self, peers: &[(PeerId, Multiaddr)]) -> Self {
+        for (peer_id, addr) in peers {
+            if !self.seed_addrs.contains(addr) {
+                self.seed_addrs.push(addr.clone());
+            }
```
matteojug, line 317: Do we want to add them to `seed_addrs`?

signer/src/storage/postgres.rs
```diff
@@ -2478,6 +2481,38 @@ impl super::DbRead for PgStore {
...
+        -- We're storing this here and as a string primarily for monitoring and ergonomics.
+        -- (The peer id is derived from the peer's keypair)
+        public_key TEXT NOT NULL,
+        -- The last known reachable multiaddress for this peer.
+        multiaddress TEXT NOT NULL,
+        -- Timestamp of when this peer was first added.
+        created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+        -- Timestamp of the last update to this peer''s record (e.g. a successful dial).
+        last_updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+    ); */
```
matteojug, line 2499: nit: leftover, also in `update_peer_connection`

signer/src/main.rs
```diff
@@ -39,6 +41,14 @@ use tracing::Span;
 // before proceeding.
 const INITIAL_BOOTSTRAP_DELAY_SECS: u64 = 3;
 
+/// The window of time in which we consider a peer to be known and valid for
+/// inclusion in bootstrapping.
+const KNOWN_PEER_WINDOW: Duration = Duration::from_secs(60 * 60 * 24 * 30); // 30 days
```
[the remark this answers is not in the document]
cylewitruk: The record's timestamp is only updated upon successful dial event at the moment, so if the signers are running for a long time without restarts and/or broken connections, this may even be too short... I think ideally we may want to periodically update the record timestamp for connections where we are the dialer, which could then motivate this being shortened to, say, 1 week (to allow for that _this_ signer has been offline a few days due to technical issues etc.).

signer/src/main.rs
```diff
@@ -248,10 +258,52 @@ async fn run_libp2p_swarm(ctx: impl Context) -> Result<(), Error> {
...
+    // of the static configuration, are generally considered trusted/more stable
+    // and as such are dialed first. Known peers are gathered from the network
+    // and are dialed using their explicit known/verified peer ID for added
+    // security.
+    let known_peers = {
+        // Fetch known peers from the database.
+        let mut db_peers = ctx.get_storage().get_p2p_peers().await.unwrap_or_default();
+
+        // Sort the peers by last successful dialed time, duration-ascending.
+        db_peers.sort_unstable_by(|a, b| b.last_dialed_at.cmp(&a.last_dialed_at));
```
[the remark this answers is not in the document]
cylewitruk: For this kind of case where we know the number of rows will be quite limited (only 1 row per pubkey), I generally prefer to keep the queries as CRUD-like as possible to increase their versatility and make the business rules more explicit in the calling code. If we were to implement other logic involving the peer store, e.g. maintaining a reputation or other metadata, then they could still use the `get_p2p_peers()` method as-is.

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -438,6 +463,27 @@ impl NetworkBehaviour for Behavior {
...
+        // subsequent polls). The overall swarm connection/concurrent dialing
+        // limits will ensure we don't dial too many peers at once and limit the
+        // total number of connections.
+        self.config.known_peers.iter().for_each(|(peer_id, addr)| {
+            // Construct the dialing options and `Dial` event which will be
+            // sent to the swarm.
+            let dial_opts = DialOpts::peer_id(*peer_id)
+                .condition(PeerCondition::DisconnectedAndNotDialing)
+                .addresses(vec![addr.clone()])
+                .build();
```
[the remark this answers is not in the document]
cylewitruk: Dialing in this way is actually _more strict_ as libp2p itself will require that the remote peer id matches before bubbling it up to `ConnectionEstablished` and will instead raise an `OutgoingConnectionFailed` event with `DialError::WrongPeerId`. We don't do this for seeds at the moment because we don't specify the peer id in the seed list (which we probably should).

signer/src/storage/postgres.rs
cylewitruk: Whoops, [77fa390](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/77fa3902a40f651854cad461babcae6ba1eb5c9c)

signer/migrations/0018__p2p_peers.sql
```diff
@@ -0,0 +1,34 @@
...
+   RETURN NEW;
+END;
+$$ language 'plpgsql';
+
+-- Trigger for `p2p_peers` table which sets the `last_updated_at` column
+-- to the current time whenever a row is updated.
+CREATE TRIGGER update_p2p_peers_last_updated_at
+BEFORE UPDATE ON p2p_peers
+FOR EACH ROW
+EXECUTE FUNCTION update_last_dialed_at_column();
```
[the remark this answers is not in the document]
cylewitruk: I had this originally when the column was `updated_at`, but with it now being more specifically `last_dialed_at` I think your suggestion is more appropriate. Revised in [77fa390](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/77fa3902a40f651854cad461babcae6ba1eb5c9c)

signer/migrations/0018__p2p_peers.sql
```diff
@@ -0,0 +1,34 @@
...
+CREATE TABLE p2p_peers (
+    -- The libp2p PeerId of the peer (base58 encoded multihash).
+    peer_id TEXT,
+    -- The public key of the peer
+    public_key BYTEA NOT NULL,
+    -- The last known reachable multiaddress for this peer.
+    address TEXT NOT NULL,
+    -- Timestamp of when this peer was first added.
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    -- Timestamp of the last update to this peer''s record (e.g. a successful
```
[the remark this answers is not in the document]
cylewitruk: [77fa390](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/77fa3902a40f651854cad461babcae6ba1eb5c9c)

signer/migrations/0018__p2p_peers.sql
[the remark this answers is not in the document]
cylewitruk: Sure, [77fa390](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/77fa3902a40f651854cad461babcae6ba1eb5c9c).

signer/src/context/signer_state.rs
```diff
@@ -537,6 +537,18 @@ impl SignerSet {
             .expect("BUG: Failed to acquire read lock")
             .len()
     }
+
+    /// Gets the signer public key for the given peer ID, returning `None` if
+    /// the peer ID is not in the signer set or otherwise unknown.
+    pub fn get_pubkey_for_peer(&self, peer_id: &PeerId) -> Option<PublicKey> {
+        #[allow(clippy::expect_used)]
```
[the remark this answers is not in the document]
cylewitruk: Nope, leftover from another version [ec369bc](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/ec369bc6addc4ea3191591dfd52deca850de5079)

signer/src/network/libp2p/swarm.rs
cylewitruk: Whoops, leftover from some debugging, removed in [b91b706](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/b91b7060949843fb16d14bea02406403657ae70e)

signer/src/main.rs
```diff
@@ -248,10 +258,52 @@ async fn run_libp2p_swarm(ctx: impl Context) -> Result<(), Error> {
...
+                let time_since_last_dialed = OffsetDateTime::now_utc() - *peer.last_dialed_at;
+                let is_seed_addr = config.signer.p2p.seeds.contains(&*peer.address);
+
+                if time_since_last_dialed > KNOWN_PEER_WINDOW || is_seed_addr {
+                    None
+                } else {
+                    let peer_id = *peer.peer_id;
+                    let peer_address = (*peer.address).clone();
+                    Some((peer_id, peer_address))
+                }
```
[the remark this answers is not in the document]
cylewitruk: I knew I was forgetting something... absolutely! [1d0ae65](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/1d0ae65701cac8d56b9599b50ff9fad202f860ae)

signer/src/error.rs
```diff
@@ -752,6 +752,10 @@ pub enum Error {
             amounts = .0.amounts, cap = .0.cap, cap_blocks = .0.cap_blocks, withdrawn_total = .0.withdrawn_total)]
     ExceedsWithdrawalCap(WithdrawalCapContext),
 
+    /// Overflow occurred when manipulating a system time value
+    #[error("an error occurred while converting to/from time types")]
+    TimeConversion,
+
```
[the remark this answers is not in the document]
cylewitruk: Nope, [a5f0a78](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/a5f0a780dce642e2b59598f77a23960a0001ecbb)

signer/src/main.rs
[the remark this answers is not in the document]
matteojug: Since this is tied to a specific peer id it's probably fine, it just seem a really long time during which anything can happen to that IP addr. But again, there's a peer id so it's probably fine (assuming no private key leak, but then we would have way worse issues than this).

signer/src/network/libp2p/swarm.rs
```diff
@@ -658,4 +680,208 @@ mod tests {
...
+
+        // Wait for the swarms to start.
+        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+
+        // Get the peers stored in context 1 after bootstrapping
+        let p2p_peers_1a = context1.get_storage().get_p2p_peers().await.unwrap();
+
+        // Verify that context 1 has the peer from signer 2. We only only validate
+        // context 1's storage because signer 1 is the dialer (incoming connections
+        // do not update stored known peers due to the risk of ephemeral ports/NAT).
```
matteojug: nit: shouldn't we check that context 2 didn't store anything?

signer/migrations/0018__p2p_peers.sql
```diff
@@ -0,0 +1,20 @@
...
+    -- Timestamp of when this peer was first added.
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    -- Timestamp of the last update to this peer's record (e.g. a successful
+    -- dial).
+    last_dialed_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+
+    -- The peer_id is derived from the public_key, so these values should always
+    -- be "connected". We log them both here to make it easier for potential
+    -- monitoring and future queries/joins which only have the public key.
+    PRIMARY KEY (peer_id, public_key)
```
djordon: I think we can just use the `peer_id` right? A particular peer_id should only be associated with one `public_key`.

signer/src/main.rs
```diff
@@ -248,10 +258,56 @@ async fn run_libp2p_swarm(ctx: impl Context) -> Result<(), Error> {
...
+    // of the static configuration, are generally considered trusted/more stable
+    // and as such are dialed first. Known peers are gathered from the network
+    // and are dialed using their explicit known/verified peer ID for added
+    // security.
+    let known_peers = {
+        // Fetch known peers from the database.
+        let mut db_peers = ctx.get_storage().get_p2p_peers().await.unwrap_or_default();
+
+        // Sort the peers by last successful dialed time, duration-ascending.
+        db_peers.sort_unstable_by(|a, b| b.last_dialed_at.cmp(&a.last_dialed_at));
```
djordon: We need to update the comment. This is sorted descending

signer/src/main.rs
```diff
@@ -248,10 +258,56 @@ async fn run_libp2p_swarm(ctx: impl Context) -> Result<(), Error> {
...
+    // failures.
+    //
+    // NOTE: We specify seed and known peers separately as seed peers are part
+    // of the static configuration, are generally considered trusted/more stable
+    // and as such are dialed first. Known peers are gathered from the network
+    // and are dialed using their explicit known/verified peer ID for added
+    // security.
+    let known_peers = {
+        // Fetch known peers from the database.
+        let mut db_peers = ctx.get_storage().get_p2p_peers().await.unwrap_or_default();
```
djordon: Don't we want to panic here?

signer/src/storage/postgres.rs
```diff
@@ -2478,6 +2481,23 @@ impl super::DbRead for PgStore {
...
     }
+
+    async fn get_p2p_peers(&self) -> Result<Vec<model::P2PPeer>, Error> {
+        sqlx::query_as::<_, model::P2PPeer>(
+            r#"
+            SELECT 
+                peer_id, 
+                public_key, 
+                address,
+                last_dialed_at
```
djordon: nit: consistent style nit with leading commas

signer/src/main.rs
```diff
@@ -248,10 +258,56 @@ async fn run_libp2p_swarm(ctx: impl Context) -> Result<(), Error> {
...
+                let is_allowed_peer = ctx
+                    .state()
+                    .current_signer_set()
+                    .is_allowed_peer(&peer.peer_id);
+
+                if time_since_last_dialed > KNOWN_PEER_WINDOW || is_seed_addr || !is_allowed_peer {
+                    None
+                } else {
+                    let peer_id = *peer.peer_id;
+                    let peer_address = (*peer.address).clone();
```
djordon, line 304: Curious, why are these dereferences necessary?

djordon: Mystery solved, the type here is a `DbMultiaddr`

signer/src/context/signer_state.rs
```diff
@@ -537,6 +537,17 @@ impl SignerSet {
             .expect("BUG: Failed to acquire read lock")
             .len()
     }
+
+    /// Gets the signer public key for the given peer ID, returning `None` if
+    /// the peer ID is not in the signer set or otherwise unknown.
+    pub fn get_pubkey_for_peer(&self, peer_id: &PeerId) -> Option<PublicKey> {
+        self.signers
+            .read()
+            .ok()?
```
djordon: Shouldn't we panic here?

signer/migrations/0018__p2p_peers.sql
cylewitruk: This constrains the data to ensure that your above "should" can't be violated :) Otherwise, the `public_key` is largely just metadata for the case that we wanted to join in here in our monitoring scripts or similar.

signer/src/main.rs
cylewitruk: Nah, a seed still will be configured, so if this (for whatever reason) errors then from the **p2p perspective** it has everything it needs to continue.

Assuming it's due to an actual underlying problem, then something else will cause a termination due to an actual fatal failure mode according to said component's business rules.

signer/src/storage/postgres.rs
cylewitruk: Was updated in [d026ba3](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/d026ba37ea8b40f530a6ee3fa70acf77fa32f6fc)

signer/src/context/signer_state.rs
cylewitruk: I didn't because this is only used in non-critical paths, but changed it to panic for consistency: [682407e](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/682407e8205ab9e05ef8194196229b62bbbe1225)

signer/src/main.rs
cylewitruk: [7751dcc](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/7751dccd06a07e4ef86e433eb661b7027b4f321c)

signer/migrations/0018__store_reclaim_script_hash.sql
matteojug, line 1: This will break devnet, just a heads-up to remember we need to fix that when we'll bump the deployment

signer/tests/integration/postgres.rs
```diff
@@ -6138,6 +6138,42 @@ async fn writing_key_rotation_transactions() {
...
+        assert_eq!(peers.len(), 1);
+        assert_eq!(*peers[0].peer_id, peer_id);
+        assert_eq!(peers[0].public_key, pub_key);
+        assert_eq!(*peers[0].address, multiaddr);
+        // Ensure that the last_dialed_at timestamp is within a reasonable
+        // timespan from utc_now.
+        assert!(*peers[0].last_dialed_at - utc_now < time::Duration::seconds(5));
+
+        dbg!(&peers);
+        dbg!(&peers[0].last_dialed_at);
```
matteojug: nit: leftover `dbg!`

matteojug (approved):

signer/migrations/0018__store_reclaim_script_hash.sql
cylewitruk: Ah, right, was hoping to get this in before it went live. You could manually add an entry in the migrations table with the extra `_` as a precaution if you wanted.

signer/migrations/0018__p2p_peers.sql
djordon: Wait, how does having this composite primary key ensure that the intended constraint can't be violated? Doesn't it explicitly allow it, since we can have two rows for a particular `peer_id` so long as the public keys are different?

signer/src/testing/network.rs
```diff
@@ -7,18 +7,21 @@ use crate::testing::get_rng;
...
+pub trait MultiaddrExt {
     /// Generates a random [`Multiaddr`] with the
     /// [`libp2p::core::transport::MemoryTransport`] transport.
-    fn random_memory() -> Multiaddr;
-}
-
-impl RandomMemoryMultiaddr for Multiaddr {
     fn random_memory() -> Multiaddr {
         Protocol::Memory(OsRng.next_u64()).into()
     }
```
djordon: unrelated, but shouldn't this take a random number generator?

signer/src/network/libp2p/bootstrap.rs
```diff
@@ -47,6 +51,19 @@ impl Config {
         self
     }
 
+    /// Adds known peers to the configuration.
+    pub fn with_known_peers<'a, T>(mut self, known_peers: T) -> Self
+    where
+        T: IntoIterator<Item = &'a (PeerId, Multiaddr)>,
```
djordon: Why take a reference when we actually want ownership? Perhaps the caller has an iterator of owned items to feed in here, in which case they can give it up. If they don't then they can do the cloning.

Also, we could just require a `Vec`, since this function is called with `Vec`s.

cylewitruk: Yeah good point. I think my original plan was to do a deep clone of the `addr` because `Multiaddr` is internally an `Arc<Vec<u8>>`, which could be modified "out of band", but there was no simple way to do that.

> Also, we could just require a Vec, since this function is called with Vecs.

I generally avoid having methods take concrete collection types, in this case if a collection of owned items is passed then the items will simply be moved.

Anyway, refactored a bit in [ec96b48](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/ec96b48607b6f01677de5cdeed3e2637d84c9e8d)

signer/src/testing/network.rs
cylewitruk: Yeah I think I had noticed that and dodged it, but yeah just as well: [8f88f86](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/8f88f86849cd9a7e9eb5a07ae64664ea2c7131c6)

signer/tests/integration/postgres.rs
cylewitruk: Took care of that one in [8f88f86](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/8f88f86849cd9a7e9eb5a07ae64664ea2c7131c6)

signer/migrations/0018__p2p_peers.sql
cylewitruk: Yeah, sorry, I went through this too quickly -- you're correct. I think my reasoning when I wrote this was that they are indeed 1:1. Perhaps we should instead make the `public_key` the PK since the `peer_id` is derived from it.

signer/src/main.rs
djordon: Ahh okay, well we should log the error here though.

signer/migrations/0018__p2p_peers.sql
djordon: Yeah, I think making the `public_key` the primary key is the way to go, but I'm not sure. Although a particular signer's IP (and thus `peer_id`) can change, we want to track the last known `peer_id` that we connected to, since that will be the one that we can contact (or something like that?).

signer/src/network/libp2p/network.rs
```diff
@@ -224,10 +225,10 @@ mod tests {
     async fn connected_peers_gossip_to_one_another() {
         clear_env();
 
-        let mut rng = get_rng();
-        let key1 = PrivateKey::new(&mut rng);
-        let key2 = PrivateKey::new(&mut rng);
-        let key3 = PrivateKey::new(&mut rng);
+        let rng = &mut get_rng();
```
djordon, line 228: Interesting, I'm half surprised that this compiled.

signer/migrations/0019__p2p_peers.sql
```diff
@@ -0,0 +1,20 @@
+-- Stores information about known libp2p peers.
+CREATE TABLE p2p_peers (
+    -- The libp2p PeerId of the peer (base58 encoded multihash).
+    peer_id TEXT NOT NULL,
+    -- The public key of the peer
+    public_key BYTEA NOT NULL,
+    -- The last known reachable multiaddress for this peer.
+    address TEXT NOT NULL,
```
djordon: Are we sure that we want to serialize the `peer_id` and `address` as strings? Having once been bitten by using strings instead of bytes, I tend to be cautious and recommend that we use bytea if that is the underlying representation, which appears to be the case for `Multiaddr`, less so for `peer_id` but I'd expect that to be more stable than their string representation.

signer/src/network/libp2p/network.rs
MCJOHN974: After looking at it I feel like it would be nice to rewrite all tests in this style =)

signer/src/network/libp2p/swarm.rs
```diff
@@ -482,7 +498,13 @@ impl SignerSwarm {
 
 #[cfg(test)]
 mod tests {
-    use crate::testing::{context::*, network::RandomMemoryMultiaddr};
+    use rand::RngCore;
```
MCJOHN974:
```suggestion
    use rand::RngCore as _;
```
nit

signer/src/testing/network.rs
```diff
@@ -1,24 +1,27 @@
 //! Test utilities for the `network` module
 
 use libp2p::{Multiaddr, multiaddr::Protocol};
-use rand::{RngCore, rngs::OsRng};
+use rand::RngCore;
```
MCJOHN974, line 4:
```suggestion
use rand::RngCore as _;
```
nit

signer/src/network/libp2p/swarm.rs
```diff
@@ -482,7 +498,13 @@ impl SignerSwarm {
...
+
+    use crate::{
+        keys::PublicKey,
+        storage::DbRead,
```
MCJOHN974:
```suggestion
        storage::DbRead as _,
```
nit

signer/src/network/libp2p/swarm.rs
```diff
@@ -658,4 +682,208 @@ mod tests {
...
+        assert!(*p2p_peers_2a[0].last_dialed_at - utc_now < time::Duration::seconds(5));
+
+        // Trigger shutdown
+        term1.signal_shutdown();
+        term2.signal_shutdown();
+    }
+
+    #[test_log::test(tokio::test)]
+    async fn libp2p_swarm_bootstraps_with_known_peers() {
+        let mut rng = get_rng();
```
MCJOHN974:
```suggestion
        let rng = &mut get_rng();
```
=)

signer/src/network/libp2p/swarm.rs
```diff
@@ -658,4 +682,208 @@ mod tests {
...
+            .enable_memory_transport(true)
+            .with_initial_bootstrap_delay(Duration::ZERO) // Connect immediately
+            .add_known_peers(&[(key2_pub.into(), swarm2_addr.clone())])
+            .add_listen_endpoint(swarm1_addr.clone())
+            .build()
+            .expect("Failed to build swarm 1");
+        let swarm2 = SignerSwarmBuilder::new(&key2)
+            .enable_mdns(false)
+            .enable_kademlia(false)
+            .enable_autonat(false)
```
MCJOHN974, line 851: Do we want to test it with this features enabled? At least we use branching based on if kademlia is on, so I think we should try with kademlia = true ?

signer/migrations/0019__p2p_peers.sql
cylewitruk: In this case both `Multiaddr`s and `PeerId`s have well-known string representations and are _intended_ to be communicated as strings (e.g. handing out a full multiaddr for config files, for example `ip4/1.2.3.4/tcp/4141/p2p/abc123abc123abc123`).

Normally I would opt for binary, but since it's fairly likely that we/others will query this table for monitoring information (or check things based on logs) it felt more prudent to use the human-form, and given the above I feel quite comfortable using their string repr.

signer/src/network/libp2p/swarm.rs
cylewitruk: For this test I'm explicitly not wanting kademlia involved because I want to make sure that the peers are being populated from the database and not some other source.

signer/src/network/libp2p/swarm.rs
cylewitruk: [7b01016](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/7b01016d14d7b896de2702734e347c34c9604730)

signer/src/network/libp2p/swarm.rs
cylewitruk: [7b01016](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/7b01016d14d7b896de2702734e347c34c9604730)

signer/src/testing/network.rs
cylewitruk: [7b01016](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/7b01016d14d7b896de2702734e347c34c9604730)

cylewitruk: We actually use this in a trait bound below

signer/src/network/libp2p/swarm.rs
cylewitruk: [7b01016](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/7b01016d14d7b896de2702734e347c34c9604730)

signer/src/network/libp2p/swarm.rs
cylewitruk: Can do, [70a6847](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/70a6847be7d6d8b4f5d197f67a534011fcf76cf6)

signer/src/main.rs
cylewitruk: Sure, [719774e](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/719774e34189cb8bad305e3482bf2c52a8f423a7)

signer/migrations/0018__p2p_peers.sql
cylewitruk: Well the `peer_id` will never change for a given `public_key`, they're 1:1, so the only thing we're really tracking is the `addr` (and some timestamps).

Here's a version with `public_key` as PK: [1c6ba87](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/1c6ba87caa62f880b2d7a8fc018184912170e44a)+ [1a6b8e3](https://github.com/stacks-sbtc/sbtc/pull/1689/commits/1a6b8e377ef73b0ceec0bf6ed1f7b0d10862bc39) (the second one feels a little better since peerid should never change)

MCJOHN974 (approved):

signer/migrations/0019__p2p_peers.sql
djordon: Yeah, I had the same exact reasoning for storing the strings when I was bitten (the case was storing stacks addresses as strings, we have a ticket briefly describing the issue in https://github.com/stacks-sbtc/sbtc/issues/1231).

djordon: Not sure what to do, just sayin we should think again

djordon: We spoke about this at stand-up and we're going with strings here. Resolving

PR #1691, opened 2025-05-27
fix: sign stacks tx once per tenure per request

matteojug:
## Description

Closes: https://github.com/stacks-sbtc/sbtc/issues/1659

## Changes

Added (yet another) LRU to tx signers keeping the stacks "requests" signed for each bitcoin tenure -- I went with a max size of 2, probably 1 would be enough as well. Also I went with the in memory solution instead of the db since it's not something we need persistence for, as we already only invoke the validation if the message comes from the current coordinator.

On stacks tx signature request validation, we first check if it's known. And after the signature broadcast, we add mark the request as signed.

~This covers only deposits and withdrawals, we could also add it for key rotations but since it was not as susceptible to the issue as the others I skipped it.~

## Testing Information

Added an integration test.

## Checklist:


signer/src/transaction_signer.rs
```diff
@@ -146,6 +147,9 @@ pub struct TxSignerEventLoop<Context, Network, Rng> {
     /// during DKG using the FROST algorithm. This is then used during the
     /// verification of the Stacks rotate-keys transaction.
     pub dkg_verification_state_machines: LruCache<StateMachineId, dkg::verification::StateMachine>,
+    /// Stacks transactions signed during a bitcoin tenure. We don't allow
+    /// signing for the same request multiple times in a tenure.
+    pub stacks_sign_request: LruCache<model::BitcoinBlockHash, HashSet<StacksSignRequestId>>,
```
cylewitruk, line 152: I guess I'm wondering if we even need an `LruCache` here at all, and instead just have a `HashSet<StacksSignRequestId>` that gets `clear()`'d on `BitcoinBlockObserved` event. Or something along those lines anyway, as this is only ever modified in a code path where the `msg.chain_tip` == the signer's canonical tip (which will change every time a `BitcoinBlockObserved` arrives).

This is fine too (functionally), my above thought is more implicit which may be less desirable given some of the complexity in this file, just felt a little overkill at first glance and would save you the added `&mut self` below 🤷

signer/tests/integration/transaction_signer.rs
```diff
@@ -250,6 +255,137 @@ async fn signer_rejects_stacks_txns_with_too_high_a_fee(
...
+        nonce: 1,
+        tx_fee: 100_000,
+        txid: Faker.fake_with_rng::<StacksTxId, _>(&mut rng).into(),
+    };
+    wallet.set_nonce(request.nonce);
+    request.txid = MultisigTx::new_tx(&request.contract_tx, &wallet, request.tx_fee)
+        .tx()
+        .txid();
+
+    // In this test we use `handle_stacks_transaction_sign_request` instead of
```
cylewitruk, line 336: This comment confused me more than it helped 😅 I'd just write "simulate handling the stacks signing request which hasn't been processed yet, which will store its id in the 'already-signed' hashset for the chain tip" or something.

matteojug: Yeah, that wasn't really helpful, simplified a bit in [cf040e6](https://github.com/stacks-sbtc/sbtc/pull/1691/commits/cf040e643281201bb1fa9bf958b519091caaf55e)

signer/src/transaction_signer.rs
matteojug: Yeah I think we should be fine with a cache of max size 1, so then we can just drop the LRU. But I'm not sure if we can rely on `BitcoinBlockObserved`, as it's not always emitted (if we err and `continue` in the block observer) and we emit it "later" than when we processed the bitcoin blocks, so we may risk getting a message that passes validation on the tx signer before we got that tenure `BitcoinBlockObserved`, which would mess things up. I guess we could just keep what's the tenure we are currently using for the hashset, but then are we really gaining something wrt using the LRU with size of 1?

cylewitruk: The above was mostly a thought as LRU's are pretty heavyweight in comparison, but I'm fine leaving it as an LRU(2); LRU(1) feels a _bit_ riskier in-case somehow some message combination makes it through and manages to evict the "correct" hashset.

cylewitruk (dismissed):

signer/src/transaction_signer.rs
```diff
@@ -160,6 +168,35 @@ struct AcceptedSigHash {
     public_key: PublicKeyXOnly,
 }
 
+/// An enum identifying requests for which we can sign for on stacks only once
+/// per tenure.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub enum StacksSignRequestId {
```
djordon, line 174: Spoke about this offline and think that we should add variants for the other stacks sign requests that we have. It doesn't hurt to do so.

djordon, line 174: Can we implement display here too, so that our logs aren't too painful on our precious eyes.

signer/src/stacks/contracts.rs
```diff
@@ -1445,7 +1445,7 @@ impl RotateKeysErrorMsg {
 
 /// A wrapper type for smart contract deployment that implements
 /// AsTxPayload.
-#[derive(Clone, Copy, Debug, Hash, PartialEq)]
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, strum::Display)]
```
djordon, line 1448: nit: I would also add
```rust
#[strum(serialize_all = "snake_case")]
```
or
```rust
#[strum(serialize_all = "SCREAMING_SNAKE_CASE")]
```
but this is all for logging at the moment so whatever is fine. But if we were to serialize this to a string in to the database then we should use `snake_case`.

djordon (approved):
Looks good

signer/src/stacks/contracts.rs
matteojug: Added in [8707546](https://github.com/stacks-sbtc/sbtc/pull/1691/commits/8707546b23435573b660458ad1162856f3e0cdb1)

djordon (approved):

Issue #1694, opened 2025-05-28
[Chore]: Prevent loading a signer set with more than 16 signers


matteojug:
# Chore - Prevent loading a signer set with more than 16 signers

## 1. Description

The smart contract currently doesn't support signer sets with more than 16 signers (see https://github.com/stacks-sbtc/sbtc/issues/1620, point 4).

To be sure we don't accidentally trigger this issue, we could add a check on bootstrap to prevent the signer from booting if we are loading more than 16 signers.

Issue #1620, opened 2025-04-22
[Feature]: Write new versions of the sBTC smart contracts


djordon:
# Feature - Write new versions of the sBTC smart contracts

## 1. Description

There are a few changes that we want to make to the smart contracts. They include:

1. In `sbtc-withdrawals.clar`, modify the `reject-withdrawal-request` function to accept a bitcoin block hash and a bitcoin block height so that we error if the current bitcoin block hash at the given height differs from the one given as input.
2. In `sbtc-withdrawals.clar`, make the dust limit depend on the recipient address instead of using the one that works for P2PKH.
3. Add a read-only function to `sbtc-bootstrap-signers.clar` that returns the first aggregate key.
4. In `sbtc-bootstrap-signers.clar`, update the `pubkeys-to-spend-script` function to produce the same address that stacks-core would construct when there are more than 16 public keys given as input.
5. In `sbtc-deposit.clar`, reduce the `dust-limit` to one.
6. Add a reasonable minimum max-fee for withdrawals (https://github.com/stacks-sbtc/sbtc/issues/1292). And perhaps the signers should be able to influence this value.
7. In `sbtc-withdrawals.clar`, the `bitcoin-txid` and the `sweep-txid` are redundant in the `accept-withdrawal-request` function. This means one of them can be removed from `complete-withdrawals` and any helpers.
8. In [`complete-individual-withdrawal-helper`](https://github.com/stacks-network/sbtc/blob/54d0a656fa91c6585e169e02a06157b57bdb939d/contracts/contracts/sbtc-withdrawal.clar#L267C240-L267C269), sweep-txid is not validated (like the other optional values) before unwrapping it which could result in a panic instead of an ERR_WITHDRAWAL indicating the index of the failed withdrawal.9. (Unlikely to be possible) Update `sbtc-registry.clar` to remove the assert on line 305.
9. The token metadata has a few typos, one mentioned in https://github.com/stacks-sbtc/sbtc/issues/1143 -- there is the missing `s` and we should remove the ".5T in" in the description. To do that we need to expose the `protocol-set-token-uri` public function in the `sbtc-token` smart contract and update it to a URI with the correct metadata.
10. Maybe do something smarter about max fees during withdrawal requests. For example, we can have the last fees of recent withdrawal requests inform the minimum max fee of withdrawal requests, or something. Or maybe we don't do anything else. Relates to https://github.com/stacks-sbtc/sbtc/issues/1292.
[...]

PR #1695, opened 2025-05-29
[feat]: restrict more then 16 signers

MCJOHN974:
## Description

Closes: #1694

## Changes

Settings will fail validation if `bootstrap_signing_set` contains more then 16 signatures.

## Testing Information

Added unittest to verify validation fail

## Checklist:

### DIFF
--- a/signer/src/config/mod.rs
+++ b/signer/src/config/mod.rs
@@ -38,6 +38,10 @@ pub const MAX_BITCOIN_PROCESSING_DELAY_SECONDS: u64 = 300;
 /// Maximum configurable delay (in seconds) before processing new SBTC requests.
 pub const MAX_REQUESTS_PROCESSING_DELAY_SECONDS: u64 = 300;
 
+/// Maximum amount of signers supported by our smart contracts
+/// See https://github.com/stacks-sbtc/sbtc/issues/1694
+pub const MAX_SIGNERS: u64 = 16;
+
 /// Trait for validating configuration values.
 trait Validatable {
     /// Validate the configuration values.
@@ -369,6 +373,11 @@ impl Validatable for SignerConfig {
             return Err(ConfigError::Message(err.to_string()));
         }
 
+        if self.bootstrap_signing_set.len() > MAX_SIGNERS as usize {
+            let err = SignerConfigError::TooManySigners(self.bootstrap_signing_set.len() as u64);
+            return Err(ConfigError::Message(err.to_string()));
+        }
+
         if self.deployer.is_mainnet() != self.network.is_mainnet() {
             let err = SignerConfigError::NetworkDeployerMismatch;
             return Err(ConfigError::Message(err.to_string()));
@@ -568,7 +577,11 @@ mod tests {
     use crate::config::serialization::try_parse_p2p_multiaddr;
     use crate::error::Error;
     use crate::testing::clear_env;
+    use crate::testing::get_rng;
     use crate::testing::set_var;
+
+    use fake::Fake;
+    use fake::Faker;
 
     use super::*;
     use test_case::test_case;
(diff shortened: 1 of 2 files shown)


signer/src/config/mod.rs
```diff
@@ -369,6 +373,11 @@ impl Validatable for SignerConfig {
             return Err(ConfigError::Message(err.to_string()));
         }
 
+        if self.bootstrap_signing_set.len() > MAX_SIGNERS as usize {
```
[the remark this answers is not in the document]
MCJOHN974: This variable used only once so it can make sense however I prefer to see explicit bitness where possible

signer/src/config/mod.rs
```diff
@@ -38,6 +38,10 @@ pub const MAX_BITCOIN_PROCESSING_DELAY_SECONDS: u64 = 300;
 /// Maximum configurable delay (in seconds) before processing new SBTC requests.
 pub const MAX_REQUESTS_PROCESSING_DELAY_SECONDS: u64 = 300;
 
+/// Maximum amount of signers supported by our smart contracts
+/// See https://github.com/stacks-sbtc/sbtc/issues/1694
+pub const MAX_SIGNERS: u64 = 16;
```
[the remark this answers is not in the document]
MCJOHN974: I think we can document it in default config file [a18631a](https://github.com/stacks-sbtc/sbtc/pull/1695/commits/a18631a5baf59057250429f8ab05516406ce6cd0)

signer/src/config/mod.rs
```diff
@@ -38,6 +38,10 @@ pub const MAX_BITCOIN_PROCESSING_DELAY_SECONDS: u64 = 300;
 /// Maximum configurable delay (in seconds) before processing new SBTC requests.
 pub const MAX_REQUESTS_PROCESSING_DELAY_SECONDS: u64 = 300;
 
+/// Maximum amount of signers supported by our smart contracts
+/// See https://github.com/stacks-sbtc/sbtc/issues/1694
```
matteojug, line 42: nit: I would refer to 1620 here for the actual issue.

matteojug (dismissed):

signer/src/config/mod.rs
MCJOHN974: Imo after reading plain 1620 I would have no idea how this related to this PR. Even with clarification that only 4th point related -- better but still I don't really like explanation, so I'd point to issue of this PR and maybe sometime somebody will post explanation under the issue

matteojug: My (very shallow) fear is that we are linking to an issue that will be closed (and forgotten) after this is merged, and it's not the one tracking the underlying issue that we need to solve, so we could avoid this indirection and just point to the actual issue. Although it's true that 1620 contains *a lot* of sutff, so pointing out to the specific point would def help.

MCJOHN974: Hmm, I don't think there is anything bad refering to closed issues, they are closed, not deleted, and if from issue description it would be unclear why 16 the one for whom it will be unclear will post updates (even to closed issue). And also this one contains link to 1620 so it provide not less informaiton

cylewitruk: I would keep the in-code issue linked to #1694 (which itself links to #1620) because it gives slightly more context, and update point 4 in #1620 to include a link to #1694 (or add it as a sub-task) which will show its status as completed.

signer/src/config/default.toml
```diff
@@ -129,6 +129,9 @@ db_endpoint = "postgresql://postgres:postgres@127.0.0.1:5432/signer"
 
 # A complete list of (compressed) public keys for known bootstrap signer
 # peers who are approved to be in the sBTC signer set.
+# Bootstrap signer set can be at most 16 signers, see
+# https://github.com/stacks-sbtc/sbtc/issues/1694 for more info.
+# Bootstrap signer set must contain public key of the signer itself.
```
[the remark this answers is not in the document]
MCJOHN974: [db8068a](https://github.com/stacks-sbtc/sbtc/pull/1695/commits/db8068ab32dcd42e5fe59b9a19f06edd42225903)

cylewitruk (approved):

PR #1699, opened 2025-06-02
refactor: centralize more metric logic

djordon:
## Description

This is part of https://github.com/stacks-sbtc/sbtc/issues/1490

## Changes

* Move some of the metrics to a more central location

## Testing Information

This is just a refactor that moves metrics logic around. We don't really test logging and stuff like that yet.

## Checklist:

### DIFF
--- a/signer/src/metrics.rs
+++ b/signer/src/metrics.rs
@@ -2,7 +2,7 @@
 //!
 
 use std::net::SocketAddr;
-use std::time::Instant;
+use std::time::Duration;
 
 use metrics_exporter_prometheus::PrometheusBuilder;
 
@@ -84,7 +84,7 @@ impl Metrics {
     /// itself included deposits and withdrawals that we know about and
     /// that there were no duplicates with the request. There could still
     /// be deposits and withdrawals that failed validation.
-    pub fn increment_presign_validation(instant: Instant, presign_result: &Result<(), Error>) {
+    pub fn increment_presign_validation(elapsed: Duration, presign_result: &Result<(), Error>) {
         let status = if presign_result.is_ok() {
             "success"
         } else {
@@ -96,7 +96,7 @@ impl Metrics {
             "kind" => "sweep-presign",
             "status" => status,
         )
-        .record(instant.elapsed());
+        .record(elapsed);
 
         metrics::counter!(
             Metrics::SignRequestsTotal,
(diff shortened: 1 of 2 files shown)


signer/src/metrics.rs
```diff
@@ -71,6 +73,81 @@ impl Metrics {
...
+    /// whether the presign validation finished successfully. Also record
+    /// the amount of time that it took to run the validation.
+    ///
+    /// # Notes
+    ///
+    /// If the presign validation result is Ok, just means that the request
+    /// itself included deposits and withdrawals that we know about and
+    /// that there were no duplicates with the request. There could still
+    /// be deposits and withdrawals that failed validation.
+    pub fn increment_presign_validation(instant: Instant, presign_result: &Result<(), Error>) {
```
cylewitruk: Wouldn't it be better to accept a `Duration` instead of `Instant` (here and below), and require the caller to pass `.elapsed()`?

djordon: Yeah, that does feel better, will update.

djordon: Done in https://github.com/stacks-sbtc/sbtc/pull/1699/commits/fe740626ed44f38da34bc362b343a6e9eae42aae

signer/src/transaction_signer.rs
```diff
@@ -765,21 +732,8 @@ where
...
-
-                        metrics::counter!(
-                            Metrics::SignRequestsTotal,
-                            "blockchain" => BITCOIN_BLOCKCHAIN,
-                            "kind" => "sweep",
-                            "status" => validation_status,
-                        )
-                        .increment(1);
+                        let sighash_result = accepted_sighash.as_ref().map(|_| ());
+                        Metrics::increment_bitcoin_validation(sighash_result);
```
MCJOHN974: Maybe allow `increment_bitcoin_validation` have some other types of parameters and avoid this mapping?

djordon: Yeah, that's what I tried first, but then I realized that the type isn't used at all, so it kinda felt like noise. But it's probably overall better to change it so let me see.

djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1699/commits/5c5d854a7b328993f42f127459ae9e7150a6ea65

MCJOHN974 (approved):

PR #1701, opened 2025-06-02
feat: generate pseudo-deterministic private shares during DKG

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1303

## Changes

* Generate secret shares using a random number generator that was seeded with the signer's private key and the bitcoin block hash and height at the time that the DKG started. This makes the secret shares deterministically generated but still pseudo-random to outside observers.
* Made a small refactor since the random number generator used when processing messages is managed by the wrapping state machine.

## Testing Information

I added an integration test that checks that the DKG generates the same secret shares when given the same inputs, but changes with a new bitcoin block.

I'm going to add a test that makes sure that the nonces generated during signing are still random. But this is good for review now.

## Checklist:


signer/src/error.rs
```diff
@@ -606,8 +606,8 @@ pub enum Error {
     TypeConversion,
 
     /// Encryption error
-    #[error("encryption error")]
-    Encryption,
+    #[error("received an error when encrypting the signer state for storate {0}")]
```
MCJOHN974:
```suggestion
    #[error("received an error when encrypting the signer state for storage {0}")]
```

signer/src/testing/wsts.rs
```diff
@@ -260,25 +262,22 @@ impl Signer {
...
-                };
-
                 let outbound_packets = self
                     .wsts_signer
-                    .process_inbound_messages(&[packet], &mut rng)
+                    .process(&wsts_msg.inner)
                     .expect("message processing failed");
 
-                for packet in outbound_packets {
+                for message in outbound_packets {
```
MCJOHN974: Nit: if we renaming `packet` => `message` then maybe rename `outbound_packets` => `outbound_messages` as well?

signer/src/error.rs
djordon: https://github.com/stacks-sbtc/sbtc/pull/1701/commits/f1578fe47d271ba597faa3b85086057538e9fc4c

signer/src/testing/wsts.rs
djordon: Makes sense, updated in https://github.com/stacks-sbtc/sbtc/pull/1701/commits/92c72eea9fef0e45cdaa853a01a9a2ac89c1b570.

Issue #1484, opened 2025-03-06
[Feature]: Store hash of reclaim script in signers databases


djordon:
# Feature - Store hash of reclaim script in signers databases

## 1. Description

Right now the signers store the entire reclaim script in the database. Storing the entire script is unnecessary, since they can make due with only a hash of the reclaim script. This ticket is to migrate over to storing only a hash of the reclaim script.

### 1.1 Context & Purpose

The purpose of this change is to limit the amount of data that gets stored in the signers' databases.

## 2. Technical Details:

Add a new column for the reclaim script hash. Then we need to populate this column with the right values going forward, and read it from the database using a new wrapped type, similar to what we do for the `ScriptPubKey` type.

The hash that we need is captured by the `bitcoin::TapNodeHash` type, so we'll need to create a wrapper type for it and add a new field to the `bitcoin::utxo::DepositRequest` and `storage::model::DepositRequest` structs.

For the backfill, we might be able to populate the new column with the correct values using only SQL. But doing so requires the pgcrypto extension for SHA256, so maybe it is off the table. If it's on the table, the hash that needs to be stored is a tagged hash of the reclaim script, where I believe the tag is "TapLeaf".

### 2.1 Acceptance Criteria:
   - [ ] The signers only store a hash of the reclaim script in their databases.

PR #1703, opened 2025-06-04
feat: store hashes of reclaim scripts in signer db

MCJOHN974:
## Description

Partially work with #1484

## Changes

This PR introduce new field in the db -- reclaim script hash. It is first step of migration to storing _only_ reclaim script hash, for now this PR store both reclaim script itself and it's hash, keeping old logic with using reclaim script itself. Migration to using new logic will be addressed in another PR. TODO: split issue in two.

## Checklist:


signer/src/storage/model.rs
```diff
@@ -1004,6 +1011,52 @@ impl PartialOrd for StacksPrincipal {
...
+    }
+
+    /// Convert a byte array into a taproot script hash.
+    pub fn from_byte_array(bytes: [u8; 32]) -> Self {
+        bitcoin::TapNodeHash::from_byte_array(bytes).into()
+    }
+
+    /// Compute a taproot script hash from a script pubkey.
+    pub fn from_script_pubkey(script_pubkey: &ScriptPubKey) -> Self {
+        // TODO: I have no idea what this LeafVersion means atm
```
MCJOHN974: I think I used correct LeafVersion here but somebody confirm pls

signer/tests/integration/block_observer.rs
```diff
@@ -1823,6 +1824,7 @@ fn make_coinbase_deposit_request(
 
     let deposit_script = deposit_inputs.deposit_script();
     let reclaim_script = reclaim_inputs.reclaim_script();
+    let reclaim_script_hash = TaprootScriptHash::from_script_pubkey(&reclaim_script.clone().into());
```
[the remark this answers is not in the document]
MCJOHN974: Not really, without `.clone` we have borrow after move here.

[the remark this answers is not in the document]
MCJOHN974: Well, I don't really like the fact that we  using both local wrappers and structs from crate but [f4e27ce](https://github.com/stacks-sbtc/sbtc/pull/1703/commits/f4e27ce803da491ea99d7885f041c547cdd81461) avoids some clones

signer/tests/integration/utxo_construction.rs
```diff
@@ -128,6 +129,7 @@ where
         amount: dep.amount,
         deposit_script: dep.deposit_script.clone(),
         reclaim_script: dep.reclaim_script.clone(),
+        reclaim_script_hash: TaprootScriptHash::new(),
```
djordon: Why not set the actual hash here with something like
```suggestion
        reclaim_script_hash: TaprootScriptHash::from_script_buf(&dep.reclaim_script),
```

signer/src/storage/postgres.rs
```diff
@@ -62,6 +62,8 @@ struct DepositStatusSummary {
     deposit_script: model::ScriptPubKey,
     /// The reclaim script for the deposit.
     reclaim_script: model::ScriptPubKey,
+    /// The hash of reclaim script for the deposit.
+    reclaim_script_hash: model::TaprootScriptHash,
```
djordon: I think this needs to be an `Option<model::TaprootScriptHash>`, since not all deposit requests in the signers database will have a value here.

signer/src/storage/model.rs
```diff
@@ -193,6 +193,8 @@ pub struct DepositRequest {
     pub spend_script: Bytes,
     /// Script spendable by the depositor.
     pub reclaim_script: Bytes,
+    /// SHA-256 hash of the reclaim script.
+    pub reclaim_script_hash: TaprootScriptHash,
```
djordon: I believe this needs to be an `Option<TaprootScriptHash>` for now, since the value in the column could be `NULL`.

signer/src/storage/model.rs
```diff
@@ -222,18 +224,21 @@ pub struct DepositRequest {
 
 impl From<Deposit> for DepositRequest {
     fn from(deposit: Deposit) -> Self {
-        let tx_input_iter = deposit.tx_info.vin.into_iter();
+        let tx_input_iter = deposit.clone().tx_info.vin.into_iter();
```
djordon: I don't think that you need this clone here.

signer/src/storage/model.rs
djordon: This is the correct `LeafVersion` version. Also, right now, the alternative here is the taproot annex version, which is essentially unused at the moment (I think it may be considered non-standard by default or something like that).

signer/src/storage/model.rs
```diff
@@ -1004,6 +1009,58 @@ impl PartialOrd for StacksPrincipal {
...
+        self.0.to_byte_array()
+    }
+
+    /// Convert a byte array into a taproot script hash.
+    pub fn from_byte_array(bytes: [u8; 32]) -> Self {
+        bitcoin::TapNodeHash::from_byte_array(bytes).into()
+    }
+
+    /// Compute a taproot script hash from a script pubkey.
+    pub fn from_script_pubkey(script_pubkey: &ScriptPubKey) -> Self {
```
djordon: The implementation here should go through `TaprootScriptHash::from_script_buf`.

MCJOHN974: [934c1eb](https://github.com/stacks-sbtc/sbtc/pull/1703/commits/934c1eb6cefc975f5d7e007b3b6267c6f10d1684)

signer/src/storage/model.rs
MCJOHN974: [934c1eb](https://github.com/stacks-sbtc/sbtc/pull/1703/commits/934c1eb6cefc975f5d7e007b3b6267c6f10d1684)

signer/src/storage/model.rs
MCJOHN974: [934c1eb](https://github.com/stacks-sbtc/sbtc/pull/1703/commits/934c1eb6cefc975f5d7e007b3b6267c6f10d1684)

signer/src/storage/postgres.rs
MCJOHN974: [934c1eb](https://github.com/stacks-sbtc/sbtc/pull/1703/commits/934c1eb6cefc975f5d7e007b3b6267c6f10d1684)

signer/src/storage/model.rs
```diff
@@ -1004,6 +1009,56 @@ impl PartialOrd for StacksPrincipal {
...
+    fn from(value: bitcoin::TapNodeHash) -> Self {
+        Self(value)
+    }
+}
+
+impl Default for TaprootScriptHash {
+    fn default() -> Self {
+        TaprootScriptHash::from_byte_array([0; 32])
+    }
+}
```
djordon: Having a Default implementation seems a bit strange here. Better to leave it out unless we need it for some reason.

signer/src/storage/model.rs
```diff
@@ -1004,6 +1009,56 @@ impl PartialOrd for StacksPrincipal {
...
+
+impl TaprootScriptHash {
+    /// Create a new taproot script hash with all zeroes
+    pub fn new() -> Self {
+        Default::default()
+    }
```
djordon: This should be feature gated for testing, since I do not think that we will use it in non-testing code.

signer/src/storage/model.rs
MCJOHN974: It was complain from clippy to have default implementation...

MCJOHN974: `clippy::new-without-default` lint

signer/src/storage/model.rs
MCJOHN974: [ef319cf](https://github.com/stacks-sbtc/sbtc/pull/1703/commits/ef319cfb8efc40416cdaeafdcd25404d46b17cf9)

cylewitruk: Do we need this at all since it's simply calling `Default::default()`? I think Clippy only complains the other way around (empty `new()` but no `Default` impl). Then you could feature-gate the default impl.

signer/src/storage/model.rs
```diff
@@ -1004,6 +1009,56 @@ impl PartialOrd for StacksPrincipal {
...
+    /// Return the inner bytes for the taproot script hash
+    pub fn into_bytes(&self) -> [u8; 32] {
```
cylewitruk: `into_*` methods generally take `self`, I'm surprised clippy didn't complain on that. I'd probably name this `to_byte_array()` to match the underlying type and avoid the general "into consumes self" assumption.

signer/src/storage/model.rs
```diff
@@ -1004,6 +1009,56 @@ impl PartialOrd for StacksPrincipal {
...
+        bitcoin::TapNodeHash::from_byte_array(bytes).into()
+    }
+
+    /// Compute a taproot script hash from a script pubkey.
+    pub fn from_script_pubkey(script_pubkey: &ScriptPubKey) -> Self {
+        Self::from_script_buf(script_pubkey)
+    }
+
+    /// Compute a taproot script hash from a scriptbuf
+    pub fn from_script_buf(script_buf: &ScriptBuf) -> Self {
```
cylewitruk: Any reason to not use `From` impls here and for some of the others?

signer/src/storage/model.rs
djordon: Yeah good point. Or just rename this function to `zeros()` or something more specific.

signer/src/storage/model.rs
MCJOHN974: [b14b3ad](https://github.com/stacks-sbtc/sbtc/pull/1703/commits/b14b3adda8e69a4bb39c8af34e12575e57923524)

signer/src/storage/model.rs
MCJOHN974: [a672f73](https://github.com/stacks-sbtc/sbtc/pull/1703/commits/a672f73d9e473fff9c234d637bcef760c95175cb)

signer/src/storage/model.rs
MCJOHN974: okay let's go with `zeros()` since it is more explicit

MCJOHN974: [8c314a7](https://github.com/stacks-sbtc/sbtc/pull/1703/commits/8c314a7267b23a5a61deedcd434ea77a35523b42)

signer/src/storage/model.rs
```diff
@@ -1004,6 +1009,61 @@ impl PartialOrd for StacksPrincipal {
...
+        Self(value)
+    }
+}
+
+#[cfg(feature = "testing")]
+impl Default for TaprootScriptHash {
+    fn default() -> Self {
+        Self::from([0; 32])
+    }
+}
```
djordon: I think you can remove this now, well, after replacing the implementation in `zeros()` with the stuff here.

MCJOHN974: [757d22c](https://github.com/stacks-sbtc/sbtc/pull/1703/commits/757d22c4fe3f1c0a89034cad82919b06d74a5e3f)

djordon (approved):
Looks good

PR #1704, opened 2025-06-04
fix: clear sources in Dockerfile before the COPY

djordon:
## Description

This is a lingering "bug" that only strikes when building a docker image locally where the local branch has conflicting changes with what is on main in a particular way.

## Changes

* Clear the sources downloaded from main before copying over our local changes.

## Testing Information

I tested this out on the `feat/sql-transactions` branch by making a conflicting change there (the one mentioned in https://github.com/stacks-sbtc/sbtc/pull/1673#discussion_r2115350409, where I rename a directory), and I was able to get it to continue past where it failed before (I didn't wait for the full build because life is short... but I should).

## Checklist:

### DIFF
--- a/docker/sbtc/Dockerfile
+++ b/docker/sbtc/Dockerfile
@@ -28,6 +28,11 @@ RUN make install-pnpm && make build
 
 # Build the codebase from the local sources.
 # -------------------------------------------
+# We first remove the existing sources so that the there is no conflict
+# when building from the sources copied over.
+RUN find sbtc signer emily emily_sidecar emily_cron blocklist-client \
+    -mindepth 1 -maxdepth 1 \
+    -exec rm -rf {} +
 COPY . /code/sbtc/
 RUN make install-pnpm && make build
 


docker/sbtc/Dockerfile
```diff
@@ -28,6 +28,11 @@ RUN make install-pnpm && make build
 
 # Build the codebase from the local sources.
 # -------------------------------------------
+# We first remove the existing sources so that the there is no conflict
+# when building from the sources copied over.
+RUN find sbtc signer emily emily_sidecar emily_cron blocklist-client \
+    -mindepth 1 -maxdepth 1 \
+    -exec rm -rf {} +
```
matteojug: qq: why not just `rm -rf sbtc signer ...`?

djordon: Oh, yeah. I started with something more complex where I removed everything except a directory and I used `find` in that case. When I went away from that I just didn't think to simplify it.

djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1704/commits/fe2eb3ee238a6a62a7dd11e97b7e559c742b232d

matteojug (dismissed):
LGTM! I'm not sure if `.generated-sources`, `protobufs` or something else may require pruning as well, but I guess we can tweak this as we go.

docker/sbtc/Dockerfile
```diff
@@ -28,6 +28,9 @@ RUN make install-pnpm && make build
 
 # Build the codebase from the local sources.
 # -------------------------------------------
+# We first remove the existing sources so that the there is no conflict
```
cylewitruk, line 31: I guess I'm pondering over if cloning and building `main` from git (above) really gives us anything, over just using `./`? If not, then we could remove these steps altogether..

cylewitruk: If we keep the `git clone`, it would be good to update the URL to the new org while we're anyways in here.

MCJOHN974 (approved):

docker/sbtc/Dockerfile
djordon: Updated the URL in https://github.com/stacks-sbtc/sbtc/pull/1704/commits/5164f18fc6979cea0b2839505e84329df33d40cf

matteojug (approved):

MCJOHN974 (approved):

PR #1707, opened 2025-06-04
fix: use the correct signer set during DKG and signing rounds

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1706, and closes https://github.com/stacks-sbtc/sbtc/issues/1714.

## Changes

1. Load the wallet using the `SignerState` in the context, saving us from an expensive query.
2. Rename the current signer set variable in the `SignerState` to be the refer to the fact that it is what is in the registry.
1. Use the bootstrap signing set when selecting the coordinator. That set is now the actual signing set, so use that all the time.
2. Use the signing set and signature threshold associated with aggregate key when loading WSTS state machines.

## Testing Information

None yet, but that will be done as part of https://github.com/stacks-sbtc/sbtc/issues/1700.

## Checklist:

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -1672,9 +1672,6 @@ where
         S: Stream<Item = Signed<SignerMessage>>,
         Coordinator: WstsCoordinator,
     {
-        // this assumes that the signer set doesn't change for the duration of this call,
-        // but we're already assuming that the bitcoin chain tip doesn't change
-        // alternately we could hit the DB every time we get a new message
         let signer_set = self.context.config().signer.bootstrap_signing_set.clone();
         tokio::pin!(signal_stream);
 


signer/src/transaction_coordinator.rs
```diff
@@ -1835,15 +1821,11 @@ where
...
-        &self,
-        bitcoin_chain_tip: &model::BitcoinBlockHash,
-        signer_public_keys: &BTreeSet<PublicKey>,
-    ) -> bool {
+    fn is_coordinator(&self, bitcoin_chain_tip: &model::BitcoinBlockHash) -> bool {
         given_key_is_coordinator(
             self.signer_public_key(),
             bitcoin_chain_tip,
-            signer_public_keys,
+            &self.context.config().signer.bootstrap_signing_set,
```
matteojug, line 1817: Using the bootstrap set (unfortunate name we are stuck with I guess) means that when `threshold` signers update the config we will be unable to confirm new requests until we reach the majority of updated configs because of missing coordinator consensus. This seems okay-ish, but I'm wondering if we should instead use the current signer set according to the smart contract -- ie, the last key rotation, together with a fallback for a signer starting fresh without any event.

matteojug (approved):
LGTM! Looking forward to nuking `.threshold` from coordinator and signer

matteojug (changes requested):
Nvm, we also need to change `inspect_msg_chain_tip` coordinator selection, right?
edit: also the coordinator check in coordinator `drive_wsts_state_machine`

signer/src/transaction_coordinator.rs
cylewitruk: Yeah, I ran into similar and ended up removing `is_coordinator` altogether to avoid ambiguity, due to needing to use different sets of keys for DKG vs signing.

And yeah, `inspect_msg_chain_tip` needs to be updated, ideally conditionally depending on the message type.

djordon: Oh right, updated in https://github.com/stacks-sbtc/sbtc/pull/1707/commits/88e8f1ca033c72537e776d5a29edafd01c8751f4

signer/src/block_observer.rs
```diff
@@ -608,12 +608,10 @@ impl<C: Context, B> BlockObserver<C, B> {
         let info = get_signer_set_info(&self.context, chain_tip).await?;
```
matteojug, line 598: nit: the `set_signer_set_info` docstring still reference the old fallbacks

signer/src/context/signer_state.rs
```diff
@@ -38,47 +38,61 @@ impl SignerState {
...
     }
 
-    /// Set the current aggregate key to the given public key.
-    pub fn set_current_aggregate_key(&self, aggregate_key: PublicKey) {
-        self.current_aggregate_key
-            .write()
-            .expect("BUG: Failed to acquire write lock")
-            .replace(aggregate_key);
+    /// Replace the current signer set with the given set of public keys.
+    pub fn update_current_signer_set(&self, public_keys: BTreeSet<PublicKey>) {
```
matteojug: nit: can we clarify in the docstring what is the _current signer set_? ie, IIUC, the set we allow talking to; maybe also in `fn current_signer_set` above

signer/src/transaction_coordinator.rs
```diff
@@ -359,14 +356,20 @@ where
         let chain_tip_hash = &bitcoin_chain_tip.block_hash;
 
         tracing::debug!("loading the signer stacks wallet");
-        let wallet = self.get_signer_wallet(chain_tip_hash).await?;
+        let wallet = self.get_signer_wallet().await?;
```
matteojug, line 359: Now that we fetch the "latest" wallet here instead of one tied to the bitcoin chain tip we are processing, it's not obvious why it's safe: what happens if when we reach this statement, the block observer got a new block (maybe validating a new rotate key? or a fork invalidating the most recent one and we need to fall back while we would be using a now stale one?) and we load a "different" wallet here, while we are still processing the now old `chain_tip_hash`?
It doesn't matter for the later part of the flow since we bail if our tenure is finished, but that's not the case for `check_and_submit_rotate_key_transaction`. It doesn't seem a clear issue but again it's not obvious that it doesn't trigger any strange situation. Just flagging this for a second check.

signer/src/transaction_coordinator.rs
```diff
@@ -1697,7 +1682,7 @@ where
         // this assumes that the signer set doesn't change for the duration of this call,
         // but we're already assuming that the bitcoin chain tip doesn't change
         // alternately we could hit the DB every time we get a new message
-        let signer_set = self.context.state().current_signer_public_keys();
+        let signer_set = self.context.config().signer.bootstrap_signing_set.clone();
```
matteojug, line 1675: nit: I guess we can remove the comment, now that we refer to the config

signer/src/transaction_coordinator.rs
```diff
@@ -2561,20 +2538,20 @@ pub async fn should_coordinate_dkg(
...
     }
 
-    // Trigger dkg if signatures_required has changed
-    if context.state().current_signatures_required() != config.signer.bootstrap_signatures_required
-    {
+    // Allow DKG if we do not have a key rotation event in the database.
+    let Some(registry_signer_info) = context.state().registry_signer_set_info() else {
+        tracing::info!("no signer set information in the registry; proceeding with DKG");
+        return Ok(true);
+    };
```
matteojug: Partially related to https://github.com/stacks-sbtc/sbtc/issues/1720, with this we will keep DKG-ing even if we already have some verified but not-submitted DKG shares in the db. The impact is limited to bootstrapping so maybe we can just keep this, or we could limit the check against `registry_signer_info` to only the case when we have a rotate key (and keep the old logic if not); also in tx signer

signer/src/transaction_coordinator.rs
```diff
@@ -2561,20 +2538,20 @@ pub async fn should_coordinate_dkg(
...
+    if registry_signer_info.signatures_required != config.signer.bootstrap_signatures_required {
         tracing::info!("signatures required has changed; proceeding with DKG");
         return Ok(true);
     }
 
-    // Trigger dkg if signer set changes
-    let signer_set_changed = !context
-        .state()
-        .current_signer_set()
-        .has_same_pubkeys(&config.signer.bootstrap_signing_set);
```
matteojug, line 2575: nit: it seems we are no longer using `has_same_pubkeys`, we could remove it

signer/tests/integration/block_observer.rs
```diff
@@ -1399,6 +1327,11 @@ async fn block_observer_updates_dkg_shares_after_observing_bitcoin_block() {
         .with_mocked_stacks_client()
         .build();
 
+    // We backfill the blockchain data in the database so that the block
+    // observer doesn't need do it, speeding up the test.
+    let chain_tip_info = get_canonical_chain_tip(rpc);
+    backfill_bitcoin_blocks(&db, rpc, &chain_tip_info.hash).await;
```
matteojug: nit: we can use the shiny `fetch_canonical_bitcoin_blockchain`

signer/tests/integration/setup.rs
```diff
@@ -266,11 +272,10 @@ impl TestSweepSetup {
         }
     }
 
-    /// Store a stacks genesis block that is on the canonical Stacks
```
matteojug: nit: unintended cut?

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -1143,7 +1143,6 @@ async fn run_dkg_if_signatures_required_changes(change_signatures_required: bool
 
     // We want current signer set be same as bootstrap_signer_set to prevent DKG trigger
     // because of changed signer set. Also we want this signer set be at least 2 signers
```
matteojug: nit: orphaned comment

signer/src/transaction_coordinator.rs
djordon: Ahhh, yeah we should change this because of bootstrapping. We can get unlucky during that phase where the time between running DKG and getting a new block, I think we've even seen this before. Easy fix.

signer/src/transaction_coordinator.rs
djordon: Oops, will update

signer/src/transaction_coordinator.rs
djordon: Hmmmm, so the case here is what to do if there is a new bitcoin block and key-rotation event in the database but we haven't finished updating the signer state in the block observer? This is a problem because we could be loading the wrong wallet, and thus constructing the wrong transaction. Other signers might load the right one when they get the request to sign the transaction, and if they do we would get a failed signing round because the transaction IDs wouldn't match and they wouldn't send the signature. If the other signers load the wrong wallet too, then the contract call will be confirmed but fail during execution, which wastes fees.

In the old code, we had a similar situation that could occur where the we know about a key-rotation event, but our stacks blockchain data isn't up to date enough and we'd load the wrong wallet. Wait, that can happen now too.

Oh I just realized that in the old code, we could have the block observer observer a bitcoin block invalidating a key rotation and but the code in the coordinator would load the wrong wallet because it was using a stale chain tip. It would be okay because the the signers would end up not signing the associated transaction there too.

Anyway, this was good to think through. It seems to me that in both cases we have some races between the different event loops. However, we cannot do anything wrong, we will just waste a little time.

signer/src/context/signer_state.rs
djordon: Yeah, and it might be worth renaming it to `p2p_signer_set` or `network_signer_set` or something like that.

signer/src/transaction_coordinator.rs
djordon: Removed in https://github.com/stacks-sbtc/sbtc/pull/1707/commits/43778d08ca2ab5ede0ae1b4f70df1082841eed89

signer/src/block_observer.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1707/commits/991243aed074377256d65e8f06b239d39d0f2709

signer/tests/integration/block_observer.rs
djordon: Done in https://github.com/stacks-sbtc/sbtc/pull/1707/commits/596795169442077d980e61a1bf0f22f0f0cd98f2

signer/src/transaction_coordinator.rs
matteojug: I was mostly thinking about the new version vs the one that was getting the key rotation event based on the chain tip we are processing, so in some sense more about the consistency of the data the coordinator is working with. But yeah, in general we could have race conditions between the block observer processing a new (fast) block and the coordinator, and I guess regardless of the cause it's fine as the other signers will respond accordingly. And yeah, worst case scenario, even if everybody is using outdated info, we will just get rejected at Stacks level, so I guess everything is fine.

signer/src/transaction_coordinator.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1707/commits/c191df2f9eb077043fcac155528f9f10ecfde088

signer/src/transaction_coordinator.rs
djordon: Removed in https://github.com/stacks-sbtc/sbtc/pull/1707/commits/ee7d1a1e96fbc2d2345d3f3d31449bd8ef184623

djordon: I lied, I updated it in https://github.com/stacks-sbtc/sbtc/pull/1707/commits/3e27e74480d6a6a0139b94fe09d4fb1d278b169b

signer/src/context/signer_state.rs
djordon: I updated the comment in https://github.com/stacks-sbtc/sbtc/pull/1707/commits/eca2908a38b6256ecd7f0e0aad416ff6e1169597, but did not do the renaming yet.

signer/tests/integration/setup.rs
djordon: yup

signer/tests/integration/transaction_coordinator.rs
djordon: Removed in https://github.com/stacks-sbtc/sbtc/pull/1707/commits/4fa16e0f1e35bdfb0b8c73c55524b5436942cce3

signer/tests/integration/setup.rs
djordon: Restored in https://github.com/stacks-sbtc/sbtc/pull/1707/commits/258bc747077004363dcf078a034f01139d6ce25b

signer/src/context/signer_state.rs
matteojug: Resolving since we can do the renaming in a different PR

signer/src/transaction_coordinator.rs
matteojug: The updated one LGTM, but this was related to a different occurrence I guess (in drive_wsts_state_machine)

djordon: Oh oops

djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1707/commits/57ac7e6f82d1ec52e993d45e3f22d5f95d5dbabf

MCJOHN974 (approved):

PR #1715, opened 2025-06-06
fix: more consistent signer set state

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1714.

## Changes

1. Load the wallet using the `SignerState` in the context, saving us from an expensive query.
2. Rename the current signer set variable in the `SignerState` to be the refer to the fact that it is what is in the registry.

## Checklist:


signer/src/context/signer_state.rs
```diff
@@ -38,47 +38,61 @@ impl SignerState {
...
+    /// Return the signer set info that is currently stored in the smart
+    /// contract.
     #[allow(clippy::unwrap_in_result)]
-    pub fn current_aggregate_key(&self) -> Option<PublicKey> {
-        self.current_aggregate_key
+    pub fn registry_signer_set_info(&self) -> Option<SignerSetInfo> {
+        self.registry_signing_set_info
             .read()
-            .expect("BUG: Failed to acquire read lock")
+            .expect("BUG: Failed to acquire read lock of signer set info")
```
MCJOHN974, line 88: Hmm, why changing this log?

MCJOHN974 (dismissed):

signer/src/context/signer_state.rs
djordon: Because it's more descriptive of what actually failed, allowing us to more easily debug any issues. I should probably update the other lines as well, since if things fail there we won't know which `RwLock` failed.

PR #1723, opened 2025-06-11
feat: switch from reclaim script to reclaim script hash

MCJOHN974:
## Description

Follow up for #1703

Closes: #1484

## NOTE

Since we use two-step deployment, #1703 should be deployed first, then some delay needed, and only then this should be deployed. Otherwise we need to provide more robust migration script.

## Changes

- drops `reclaim_script` column from signer db
- use new `reclaim_script_hash` column instead

## Testing Information

I don't think new tests needed.

## Checklist:

### DIFF
--- a/signer/tests/integration/transaction_coordinator.rs
+++ b/signer/tests/integration/transaction_coordinator.rs
@@ -13,6 +13,7 @@ use bitcoin::Amount;
 use bitcoin::BlockHash;
 use bitcoin::Transaction;
 use bitcoin::hashes::Hash as _;
+use bitcoin::relative::Height;
 use bitcoincore_rpc::RpcApi as _;
 use bitcoincore_rpc_json::GetChainTipsResultTip;
 use bitvec::array::BitArray;
@@ -6179,7 +6180,7 @@ where
             reclaim_script: req.reclaim_script,
             signers_public_key,
             recipient: deposit_inputs.recipient.clone(),
-            lock_time: bitcoin::relative::LockTime::ZERO,
+            lock_time: bitcoin::relative::LockTime::Blocks(Height::from_height(50)),
         });
 
         requests.push(DepositRequest {


MCJOHN974:
Whoa, it seems now `construct_taproot_info` is correctly implemented.

signer/src/bitcoin/utxo.rs
```diff
@@ -499,7 +497,7 @@ impl DepositRequest {
     fn construct_taproot_info(&self, ver: LeafVersion) -> TaprootSpendInfo {
         // For such a simple tree, we construct it by hand.
         let leaf1 = NodeInfo::new_leaf_with_ver(self.deposit_script.clone(), ver);
-        let leaf2 = NodeInfo::new_leaf_with_ver(self.reclaim_script.clone(), ver);
+        let leaf2 = NodeInfo::new_hidden_node(*self.reclaim_script_hash.clone().unwrap());
```
MCJOHN974: Most important change here.

MCJOHN974:
Also open question here -- what to do with `reclaim_script_hash` column. Since it is used in core logic now, it would make sense to make it `NOT NULL`, however, this will fail unless we compute hashes for old rows. 

For now I just removed `Option` from rust code, making it failing just after querying db if query results have old entryes, but I  think it would be nice to eventually make this column `NOT NULL`. Probably just wait for a while and drop all rows...

signer/migrations/0019_drop_reclaim_script_column.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE sbtc_signer.deposit_requests
+DROP COLUMN reclaim_script;
```
djordon: nit: newline here please

djordon (commented):
This basically looks good. We still need to decide on whether we want to make the `reclaim_script_hash` column `NOT NULL` (which requires deleting old deposit requests) or not.

signer/migrations/0019_drop_reclaim_script_column.sql
djordon: Hmmm. Are we also going to delete rows where `reclaim_script_hash` is `NULL` and enforce `NOT NULL` on the column? We'd also have to delete rows from the `deposit_signers` table too if we were to do so because of the foreign key constraint. I could think of one alternative approach where we would know that things would always be correct, but it is simpler for us to just delete these "problematic" rows.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -2134,13 +2137,13 @@ async fn sign_bitcoin_transaction() {
...
     let max_fee = amount / 2;
-    let (deposit_tx, deposit_request, _) =
+    let (deposit_tx, deposit_request, deposit_info) =
         make_deposit_request(&depositor, amount, utxo, max_fee, signers_public_key);
     rpc.send_raw_transaction(&deposit_tx).unwrap();
 
     assert_eq!(deposit_tx.compute_txid(), deposit_request.outpoint.txid);
 
-    let body = deposit_request.as_emily_request(&deposit_tx);
+    let body = deposit_request.as_emily_request(&deposit_tx, deposit_info.reclaim_script);
```
djordon: It feels more natural to move the `as_emily_request` method to the `DepositInfo` struct, since that has everything we need (besides the raw transaction) and we always have one available when we want to make an Emily request.

signer/migrations/0020__drop_reclaim_script_column.sql
```diff
@@ -0,0 +1,26 @@
+BEGIN;
+-- We want to set reclaim_scipt_hash to NOT NULL, since we dropping
+-- reclaim_script and reclaim_script_hash now will replace the reclaim_script
+-- functionality. In order to do so, we need to delete all rows where 
+-- reclaim_script_hash IS NULL, and rows from another tables pointing via
+-- FOREIGHN KEY to rows where reclaim_script_hash IS NULL.
```
djordon:
```suggestion
-- We want to add a NOT NULL constraint on the deposit_request.reclaim_scipt_hash 
-- column and drop the deposit_request.reclaim_script column. In order to do so,
-- we need to delete all rows where the reclaim_script_hash column IS NULL, as
-- well as all rows from other tables with FOREIGN KEY pointers to the
-- deposit_request table where reclaim_script_hash column IS NULL.
```

signer/migrations/0020__drop_reclaim_script_column.sql
```diff
@@ -0,0 +1,26 @@
...
+
+
+-- Since deposit_signers references deposit requests with ON DELETE CASCADE
+-- this should delete nothing, just added for consistency.
```
djordon: This should be updated, looks like you removed the ON DELETE CASCADE part of the query.

signer/migrations/0020__drop_reclaim_script_column.sql
```diff
@@ -0,0 +1,26 @@
+BEGIN;
```
djordon: I don't think the `BEGIN` and `COMMIT` statements matter for our default migration flow. However some migration tools, like flyway, may complain, so we should remove it in case one of our signing partners uses one of those tools to manage migrations.

signer/migrations/0020__drop_reclaim_script_column.sql
MCJOHN974: Nope, I mean this https://github.com/stacks-sbtc/sbtc/blob/5883e75a40f4e7fca69ed4a48e1c922cd31bab68/signer/migrations/0003__create_tables.sql#L67 `ON DELETE CASCADE` and I don't think I removed it.

signer/migrations/0020__drop_reclaim_script_column.sql
MCJOHN974: Hmmm, this migration will delete some data from the db, and I expect this `BEGIN` + `COMMIT` to rollback all deletions if something goes wrong, why wouldn't they matter in our flow? Or are we already doing some transactioning on a higher level?

signer/migrations/0020__drop_reclaim_script_column.sql
MCJOHN974: ooohhh, english =(

signer/tests/integration/transaction_coordinator.rs
MCJOHN974: Yeas, but `DepositInfo` is described in the `sbtc`, and I don't think that we want to modify `sbtc`. So to do so we will need to create a local copy of `DepositInfo` or move `as_emily_request` from being the `DepositRequest` method to being a function (which imo is the best option here).

Do you think we want to make `as_emily_request` a standalone function?

djordon: I wouldn't touch the `sbtc` crate, so that leaves two ways to do it. One is to add it via a new trait in the signer crate (probably in the testing module or someplace like that), and another is to create a stand alone function. Whichever is fine.

signer/migrations/0020__drop_reclaim_script_column.sql
djordon: Ahh, I see. Oh so we can safely delete the rows from the `deposit_requests` table and everything will be fine. 🤔 Well what is here should be equivalent to just deleting the rows in the `deposit_requests` table. I do like being explicit but having the database correctly handle this is also appealing. I guess either is fine.

MCJOHN974: I'd leave this explicit.

 It's both shows that some more columns was deleted to reader and also saves us if this `ON DELETE CASCADE` will not work as expected (and this is possible in some scenarious but I don't think we are such scenario anyways)

signer/migrations/0020__drop_reclaim_script_column.sql
djordon: We run our migrations in a transaction based connection (probably not the right term), so failures get rolled back by default. Similarly, other tools that run migrations already run things in a transaction so that they can safely rollback as well.

signer/tests/integration/transaction_coordinator.rs
MCJOHN974: [87db3fa](https://github.com/stacks-sbtc/sbtc/pull/1723/commits/87db3fa0da4ffcaf3ff777dd80ad9770f62d7481) standalone function added here

signer/migrations/0020__drop_reclaim_script_column.sql
```diff
@@ -0,0 +1,23 @@
+-- We want to add a NOT NULL constraint on the deposit_request.reclaim_scipt_hash 
+-- column and drop the deposit_request.reclaim_script column. In order to do so,
+-- we need to delete all rows where the reclaim_script_hash column IS NULL, as
+-- well as all rows from other tables with FOREIGN KEY pointers to the
+-- deposit_request table where reclaim_script_hash column IS NULL.
+
+
+-- Since deposit_signers references deposit requests with ON DELETE CASCADE
+-- this should delete nothing, just added for consistency.
```
djordon: I would update this to state the purpose a little more clearly. Here is my suggestion
```suggestion
-- Since the deposit_signers table references the deposit_requests table with
-- ON DELETE CASCADE, the following DELETE FROM query is not strictly necessary.
-- However, it was added to make the deletion explicit.
```

djordon (dismissed):
Looks good.

MCJOHN974:
@matteojug ptal

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -6161,26 +6163,37 @@ where
...
 
+        infos.push(DepositInfo {
+            outpoint: req.outpoint,
+            max_fee,
+            deposit_script: req.deposit_script,
+            amount: *amount,
+            reclaim_script: req.reclaim_script,
+            signers_public_key,
+            recipient: deposit_inputs.recipient.clone(),
+            lock_time: bitcoin::relative::LockTime::ZERO,
```
matteojug: nit: shouldn't the locktime be 50?

matteojug (dismissed):

signer/tests/integration/transaction_coordinator.rs
djordon: Yeah it should be, good catch.

MCJOHN974: Hmmm, is it ok that tests didn't catch it 🤔

matteojug (approved):

PR #1727, opened 2025-06-11
refactor: restructure/centralize handling of coordinator key selection algorithm

cylewitruk:
## Description

Something that's been nagging at me for a while, so I went ahead and restructured/centralized the "is coordinator pubkey for" logic a little bit.

## Changes

- Moves the core logic into the `keys` module, which feels more natural to me as it's a cross-cutting "key-related" concern (and helps to reduce the size of `transaction_coordinator.rs` slightly).
- Adds tests for the core logic.
- The updated implementation is slightly more flexible by taking generic parameters (`Borrow<bitcoin::BlockHash>` due to its blanket implementation, allowing both `bitcoin::BlockHash` and our `BitcoinBlockHash` and `BitcoinBlockRef` types to be used interchangeably, which is nice in various calling scenarios, and `Into<PublicKey>` so that we can pass most of our pubkey types depending on what you have at the callsite).
- Centralized the "select coordinator" logic for the `SignerInfo` test helper.
- Minor tweaks around the codebase due to the above.

## Checklist:


signer/src/testing/transaction_signer.rs
```diff
@@ -10,6 +10,7 @@ use crate::context::SignerEvent;
 use crate::context::SignerSignal;
 use crate::context::TxSignerEvent;
 use crate::error::Error;
+use crate::keys::CoordinatorPublicKey;
```
djordon: nit: we should alias it `as _`, like we do for most (all?) traits. Same thing for `signer/src/testing/wsts.rs`, `signer/src/transaction_coordinator.rs`, and `signer/src/transaction_signer.rs` and maybe a few other places.
```suggestion
use crate::keys::CoordinatorPublicKey as _;
```

signer/tests/integration/emily.rs
```diff
@@ -12,6 +12,7 @@ use bitcoincore_rpc_json::Utxo;
 use fake::Fake as _;
 use futures::future::join_all;
 use signer::testing::storage::model::TestBitcoinTxInfo;
+use signer::testing::wsts::SelectCoordinatorPrivateKey;
```
djordon: nit:
```suggestion
use signer::testing::wsts::SelectCoordinatorPrivateKey as _;
```

signer/src/storage/memory/read.rs
```diff
@@ -592,7 +592,7 @@ impl DbRead for SharedStore {
         // Get the blockchain
         let bitcoin_blocks = std::iter::successors(Some(chain_tip), |block_hash| {
             db.bitcoin_blocks
-                .get(block_hash)
+                .get(*block_hash)
```
djordon, line 603: Any idea on why is the dereference necessary now? It seemed to work fine with the doubly reference block hash before.

signer/src/keys.rs
```diff
@@ -467,11 +471,82 @@ impl SignerScriptPubKey for secp256k1::XOnlyPublicKey {
...
+    ) -> Option<PublicKey>
+    where
+        B: Borrow<bitcoin::BlockHash>,
+        K: Copy + Into<PublicKey>,
+    {
+        let num_signers = signer_public_keys.len();
+        if num_signers == 0 {
+            // Handle empty set to avoid panic with % 0
+            return None;
+        }
```
djordon, line 499: Nice catch.

signer/src/keys.rs
```diff
@@ -28,14 +28,18 @@
...
 use std::ops::Deref;
 use std::str::FromStr;
 
 use bitcoin::ScriptBuf;
 use bitcoin::TapTweakHash;
+use bitcoin::hashes::Hash as _;
 use secp256k1::SECP256K1;
 use serde::Deserialize;
 use serde::Serialize;
+use sha2::Digest;
```
djordon: nit
```suggestion
use sha2::Digest as _;
```

signer/src/testing/wsts.rs
```diff
@@ -39,6 +41,48 @@ pub struct SignerInfo {
...
+
+/// Implementation of `SelectCoordinatorPrivateKey` for any type that can be
+/// converted into an iterator of `SignerInfo`.
+impl<I> SelectCoordinatorPrivateKey for I
+where
+    I: ?Sized,
+    for<'a> &'a I: IntoIterator<Item = &'a SignerInfo>,
+{
+    #[track_caller]
+    fn select_coordinator_private_key<B>(&self, bitcoin_chain_tip: B) -> PrivateKey
```
djordon: What does this `#[track_caller]` do?

signer/src/testing/transaction_signer.rs
djordon: Turns out that 90% of my comments so far are like this 🤷🏽‍♂️

signer/src/storage/memory/read.rs
cylewitruk: I played around with this for a while and it seems to be some sort of compiler trait resolution confusion (hashmaps etc. depend on `Borrow<T>` as well), perhaps due to some interaction with implementations in the `bitcoin` crate itself or the `Eq + Hash` bounds somewhere... According to everything that I know and have read it _should_ still work. But it was only a handful of places where the deref from `&&` was needed, so I didn't put too much more effort on it. We could implement an explicit `Borrow<T> for &&` if we really want to avoid the deref, but that felt overkill vs. a couple of `*`'s in test code.

signer/src/testing/wsts.rs
cylewitruk: A recent discovery -- if (sync) methods attributed with `#[track_caller]` panic, the panic source will be at the function's call-site instead of the panic location _within_ the function. In other words, the panic in a test will point to the line in the test at which this function was called instead of the line within this function that panicked, which is usually pretty useless and mostly frustrating. So it's a very nice feature for smaller, panicky test methods vs. needing to return a `Result` and `unwrap()`/`expect()` everything in tests to get the correct (or rather, "interesting") call-site.

It's apparently useless for `async` methods, however.

signer/src/testing/transaction_signer.rs
cylewitruk: Ah, yeah, I've addressed a number of these elsewhere, we're missing quite a few of them.

signer/src/storage/memory/read.rs
cylewitruk: Worth mentioning that I originally wanted to use `AsRef`, but it lacks the blanket impl for itself, so you couldn't pass `(&)bitcoin::BlockHash` (which we do in several places). That's why it's using `Borrow`.

signer/src/keys.rs
cylewitruk: [1f993ca](https://github.com/stacks-sbtc/sbtc/pull/1727/commits/1f993ca191d7a28803ac9b0af3db6805905b91ad)

signer/tests/integration/emily.rs
cylewitruk: [1f993ca](https://github.com/stacks-sbtc/sbtc/pull/1727/commits/1f993ca191d7a28803ac9b0af3db6805905b91ad)

signer/src/testing/transaction_signer.rs
cylewitruk: Attacked a number of them in [1f993ca](https://github.com/stacks-sbtc/sbtc/pull/1727/commits/1f993ca191d7a28803ac9b0af3db6805905b91ad)

signer/src/testing/wsts.rs
cylewitruk: Moved it to the trait fn definition in [1f993ca](https://github.com/stacks-sbtc/sbtc/pull/1727/commits/1f993ca191d7a28803ac9b0af3db6805905b91ad) btw.

MCJOHN974 (dismissed):

MCJOHN974 (approved):

signer/src/keys.rs
```diff
@@ -689,4 +764,124 @@ mod tests {
...
+        let mut bytes = [0u8; 32];
+        bytes[0] = seed; // Simple differentiation
+        // For more varied hashes, could use a simple counter or fill more bytes.
+        // For example, to make it more "hash-like" for testing distribution:
+        let mut full_seed = [0u8; 32];
+        full_seed[0] = seed;
+        full_seed[1] = seed.wrapping_add(1);
+        full_seed[2] = seed.wrapping_add(2);
+        // ... and so on, or use a seeded RNG to fill `bytes`.
+        // For this test, simple differentiation is likely sufficient.
```
matteojug: nit: deadcode?

signer/src/testing/wsts.rs
```diff
@@ -39,6 +41,48 @@ pub struct SignerInfo {
     pub signer_public_keys: BTreeSet<PublicKey>,
 }
 
+/// Trait for selecting the coordinator's private key from a list of items which
+/// contain keypairs.
+pub trait SelectCoordinatorPrivateKey {
+    /// Select the coordinator's private key based on the given list of `SignerInfo`.
```
matteojug: nit: mentioning `SignerInfo` here seems unexpected

matteojug (approved):
LGTM, just some nits

MCJOHN974 (approved):

signer/src/keys.rs
```diff
@@ -476,11 +480,82 @@ impl SignerScriptPubKey for secp256k1::XOnlyPublicKey {
...
 }
 
+/// Utility methods for determining the coordinator public key based on the
+/// underlying set of signers and the bitcoin chain tip.
+pub trait CoordinatorPublicKey {
+    /// Find the coordinator public key
+    fn determine_for<B, K>(
+        bitcoin_chain_tip: B,
+        signer_public_keys: &BTreeSet<K>,
+    ) -> Option<PublicKey>
```
djordon, line 490: Why tie this to the trait instead instead of having this as a free function? Also, I can see why making it generic over `B` is useful, but why make it generic over `K`?

signer/src/keys.rs
djordon: Why not sha256 the seed? Just curious, what's here is fine too.

signer/src/testing/transaction_signer.rs
djordon: Awesome, thanks.

signer/src/keys.rs
cylewitruk: `full_seed` was dead code so I just removed it, the above was sufficient for the simple `u8` differentiator (and makes it easy to see which seed caused any potential issues directly in the block hash). [c8db5e9](https://github.com/stacks-sbtc/sbtc/pull/1727/commits/c8db5e938b17effb2ab1211414b257afd5218064)

cylewitruk: yup [c8db5e9](https://github.com/stacks-sbtc/sbtc/pull/1727/commits/c8db5e938b17effb2ab1211414b257afd5218064)

signer/src/testing/wsts.rs
cylewitruk: yeah, missed to update it, [4c577f5](https://github.com/stacks-sbtc/sbtc/pull/1727/commits/4c577f589c91ae212f328ce24cd99f2b7ae28dd3)

signer/src/keys.rs
cylewitruk: > Why tie this to the trait instead instead of having this as a free function?

Just a preference of taste, I suppose. `CoordinatorPublicKey::determine_for(tip, keys)` gives nice and encapsulated context (though e.g. `BTreeSet::<PublicKey>::determine_for(tip, keys)` is a bit ambiguous, it's not very often one calls associated methods on the collection type like that, and intellisense anyway highlights that it's from the `CoordinatorPublicKey` trait). 🤷 

> Also, I can see why making it generic over B is useful, but why make it generic over K?

No strong argument, largely just ergonomics as we sometimes use `secp256k1` etc. types directly in tests, so you can call `CoordinatorPublicKey::determine_for()` if you've got a set of those and want to find the coordinator key. This will result in something like 5 concrete impls, but optimized they should be very small and/or optimized away if unused, and only the final`nth` key is copied and mapped, so no real measurable performance impact.

djordon-stacks:

PR #1726, opened 2025-06-11
feat: additional test helpers & utilities (part 1)

cylewitruk:
## Description

A collection of some smaller test helper/ergonomics improvements. Not everything is necessarily used here but will be more evident in the next parts.

## Changes

- Removed some outdated `todo`'s from `default.toml`.
- New, versatile `TestUtilityError` type and error variant to make it easy to return arbitrary errors from test utilities without needing to create new error variants.
- New utility/helper traits, types and methods, including `SleepAsyncExt`, `ResultIterExt`, `FutureExt`, `FuturesIterExt` and `Sleep`. These are very minor, but provide small ergonomic/verbosity wins in test code. Refactored `tests/integration/transaction_coordinator.rs` to show their usages.
- Additional `wait_for_latest_dkg_to_become_verified()` and `wait_for_key_rotation_event` test-wait helpers.
- Introduce `#[track_caller]` to more test utility functions and having said methods return better panic messages. This attribute places the actual panic at the _call site_ of the function instead of within the utility function itself.
- Changed a number of `.unwrap()`s to panic with better messages.
- Changed a number of `.generate_blocks(1).pop().unwrap()` and similar usages to `.generate_block()`.
- Other small tweaks/adjustments/etc.

## Checklist:

- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #1729, opened 2025-06-11
feat: new test helpers for parsing contract calls from stacks transactions

cylewitruk:
## Description

Adds test helper/utilities for parsing `StacksTransaction`s into their respective contract call types. 

Their usage isn't obvious in this PR, but will be used in subsequent PR's introducing `MockStacksInteract` adapters, optionally allowing submitted `StacksTransaction`s to be parsed and update shared state during the ongoing operation of sets of signers in tests, making the mocks behave more like a real stacks node.

## Notes

- This only includes parsing for `CompleteDepositV1` and `RotateKeysV1` as that's what was used in my grow-signer-set test. The other calls are fairly trivial to add in a followup PR.
- The diff is roughly 40% tests, testing the test code...
- This PR targets PR #1726.

## Checklist:


signer/src/testing/stacks/contracts.rs
```diff
@@ -0,0 +1,767 @@
...
+    /// Tries to parse the argument at `pos` as a Bitcoin block hash ([`bitcoin::BlockHash`]).
+    fn try_parse_bitcoin_block_hash(
+        &self,
+        pos: usize,
+        contract_name: &'static str,
+        function_name: &'static str,
+    ) -> Result<bitcoin::BlockHash, ParseContractCallError> {
+        self.try_get_arg(pos, contract_name, function_name)?
+            .clone()
+            .expect_buff(32) // StacksBlockHash is 32 bytes
```
MCJOHN974, line 396:
```suggestion
            .expect_buff(32) // BitcoinBlockHash is 32 bytes
```

MCJOHN974 (approved):

djordon-stacks:

PR #1730, opened 2025-06-12
feat: include bitcoin block hash in tenure-completed event

cylewitruk:
## Description

Modifies the tenure completion signal (`TxCoordinatorEvent::TenureComplete`) to contain an `Option<BitcoinBlockHash>` to allow for more specifically identifying which block a coordinator has completed its tenure for.

This is largely to enable stronger assertions in tests. Primarily in scenarios where a number of signers are potentially processing a number of blocks (i.e. from a `.generate_blocks(n)` call) and you wait to wait until processing for the last block has completed.

This PR targets PR https://github.com/stacks-sbtc/sbtc/pull/1726.

## Changes

- The above mentioned modification to `TxCoordinatorEvant`.
- Minor refactoring in the transaction coordinator to include the block hash for successfully completed tenures (including non-coordinator tenures).
- A little bit of re-ordering in the transaction coordinator validation logic and minor cleanup + comments (see [this comment](https://github.com/stacks-sbtc/sbtc/pull/1730#discussion_r2152789309)).
- Fixes broken `skip_smart_contract_deployment_and_key_rotation_if_up_to_date` test (which without these changes incorrectly passed).
- Adapted existing tests.

## Checklist:

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -300,7 +300,6 @@ where
     /// ## Returns
     ///
     /// * `Ok(Some(block_hash))` with the processed Bitcoin block hash on success,
-    /// * `Ok(None)` if not in the correct Stacks epoch (epoch 3 or later),
     /// * `Err(Error)` if any step fails with an unhandled error.
     #[tracing::instrument(skip_all, fields(
         public_key = %self.signer_public_key(),


signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4403,7 +4402,8 @@ async fn process_rejected_withdrawal(is_completed: bool, is_in_mempool: bool) {
         .expect("failed to signal");
 
     // Await for tenure completion
-    let tenure_completed_signal = TxCoordinatorEvent::TenureCompleted.into();
+    let tenure_completed_signal =
+        TxCoordinatorEvent::TenureCompleted(Some(bitcoin_chain_tip.block_hash)).into();
```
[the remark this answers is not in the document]
cylewitruk: `IntoOption` is removed.

signer/src/transaction_coordinator.rs
```diff
@@ -236,15 +239,18 @@ where
                         event
                     {
                         tracing::debug!("received signal; processing requests");
-                        if let Err(error) = self.process_new_blocks().await {
```
cylewitruk, line 239: A few tweaks here (mostly re-ordering):
- Attempt to get the chain tip straight away from the database, for a few reasons:
    - When the coordinator starts its processing, let's use the bitcoin chain tip as close to the bitcoin-block-observed notification as possible. If we take the chain tip after the "processing delay" and there's a super-flash-block:
        - There's a subtle risk that it'll pick something up that the block observer isn't finished with (though this should be mitigated now that the block observer writes bitcoin blocks in transactions).
        - It's possible that the "processing delay" is nullified anyway if the chain tip has progressed between the block observation and when the chain tip is fetched (there's no delay in the block observer saving chain tips to the database), so it could have waited for 15 or so seconds only to pick a seconds-old block anyway, which a number of signers potentially haven't received yet.
    - So we can return an `Ok(block_hash)` if we're not in epoch 3, in case any test wants to verify this.
- If we are in epoch 3, perform the coordinator check immediately instead of waiting for the processing delay to finish. Might as well exit quickly if we're not the coordinator.
- Added comments and did some general cleanup-refactoring.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -265,27 +268,49 @@ where
...
         .join_all()
-        .await;
+        .await
+        .into_iter()
+        .filter_map(Result::err)
+        .collect::<Vec<_>>();
+
+    if !error_messages.is_empty() {
+        return Err(format!(
+            "Not all signers completed their coordinator tenure for bitcoin block '{bitcoin_block_hash:?}' within the expected time of {wait_duration:?}. Errors: {error_messages:?}"
```
djordon, line 305: nit: Can we make this a multi-line string.

signer/src/transaction_coordinator.rs
```diff
@@ -285,86 +290,106 @@ where
...
+    ///
+    /// This function checks if the node is eligible to process new blocks and, if so,
+    /// performs all necessary coordinator duties for the current Bitcoin chain tip.
+    /// These include DKG coordination, smart contract deployment, key rotation,
+    /// and constructing and signing Bitcoin and Stacks transactions as needed.
+    ///
+    /// ## Returns
+    ///
+    /// * `Ok(Some(block_hash))` with the processed Bitcoin block hash on success,
+    /// * `Ok(None)` if not in the correct Stacks epoch (epoch 3 or later),
```
djordon: Why support the distinction between epochs? I would think that we would always return the block hash that was "processed". Also, the current code should be able to always return a block hash, since we query the database first thing.

signer/src/transaction_coordinator.rs
```diff
@@ -285,86 +290,106 @@ where
...
-        let bitcoin_processing_delay = self.context.config().signer.bitcoin_processing_delay;
-        if bitcoin_processing_delay > Duration::ZERO {
-            tracing::debug!("sleeping before processing new bitcoin block");
-            tokio::time::sleep(bitcoin_processing_delay).await;
-        }
-
-        let bitcoin_chain_tip = self
+    pub async fn process_new_blocks(&mut self) -> Result<Option<BitcoinBlockHash>, Error> {
+        // Attempt to get the current Bitcoin canonical chain tip from the database.
+        let maybe_bitcoin_chain_tip = self
```
djordon, line 311: I’m going to think out loud here about the reordering. The proposed change is a solution to the following scenario

1. A signer observes bitcoin chain tip with block hash A.
2. The signer starts doing coordinator things in `process_new_block`, and they start sleeping for X seconds.
3. After T seconds, other parts of the signer observes a new bitcoin chain tip with block hash B.
4. Other signers observe this bitcoin block at different times, some before and some after the current signer.
5. The coordinator for block hash B only slept for X - T seconds before starting to do coordinator things, and perhaps the other signers haven’t observed bitcoin block with block hash B yet.

If we fetch the chain tip at the start of the event loop, the signer who is coordinator for tenure A (tenure during block hash A) will believe that they are coordinator during some parts of tenure B. They will get ignored and eventually stop coordinator activities when they realize that their tenure has ended. The signer who is coordinator for tenure B will wait for the appropriate amount of time. If we fetch the chain tip after the sleep, the signer who is coordinator for tenure A will correctly realize that their tenure has ended. While the signer who is coordinator for tenure B will not wait the appropriate amount of time before beginning their duties.

djordon (commented):
Nice, I think you’ve identified a potential issue with how we currently do things. The issue is that flash blocks can cause the coordinator to not wait the correct amount of time before starting their duties. The other event loops might not finish doing their thing either before the coordinator does its thing too. Querying for the bitcoin block before the pause fixes that issue in a subtle way, at the cost of slightly noisier logs during flash blocks.

I think that I’d like to change this a bit. Here is what I’m thinking

1. Have `TxCoordinatorEventLoop::process_new_block` take a `BitcoinBlockHash` or `BitcoinBlockRef` as an input. This change requires us to make the `BlockObserved` and the `NewRequestsHandled` events take a parameter. But I'm not sure if this change is worth it.
2. Have `process_new_block` check that the block hash that is given as an input matches the one in the state and bail if they differ. Doing it at the start and after sleeping makes sense.
3. Keep the return type of `process_new_block` the a `Result<(), Error>` but `TenureCompleted` would send a block hash/ref, not an option. Alternatively, change the return type of `process_new_block` to a `Result<BitcoinBlockRef, Error>` and still have `TenureCompleted` send a block hash/ref, not an option.

I think these changes would make everything more sound and easier to reason about. The design would be that the `process_new_block` function is called once per bitcoin block, and it returns “as soon as possible” whenever a new bitcoin block has been observed. The current code kinda sorta does that, but it has the drawback that it doesn’t reliably wait for the necessary amount of time for each bitcoin block.

signer/src/transaction_coordinator.rs
cylewitruk: Yeah this comment was from before I moved the fetching of block hash before the epoch check, updated in [56bc08b](https://github.com/stacks-sbtc/sbtc/pull/1730/commits/56bc08b183bd15403860c43410cb9eb39fb179c8). The option was needed because of the ordering of operations, e.g. some return points (epoch3 check) were pre-wait so the block being processed wasn't known.

PR #1734, opened 2025-06-13
feat: use stacks core for signer set data

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1733

## Changes

* Replace the `StacksInteract::get_current_signer_set` method with a `StacksInteract::get_current_signer_set_info` method, that returns the signer set, the aggregate key, and the signature threshold from the registry.
* Use the new `StacksInteract::get_current_signer_set_info` function in the block observer to update the state.

## Testing Information

I repurposed one of our existing tests after this change and spun up devenv to make sure that everything still worked as expected.

## Checklist:


MCJOHN974 (approved):

signer/src/stacks/api.rs
```diff
@@ -135,17 +137,36 @@ pub enum FeePriority {
...
+pub struct SignerSetInfo {
+    /// The aggregate key of the most recently confirmed key rotation
+    /// contract call on Stacks.
+    pub aggregate_key: PublicKey,
+    /// The set of sBTC signers public keys.
+    pub signer_set: BTreeSet<PublicKey>,
+    /// The number of signatures required to sign a transaction.
+    /// This is the number of signature shares necessary to successfully sign a
+    /// bitcoin transaction spending a UTXO locked with the above aggregate key,
+    /// or the number of signers necessary to sign a Stacks transaction under
```
MCJOHN974:
```suggestion
    /// and the number of signers necessary to sign a Stacks transaction under
```

djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1734/commits/da6a92bd97ff79d61611143ea4cd96e70393cf87

MCJOHN974 (dismissed):

signer/src/testing/context.rs
```diff
@@ -422,15 +422,15 @@ impl BitcoinInteract for WrappedMockBitcoinInteract {
     }
 }
 
-impl StacksInteract for WrappedMockStacksInteract {
-    async fn get_current_signer_set(
+impl StacksInteract for WrappedMock<MockStacksInteract> {
```
matteojug: nit: we lost the alias

signer/src/transaction_coordinator.rs
```diff
@@ -432,9 +432,9 @@ where
 
         let current_aggregate_key = self
             .context
-            .get_stacks_client()
-            .get_current_signers_aggregate_key(&self.context.config().signer.deployer)
-            .await?;
+            .state()
+            .registry_signer_set_info()
+            .map(|info| info.aggregate_key);
```
matteojug, line 446: Should we fetch it again, to be sure we have the most up to date info in here? eg, if we have a rotate key tx in the mempool but it gets confirmed only after the block observer refreshed the state, here we may try to submit another key rotation (signed by the "old" signer set). We are not solving the issue since the same could occur between here and when we actually submit the tx, it's just to reduce that time window.

signer/src/stacks/api.rs
```diff
@@ -1275,45 +1296,113 @@ where
...
+/// Extract a signature threshold from a Clarity value.
+///
+/// In the sbtc-registry smart contract, the signature threshold is stored
+/// in the `current-signature-threshold` data var and is initialized to 0,
+/// allowing use to distinguish between the initial value and an actual
+/// signature threshold. Ok(None) is returned if the value is the initial
+/// value.
+fn extract_signatures_required(value: Value) -> Result<Option<u16>, Error> {
+    match value {
+        Value::UInt(0) => Ok(None),
```
MCJOHN974, line 1363: Do we really want represent 0 signatures required as `None` and do not just use 0 ?

djordon: Yeah I think we want it to be `None`. My reasoning is that the signer's behavior depends on whether the registry has signer set data or not. Since the signers cannot set the `signatures-required` to zero, so seeing zero here means that it's unset and that the registry does not have signer set data. Therefore, returning `None` is better than returning `0`. Does that make sense?

MCJOHN974 (dismissed):

signer/src/stacks/api.rs
```diff
@@ -1275,45 +1296,113 @@ where
...
+/// In the sbtc-registry smart contract, the aggregate key is stored in the
+/// `current-aggregate-pubkey` data var and is initialized to the 0x00
+/// byte, allowing use to distinguish between the initial value and an
+/// actual public key in that case. Ok(None) is returned if the value is
+/// the initial value.
+fn extract_public_key(value: Value) -> Result<Option<PublicKey>, Error> {
+    match value {
+        Value::Sequence(SequenceData::Buffer(BuffData { data })) => {
+            // The initial value of the data var is all zeros
+            if data.iter().all(|v| *v == 0) {
```
matteojug: before the first rotation, isn't this a single `0x00` byte? With this check we would consider "bootstrap" also other values (sequences of `0x00` bytes), which is probably not an issue since we are not going to use them, but OTOH I don't see why "risking" having issues related to this difference in behaviour.

signer/src/stacks/api.rs
```diff
@@ -1275,45 +1296,113 @@ where
...
+///
+/// The value is expected to be a sequence of 33 byte buffers, each of
+/// which represents a compressed public key.
+fn extract_signer_set(value: Value) -> Result<BTreeSet<PublicKey>, Error> {
+    match value {
+        // Iterate through each record in the list and convert it to a
+        // public key. If the record is not a buffer, then return an error.
+        Value::Sequence(SequenceData::List(ListData { data, .. })) => data
+            .into_iter()
+            .flat_map(|item| extract_public_key(item).transpose())
```
matteojug: while it's probably fine, we shouldn't skip anything in here, while `extract_public_key` will skip over `0x00` as key. Similar to the other comment, probably fine but since this is kinda important we should probably avoid having different behaviours wrt the smart contract.

signer/src/stacks/api.rs
```diff
@@ -1275,45 +1296,113 @@ where
...
+/// signature threshold. Ok(None) is returned if the value is the initial
+/// value.
+fn extract_signatures_required(value: Value) -> Result<Option<u16>, Error> {
+    match value {
+        Value::UInt(0) => Ok(None),
+        Value::UInt(threshold) => Ok(Some(
+            threshold.try_into().map_err(|_| Error::TypeConversion)?,
+        )),
+        _ => Err(Error::InvalidStacksResponse(
+            "expected a buffer but got something else",
```
matteojug: nit: 
```suggestion
            "expected a uint but got something else",
```

signer/src/stacks/api.rs
djordon: Yeah, these changes are just meant to reorganize the logic that was already here. I don't mind updating it to check for the sentinal value of a buffer with one `0x00` byte.

signer/src/testing/context.rs
djordon: Oh interesting. I don't recall doing that explicitly, maybe my IDE did it for me?

signer/src/stacks/api.rs
djordon: Good point, will update

signer/src/stacks/api.rs
djordon: Corrected in https://github.com/stacks-sbtc/sbtc/pull/1734/commits/0d6b98c5dbb43624a88a327895093993b5347088

signer/src/stacks/api.rs
djordon: Changed in https://github.com/stacks-sbtc/sbtc/pull/1734/commits/f62e85ae0b6b9ffbdb7741796d966f086fdff371.

signer/src/stacks/api.rs
djordon: Yeah this is better. Fixed in https://github.com/stacks-sbtc/sbtc/pull/1734/commits/4f60d3ebd11641182e585e125343a30cf6701655

signer/src/testing/context.rs
djordon: Changed back in https://github.com/stacks-sbtc/sbtc/pull/1734/commits/d4be0e74e3035d93fc515be4371875ada189abed

signer/src/transaction_coordinator.rs
djordon: > Should we fetch it again, to be sure we have the most up to date info in here?

Hmmm, I think so, but I'm not sure. My first though was "yeah, that makes sense, that seems more correct", but after thinking about it some more something seems off.

The thing that came to mind is that other parts of the system use different sources of information for the same kind of "check". Basically, we should also reach out to the stacks node in the tx-signer, as well as in other places of the tx-coordinator, when we need the signer set info. These aren't bad changes either, that might be what we should do, I'm just highlighting the inconsistency that made me feel like something is off.

I'm not sure what originally made me make this change, so I think we can keep the original behavior, it just feels a little off. And it's not a problem either way, we'll be fine with any of these implementations so I'm cool with whatever. Anyway, thoughts?

matteojug (approved):

signer/src/transaction_coordinator.rs
matteojug: I suggested this in a similar spirit to how we check if a request was completed according to the SC, regardless of the events in the db, but yeah it does seem a bit off. I guess we can roll with this for the moment, keeping the original behaviour, and then reevaluate if/where we could do more "updated/fresh" checks. In the end it's mostly an optimization to avoid wasting fees, it's not a safety issue.

cylewitruk: I'd be pro a "check and bail if mismatch" just to save a signing round and potential failed tx submission, but this is probably an edge case (and key rotations are rare in comparison), so I don't have any strong opinions on the strategy here given the current constraints.

djordon: 👍🏽 yeah. We'd need the other signers to also get it wrong when they do their checks, which is very possible since they all get the signer set info from their state object.

Also, I'm less worried about fees, since the key rotation must be one where the stacks wallet remains the same, which hopefully is rare.

One thing that I think is worthwhile is adding metrics around how long this call takes.

matteojug: Resolving since the original issue seems like something we may tackle in a different PR

PR #1735, opened 2025-06-13
fix: don't decrement nonce on mempool conflict

matteojug:
## Description

Closes: https://github.com/stacks-sbtc/sbtc/issues/1711

## Changes

When submitting a stacks tx, if the coordinator gets a `ConflictingNonceInMempool` error from the stacks node, it will not decrement the nonce to avoid failing also the following transactions.

## Testing Information

Add a devenv-based integration test.

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -52,8 +52,10 @@ use crate::network;
 use crate::signature::TaprootSignature;
 use crate::stacks::api::FeePriority;
 use crate::stacks::api::GetNakamotoStartHeight;
+use crate::stacks::api::RejectionReason;
 use crate::stacks::api::StacksInteract;
 use crate::stacks::api::SubmitTxResponse;
+use crate::stacks::api::TxRejection;
 use crate::stacks::contracts::AcceptWithdrawalV1;
 use crate::stacks::contracts::AsTxPayload;
 use crate::stacks::contracts::CompleteDepositV1;
@@ -793,8 +795,8 @@ where
                 }
             };
 
-            // If we fail to sign the transaction for some reason, we
-            // decrement the nonce by one, and try the next transaction.
+            // If we fail to sign the transaction for some reason, we adjust the
+            // nonce and try the next transaction.
             // This is not a fatal error, since we could fail to sign the
             // transaction because someone else is now the coordinator, and
             // all the signers are now ignoring us.
@@ -808,7 +810,7 @@ where
                 }
                 Err(error) => {
                     tracing::warn!(%error, %outpoint, "could not process the stacks sign request for a deposit");
-                    wallet.set_nonce(wallet.get_nonce().saturating_sub(1));
+                    adjust_nonce(wallet, &error);
                     "failure"
                 }
             };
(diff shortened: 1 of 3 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -2634,6 +2636,19 @@ pub fn assert_rotate_key_action(
...
+        // For `ConflictingNonceInMempool` we don't want to decrement the nonce
+        // to avoid failing also the following submissions
+        Error::StacksTxRejection(TxRejection {
+            reason: RejectionReason::ConflictingNonceInMempool,
+            ..
+        }) => (),
+        _ => wallet.set_nonce(wallet.get_nonce().saturating_sub(1)),
+    }
+}
+
```
MCJOHN974, line 2616: Maybe return from this function `Result<(), Error>` and return `Error::StacksTxRejection` to highlight that we actually didn't change anything?

matteojug: mmm it seems a bit strange to return an error from this, since we are doing some logic on the specific `StacksTxRejection` `ConflictingNonceInMempool` error. Like, the caller already received and handled the error, we are operating on that error and both outcomes (decrement or not) are "happy" paths for this fn. Returning an error would signal that something went wrong and we may need to act on it, while it's not really the case imho.

signer/tests/integration/contracts.rs
```diff
@@ -48,16 +48,16 @@ fn make_signatures(tx: &StacksTransaction, keys: &[Keypair]) -> Vec<RecoverableS
         .collect()
 }
 
-pub struct SignerStxState {
+pub struct SignerStxState<S: StacksInteract> {
```
djordon: We should remove the trait bound in the struct definition.

signer/src/transaction_coordinator.rs
MCJOHN974: Also not related to this pr but it is sad a bit that we mutate wallet by immutable reference =(

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -5014,3 +5030,465 @@ async fn should_handle_dkg_coordination_failure() {
...
+    // =========================================================================
+    // Wait for DKG + key rotation
+    // =========================================================================
+    faucet.generate_block();
+    wait_for_signers(&signers).await;
+
+    let shares = db.get_latest_encrypted_dkg_shares().await.unwrap().unwrap();
+    assert_eq!(shares.dkg_shares_status, DkgSharesStatus::Verified);
+
+    wait_next_stacks_block(&stacks).await;
```
djordon, line 5399: This will be tricky some time in the future, when we run this in CI, since we'll probably turn off automatic stacks block production.

djordon (dismissed):
Looks good

MCJOHN974 (approved):

signer/tests/integration/contracts.rs
matteojug: Done in [dc011d2](https://github.com/stacks-sbtc/sbtc/pull/1735/commits/dc011d28092021ded7222550389b442c622f9e4e)

PR #1743, opened 2025-06-17
start changes

MCJOHN974:
## Description

Closes: #?

## Checklist:

### DIFF
--- a/emily/handler/src/api/handlers/deposit.rs
+++ b/emily/handler/src/api/handlers/deposit.rs
@@ -509,10 +509,12 @@ async fn update_deposits(
     // Loop through all updates and execute.
     for (index, update) in validated_request.deposits {
         if update.is_err() {
+            // We just checked that update is err, so this unwrap is fine.
+            let deposit = Err(update.unwrap_err().to_string());
             updated_deposits.push((
                 index,
                 DepositWithStatus {
-                    deposit: Deposit::default(),
+                    deposit,
                     status: StatusCode::BAD_REQUEST.as_u16(),
                 },
             ));
@@ -547,7 +549,7 @@ async fn update_deposits(
                 updated_deposits.push((
                     index,
                     DepositWithStatus {
-                        deposit: Deposit::default(),
+                        deposit: Err(Error::NotFound.to_string()),
                         status: StatusCode::NOT_FOUND.as_u16(),
                     },
                 ));
@@ -562,7 +564,7 @@ async fn update_deposits(
                 updated_deposits.push((
                     index,
                     DepositWithStatus {
-                        deposit: Deposit::default(),
+                        deposit: Err(Error::Forbidden.to_string()),
                         status: StatusCode::FORBIDDEN.as_u16(),
                     },
                 ));
@@ -578,7 +580,7 @@ async fn update_deposits(
                 updated_deposits.push((
                     index,
                     DepositWithStatus {
-                        deposit: Deposit::default(),
+                        deposit: Err(error.to_string()),
                         status: StatusCode::INTERNAL_SERVER_ERROR.as_u16(),
                     },
                 ));
@@ -599,7 +601,7 @@ async fn update_deposits(
         updated_deposits.push((
             index,
             DepositWithStatus {
-                deposit,
+                deposit: Ok(deposit),
                 status: StatusCode::OK.as_u16(),
             },
         ));
--- a/emily/handler/src/api/handlers/withdrawal.rs
+++ b/emily/handler/src/api/handlers/withdrawal.rs
@@ -405,10 +405,12 @@ async fn update_withdrawals(
     // Loop through all updates and execute.
     for (index, update) in validated_request.withdrawals {
         if update.is_err() {
+            // We just checked that update is err so unwrap here is fine.
+            let withdrawal = update.unwrap_err().to_string();
             updated_withdrawals.push((
                 index,
                 WithdrawalWithStatus {
-                    withdrawal: Withdrawal::default(),
+                    withdrawal,
                     status: StatusCode::BAD_REQUEST.as_u16(),
                 },
             ));
@@ -435,7 +437,7 @@ async fn update_withdrawals(
                 updated_withdrawals.push((
                     index,
                     WithdrawalWithStatus {
-                        withdrawal: Withdrawal::default(),
+                        withdrawal: Err(Error::NotFound.to_string()),
                         status: StatusCode::NOT_FOUND.as_u16(),
                     },
                 ));
@@ -449,7 +451,7 @@ async fn update_withdrawals(
                 updated_withdrawals.push((
                     index,
                     WithdrawalWithStatus {
-                        withdrawal: Withdrawal::default(),
+                        withdrawal: Err(Error::Forbidden.to_string()),
                         status: StatusCode::FORBIDDEN.as_u16(),
                     },
                 ));
@@ -464,7 +466,7 @@ async fn update_withdrawals(
                 updated_withdrawals.push((
                     index,
                     WithdrawalWithStatus {
-                        withdrawal: Withdrawal::default(),
+                        withdrawal: Err(error.to_string()),
                         status: StatusCode::INTERNAL_SERVER_ERROR.as_u16(),
                     },
                 ));
@@ -484,7 +486,7 @@ async fn update_withdrawals(
         updated_withdrawals.push((
             index,
             WithdrawalWithStatus {
-                withdrawal,
+                withdrawal: Ok(withdrawal),
                 status: StatusCode::OK.as_u16(),
             },
         ));
--- a/emily/handler/src/api/models/deposit/responses.rs
+++ b/emily/handler/src/api/models/deposit/responses.rs
@@ -34,11 +34,13 @@ pub struct UpdateDepositsResponse {
 }
 
 /// Wrapper for deposit with status code. Used for multi-status responses.
-#[derive(Clone, Default, Debug, PartialEq, Hash, Serialize, Deserialize, ToSchema, ToResponse)]
+#[derive(Clone, Debug, PartialEq, Hash, Serialize, Deserialize, ToSchema, ToResponse)]
 #[serde(rename_all = "camelCase")]
 pub struct DepositWithStatus {
-    /// The fully extracted and validated deposit request.
-    pub deposit: Deposit,
+    /// The fully extracted and validated deposit request, or
+    /// an error message.
+    pub deposit: Result<Deposit, String>,
     /// HTTP status code, returned as part of multi-status responses.
     pub status: u16,
 }
+
--- a/emily/handler/src/api/models/withdrawal/responses.rs
+++ b/emily/handler/src/api/models/withdrawal/responses.rs
@@ -24,11 +24,11 @@ pub struct UpdateWithdrawalsResponse {
 }
 
 /// Wrapper for withdrawal with status code. Used for multi-status responses.
-#[derive(Clone, Default, Debug, PartialEq, Hash, Serialize, Deserialize, ToSchema, ToResponse)]
+#[derive(Clone, Debug, PartialEq, Hash, Serialize, Deserialize, ToSchema, ToResponse)]
 #[serde(rename_all = "camelCase")]
 pub struct WithdrawalWithStatus {
     /// The fully extracted and validated withdrawal.
-    pub withdrawal: Withdrawal,
+    pub withdrawal: Result<Withdrawal, String>,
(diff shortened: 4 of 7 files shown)


.generated-sources/emily/client/rust/public/src/models/deposit_with_status_schemed.rs
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * The version of the OpenAPI document: 0.1.0
+ *
+ * Generated by: https://openapi-generator.tech
+ */
+
+use crate::models;
+use serde::{Deserialize, Serialize};
+
+/// DepositWithStatusSchemed : Workaround to make utopia generate openapi. Used only as last step before sending UpdateDepositsResponce
```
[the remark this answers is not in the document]
MCJOHN974: You pointed to autogenerated files but actually it was caused by typo in source code, thanks! [31f52bc](https://github.com/stacks-sbtc/sbtc/pull/1743/commits/31f52bc99970a8aadb0ce48e3c0517611b497b42)

.generated-sources/emily/client/rust/private/src/models/withdrawal_with_status_schemed.rs
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * The version of the OpenAPI document: 0.1.0
+ *
+ * Generated by: https://openapi-generator.tech
+ */
+
+use crate::models;
+use serde::{Deserialize, Serialize};
+
+/// WithdrawalWithStatusSchemed : Workaround to make utopia generate openapi. Used only as last step before sending UpdateWithdrawalsResponce
```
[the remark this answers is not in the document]
MCJOHN974: Same as above

emily/handler/src/api/models/deposit/responses.rs
```diff
@@ -30,15 +30,47 @@ pub struct GetDepositsResponse {
...
+
+/// Workaround to make utopia generate openapi. Used only as last step before sending
+/// UpdateDepositsResponce
+#[derive(Clone, Default, Debug, PartialEq, Hash, Serialize, Deserialize, ToSchema, ToResponse)]
+#[serde(rename_all = "camelCase")]
+pub struct DepositWithStatusSchemed {
+    deposit: Option<Deposit>,
+    error: Option<String>,
+    status: u16,
+}
```
djordon: Hmmm, this type seems like the one that we should use. Although it is weaker, in the sense that the `Result<_, _>` conveys the information that we want, this the `DepositWithStatus` struct is used just as a response, so I'd document the expected condition that exactly one of the `deposit` field or the `error` field will be `None`.

So, I think this struct should be the actual `DepositWithStatus`. I'd remove the existing `DepositWithStatus` and rename this one as `DepositWithStatus`. And we'll need to update the existing comments and add a comment to the new error field.

djordon: Oh and same thing for withdrawals.

MCJOHN974: [02b8ff5](https://github.com/stacks-sbtc/sbtc/pull/1743/commits/02b8ff58828143761ce4db809fa75497193875b6)

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -509,10 +509,13 @@ async fn update_deposits(
     // Loop through all updates and execute.
     for (index, update) in validated_request.deposits {
         if update.is_err() {
+            // We just checked that update is err, so this unwrap is fine.
+            let error = update.unwrap_err().to_string();
```
djordon: I would change this to be something like
```suggestion
        if let Err(error) = update {
```
These lines could also benefit from some more `match` statements, but no need to make that change.

emily/handler/src/api/models/deposit/responses.rs
```diff
@@ -34,11 +34,17 @@ pub struct UpdateDepositsResponse {
...
+/// None, and exactly one of them should be Some. It would be nice to find a way to use
+/// `Result` here.
 #[derive(Clone, Default, Debug, PartialEq, Hash, Serialize, Deserialize, ToSchema, ToResponse)]
 #[serde(rename_all = "camelCase")]
 pub struct DepositWithStatus {
     /// The fully extracted and validated deposit request.
-    pub deposit: Deposit,
-    /// HTTP status code, returned as part of multi-status responses.
+    pub deposit: Option<Deposit>,
+    /// String explaining error occured during updating the deposit.
```
djordon:
```suggestion
    /// A string explaining the error that occurred during the deposit update.
```

emily/handler/src/api/models/withdrawal/responses.rs
```diff
@@ -24,11 +24,17 @@ pub struct UpdateWithdrawalsResponse {
 }
 
 /// Wrapper for withdrawal with status code. Used for multi-status responses.
+/// TODO: utopia, which we use for generating OpenAPI spec does not support
+/// [`Result`] in structs, however, logically exactly one of `withdrawal` or `error` should be
+/// None, and exactly one of them should be Some. It would be nice to find a way to use
+/// `Result` here.
```
djordon: I think the returned JSON cannot really support returning a result, since it would need to be an object, and then you arrive at something similar to what we have here. So I would just note that, logically, exactly one of `withdrawal` or `error` will be `None` and I'd remove the TODO since there is nothing TODO.

If we want to enforce this only-one-will-be-none invariant in code, which I think we want to do, we can make it so that the only way to construct this type would be through a `WithdrawalWithStatus::from_result` function that takes a `Result` and maybe a status code, and properly creates the struct. This way the struct enforces its invariants, and we cannot mess it up. Same thing for the `DepositWithStatus` struct.

MCJOHN974: Well, on json level it really sounds like impossible to have a result. But maybe it can be done somehow on rust level in such a way that during json parsing an error will be raised if this json can not be parsed into valid `Result`

emily/handler/src/api/handlers/deposit.rs
MCJOHN974: [58310c7](https://github.com/stacks-sbtc/sbtc/pull/1743/commits/58310c7d823f1fea102c96120a0fbde811e8b772)

emily/handler/src/api/models/withdrawal/responses.rs
djordon: Yeah, it's not clear whether such a thing is possible. If you think it is, then we might as well document it in a ticket and write the ticket number next to the TODO. Many of our TODOs in the signer crate, though not all, follow the `TODO(ticket-number)` pattern. But if you suspect that it's impossible or just not worth it then this isn't a TODO so I'd modify the comment accordingly.

MCJOHN974: [f5487cf](https://github.com/stacks-sbtc/sbtc/pull/1743/commits/f5487cff8dfbbacdd95222e17605c06dc99ffc0f)

djordon: Same thing for deposits.

MCJOHN974: [0ba63ef](https://github.com/stacks-sbtc/sbtc/pull/1743/commits/0ba63ef59ca21e2f84d912823c3166b4a836e033)

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -508,11 +508,12 @@ async fn update_deposits(
...
     // Loop through all updates and execute.
     for (index, update) in validated_request.deposits {
-        if update.is_err() {
+        if let Err(error) = update {
             updated_deposits.push((
                 index,
                 DepositWithStatus {
-                    deposit: Deposit::default(),
+                    deposit: None,
+                    error: Some(error.to_string()),
```
matteojug, line 518: Should we add some comments in `ValidationError` to make it clear that those errors may be returned to the client, so we shouldn't have any error leaking "sensitive" info in there? Can be done in a different PR

emily/handler/src/api/handlers/withdrawal.rs
```diff
@@ -405,10 +405,13 @@ async fn update_withdrawals(
     // Loop through all updates and execute.
     for (index, update) in validated_request.withdrawals {
         if update.is_err() {
+            // We just checked that update is err so unwrap here is fine.
+            let error = update.unwrap_err().to_string();
```
matteojug: nit: `if let Err` as in deposit

emily/handler/src/api/handlers/deposit.rs
```diff
@@ -578,7 +581,8 @@ async fn update_deposits(
                 updated_deposits.push((
                     index,
                     DepositWithStatus {
-                        deposit: Deposit::default(),
+                        deposit: None,
+                        error: Some(error.to_string()),
```
matteojug: This may leak sensitive info (based on the error), we should use `.into_production_error` before stringifying it. Same in withdrawals

emily/handler/tests/integration/deposit.rs
```diff
@@ -1628,13 +1651,23 @@ async fn emily_process_deposit_updates_when_some_of_them_are_unknown() {
...
 
     // Check that multistatus response is returned correctly.
-    assert!(update_responce.deposits.iter().all(|deposit| {
-        if deposit.deposit.bitcoin_txid == create_deposit_body1.bitcoin_txid {
-            deposit.status == 200
-        } else {
-            deposit.status == 404
-        }
-    }));
+    update_responce
```
matteojug: nit: also in withdrawal
```suggestion
    update_response
```

emily/handler/tests/integration/deposit.rs
```diff
@@ -1628,13 +1651,23 @@ async fn emily_process_deposit_updates_when_some_of_them_are_unknown() {
...
+        .deposits
+        .iter()
+        .for_each(|deposit| match &deposit.deposit {
+            Some(Some(inner)) => {
+                assert_eq!(inner.bitcoin_txid, create_deposit_body1.bitcoin_txid);
+                assert_eq!(deposit.status, 200);
```
matteojug: nit: for completeness, can we also check that `error` is none? also in withdrawal

emily/handler/src/api/handlers/withdrawal.rs
```diff
@@ -404,11 +404,12 @@ async fn update_withdrawals(
...
     // Loop through all updates and execute.
     for (index, update) in validated_request.withdrawals {
-        if update.is_err() {
+        if let Err(error) = update {
             updated_withdrawals.push((
                 index,
                 WithdrawalWithStatus {
-                    withdrawal: Withdrawal::default(),
+                    withdrawal: None,
+                    error: Some(error.to_string()),
```
djordon, line 414: Don't we need `.into_production_error()` here too?

djordon: Oh no we don't. Maybe add another comment about how this is a `ValidationError`? Actually, it's fine as is.

emily/handler/tests/integration/withdrawal.rs
```diff
@@ -1006,21 +1012,32 @@ async fn emily_process_withdrawal_updates_when_some_of_them_are_unknown() {
...
     // Check that multistatus response is returned correctly.
-    assert!(update_responce.withdrawals.iter().all(|withdrawal| {
-        if withdrawal.withdrawal.request_id == create_withdrawal_body1.request_id {
-            withdrawal.status == 200
-        } else {
-            withdrawal.status == 404
-        }
-    }));
+    update_response
+        .withdrawals
```
djordon: We probably want to check that `update_response.withdrawals` is nonempty too.

djordon (dismissed):
Looks good, had one nitty comment that you can feel free to ignore.

emily/handler/src/api/handlers/withdrawal.rs
matteojug: Added a comment in [a911bf7](https://github.com/stacks-sbtc/sbtc/pull/1743/commits/a911bf789f1c5db3fdd64c03895433983b512eae), it doesn't hurt.

emily/handler/tests/integration/withdrawal.rs
matteojug: Yeah, changed the check in favour of a more explicit one on the returned values in [a911bf7](https://github.com/stacks-sbtc/sbtc/pull/1743/commits/a911bf789f1c5db3fdd64c03895433983b512eae)

djordon (approved):
✅

matteojug (approved):

Issue #1693, opened 2025-05-28
[Chore]: Split Emily Status enum for deposits and withdrawals


MCJOHN974:
# Chore - Split Emily Status enum for deposits and withdrawals

## 1. Description

Currently Emily have `Status` enum which represents status of deposit and withdrawal events. However, this events behave slightly differently, and now we have `RBF` enum variant which is applicable only for deposits. Using same structure for both deposits and withdrawals leads us to hacky workarounds -- `WithdrawalRBF` error type, etc. I think in future there can be even more difference so we better create two separate enums for deposits and withdrawals.

Thread which created this issue: https://github.com/stacks-sbtc/sbtc/pull/1686#discussion_r2111629899

## 2. Why It's Necessary

Would make our code more clear and remove hacky solutions dictated by having single struct.

## 3. Execution Checklist

- [ ] `Status` enum eliminated
- [ ] `DepositStatus` and `WithdrawalStatus` enums introduced.

PR #1745, opened 2025-06-17
chore: Split deposit and withdrawal status on Emily

MCJOHN974:
## Description

Closes: #1693

## Changes

Now Emily have two structs `DepositStatus` and `WithdrawalStatus` instead of old `Status` (same for `StatusEntry`). As for now they differ because deposits can have `RBF` status while withdrawals can't, and this split protect us from invalid state (withdrawal with `RBF` status) on compiler level.

## Testing Information

Refactoring. No new testing needed

## Checklist:

### DIFF
--- a/emily/handler/src/api/models/common/mod.rs
+++ b/emily/handler/src/api/models/common/mod.rs
@@ -8,7 +8,7 @@ pub mod requests;
 
 // Common Types ----------------------------------------------------------------
 
-/// The status of the in-flight sBTC operation.
+/// The status of the in-flight sBTC deposit.
 #[derive(
     Clone,
     Default,
@@ -24,7 +24,7 @@ pub mod requests;
     ToResponse,
 )]
 #[serde(rename_all = "lowercase")]
-pub enum Status {
+pub enum DepositStatus {
     /// Transaction hasn't yet been addressed by the sBTC Signers.
     #[default]
     Pending,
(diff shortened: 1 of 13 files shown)


emily/handler/src/api/models/common/mod.rs
```diff
@@ -52,6 +52,48 @@ pub enum Status {
...
+    ToResponse,
+)]
+#[serde(rename_all = "lowercase")]
+pub enum WithdrawalStatus {
+    /// Transaction hasn't yet been addressed by the sBTC Signers.
+    #[default]
+    Pending,
+    /// Transaction was dealt with by the signers at one point but is now being
+    /// reprocessed. The Signers are aware of the operation request.
+    Reprocessing,
```
djordon, line 80: I suppose we could preemptively remove this here too. Or not, do whichever is easier.

emily/handler/src/database/accessors.rs
```diff
@@ -119,14 +122,22 @@ pub async fn get_deposit_entries_by_reclaim_pubkeys_hash(
...
+    DepositStatus::Accepted,
+    DepositStatus::Confirmed,
+    DepositStatus::Failed,
+    DepositStatus::Pending,
+    DepositStatus::Reprocessing,
+    DepositStatus::RBF,
+];
+/// Hacky exhaustive list of all possible withdrawal statuses that we will iterate over in order to
+/// get every withdrawal present.
+const ALL_WITHDRAWAL_STATUSES: &[WithdrawalStatus] = &[
```
djordon: If we wanted, we could also use `strum::EnumIter` here on the type, so that an iterator gets automatically generated for us. The main benefit of doing it this way is that if we add or remove a variant, we cannot forget to update these arrays.

emily/handler/src/api/models/common/mod.rs
MCJOHN974: I already have another PR for doing this and you approved it already, so I think we can close this thread for now

emily/handler/src/database/accessors.rs
MCJOHN974: [86dcd0b](https://github.com/stacks-sbtc/sbtc/pull/1745/commits/86dcd0bfe2073d0b27db5f9e1de7536172701868) and [94321f5](https://github.com/stacks-sbtc/sbtc/pull/1745/commits/94321f5899dea99b48a1a930d8af1030d7565bc1) add strum usage

PR #1746, opened 2025-06-17
feat: generate recoverable private shares during dkg

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1303

## Changes

* Generate secret shares using a random number generator that was seeded with the signer's private key and the bitcoin block hash and height at the time that the DKG started. This makes the secret shares deterministically generated but still pseudo-random to outside observers.
* Made a small refactor since the random number generator used when processing messages is managed by the wrapping state machine.

## Testing Information

I added an integration test that checks that the DKG generates the same secret shares when given the same inputs, but changes with a new bitcoin block.

I'm going to add a test that makes sure that the nonces generated during signing are still random. But this is good for review now.

## Checklist:


signer/src/wsts_state_machine.rs
```diff
@@ -465,41 +484,109 @@ impl SignerStateMachine {
...
+    /// block reference and a private key.
+    fn create_rng(started_at: &BitcoinBlockRef, private_key: PrivateKey) -> StdRng {
+        let seed_bytes: [u8; 32] = sha2::Sha256::new_with_prefix("DKG_RNG")
+            .chain_update(started_at.block_hash.into_bytes())
+            .chain_update(started_at.block_height.to_le_bytes())
+            .chain_update(private_key.to_bytes())
+            .finalize()
+            .into();
+
+        StdRng::from_seed(seed_bytes)
```
cylewitruk: Should we opt to use an explicit RNG here (e.g. from `rand_chacha`) instead of the opaque `StdRng`? IIUC the underlying algorithm for `StdRng` isn't guaranteed between versions (as a general-purpose RNG). I don't know if its `SeedableRng` implementation offers other guarantees, though.

djordon: Good call. Will do.

Cargo.toml
```diff
@@ -57,6 +57,7 @@ p256k1 = { version = "7.2.2", default-features = false }
 proptest = { version = "1.6.0", default-features = false, features = ["std"] }
 prost = { version = "0.13.4", default-features = false, features = ["derive"] }
 rand = { version = "0.8.5", default-features = false }
+rand_chacha = { version = "0.3.0", default-features = false }
```
cylewitruk: `0.3.1` maybe? I'd say we should try to bump to `0.9.0` but then we have to bump other stuff and that would grow the scope of this PR (but we should probably look into that later, just noting it here).

signer/src/error.rs
```diff
@@ -613,8 +613,8 @@ pub enum Error {
     TypeConversion,
 
     /// Encryption error
-    #[error("encryption error")]
-    Encryption,
+    #[error("received an error when encrypting the signer state for storage {0}")]
+    Encryption(#[source] wsts::errors::EncryptionError),
```
cylewitruk: - `WstsEncrypt` maybe (in-case we have other encryption-style errors in the future)? 
- Consider `#[from]` as this is a very specific error without additional context? Alt. add additional context, e.g. the aggregate key.

signer/src/util.rs
```diff
@@ -237,7 +237,7 @@ mod tests {
             } else {
                 // Just picked a random error here (which isn't
                 // a AllFailoverClientsFailed)
-                Err(Error::Encryption)
+                Err(Error::DivideByZero)
```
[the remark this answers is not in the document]
cylewitruk: We have the `Error::Dummy` testing variant specifically for this :)

signer/src/wsts_state_machine.rs
cylewitruk: Resolved in [1102863](https://github.com/stacks-sbtc/sbtc/pull/1746/commits/11028635ed1cfdf0011de17f4ea97ed0b97c8d64), closing

signer/src/error.rs
djordon: Good call. I ended up adding context in https://github.com/stacks-sbtc/sbtc/pull/1746/commits/74883bc601d1c53abf60d84796321772cccbfe85

Cargo.toml
djordon: Bumped in https://github.com/stacks-sbtc/sbtc/pull/1746/commits/889fb434b261c3c628b1c6612fca63bcaebb9c78

signer/src/util.rs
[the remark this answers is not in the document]
djordon: Made the switch in https://github.com/stacks-sbtc/sbtc/pull/1746/commits/4dd4e71fb7ef5b2c5f57567ab110f04397b74a21

cylewitruk (dismissed):
LGTM 👍

signer/src/wsts_state_machine.rs
```diff
@@ -465,41 +485,108 @@ impl SignerStateMachine {
...
             dkg_threshold,
             num_parties,
             num_keys,
             id,
             key_ids,
-            signer_private_key.into(),
+            private_key.into(),
             public_keys,
-            &mut rng,
+            &mut OsRng,
```
MCJOHN974, line 497: Do we want use some `ChaCha20Rng` instead `OsRng` as well?

djordon: Maybe, it's useful to think about why though. I believe that `OsRng`, `StdRng`, and `ChaCha20Rng` are cryptographically secure pseudo random number generators (CSPRNG) so I think either will be fine. What do you think we'd gain from switching?

MCJOHN974: Yea, I was thinking that rust don't guarantee that `OsRng` is CSPRNG, but just double checked and it seems that rust guarantee it.

signer/src/error.rs
```diff
@@ -613,8 +613,12 @@ pub enum Error {
     TypeConversion,
 
     /// Encryption error
```
matteojug: nit: old comment? It's probably still valid, but we could scope it better like in `WstsDecrypt`

signer/tests/integration/transaction_signer.rs
```diff
@@ -840,6 +844,182 @@ async fn new_state_machine_per_valid_sighash() {
...
+async fn nonce_response_unique_nonces() {
+    let db = testing::storage::new_test_database().await;
+
+    let mut rng = get_rng();
+    // Build the test context with mocked clients
+    let ctx = TestContext::builder()
+        .with_storage(db.clone())
+        .with_mocked_bitcoin_client()
+        .with_mocked_emily_client()
+        .with_mocked_stacks_client()
```
matteojug: nit: `with_mocked_clients`

signer/src/wsts_state_machine.rs
```diff
@@ -465,41 +485,108 @@ impl SignerStateMachine {
...
-    /// Create a state machine from loaded DKG shares for the given aggregate key
+    /// Create a state machine from loaded DKG shares for the given
+    /// aggregate key
+    ///
+    /// # Note
+    ///
+    /// Loaded state machines should only be used for signing rounds and
+    /// not for DKG, since they will always create the same secret shares.
+    /// TODO: Make it so that we have separate state machines for signing
+    /// and DKG.
```
matteojug, line 571: This should be fine at the moment since we use `load` only on `NonceRequest` message, and we prevent creating state machines for `WstsMessageId::Dkg` in there. But still, it does seem dangerous, I'm wondering if we can do something to avoid this behaviour (eg, setting `created_at` to a random value)

signer/src/wsts_state_machine.rs
```diff
@@ -542,14 +631,19 @@ impl SignerStateMachine {
...
-        let public_shares = self.dkg_public_shares.clone().encode_to_vec();
-
-        // After DKG, each of the signers will have "new public keys".
-        let encrypted_private_shares =
-            wsts::util::encrypt(&self.0.network_private_key.to_bytes(), &encoded, rng)
-                .map_err(|_| error::Error::Encryption)?;
+        let public_shares = self.inner.dkg_public_shares.clone().encode_to_vec();
+
+        // After DKG, each of the signers will have "new public keys". The
+        // can error if we are encrypting more than 68719476752 bytes.
```
matteojug: nit: missing a word, `The ? can error`

signer/src/wsts_state_machine.rs
djordon: I don't know. My first thought was that it shouldn't hurt to do something, like randomizing the `created_at`, but now I'm not so sure. If we footgun ourselves in this way, running DKG with a loaded state machine, it may be harmless if the same signer set is unchanged and everyone uses the same private keys because the aggregate key would be the same. However, if the signer set changes or a private key changes then we'll regenerate the private polynomial coefficients, which feels worse, even though that is exactly what this code is designed to do.

My second thought is that I'd prefer to make such a mistake impossible using the type system.

signer/src/error.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1746/commits/5934815d617b5b958035d9d332887e4d75c4959b

signer/src/wsts_state_machine.rs
djordon: Whoops, finished my thought in https://github.com/stacks-sbtc/sbtc/pull/1746/commits/50f69854eb66d6165403d881c7ff146c8502a31f

matteojug (approved):

evonide:
This is a summary of a deep dive and detailed analysis with additional feedback from @djordon.

This change alters the Distributed Key Generation (DKG) process by making the creation of each participant's secret polynomial deterministic. It replaces the use of `OsRng` with `ChaCha20Rng` a cryptographically secure pseudo-random number generator. This generator is seeded with a value derived from the Bitcoin block hash that triggered the DKG round and the signer's own long-term private key.

The main motivation of this change is to provide the potential to recover from catastrophic failure scenarios where multiple signers simultaneously lose their databases, such as widespread datacenter outages affecting signers running on the same cloud infrastructure. This enables the possibility of writing cooperative recovery code in emergency situations without requiring such code to exist proactively.

Crucially, this allows a signer to perfectly reconstruct their **own secret polynomial** for any past DKG round, but not their final private key share. A signer's final private key share is the sum of shares received from all other participants during that DKG round. Therefore, reconstructing it requires a **"cooperative recovery"** process where the entire group of signers must agree to help. They would re-initiate the DKG process using the original block hash, with each participant using their long-term key and that old block hash to regenerate their identical, original secret polynomial. They would then re-calculate and securely re-send the specific share owed to the recovering member, allowing them to participate in signing again without the group having to abandon the established aggregate key.

For reference the DKG protocol works like this:
1. **Initialization** - Agree on set of signers and threshold t (minimum required to create signature)
1. **Generate secret polynomial** - Each signer generates a unique, ~unpredictable~ **predictable** (given the private key and block height) set of coefficients for their secret polynomial.
1. **Broadcast commitments** - Each signer creates and broadcasts public commitment to their secret polynomial. Will be used to verify actions without revealing actual secrets
1. **Distribute secret shares** - Each signer calculates a secret share for every other participant by evaluating the secret polynomial at the other participant's index. The shares are then communicated over encrypted channels.
1. **Verification** - Each signer verifies received secret shares by checking them against the public commitments they received
1. **Key creation** - Each signer combines valid secret shares to create the final private key share used for signing. The group's public key is computed from all public commitments.

Note that this PR only affects the second step here. Finally, regarding the security impact we accept a cryptographic trade-off. The core risk is a weakening of forward secrecy. The DKG rounds are no longer independent sessions; they are permanently tied to the signers' long-term keys. If an attacker compromises a signer's long-term private key, they can perfectly reconstruct that signer's secret polynomial for all past and future DKG rounds, but crucially they still cannot calculate that signer's final private key shares on their own. However, the compromised polynomial would allow the attacker to impersonate the compromised signer within the DKG protocol, correctly generating all expected messages and secret shares as if they were the legitimate participant.

I agree that it's fine to drop the forward secrecy which we don't need to improve our disaster recovery and this looks like a thought through implementation to me.

MCJOHN974 (approved):

PR #1748, opened 2025-06-18
chore: introduce make fix command

MCJOHN974:
## Description

I widely use `cargo clippy --fix`, it can easily eliminate all unused imports or other machine-applicable clippy lints. However, I don't remember sometimes whole long command we use for clippy in CI, which makes me open our Makefile, copy command from it, edit adding --fix flag, which is annoying. This PR aims to solve the problem.

Closes: no issue

## Changes

Added two new Makefile commands:

`make fix` -- runs `cargo fmt` and `clippy --fix` for all targets we need. If you have any uncommited changes it will bail and only `cargo fmt` will be runned.
`make fix-uncommited` -- same as above, but this command won't bail if you have any uncommited changes and will perform it's work.

## Note

I'm not sure if we need both versions of the command since I usually don't afraid to run clippy on uncommited changes if same for you then probably we want just one command.

## Testing Information

It is developer tools, no testing needed. I tryied them locally and it seems they do what they should.

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,8 +46,12 @@ lint:
 	pnpm --recursive run lint
 
 fix:
-	cargo $(CARGO_FLAGS) fmt --all
-	cargo $(CARGO_FLAGS) clippy --workspace --all-targets --all-features --no-deps --fix -- -D warnings
+	cargo fmt --all
+	cargo clippy --workspace --all-targets --all-features --no-deps --fix -- -D warnings
+
+fix-uncommited:
+	cargo fmt --all
+	cargo clippy --workspace --all-targets --all-features --no-deps --fix --allow-dirty -- -D warnings
 
 format:
 	cargo $(CARGO_FLAGS) fmt
--- a/Makefile
+++ b/Makefile
@@ -46,8 +46,8 @@ lint:
 	pnpm --recursive run lint
 
 fix:
-	cargo fmt --all
-	cargo clippy --workspace --all-targets --all-features --no-deps --fix -- -D warnings
+	cargo $(CARGO_FLAGS) fmt --all
+	cargo $(CARGO_FLAGS) clippy --workspace --all-targets --all-features --no-deps --fix -- -D warnings
 
 fix-uncommitted:
 	cargo fmt --all
--- a/Makefile
+++ b/Makefile
@@ -50,8 +50,8 @@ fix:
 	cargo $(CARGO_FLAGS) clippy --workspace --all-targets --all-features --no-deps --fix -- -D warnings
 
 fix-uncommitted:
-	cargo fmt --all
-	cargo clippy --workspace --all-targets --all-features --no-deps --fix --allow-dirty -- -D warnings
+	cargo $(CARGO_FLAGS) fmt --all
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -45,6 +45,14 @@ lint:
 	cargo $(CARGO_FLAGS) clippy --workspace --all-targets --all-features --no-deps -- -D warnings
 	pnpm --recursive run lint
 
+fix:
+	cargo fmt --all
+	cargo clippy --workspace --all-targets --all-features --no-deps --fix -- -D warnings
```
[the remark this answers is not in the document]
MCJOHN974: I intentionally removed it because it makes everything slower and I encountered problems running without locked mode only in integration tests, not clippy nor fmt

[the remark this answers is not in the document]
MCJOHN974: Keeping it open for now in case any human prefer locked mode for clippy or fmt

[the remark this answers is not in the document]
aldur: If we don't want that, you should explicitly add a comment stating that it is intended.

Makefile
```diff
@@ -45,6 +45,14 @@ lint:
...
+
+fix-uncommitted:
+	cargo fmt --all
+	cargo clippy --workspace --all-targets --all-features --no-deps --fix --allow-dirty -- -D warnings
```
aldur: With a bit of `Makefile`-fu one could probably avoid the duplication here

Makefile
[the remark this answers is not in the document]
cylewitruk: Yeah `--locked` is generally preferred because otherwise `cargo` is free to pull other versions of transient dependencies, which can result in unexpected lockfile updates (and thus version changes).

I see no reason why it should make anything slower, though...

If you're running into issues due to `--locked` in integration tests then I think that points to something else in your setup that might be a bit "off"...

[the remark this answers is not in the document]
MCJOHN974: Okay lets add locked then

Makefile
MCJOHN974: [cb3e46b](https://github.com/stacks-sbtc/sbtc/pull/1748/commits/cb3e46bfd2a6200b574efb7076210b18b7446e2b)

cylewitruk: lol I preferred the more explicit version tbh 🙈

MCJOHN974: I don't really have strong opinion on this, downside of current version is that it is a little bit harder to eye-parse it and understand what command we actually run, but it's not so hard so I'm ok with this downside. Downside of explicit version is probability to have different commands in `fix` and `lint` but I'm ok with this as well since it won't harm our code quality if we review CI changes properly and I believe we are strict enough in reviewing them

MCJOHN974: @djordon @matteojug wdyt? And also about if we need both commands for fixing commited and uncommited (I personally never seen `clippy fix` breaking anything so I'm ok to let it fix uncommited code

matteojug: No strong opinion either way, but since we will have 3 targets using small variations I would be more inclined to go with the current (deduped) one.

djordon (approved):

Makefile
MCJOHN974: @aldur do you think we can resolve this thread?

djordon: I do not have a strong opinion either way, but lean toward the old, more explicit, version. But fine with whichever.

PR #1752, opened 2025-06-18
fix: support brand new bootstrap signer set

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1751

## Changes

1. Load the aggregate key from the last DKG shares if we did not run DKG and do not have any signer set info in the registry.
2. Signers proceed with verification using the latest verified DKG shares if they do not have info in the registry. They fall back to the latest shares if this is a contract deployment and the contracts have not been deployed.

## Testing Information

Tested this on devenv with a fresh build and was able to sweep a deposit without any issues. 

Also, I redid one of our previously ignored tests and completely removed two of the variants that were ignored. I couldn't get them to work and got tired of attempting to fix them. Perhaps it would be better to keep them around but ignored. I'm okay either way.

## Checklist:


signer/src/transaction_signer.rs
```diff
@@ -552,20 +552,29 @@ where
...
+        let signer_set_info = match state.registry_signer_set_info() {
             Some(info) => info,
-            None => db
-                .get_latest_verified_dkg_shares()
-                .await?
-                .map(SignerSetInfo::from)
-                .ok_or_else(|| Error::NoDkgShares)?,
+            None => match db.get_latest_verified_dkg_shares().await? {
+                Some(info) => info.into(),
+                None if !state.sbtc_contracts_deployed() => match &request.contract_tx {
```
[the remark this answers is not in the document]
djordon: Fair

signer/src/transaction_coordinator.rs
```diff
@@ -346,13 +342,28 @@ where
...
             }
         } else {
-            registry_aggregate_key
-                .ok_or(Error::MissingAggregateKey(*bitcoin_chain_tip.block_hash))?
+            // If we do not have signer set info in the registry, then we
+            // are in the bootstrap phase. Our latest DKG shares may be
+            // 'Unverified', but if we are here then they are not 'Failed',
+            // since we will coordinate DKG if our last DKG shares are
+            // 'Failed'.
+            match registry_signer_set_info.as_ref() {
```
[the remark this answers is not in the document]
djordon: Fair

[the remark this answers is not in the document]
matteojug: I would vote to keep it as is, it's easier to get the full picture without having to follow an indirection.

signer/src/transaction_coordinator.rs
```diff
@@ -346,13 +342,28 @@ where
...
                 }
             }
         } else {
-            registry_aggregate_key
-                .ok_or(Error::MissingAggregateKey(*bitcoin_chain_tip.block_hash))?
+            // If we do not have signer set info in the registry, then we
+            // are in the bootstrap phase. Our latest DKG shares may be
+            // 'Unverified', but if we are here then they are not 'Failed',
+            // since we will coordinate DKG if our last DKG shares are
+            // 'Failed'.
```
matteojug: IIUC we can have a race here, given that between `coordinate_dkg` getting the shares count and the `get_latest_encrypted_dkg_shares` below we could have the block observer marking the latest shares as failed, so this may not be always true. But I guess it's fine regardless, since then in `check_and_submit_rotate_key_transaction` we would err with `DkgVerificationFailed` and skip the tenure. Maybe consider updating the comment?

signer/src/transaction_signer.rs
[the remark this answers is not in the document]
matteojug: As the other, I would vote to keep it as is, or maybe use `matches!` to include the is `SmartContract` in this arm and remove one of the dup `Err(Error::NoVerifiedDkgShares)`

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -673,166 +493,315 @@ async fn process_complete_deposit() {
...
-    F: FnOnce(u64, Sender<StacksTransaction>) -> Box<dyn FnOnce(&mut MockStacksInteract)>,
-{
-    let db = testing::storage::new_test_database().await;
-    let mut rng = get_rng();
+async fn deploy_smart_contracts_coordinator() {
+    let (_, signer_key_pairs): (_, [Keypair; 3]) = testing::wallet::regtest_bootstrap_wallet();
+    let (rpc, faucet) = regtest::initialize_blockchain();
 
-    let num_messages = smart_contracts.len();
+    signer::logging::setup_logging("info,signer=debug", false);
```
matteojug: nit: leftover?

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -673,166 +493,315 @@ async fn process_complete_deposit() {
...
 
-    let bitcoin_block: model::BitcoinBlock = Faker.fake_with_rng(&mut rng);
-    db.write_bitcoin_block(&bitcoin_block).await.unwrap();
+    // We need to populate our databases, so let's fetch the data.
+    let emily_client = EmilyClient::try_new(
+        &Url::parse("http://testApiKey@localhost:3031").unwrap(),
+        Duration::from_secs(1),
+        None,
+    )
+    .unwrap();
```
matteojug, line 659: nit: do we need the real Emily for this test?

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -673,166 +493,315 @@ async fn process_complete_deposit() {
...
+        // have verified shares.
+        client
+            .expect_get_current_signers_aggregate_key()
+            .returning(move |_| {
+                let db = db.clone();
+                Box::pin(async move {
+                    let shares = db.get_latest_verified_dkg_shares().await?;
+                    Ok(shares.map(|sh| sh.aggregate_key))
+                })
+            });
```
matteojug, line 613: nit: isn't this a nop given we override it right below?

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -263,198 +260,21 @@ fn mock_deploy_all_contracts(
...
             Box::pin(async move {
                 Ok(AccountInfo {
                     balance: 1_000_000,
                     locked: 0,
                     unlock_height: 0u64.into(),
-                    nonce,
+                    nonce: 16,
                 })
             })
         });
```
matteojug: nit: given we override those in `mock_stacks_core`, should we remove them from `mock_deploy_all_contracts`?

matteojug (commented):
LGTM, just some nits

signer/tests/integration/transaction_coordinator.rs
djordon: Kinda, this test spins up a block observer so we need either a real emily or a mocked one and a real one is easier.

signer/tests/integration/transaction_coordinator.rs
djordon: I need to delete the one below because that one is the noop. It doesn't override this one.

signer/tests/integration/transaction_coordinator.rs
djordon: They shouldn't be overridden, they should get used after these mocks. But since these mocks go on for forever, they later ones are effectively ignored.

signer/tests/integration/transaction_coordinator.rs
djordon: https://github.com/stacks-sbtc/sbtc/pull/1752/commits/565968315b786bbf4ac46b2271987b8933d66fab

signer/tests/integration/transaction_coordinator.rs
djordon: I ended up removing this one here and adding it back here. The last addition happened in https://github.com/stacks-sbtc/sbtc/pull/1752/commits/72ca28fe68bae83026cd35dbd12d61ab0e461934

signer/src/transaction_signer.rs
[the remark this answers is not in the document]
djordon: I switched to matches in https://github.com/stacks-sbtc/sbtc/pull/1752/commits/491f645b2de18b4f112a7b37029e38102af777ca, it looked a little easier to follow

Issue #1750, opened 2025-06-18
[Feature]: Only consider one pre-sign request per bitcoin block


djordon:
# Feature - Only consider one pre-sign request per bitcoin block

## 1. Description

The signers should only consider one pre-sign request per bitcoin block. This is to protect the signers from doing unnecessary work. If they get two, it's probably an error or something, and can be safely ignored.

## 2. Technical Details:

The signers should keep track of the pre-sign requests from the coordinator for each bitcoin block and ignore the request if you get more than one.

### 2.1 Acceptance Criteria:
   - [ ] The signers consider at more one pre-sign request per bitcoin block

PR #1757, opened 2025-06-20
[feat] allow one pre-sign request per block

MCJOHN974:
## Description

Closes: #1750

## Changes

WIP
## Checklist:


signer/src/transaction_signer.rs
```diff
@@ -425,6 +429,14 @@ where
...
         let db = self.context.get_storage_mut();
 
+        if chain_tip.block_height == self.last_presign_height {
+            tracing::debug!(
+                ?chain_tip,
+                "already processed presign request for this stacks chain tip"
+            );
+            return Ok(());
+        }
+
```
MCJOHN974: Maaaybe we want this to be `<` instead of `==` and maaaybe it is better to store hash here instead of height in case of reorgs

signer/src/lib.rs
```diff
@@ -116,7 +122,7 @@ pub const WITHDRAWAL_DUST_LIMIT: u64 = 546;
 /// The value here was taken from the last paragraph of the opening comment
 /// of https://github.com/stacks-network/sbtc/discussions/12 and in the
 /// comments of https://github.com/stacks-network/sbtc/issues/16.
-pub const WITHDRAWAL_MIN_CONFIRMATIONS: u64 = 6;
+pub const WITHDRAWAL_MIN_CONFIRMATIONS: u64 = BITCOIN_CONFIRMATION_DELAY;
```
MCJOHN974: This change and one below are out of scope of this PR, but I think they are pretty small to include them here.

signer/src/transaction_signer.rs
```diff
@@ -471,6 +484,20 @@ where
...
+        tracing::debug!(?chain_tip, "updating used for presign blocks",);
+        self.used_presign_blocks.insert(*chain_tip);
+        // We don't want [`self.used_presign_blocks`] to grow infinitely, so saving here
+        // only the last `BITCOIN_CONFIRMATION_DELAY` blocks, since probability of receiving presign
+        // requests for blocks older than that is very low. Also, if due to this we process
+        // presign requests for the same block multiple times, it is not a problem, since
+        // since it will just some waste of signer time, not a functionality issue.
+        if self.used_presign_blocks.len() > crate::BITCOIN_CONFIRMATION_DELAY as usize {
+            self.used_presign_blocks.retain(|b| {
+                b.block_height >= chain_tip.block_height - crate::BITCOIN_CONFIRMATION_DELAY
```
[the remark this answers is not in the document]
MCJOHN974: We probably will never encounter bitcoin heights smaller then 6 but for general robustness makes sense

signer/src/transaction_signer.rs
```diff
@@ -471,6 +484,20 @@ where
 
         self.send_message(BitcoinPreSignAck, &chain_tip.block_hash)
             .await?;
+
+        tracing::debug!(?chain_tip, "updating used for presign blocks",);
```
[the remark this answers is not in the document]
MCJOHN974: I don't really like both your and mine wordings so let's go with something like this [bd95d96](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/bd95d9687820573d0b1e3461c3f206d967c406b9)

signer/src/lib.rs
```diff
@@ -131,7 +137,7 @@ pub const WITHDRAWAL_BLOCKS_EXPIRY: u64 = 24;
 ///
 /// This is the value in https://github.com/stacks-network/sbtc/issues/1363's
 /// proposed "buffer for expiring requests" section.
-pub const WITHDRAWAL_EXPIRY_BUFFER: u64 = 6;
+pub const WITHDRAWAL_EXPIRY_BUFFER: u64 = BITCOIN_CONFIRMATION_DELAY;
```
matteojug: I don't think this `6` was related to the standard bitcoin confirmation, it was kinda an arbitrary buffer value. And if so, I think we should revert the changes in here since we would use the new constant in just a single place.

signer/src/transaction_signer.rs
```diff
@@ -140,6 +141,9 @@ pub struct TxSignerEventLoop<Context, Network, Rng> {
     pub wsts_state_machines: LruCache<StateMachineId, SignerStateMachine>,
     /// The threshold for the signer
     pub threshold: u32,
+    /// Set of bitcoin blocks for which the signer has already processed
+    /// presign request.
+    pub used_presign_blocks: HashSet<BitcoinBlockRef>,
```
matteojug: I think we can simplify this and just keep the latest chain tip hash for which we signed a pre sign: we enter in `handle_bitcoin_pre_sign_request` only if the message is for a `ChainTipStatus::Canonical` chain tip and sent by the coordinator, so if we just keep track of the latest chain tip we signed for and bail if the new request matches it should be enough.

signer/src/transaction_signer.rs
```diff
@@ -425,6 +430,14 @@ where
     ) -> Result<(), Error> {
         let db = self.context.get_storage_mut();
 
+        if self.used_presign_blocks.contains(chain_tip) {
+            tracing::debug!(
+                ?chain_tip,
+                "already processed presign request for this block"
+            );
+            return Ok(());
+        }
```
matteojug: We should probably return an error here (and then the top level `run` will take care of logging it): a coordinator is supposed to send only one presign request per tenure, if it tries to send more we should get angry about it.

signer/tests/integration/transaction_signer.rs
```diff
@@ -715,6 +721,129 @@ pub async fn presign_requests_with_dkg_shares_status(status: DkgSharesStatus, is
...
+        .with_mocked_emily_client()
+        .with_mocked_stacks_client()
+        .build();
+
+    let (rpc, faucet) = sbtc::testing::regtest::initialize_blockchain();
+
+    let signers = TestSignerSet::new(&mut rng);
+    // Create a test setup object so that we can simply create proper DKG
+    // shares in the database. Note that calling TestSweepSetup2::new_setup
+    // creates two bitcoin block.
```
djordon:
```suggestion
    // Create a test setup object so that we can easily create proper DKG
    // shares in the database. Note that calling TestSweepSetup2::new_setup
    // creates two bitcoin blocks.
```

signer/src/transaction_signer.rs
```diff
@@ -471,6 +484,23 @@ where
 
         self.send_message(BitcoinPreSignAck, &chain_tip.block_hash)
             .await?;
+
+        tracing::debug!(?chain_tip, "updating used_presign_blocks",);
+        self.used_presign_blocks.insert(*chain_tip);
```
djordon: I think that we want to note that you've seen a message for this block immediately after checking whether we've seen something for this block, not after processing the message. A malicious coordinator could send junk that causes us to bail when we call `construct_package_sighashes`. Doing so would allow them to continually waste our compute as the signer processes their crappy request.

signer/src/transaction_signer.rs
```diff
@@ -425,6 +430,14 @@ where
     ) -> Result<(), Error> {
         let db = self.context.get_storage_mut();
 
+        if self.used_presign_blocks.contains(chain_tip) {
+            tracing::debug!(
+                ?chain_tip,
```
djordon: The bitcoin chain tip block hash is already in our logs, no need to add it again.

signer/tests/integration/transaction_signer.rs
```diff
@@ -715,6 +721,129 @@ pub async fn presign_requests_with_dkg_shares_status(status: DkgSharesStatus, is
...
+    // it should return an error, but, since we already processed this
+    // block, it should return early with Ok(()).
+    // Seing Ok(()) here indicates that we ignored the request.
+
+    set_verification_status(&db, aggregate_key, DkgSharesStatus::Unverified).await;
+    let result = tx_signer
+        .handle_bitcoin_pre_sign_request(&sbtc_context, &chain_tip)
+        .await;
+
+    assert!(result.is_ok());
```
djordon: As @matteojug mentioned, we want to error when we get a duplicate message. No need to change the status of the shares, they shouldn't matter for this test.

MCJOHN974: [87b0059](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/87b005954edf255d90eeee1e8588c27ed76d80ac)

signer/src/transaction_signer.rs
MCJOHN974: [87b0059](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/87b005954edf255d90eeee1e8588c27ed76d80ac)

signer/src/transaction_signer.rs
MCJOHN974: [87b0059](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/87b005954edf255d90eeee1e8588c27ed76d80ac)

signer/src/lib.rs
MCJOHN974: [87b0059](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/87b005954edf255d90eeee1e8588c27ed76d80ac)

signer/src/transaction_signer.rs
MCJOHN974: [87b0059](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/87b005954edf255d90eeee1e8588c27ed76d80ac)

signer/src/transaction_signer.rs
MCJOHN974: [87b0059](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/87b005954edf255d90eeee1e8588c27ed76d80ac)

signer/tests/integration/transaction_signer.rs
MCJOHN974: We actually have a bit more of this =)

[454ca99](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/454ca997562307e5af989cccff2581f7d2754fbe)

MCJOHN974:
Hmm it looks like it run CI on wrong commit....

```shell
$ git checkout 454ca997562307e5af989cccff2581f7d2754fbe && sed -n '742p' signer/tests/integration/transaction_coordinator.rs

Note: switching to '454ca997562307e5af989cccff2581f7d2754fbe'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 454ca997 english in comments
        for _ in 0..num_messages {
```

signer/src/error.rs
```diff
@@ -396,6 +396,11 @@ pub enum Error {
     #[error("could not recover the public key from the signature: {0}, digest: {1}")]
     InvalidRecoverableSignature(#[source] secp256k1::Error, secp256k1::Message),
 
+    /// This is thrown when we attempt to process a presign request for
+    /// a block for which we already processed presign request.
+    #[error("Recieved presign request for already processed block {0}")]
+    InvalidPresignRequest(String),
```
djordon: Why a string instead of the block hash? I usually prefer something other than a string here, since it's clear what context we are actually being provided by looking at the variant.

signer/src/error.rs
```diff
@@ -396,6 +396,11 @@ pub enum Error {
     #[error("could not recover the public key from the signature: {0}, digest: {1}")]
     InvalidRecoverableSignature(#[source] secp256k1::Error, secp256k1::Message),
 
+    /// This is thrown when we attempt to process a presign request for
+    /// a block for which we already processed presign request.
```
djordon:
```suggestion
    /// This is thrown when we attempt to process a presign request for
    /// a block for which we have already processed a presign request.
```

signer/tests/integration/transaction_signer.rs
```diff
@@ -877,6 +1040,10 @@ async fn max_one_state_machine_per_bitcoin_block_hash_for_dkg() {
         wsts_state_machines: LruCache::new(NonZeroUsize::new(100).unwrap()),
         signer_private_key: ctx.config().signer.private_key,
         threshold: 2,
+        last_presign_block: BitcoinBlockRef {
+            block_hash: BitcoinBlockHash::from([0; 32]),
+            block_height: 0u64.into(),
+        },
```
djordon: Maybe we should add a testing function impl for `BitcoinBlockRef` that returns this.

signer/src/transaction_signer.rs
```diff
@@ -140,6 +142,9 @@ pub struct TxSignerEventLoop<Context, Network, Rng> {
     pub wsts_state_machines: LruCache<StateMachineId, SignerStateMachine>,
     /// The threshold for the signer
     pub threshold: u32,
+    /// Last bitcoin block for which the signer has already processed
+    /// presign request.
+    pub last_presign_block: BitcoinBlockRef,
```
djordon: Do we gain anything from using the block ref instead of the block hash? I suppose it doesn't matter much, so I'm mainly just curious, but it seems simpler to use the block hash.

Also, using an `Option<BitcoinBlockWhatever>` here seems more "correct".

signer/tests/integration/transaction_signer.rs
```diff
@@ -715,6 +739,141 @@ pub async fn presign_requests_with_dkg_shares_status(status: DkgSharesStatus, is
...
+    // Check that we store information that we processed this block
+    assert_eq!(tx_signer.last_presign_block, chain_tip);
+
+    // Now we will try to process the same block again, but this time we will
+    // Put unverified shares in the database. Since shares are not verified
+    // it should return an error, but, since we already processed this
+    // block, it should return early with Ok(()).
+    // Seing Ok(()) here indicates that we ignored the request.
+
+    set_verification_status(&db, aggregate_key, DkgSharesStatus::Verified).await;
```
djordon: I think we can remove this now.

signer/src/error.rs
MCJOHN974: Yeaa, was too long on Emily side =)

[0499863](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/049986319498813d2e5b31928901dcbbcc924fb4)

signer/src/transaction_signer.rs
MCJOHN974: [32ed3e9](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/32ed3e943814827fa4b179df8cad024e70356800)

signer/tests/integration/transaction_signer.rs
MCJOHN974: Since we moved to option we simply use `None` here and do not need any helpers for this

signer/tests/integration/transaction_signer.rs
MCJOHN974: Actually code here is ok, just the comment is artifact from some old impl: [837d223](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/837d2238dea198a4b4efbf90b72c3e5089f51166)

signer/src/error.rs
MCJOHN974: [d521d72](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/d521d72e6e1bd97e4b8c6793bc30c3f88816a05d)

signer/src/transaction_signer.rs
```diff
@@ -430,12 +430,10 @@ where
...
         let db = self.context.get_storage_mut();
 
-        if self.used_presign_blocks.contains(chain_tip) {
-            tracing::debug!(
-                ?chain_tip,
-                "already processed presign request for this block"
-            );
-            return Ok(());
+        if let Some(processed_block) = &self.last_presign_block {
+            if processed_block == &chain_tip.block_hash {
```
matteojug: nit: we could directly check if `self.last_presign_block == Some(chain_tip.block_hash)` (with some `&` around) for better readability

signer/tests/integration/transaction_signer.rs
```diff
@@ -715,6 +721,131 @@ pub async fn presign_requests_with_dkg_shares_status(status: DkgSharesStatus, is
...
+pub async fn presign_request_ignore_request_if_already_processed_this_block() {
+    let db = testing::storage::new_test_database().await;
+
+    let mut rng = get_rng();
+    // Build the test context with mocked clients
+    let ctx = TestContext::builder()
+        .with_storage(db.clone())
+        .with_mocked_bitcoin_client()
+        .with_mocked_emily_client()
+        .with_mocked_stacks_client()
```
matteojug: nit: we can use `with_mocked_clients` if we want to mock all the clients

signer/tests/integration/transaction_signer.rs
```diff
@@ -715,6 +721,131 @@ pub async fn presign_requests_with_dkg_shares_status(status: DkgSharesStatus, is
...
+        request_package: vec![sbtc_requests],
+        fee_rate: 2.0,
+        last_fees: None,
+    };
+
+    let result = tx_signer
+        .handle_bitcoin_pre_sign_request(&sbtc_context, &chain_tip)
+        .await;
+
+    // We processing this block with verified shares and thus it should be ok.
```
matteojug: nit?
```suggestion
    // We are processing this block with verified shares and thus it should be ok.
```

signer/tests/integration/transaction_signer.rs
```diff
@@ -715,6 +721,131 @@ pub async fn presign_requests_with_dkg_shares_status(status: DkgSharesStatus, is
...
+    assert!(result.is_ok());
+
+    // Check that we store information that we processed this block
+    assert_eq!(tx_signer.last_presign_block, Some(chain_tip.block_hash));
+
+    // Now we will try to process the same block again, but since we already
+    // processed it, we should get an error.
+    set_verification_status(&db, aggregate_key, DkgSharesStatus::Verified).await;
```
matteojug: nit: we already verified the shares above so this shouldn't be needed
```suggestion
```

matteojug (approved):
LGTM, some small nits

signer/tests/integration/transaction_signer.rs
MCJOHN974: [08785b2](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/08785b2418705642f741f044b919a75487d28972)

signer/tests/integration/transaction_signer.rs
MCJOHN974: [0578bc6](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/0578bc6d80a3cf9ddfdb8996861e9e937dec0405)

signer/tests/integration/transaction_signer.rs
MCJOHN974: [883eac4](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/883eac4a55136e16c2264fff0c579ba0ea0d20e5)

signer/src/transaction_signer.rs
MCJOHN974: [75bae06](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/75bae065562690ba6061cb3ab66120c2ace84c13)

matteojug (dismissed):

djordon (approved):
✅ Looks good

signer/src/transaction_signer.rs
```diff
@@ -471,6 +480,9 @@ where
 
         self.send_message(BitcoinPreSignAck, &chain_tip.block_hash)
             .await?;
+
+        self.last_presign_block = Some(chain_tip.block_hash);
```
djordon: This should be moved to be immediately after the `self.last_presign_block == Some(chain_tip.block_hash)` check above.

djordon (changes requested):

signer/src/transaction_signer.rs
MCJOHN974: [5b0262d](https://github.com/stacks-sbtc/sbtc/pull/1757/commits/5b0262d9049b227493b7a1f53b98fad94b0da651)

matteojug (approved):

PR #1762, opened 2025-06-24
fix: specify the event name in the wait-list of wait-other-jobs

djordon:
## Description

This may fix that issue that we've been seeing in CI. Doesn't address the warnings though.

The underlying action can be found here https://github.com/kachick/wait-other-jobs/tree/215c93f93f91e8f9a5ccb7fc0cc78220c9fd8a99?tab=readme-ov-file#usage, while our use of it can be found here https://github.com/stacks-sbtc/actions/blob/181f8c67da2707c66b5e31f24e7418c47adefdd1/wait-other-jobs/action.yml

## Changes

1. Specify the `eventName` in the wait list, in the same way that the usage docs demonstrate. But I'm not 100% sure that this fixes it, also not sure whether this introduces other issues.

## Checklist

### DIFF
--- a/.github/workflows/on-push.yaml
+++ b/.github/workflows/on-push.yaml
@@ -288,6 +288,7 @@ jobs:
               {
                 "workflowFile": "on-push.yaml",
                 "jobName": "Build Test Artifacts",
+                "eventName": "${{ github.event_name }}",
                 "optional": false,
                 "startupGracePeriod": {
                   "minutes": 5


BowTiedDevOps (approved):
LGTM - I believe this won't induce any new issues in the action, and it should resolve the current failure issue. P.S. Regarding the warnings, they can't be fixed as of now because the next major release (as specified in the warnings) doesn't yet exist.

fabgtc:
Do we want to fix also these warnings?

_Warning: Input 'wait-seconds-before-first-polling' has been deprecated with message: This option will be removed in next major release. Use warmup-delay instead
Warning: Input 'min-interval-seconds' has been deprecated with message: This option will be removed in next major release. Use minimum-interval instead_

fabergat:
Do we want to fix also these warnings?

Warning: Input 'wait-seconds-before-first-polling' has been deprecated with message: This option will be removed in next major release. Use warmup-delay instead
Warning: Input 'min-interval-seconds' has been deprecated with message: This option will be removed in next major release. Use minimum-interval instead

matteojug (approved):

MCJOHN974 (approved):

djordon:
> Do we want to fix these warnings?
> 
> _Warning: Input 'wait-seconds-before-first-polling' has been deprecated with message: This option will be removed in next major release. Use warmup-delay instead Warning: Input 'min-interval-seconds' has been deprecated with message: This option will be removed in next major release. Use minimum-interval instead_

Yeah we do, but we cannot fix them in this repo.

PR #1765, opened 2025-06-26
feat: support updates to the signing threshold parameter

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1617.

Well, we already supported changes to the signing threshold, that functionality was probably completed in https://github.com/stacks-sbtc/sbtc/pull/1707. This cleans up the event loops and add some tests.

The one caveat is that we do not support changing the signature threshold to a value that is greater than the size of the previous signer set. This is due to the fact that new signers do not currently "know" the public key of the last signing set.

## Changes

* Remove the signature share threshold parameters from the tx-coordinator and tx-signer event loops.
* Update the testing structs to function without depending on the signature threshold and to use the context instead.
* Added four new integration tests, two that test adding a new signer and changing the threshold, and two for just changing the threshold.

## Testing Information

The two integration tests take quite some time to run. Perhaps `testcontainers` to the rescue?

## Checklist

### DIFF
--- a/signer/src/testing/btc.rs
+++ b/signer/src/testing/btc.rs
@@ -87,7 +87,10 @@ pub async fn new_zmq_block_hash_stream(endpoint: &str) -> ReceiverStream<Result<
     tokio::spawn(async move {
         let mut stream = zmq_stream.to_block_hash_stream();
         while let Some(block) = stream.next().await {
-            sender.send(block).await.unwrap();
+            if let Err(error) = sender.send(block).await {
+                tracing::warn!(%error, "could not send bitcoin block hash from zmq stream; exiting");
+                break;
+            }
         }
     });
 
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -158,8 +158,6 @@ pub struct TxCoordinatorEventLoop<Context, Network> {
     pub network: Network,
     /// Private key of the coordinator for network communication.
     pub private_key: PrivateKey,
-    /// the number of signatures required.
-    pub threshold: u16,
     /// How many bitcoin blocks back from the chain tip the signer will
     /// look for requests.
     pub context_window: u16,
@@ -535,6 +533,7 @@ where
 
         // Create a signal stream with the defined filter
         let signal_stream = self.context.as_signal_stream(presign_ack_filter);
+        let signature_threshold = self.context.config().signer.bootstrap_signatures_required;
 
         // Send the presign request message
         self.send_message(sbtc_requests, bitcoin_chain_tip).await?;
@@ -544,7 +543,7 @@ where
             let target_tip = *bitcoin_chain_tip;
             let mut acknowledged_signers = HashSet::new();
 
-            while acknowledged_signers.len() < self.threshold as usize {
+            while acknowledged_signers.len() < signature_threshold as usize {
                 match signal_stream.next().await {
                     None => {
                         tracing::warn!("signer signal stream closed unexpectedly, shutting down");
@@ -1637,8 +1636,9 @@ where
         tracing::info!("Coordinating DKG");
         // Get the current signer set for running DKG.
         let signer_set = self.context.config().signer.bootstrap_signing_set.clone();
+        let threshold = self.context.config().signer.bootstrap_signatures_required;
 
-        let mut state_machine = FireCoordinator::new(signer_set, self.threshold, self.private_key);
+        let mut state_machine = FireCoordinator::new(signer_set, threshold, self.private_key);
 
         // Okay let's move the coordinator state machine to the beginning
         // of the DKG phase.
@@ -2137,13 +2137,14 @@ where
 
         // Get the current sBTC limits (caps).
         let sbtc_limits = self.context.state().get_current_limits();
+        let signature_threshold = config.signer.bootstrap_signatures_required;
 
         // Setup the parameters for fetching pending requests.
         let params = GetPendingRequestsParams {
             bitcoin_chain_tip,
             stacks_chain_tip,
             aggregate_key,
-            signature_threshold: self.threshold,
+            signature_threshold,
             sbtc_limits: &sbtc_limits,
         };
 
@@ -2186,7 +2187,7 @@ where
             deposits,
             withdrawals,
             signer_state,
-            accept_threshold: self.threshold,
+            accept_threshold: signature_threshold,
             num_signers,
             sbtc_limits,
             max_deposits_per_bitcoin_tx,
@@ -2681,6 +2682,9 @@ mod tests {
         let context = TestContext::builder()
             .with_in_memory_storage()
             .with_mocked_clients()
+            .modify_settings(|settings| {
+                settings.signer.bootstrap_signatures_required = 3;
+            })
             .build();
 
         // TODO: fix tech debt #893 then raise threshold to 5
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -2692,7 +2692,6 @@ mod tests {
             context,
             context_window: 5,
             num_signers: 7,
-            signing_threshold: 3,
             test_model_parameters,
         }
     }
--- a/signer/tests/integration/setup.rs
+++ b/signer/tests/integration/setup.rs
@@ -1,6 +1,5 @@
 use std::collections::BTreeMap;
 use std::collections::HashSet;
-use std::ops::Deref;
 
 use bitcoin::AddressType;
 use bitcoin::Amount;
@@ -43,7 +42,6 @@ use signer::keys::PublicKey;
 use signer::keys::SignerScriptPubKey;
 use signer::stacks::api::MockStacksInteract;
 use signer::stacks::wallet::SignerWallet;
-use signer::storage::DbRead;
 use signer::storage::DbWrite as _;
 use signer::storage::model;
 use signer::storage::model::BitcoinBlock;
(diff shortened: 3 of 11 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -535,6 +533,7 @@ where
 
         // Create a signal stream with the defined filter
         let signal_stream = self.context.as_signal_stream(presign_ack_filter);
+        let signature_threshold = self.context.config().signer.bootstrap_signatures_required;
```
matteojug, line 551: Can we add a comment that this may not be the "correct" threshold, but we don't really care since this is already a best effort check? eg, if we still need to rotate to the new threshold, we may check against the new (still unused) threshold. We should use something like the max of the thresholds of the DKG shares required to sign the inputs of the package, but assuming we will not do unreasonable changes to the threshold the config one should be fine.

signer/src/transaction_coordinator.rs
```diff
@@ -2137,13 +2137,14 @@ where
 
         // Get the current sBTC limits (caps).
         let sbtc_limits = self.context.state().get_current_limits();
+        let signature_threshold = config.signer.bootstrap_signatures_required;
```
matteojug, line 2148: As with the presign one, this may not be the correct one: it should probably a mix of each deposit aggregate key shares threshold and the registry (or maybe the config?) one for the stacks side. I'm not sure if we can safely roll with the config one in this case, eg if we lower the threshold (say, from 11 to 10) and we have a deposit with 10 for and 4 against spent to the old aggregate key (the one with 11 as threshold), using the new threshold we may try to include it, but we would never manage to sign for it. And similarly, if we increase the threshold we may cut off valid deposits using spent to a lower threshold aggregate key.

signer/tests/integration/setup.rs
```diff
@@ -457,9 +455,6 @@ pub async fn backfill_bitcoin_blocks(db: &PgStore, rpc: &Client, chain_tip: &bit
         db.write_bitcoin_block(&bitcoin_block).await.unwrap();
         block_header = rpc.get_block_header_info(&parent_header_hash).unwrap();
     }
-
-    let block_hash = db.get_bitcoin_canonical_chain_tip().await.unwrap().unwrap();
-    assert_eq!(block_hash.deref(), chain_tip);
```
matteojug, line 462: qq: why this deletion? If we are backfilling in a test from a tip and end up with a fork as canonical chain tip, there's probably something strange with the setup (or we risk having gaps in the chain, if we have no forks but backfilled from not a chain tip but we have a different tip in the db).

signer/src/testing/btc.rs
```diff
@@ -87,7 +87,10 @@ pub async fn new_zmq_block_hash_stream(endpoint: &str) -> ReceiverStream<Result<
     tokio::spawn(async move {
         let mut stream = zmq_stream.to_block_hash_stream();
         while let Some(block) = stream.next().await {
-            sender.send(block).await.unwrap();
+            if let Err(error) = sender.send(block).await {
+                tracing::warn!(%error, "could not send bitcoin block hash from zmq stream; exiting");
```
MCJOHN974: nit: should it be error level log?

signer/src/transaction_coordinator.rs
djordon: Yeah, will add.

signer/src/transaction_coordinator.rs
djordon: Yeah, all good points. This issue is quite complicated because it's not clear that we do the computations optimally at the packaging level (I think we do it incorrectly). We have two issues there:
1. We treat "I vote against" the same as "I cannot sign for this even if I wanted to" at the packaging level, which can lead to deposits falling through the cracks here and there. So even if we used the correct threshold, we count votes incorrectly as well.
2. We do not distinguish between the different thresholds for each of the inputs.

signer/tests/integration/setup.rs
djordon: I think the issue is that this function doesn't care about whether the chain is canonical or whether it is actually at the tip of any particular block chain. This function backfills from the given block hash to some minimum height, it doesn't really care about anything else. Since the given block hash need not be the actual chain tip,  the chain tip in the database does not need to match it, and that is the problem with the check. 

So, I should probably move the check to the `fetch_canonical_bitcoin_blockchain` helper, but it shouldn't be here. I should also rename the variable, I forgot to come back and do that.

matteojug: In general I agree -- this function does the backfill from a block, why should it care about what's in the db -- but without the check we risk ending up with gaps in the db view of the chain, which I think could be more annoying than just fixing the test setup to avoid triggering this assert. But I may be biased because I didn't encounter any test scenario where this check was triggering, so I may be missing some good reasons to remove this (other than the general one).

djordon: Hmmm, yeah it might be easier to fix the test. But I still feel like the existing check should change, just that the function should care whether the blockchain identified by the given block hash is in the database instead.

The specific scenario happens in the new test that onboards a new signer. I want the new signer to do the normal bootstrapping through the block observer. In this case I back fill the bitcoin blocks until before the existing signer set has come online so that the new signer doesn't fetch useless blocks in the block observer during the test (I need to add a comment about that). However, I do this all sloppily and call `backfill_bitcoin_blocks` for all signers in a loop, not just the new signer. When I call `backfill_bitcoin_blocks` for an existing signer who is all caught up, it used to panic because the check doesn't expect the chain tip in the signer's database to have advanced past the given input, which was the case for the existing signers. 

In general though I think that this function should panic if the database doesn't have the block headers, so maybe the check should be that the expected block headers are in the database? I think that will address all of our concerns.

signer/src/testing/btc.rs
djordon: I've only triggered this in a situation where it is not actually an error. I think the receivers were all correctly dropped so sending would correctly panic. Making this an ERROR level log seemed incorrect, it might be fine, but I made it a warn in case it is an error in some future test.

signer/src/transaction_coordinator.rs
djordon: I've created https://github.com/stacks-sbtc/sbtc/issues/1767 for this issue.

signer/tests/integration/setup.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1765/commits/265c8278316039067ac5d1d862f9934c2c49072f. Let me know what you think.

djordon: Oh I forgot to change some things, one sec.

signer/src/transaction_coordinator.rs
djordon: Added a comment in https://github.com/stacks-sbtc/sbtc/pull/1765/commits/4d11caf815de93f3519c1ee17e8c86f9d8c27f4b

signer/tests/integration/setup.rs
djordon: Oh apparently I made one of the changes that I was thinking about. I just didn't check that the headers were in the database, and instead just relied on our tests of the `write_bitcoin_block` function and skipped the check (at least that's what I remember thinking when I made the commit). I don't mind adding a header check though.

matteojug: Resolving as per offline chat: the current approach seems fine and to trigger the situation I was scared of we need to mess things pretty badly in the test setup, so it was probably a non issue in the first place.

matteojug (approved):

matteojug (approved):

signer/src/transaction_coordinator.rs
```diff
@@ -535,6 +533,13 @@ where
 
         // Create a signal stream with the defined filter
         let signal_stream = self.context.as_signal_stream(presign_ack_filter);
+        // This value may not be the "correct" threshold to use. The
+        // threshold should be the max of all thresholds for the inputs in
+        // the sweep transaction package. Since the signer set is currently
+        // stable, using this value won't cause any issues. However we can
+        // have a bug here if we open up the signer set and allow a large
+        // increase in the signatures required parameter.
```
cylewitruk, line 550: We could also defensively bail with an informative message if a generated presign request happens to contain any inputs whose associated shares' `threshold` is >= the number of signers in the current signing set (shrunken set case), otherwise we'd just do unnecessary work and time out below with "not enough signatures" as `acknowledged_signers` would never be >= `signature_threshold`.

Shouldn't happen if packaging is fixed to properly account for differing thresholds, though (just saw https://github.com/stacks-sbtc/sbtc/issues/1767).

cylewitruk (approved):
Looks fine, though yeah there's still a little bit of work to make threshold changes truly "production ready".

signer/src/transaction_coordinator.rs
djordon: Yeah, I figure that fix is better left for a separate PR, such as one that targets https://github.com/stacks-sbtc/sbtc/issues/1777

matteojug (dismissed):

MCJOHN974 (approved):

signer/src/transaction_coordinator.rs
djordon: Resolving, since we'll finalizing the handling of the threshold later.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -291,7 +304,37 @@ async fn wait_for_signers<S>(
...
+    let coordinator_ctx = signers
+        .iter()
+        .find(|ctx| {
+            let public_key = ctx.config().signer.public_key();
+            given_key_is_coordinator(public_key, chain_tip, &signer_public_keys)
+        })
+        .unwrap();
+
+    let expected_signal = TxCoordinatorEvent::TenureCompleted.into();
+    let wait_duration = Duration::from_secs(10);
```
MCJOHN974, line 334: Isn't 10 too much here?

djordon: I think I originally wanted to choose the same number that we used in `wait_for_signers`, and we use 15 seconds there. I don't know why I went with 10 and not 15. They both could probably be reduced to 5, but I'm not sure. I just didn't want to choose a number that leads to false positives.

djordon:
Closing this, since it's such a large and old PR. I'm going to PR these changes but in pieces, with https://github.com/stacks-sbtc/sbtc/pull/1893 and https://github.com/stacks-sbtc/sbtc/pull/1892 already opened tackling the low hanging fruit.

PR #1766, opened 2025-06-27
refactor: extract common pieces of several integration tests

djordon:
## Description

This is a refactor that aims to simplify many of our integrations tests in `signer/tests/integration/transaction_coordinator.rs`.

## Changes

* Add a function for creating an `EmilyClient` for our integration tests.
* Add a function that starts our four event loops that can talk to each other as if they are running in the same binary.
* Simplify the `signers` variable into just a vector of contexts.
* Clean up some of the comments in our tests, since sometimes they no longer apply.

## Testing Information

This should be entirely test changes.

## Checklist

### DIFF
--- a/signer/tests/integration/transaction_coordinator.rs
+++ b/signer/tests/integration/transaction_coordinator.rs
@@ -1626,12 +1626,7 @@ async fn pseudo_random_dkg() {
     let (rpc, faucet) = regtest::initialize_blockchain();
 
     // We need to populate our databases, so let's fetch the data.
-    let emily_client = EmilyClient::try_new(
-        &Url::parse("http://testApiKey@localhost:3031").unwrap(),
-        Duration::from_secs(1),
-        None,
-    )
-    .unwrap();
+    let emily_client = EmilyClient::new_test_client();
 
     testing_api::wipe_databases(&emily_client.config().as_testing())
         .await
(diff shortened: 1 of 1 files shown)


signer/src/emily_client.rs
```diff
@@ -195,6 +195,17 @@ impl EmilyClient {
                 .map_err(Error::DecodeHexScript)?,
         })
     }
+
+    /// Create a new Emily Client for use in our integration tests.
+    #[cfg(any(test, feature = "testing"))]
+    pub fn new_text_client() -> Self {
```
cylewitruk: - `new_test_client`. 
- I personally prefer to put these in a separate `impl` block with the `cfg` attribute just to keep test utility methods together and out of the way of the main impls. No strong feelings here though.

djordon: Good points, will update.

djordon: Addressed in https://github.com/stacks-sbtc/sbtc/pull/1766/commits/f9e2406e9a817350f3b6f926a748f61071a96814 and https://github.com/stacks-sbtc/sbtc/pull/1766/commits/291d77a7fa1b6fd068babb57e857d70ec4abaddf

MCJOHN974 (approved):

Issue #1721, opened 2025-06-10
[Bug]: Don't verify already verified shares before submitting a key rotation


matteojug:
# Bug - Don't verify already verified shares before submitting a key rotation

## 1. Description

Currently if we need to issue a rotate key rotation, we also always verify the shares, even if those are already verified. This was to ensure that if some signer failed to get the previous verification (eg, unfortunate restart), it will get a new chance to get them verified.

OTOH, the current behaviour could waste time in verifying something that everybody already verified, or if we pass the 10 blocks verification window since the shares creation we would not be able to verify them again, blocking the key rotation. So it seems we may want to revert this behaviour and skip verification if shares are already verified.

PR #1768, opened 2025-06-30
Remove Verification If Past Verification Window (draft)

setzeus:
## Description
Currently, during a key rotation, we verify shares *twice.* Here, if we're past the verification window we forgo verifying again since it'd fail & further delay rotating to the new key.

Closes: #1721 

## Changes
signer/src/transaction_coordinator.rs, specifically the 'assert_rotate_key_action.'

## Testing Information
Three new tests were added.

## Checklist

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -351,6 +351,12 @@ where
                 }
             }
         } else {
+            // If we do not have signer set info in the registry, then we
+            // are in the bootstrap phase. Our latest DKG shares may be
+            // 'Unverified', but if we are here then they were not 'Failed'
+            // when we made to call to `should_coordinate_dkg`, since we
+            // will coordinate DKG if our last DKG shares are 'Failed'. But
+            // we could be loading 'Failed' shares here.
             registry_aggregate_key
                 .ok_or(Error::MissingAggregateKey(*bitcoin_chain_tip.block_hash))?
         };
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -316,7 +316,11 @@ where
         );
         span.record("bitcoin_tip_height", *bitcoin_chain_tip.block_height);
 
-        let registry_signer_set_info = self.context.state().registry_signer_set_info();
+        let registry_aggregate_key = self
+            .context
+            .state()
+            .registry_signer_set_info()
+            .map(|info| info.aggregate_key);
 
         // If we are not the coordinator, then we have no business
         // coordinating DKG or constructing bitcoin and stacks
@@ -342,29 +346,13 @@ where
                 Ok(key) => key,
                 Err(error) => {
                     tracing::error!(%error, "failed to coordinate DKG; using existing aggregate key");
-                    registry_signer_set_info
-                        .as_ref()
-                        .map(|info| info.aggregate_key)
+                    registry_aggregate_key
                         .ok_or(Error::MissingAggregateKey(*bitcoin_chain_tip.block_hash))?
                 }
             }
         } else {
-            // If we do not have signer set info in the registry, then we
-            // are in the bootstrap phase. Our latest DKG shares may be
-            // 'Unverified', but if we are here then they were not 'Failed'
-            // when we made to call to `should_coordinate_dkg`, since we
-            // will coordinate DKG if our last DKG shares are 'Failed'. But
-            // we could be loading 'Failed' shares here.
-            match registry_signer_set_info.as_ref() {
-                Some(info) => info.aggregate_key,
-                None => self
-                    .context
-                    .get_storage()
-                    .get_latest_encrypted_dkg_shares()
-                    .await?
-                    .map(|shares| shares.aggregate_key)
-                    .ok_or(Error::NoDkgShares)?,
-            }
+            registry_aggregate_key
+                .ok_or(Error::MissingAggregateKey(*bitcoin_chain_tip.block_hash))?
         };
 
         let chain_tip_hash = &bitcoin_chain_tip.block_hash;
@@ -390,7 +378,10 @@ where
             return Ok(());
         }
 
-        let signer_public_keys = registry_signer_set_info
+        let signer_public_keys = self
+            .context
+            .state()
+            .registry_signer_set_info()
             .map(|info| info.signer_set)
             .ok_or_else(|| Error::NoKeyRotationEvent)?;
 
@@ -446,7 +437,7 @@ where
             .await?;
 
         let (needs_verification, needs_rotate_key) =
-            assert_rotate_key_action(&last_dkg, current_aggregate_key)?;
+            assert_rotate_key_action(&self.context, &last_dkg, current_aggregate_key).await?;
         if !needs_verification && !needs_rotate_key {
             tracing::debug!(
                 "stacks node is up to date with the current aggregate key and no DKG verification required"
@@ -2606,14 +2597,33 @@ pub async fn should_coordinate_dkg(
 
 /// Assert, given the last dkg and smart contract current aggregate key, if we
 /// need to verify the shares and/or issue a rotate key call.
-pub fn assert_rotate_key_action(
+pub async fn assert_rotate_key_action<C>(
+    context: &C,
     last_dkg: &model::EncryptedDkgShares,
     current_aggregate_key: Option<PublicKey>,
-) -> Result<(bool, bool), Error> {
+) -> Result<(bool, bool), Error>
+where
+    C: Context,
+{
     let needs_rotate_key = Some(last_dkg.aggregate_key) != current_aggregate_key;
 
+    // Check if past verification window, if we are, skip verification
+    let dkg_verification_window = context.config().signer.dkg_verification_window;
+    let max_verification_height = last_dkg
+        .started_at_bitcoin_block_height
+        .saturating_add(dkg_verification_window as u64);
+
+    // Get the current bitcoin chain tip to check against the verification window
+    let current_bitcoin_chain_tip = context
+        .get_storage()
+        .get_bitcoin_canonical_chain_tip_ref()
+        .await?
+        .ok_or(Error::NoChainTip)?;
+
+    let past_verification_window = current_bitcoin_chain_tip.block_height > max_verification_height;
+
     let needs_verification = match last_dkg.dkg_shares_status {
-        model::DkgSharesStatus::Unverified => true,
+        model::DkgSharesStatus::Unverified => !past_verification_window,
         model::DkgSharesStatus::Verified => needs_rotate_key,
         model::DkgSharesStatus::Failed => {
             return Err(Error::DkgVerificationFailed(last_dkg.aggregate_key.into()));
@@ -2856,7 +2866,7 @@ mod tests {
             current_aggregate_key_seed: None,
             needs_verification: true,
             needs_rotate_key: true,
-        }; "unverified, no key")]
+        }; "unverified, no key, within window")]
     #[test_case(
         RotateKeyActionTest {
             shares_status: model::DkgSharesStatus::Verified,
@@ -2872,7 +2882,7 @@ mod tests {
             current_aggregate_key_seed: Some(1),
             needs_verification: true,
             needs_rotate_key: false,
-        }; "unverified, key up to date")]
+        }; "unverified, key up to date, within window")]
     #[test_case(
         RotateKeyActionTest {
             shares_status: model::DkgSharesStatus::Verified,
@@ -2888,7 +2898,7 @@ mod tests {
             current_aggregate_key_seed: Some(1),
             needs_verification: true,
             needs_rotate_key: true,
-        }; "unverified, new key")]
+        }; "unverified, new key, within window")]
     #[test_case(
         RotateKeyActionTest {
             shares_status: model::DkgSharesStatus::Verified,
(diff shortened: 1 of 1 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -355,16 +357,8 @@ where
...
-                None => self
-                    .context
-                    .get_storage()
-                    .get_latest_encrypted_dkg_shares()
-                    .await?
-                    .map(|shares| shares.aggregate_key)
-                    .ok_or(Error::NoDkgShares)?,
-            }
+            registry_aggregate_key
+                .ok_or(Error::MissingAggregateKey(*bitcoin_chain_tip.block_hash))?
```
djordon: ? This seems accidental. I think you need to merge in main.

signer/src/transaction_coordinator.rs
```diff
@@ -2606,14 +2603,33 @@ pub async fn should_coordinate_dkg(
...
+    let max_verification_height = last_dkg
+        .started_at_bitcoin_block_height
+        .saturating_add(dkg_verification_window as u64);
+
+    // Get the current bitcoin chain tip to check against the verification window
+    let current_bitcoin_chain_tip = context
+        .get_storage()
+        .get_bitcoin_canonical_chain_tip_ref()
+        .await?
+        .ok_or(Error::NoChainTip)?;
```
djordon: We should pass in the chain tip here instead of fetching it from the database.

signer/src/transaction_coordinator.rs
```diff
@@ -2606,14 +2603,33 @@ pub async fn should_coordinate_dkg(
...
+
+    let past_verification_window = current_bitcoin_chain_tip.block_height > max_verification_height;
+
     let needs_verification = match last_dkg.dkg_shares_status {
-        model::DkgSharesStatus::Unverified => true,
+        model::DkgSharesStatus::Unverified => !past_verification_window,
```
djordon: Ahh, so this state is possible right now, but unlikely since the block observer marks unverified shares as failed after the verification window. The edge case has to do with things being out of sync sometimes. If we do some of the changes proposed in https://github.com/stacks-sbtc/sbtc/pull/1730, and pass in the bitcoin block here then I think that this state would be impossible.

Is there some other case where we can get to this kind of state with unverified shares past the verification window? cc @matteojug

matteojug: Yeah if we pass the chain tip (and it's the one passed by the block observer) it shouldn't be possible: we could still get failed shares in here because of timing, but since shares cannot go from non-unverified to unverified and we are processing according to (an eventual past) block observer view of things I don't see how we could end up with unverified shares past the verification window. I would revert this back to `true`.

signer/src/transaction_coordinator.rs
```diff
@@ -2606,14 +2603,33 @@ pub async fn should_coordinate_dkg(
...
         model::DkgSharesStatus::Verified => needs_rotate_key,
```
matteojug: This (`model::DkgSharesStatus::Verified`) is the state where we want to avoid verifying if `past_verification_window`

signer/src/transaction_coordinator.rs
```diff
@@ -390,7 +390,10 @@ where
             return Ok(());
         }
 
-        let signer_public_keys = registry_signer_set_info
+        let signer_public_keys = self
+            .context
+            .state()
+            .registry_signer_set_info()
```
matteojug: This also seems a merge leftover -- is there a reason we don't want to reuse `registry_signer_set_info`?

signer/src/transaction_coordinator.rs
setzeus: Copy, refactored & passed in.

signer/src/transaction_coordinator.rs
setzeus: Copy, removed.

signer/src/transaction_coordinator.rs
```diff
@@ -445,8 +445,21 @@ where
...
 
-        let (needs_verification, needs_rotate_key) =
-            assert_rotate_key_action(&last_dkg, current_aggregate_key)?;
+        // Get the bitcoin chain tip ref from the database to pass to assert_rotate_key_action
+        let bitcoin_chain_tip_ref = self
+            .context
+            .get_storage()
+            .get_bitcoin_canonical_chain_tip_ref()
+            .await?
+            .ok_or(Error::NoChainTip)?;
```
djordon: Instead of making the database query here, change the function signature of `check_and_submit_rotate_key_transaction` to require a `BitcoinBlockRef` type for the `bitcoin_chain_tip` and use that in the call to `assert_rotate_key_action`.

signer/src/transaction_coordinator.rs
```diff
@@ -2606,15 +2619,28 @@ pub async fn should_coordinate_dkg(
 
 /// Assert, given the last dkg and smart contract current aggregate key, if we
 /// need to verify the shares and/or issue a rotate key call.
-pub fn assert_rotate_key_action(
+pub async fn assert_rotate_key_action<C>(
```
djordon: I don't think this function needs to be `async` anymore.

signer/src/transaction_coordinator.rs
```diff
@@ -2976,17 +2980,28 @@ mod tests {
         } else {
             101u64
         };
+        let bitcoin_chain_tip = model::BitcoinBlockHash::from(Faker.fake::<[u8; 32]>());
```
matteojug: very nit, also below
```suggestion
        let bitcoin_chain_tip: model::BitcoinBlockHash = Faker.fake();
```

signer/src/transaction_coordinator.rs
```diff
@@ -2897,33 +2923,130 @@ mod tests {
...
 
-        let (needs_verification, needs_rotate_key) =
-            assert_rotate_key_action(&last_dkg, current_aggregate_key).unwrap();
+        // Write a bitcoin block at the appropriate height to simulate the current chain tip
+        let storage = context.get_storage_mut();
+        let chain_tip_height = if scenario.needs_verification {
+            100u64
+        } else {
+            101u64
+        };
```
matteojug: nit: consider moving this together with the `started_at_bitcoin_block_height` above for better setup grouping. Or we could just remove the conditional value and just use a constant for this (eg, 100), given we are already setting different values for `started_at_bitcoin_block_height`

signer/src/transaction_coordinator.rs
```diff
@@ -2897,33 +2923,130 @@ mod tests {
...
+        let bitcoin_chain_tip = model::BitcoinBlockHash::from(Faker.fake::<[u8; 32]>());
+        storage
+            .write_bitcoin_block(&model::BitcoinBlock {
+                block_hash: bitcoin_chain_tip,
+                block_height: chain_tip_height.into(),
+                parent_hash: Faker.fake(),
+            })
+            .await
+            .unwrap();
```
matteojug: nit: we shouldn't need this anymore, also in `test_assert_rotate_key_action_failure` and `test_assert_rotate_key_action_verification_window_elapsed`

signer/src/transaction_coordinator.rs
```diff
@@ -2897,33 +2923,130 @@ mod tests {
...
+
+        // Set the DKG start height based on whether we expect verification or not
+        // If we expect verification (needs_verification = true), set within window
+        // If we don't expect verification (needs_verification = false), set past window
+        if scenario.needs_verification {
+            // Set within verification window (10 blocks before current height of 100)
+            last_dkg.started_at_bitcoin_block_height = 90u64.into();
+        } else {
+            // Set past verification window (21 blocks before current height of 101)
+            last_dkg.started_at_bitcoin_block_height = 80u64.into();
```
matteojug: nit: can we check 11 blocks instead? To ensure the check is as tight as expected

signer/src/transaction_coordinator.rs
setzeus: Copy, updated in [988acf5](https://github.com/stacks-sbtc/sbtc/pull/1768/commits/988acf5f2cb6d17086d23e445427565b42a6c47f)

signer/src/transaction_coordinator.rs
matteojug: there are still some occurrences of this

signer/src/transaction_coordinator.rs
matteojug: sorry I meant checking 11 blocks before the current height, so that if `needs_verification` we check the largest okay value, and if not we check the smallest not okay value

setzeus: Addressed in [8ac80fe](https://github.com/stacks-sbtc/sbtc/pull/1768/commits/8ac80fe02ca6ab3574e2194d9b56c733848a0602)

signer/src/transaction_coordinator.rs
setzeus: Addressed in [8ac80fe](https://github.com/stacks-sbtc/sbtc/pull/1768/commits/8ac80fe02ca6ab3574e2194d9b56c733848a0602)

signer/src/transaction_coordinator.rs
setzeus: Addressed in [8ac80fe](https://github.com/stacks-sbtc/sbtc/pull/1768/commits/8ac80fe02ca6ab3574e2194d9b56c733848a0602)

signer/src/transaction_coordinator.rs
```diff
@@ -375,8 +375,12 @@ where
         self.deploy_smart_contracts(chain_tip_hash, &wallet, &aggregate_key)
             .await?;
 
-        let rotate_key_txid =
-            self.check_and_submit_rotate_key_transaction(chain_tip_hash, &wallet, &aggregate_key);
+        // Pass the full bitcoin_chain_tip ref
```
matteojug: nit: can we remove both this and `Use the passed-in bitcoin_chain_tip directly` comment below?

signer/src/transaction_coordinator.rs
```diff
@@ -2933,4 +3028,39 @@ mod tests {
...
+        let context = TestContext::builder()
+            .with_in_memory_storage()
+            .with_mocked_clients()
+            .build();
+
+        let mut last_dkg: model::EncryptedDkgShares = Faker.fake();
+        last_dkg.dkg_shares_status = model::DkgSharesStatus::Unverified;
+        last_dkg.aggregate_key = public_key_from_seed(1);
+        // Set the DKG start height to be outside the verification window
+        last_dkg.started_at_bitcoin_block_height = 79u64.into(); // 11 blocks before current height of 100
```
matteojug: nit, or update comment
```suggestion
        last_dkg.started_at_bitcoin_block_height = 89u64.into(); // 11 blocks before current height of 100
```

signer/src/transaction_coordinator.rs
```diff
@@ -2606,15 +2615,28 @@ pub async fn should_coordinate_dkg(
...
+        .started_at_bitcoin_block_height
+        .saturating_add(dkg_verification_window as u64);
+
+    let past_verification_window = bitcoin_chain_tip.block_height > max_verification_height;
+
     let needs_verification = match last_dkg.dkg_shares_status {
-        model::DkgSharesStatus::Unverified => true,
-        model::DkgSharesStatus::Verified => needs_rotate_key,
+        model::DkgSharesStatus::Unverified => !past_verification_window,
+        model::DkgSharesStatus::Verified => needs_rotate_key && !past_verification_window,
```
djordon: I think we should do something similar for the `needs_rotate_key` logic, where we say it's false if the shares are `Unverified` and we are past the max verification height.

signer/src/transaction_coordinator.rs
```diff
@@ -2606,15 +2613,37 @@ pub async fn should_coordinate_dkg(
...
-        model::DkgSharesStatus::Unverified => true,
-        model::DkgSharesStatus::Verified => needs_rotate_key,
+        model::DkgSharesStatus::Unverified => !past_verification_window,
+        model::DkgSharesStatus::Verified => base_needs_rotate_key && !past_verification_window,
+        model::DkgSharesStatus::Failed => {
+            return Err(Error::DkgVerificationFailed(last_dkg.aggregate_key.into()));
+        }
+    };
+
+    // Similar logic to needs_rotate_key: if shares are Unverified *&* past verification window, don't rotate
```
djordon:
```suggestion
    // Similar logic to needs_rotate_key: if shares are Unverified and 
    // past the verification window, don't attempt to submit a key
    // rotation transaction.
```

signer/src/transaction_coordinator.rs
```diff
@@ -2933,4 +3035,39 @@ mod tests {
...
+        let bitcoin_chain_tip: model::BitcoinBlockHash = Faker.fake();
+
+        let bitcoin_chain_tip_ref = model::BitcoinBlockRef {
+            block_hash: bitcoin_chain_tip,
+            block_height: 100u64.into(),
+        };
+
+        let result = assert_rotate_key_action(&context, &last_dkg, None, &bitcoin_chain_tip_ref);
+
+        // Now we expect success: no verification needed (false), and no rotation needed (false) since past verification window
```
djordon:
```suggestion
        // Now we expect success: neither verification nor key
        // rotation are needed since we are past the verification
        // window of 10 bitcoin blocks
```

signer/src/transaction_coordinator.rs
```diff
@@ -2897,33 +2926,106 @@ mod tests {
...
+            current_aggregate_key_seed: Some(1),
+            needs_verification: false,
+            needs_rotate_key: false,
+        }; "unverified, new key, past window")]
+    #[tokio::test]
+    async fn test_assert_rotate_key_action(scenario: RotateKeyActionTest) -> Result<(), Error> {
+        let context = TestContext::builder()
+            .with_in_memory_storage()
+            .with_mocked_clients()
+            .build();
```
djordon, line 2969: One thing that we should either explicitly set in these tests or comment about in each test is the value of the `dkg_verification_window`.

signer/src/transaction_coordinator.rs
setzeus: Addressed in [94957b7](https://github.com/stacks-sbtc/sbtc/pull/1768/commits/94957b78eea3d74ca97105a8958e917f4f025325)

signer/src/transaction_coordinator.rs
setzeus: Addressed in [94957b7](https://github.com/stacks-sbtc/sbtc/pull/1768/commits/94957b78eea3d74ca97105a8958e917f4f025325)

signer/src/transaction_coordinator.rs
setzeus: Addressed in [94957b7](https://github.com/stacks-sbtc/sbtc/pull/1768/commits/94957b78eea3d74ca97105a8958e917f4f025325)

djordon (approved):
✅

PR #1769, opened 2025-07-02
feat: add metrics for smart contract read requests

djordon:
## Changes

* Adds metrics around requests to the stacks node that call read-only functions, read data variables, and read map entries.
* Changed the function signatures to take ownership of the string types, since it should be slightly more efficient.

## Testing Information

~~I still need to check that this works on devenv.~~ Ran this branch on devenv and everything looked fine. The only thing that is not tested is the new `Metrics::record_data_var`, because we no longer read data variables after https://github.com/stacks-sbtc/sbtc/pull/1734. Also, we can probably remove that function, as well as remove `get_current_signers_aggregate_key`. Let's do that another day.

## Checklist

### DIFF
--- a/signer/src/stacks/api.rs
+++ b/signer/src/stacks/api.rs
@@ -768,8 +768,8 @@ impl StacksClient {
     pub async fn get_map_entry(
         &self,
         contract_principal: &StacksAddress,
-        contract_name: &ContractName,
-        map_name: &ClarityName,
+        contract_name: ContractName,
+        map_name: ClarityName,
         map_entry: &Value,
     ) -> Result<Option<Value>, Error> {
         let path = format!("/v2/map_entry/{contract_principal}/{contract_name}/{map_name}?proof=0");
@@ -791,6 +791,7 @@ impl StacksClient {
             .map_err(Box::new)
             .map_err(Error::ClarityValueSerialization)?;
 
+        let instant = Instant::now();
         let response = self
             .client
             .post(url)
@@ -800,6 +801,7 @@ impl StacksClient {
             .await
             .map_err(Error::StacksNodeRequest)?;
 
+        Metrics::record_map_entry_duration(instant.elapsed(), contract_name, map_name, &response);
         // It looks like the stacks node returns a 404 if the data is not
         // available, see
         // https://github.com/stacks-network/stacks-core/blob/c1a1f50fddcbc11054fae537103423e21221665a/stackslib/src/net/api/getmapentry.rs#L223-L225C22
@@ -1492,7 +1494,7 @@ impl StacksInteract for StacksClient {
 
         let map_entry = Value::UInt(request_id as u128);
         let result = self
-            .get_map_entry(deployer, &contract_name, &map_name, &map_entry)
+            .get_map_entry(deployer, contract_name, map_name, &map_entry)
             .await?;
 
         // This map `withdrawal-status` in the smart contract stores
--- a/signer/src/stacks/api.rs
+++ b/signer/src/stacks/api.rs
@@ -97,29 +97,9 @@ const DUMMY_STX_TRANSFER_PAYLOAD: TransactionPayload = TransactionPayload::Token
 /// The names of all the read-only functions used in the signers for any of
 /// the sbtc smart contracts.
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
-pub struct ReadOnlyFnName(pub &'static str);
+pub struct ClarityName(pub &'static str);
 
-impl std::fmt::Display for ReadOnlyFnName {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
-    }
-}
-
-/// A wrapper around the contract name string.
-#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
-pub struct ClarityMapName(pub &'static str);
-
-impl std::fmt::Display for ClarityMapName {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
-    }
-}
-
-/// A wrapper around the contract name string.
-#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
-pub struct DataVarName(pub &'static str);
-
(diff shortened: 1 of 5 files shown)


signer/src/stacks/api.rs
```diff
@@ -80,6 +74,47 @@ const DUMMY_STX_TRANSFER_PAYLOAD: TransactionPayload = TransactionPayload::Token
     TokenTransferMemo([0; 34]),
 );
 
+/// The names of all the read-only functions used in the signers for any of
+/// the sbtc smart contracts.
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, strum::Display, strum::IntoStaticStr)]
+#[strum(serialize_all = "kebab-case")]
```
matteojug: While functionally equivalent, I'm wondering if we should opt to use an explicit `serialize` (or `to_string`) `strum` attribute on each variant, for a more "obvious" mapping of the variant to the method called; also for `ClarityMapName` and `DataVarName`.

signer/src/stacks/api.rs
```diff
@@ -80,6 +74,47 @@ const DUMMY_STX_TRANSFER_PAYLOAD: TransactionPayload = TransactionPayload::Token
...
+pub enum ReadOnlyFnName {
```
matteojug: It seems a bit strange that we are putting in here function names from different smart contracts, but I don't have an actual argument for not doing it, it just feels a bit off; but I wanted to flag regardless, in case there was a "better" way.

In general I'm having troubles seeing if we are gaining something using a wrapper: we "ensure" you don't mix different `ClarityName`s between fn names/data/maps, but we are forced to declare testing functions (or datavars) in here. Is this really worth it?

signer/src/stacks/api.rs
```diff
@@ -80,6 +74,47 @@ const DUMMY_STX_TRANSFER_PAYLOAD: TransactionPayload = TransactionPayload::Token
...
+    /// This is the name of the read-only function in the sbtc-token smart
+    /// for getting the total supply.
+    GetTotalSupply,
+    /// This is the name of the read-only function in the sbtc-token smart
+    /// contract that returns the balance of a given principal.
+    #[cfg(any(test, feature = "testing"))]
+    GetBalance,
+}
+
+/// A wrapper around the contract name string.
```
matteojug: nit: not a _contract name_, also in `DataVarName`

signer/src/stacks/api.rs
```diff
@@ -80,6 +74,47 @@ const DUMMY_STX_TRANSFER_PAYLOAD: TransactionPayload = TransactionPayload::Token
...
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, strum::Display, strum::IntoStaticStr)]
+#[strum(serialize_all = "kebab-case")]
+pub enum DataVarName {
+    /// The name of the data variable in the sbtc-registry smart contract
+    /// that stores the current aggregate public key of the signers.
+    CurrentAggregatePubkey,
+    /// The name of the data variable in the sbtc-registry smart contract
```
matteojug: nit: missing the end of the comment

signer/src/stacks/contracts.rs
```diff
@@ -1466,7 +1466,7 @@ impl RotateKeysErrorMsg {
 /// A wrapper type for smart contract deployment that implements
 /// AsTxPayload.
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, strum::Display)]
-#[strum(serialize_all = "snake_case")]
+#[strum(serialize_all = "kebab_case")]
```
matteojug: nit: `kebab-case` (according to https://docs.rs/strum/latest/strum/additional_attributes/index.html)

signer/src/stacks/api.rs
djordon: > It seems a bit strange that we are putting in here function names from different smart contracts, but I don't have an actual argument for not doing it, it just feels a bit off; but I wanted to flag regardless, in case there was a "better" way.

It is off, I thought the same thing. One "better" way is to tie the function name to the smart contract. So get ride of the smart contract in the function signature and and have the function names be globally unique. Not an issue for us since our function names are globally unique right now, but might be slightly more complex later.



> In general I'm having troubles seeing if we are gaining something using a wrapper: we "ensure" you don't mix different `ClarityName`s between fn names/data/maps, but we are forced to declare testing functions (or datavars) in here.

My first thought was to change this to `&'static str`s. This is basically what we use now but what we gained from using `ClarityName` is that calling the function is harder to do incorrectly the way that it is. I wanted to preserve that.

> Is this really worth it?

Not sure myself, I've been going back and forth on this. I'd prefer that we use our own types instead of `ClarityName`, since we can implement foreign traits on it (which is what I wanted to do). Well, instead of an enum we can us a type that wraps a `&'static str`, then we don't have any of the issues that we were talking about.

signer/src/stacks/api.rs
djordon: oops

signer/src/stacks/contracts.rs
djordon: I thought I changed that. I guess I just missed it.

Just an FYI, they both work, I noticed this mistake in some other struct and thought that it shouldn't work, but then I found https://github.com/Peternator7/strum/blob/v0.26.3/strum_macros/src/helpers/case_style.rs#L68. But yeah, I did intend to change it, and still do, I just missed it.

signer/src/stacks/api.rs
djordon: Yeah, I agree. I did it like how it is now, changed it to how you're suggesting, and then changed it back. I might have done that two times 😞. I think I'm going to move away from the `enum` though.

signer/src/stacks/contracts.rs
matteojug: I would also suggest to use explicit `serialize`/`to_string` attributes on the variants instead of the enum attribute to make the mapping "obvious" between those and the contract name

djordon: Oh, I just didn't get to this yet. Will do

djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1769/commits/f94520c3c171735d493e95a837cef5950b5b11f9

MCJOHN974 (approved):

Issue #1720, opened 2025-06-10
[Bug]: Avoid recomputing DKG from scratch on key rotation submit failure


matteojug:
# Bug - Avoid recomputing DKG from scratch on key rotation submit failure

## 1. Description

Currently if the bootstrap signer set changes we do a new DKG round and verify them. If then we fail to submit a rotate key (eg, because of a fast block or a timeout) at the next tenure we will again try to DKG again from scratch since the bootstrap signer set is still changed wrt the registry.

Since we have verified shares matching the bootstrap set, we should probably just skip DKG and try to submit a rotate key for those.

PR #1773, opened 2025-07-08
Avoid Rerunning DKG (draft)

setzeus:
## Description
This PR aims to reduce the amount of unnecessary DKGs ran while rotating keys. If there's a rotate key event & we have already verified those shares, then we should skip DKG.

This is currently in draft as a WIP for a sanity check, will add tests shortly.

Closes: #1720 

## Changes
For now the sole change is to the 'should_coordinate_dkg' function found in signer/src/transaction_coordinator.rs.

## Testing Information
No tests for now, in draft as a sanity check.

## Checklist

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -2564,16 +2564,23 @@ pub async fn should_coordinate_dkg(
         // Trigger DKG if signer set changes
         if registry_signer_info.signer_set != config.signer.bootstrap_signing_set {
             // Check if we have verified shares that match the current bootstrap set
-            if let Some(latest_shares) = storage.get_latest_encrypted_dkg_shares().await? {
+            if let Some(latest_shares) = &latest_dkg_shares {
                 if latest_shares.dkg_shares_status == model::DkgSharesStatus::Verified
                     && latest_shares.signer_set_public_keys() == config.signer.bootstrap_signing_set
                 {
-                    // We have verified shares for the current bootstrap set, skip DKG
-                    return Ok(false);
+                    // We have verified shares for the current bootstrap set, but continue
+                    // to check other conditions (like height requirements)
+                    tracing::info!(
+                        "signer set has changed but we have verified shares for current set; checking other conditions"
+                    );
+                } else {
+                    // No verified shares for current set, proceed with DKG
+                    tracing::info!("signer set has changed; proceeding with DKG");
                 }
+            } else {
+                // No shares available, proceed with DKG
+                tracing::info!("signer set has changed; proceeding with DKG");
             }
-            // No verified shares for current set, proceed with DKG
-            return Ok(true);
         }
     }
 
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -2563,7 +2563,16 @@ pub async fn should_coordinate_dkg(
 
         // Trigger DKG if signer set changes
         if registry_signer_info.signer_set != config.signer.bootstrap_signing_set {
-            tracing::info!("signer set has changed; proceeding with DKG");
+            // Check if we have verified shares that match the current bootstrap set
+            if let Some(latest_shares) = storage.get_latest_encrypted_dkg_shares().await? {
+                if latest_shares.dkg_shares_status == model::DkgSharesStatus::Verified
+                    && latest_shares.signer_set_public_keys() == config.signer.bootstrap_signing_set
+                {
+                    // We have verified shares for the current bootstrap set, skip DKG
+                    return Ok(false);
+                }
+            }
+            // No verified shares for current set, proceed with DKG
             return Ok(true);
         }
     }
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -2547,7 +2547,9 @@ pub async fn should_coordinate_dkg(
     // instead of doing new DKG rounds. If we fail to do so they will eventually
     // be marked as failed, and we will resume DKG-ing.
     let latest_dkg_shares = storage.get_latest_encrypted_dkg_shares().await?;
-    if latest_dkg_shares.map(|s| s.dkg_shares_status) == Some(model::DkgSharesStatus::Unverified) {
+    if latest_dkg_shares.as_ref().map(|s| s.dkg_shares_status)
+        == Some(model::DkgSharesStatus::Unverified)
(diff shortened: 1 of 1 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -2563,7 +2563,16 @@ pub async fn should_coordinate_dkg(
 
         // Trigger DKG if signer set changes
         if registry_signer_info.signer_set != config.signer.bootstrap_signing_set {
-            tracing::info!("signer set has changed; proceeding with DKG");
+            // Check if we have verified shares that match the current bootstrap set
+            if let Some(latest_shares) = storage.get_latest_encrypted_dkg_shares().await? {
```
matteojug: nit: we already have fetched them in `latest_dkg_shares`

signer/src/transaction_coordinator.rs
```diff
@@ -2563,7 +2563,16 @@ pub async fn should_coordinate_dkg(
...
+                if latest_shares.dkg_shares_status == model::DkgSharesStatus::Verified
+                    && latest_shares.signer_set_public_keys() == config.signer.bootstrap_signing_set
+                {
+                    // We have verified shares for the current bootstrap set, skip DKG
+                    return Ok(false);
+                }
```
matteojug: We should do something similar also for the threshold change

signer/src/transaction_coordinator.rs
```diff
@@ -2563,7 +2563,16 @@ pub async fn should_coordinate_dkg(
...
         // Trigger DKG if signer set changes
         if registry_signer_info.signer_set != config.signer.bootstrap_signing_set {
-            tracing::info!("signer set has changed; proceeding with DKG");
+            // Check if we have verified shares that match the current bootstrap set
+            if let Some(latest_shares) = storage.get_latest_encrypted_dkg_shares().await? {
+                if latest_shares.dkg_shares_status == model::DkgSharesStatus::Verified
+                    && latest_shares.signer_set_public_keys() == config.signer.bootstrap_signing_set
+                {
+                    // We have verified shares for the current bootstrap set, skip DKG
+                    return Ok(false);
```
matteojug: While it's probably okay, it may be better to not return anything and continuing executing the latter part of the function that checks for other conditions -- eg, we may not want to trigger DKG because of the changed set but we may want to run it because of the bitcoin height (the `match` at the end of the function).

signer/src/transaction_coordinator.rs
setzeus: Copy, fixed.

signer/src/transaction_coordinator.rs
setzeus: Don't think I follow exactly, are you saying we should check that the threshold is the same?

signer/src/transaction_coordinator.rs
setzeus: I *think* that broke a test? https://github.com/stacks-sbtc/sbtc/actions/runs/16416881743/job/46384895207?pr=1773

signer/src/transaction_coordinator.rs
matteojug: We rerun DKG if the threshold or the signer set changed, right now this PR only addressed the latter: we should do a similar _do we already have DKG shares addressing this change_ check before triggering a DKG because of a change in the threshold.

signer/src/transaction_coordinator.rs
matteojug: As of [b41e84e](https://github.com/stacks-sbtc/sbtc/pull/1773/commits/b41e84eba0fb2cb0e4cba310c2d799c0f164e703), we no longer run DKG for changes in the signer set (we only have logs); that should have broken some tests (if not, we are missing some good tests)

signer/src/transaction_coordinator.rs
setzeus: Changed in [5acd375](https://github.com/stacks-sbtc/sbtc/pull/1773/commits/5acd375119d68d62f03376b140b8b3079c54f8ab), added a test as well!

signer/src/transaction_coordinator.rs
```diff
@@ -2557,14 +2559,47 @@ pub async fn should_coordinate_dkg(
...
+                        "signer set has changed but we have verified shares for current set; checking other conditions"
+                    );
+                } else {
+                    // No verified shares for current set, proceed with DKG
+                    tracing::info!("signer set has changed; proceeding with DKG");
+                }
+            } else {
+                // No shares available, proceed with DKG
+                tracing::info!("signer set has changed; proceeding with DKG");
+            }
```
djordon, line 2602: I think that we should simplify this to something like:
```rust
    let Some(latest_dkg_shares) = latest_dkg_shares else {
        return Ok(true);
    }
    if latest_dkg_shares.dkg_shares_status == model::DkgSharesStatus::Unverified {
        tracing::debug!("latest shares are unverified; skipping DKG");
        return Ok(false);
    }

    // If the registry does not have signer set info, we will
    // allow DKG below if we have not run DKG yet.
    if let Some(registry_signer_info) = context.state().registry_signer_set_info() {
        // Trigger DKG if signatures_required has changed
        if registry_signer_info.signatures_required != config.signer.bootstrap_signatures_required || 
            registry_signer_info.signer_set != config.signer.bootstrap_signing_set {
            // Check if we have verified shares that match the current threshold
            
            if latest_dkg_shares.signature_share_threshold != config.signer.bootstrap_signatures_required ||
                latest_dkg_shares.signer_set != config.signer.bootstrap_signing_set {
                // to check other conditions (like height requirements)
                tracing::info!(
                    "signatures required has changed but we have verified shares for current threshold; checking other conditions"
                );
                return Ok(true);
            }
        }
    }
```
I didn't update the comments or the logs, but I think they need to be updated as well.

signer/src/transaction_coordinator.rs
```diff
@@ -2547,7 +2547,9 @@ pub async fn should_coordinate_dkg(
     // instead of doing new DKG rounds. If we fail to do so they will eventually
     // be marked as failed, and we will resume DKG-ing.
     let latest_dkg_shares = storage.get_latest_encrypted_dkg_shares().await?;
```
djordon, line 2549: We should use a function that gets non-failed shares here. We'll have to add a query.

djordon:
This was superceded by https://github.com/stacks-sbtc/sbtc/pull/1795

PR #1776, opened 2025-07-17
feat: set the sign and dkg IDs in wsts coordinator state machines

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1177 and closes https://github.com/stacks-sbtc/sbtc/issues/1775.

## Changes

* Set the `dkg_id` in the WSTS state machines. The `dkg_id` is set to the bitcoin block height associated with the DKG round that generated the keys.
* Set the `sign_id` during all signing rounds. It is set to the first 8 bytes of the SHA-256 of the message and the bitcoin chain tip at the time that the signing round took place.

These values are only set in the coordinator state machine and are inherited in the signer state machine. This ends up being okay, since out checks in the tx-signer is validation that we would do in the state machine.

### Notes

We do not set the `sign_id` in the FROST coordinator because that field is private there. This is okay, because I want to remove our use of the FROST coordinator (and remove it generally https://github.com/stacks-sbtc/wsts/issues/180), which we can do once https://github.com/stacks-sbtc/wsts/issues/174 is done.

Also, WSTS internally changes the sign ID for you, which is kinda helpful but not at the same time. For us that means that the expected sign ID is always off by one.

## Testing Information

I added an integration test that checks that the values send over the wire have the expected `dkg_id` and `sign_id`.

## Checklist

### DIFF
--- a/signer/src/wsts_state_machine.rs
+++ b/signer/src/wsts_state_machine.rs
@@ -94,12 +94,13 @@ fn construct_signing_round_id(message: &[u8], bitcoin_chain_tip: &BitcoinBlockHa
         .finalize()
         .into();
 
-    // Use the first 8 bytes of the digest to create a u64 index. Since `digest`
-    // is 64 bytes and we explicitly take the first 8 bytes, this is safe.
+    // Use the first 8 bytes of the digest to create a u64 index. Since
+    // `digest` is 32 bytes and we explicitly take the first 8 bytes, this
+    // is safe.
     #[allow(clippy::expect_used)]
     let u64_bytes: [u8; 8] = digest[..8]
         .try_into()
-        .expect("BUG: failed to take first 4 bytes of digest");
+        .expect("BUG: failed to take first 8 bytes of digest");
 
     u64::from_le_bytes(u64_bytes)
 }
--- a/signer/src/wsts_state_machine.rs
+++ b/signer/src/wsts_state_machine.rs
@@ -88,7 +88,7 @@ impl From<SigHash> for StateMachineId {
 /// The signing round id is a u64 that is used to identify the signing round.
 /// It is constructed by hashing the message and bitcoin chain tip together.
 /// The first 8 bytes of the hash are used as the u64.
-fn construct_signing_round_id(message: &[u8], bitcoin_chain_tip: &BitcoinBlockHash) -> u64 {
+pub fn construct_signing_round_id(message: &[u8], bitcoin_chain_tip: &BitcoinBlockHash) -> u64 {
     let digest: [u8; 32] = Sha256::new()
         .chain_update(message)
         .chain_update(bitcoin_chain_tip.into_bytes())
--- a/signer/src/wsts_state_machine.rs
+++ b/signer/src/wsts_state_machine.rs
@@ -13,12 +13,15 @@ use crate::keys::PublicKeyXOnly;
 use crate::keys::SignerScriptPubKey as _;
 use crate::storage;
 use crate::storage::model;
+use crate::storage::model::BitcoinBlockHash;
 use crate::storage::model::DkgSharesStatus;
 use crate::storage::model::SigHash;
 
 use hashbrown::HashMap;
 use hashbrown::HashSet;
 use rand::rngs::OsRng;
+use sha2::Digest as _;
+use sha2::Sha256;
 use wsts::common::PolyCommitment;
 use wsts::net::Message;
 use wsts::net::Packet;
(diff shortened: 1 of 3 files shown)


xoloki:
@djordon `fn Coordinator::start_dkg_round` takes an optional `dkg_id`, and if it's set we don't increment the `dkg_id`.  I think we should make the same change for `fn Coordinator::start_signing_round`, which would fix your issues here with both the `frost::Coordinator` and also incrementing the `sign_id`.  

If you agree, I'll make that change now and we can base this PR on it.

djordon:
> @djordon `fn Coordinator::start_dkg_round` takes an optional `dkg_id`, and if it's set we don't increment the `dkg_id`.

That change is not in the current version of WSTS that we are using, but I should add a comment that this can change when we change which version of WSTS we use.




> I think we should make the same change for `fn Coordinator::start_signing_round`, which would fix your issues here with both the `frost::Coordinator` and also incrementing the `sign_id`.
> 
> If you agree, I'll make that change now and we can base this PR on it.

Yeah, we should do that as well.

djordon:
> It seems like this is a good reason to upgrade the WSTS dependency.

Yeah, maybe. Probably not in this PR though.

signer/src/wsts_state_machine.rs
```diff
@@ -79,6 +83,29 @@ impl From<SigHash> for StateMachineId {
...
+
+    // Use the first 8 bytes of the digest to create a u64 index. Since
+    // `digest` is 32 bytes and we explicitly take the first 8 bytes, this
+    // is safe.
+    #[allow(clippy::expect_used)]
+    let u64_bytes: [u8; 8] = digest[..8]
+        .try_into()
+        .expect("BUG: failed to take first 8 bytes of digest");
+
+    u64::from_le_bytes(u64_bytes)
```
xoloki, line 109: nit qq: why `from_le_bytes` and not `from_be_bytes`?

signer/src/wsts_state_machine.rs
```diff
@@ -231,7 +276,8 @@ impl WstsCoordinator for FireCoordinator {
             signer_public_keys,
         };
 
-        let wsts_coordinator = fire::Coordinator::new(config);
+        let mut wsts_coordinator = fire::Coordinator::new(config);
+        wsts_coordinator.current_dkg_id = *block_height;
```
xoloki, line 283: nit: will need to revisit this once you update the wsts dependency

signer/src/wsts_state_machine.rs
```diff
@@ -292,16 +343,23 @@ impl WstsCoordinator for FireCoordinator {
     fn start_signing_round(
         &mut self,
         message: &[u8],
+        bitcoin_chain_tip: &BitcoinBlockHash,
         signature_type: SignatureType,
     ) -> Result<Packet, Error> {
+        self.0.current_sign_id = construct_signing_round_id(message, bitcoin_chain_tip);
```
xoloki, line 354: revisit this too

signer/src/wsts_state_machine.rs
```diff
@@ -396,8 +460,11 @@ impl WstsCoordinator for FrostCoordinator {
     fn start_signing_round(
         &mut self,
         message: &[u8],
+        _: &BitcoinBlockHash,
         signature_type: SignatureType,
     ) -> Result<Packet, Error> {
+        // The current sign ID is private in the FROST coordinator so we
```
xoloki, line 473: Mark this to revisit as well.  

We could also make a WSTS branch with relatively safe fixes, and update to a git rev from that branch.   I'd argue for that rather than leave this here, but it's not my repo.

xoloki (dismissed):
LGTM.  Feel free to respond or just resolve the comment threads.  The only one I cared about was about updating the WSTS dependency, but we've already discussed that.

signer/src/wsts_state_machine.rs
djordon: I couldn't think I had a reason for preferring one over the other, so I went with whatever my autocomplete was suggesting.

signer/src/wsts_state_machine.rs
djordon: Added in https://github.com/stacks-sbtc/sbtc/pull/1776/commits/34526d4cfd3138714e73c790735d8494a9f45fc9

signer/src/wsts_state_machine.rs
djordon: Added in https://github.com/stacks-sbtc/sbtc/pull/1776/commits/34526d4cfd3138714e73c790735d8494a9f45fc9

signer/src/wsts_state_machine.rs
djordon: Added in https://github.com/stacks-sbtc/sbtc/pull/1776/commits/34526d4cfd3138714e73c790735d8494a9f45fc9

xoloki (approved):
LGTM, thanks!

signer/src/wsts_state_machine.rs
```diff
@@ -335,7 +395,10 @@ impl WstsCoordinator for FrostCoordinator {
             signer_public_keys,
         };
 
-        let wsts_coordinator = frost::Coordinator::new(config);
+        let mut wsts_coordinator = frost::Coordinator::new(config);
+        // TODO: Revisit when https://github.com/stacks-sbtc/wsts/pull/198
```
MCJOHN974, line 402: nit: Maybe open issue to not forget about this TODO

MCJOHN974 (approved):

signer/src/wsts_state_machine.rs
djordon: Oh right, I created https://github.com/stacks-sbtc/sbtc/issues/1778 for this.

xoloki (approved):
LGTM

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -2724,6 +2728,354 @@ async fn sign_bitcoin_transaction_multiple_locking_keys() {
...
+/// make integration-env-up-ci
+/// ```
+///
+/// then, once everything is up and running, run the test.
+#[tokio::test]
+async fn wsts_ids_set_during_dkg_and_signing_rounds() {
+    let (_, signer_key_pairs): (_, [Keypair; 3]) = testing::wallet::regtest_bootstrap_wallet();
+    let (rpc, faucet) = regtest::initialize_blockchain();
+
+    signer::logging::setup_logging("info,signer=debug", false);
```
matteojug: nit: debug leftover?

matteojug (dismissed):

signer/tests/integration/transaction_coordinator.rs
djordon: Oops

djordon: Removed in https://github.com/stacks-sbtc/sbtc/pull/1776/commits/a0a2e8796ced5d82142c9fed807fc5ae1943af95

xoloki (approved):
LGTM.  But somehow removing this line broke integration tests XD

PR #1779, opened 2025-07-23
testing: skip importing already-loaded descriptors when initializing the local faucet

cylewitruk:
## Description

Before importing the `Faucet`'s descriptors, first check if they are already loaded (which is highly likely for all but the first test when running integration tests) and skip it if already loaded.

Before I realized that Docker had switched me to their new VMM hypervisor, I had a lot of scans in my bitcoind logs which is triggered by importing the faucet's descriptors over and over again which were taking ~13s, so I had tried this change with little effect. But after switching back to Apple Virtualization with Rosetta, this change netted me about 40s savings on local integration test runs:

```
Before: 318.699s
After: 277.045s
```

I would guess that this might have been a contributing factor to @djordon describing that "bitcoin became slow" if you run it too long with many integration test runs and without restarts in-between.

## Changes

Makes a simple `listdescriptors` RPC call and returns early from `Faucet::track_address` if it's already loaded, skipping the scan.

## Checklist

### DIFF
--- a/.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
+++ b/.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
@@ -1,4 +1,4 @@
-FROM rust:1.87.0-slim-bookworm AS build
+FROM rust:1.88.0-slim-bookworm AS build
 
 ARG CARGO_BUILD_ARGS="--release --locked"
 
--- a/.github/actions/dockerfiles/Dockerfile.signer.debian
+++ b/.github/actions/dockerfiles/Dockerfile.signer.debian
@@ -1,4 +1,4 @@
-FROM rust:1.87.0-slim-bookworm AS build
+FROM rust:1.88.0-slim-bookworm AS build
 
 ARG GIT_COMMIT
 RUN test -n "$GIT_COMMIT" || (echo "GIT_COMMIT not set" && false)
--- a/.github/workflows/image-build-and-draft-release.yaml
+++ b/.github/workflows/image-build-and-draft-release.yaml
@@ -87,7 +87,7 @@ jobs:
       ## Build docker image for release
       - name: Build and Push ( ${{ matrix.dist }} ${{ matrix.docker_target }} )
         id: docker_build
-        uses: stacks-sbtc/actions/docker/build-push-action@181f8c67da2707c66b5e31f24e7418c47adefdd1
+        uses: stacks-sbtc/actions/docker/build-push-action@942832cc7f86ead03c05b7035cd4b9573dd6c617
         with:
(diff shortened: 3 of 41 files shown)


sbtc/src/testing/regtest.rs
```diff
@@ -269,6 +269,26 @@ impl Faucet {
...
+        // This isn't part of the bitcoincore_rpc API, unfortunately.
+        let list_descriptors_request = self
+            .rpc
+            .get_jsonrpc_client()
+            .build_request("listdescriptors", None);
+        let list_descriptors_response = self
+            .rpc
+            .get_jsonrpc_client()
+            .send_request(list_descriptors_request)
+            .expect("failed to list descriptors");
```
matteojug: maybe nit: we could use `self.rpc.call` here instead of the underlying client, but we need to define the response struct with at least the fields we are interested in. It's probably more robust than the current _does the json str contains the descriptor name_, but OTOH this is test code that works so maybe it's not worth the effort and we can just keep it as is. Just wanted to flag this in case anybody had strong feelings about it. I'm fine with the current impl.

matteojug (approved):

sbtc/src/testing/regtest.rs
cylewitruk: Yeah I thought about doing that but ended up with "meh", but if someone really wants it then I can add it.

djordon: Yeah, I think I care. Can we add a struct for the fields that we want and use `self.rpc.call` instead?

djordon: Also, if this starts to take too long then just go with what is here.

MCJOHN974 (approved):

sbtc/src/testing/regtest.rs
matteojug: Done in [502074f](https://github.com/stacks-sbtc/sbtc/pull/1779/commits/502074f3b6b5543323b92c214219a80496e5170b). Also confirmed that locally this saves about 40s (on a M1), which is nice!

sbtc/src/testing/regtest.rs
```diff
@@ -269,6 +269,26 @@ impl Faucet {
...
+
+        if list_descriptors_response
+            .result
+            .is_some_and(|f| f.get().contains(&descriptor_info.descriptor))
+        {
```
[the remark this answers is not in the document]
cylewitruk: Resolved by [502074f](https://github.com/stacks-sbtc/sbtc/pull/1779/commits/502074f3b6b5543323b92c214219a80496e5170b)

MCJOHN974 (approved):

Issue #1738, opened 2025-06-16
[Feature]: Improve bitcoin core ZMQ stream resiliency


cylewitruk:
[Feature]: Improve bitcoin core ZMQ stream resiliency

## 1. Description

We've seen a number of cases where the ZMQ stream for mainnet signers has randomly "stopped" receiving new block notifications, despite no disconnect being logged. Reconnect-after-disconnect functionality exists and works, however we have no mechanism to detect and attempt to restart a stream if `hashblock` notifications stop being received despite a reportedly connected stream.

To increase the resiliency of the signers, our ZMQ client should be able to handle this scenario.

### 1.1 Context & Purpose

If the signers stop receiving bitcoin block notifications, they effectively "go offline" due to being out-of-date in relation to the current bitcoin chain tip. This would add a recovery mode for ZMQ streams which go "stale" and stop receiving notifications, _avoiding signer partners needing to manually restart their sBTC signer_.

## 2. Technical Details:

The simplest solution would be to implement a time-based check within the ZMQ subscriber itself which attempts to reconnect the ZMQ stream if a `hashblock` notification hasn't been received within a configurable timeframe.

PR #1780, opened 2025-07-24
feat: improve zmq resiliency

cylewitruk:
## Description

Closes: #1738

Improves the resiliency of the ZMQ subscriptions/connections by introducing a timeout since the last received bitcoin block notification (currently hardcoded at 1hr), after which the ZMQ stream will be fully restarted (disconnected/reconnected/resubscribed). Existing reconnect functionality upon disconnect remains the same.

## Changes

- Introduced a new `BlockHashStreamProvider` trait, implemented by types which can provide streams of block hashes.
- New `BitcoinCoreMessageDispatcher` type which can dispatch messages from bitcoin ZMQ. This now internally buffers messages in a broadcast channel so it's also `Clone`-able (i.e. can have multiple subscribers without opening multiple ZMQ subscriptions). Handles connection events like before, but also handles the "timeout" mentioned in the description with a forced reconnect & resubscribe.
- Changed the `BlockObserver` to take a `BlockHashStreamProvider` instead of a raw stream, making testing easier for specific scenarios (e.g. in `emily.rs` integration tests, but also in other tests yet to be PR'd).

## Testing Information

Difficult to test via integration tests, but I've tested it quite a bit in devenv. I can't recreate the exact issue which some partners have experienced, but the "timeout" case simulates it (because in devenv when you stop the bitcoin node, the bitcoin miner dies and there will be no more block notifications after bitcoin re-start).

- The traditional "reconnect" case:
  ```
  2025-07-24T14:25:37.733683Z  WARN signer::bitcoin::zmq: disconnected from ZMQ endpoint fd=28 endpoint=tcp://127.0.0.1:28332
  2025-07-24T14:25:37.889629Z  INFO signer::bitcoin::zmq: connected to ZMQ endpoint fd=28 endpoint=tcp://127.0.0.1:28332
  2025-07-24T14:25:37.890087Z  WARN signer::bitcoin::zmq: disconnected from ZMQ endpoint fd=28 endpoint=tcp://127.0.0.1:28332
  2025-07-24T14:25:38.064653Z  INFO signer::bitcoin::zmq: connected to ZMQ endpoint fd=28 endpoint=tcp://127.0.0.1:28332
  2025-07-24T14:25:38.078504Z  WARN signer::bitcoin::zmq: disconnected from ZMQ endpoint fd=28 endpoint=tcp://127.0.0.1:28332
  2025-07-24T14:25:41.422688Z  INFO signer::bitcoin::zmq: connected to ZMQ endpoint fd=28 endpoint=tcp://127.0.0.1:28332
  2025-07-24T14:25:41.423391Z  WARN signer::bitcoin::zmq: disconnected from ZMQ endpoint fd=28 endpoint=tcp://127.0.0.1:28332
  2025-07-24T14:25:41.601082Z  INFO signer::bitcoin::zmq: connected to ZMQ endpoint fd=28 endpoint=tcp://127.0.0.1:28332
  ```
- The new "timeout" case:
  ```
  2025-07-24T12:38:40.184808Z  WARN signer::bitcoin::zmq: ZMQ stream inactive; attempting to reconnect. endpoint=tcp://127.0.0.1:28332 timeout_secs=60
  2025-07-24T12:38:40.187517Z  INFO signer::bitcoin::zmq: waiting 1s before next ZMQ connection attempt. endpoint=tcp://127.0.0.1:28332
  2025-07-24T12:38:41.191217Z  INFO signer::bitcoin::zmq: ZMQ poller connected. endpoint=tcp://127.0.0.1:28332
  2025-07-24T12:38:41.19134Z  INFO signer::bitcoin::zmq: connected to ZMQ endpoint fd=26 endpoint=tcp://127.0.0.1:28332
  ```

## Checklist

### DIFF
--- a/sbtc/src/testing/regtest.rs
+++ b/sbtc/src/testing/regtest.rs
@@ -44,6 +44,8 @@ use std::sync::OnceLock;
 pub const BITCOIN_CORE_RPC_USERNAME: &str = "devnet";
 /// The password for RPC calls in bitcoin-core
 pub const BITCOIN_CORE_RPC_PASSWORD: &str = "devnet";
+/// Default ZMQ endpoint for regtest bitcoin-core
+pub const BITCOIN_CORE_ZMQ_ENDPOINT: &str = "tcp://localhost:28332";
 
 /// The fallback fee in bitcoin core
 pub const BITCOIN_CORE_FALLBACK_FEE: Amount = Amount::from_sat(1000);
--- a/signer/src/bitcoin/zmq.rs
+++ b/signer/src/bitcoin/zmq.rs
@@ -26,96 +26,371 @@
 //!   docker compose, this means that you should use the `stop` command and not
 //!   the `down` command.
 
-use std::future::ready;
-use std::pin::Pin;
-use std::task::Context;
-use std::task::Poll;
+use std::sync::Arc;
 use std::time::Duration;
 
-use bitcoin::Block;
 use bitcoin::BlockHash;
 use bitcoincore_zmq::Message;
 use bitcoincore_zmq::SocketEvent;
 use bitcoincore_zmq::SocketMessage;
-use bitcoincore_zmq::subscribe_async_monitor_stream::MessageStream;
+use futures::TryStreamExt as _;
 use futures::stream::Stream;
 use futures::stream::StreamExt as _;
+use tokio::sync::broadcast;
+use tokio::sync::oneshot;
+use tokio::task::JoinHandle;
+use tokio_stream::wrappers::BroadcastStream;
+use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
 
 use crate::error::Error;
+use crate::util::FutureExt as _;
+use crate::util::SleepAsyncExt as _;
 
-/// A struct for messages over bitcoin-core's ZeroMQ interface.
-pub struct BitcoinCoreMessageStream {
-    /// The inner stream we're wrapping.
-    inner_stream: MessageStream,
+const DEFAULT_BROADCAST_CAPACITY: usize = 100;
+const INITIAL_RECONNECT_DELAY: Duration = Duration::from_secs(1);
+const MAX_RECONNECT_DELAY: Duration = Duration::from_secs(60);
+
+/// The factor by which the reconnect delay increases after each failed attempt.
+const RECONNECT_BACKOFF_FACTOR: f32 = 1.5;
+
+/// Timeout for the initial ZMQ connection attempt in `new_from_endpoint`.
+const INITIAL_CONNECTION_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// The number of seconds to wait before considering a ZMQ connection "stale",
+/// triggering a reconnection attempt.
+const ZMQ_INACTIVITY_TIMEOUT: Duration = Duration::from_secs(3600); // 1 hour
+
+/// Error type for the Bitcoin ZMQ module, encapsulating errors related to
+/// the ZMQ data source or subscriber issues.
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum BitcoinZmqError {
+    /// An error originating from the upstream ZMQ data source or poller.
+    /// The inner String contains a description of the original error.
+    #[error("internal error: {0}")]
+    Internal(String),
+
+    /// An error related to the Bitcoin Core ZMQ interface.
+    #[error("bitcoin Core ZMQ error: {0}")]
+    BitcoinCoreZmq(String),
+
+    /// A subscriber to the ZMQ block hash broadcast lagged too far behind
+    /// and missed messages. The inner u64 is the number of messages missed.
+    #[error("subscriber lagged behind ZMQ broadcast: {0} messages missed")]
+    SubscriberLagged(u64),
 }
 
-impl BitcoinCoreMessageStream {
-    /// Create a new one using the given endpoint.
-    pub async fn new_from_endpoint(endpoint: &str) -> Result<Self, Error> {
-        let inner_stream = tokio::time::timeout(Duration::from_secs(10), async {
-            bitcoincore_zmq::subscribe_async_monitor(&[endpoint])
-        })
-        .await
-        .map_err(|_| Error::BitcoinCoreZmqConnectTimeout(endpoint.to_string()))?
-        .map_err(Error::BitcoinCoreZmq)?;
+/// Represents the control flow decision after processing a ZMQ message.
+#[derive(Debug, Clone)]
+enum ControlFlow {
+    /// Continue processing messages from the current stream.
+    Continue,
+    /// The stream has ended or an error occurred; break to attempt reconnection.
+    Reconnect,
+    /// A fatal, unrecoverable error was met; shut down the task.
+    Shutdown,
+}
 
-        Ok(Self { inner_stream })
-    }
+/// A trait for providing a stream of block hashes to be used by the block observer.
+///
+/// Implementors of this trait are responsible for sourcing block hash notifications,
+/// typically from a Bitcoin Core ZMQ endpoint, and making them available as an
+/// asynchronous stream. This abstraction allows different components, such as
+/// block observers or test utilities, to consume block hash events without being
+/// coupled to the specific mechanism of how those events are obtained.
+pub trait BlockHashStreamProvider: Send + Sync {
+    /// Subscribes to the block hash stream, returning a new stream that emits
+    /// block hashes as they are received.
+    ///
+    /// The returned stream will yield `Result<BlockHash, BitcoinZmqError>`.
+    ///
+    /// Consumers of this stream are responsible for mapping `BitcoinZmqError`
+    /// to a broader application error type (like `crate::error::Error`) if needed.
+    fn get_block_hash_stream(
+        &self,
+    ) -> impl futures::Stream<Item = Result<BlockHash, BitcoinZmqError>> + Send + Sync + Unpin + 'static;
+}
 
-    /// Method we use to inspect incoming messages and log things.
-    fn inspect_message(msg: &Result<SocketMessage, Error>) {
-        match msg {
-            Ok(SocketMessage::Event(event)) => match event.event {
-                SocketEvent::Connected { fd } => {
-                    tracing::info!(%fd, endpoint = event.source_url, "connected to ZeroMQ endpoint");
+/// Encapsulates the state and logic for the ZMQ polling task.
+struct Poller {
+    endpoint: String,
+    broadcast_tx: broadcast::Sender<Result<BlockHash, BitcoinZmqError>>,
+    initial_conn_signal_tx: Option<oneshot::Sender<Result<(), BitcoinZmqError>>>,
+}
+
+impl Poller {
+    /// The main entry point for the poller task. Contains the reconnection loop.
+    async fn run(mut self) {
+        let mut reconnect_attempt: u32 = 0;
+        loop {
+            // This function now returns a variant that explicitly tells us what to do next.
+            match self.connect_and_process_stream().await {
+                ControlFlow::Shutdown => {
+                    tracing::info!("Poller shutting down.");
+                    return;
                 }
-                SocketEvent::Disconnected { fd } => {
-                    tracing::warn!(%fd, endpoint = event.source_url, "disconnected from ZeroMQ endpoint");
+                ControlFlow::Reconnect => {
+                    // A reconnect is needed. Apply backoff and try again.
+                    let backoff_duration = calculate_backoff_duration(reconnect_attempt);
+                    tracing::info!(
+                        endpoint = %self.endpoint,
+                        "waiting {backoff_duration:?} before next ZMQ connection attempt."
+                    );
+                    backoff_duration.sleep().await;
+                    reconnect_attempt = reconnect_attempt.saturating_add(1);
                 }
-                _ => {}
-            },
-            Ok(SocketMessage::Message(msg)) => match msg {
-                Message::Block(block, height) => {
-                    tracing::trace!(block_hash = %block.block_hash(), block_height = %height, "received block");
+                ControlFlow::Continue => {
+                    // This case should not be returned by connect_and_process_stream.
+                    // We'll log a warning and treat it as a reconnect to be safe.
+                    tracing::warn!(
+                        "connect_and_process_stream returned Continue; treating as Reconnect."
+                    );
+                    reconnect_attempt = 0; // Reset backoff as this is unexpected.
                 }
-                Message::HashBlock(hash, height) => {
-                    tracing::trace!(block_hash = %hash, block_height = %height, "received block hash");
-                }
-                _ => {}
-            },
-            Err(error) => {
-                tracing::error!(%error, "error receiving message from ZeroMQ");
             }
         }
     }
 
-    /// Convert this stream into one that returns only blocks
-    pub fn to_block_stream(self) -> impl Stream<Item = Result<Block, Error>> {
-        self.inspect(Self::inspect_message)
-            .filter_map(|msg| match msg {
-                Ok(SocketMessage::Message(Message::Block(block, _))) => ready(Some(Ok(block))),
-                Err(err) => ready(Some(Err(err))),
-                Ok(_) => ready(None),
-            })
+    /// Attempts to connect to the ZMQ endpoint and, if successful, enters the
+    /// message processing loop.
+    ///
+    /// Returns a `ControlFlow` variant indicating whether to shut down or reconnect.
+    async fn connect_and_process_stream(&mut self) -> ControlFlow {
+        let zmq_stream = match bitcoincore_zmq::subscribe_async_monitor(&[&self.endpoint]) {
+            Ok(stream) => stream,
+            Err(error) => {
+                tracing::warn!(
+                    endpoint = %self.endpoint,
+                    %error,
+                    "failed to connect ZMQ; will retry..."
+                );
+                // If we fail to connect, we must signal the constructor if it's still waiting.
+                if let Some(tx) = self.initial_conn_signal_tx.take() {
+                    let err_msg = format!("initial ZMQ connection attempt failed: {error}");
+                    let _ = tx.send(Err(BitcoinZmqError::Internal(err_msg)));
+                }
+                return ControlFlow::Reconnect;
+            }
+        };
+
+        tracing::info!(endpoint = %self.endpoint, "ZMQ poller connected.");
+        let mut poller_stream =
+            zmq_stream.map_err(|e| BitcoinZmqError::BitcoinCoreZmq(e.to_string()).into());
+
+        // Loop to process messages from the connected stream.
+        loop {
+            let message_result = poller_stream
+                .next()
+                .with_timeout(ZMQ_INACTIVITY_TIMEOUT)
+                .await;
+
+            let decision = match message_result {
+                // A message was successfully received from the stream.
+                Ok(Some(message)) => self.handle_message(message),
+                // The stream ended gracefully.
+                Ok(None) => {
+                    tracing::info!(
+                        endpoint = %self.endpoint,
+                        "ZMQ stream ended; will attempt to reconnect."
+                    );
+                    ControlFlow::Reconnect
+                }
+                // The stream timed out waiting for a message.
+                Err(_) => {
+                    tracing::warn!(
+                        endpoint = %self.endpoint,
+                        timeout_secs = ZMQ_INACTIVITY_TIMEOUT.as_secs(),
+                        "ZMQ stream inactive; attempting to reconnect."
+                    );
+                    ControlFlow::Reconnect
+                }
+            };
+
+            // Act on the decision from handling the message or the stream event.
+            match decision {
+                ControlFlow::Continue => continue,
+                // For any other decision, we exit this function and let the main
+                // run loop handle it.
+                other => return other,
+            }
+        }
     }
 
-    /// Convert this stream into one that returns only block hashes
-    pub fn to_block_hash_stream(self) -> impl Stream<Item = Result<BlockHash, Error>> {
-        self.inspect(Self::inspect_message)
-            .filter_map(|msg| match msg {
-                Ok(SocketMessage::Message(Message::HashBlock(hash, _))) => ready(Some(Ok(hash))),
-                Err(err) => ready(Some(Err(err))),
-                Ok(_) => ready(None),
-            })
+    /// Processes a single message or event and determines the next action.
+    fn handle_message(&mut self, msg: Result<SocketMessage, Error>) -> ControlFlow {
+        inspect_zmq_message(&msg);
+
+        match msg {
+            Ok(SocketMessage::Event(event)) => {
+                tracing::trace!(?event, "received ZMQ event");
+                // Only attempt to take and send the signal if the event is `Connected`.
+                if let SocketEvent::Connected { .. } = event.event {
+                    // Now that we know it's the right event, we can take the sender.
+                    if let Some(tx) = self.initial_conn_signal_tx.take() {
+                        // On first successful connection, signal the constructor.
+                        if tx.send(Ok(())).is_err() {
+                            tracing::warn!("failed to send initial ZMQ connection success signal.");
+                        }
+                    }
+                }
+            }
+            Ok(SocketMessage::Message(Message::HashBlock(hash, _))) => {
+                tracing::trace!(block_hash = %hash, "received ZMQ block hash");
+                // If send fails, it means there are no subscribers. This is a normal
+                // condition during startup. We can just log it at trace level and continue.
+                if self.broadcast_tx.send(Ok(hash)).is_err() {
+                    tracing::trace!("no subscribers, dropping ZMQ block hash message");
+                }
+            }
+            Ok(_) => {} // Ignore other message types
+            Err(Error::BitcoinCoreZmq(error)) => {
+                tracing::warn!(endpoint = %self.endpoint, %error, "ZMQ stream error; reconnecting.");
+                // If a stream error occurs before the initial connection is confirmed,
+                // signal the failure to the constructor.
+                if let Some(tx) = self.initial_conn_signal_tx.take() {
+                    let _ = tx.send(Err(error));
+                }
+                return ControlFlow::Reconnect;
+            }
+            Err(error) => {
+                // For any other unexpected error, we treat it as fatal.
+                tracing::error!(endpoint = %self.endpoint, "unexpected poller error: {error}; shutting down.");
+                let bcast_err = BitcoinZmqError::Internal(error.to_string());
+                if let Some(tx) = self.initial_conn_signal_tx.take() {
+                    let _ = tx.send(Err(bcast_err.clone()));
+                }
+                let _ = self.broadcast_tx.send(Err(bcast_err));
+                return ControlFlow::Shutdown;
+            }
+        }
+        ControlFlow::Continue
     }
 }
 
-impl Stream for BitcoinCoreMessageStream {
-    type Item = Result<SocketMessage, Error>;
+/// A stream provider that connects to a Bitcoin Core ZMQ endpoint,
+/// polls for block hash messages, and broadcasts them to multiple subscribers.
+#[derive(Clone)]
+pub struct BitcoinCoreMessageDispatcher {
+    // The broadcast channel now sends Result<BlockHash, BitcoinZmqError>
+    broadcast_tx: broadcast::Sender<Result<BlockHash, BitcoinZmqError>>,
+    // Keep the task handle to ensure the poller task isn't dropped prematurely
+    // and potentially for graceful shutdown in the future.
+    _poller_task_handle: Arc<JoinHandle<()>>,
+}
 
-    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        self.inner_stream
-            .poll_next_unpin(cx)
-            .map_err(Error::BitcoinCoreZmq)
+/// Inspects a ZMQ message and logs its content based on the type of message.
+fn inspect_zmq_message(msg: &Result<SocketMessage, Error>) {
+    match msg {
+        Ok(SocketMessage::Event(event)) => match event.event {
+            SocketEvent::Connected { fd } => {
+                tracing::info!(%fd, endpoint = %event.source_url, "connected to ZMQ endpoint");
+            }
+            SocketEvent::Disconnected { fd } => {
+                tracing::warn!(%fd, endpoint = %event.source_url, "disconnected from ZMQ endpoint");
+            }
+            _ => {}
+        },
+        Ok(SocketMessage::Message(message)) => match message {
+            Message::Block(block, height) => {
+                tracing::trace!(block_hash = %block.block_hash(), %height, "received ZMQ block");
+            }
+            Message::HashBlock(hash, height) => {
+                tracing::trace!(block_hash = %hash, %height, "received ZMQ block hash");
+            }
+            // Other message types like full blocks are ignored by this poller
+            _ => {}
+        },
+        Err(error) => {
+            // This logging is for errors from the ZmqMessagePoller stream itself.
+            // Specific handling (reconnect or broadcast error) happens in the poller loop.
+            tracing::debug!(%error, "ZMQ poller received an error from inner stream");
+        }
+    }
+}
+
+/// Calculates the backoff duration based on attempt number.
+fn calculate_backoff_duration(attempt: u32) -> Duration {
+    if attempt == 0 {
+        INITIAL_RECONNECT_DELAY
+    } else {
+        // Calculate delay using f32 arithmetic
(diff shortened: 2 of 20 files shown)


signer/src/bitcoin/zmq.rs
```diff
@@ -26,96 +26,371 @@
...
+        },
+        Err(error) => {
+            // This logging is for errors from the ZmqMessagePoller stream itself.
+            // Specific handling (reconnect or broadcast error) happens in the poller loop.
+            tracing::debug!(%error, "ZMQ poller received an error from inner stream");
+        }
     }
+}
+
+/// Calculates the backoff duration based on attempt number.
```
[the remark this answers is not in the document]
cylewitruk: Since this is an internal method, and the method is small and readable, I opted to not over-document it as exponential backoffs are widely "grasped".

signer/src/bitcoin/zmq.rs
```diff
@@ -26,96 +26,371 @@
...
-                Ok(_) => ready(None),
-            })
+        // Clamp the result between the initial and maximum delay values.
+        Duration::from_secs_f32(delay).clamp(INITIAL_RECONNECT_DELAY, MAX_RECONNECT_DELAY)
     }
 }
 
-impl Stream for BitcoinCoreMessageStream {
-    type Item = Result<SocketMessage, Error>;
+/// Encapsulates the state and logic for the ZMQ polling task.
```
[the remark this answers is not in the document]
cylewitruk: [57e3cfb](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/57e3cfb345c1bfb5a406004d2bbd24c503c9fe9c)

signer/src/bitcoin/zmq.rs
```diff
@@ -26,96 +26,371 @@
...
+            Ok(stream) => stream,
+            Err(error) => {
+                tracing::warn!(
+                    endpoint = %self.endpoint,
+                    %error,
+                    "failed to connect ZMQ; will retry..."
+                );
+                // If we fail to connect, we must signal the constructor if it's still waiting.
+                if let Some(tx) = self.initial_conn_signal_tx.take() {
+                    let err_msg = format!("initial ZMQ connection attempt failed: {error}");
```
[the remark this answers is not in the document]
cylewitruk: [9c0fb10](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/9c0fb101741c15d24d319912652861ef884cd0ae)

signer/src/bitcoin/zmq.rs
```diff
@@ -26,96 +26,371 @@
...
+
+/// The factor by which the reconnect delay increases after each failed attempt.
+const RECONNECT_BACKOFF_FACTOR: f32 = 1.5;
+
+/// Timeout for the initial ZMQ connection attempt in `new_from_endpoint`.
+const INITIAL_CONNECTION_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// The number of seconds to wait before considering a ZMQ connection "stale",
+/// triggering a reconnection attempt.
+const ZMQ_INACTIVITY_TIMEOUT: Duration = Duration::from_secs(3600); // 1 hour
```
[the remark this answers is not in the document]
cylewitruk: I thought about this as well when I was reproducing in devenv. It's good to do, and I should, but wanted to get the PR up first.

sbtc/src/testing/regtest.rs
```diff
@@ -44,6 +44,8 @@ use std::sync::OnceLock;
 pub const BITCOIN_CORE_RPC_USERNAME: &str = "devnet";
 /// The password for RPC calls in bitcoin-core
 pub const BITCOIN_CORE_RPC_PASSWORD: &str = "devnet";
+/// Default ZMQ endpoint for regtest bitcoin-core
+pub const BITCOIN_CORE_ZMQ_ENDPOINT: &str = "tcp://localhost:28332";
```
djordon: Feels strange to put this here, since this crate doesn't interact with bitcoin-core's ZMQ stream.

djordon: Spoke offline about this, and we already use the variables here, so it makes sense to colocate this variable here as well.

signer/src/bitcoin/zmq.rs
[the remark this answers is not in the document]
cylewitruk: Removed in new polling model

signer/src/bitcoin/zmq.rs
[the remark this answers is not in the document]
cylewitruk: Irrelevant now in new polling model

signer/src/bitcoin/zmq.rs
```diff
@@ -26,96 +26,371 @@
...
+
```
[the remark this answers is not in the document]
cylewitruk: Irrelevant now in new polling model

signer/src/config/mod.rs
```diff
@@ -130,9 +136,48 @@ pub struct BitcoinConfig {
...
+                ));
+            }
+        }
+
+        if self.chain_tip_polling_interval.as_secs()
+            > MAX_BITCOIN_CHAIN_TIP_POLLING_INTERVAL_SECONDS
+        {
+            return Err(ConfigError::Message(
+                "[bitcoin.chain_tip_polling_interval] Maximum polling interval exceeded"
+                    .to_string(),
```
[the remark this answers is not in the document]
cylewitruk: Mmm... I'm pretty sure the error message provides this?

signer/src/bitcoin/poller.rs
```diff
@@ -0,0 +1,235 @@
...
+    loop {
+        polling_interval.sleep().await;
+
+        match rpc.get_best_block_hash().await {
+            Ok(current_hash) => {
+                if current_hash != last_seen_hash {
+                    tracing::trace!(new_hash = %current_hash, "detected new best block hash");
+                    last_seen_hash = current_hash;
+                    if broadcast_tx.send(current_hash).is_err() {
+                        tracing::warn!("broadcasting new block hash failed; no subscribers?");
```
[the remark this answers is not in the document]
cylewitruk: I'm not sure this actually provides much value as it's really not expected to occur except maybe during startup/shutdown races, and the block hash doesn't provide much debug help?

signer/src/bitcoin/poller.rs
```diff
@@ -0,0 +1,235 @@
...
+                    tokio::time::sleep(Duration::from_secs(1)).await;
+                }
+            }
+        };
+
+        let (broadcast_tx, _rx) = broadcast::channel::<BlockHash>(DEFAULT_BROADCAST_CAPACITY);
+
+        // Broadcast the initial hash.
+        if broadcast_tx.send(initial_hash).is_err() {
+            tracing::warn!("no subscribers at initial broadcast, continuing anyway");
```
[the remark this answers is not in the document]
cylewitruk: I'm not sure this actually provides much value as it's really not expected to occur except maybe during startup/shutdown races, and the block hash doesn't provide much debug help?

docker/mainnet/nodes/docker-compose.chains.yml
```diff
@@ -2,8 +2,6 @@ x-common-vars:
   # Bitcoin RPC
   - &BITCOIN_RPC_USERNAME ${BITCOIN_RPC_USERNAME} # Required
   - &BITCOIN_RPC_PASSWORD ${BITCOIN_RPC_PASSWORD} # Required
-  - &BITCOIN_RPC_PORT ${BITCOIN_RPC_PORT:-8332}
```
matteojug, line 5: Is the `BITCOIN_RPC_PORT` removal expected? Also below

signer/tests/service-configs/bitcoin.conf
```diff
@@ -446,10 +446,6 @@ txindex=1
 
 ### ZeroMQ notification options
 
-
-# Enable publish hash block in <address>
-zmqpubhashblock=tcp://*:28332
```
matteojug: nit: I would just comment this and `zmqpubrawblock` below

signer/src/bitcoin/rpc.rs
```diff
@@ -651,6 +651,13 @@ impl BitcoinCoreClient {
             .map_err(Error::BitcoinCoreRpc)
     }
 
+    /// Gets the best block height from the Bitcoin node.
```
matteojug: nit
```suggestion
    /// Gets the best block hash from the Bitcoin node.
```

signer/src/main.rs
```diff
@@ -33,12 +33,18 @@ use tower_http::trace::TraceLayer;
...
+/// bootstrap (i.e. connect to other peers). Three seconds is a sane default
+/// value, giving the swarm a few seconds to start up and bind listener(s)
+/// before proceeding.
 const INITIAL_BOOTSTRAP_DELAY_SECS: u64 = 3;
 
+/// The timeout for the Bitcoin chain tip poller to initialize. This is the
+/// maximum time the poller will wait for the initial block hash to be
+/// received from the Bitcoin Core RPC. If this timeout is reached, the
+/// poller will return an error and shut down.
+const BITCOIN_POLLER_INITIALIZATION_TIMEOUT: Duration = Duration::from_secs(120);
```
matteojug: Do we want a init timeout? IIUC if the bitcoin node is not reachable for the first 2 minutes we will stop the sbtc signer, which may be bad?

signer/src/bitcoin/poller.rs
```diff
@@ -0,0 +1,235 @@
...
+    {
+        let start_time = tokio::time::Instant::now();
+        let mut last_error;
+
+        // Try to fetch the initial hash, retrying until the timeout is reached.
+        let initial_hash = loop {
+            match rpc.get_best_block_hash().await {
+                Ok(hash) => break hash,
+                Err(e) => {
+                    last_error = e;
```
matteojug: nit: we could remove `last_error` and just `e` around, since it's not used outside this arm

signer/src/bitcoin/poller.rs
```diff
@@ -0,0 +1,235 @@
...
+                    tracing::warn!(error = %last_error, "failed to get initial block hash, retrying...");
+                    tokio::time::sleep(Duration::from_secs(1)).await;
+                }
+            }
+        };
+
+        let (broadcast_tx, _rx) = broadcast::channel::<BlockHash>(DEFAULT_BROADCAST_CAPACITY);
+
+        // Broadcast the initial hash.
+        if broadcast_tx.send(initial_hash).is_err() {
```
matteojug: Since this will not be received by anyone, why are we sending it?

signer/src/bitcoin/poller.rs
```diff
@@ -0,0 +1,235 @@
...
+                    last_seen_hash = current_hash;
+                    if broadcast_tx.send(current_hash).is_err() {
+                        tracing::warn!("broadcasting new block hash failed; no subscribers?");
+                    }
+                }
+            }
+            Err(e) => {
+                // On a transient error, log it and continue polling. Do not send the
+                // error to consumers, as they cannot act on it.
+                tracing::error!(error = %e, "failed to get best block hash during polling; will retry.");
```
matteojug: This should probably be a warning, or an error if we don't get a success in the last X minutes? Having an error for each possible transient errors may be annoying

signer/src/bitcoin/poller.rs
matteojug: nit: add the error to the log?

signer/src/config/mod.rs
```diff
@@ -747,17 +793,9 @@ mod tests {
                 .rpc_endpoints
                 .contains(&url("http://foo:bar@localhost:5678"))
         );
-        assert!(
-            settings
-                .bitcoin
-                .block_hash_stream_endpoints
-                .contains(&url("tcp://localhost:1234"))
-        );
```
matteojug, line 755: nit: there are still some occurrences around of `block_hash_stream_endpoints` (`signer-config.toml.sample`, `signer-*.env` and in here)

signer/src/config/mod.rs
```diff
@@ -130,9 +136,48 @@ pub struct BitcoinConfig {
...
+
+            if endpoint.host_str().is_none() {
+                return Err(ConfigError::Message(
+                    "[bitcoin.rpc_endpoints] Invalid URL: host is required".to_string(),
+                ));
+            }
+        }
+
+        if self.chain_tip_polling_interval.as_secs()
+            > MAX_BITCOIN_CHAIN_TIP_POLLING_INTERVAL_SECONDS
```
matteojug, line 171: nit: can we add a test for this?

signer/src/testing/btc.rs
```diff
@@ -103,3 +85,87 @@ pub fn get_canonical_chain_tip(rpc: &Client) -> GetChainTipsResultTip {
...
+    pub fn with_capacity(capacity: usize) -> Self {
+        let (sender, _receiver) = broadcast::channel(capacity);
+        Self { sender }
+    }
+
+    /// Sends a block hash result to all subscribers of this provider.
+    ///
+    /// ## Returns
+    /// * `Ok(usize)`: The number of active receivers the message was sent to.
+    /// * `Err(broadcast::error::SendError<...>)`: If there are no active receivers.
```
matteojug: nit: it doesn't return (anymore?)

signer/tests/integration/bitcoin_poller.rs
```diff
@@ -0,0 +1,75 @@
+use bitcoin::BlockHash;
+use futures::StreamExt;
+use sbtc::testing::regtest;
+use signer::bitcoin::BitcoinBlockHashStreamProvider as _;
+use signer::bitcoin::poller::BitcoinChainTipPoller;
+use test_log::test;
+use tokio::sync::mpsc::error::TryRecvError;
+
+/// This tests that out bitcoin block hash stream receives new block hashes
```
matteojug:
```suggestion
/// This tests that our bitcoin block hash stream receives new block hashes
```

signer/tests/integration/bitcoin_poller.rs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            last_received_hash = Some(received);
+            if last_received_hash.as_ref() == Some(final_tip_hash) {
+                break;
+            }
+        }
+    }
+
+    // The last hash we received must be the final chain tip.
+    assert_eq!(Some(*final_tip_hash), last_received_hash);
```
matteojug, line 75: nit: can we also sleep a bit and check that the stream is empty (and thus `last_received_hash` is actually the last block received)?

signer/src/bitcoin/poller.rs
cylewitruk: There is literally only one error and that is that there are no subscribers:

```rust
    /// Error returned by the [`send`] function on a [`Sender`].
    ///
    /// A **send** operation can only fail if there are no active receivers,
    /// implying that the message could never be received. The error contains the
    /// message being sent as a payload so it can be recovered.
    #[derive(Debug)]
    pub struct SendError<T>(pub T);
```

The payload of what we tried to send is largely uninteresting because we'll try again in x seconds.

matteojug: Got it, but the _we'll try again in x seconds_ is not true IIUC: since we set `last_seen_hash` before the send, if we fail to send we lose that hash and will not try again until a new tip appears. But I guess it's fine regardless.

docker/mainnet/nodes/docker-compose.chains.yml
cylewitruk: Whoops, fat-fingered that one: [ddc4f13](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/ddc4f13858ebef169a6a5354005476679c921961)

signer/tests/integration/bitcoin_poller.rs
cylewitruk: [a269262](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/a26926236048270428d1eb8142ce35b2a25a3115)

signer/src/bitcoin/rpc.rs
cylewitruk: [a269262](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/a26926236048270428d1eb8142ce35b2a25a3115)

signer/src/bitcoin/poller.rs
cylewitruk: Good call, [bf9eb10](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/bf9eb106abb10d1d98b5eaf7687069d5d8a4a556)

signer/src/bitcoin/poller.rs
cylewitruk: Leftover from an earlier version, removed in [f828f65](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/f828f65ead6447c751d5c5408789a84cd6e77701)

signer/src/bitcoin/poller.rs
cylewitruk: Yeah, I guess if we get error alarms it wouldn't be so nice: [ef507e7](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/ef507e760128a3a8de4115588db4bd36dabb1c06). We'll get errors from the usual suspects when the signer falls behind due to polling failing.

docker/mainnet/nodes/docker-compose.chains.yml
matteojug: There's also the `BITCOIN_RPC_PORT: *BITCOIN_RPC_PORT` at (ex) line 29 that was removed

signer/src/bitcoin/poller.rs
cylewitruk: Yeah that's true, it should hopefully basically never happen, but it's pretty easy to make it a little more resilient (i.e. address the "we'll try again in x seconds"):

[0c9c0d5](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/0c9c0d58bc566cbf081872ed8f1d70e2bc2db3ef)

signer/src/main.rs
cylewitruk: My thinking was that if, when the signer is starting up, it can't reach bitcoind within the timeout, that this will cause the container/signer to exit with an error which (hopefully) will trigger the more "standard operational monitoring" alarms at operators (i.e. orchestrator alarms, not log-based) but also alarms that we're not getting votes, etc.

I'm fine with removing it too, though, I was on the fence.

signer/tests/service-configs/bitcoin.conf
cylewitruk: [fe99854](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/fe998546b999e1497b2db4c4016d9f3bd89409ac)

docker/mainnet/nodes/docker-compose.chains.yml
cylewitruk: Very true, sorry read too quickly [00a7201](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/00a7201909b6d67c67046cd48e3d95a9bdc6bf84)

signer/src/config/mod.rs
cylewitruk: Good catch, [ea1442d](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/ea1442d1846d906702bc987e931ab43ce3d27712)

signer/src/bitcoin/poller.rs
matteojug: A downside of this new version is that if we don't have a subscriber we will log the error *a lot*. But in that case we are probably broken very badly or just in startup, so it seems fine

signer/src/main.rs
matteojug: What I find "strange" is that so far you could start the sbtc signer without any dependency on the stacks/bitcoin node, and it will start working as soon as those are reachable. With this change, we put a strict dependency on the bitcoin node.

It's true that without it you are not going to do anything meaningful, but still it's a different behaviour wrt the current more lenient one, and I'm not entirely sure it's a safe assumption to make. And since it also seem a kind of sanity check more than something we require in order to get the new things working, I'm a bit on the fence about it.

signer/tests/integration/bitcoin_poller.rs
cylewitruk: Sure, [5ad8e6c](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/5ad8e6c4cdf5c7512f00de22f674c173aef38ff4)

signer/src/config/mod.rs
cylewitruk: Good thing you asked, because I had forgotten to call `validate()` 🙈 [4a56f88](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/4a56f88dbaa60eef0f6e6740824abd2e8e357f97)

signer/src/config/mod.rs
```diff
@@ -776,6 +814,44 @@ mod tests {
...
+    fn config_errors_if_no_bitcoin_rpc_endpoints() {
+        clear_env();
+
+        set_var("SIGNER_BITCOIN__RPC_ENDPOINTS", "");
+
+        let error = Settings::new_from_default_config().unwrap_err();
+
+        match error {
+            ConfigError::Message(msg) => {
+                dbg!(&msg);
```
matteojug: nit: leftover `dbg`

matteojug (dismissed):

signer/src/config/serialization.rs
```diff
@@ -17,6 +17,9 @@ where
 {
     let mut v = Vec::new();
     for s in Vec::<String>::deserialize(deserializer)? {
+        if s.is_empty() {
+            continue;
+        }
```
djordon: Why continue on an empty string? I would think that an empty string is not a valid URL, so parsing should fail and therefore so should deserialization.

signer/src/main.rs
cylewitruk: Well, let's remove it and restore to the previous "startup no matter what" logic

signer/src/block_observer.rs
```diff
@@ -131,28 +132,38 @@ pub trait DepositRequestValidator {
...
+                tracing::debug!("block observer has received a shutdown signal");
                 break;
             }
 
             // Bitcoin blocks will generally arrive in ~10 minute intervals, so
             // we don't need to be so aggressive in our timeout here.
-            let poll = tokio::time::timeout(Duration::from_millis(100), self.bitcoin_blocks.next());
+            let poll = bitcoin_blocks
+                .next()
+                .with_timeout(Duration::from_millis(100));
```
djordon, line 162: In general, I prefer the `tokio::time::timeout(duration, future)` call over the abstracted variant, since it's already clear and takes up the same amount of space.

signer/src/config/mod.rs
cylewitruk: [a1bbad4](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/a1bbad4bf3200a8e74b56b00280e7955e2df74fe)

signer/src/main.rs
cylewitruk: Removed and simplified in [a617921](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/a617921c7c5e5f751e824af3c68233c7d33fb383)

signer/src/bitcoin/poller.rs
```diff
@@ -0,0 +1,137 @@
...
+    polling_interval: Duration,
+) where
+    Bitcoin: BitcoinInteract,
+{
+    let mut last_seen_hash = None;
+
+    loop {
+        match rpc.get_best_block_hash().await {
+            Ok(current_hash) => {
+                if Some(current_hash) != last_seen_hash {
```
djordon, line 72: 👍🏽 , much better

signer/src/bitcoin/poller.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /// Stops the background polling task.
+    pub fn stop(self) {
+        self.poller_task_handle.abort();
+    }
+}
+
+impl BitcoinBlockHashStreamProvider for BitcoinChainTipPoller {
+    type Error = BitcoinChainTipPollerError;
```
djordon, line 120: I think that we should keep things simple and use the regular crate `Error` here.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4775,8 +4776,7 @@ async fn sign_bitcoin_transaction_withdrawals() {
 
         let block_observer = BlockObserver {
             context: ctx.clone(),
-            bitcoin_blocks: testing::btc::new_zmq_block_hash_stream(BITCOIN_CORE_ZMQ_ENDPOINT)
-                .await,
+            bitcoin_block_source: bitcoin_chain_tip_poller.clone(),
```
djordon, line 4779: Why clone, can't we just do:
```suggestion
            bitcoin_block_source: BitcoinChainTipPoller::start_for_regtest().await,
```

signer/src/bitcoin/mod.rs
```diff
@@ -147,4 +147,30 @@ pub trait BitcoinInteract: Sync + Send {
...
+/// A trait for providing a stream of block hashes to be used by the block observer.
+///
+/// Implementors of this trait are responsible for sourcing block hash notifications,
+/// such as new chain tips from a Bitcoin Core node, and making them available as an
+/// asynchronous stream. This abstraction allows different components, such as
+/// block observers or test utilities, to consume block hash events without being
+/// coupled to the specific mechanism of how those events are obtained.
+pub trait BitcoinBlockHashStreamProvider: Send + Sync {
+    /// The error type that this provider can return.
+    type Error: std::error::Error;
```
djordon, line 164: I suppose it doesn't hurt, but why give the choice? Seems simpler to always return a specific error. In this case, we should just return the crate error here.

signer/src/bitcoin/poller.rs
cylewitruk: We can't, broadcast channels require messages to be `Clone`, which `Error` isn't.

signer/src/config/serialization.rs
cylewitruk: There's nothing inherently wrong with an empty string (extra or trailing comma in env variable, etc.), but what's wrong here is that if it hits an empty string, instead of hitting the actual "can't be empty" checks and outputting a user-friendly message, it errors with an obscure parsing error.

So, skipping empty strings let's the deserialization get past that and the validators to do their job.

signer/src/bitcoin/poller.rs
djordon: Well, we could use a mpsc based channel, since they do not have that limitation. If we had multiple consumers then this would be a no-go, but I don't think that we do.

signer/src/block_observer.rs
cylewitruk: Hm, I'm on the other side of the fence on this one. I find this syntax much clearer/easier on the eyes, more aligned with the functional, combinator-style patterns that `tokio` and `futures` already heavily promote, and it feels more idiomatic, being consistent with how we handle things like iterators etc.

signer/src/bitcoin/poller.rs
cylewitruk: The broadcast channel is used by e.g. tests to avoid creating many simultaneous pollers for multiple signer instances, and could be used by other tasks in the future, for example a metrics-updating task, etc.

I don't see any benefit to using `Error` -- it's an unnecessarily large and complex type to need to allocate for each message when the only error which can occur is a "lagging subscriber" error, and that's not an error that we (currently) propagate, only log.

signer/tests/integration/transaction_coordinator.rs
cylewitruk: To be able to `stop()` the poller once the block observer has exited so as to not leave dangling tasks around.

signer/src/config/serialization.rs
djordon: > [I]f it hits an empty string, instead of hitting the actual "can't be empty" checks and outputting a user-friendly message, it errors with an obscure parsing error.

Good point. Instead of continuing, we print a helpful message.

With this, if someone tries to configure more than one URL but only sets one with a trailing comma (or something), then this just silently ignores it. Although I'm not particularly worried about this scenario, I still feel uneasy about silently ignoring something that seems like a known invalid configuration. We should at least log a warning about this case, so that the operator has a chance to know about it. I'm on the fence about continuing though.

signer/src/bitcoin/mod.rs
cylewitruk: Same as above, it can't be the crate `Error` with broadcast streams. The primary reason for the associated type is to allow test implementations to use the error most convenient/appropriate for their context.

signer/tests/integration/transaction_coordinator.rs
djordon: All tasks are automatically cleaned up when the test ends and the runtime is dropped. There is no need to call stop or abort on any tasks.

signer/src/bitcoin/poller.rs
djordon: > The broadcast channel is used by e.g. tests to avoid creating many simultaneous pollers for multiple signer instances...

I didn't think that we did that in any of our tests.



> I don't see any benefit to using `Error` -- it's an unnecessarily large and complex type to need to allocate for each message when the only error which can occur is a "lagging subscriber" error, and that's not an error that we (currently) propagate, only log.

Good point. I think you have the better argument. Do forgive me, you know I love my big fat `Error` type.

signer/src/block_observer.rs
djordon: Yeah, it is quite nice actually.

We could also lean into existing functionality that we have from `tokio_stream` and friends. In particular, we could use [timeout_repeating](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.timeout). I'm not advocating that we do so, just that we have this functionality already with one of our dependencies.

The `tokio_stream` stuff is nice and all but it adds an additional layer on top of something that we already know, namely the `tokio::time::timeout(duration, future)` and our local stuff which are thin layers on top of it. So I'd feel roughly the same way about that as I do with our local extensions.

signer/tests/integration/transaction_coordinator.rs
cylewitruk: Yeah, I was just trying to be explicit about cleaning up since our other tasks generally listen to a `TerminationHandle` and shut themselves down. I don't really care either way here, though.

signer/src/config/serialization.rs
cylewitruk: The issue is that we can't really print a helpful message because we don't know the context of the field we're deserializing within the `Deserialize` impl.

For example, if I comment out the `continue` with the env var `SIGNER_BITCOIN__RPC_ENDPOINTS=""` then I'm left with:
```
&result = Err(
    relative URL without a base,
)
```
Which for an operator is quite ambiguous, vs. letting the validators do their thing:
```
&result = Err(
    [bitcoin.rpc_endpoints] At least one Bitcoin RPC endpoint must be provided,
)
```

Similarly, with the `continue` commented out and the env var `SIGNER_BITCOIN__RPC_ENDPOINTS="http://localhost:1234,"` (trailing comma, which I would consider valid):
```
&result = Err(
    relative URL without a base,
)
```
vs. successfully parsing to a `Vec` with one item with the `continue` in place.

signer/src/block_observer.rs
cylewitruk: Well, if you think about it, the call to `.next()` above is a combinator from `futures::StreamExt`. Same with `future_map()` and co. So it's a pattern that we already use fairly extensively.

Yeah the `tokio_stream` stuff is nice, but it doesn't always make for the easiest read. For this specific case, the simplicity and clear intent of `.with_timeout()`, `.sleep()` and friends make, imo, the light abstractions a worthwhile ergonomics/readability win without introducing complexity.

signer/src/config/serialization.rs
djordon: Well, we can say something, I just don't think that we should silently ignore anything.

signer/src/bitcoin/mod.rs
djordon: Right, we need clone for broadcast streams.

signer/src/config/serialization.rs
cylewitruk: I just don't see an empty string as an error, an extra or trailing comma is a common accidental "artifact" which in general is gracefully handled (look at rust or javascript, for example). It's better that the validators do their job, and the deserializer do its, especially when we can't provide any context in the deserializer.

djordon: I'm not sure that I agree here. For example, I would expect, say,
```toml
[emily]
endpoints = [
    "http://testApiKey@127.0.0.1:3031",
    "",
    ""
]
```
to be a configuration error when the application starts. If we're moving to explicitly accepting this, then we should log _something_. The log doesn't necessarily need to be here at the parsing/deserialization step, but I would think that it's uncommon to allow this.

matteojug (approved):

signer/src/config/serialization.rs
cylewitruk: Removed some of the stuff in [f14f5f6](https://github.com/stacks-sbtc/sbtc/pull/1780/commits/f14f5f671dffa7c520bb57592e1d6f9c6f64141b)

djordon (approved):

Issue #1690, opened 2025-05-27
[Chore]: Restrict fullfillment for non-completed Emily events


MCJOHN974:
# Chore - Restrict fullfillment for non-completed Emily events

## 1. Description

In current implementation only confirmed deposit requests can (from logical side of view) have a fulfillment attached to them. However we don't forbid to send update request with `fulfillment` and with not-completed status. We should change this, making Emily returning an error if it receives update for non-completed deposit with fulfillment.

Note: currently, Emily silently ignores `fulfillment` for non-completed events, so this PR is more about notifying request sender that they are doing something wrong, rather then caring about Emily state correctness. (However logically wrong requests still is a bad thing and it is nice to bail when we see them).

## 2. Why It's Necessary

We want Emily to report that this update request is invalid, instead of silently save it, dropping the `fulfillment`.

## 3. Execution Checklist

- [ ] Emily now bail on non-completed deposits with fulfillment
- [ ] Tests for this bailing

PR #1781, opened 2025-07-25
[chore]: restrict fullfillment for non-completed Emily events

MCJOHN974:
## Description

Closes: #1690

Logically, only `Confirmed` events can have a `fullfilment` field. With this PR Emily will reject update requests with non-None fullfillment and non-Confirmed status.

## Changes
- added new `ValidationError` enum variants
- return them with 400 Bad Request for requests with non-None fullfillment and non-Confirmed status
- tests for this

## Testing Information
- added 4 new tests

## Checklist

### DIFF
--- a/emily/handler/src/api/models/deposit/requests.rs
+++ b/emily/handler/src/api/models/deposit/requests.rs
@@ -150,11 +150,19 @@ impl DepositUpdate {
             bitcoin_tx_output_index: self.bitcoin_tx_output_index,
             bitcoin_txid: self.bitcoin_txid.clone(),
         };
-        // Only RBF transactions can have a replaced_by_tx.
+        // Only RBF deposits can have a replaced_by_tx.
         if self.status != DepositStatus::Rbf && self.replaced_by_tx.is_some() {
             return Err(error::ValidationError::InvalidReplacedByTxStatus(
                 self.status,
                 self.bitcoin_txid,
+                self.bitcoin_tx_output_index,
+            ));
+        }
+        // Only confirmed deposits can have a fulfillment.
+        if self.status != DepositStatus::Confirmed && self.fulfillment.is_some() {
+            return Err(error::ValidationError::DepositFulfillmentNotConfirmed(
+                self.status,
+                self.bitcoin_txid.clone(),
                 self.bitcoin_tx_output_index,
             ));
         }
@@ -228,6 +236,21 @@ impl UpdateDepositsRequestBody {
                         %bitcoin_txid,
                         bitcoin_tx_output_index,
                         "failed to update deposit: request missing fulfillment for completed request."
+                    );
+                    deposits.push((index, Err(error.clone())));
+                }
+                Err(
+                    ref error @ ValidationError::DepositFulfillmentNotConfirmed(
+                        ref status,
+                        ref txid,
+                        outindex,
+                    ),
+                ) => {
+                    tracing::warn!(
+                        %txid,
+                        bitcoin_tx_output_index = outindex,
+                        ?status,
+                        "failed to update deposit: fulfillment is only allowed for confirmed deposits.",
                     );
                     deposits.push((index, Err(error.clone())));
                 }
--- a/emily/handler/src/api/models/withdrawal/requests.rs
+++ b/emily/handler/src/api/models/withdrawal/requests.rs
@@ -75,6 +75,13 @@ impl WithdrawalUpdate {
         self,
         chainstate: Chainstate,
     ) -> Result<ValidatedWithdrawalUpdate, error::ValidationError> {
+        // Only confirmed withdrawals can have a fulfillment.
+        if self.status != WithdrawalStatus::Confirmed && self.fulfillment.is_some() {
+            return Err(ValidationError::WithdrawalFulfillmentNotConfirmed(
+                self.status,
+                self.request_id,
+            ));
+        }
         // Make status entry.
         let status_entry: WithdrawalStatusEntry = match self.status {
             WithdrawalStatus::Confirmed => {
@@ -136,6 +143,19 @@ impl UpdateWithdrawalsRequestBody {
                     tracing::warn!(
                         request_id,
                         "failed to update withdrawal: request missing fulfillment for completed request."
+                    );
+                    withdrawals.push((index, Err(error.clone())));
+                }
+                Err(
+                    ref error @ ValidationError::WithdrawalFulfillmentNotConfirmed(
+                        ref status,
+                        request_id,
+                    ),
+                ) => {
+                    tracing::warn!(
+                        request_id,
+                        ?status,
+                        "failed to update withdrawal: withdrawal is not confirmed, but has a fulfillment."
                     );
                     withdrawals.push((index, Err(error.clone())));
                 }
--- a/emily/handler/src/common/error.rs
+++ b/emily/handler/src/common/error.rs
@@ -19,7 +19,10 @@ use utoipa::ToSchema;
 use warp::{reject::Reject, reply::Reply};
 
 use crate::{
-    api::models::{chainstate::Chainstate, common::DepositStatus},
+    api::models::{
+        chainstate::Chainstate,
+        common::{DepositStatus, WithdrawalStatus},
+    },
     database::entries::{
         chainstate::ChainstateEntry, deposit::DepositEntryKey, withdrawal::WithdrawalEntryKey,
     },
@@ -67,6 +70,20 @@ pub enum ValidationError {
     /// The deposit has status RBF but is missing the replaced_by_tx field.
     #[error("missing replaced_by_tx for RBF deposit with txid: {0}, vout: {1}")]
     DepositMissingReplacementTx(String, u32),
+
+    /// The withdrawal have fulfillment data, but is not confirmed. Only confirmed withdrawals can have
+    /// fulfillment data.
+    #[error(
+        "withdrawal with fulfillment data must be confirmed, but got status {0:?} for request id: {1}"
+    )]
+    WithdrawalFulfillmentNotConfirmed(WithdrawalStatus, u64),
+
+    /// The deposit have fulfillment data, but is not confirmed. Only confirmed deposits can have
+    /// fulfillment data.
+    #[error(
+        "deposit with fulfillment data must be confirmed, but got status {0:?} for txid: {1}, vout: {2}"
+    )]
+    DepositFulfillmentNotConfirmed(DepositStatus, String, u32),
 }
 
 /// Errors from the internal API logic.
--- a/emily/handler/tests/integration/deposit.rs
+++ b/emily/handler/tests/integration/deposit.rs
@@ -1056,6 +1056,7 @@ async fn update_deposits_is_forbidden_for_signer(
         assert_eq!(deposits.len(), 1);
         let deposit = deposits.first().unwrap();
         assert_eq!(deposit.status, 403);
+        println!("Deposit: {:?}", deposit);
         assert!(deposit.deposit.clone().unwrap().is_none());
         assert_eq!(deposit.error.clone().unwrap().unwrap(), "Forbidden");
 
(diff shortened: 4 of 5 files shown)


emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -150,14 +150,22 @@ impl DepositUpdate {
...
                 self.status,
                 self.bitcoin_txid,
                 self.bitcoin_tx_output_index,
             ));
         }
+        // Only confirmed deposits can have a fulfillment.
+        if self.status != DepositStatus::Confirmed && self.fulfillment.is_some() {
+            return Err(error::ValidationError::DepositFulfillmentNotConfirmed(
+                self.status,
+                self.bitcoin_txid.clone(),
```
matteojug: nit: no need to clone

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -231,6 +239,21 @@ impl UpdateDepositsRequestBody {
                     );
                     deposits.push((index, Err(error.clone())));
                 }
+                Err(
+                    ref error @ ValidationError::DepositFulfillmentNotConfirmed(
+                        ref status,
+                        ref txid,
+                        outindex,
```
matteojug: nit: for consistency with the other arms
```suggestion
                        ref status,
                        ref bitcoin_txid,
                        bitcoin_tx_output_index,
```

emily/handler/tests/integration/deposit.rs
```diff
@@ -1056,6 +1056,7 @@ async fn update_deposits_is_forbidden_for_signer(
         assert_eq!(deposits.len(), 1);
         let deposit = deposits.first().unwrap();
         assert_eq!(deposit.status, 403);
+        println!("Deposit: {:?}", deposit);
```
matteojug: nit: leftover print

emily/handler/tests/integration/deposit.rs
```diff
@@ -1675,3 +1676,249 @@ async fn emily_process_deposit_updates_when_some_of_them_are_unknown() {
     .expect("Received an error after making a valid get deposits api call.");
     assert_eq!(deposits.deposits.len(), 1);
 }
+
+// All other deposit status updates are forbidden to signer and I don't think we
+// want strictly pin behaviour where Emily first validate request about fulfillment and
+// only after check if such request is allowed for signer.
+#[test_case(DepositStatus::Accepted; "accepted")]
+#[tokio::test]
+async fn only_completed_deposit_can_have_fulfillment_signer(status: DepositStatus) {
```
matteojug: this and the sidecar version of the test are identical except for which function is called during the updated. What about merging the two by adding another test parameter (eg, `is_sidecar`)? Also for withdrawal tests

emily/handler/tests/integration/deposit.rs
```diff
@@ -1675,3 +1676,249 @@ async fn emily_process_deposit_updates_when_some_of_them_are_unknown() {
...
+    .await;
+
+    if status != DepositStatus::Confirmed {
+        // Check response correctness
+        let response = response.expect("Batch update should return 200 OK");
+        let deposits = response.deposits;
+        assert_eq!(deposits.len(), 1);
+        let deposit = deposits.first().unwrap();
+        assert_eq!(deposit.status, 400);
+        println!("Deposit: {:?}", deposit);
```
matteojug: nit: leftover print

emily/handler/tests/integration/deposit.rs
```diff
@@ -1675,3 +1676,249 @@ async fn emily_process_deposit_updates_when_some_of_them_are_unknown() {
...
+        },
+    )
+    .await;
+
+    if status != DepositStatus::Confirmed {
+        // Check response correctness
+        let response = response.expect("Batch update should return 200 OK");
+        let deposits = response.deposits;
+        assert_eq!(deposits.len(), 1);
+        let deposit = deposits.first().unwrap();
```
matteojug: nit: this part seems shared between the two cases, we can move it before the if; also for withdrawal tests

emily/handler/tests/integration/deposit.rs
```diff
@@ -1675,3 +1676,249 @@ async fn emily_process_deposit_updates_when_some_of_them_are_unknown() {
...
+            .chars()
+            .enumerate()
+            .map(|(i, c)| {
+                if i == 0 {
+                    c.to_uppercase().next().unwrap()
+                } else {
+                    c
+                }
+            })
+            .collect();
```
matteojug: nit: we have a similar version of this around (used in `replaced_by_tx_for_not_rbf_transactions_is_bad_request`): can we move it to a test function (eg, `capitalize_str`) and use that around? Or maybe even better: can't we directly serialize a `DepositFulfillmentNotConfirmed` error and use that in the comparison? also for withdrawal tests

emily/handler/tests/integration/deposit.rs
```diff
@@ -1675,3 +1676,249 @@ async fn emily_process_deposit_updates_when_some_of_them_are_unknown() {
...
+
+        // Check that deposit wasn't updated
+        let response = apis::deposit_api::get_deposit(
+            &user_configuration,
+            &bitcoin_txid,
+            &bitcoin_tx_output_index.to_string(),
+        )
+        .await
+        .expect("Received an error after making a valid get deposit api call.");
+        assert_eq!(response.bitcoin_txid, bitcoin_txid);
```
matteojug: Shouldn't this check the `status` (and eventually also the fulfillment being `None`)? also for withdrawal tests

emily/handler/tests/integration/withdrawal.rs
```diff
@@ -1050,3 +1050,223 @@ async fn emily_process_withdrawal_updates_when_some_of_them_are_unknown() {
...
+        stacks_txid: "test_fulfillment_stacks_txid".to_string(),
+    })));
+
+    let response = apis::withdrawal_api::update_withdrawals_sidecar(
+        &user_configuration,
+        UpdateWithdrawalsRequestBody {
+            withdrawals: vec![WithdrawalUpdate {
+                request_id,
+                fulfillment: fulfillment.clone(),
+                status: status.clone(),
```
matteojug: nit: locally clippy complains about this: (also below)
```
try removing the `clone` call: `status`
```

emily/handler/tests/integration/deposit.rs
MCJOHN974: [51f6882](https://github.com/stacks-sbtc/sbtc/pull/1781/commits/51f68824c1444a36a65f82ef9de338a67eeb52a0)

MCJOHN974: and [fa1932f](https://github.com/stacks-sbtc/sbtc/pull/1781/commits/fa1932f73f2f654f0c69c2117ba9f8818d807739)

emily/handler/tests/integration/deposit.rs
MCJOHN974: Hmm, yes and no. The caveat here is that `update_deposits_signer` and `update_deposits_sidecar` actually have different return types (`Result<_, testing_emily_client::apis::Error<UpdateDepositsSidecarError>>`
   and `Result<_, testing_emily_client::apis::Error<UpdateDepositsSignerError>>`), thus, adding this `is_sidecar` parameter is not as easy as just write `let x = if is_sidecar { f(y) } else { g(y) }`, so probably code duplication here will be simpler.

emily/handler/tests/integration/deposit.rs
MCJOHN974: [c04b3a9](https://github.com/stacks-sbtc/sbtc/pull/1781/commits/c04b3a909be12f6dca8f1580cb0e1f64ccc95250)

emily/handler/tests/integration/deposit.rs
MCJOHN974: [2e9322b](https://github.com/stacks-sbtc/sbtc/pull/1781/commits/2e9322b62b9392e8abf02796d35d09bbe17b4bf3)

emily/handler/src/api/models/withdrawal/requests.rs
```diff
@@ -139,6 +146,19 @@ impl UpdateWithdrawalsRequestBody {
...
+                Err(
+                    ref error @ ValidationError::WithdrawalFulfillmentNotConfirmed(
+                        ref status,
+                        request_id,
+                    ),
+                ) => {
+                    tracing::warn!(
+                        request_id,
+                        ?status,
+                        "failed to update withdrawal: withdrawal is not confirmed, but has a fulfillment."
```
djordon: I think you meant this right?
```suggestion
                        "failed to update withdrawal: fulfillment data is only allowed for confirmed withdrawals."
```

emily/handler/src/common/error.rs
```diff
@@ -67,6 +70,20 @@ pub enum ValidationError {
     /// The deposit has status RBF but is missing the replaced_by_tx field.
     #[error("missing replaced_by_tx for RBF deposit with txid: {0}, vout: {1}")]
     DepositMissingReplacementTx(String, u32),
+
+    /// The withdrawal have fulfillment data, but is not confirmed. Only confirmed withdrawals can have
+    /// fulfillment data.
```
djordon:
```suggestion
    /// A withdrawal update had fulfillment data where the status was not `Confirmed`.
    /// Only confirmed withdrawals can have fulfillment data.
```

emily/handler/src/common/error.rs
```diff
@@ -67,6 +70,20 @@ pub enum ValidationError {
...
+    #[error(
+        "withdrawal with fulfillment data must be confirmed, but got status {0:?} for request id: {1}"
+    )]
+    WithdrawalFulfillmentNotConfirmed(WithdrawalStatus, u64),
+
+    /// The deposit have fulfillment data, but is not confirmed. Only confirmed deposits can have
+    /// fulfillment data.
```
djordon:
```suggestion
    /// A deposit update had fulfillment data where the status was not `Confirmed`.
    /// Only Confirmed deposits can have fulfillment data.
```

emily/handler/tests/integration/deposit.rs
djordon: Wait, it does seem like you can combine these by doing something like
```rust
let deposits = if is_sidecar {
    update_deposits_signer(y).deposits
} else {
    update_deposits_sidecar(y).deposits
}
```
These tests really seem identical except for that bit right?

djordon: Actually, they aren't identical except for that bit because of how they handle the different status updates. But still, might be worth taking an extra look to see if they can be combined.

emily/handler/src/api/models/deposit/requests.rs
```diff
@@ -231,6 +239,21 @@ impl UpdateDepositsRequestBody {
...
+                        ref status,
+                        ref bitcoin_txid,
+                        bitcoin_tx_output_index,
+                    ),
+                ) => {
+                    tracing::warn!(
+                        %bitcoin_txid,
+                        bitcoin_tx_output_index,
+                        ?status,
+                        "failed to update deposit: fulfillment is only allowed for confirmed deposits.",
```
djordon:
```suggestion
                        "failed to update deposit: fulfillment data is only allowed for confirmed deposits.",
```

emily/handler/src/api/models/withdrawal/requests.rs
MCJOHN974: tbh, for me first version of this error message sounds clearer, but English is not my native language, so I believe that your version is better =)

emily/handler/tests/integration/deposit.rs
```diff
@@ -1673,3 +1673,120 @@ async fn emily_process_deposit_updates_when_some_of_them_are_unknown() {
...
+        assert!(matches!(response.status, DepositStatus::Pending));
+        assert!(response.fulfillment.is_none());
+    } else {
+        let deposit = deposit.deposit.clone().unwrap().unwrap();
```
djordon, line 1788: We should probably check that `assert_eq!(deposit.status, 200)` before the unwrap right? Same thing for withdrawals.

djordon (commented):
Looks good. Just had one small comment about the tests.

emily/handler/tests/integration/deposit.rs
MCJOHN974: [6b1dabd](https://github.com/stacks-sbtc/sbtc/pull/1781/commits/6b1dabdfce9ed99f92ec15ae9d0a862ef128a264)

djordon: We should add the same thing for the withdrawal tests.

MCJOHN974: [6b1dabd](https://github.com/stacks-sbtc/sbtc/pull/1781/commits/6b1dabdfce9ed99f92ec15ae9d0a862ef128a264)

matteojug (approved):

Issue #1742, opened 2025-06-17
[Feat]: Add periodic logging of stacks and bitcoin node details


cylewitruk:
# Feat - Add periodic logging of stacks and bitcoin node details

## 1. Description

To help with troubleshooting, we should periodically log the Stacks and Bitcoin node details (version, etc). 

## 2. Why It's Necessary

This would have been helpful recently when one of the signers' Stacks nodes ran into issues during a Bitcoin fork. Being able to determine which node versions they were running would have helped to better understand if the issue was caused to a missed upgrade or something else.

Example code for retrieving this information from the nodes exists in the `api/info` endpoint in the signer, making this a trivial add.

## 3. Technical Details / Acceptance Criteria

- The Bitcoin and Stacks node versions, and potentially their current chain tip/height, should be `debug`-logged upon [some trigger], where the trigger is...
    - Perhaps a new top-level async task that logs "things" at an interval. This would prevent disruptions in e.g. block observation from impeding the logging.
    - Or within the block observer, with the known caveat that if we stop getting blocks for whatever reason that the information will not be logged.
- We could also push this as a metric string (I think?) if that could be useful for operators. We don't receive metric telemetry from partner signers, so it would be of less use to us directly.

PR #1787, opened 2025-07-28
[feat]: add node info logger && enable debug logs in devenv

MCJOHN974:
## Description

Closes: #1742

## Changes

- adds `BlockchainNodeInfoLogger` which runs in it's own thread, polls bitcoin and stacks node time to time and log info about them
- mostly reusing code from `api/info`

## Testing Information

I manually runned devenv and confirmed that necessary logs are printed.

## Note

This PR enables debug logging for signer in devenv. It was not stated in the issue, but change is relatively small so I added it to this PR.

## Checklist


signer/src/logging.rs
```diff
@@ -43,3 +49,43 @@ fn setup_logging_pretty(directives: &str) {
         .with(main_layer)
         .init()
 }
+
+/// Logs to standard logging stream information about Bitcoin and Stacks
+/// node versions, chaintips, dkg rounds, etc.
+async fn log_blockchain_nodes_info<C: Context>(ctx: &C) {
+    let info = build_info(ctx).await;
+    tracing::debug!(?info, "logging blockchain info",);
+}
```
matteojug: This will log a lot of info, I'm wondering if we really need those (eg, the config). Also note that it runs `get_stacks_chain_tip`, which unfortunately is not a fast query. 

Maybe instead of the full info we could call only `populate_bitcoin_node_info` and `populate_stacks_node_info`.

signer/src/main.rs
```diff
@@ -344,6 +346,14 @@ async fn run_block_observer(ctx: impl Context) -> Result<(), Error> {
     block_observer.run().await
 }
 
+/// Run the blockchain info logger event loop.
+async fn run_blockchain_info_logger(ctx: impl Context) -> Result<(), Error> {
+    // TODO: make timout a config parameter.
+    let logger = BlockchainInfoLogger::new(ctx, Duration::from_secs(60));
```
matteojug: Can you move this `60` to a top level const? And maybe bump it to 10m? We probably are fine with less frequent logs for this

signer/src/main.rs
```diff
@@ -146,6 +147,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         run_checked(run_request_decider, &context),
         run_checked(run_transaction_coordinator, &context),
         run_checked(run_transaction_signer, &context),
+        run_checked(run_blockchain_info_logger, &context),
```
matteojug: We shouldn't use `run_checked` for this: if `run_blockchain_info_logger` fails I don't think we want to stop everything else.

signer/src/logging.rs
```diff
@@ -43,3 +49,43 @@ fn setup_logging_pretty(directives: &str) {
...
+where
+    C: Context,
+{
+    /// Creates new BlockchainInfoLogger with given context and timeout.
+    pub fn new(context: C, timeout: Duration) -> Self {
+        Self { context, timeout }
+    }
+    /// Runs BlockchainInfoLogger which will log info about blockchain nodes
+    /// each timeout.
+    pub async fn run(self) -> Result<(), Error> {
```
matteojug: Why a result for `run`? It doesn't seem to be fallible

signer/src/logging.rs
MCJOHN974: I agree that a lot of this info looks not necessary, but maybe it could be case where it can be useful, and about node performance given the fact that we don't really have a lot of deposits/withdrawals and we do not expect process them in seconds, maybe I'd choose here to have more info.

signer/src/main.rs
cylewitruk: I'd say even 1hr would be fine

signer/src/logging.rs
```diff
@@ -43,3 +49,43 @@ fn setup_logging_pretty(directives: &str) {
...
+async fn log_blockchain_nodes_info<C: Context>(ctx: &C) {
+    let info = build_info(ctx).await;
+    tracing::debug!(?info, "logging blockchain info",);
+}
+
+/// Simple struct for time to time writing logs
+/// about Stacks and Bitcoin nodes state.
+pub struct BlockchainInfoLogger<Context> {
+    /// Signer context.
+    context: Context,
```
djordon: We should use a different type parameter name here, since `Context` is often a common struct name. Instead, we should use the usual convention of using a single capital letter for the type parameter, especially since there is only one here.

Alternatively, we could use the expected concrete type since we there is only one.

signer/src/logging.rs
MCJOHN974: [24be6d0](https://github.com/stacks-sbtc/sbtc/pull/1787/commits/24be6d056cfd911089de863a7e88f12ac4c2a580)

signer/src/main.rs
MCJOHN974: [8f185f1](https://github.com/stacks-sbtc/sbtc/pull/1787/commits/8f185f104e51a875b4e5a11b8cf95ad9228a8902)

signer/src/main.rs
MCJOHN974: [8f185f1](https://github.com/stacks-sbtc/sbtc/pull/1787/commits/8f185f104e51a875b4e5a11b8cf95ad9228a8902)

signer/src/logging.rs
```diff
@@ -43,3 +49,43 @@ fn setup_logging_pretty(directives: &str) {
...
+    }
+    /// Runs BlockchainInfoLogger which will log info about blockchain nodes
+    /// each timeout.
+    pub async fn run(self) -> Result<(), Error> {
+        let term = self.context.get_termination_handle();
+        loop {
+            if term.shutdown_signalled() {
+                break;
+            }
+            tokio::time::sleep(self.timeout).await;
```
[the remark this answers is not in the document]
MCJOHN974: When timeout was a minute it was not so a problem but now it totally makes sense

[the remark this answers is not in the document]
MCJOHN974: [7ad0b45](https://github.com/stacks-sbtc/sbtc/pull/1787/commits/7ad0b45c0dd5627952fab6c175775025a9cb18c8)

signer/src/main.rs
```diff
@@ -344,6 +346,14 @@ async fn run_block_observer(ctx: impl Context) -> Result<(), Error> {
     block_observer.run().await
 }
 
+/// Run the blockchain info logger event loop.
+async fn run_blockchain_info_logger(ctx: impl Context) -> Result<(), Error> {
+    // TODO: make timout a config parameter.
```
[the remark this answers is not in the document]
MCJOHN974: [5e87fbc](https://github.com/stacks-sbtc/sbtc/pull/1787/commits/5e87fbcbdd69a2da3046cc995dc6d06c4a308a78)

signer/src/logging.rs
MCJOHN974: Esplecially since we decided to move with 1h timeout I don't think 1 extra heavy request per hour will make any visible difference, as well as some extra logs once an hour will not make logs dirty

signer/src/main.rs
matteojug: Can you add a comment to separate this from the run_checked ones above? To make it explicit that the signer doesn't need run_blockchain_info_logger to be operational

signer/src/logging.rs
```diff
@@ -43,3 +48,53 @@ fn setup_logging_pretty(directives: &str) {
...
+/// Logs to standard logging stream information about Bitcoin and Stacks
+/// node versions, chaintips, dkg rounds, etc.
+async fn log_blockchain_nodes_info<C: Context>(ctx: &C) {
+    let info = build_info(ctx).await;
+    tracing::debug!(?info, "logging blockchain info",);
+}
+
+/// Simple struct for time to time writing logs
+/// about Stacks and Bitcoin nodes state.
+pub struct BlockchainInfoLogger<Context> {
```
matteojug: nit: this is not limited to nodes state, also in `run` docstring below

signer/src/main.rs
```diff
@@ -39,6 +40,11 @@ use tracing::Span;
 // before proceeding.
 const INITIAL_BOOTSTRAP_DELAY_SECS: u64 = 3;
 
+// Timeout after which blockchain info logger will print new log.
+// Currently chosen to be 1 hour.
+// TODO: make timeout a config parameter.
+const BLOCKCHAIN_INFO_LOGGER_TIMEOUT: Duration = Duration::from_secs(3600);
```
matteojug: nit: s/timeout/interval/? also in `signer/src/logging.rs`

signer/src/logging.rs
```diff
@@ -43,3 +48,53 @@ fn setup_logging_pretty(directives: &str) {
         .with(main_layer)
         .init()
 }
+
+/// Logs to standard logging stream information about Bitcoin and Stacks
+/// node versions, chaintips, dkg rounds, etc.
+async fn log_blockchain_nodes_info<C: Context>(ctx: &C) {
+    let info = build_info(ctx).await;
+    tracing::debug!(?info, "logging blockchain info",);
```
matteojug: nit?
```suggestion
    tracing::debug!(?info, "signer info");
```

signer/src/logging.rs
```diff
@@ -43,3 +48,53 @@ fn setup_logging_pretty(directives: &str) {
...
+    /// Runs BlockchainInfoLogger which will log info about blockchain nodes
+    /// each timeout.
+    pub async fn run(self) {
+        let term = self.context.get_termination_handle();
+
+        // Without this counter and each_nth_iteration_print_log,
+        // if self.timeout is big, tokio::time::sleep will sleep for
+        // full timeout, and it will not check if termination was activated
+        // for a long time.
+        let each_nth_iteration_print_log = self.timeout.as_secs();
```
matteojug: We can use a `tokio::select!`, we do something similar for the p2p event loop (https://github.com/stacks-sbtc/sbtc/blob/9507bd4a41bee4cf038fa30f02628df1aeb14bb9/signer/src/network/libp2p/event_loop.rs#L255-L256)

signer/src/main.rs
MCJOHN974: [1dbf501](https://github.com/stacks-sbtc/sbtc/pull/1787/commits/1dbf501230d1ef82b3dc4fef1c5e760dc2fb7382)

signer/src/logging.rs
MCJOHN974: [1dbf501](https://github.com/stacks-sbtc/sbtc/pull/1787/commits/1dbf501230d1ef82b3dc4fef1c5e760dc2fb7382)

signer/src/main.rs
MCJOHN974: [1dbf501](https://github.com/stacks-sbtc/sbtc/pull/1787/commits/1dbf501230d1ef82b3dc4fef1c5e760dc2fb7382)

signer/src/logging.rs
```diff
@@ -43,3 +48,54 @@ fn setup_logging_pretty(directives: &str) {
         .with(main_layer)
         .init()
 }
+
+/// Logs to standard logging stream information about Bitcoin and Stacks
+/// node versions, chaintips, dkg rounds, etc.
+async fn log_blockchain_nodes_info<C: Context>(ctx: &C) {
+    let info = build_info(ctx).await;
+    tracing::debug!(?info, "signer info");
```
djordon: It would be nice to log this as a JSON object or something, but my searches having come up with anything nice and easy. So we can ship with this and change it later.

signer/src/logging.rs
MCJOHN974: [e5fc0f9](https://github.com/stacks-sbtc/sbtc/pull/1787/commits/e5fc0f9c2ff80f18c6a325b350109420190a7cb2)

signer/src/logging.rs
MCJOHN974: [59a3ffe](https://github.com/stacks-sbtc/sbtc/pull/1787/commits/59a3ffe72b3883c1a7477c424623a2b27dc11a38)

matteojug (approved):

djordon (dismissed):

signer/src/logging.rs
MCJOHN974: I think [02713cb](https://github.com/stacks-sbtc/sbtc/pull/1787/commits/02713cb674193ab4b5ae6ed79609df061467f925) can do what we want but I want to test in devenv before merge

signer/src/logging.rs
```diff
@@ -43,3 +48,47 @@ fn setup_logging_pretty(directives: &str) {
         .with(main_layer)
         .init()
 }
+
+/// Logs to standard logging stream information about Bitcoin and Stacks
+/// node versions, chaintips, dkg rounds, etc.
+async fn log_blockchain_nodes_info<C: Context>(ctx: &C) {
+    let info = build_info(ctx).await;
+    let json = serde_json::to_string(&info).unwrap_or_else(|_| "{}".to_string());
```
djordon: I think you want this to be:
```suggestion
    let json = serde_json::to_string(&info).unwrap_or_else(|_| format!("{info:?}"));
```

signer/src/logging.rs
```diff
@@ -43,3 +48,48 @@ fn setup_logging_pretty(directives: &str) {
         .with(main_layer)
         .init()
 }
+
+/// Logs to standard logging stream information about Bitcoin and Stacks
+/// node versions, chaintips, dkg rounds, etc.
+async fn log_blockchain_nodes_info<C: Context>(ctx: &C) {
+    let info = build_info(ctx).await;
+    let json = serde_json::to_string(&info).unwrap_or_else(|_| format!("{info:?}"));
+    tracing::debug!(%json, "signer info");
```
djordon:
```suggestion
    let json = serde_json::to_string(&info).unwrap_or_else(|_| format!("{info:?}"));
    tracing::debug!(info = %json, "signer info");
```

djordon (approved):

Issue #1778, opened 2025-07-22
[Feature]: Upgrade WSTS dependency version


djordon:
# Feature - Update WSTS dependency version

## 1. Description

Once https://github.com/stacks-sbtc/wsts/pull/168 and https://github.com/stacks-sbtc/wsts/pull/165 are merged, we'll want to use the new version of WSTS since the `main` branch on WSTS includes the following security enhancements:
* https://github.com/stacks-sbtc/wsts/pull/159
* https://github.com/stacks-sbtc/wsts/pull/157 (parts of it, it's been partially reverted)
* https://github.com/stacks-sbtc/wsts/pull/164
* https://github.com/stacks-sbtc/wsts/pull/160

## 2. Technical Details

There have been one or two API breaking changes and https://github.com/stacks-sbtc/wsts/pull/168 is a protocol breaking change, so we'll need to update the protobufs and the conversion implementations when we do the update.

### 2.1 Acceptance Criteria

- [ ] We are using a newer version of WSTS with the desired changes
- [ ] We pass in the sign ID and DKG ID with the new API

PR #1789, opened 2025-07-28
chore: update wsts dependency

xoloki:
## Description
Update the `WSTS` dependency

Closes: #1778 

## Changes
This PR updates the `WSTS` dependency to include numerous security audit fixes and feature additions.  

## Testing Information
No new functionality has been added to `sBTC`, so as long as existing tests pass the new dependency should be working.

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ stackslib = { git = "https://github.com/stacks-network/stacks-core", rev = "4977
 stacks-common = { git = "https://github.com/stacks-network/stacks-core", rev = "49777d3fd73a6dbb610be80c376b7d9389c9871a", default-features = false, features = ["canonical"] }
 
 # Trust Machines Dependencies
-wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "6cf8f349234279b82a7cf8867812a65697d5706b" }
+wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "afec464dbcf4321c93a56aa92a08c3f9bb586705", features = ["testing"] }
 
 # Crates.io
 aquamarine = { version = "0.6.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ stackslib = { git = "https://github.com/stacks-network/stacks-core", rev = "4977
 stacks-common = { git = "https://github.com/stacks-network/stacks-core", rev = "49777d3fd73a6dbb610be80c376b7d9389c9871a", default-features = false, features = ["canonical"] }
 
 # Trust Machines Dependencies
-wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "afec464dbcf4321c93a56aa92a08c3f9bb586705", features = ["testing"] }
+wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "f354dedeb0f7bede14575553be6d35454309cf02", features = ["testing"] }
 
 # Crates.io
 aquamarine = { version = "0.6.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ stackslib = { git = "https://github.com/stacks-network/stacks-core", rev = "4977
 stacks-common = { git = "https://github.com/stacks-network/stacks-core", rev = "49777d3fd73a6dbb610be80c376b7d9389c9871a", default-features = false, features = ["canonical"] }
 
 # Trust Machines Dependencies
-wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "f354dedeb0f7bede14575553be6d35454309cf02", features = ["testing"] }
+wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "f354dedeb0f7bede14575553be6d35454309cf02" }
 
 # Crates.io
 aquamarine = { version = "0.6.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ stackslib = { git = "https://github.com/stacks-network/stacks-core", rev = "4977
 stacks-common = { git = "https://github.com/stacks-network/stacks-core", rev = "49777d3fd73a6dbb610be80c376b7d9389c9871a", default-features = false, features = ["canonical"] }
 
 # Trust Machines Dependencies
-wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "11e2316aa0edf996f8e8f9ccbbbfd4bd2975ed1d" }
+wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "6cf8f349234279b82a7cf8867812a65697d5706b" }
 
 # Crates.io
 aquamarine = { version = "0.6.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ stackslib = { git = "https://github.com/stacks-network/stacks-core", rev = "4977
 stacks-common = { git = "https://github.com/stacks-network/stacks-core", rev = "49777d3fd73a6dbb610be80c376b7d9389c9871a", default-features = false, features = ["canonical"] }
 
 # Trust Machines Dependencies
-wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "f354dedeb0f7bede14575553be6d35454309cf02" }
+wsts = { git = "https://github.com/stacks-sbtc/wsts", rev = "f354dedeb0f7bede14575553be6d35454309cf02" }
 
 # Crates.io
 aquamarine = { version = "0.6.0", default-features = false }
--- a/protobufs/crypto/wsts/wsts.proto
+++ b/protobufs/crypto/wsts/wsts.proto
@@ -24,7 +24,7 @@ message SignerDkgPublicShares {
   uint32 signer_id = 2;
   // List of (signer_id, commitment)
   repeated PartyCommitment commitments = 3;
-  // kex exchange public key
+  // The public key used for exchanging secret shares during DKG
   crypto.Point kex_public_key = 4;
 }
 
--- a/protobufs/crypto/wsts/wsts.proto
+++ b/protobufs/crypto/wsts/wsts.proto
@@ -28,6 +28,22 @@ message SignerDkgPublicShares {
   crypto.Point kex_public_key = 4;
 }
 
+// DKG public shares message from a signer to all signers and coordinator.
+// This type is used by the signers for reading polynomial commitments from
+// the database. For sBTC, this vector should always have a length of 1.
+// This maps to this type:
+// <https://github.com/Trust-Machines/wsts/blob/2d6cb87218bb8dd9ed0519356afe57a0b9a697cb/src/net.rs#L137-L146>
+message DkgSignerCommitment {
+  // List of (signer_id, commitment)
+  repeated PartyCommitment commitment = 3;
+}
+
+// All of the polynomial commitments received by a signer during a DKG
+// round.
+message DkgPolynomialCommitments {
+  map<uint32, DkgSignerCommitment> commitments = 1;
+}
+
 // The public polynomial committed to by one of the party members who are
 // participating in distributed key generation.
 // This maps to this type <https://github.com/Trust-Machines/wsts/blob/2d6cb87218bb8dd9ed0519356afe57a0b9a697cb/src/net.rs#L144-L145>
--- a/protobufs/crypto/wsts/wsts.proto
+++ b/protobufs/crypto/wsts/wsts.proto
@@ -24,6 +24,8 @@ message SignerDkgPublicShares {
   uint32 signer_id = 2;
   // List of (signer_id, commitment)
   repeated PartyCommitment commitments = 3;
+  // kex exchange public key
+  crypto.Point kex_public_key = 4;
 }
 
 // The public polynomial committed to by one of the party members who are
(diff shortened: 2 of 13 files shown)


supply-chain/audits.toml
```diff
@@ -48,3 +48,9 @@ who = "djordon <dan.jordon@gmail.com>"
 criteria = "safe-to-deploy"
 delta = "14.0.0 -> 14.0.0@git:11e2316aa0edf996f8e8f9ccbbbfd4bd2975ed1d"
 importable = false
+
+[[audits.wsts]]
+who = "djordon <dan.jordon@gmail.com>"
```
djordon: How did my name end up here?

signer/src/wsts_state_machine.rs
```diff
@@ -569,6 +592,8 @@ impl SignerStateMachine {
         )
         .map_err(Error::Wsts)?;
 
+        inner.verify_packet_sigs = false;
```
djordon: We should add a comment here explaining why this is the correct thing for us to do.

signer/src/proto/convert.rs
```diff
@@ -1593,14 +1593,20 @@ impl From<DkgPublicShares> for proto::SignerDkgPublicShares {
...
     }
 }
 
 impl TryFrom<proto::SignerDkgPublicShares> for DkgPublicShares {
     type Error = Error;
     fn try_from(value: proto::SignerDkgPublicShares) -> Result<Self, Self::Error> {
+        let kex_public_key = match value.kex_public_key {
+            Some(key) => Point::try_from(key)?,
+            None => Point::new(),
+        };
```
djordon: We should use `Point::identity()` instead of `Point::new()`, since it's a little clearer what we're getting. Also, we should add a comment about why we're defaulting to this point and how it is very likely to be okay because peers cannot deserialize the point at infinity.

Cargo.toml
```diff
@@ -28,7 +28,7 @@ stackslib = { git = "https://github.com/stacks-network/stacks-core", rev = "4977
 stacks-common = { git = "https://github.com/stacks-network/stacks-core", rev = "49777d3fd73a6dbb610be80c376b7d9389c9871a", default-features = false, features = ["canonical"] }
 
 # Trust Machines Dependencies
-wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "11e2316aa0edf996f8e8f9ccbbbfd4bd2975ed1d" }
+wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "f354dedeb0f7bede14575553be6d35454309cf02", features = ["testing"] }
```
djordon: We should add the testing feature here, but instead add the line
```toml
wsts = { workspace = true, features = ["testing"] }
```
to the dev-dependencies section in `signer/Cargo.toml`.

signer/src/transaction_coordinator.rs
```diff
@@ -1725,7 +1725,16 @@ where
...
+                .get_config()
+                .public_keys
+                .signers
+                .iter()
+                .map(|(id, key)| {
+                    let public_key = PublicKey::from(key);
+                    (*id, p256k1::point::Point::from(public_key))
+                })
+                .collect();
             let public_key_point = p256k1::point::Point::from(msg_public_key);
```
djordon: This can be simplified by just embracing the p256k1 `PublicKey` type here and doing something like
```suggestion
            let public_keys = &coordinator.get_config().public_keys.signers;
            let public_key_point = p256k1::keys::PublicKey::from(msg_public_key);
```
and then modifying the signature of `authenticate_message` to expect `p256k1::keys::PublicKey`s instead of `p256k1::point::Point`s.

signer/src/wsts_state_machine.rs
```diff
@@ -249,21 +249,35 @@ impl WstsCoordinator for FireCoordinator {
...
+            .map(|(id, key)| (id + 1, key))
             .collect();
+        let signer_key_ids: HashMap<u32, HashSet<u32>> = signers
+            .iter()
+            .map(|(&signer_id, _)| {
+                let mut keys = HashSet::new();
+                keys.insert(signer_id + 1);
+                (signer_id, keys)
+            })
+            .collect();
```
djordon: What is wrong with what was there before?

supply-chain/audits.toml
xoloki: You were the primary reviewer for all of the PRs in this series of commits.  How do you think we should represent that?

signer/src/wsts_state_machine.rs
xoloki: Done

signer/src/proto/convert.rs
xoloki: Done

Cargo.toml
xoloki: Done

signer/src/transaction_coordinator.rs
xoloki: Done

supply-chain/audits.toml
xoloki: Done.  

I noticed that most entries in this file use github usernames, while the ones which `cargo vet inspect` added had my full name.  Should I replace all of those with github usernames?

signer/src/wsts_state_machine.rs
xoloki: `signer_key_ids` has a different format in the `state_machine::PublicKeys` struct than it did in the old `coordinator::Config` struct

xoloki: Oh wait, not it doesn't.  Maybe I just copypastad this unnecessarily?

xoloki: Removed this an another apparently unnecessary change from this fn

xoloki: Removed the same changes from `wsts.rs`

xoloki: Found same unnecessary changes in the FrostCoordinator part of `wsts_state_machine.rs`

supply-chain/audits.toml
xoloki: I went ahead and changed all `cargo vet inspect` added name fields to use github usernames, as per the existing entries

protobufs/crypto/wsts/wsts.proto
```diff
@@ -24,6 +24,8 @@ message SignerDkgPublicShares {
   uint32 signer_id = 2;
   // List of (signer_id, commitment)
   repeated PartyCommitment commitments = 3;
+  // kex exchange public key
```
djordon:
```suggestion
  // The public key used for exchanging secret shares during DKG.
```

signer/src/proto/convert.rs
```diff
@@ -1593,14 +1593,21 @@ impl From<DkgPublicShares> for proto::SignerDkgPublicShares {
...
         }
     }
 }
 
 impl TryFrom<proto::SignerDkgPublicShares> for DkgPublicShares {
     type Error = Error;
     fn try_from(value: proto::SignerDkgPublicShares) -> Result<Self, Self::Error> {
+        let kex_public_key = match value.kex_public_key {
+            Some(key) => Point::try_from(key)?,
+            // use point at infinity since this value will fail deserialization
```
djordon: This would make me have even more questions, like "why would we want to fail deserialization", and "what does this have to do with deserialization, we've already deserialized the protobuf by the time that we are here", and things like that. In a few months, I may not be able to remember enough context to answer these questions and it may not be clear that what is here is correct -- or correct anymore given some other changes that have taken place. So the goal is to write enough so that someone who wasn't around for the teams discussions or doesn't recall them, like myself in a few months, could know whether this is correct and non-harmful.

supply-chain/audits.toml
djordon: Either is fine, I think that I always went with whatever cargo vet filled in for me.

protobufs/crypto/wsts/wsts.proto
xoloki: Done. I didn't commit the suggestion because it ended in a period.

signer/src/proto/convert.rs
xoloki: Done

signer/src/proto/convert.rs
```diff
@@ -1593,14 +1593,25 @@ impl From<DkgPublicShares> for proto::SignerDkgPublicShares {
...
 
 impl TryFrom<proto::SignerDkgPublicShares> for DkgPublicShares {
     type Error = Error;
     fn try_from(value: proto::SignerDkgPublicShares) -> Result<Self, Self::Error> {
+        let kex_public_key = match value.kex_public_key {
+            Some(key) => Point::try_from(key)?,
+            // If the protobuf doesn't have a kex_public_key, that means it is an artifact
+            // from an old DKG run before we had ephemeral DKG encryption keys.  Using the
+            // point at infinity allows us to distinguish this case in code if desired, and
+            // also prevents us from accidentally using these shares in a future DKG found,
```
djordon: 🙏🏽 
```suggestion
            // also prevents us from accidentally using these shares in a future DKG round,
```

signer/src/proto/convert.rs
```diff
@@ -1593,14 +1593,25 @@ impl From<DkgPublicShares> for proto::SignerDkgPublicShares {
...
+            // since the point at infinity will not deserialize if sent over the wire.
+            None => Point::identity(),
```
djordon: Hmmmm, suppose a malicious signer does not send `kex_public_key` during DKG, can they end up revealing some information somehow?

djordon: If it's possible for someone to select a "dangerous public key", such as the generator, or the point at infinity, or the negation of someone else's public key, then we should also change the key exchange protocol to include a proof that the signer controls the corresponding private key. Not sure how necessary that is though.

signer/src/proto/convert.rs
xoloki: Done.  I updated some other text to make things more consistent and clear.

signer/src/proto/convert.rs
xoloki: I can't think of any way that sending a protobuf with a null `kex_public_key` would expose information.  With 100% participation requirement DKG would halt, with less than 100% requirement the user would be excluded.  Either way no one is sending data to that user.

If a malicious signer instead used a "dangerous" public key, like the generator or some small multiple of it, then all of the private shares to and from the signer in question would be exposed.  But this is information that the malicious signer would have anyway.  A PoK of the private key wouldn't help, because the private key would be known (it's just be a small scalar).  

Given the cardinality of a scalar, I don't think we need to worry about this happening accidentally.  If we're worried about 1 in 2^256 probabilities, then everything in BTC, STX, and all of crypto is suspect.

djordon: Oh right, that makes sense. They'd reveal what would be private to them, and perhaps some bits of other signer's key exchange private key. But that would be fine, since they key exchange private key is ephermeral.

signer/src/proto/convert.rs
```diff
@@ -1593,14 +1593,25 @@ impl From<DkgPublicShares> for proto::SignerDkgPublicShares {
...
+            kex_public_key: Some(proto::Point::from(value.kex_public_key)),
         }
     }
 }
 
 impl TryFrom<proto::SignerDkgPublicShares> for DkgPublicShares {
     type Error = Error;
     fn try_from(value: proto::SignerDkgPublicShares) -> Result<Self, Self::Error> {
+        let kex_public_key = match value.kex_public_key {
+            Some(key) => Point::try_from(key)?,
```
djordon: I just realized that there is a simpler way. We should:
1. Create a new protobuf that is like the existing `proto::SignerDkgPublicShares` type.
2. Create a new local-to-signer crate type that is like the existing `wsts::net::DkgPublicShares` type.
3. Use the new types for serializing and deserializing to and from the database.
4. Add implementations for transforming the new protobuf type to the new local `DkgPublicShares` type.
5. Require the `kex_public_key` here.

This seems like a cleaner thing than we are currently going for, even though it's more complex. I don't mind doing the leg work on this one, but before I do so, do you have any thoughts on this?

xoloki: It sounds like a much bigger diff, and I'm not sure what it gains us.  Letting protobuf handle optional fields seems like a reasonable path here and in any future protocol changes with existing data that's been serialized to the DB.  This is pretty much the use case for that functionality.

But if it appeals to you from a cleanliness point of view, and you're willing to do the legwork, I woudn't argue against it strongly.

djordon: Yeah, but I really want to make the `kex_public_key` a required field. I'll branch off of your branch, write the code and see what people think. If it looks like too much we can scrap it.

signer/src/proto/convert.rs
```diff
@@ -1587,12 +1588,26 @@ impl TryFrom<proto::PartyCommitment> for (u32, PolyCommitment) {
...
+    }
+}
+
 impl From<DkgPublicShares> for proto::SignerDkgPublicShares {
     fn from(value: DkgPublicShares) -> Self {
         proto::SignerDkgPublicShares {
             dkg_id: value.dkg_id,
             signer_id: value.signer_id,
             commitments: value.comms.into_iter().map(|v| v.into()).collect(),
+            kex_public_key: Some(proto::Point::from(value.kex_public_key)),
```
matteojug: nit:
```suggestion
            kex_public_key: Some(value.kex_public_key.into()),
```

Cargo.toml
```diff
@@ -28,7 +28,7 @@ stackslib = { git = "https://github.com/stacks-network/stacks-core", rev = "4977
 stacks-common = { git = "https://github.com/stacks-network/stacks-core", rev = "49777d3fd73a6dbb610be80c376b7d9389c9871a", default-features = false, features = ["canonical"] }
 
 # Trust Machines Dependencies
-wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "11e2316aa0edf996f8e8f9ccbbbfd4bd2975ed1d" }
+wsts = { git = "https://github.com/Trust-Machines/wsts", rev = "f354dedeb0f7bede14575553be6d35454309cf02" }
```
matteojug: nit
```suggestion
wsts = { git = "https://github.com/stacks-sbtc/wsts", rev = "f354dedeb0f7bede14575553be6d35454309cf02" }
```

matteojug (approved):

matteojug (approved):

djordon (approved):

PR #1790, opened 2025-07-28
chore: bump stacks to 3.2

matteojug:
## Description

Porting of https://github.com/stacks-sbtc/sbtc/pull/1782 for main

> This PR updates the Stacks bindings in preparation for Epoch 3.2. This is needed primarily due to our use of the stacks-core API response types which use enums (e.g. `StacksBlockId::Epoch32` does not exist and results in a deserialization error).

## Testing Information

Devenv test run worked fine.

### DIFF
--- a/supply-chain/config.toml
+++ b/supply-chain/config.toml
@@ -306,7 +306,7 @@ version = "0.10.1"
 criteria = "safe-to-deploy"
 
 [[exemptions.chrono]]
-version = "0.4.38"
+version = "0.4.19"
 criteria = "safe-to-deploy"
 
 [[exemptions.clang-sys]]
@@ -339,10 +339,6 @@ criteria = "safe-to-deploy"
 
 [[exemptions.const-oid]]
 version = "0.9.6"
-criteria = "safe-to-deploy"
-
-[[exemptions.const_fn]]
-version = "0.4.10"
 criteria = "safe-to-deploy"
 
 [[exemptions.core2]]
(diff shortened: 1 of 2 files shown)


signer/src/stacks/api.rs
```diff
@@ -88,11 +89,13 @@ const CURRENT_AGGREGATE_PUBKEY_DATA_VAR_NAME: &str = "current-aggregate-pubkey";
...
-    0,
-    TokenTransferMemo([0; 34]),
-);
+static DUMMY_STX_TRANSFER_PAYLOAD: LazyLock<TransactionPayload> = LazyLock::new(|| {
+    TransactionPayload::TokenTransfer(
+        PrincipalData::Standard(StandardPrincipalData::null_principal()),
+        0,
+        TokenTransferMemo([0; 34]),
+    )
+});
```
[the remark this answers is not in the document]
matteojug: we can't directly init a principal anymore

cylewitruk (approved):
I haven't run it either, but looks to be in-line with the hotfix.

signer/src/stacks/api.rs
```diff
@@ -88,11 +89,13 @@ const CURRENT_AGGREGATE_PUBKEY_DATA_VAR_NAME: &str = "current-aggregate-pubkey";
 
 /// This is a dummy STX transfer payload used only for estimating STX
 /// transfer costs.
-const DUMMY_STX_TRANSFER_PAYLOAD: TransactionPayload = TransactionPayload::TokenTransfer(
-    PrincipalData::Standard(StandardPrincipalData(0, [0; 20])),
-    0,
-    TokenTransferMemo([0; 34]),
-);
+static DUMMY_STX_TRANSFER_PAYLOAD: LazyLock<TransactionPayload> = LazyLock::new(|| {
```
matteojug, line 92: The only difference from the original one, since clippy complained about internally mutable const.

cylewitruk: Ah right, makes sense

PR #1793, opened 2025-07-29
feat: consistent waiting per observed bitcoin block

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1792

## Changes

1. Pass the observed chain tip from the block observer to the other processes
2. Bail if the chain tip in the state differs from the passed chain tip after sleeping in the request decider and the transaction coordinator.

## Checklist

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -326,7 +326,7 @@ where
         // The bitcoin chain tip could have changed since we observed the
         // bitcoin block given as an input here. If so, we can safely skip
         // processing this block since other signers are likely to ignore
-        // us if we are the coordinator.
+        // us.
         if bitcoin_chain_tip != state_chain_tip {
             tracing::info!(
                 state_bitcoin_tip_hash = %state_chain_tip.block_hash,


signer/src/request_decider.rs
```diff
@@ -125,27 +126,29 @@ where
...
-            .state()
-            .bitcoin_chain_tip()
-            .ok_or(Error::NoChainTip)?
-            .block_hash;
+        // If the bitcoin chain tip has changed since the we observed
+        // the bitcoin block given as an input here, then we can safely
+        // bail, as we will be processing the new block shortly.
+        let state_chain_tip = self.context.state().bitcoin_chain_tip();
+        if Some(block_ref) != state_chain_tip {
+            tracing::debug!("chain tip has changed, skipping request processing");
```
matteojug: Bailing here we may skip some decision broadcast retries, and since the default retry window is pretty short (3 blocks) I'm not completely sure we can "safely" bail.

matteojug: nit: can we add `state_chain_tip` to the debug log?

signer/src/context/signer_context.rs
```diff
@@ -237,10 +238,10 @@ mod tests {
             task_started_clone.notify_one();
             let signal = cloned_receiver.recv().await.unwrap();
 
-            assert_eq!(
+            assert!(matches!(
```
matteojug: nit: recently we added `assert_matches` to the signers, we could start using it in new code (and eventually migrate old occurrences)

signer/src/transaction_coordinator.rs
```diff
@@ -302,20 +302,16 @@ where
...
-            "bitcoin_tip_hash",
-            tracing::field::display(bitcoin_chain_tip.block_hash),
-        );
-        span.record("bitcoin_tip_height", *bitcoin_chain_tip.block_height);
+        // The bitcoin chain tip could have changed since the we observed
+        // the bitcoin block given as an input here. If so, we can safely
+        // skip processing this block since other signers are likely to
+        // ignore us if we are the coordinator.
+        if Some(bitcoin_chain_tip) != self.context.state().bitcoin_chain_tip() {
+            tracing::debug!("bitcoin chain tip has changed, skipping processing");
```
matteojug: nit: can we add the state chain tip in the log?

signer/src/transaction_coordinator.rs
```diff
@@ -2755,6 +2753,60 @@ mod tests {
         more_asserts::assert_gt!(start.elapsed(), delay + baseline_elapsed);
     }
 
+    /// Check that we skip processing bitcoin blocks if the chain tip in
+    /// the state doesn't match the block hash passed in.
+    ///
+    /// Note: this test is a little sensitive to the current logic that
+    /// checks for smart contract deployment after checking whether the
+    /// chain tip is up to date.
```
matteojug, line 2821: I'm wondering if we should move the `if !self.is_coordinator(bitcoin_chain_tip.as_ref())` block before the sleep:
 - if we are not the coordinator, there's no point in sleeping
 - if we think we are the coordinator, then after the sleep we will check against the state for stale chain tips

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -5950,7 +5969,7 @@ async fn should_handle_dkg_coordination_failure() {
     // that 'should_coordinate_dkg' will trigger and we set the
     // 'dkg_max_duration' to 10 milliseconds we expect that
     // DKG will run & fail
-    let result = coordinator.process_new_blocks().await;
+    let result = coordinator.process_new_blocks(chain_tip).await;
```
matteojug: qq (not related to the change): is the `coordinator` actually the coordinator for `chain_tip`? Otherwise this test could succeed just because coordinator early return not being the coordinator for this tenure

signer/src/request_decider.rs
djordon: Good point, easy remove.

djordon: I'm going to remove this.

signer/src/transaction_coordinator.rs
djordon: Good point, easy win.

signer/tests/integration/transaction_coordinator.rs
djordon: Good point, I think it needs to check so I'll update it. I need to update my test for this as well.

djordon: BTW, the event loop definitely wasn't coordinator because it was configured to use a random private key, so it wasn't even in the signing set.

signer/src/request_decider.rs
djordon: I removed the waiting in https://github.com/stacks-sbtc/sbtc/pull/1793/commits/ae297b93e2f22fb4d37b8b5a7810ed6354f6cefc

djordon: Waiting, and the log, is not longer there https://github.com/stacks-sbtc/sbtc/pull/1793/commits/ae297b93e2f22fb4d37b8b5a7810ed6354f6cefc

signer/src/context/signer_context.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1793/commits/473442fb74a4465d37c8489e70b03639d649fb32

signer/src/transaction_coordinator.rs
djordon: Added in https://github.com/stacks-sbtc/sbtc/pull/1793/commits/9af6f1f86af616ba89b84b3a63f17c0efb3ce2cd

signer/src/transaction_coordinator.rs
djordon: Made the change in https://github.com/stacks-sbtc/sbtc/pull/1793/commits/8d645952329d87f3384ee18aa55b15819d64c4c3

signer/tests/integration/transaction_coordinator.rs
djordon: Fixed this test in https://github.com/stacks-sbtc/sbtc/pull/1793/commits/fff839666498bf87fc37809dc18a6bcb11edb491. I also updated the new unit test in https://github.com/stacks-sbtc/sbtc/pull/1793/commits/0c717eece995d7fa45d7d771dac0893c44a2ef11

signer/src/transaction_coordinator.rs
```diff
@@ -2753,14 +2758,62 @@ mod tests {
...
+        assert!(ev.is_coordinator(&chain_tip2.block_hash));
+
+        // We should bail if the chain tip is not set.
+        let error = ev.process_new_blocks(chain_tip2).await.unwrap_err();
+        assert_matches::assert_matches!(error, Error::NoChainTip);
+
+        // Now the chain tip in the state is set but it does not match the
+        // chain tip passed in, so we should bail early and return Ok(())
+        ctx.state().set_bitcoin_chain_tip(chain_tip1);
+        ev.process_new_blocks(chain_tip2).await.unwrap();
```
matteojug, line 2807: Can you also add the match err if we try to process `chain_tip2` after setting that in the state (or process `chain_tip1`)? Otherwise is not obvious if/what should fail, and isn't failing because of the early return.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -5945,34 +5986,41 @@ async fn should_handle_dkg_coordination_failure() {
...
+    // continues since there is signer set info to fallback on. Note that
+    // the coordinator should be emitting lots of error logs when
+    // constructing bitcoin or stacks transactions, but the coordinator
+    // should continue through all of that.
+    coordinator
+        .process_new_blocks(chain_tip)
+        .await
+        .expect("process_new_blocks should complete successfully even with DKG failure");
+
+    testing::storage::drop_db(db).await;
```
matteojug: This seems doing the same check as the previous `process_new_blocks`, how does that differ exactly? The fallback that is mentioned shouldn't apply also to the previous iteration?

matteojug (approved):
New changes LGTM, just a nit

signer/tests/integration/transaction_coordinator.rs
djordon: I don't think that it differs at all either. I suspect, as it was before this change, was supposed to change something and then call `process_new_blocks` again but I'm not sure what that was supposed to be. I'll just remove it.

signer/src/transaction_coordinator.rs
djordon: Yeah, I had it earlier, and removed it because the test seems a little fragile, but I can try adding it back in.

djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1793/commits/d4ef43eee0f43155038d6c839481709766b4bf05

signer/tests/integration/transaction_coordinator.rs
djordon: I removed it in https://github.com/stacks-sbtc/sbtc/pull/1793/commits/cbc450ee3b96693c64e5995d5f46b9806f6a971c

signer/src/context/signer_context.rs
cylewitruk: For context, `assert_matches!` is superior because `assert!(matches!(..))` will only tell you that the `matches!(..)` returned `false`, whereas `assert_matches!` will actually expand the match statement so you can see the contents which failed.

signer/src/transaction_coordinator.rs
matteojug: ~Nice, just the comment is maybe missing some word or some copypaste went wrong?~
Nvm, I forgot how to read

signer/src/transaction_coordinator.rs
```diff
@@ -330,6 +309,35 @@ where
...
+
+        // If we need to bail here then there is some bug in the code,
+        // since `process_new_blocks` should only be called after the state
+        // has been updated with the bitcoin chain tip.
+        let Some(state_chain_tip) = self.context.state().bitcoin_chain_tip() else {
+            tracing::error!("no bitcoin chain tip in state, skipping processing");
+            return Err(Error::NoChainTip);
+        };
+
+        // The bitcoin chain tip could have changed since the we observed
```
cylewitruk:
```suggestion
        // The bitcoin chain tip could have changed since we observed
```

signer/src/transaction_coordinator.rs
```diff
@@ -1826,19 +1834,12 @@ where
...
-    // it typically starts with a lot of leading zeros.
-    //
-    // Note that this function is technically not fallible,
-    // but for now we have chosen to return phantom errors
-    // instead of adding expects/unwraps in the code.
-    // Ideally the code should be formulated in a way to guarantee
-    // it being infallible without relying on sequentially coupling
-    // expressions. However, that is left for future work.
-    fn is_coordinator(&self, bitcoin_chain_tip: &model::BitcoinBlockHash) -> bool {
+    /// Determine if this signer is the signer set's coordinator.
```
cylewitruk:
```suggestion
    /// Determine if this signer is the signer set's coordinator for the specified Bitcoin block hash.
```

MCJOHN974 (approved):

signer/src/transaction_coordinator.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1793/commits/bbfa917b12ec394c9c9bd3420a86ae4f694c96db

signer/src/transaction_coordinator.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1793/commits/bbfa917b12ec394c9c9bd3420a86ae4f694c96db

matteojug (approved):

cylewitruk (approved):

PR #1795, opened 2025-07-29
feat: skip running DKG if config differs from registry but latest shares matches config

matteojug:
## Description

Closes: https://github.com/stacks-sbtc/sbtc/issues/1720
Superseeds: https://github.com/stacks-sbtc/sbtc/pull/1773

## Changes

If the config doesn't match the registry we run DKG, but only if latest shares do not match already the new config. Also ignores failed shares.

Add `get_latest_non_failed_dkg_shares` which does what you can guess.

## Testing Information

Add tests for `get_latest_non_failed_dkg_shares`, refactored a test to include the new condition.

## Checklist

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -2537,32 +2537,46 @@ pub async fn should_coordinate_dkg(
     let storage = context.get_storage();
     let config = context.config();
 
+    let latest_dkg_shares = storage.get_latest_non_failed_dkg_shares().await?;
+    let Some(latest_dkg_shares) = latest_dkg_shares else {
+        tracing::info!("no non-failed DKG shares exist; proceeding with DKG");
+        return Ok(true);
+    };
+
     // If the latest shares are unverified, we want to prioritize verifying them
     // instead of doing new DKG rounds. If we fail to do so they will eventually
     // be marked as failed, and we will resume DKG-ing.
-    let latest_dkg_shares = storage.get_latest_encrypted_dkg_shares().await?;
-    if latest_dkg_shares.map(|s| s.dkg_shares_status) == Some(model::DkgSharesStatus::Unverified) {
+    if latest_dkg_shares.dkg_shares_status == model::DkgSharesStatus::Unverified {
         tracing::debug!("latest shares are unverified; skipping DKG");
         return Ok(false);
     }
 
-    // If we do not have a key rotation event in the database, we will
-    // allow DKG below if we have not run DKG yet.
+    // If the registry has signer set info, we may need to run DKG based on it
     if let Some(registry_signer_info) = context.state().registry_signer_set_info() {
-        // Trigger DKG if signatures_required has changed
-        if registry_signer_info.signatures_required != config.signer.bootstrap_signatures_required {
-            tracing::info!("signatures required has changed; proceeding with DKG");
-            return Ok(true);
-        }
-
-        // Trigger DKG if signer set changes
-        if registry_signer_info.signer_set != config.signer.bootstrap_signing_set {
-            tracing::info!("signer set has changed; proceeding with DKG");
-            return Ok(true);
+        // If the registry differs from the config we may need to run DKG
+        if registry_signer_info.signatures_required != config.signer.bootstrap_signatures_required
+            || registry_signer_info.signer_set != config.signer.bootstrap_signing_set
+        {
+            // If we don't have new shares for the config already, we need DKG
+            if latest_dkg_shares.signature_share_threshold
+                != config.signer.bootstrap_signatures_required
+                || latest_dkg_shares.signer_set_public_keys() != config.signer.bootstrap_signing_set
+            {
+                tracing::info!(
+                    "signer set config differs from registry and latest DKG shares; proceeding with DKG"
+                );
+                return Ok(true);
+            } else {
+                tracing::debug!(
+                    "signer set config differs from registry, but we alreay have verified shares for it; checking other conditions"
+                );
+            }
         }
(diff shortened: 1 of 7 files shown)


signer/tests/integration/transaction_coordinator.rs
```diff
@@ -1169,26 +1155,45 @@ async fn run_dkg_if_signatures_required_changes(change_signatures_required: bool
...
+    if scenario.latest_shares_matches {
+        // Latest shares must match the config
+        let mut signer_set: Vec<PublicKey> = ctx
+            .config()
+            .signer
+            .bootstrap_signing_set
+            .iter()
+            .copied()
+            .collect();
+        signer_set.sort();
```
djordon: I don't think the sort is necessary since we are iterating over a `BTreeSet`, but I suppose it doesn't hurt.

djordon (approved):
Nice, looks good.

signer/tests/integration/transaction_coordinator.rs
matteojug: Oh right, removed in [357097a](https://github.com/stacks-sbtc/sbtc/pull/1795/commits/357097a45684e7eb30df19ca22596bc326e1f661)

djordon (approved):

signer/src/transaction_coordinator.rs
```diff
@@ -2537,32 +2537,46 @@ pub async fn should_coordinate_dkg(
...
+            } else {
+                tracing::debug!(
+                    "signer set config differs from registry, but we already have verified shares for it; checking other conditions"
+                );
+            }
         }
     }
 
-    // Get the number of DKG shares that have been stored
+    // Finally, we may need to run DKG because of config target rounds.
```
cylewitruk, line 2577: Tbh I think I'd prefer to get rid of the "target rounds" altogether and instead have an implicit "1 DKG round (verified|unverified) allowed >= the configured `dkg_min_bitcoin_block_height`".

We had the target round impl as a sanity/safety-guard for the first re-dkg's, but at this point it complicates the config update communications with signers (short of just saying "increment by 1").

Maybe we'll need to bite the bullet for this next release since this needs to get in, so just throwing it out there, feel free to resolve.

signer/src/transaction_signer.rs
```diff
@@ -1588,32 +1588,46 @@ pub async fn assert_allow_dkg_begin(
...
-    if latest_dkg_shares.map(|s| s.dkg_shares_status) == Some(model::DkgSharesStatus::Unverified) {
+    if latest_dkg_shares.dkg_shares_status == model::DkgSharesStatus::Unverified {
         tracing::warn!("latest shares are unverified; aborting");
         return Err(Error::DkgHasAlreadyRun);
     }
 
-    // If we do not have a key rotation event in the database, we will
-    // allow DKG below if we have not run DKG yet.
+    // If the registry has signer set info, we may need to run DKG based on it
     if let Some(registry_signer_info) = context.state().registry_signer_set_info() {
```
cylewitruk, line 1606: nit: Feels like we're duplicating a fair bit between here and the coordinator, maybe we could extract some of these checks to a common "place" to be more DRY?

signer/tests/integration/postgres.rs
```diff
@@ -2064,6 +2064,74 @@ async fn get_last_verified_dkg_shares_does_whats_advertised() {
     signer::testing::storage::drop_db(db).await;
 }
 
+/// The [`DbRead::get_latest_non_failed_dkg_shares`] function is supposed to
+/// fetch the last encrypted DKG shares with status not 'failed' from the
+/// database.
+#[tokio::test]
+async fn get_latest_non_failed_dkg_shares_does_whats_advertised() {
```
cylewitruk, line 2071: :) (nit, feel free to skip)
```suggestion
async fn get_latest_non_failed_dkg_shares_behaves_as_advertised() {
```

signer/src/transaction_coordinator.rs
matteojug: Yeah, especially now that we can have DKG for the signer set change the target round is quite a mess to handle, what you're suggesting was the consensus IIUC last time we talked about it.

But I would keep that out of this release, to avoid adding more stuff to test before the release -- if we need to do another DKG right after we have a way to do it, it's just that everybody may need a custom value for target.

signer/src/transaction_signer.rs
matteojug: We could move the `registry_signer_info` block out and use the same, but ideally we should have the whole logic in a single place, and then handle the different logs in the caller.

I suggest doing the full refactoring when we remove the target rounds. But happy to move that part out in this PR if anybody prefers it.

signer/tests/integration/postgres.rs
matteojug: Lazy copied from `get_last_verified_dkg_shares_does_whats_advertised`, I'll change in this PR based on the other comments, otherwise I'll squeeze it in another PR

signer/src/transaction_coordinator.rs
matteojug: Opened https://github.com/stacks-sbtc/sbtc/issues/1797 to track it

signer/src/transaction_signer.rs
matteojug: Opened https://github.com/stacks-sbtc/sbtc/issues/1797 to track it

signer/tests/integration/postgres.rs
matteojug: No changes, so I'll do it later 🙏

PR #1796, opened 2025-07-30
feat: make key exchange public key required

djordon:
## Description

This builds off of https://github.com/stacks-sbtc/sbtc/pull/1789. The goal here was to make the key exchange public key required during DKG.

The issue is that we cannot just require key exchange public key on our existing types, since that wouldn't allow us to read existing shares from the database. So, to solve that issues, we read a protobuf type that is different from the one that is written. This allows us to better separate the relevant WSTS type during DKG from is read from in the database.

Let me know if you don't think that going down this route is a good idea. We can keep https://github.com/stacks-sbtc/sbtc/pull/1789 as is or do something else. I don't feel strongly about it.

## Changes

* Add new protobuf types that are used when reading from the database.
* Make the key exchange public key a required field in `wsts::net::DkgPublicShares`.

## Testing Information

I added a unit test that checks the relevant logic of serializing one thing and deserializing another. Also, the existing unit tests should exercise this easily enough.

## Checklist

### DIFF
--- a/protobufs/crypto/wsts/wsts.proto
+++ b/protobufs/crypto/wsts/wsts.proto
@@ -28,6 +28,22 @@ message SignerDkgPublicShares {
   crypto.Point kex_public_key = 4;
 }
 
+// DKG public shares message from a signer to all signers and coordinator.
+// This type is used by the signers for reading polynomial commitments from
+// the database. For sBTC, this vector should always have a length of 1.
+// This maps to this type:
+// <https://github.com/Trust-Machines/wsts/blob/2d6cb87218bb8dd9ed0519356afe57a0b9a697cb/src/net.rs#L137-L146>
+message DkgSignerCommitment {
+  // List of (signer_id, commitment)
+  repeated PartyCommitment commitment = 3;
+}
+
+// All of the polynomial commitments received by a signer during a DKG
+// round.
+message DkgPolynomialCommitments {
+  map<uint32, DkgSignerCommitment> commitments = 1;
+}
+
 // The public polynomial committed to by one of the party members who are
 // participating in distributed key generation.
 // This maps to this type <https://github.com/Trust-Machines/wsts/blob/2d6cb87218bb8dd9ed0519356afe57a0b9a697cb/src/net.rs#L144-L145>
--- a/signer/src/wsts_state_machine.rs
+++ b/signer/src/wsts_state_machine.rs
@@ -3,6 +3,7 @@
 use std::collections::BTreeMap;
 use std::future::Future;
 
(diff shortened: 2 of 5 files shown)


signer/src/wsts_state_machine.rs
```diff
@@ -40,6 +43,31 @@ use wsts::state_machine::coordinator::frost;
 use wsts::traits::Signer as _;
 use wsts::v2::Aggregator;
 
+/// A database model for storing DKG public shares.
+///
+/// This is used to store the DKG public shares in the database.
+#[derive(Clone, Debug, PartialEq, Default)]
+pub struct DkgPublicSharesDb {
```
cylewitruk: I'd name this `DbDkgPublicShares` (or `WstsPublicShares` or just `PublicShares`) and move to `storage/model` (suffixing with `Db` makes it sound like it itself is a database).

cylewitruk (commented):
I agree in principal (application owning its db models), though I would probably go all-the-way and create separate model types specifically for what we need to store in the database vs. involving protobuf and other wsts message types (which have the potential to change as well).

protobufs/crypto/wsts/wsts.proto
```diff
@@ -28,6 +28,22 @@ message SignerDkgPublicShares {
   crypto.Point kex_public_key = 4;
 }
 
+// DKG public shares message from a signer to all signers and coordinator.
+// This type is used by the signers for reading polynomial commitments from
+// the database. For sBTC, this vector should always have a length of 1.
+// This maps to this type:
+// <https://github.com/Trust-Machines/wsts/blob/2d6cb87218bb8dd9ed0519356afe57a0b9a697cb/src/net.rs#L137-L146>
+message DkgSignerCommitment {
+  // List of (signer_id, commitment)
```
xoloki, line 37: If a signer sends more than one of these, it's indexed on `party_id` not `signer_id`:
```suggestion
  // List of (party_id, commitment)
```

djordon: Although calling it the `party_id` is more correct from a WSTS perspective, sBTC only uses the WSTS v2 system, so being more specific and calling it the `signer_id` seems to be a bit better.

signer/src/wsts_state_machine.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1796/commits/37f2403f0c7e6b00db77cd9b8d36a5a20198d1a7

matteojug (approved):
Approach LGTM, waiting for this to land into the other for the full review

Issue #1749, opened 2025-06-18
[Chore]: Fix `fetch_unknown_ancestors` to support testnet backfilling from scratch


matteojug:
# Chore - Fix `fetch_unknown_ancestors` to support testnet backfilling from scratch 

## 1. Description

In the current implementation of `fetch_unknown_ancestors`, we assume _continuous_ bitcoin blocks anchors around the nakamoto start height: if a stacks block is anchored to a bitcoin block with height greater than `nakamoto_start_height`, then its parent will be anchored to a bitcoin block of height greater than or equal `nakamoto_start_height`. We rely on this assumption to then check if we reached an anchor at height `nakamoto_start_height` and, if so, stop the backfilling.

The issue is that this assumption is no longer true for the current Stacks testnet:
> The height of the first bitcoin block anchoring stacks blocks in Epoch 3 is block 1901, where there are lots of stacks blocks but none in the bitcoin block immediately before and none after for about 98 bitcoin blocks. Here are the [stacks blocks for bitcoin block 1901](https://explorer.hiro.so/btcblock/0x5cb8fc024a7c7f40c8890c92f15803dfb4fe3406d047db9e2a8492727127f00c?chain=testnet), the last stacks block in that tenure is block 320. The previous stacks block, block 319, is indeed [anchored to bitcoin block 1865](https://explorer.hiro.so/block/0x0d7cb8c66040d87fc17f39e1b5c36bc7fb5c4d97cc611a168e2cca186848be1e?chain=testnet).

So, when trying to fetch the tenure of the parent of the last Stacks block anchored to a bitcoin block with height more than `nakamoto_start_height`, we try to get the tenure for a pre-Nakamoto block, getting an error from the Stacks node.

## 2. Why It's Necessary

Booting up a sBTC signer on testnet, it will not be possible to successfully finish the Stacks blocks backfill. To fix this, we usually insert some blocks manually, eg:
```sql
insert into "sbtc_signer"."stacks_blocks" ("bitcoin_anchor", "block_hash", "block_height", "created_at", "parent_hash") 
values 
[...]

PR #1803, opened 2025-08-01
Test Backfill Fix (draft)

setzeus:
## Description
This PR introduces a check into the signer/stacks/api that verifies whether there is a large gap in pre-Nakamoto anchor blocks. As seen in #1749, these gaps in testnet require a manual massage / inserting to continue syncing. 

This initial lay down contains logical checks for *when* blocks should be manually inserted for a successful testnet run.

Closes: #1749 

## Changes
The 'fetch_unknown_ancestor' in the signer/stacks/api.rs file.

## Testing Information
No tests as of now but this will change.

## Checklist

### DIFF
--- a/signer/src/stacks/api.rs
+++ b/signer/src/stacks/api.rs
@@ -1345,16 +1345,12 @@ where
 
 // Helper function for backfill testnet logic, checks if an
 // error indicates we're trying to fetch a pre-Nakamoto block
-fn is_pre_nakamoto_block_error(
-    error: &Error,
-    network: NetworkKind,
-) -> bool {
+fn is_pre_nakamoto_block_error(error: &Error, network: NetworkKind) -> bool {
     match error {
         // If we get a 404, it's likely because the block ID refers to a pre-Nakamoto block
         // But only if we're not on mainnet (since this is a testnet-specific issue)
         Error::StacksNodeResponse(req_error) => {
-            req_error.status() == Some(reqwest::StatusCode::NOT_FOUND)
-                && !network.is_mainnet()
+            req_error.status() == Some(reqwest::StatusCode::NOT_FOUND) && !network.is_mainnet()
         }
         // For other errors, we're not sure so we don't assume it's pre-Nakamoto
         _ => false,


signer/src/stacks/api.rs
```diff
@@ -1309,15 +1309,84 @@ where
...
+                        parent_anchor_height = %parent_tenure.anchor_block_height,
+                        gap = %anchor_gap,
+                        "detected large gap in bitcoin anchor heights during backfill"
+                    );
+                    // TODO: this is where we'll manually massage / insert blocks if in testnet
+                    // Increment a metric for monitoring
+                    metrics::counter!(
+                        "stacks_backfill_anchor_gaps_total",
+                        "gap_size" => if anchor_gap > 100 { "large" } else { "medium" }
+                    ).increment(1);
```
djordon: We can probably just remove the logging and metrics here. I don't think we really care if all is well.

signer/src/stacks/api.rs
```diff
@@ -1309,15 +1309,84 @@ where
...
+                        parent_block_id = %header.parent_block_id,
+                        current_anchor_height = %tenure.anchor_block_height,
+                        error = %error,
+                        "parent block appears to be pre-Nakamoto, stop backfill"
+                    );
+
+                    // Increment metric for monitoring pre-Nakamoto stops
+                    metrics::counter!(
+                        "stacks_backfill_pre_nakamoto_stops_total"
+                    ).increment(1);
```
djordon: No need for the metric, this should only happen once.

signer/src/stacks/api.rs
```diff
@@ -1309,15 +1309,84 @@ where
...
+                    metrics::counter!(
+                        "stacks_backfill_anchor_gaps_total",
+                        "gap_size" => if anchor_gap > 100 { "large" } else { "medium" }
+                    ).increment(1);
+                }
+                headers.push(parent_tenure);
+            }
+            Err(error) => {
+                // Check if it's a pre-Nakamoto block error
+                if is_pre_nakamoto_block_error(&error) {
```
djordon:
```suggestion
            Err(error) if is_pre_nakamoto_block_error(&error) => {
```

signer/src/stacks/api.rs
```diff
@@ -1309,15 +1309,84 @@ where
...
+                    break;
+                } else {
+                    // Propogate other errors
+                    tracing::error!(
+                        parent_block_id = %header.parent_block_id,
+                        error = %error,
+                        "failed to fetch parent tenure during backfill"
+                    );
+                    return Err(error);
+                }
```
djordon: We probably don't need this log, since we'll get one later as well.
```suggestion
                Err(error) => return Err(error),
```

signer/src/stacks/api.rs
```diff
@@ -1309,15 +1309,84 @@ where
...
+        Error::StacksNodeResponse(req_error) => {
+            req_error.status() == Some(reqwest::StatusCode::NOT_FOUND)
+        }
+        // Deserialization errors can indicate pre-Nakamoto blocks, but they could
+        // also indicate corruption or network issues, so we're being conservative here
+        Error::DecodeNakamotoBlock(_, _) | Error::DecodeNakamotoTenure(_, _) => {
+            // Note: These could also indicate corrupted data, not just pre-Nakamoto blocks
+            // The caller logs this as a warning to help with debugging
+            true
+        }
```
djordon: I don't think we want these in here.

signer/src/stacks/api.rs
```diff
@@ -1309,15 +1309,84 @@ where
...
 }
 
+// Helper function for backfill testnet logic, checks if an 
+// error indicates we're trying to fetch a pre-Nakamoto block
+fn is_pre_nakamoto_block_error(error: &Error) -> bool {
+    match error {
+        // If we get a 404, it's likely because the block ID refers to a pre-Nakamoto block
+        Error::StacksNodeResponse(req_error) => {
+            req_error.status() == Some(reqwest::StatusCode::NOT_FOUND)
+        }
```
djordon: Hmm, there must be some other check that we can do to make this whole thing tighter; right now every missing block ID is treated as if we are in pre-Nakamoto territory, which doesn't seem right.

djordon (commented):
Had some preliminary comments.

signer/src/stacks/api.rs
djordon: One thing that we could do is add another variable for bootstrapping. Another is use the fact that this is a testnet check, since this isn't a problem on mainnet.

signer/src/stacks/api.rs
setzeus: Addressed in [ba94bc1](https://github.com/stacks-sbtc/sbtc/pull/1803/commits/ba94bc1df3d446f68001106606e4be52ff4c2001)

signer/src/stacks/api.rs
setzeus: Addressed in [ba94bc1](https://github.com/stacks-sbtc/sbtc/pull/1803/commits/ba94bc1df3d446f68001106606e4be52ff4c2001)

signer/src/stacks/api.rs
setzeus: Addressed in [ba94bc1](https://github.com/stacks-sbtc/sbtc/pull/1803/commits/ba94bc1df3d446f68001106606e4be52ff4c2001)

signer/src/stacks/api.rs
setzeus: Addressed in [ba94bc1](https://github.com/stacks-sbtc/sbtc/pull/1803/commits/ba94bc1df3d446f68001106606e4be52ff4c2001)

signer/src/stacks/api.rs
setzeus: Addressed in [ba94bc1](https://github.com/stacks-sbtc/sbtc/pull/1803/commits/ba94bc1df3d446f68001106606e4be52ff4c2001)

signer/src/stacks/api.rs
```diff
@@ -1309,15 +1309,67 @@ where
...
+    nakamoto_start_height: BitcoinBlockHeight,
+    network: NetworkKind,
+) -> bool {
+    match error {
+        // If we get a 404, it's likely because the block ID refers to a pre-Nakamoto block
+        // But only if we're already close to or at the Nakamoto start height
+        // AND we're not in mainnet (since this is what we're addressing)
+        Error::StacksNodeResponse(req_error) => {
+            req_error.status() == Some(reqwest::StatusCode::NOT_FOUND)
+                && current_anchor_height <= nakamoto_start_height + 1
```
djordon: Hmmmm, I don't think that this logic is correct. If it were, then we wouldn't need the check for the network (so I'm hoping that I'm wrong here). Basically, I think that it's possible for the current anchor height to be the `nakamoto_start_height + X`, where `X` can be any nonnegative integer.

If my thinking is correct, then we should remove this check and just rely on the network.

signer/src/stacks/api.rs
```diff
@@ -1309,15 +1309,67 @@ where
...
+        match stacks.get_tenure(header.parent_block_id).await {
+            Ok(parent_tenure_blocks) => {
+                let parent_tenure: TenureBlockHeaders = parent_tenure_blocks.into();
+                headers.push(parent_tenure);
+            }
+            Err(error) => {
+                // Get network configuration to determine if we should treat 404 as pre-Nakamoto
+                let network = Settings::new_from_default_config()
+                    .map(|settings| settings.signer.network)
+                    .unwrap_or(NetworkKind::Regtest); // Default to Regtest if config fails
```
djordon, line 1324: I think that it would be better to pass in this information. I'd change the type signature of `fetch_unknown_ancestors` to take a generic context and a Stacks block ID, similar to this function https://github.com/stacks-sbtc/sbtc/blob/d1f4ef6e4e348543bb711be2090d9135d5ec011d/signer/src/block_observer.rs#L721-L724 Once you have the context you can call `.config()` to return a settings object.

djordon: Actually, I'm beginning to rethink using the network here. Perhaps what we should do is just check to see if we have any stacks blocks in the database associated with the current bitcoin chain tip and bail if we get an error, and continue otherwise. This turns the check into one where we ask if we're in a "bootstrapping" phase.

So a couple of notes
1. This function should probably take a generic context still.
2. It should also take a bitcoin chain tip for calculating the stacks chain tip.
3. If the stacks chain tip doesn't exist, then we are likely bootstrapping and can tolerate a 404 Not Found.

setzeus: Copy on the context part.

For 3., not sure I follow why we wouldn't want a testnet filter here though? If stacks chain tip doesn't exist on mainnet we def wouldn't want to tolerate that no?

signer/src/stacks/api.rs
setzeus: Copy. Updated in [bb70c82](https://github.com/stacks-sbtc/sbtc/pull/1803/commits/bb70c82126fb6fc8f30d27441caf476b845d3e28)

signer/src/stacks/api.rs
djordon: > For 3., not sure I follow why we wouldn't want a testnet filter here though?

Because we're aiming for logic that is always true and adding an "is testnet" filter would be unnecessary with the logic that is outlines above.

PR #1812, opened 2025-08-04
feat: move coordinator checks for smart contract deployment to block observer

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1811

## Changes

* Move the check for smart contract deployment to the block observer. It's more natural here, and doesn't result in a noisy start when starting the application.
* Updated the tests affected by the change.

## Testing Information

I repurposed one of our existing tests to test some of the new functionality. This is the new `get_signer_set_info_checks_for_contract_deployment` test. ~~I'm going to test this on devenv, just waiting for the build at the moment~~ I ran this on devenv and things seemed mostly fine. I saw some "concurrent WSTS activity observed" logs early on and also saw that my first deposit took two bitcoin block to be swept instead of one. The later deposits were promptly swept and those logs didn't show up after a while. I think this is an artifact of the 1 second bitcoin blocks pre-Nakamoto and 5 second polling of bitcoin core.

## Checklist

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -374,8 +374,12 @@ where
         tracing::debug!("loading the signer stacks wallet");
         let wallet = self.get_signer_wallet().await?;
 
-        self.deploy_smart_contracts(chain_tip_hash, &wallet, &aggregate_key)
-            .await?;
+        if !self.context.state().sbtc_contracts_deployed() {
+            self.deploy_smart_contracts(chain_tip_hash, &wallet, &aggregate_key)
+                .await?;
+
+            return Ok(());
+        }
 
         let rotate_key_txid = self.check_and_submit_rotate_key_transaction(
             &bitcoin_chain_tip,
@@ -429,10 +433,6 @@ where
         wallet: &SignerWallet,
         aggregate_key: &PublicKey,
     ) -> Result<Option<StacksTxId>, Error> {
-        if !self.context.state().sbtc_contracts_deployed() {
-            return Ok(None);
-        }
-
         let last_dkg = self
             .context
             .get_storage()
@@ -2219,6 +2219,7 @@ where
     }
 
     /// Deploy an sBTC smart contract to the stacks node.
+    #[tracing::instrument(skip_all, fields(smart_contract = %contract_deploy))]
     async fn deploy_smart_contract(
         &mut self,
         contract_deploy: SmartContract,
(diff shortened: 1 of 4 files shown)


signer/src/transaction_coordinator.rs
```diff
@@ -434,7 +429,7 @@ where
         wallet: &SignerWallet,
         aggregate_key: &PublicKey,
     ) -> Result<Option<StacksTxId>, Error> {
-        if !self.all_smart_contracts_deployed().await? {
+        if !self.context.state().sbtc_contracts_deployed() {
```
cylewitruk: We call `deploy_smart_contracts()` literally the line before this method is called. I wonder if it wouldn't make more sense to stop the coordinator's tenure after a call to `deploy_smart_contracts()` where contracts were deployed -- it doesn't seem likely that the check here will succeed immediately after contract deployment (submission). If we did that then the check here would be unnecessary and functions can work under the assumption that contracts are indeed deployed.

signer/src/transaction_coordinator.rs
```diff
@@ -2822,24 +2799,18 @@ mod tests {
...
-    /// chain tip is up to date.
     #[tokio::test]
     async fn should_skip_processing_bitcoin_blocks_if_not_coordinator() {
         let mut rng = testing::get_rng();
         let ctx = TestContext::builder()
             .with_in_memory_storage()
             .with_mocked_clients()
+            .modify_settings(|settings| {
+                // If we are the coordinator then we will wait for 10 seconds.
+                settings.signer.bitcoin_processing_delay = Duration::from_secs(10);
```
cylewitruk, line 2811: Maybe add a little more context to the comment here from the comment below (i.e. explaining that this is to ensure that the short timeout further down actually times out). Reading top-down I started this comment with "do we really need 10s?!" 😅

cylewitruk (approved):
Looks good

signer/src/transaction_coordinator.rs
djordon: Yeah, good idea. Will update.

signer/src/transaction_coordinator.rs
djordon: 😄 Yeah makes sense, will do.

matteojug (approved):
LGTM (assuming Cyle's nits)

signer/src/transaction_coordinator.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1812/commits/318b2605dd289e1e7a4dd4acb139fbb4be332fab

signer/src/transaction_coordinator.rs
djordon: Moved things around in https://github.com/stacks-sbtc/sbtc/pull/1812/commits/6dbfd82a7b051dcb1c687d64be671062642220ee

matteojug (approved):

PR #1813, opened 2025-08-04
fix: no more flaky `sign_bitcoin_transaction` integration test

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1810

I think the issue with the test was that the coordinator hasn't received everyone's votes for the deposit by the time that they go to process it. This meant that the coordinator don't see any deposits that have enough votes and move on. With a sleep, that gives everyone enough time to send their votes and for the coordinator to receive and process them before beginning their coordinator duties. Earlier versions of the signer wouldn't have this issue, because the coordinator would query the database for information (like the bitcoin chain tip for example), causing a delay and giving the signers more time to process the deposit in the test.

## Changes

* Add a coordinator delay to the `sign_bitcoin_transaction` and friends tests.

## Testing Information

After warming up bitcoin core by running `complete_deposit_validation_happy_path` 5 times in a row, I ran the `sign_bitcoin_transaction` test 10 times in a row and they all passed.

## Checklist

### DIFF
--- a/signer/tests/integration/transaction_coordinator.rs
+++ b/signer/tests/integration/transaction_coordinator.rs
@@ -1975,6 +1975,9 @@ async fn sign_bitcoin_transaction() {
             .with_first_bitcoin_core_client()
             .with_emily_client(emily_client.clone())
             .with_mocked_stacks_client()
+            .modify_settings(|settings| {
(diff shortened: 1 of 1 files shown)


cylewitruk (approved):
Ah, more sleeps, yay 🫠

matteojug (approved):

PR #1814, opened 2025-08-04
We need a minimum for the polling interval

djordon:
## Description

Maybe we should add this check for the config instead:
```rust
if self.bitcoin.chain_tip_polling_interval > self.signer.bitcoin_processing_delay {
    return Err(ConfigError::Message(
        "[bitcoin.chain_tip_polling_interval] must be less than or equal to [signer.bitcoin_processing_delay]".to_string(),
    ));
}
```

## Changes

* We should check the polling interval against zero, not sure about the current maximum though.

## Checklist

### DIFF
--- a/signer/src/config/mod.rs
+++ b/signer/src/config/mod.rs
@@ -167,12 +167,9 @@ impl Validatable for BitcoinConfig {
             }
         }
 
-        if self.chain_tip_polling_interval.as_secs()
-            > MAX_BITCOIN_CHAIN_TIP_POLLING_INTERVAL_SECONDS
-        {
+        if self.chain_tip_polling_interval == std::time::Duration::ZERO {
             return Err(ConfigError::Message(
-                "[bitcoin.chain_tip_polling_interval] Maximum polling interval exceeded"
-                    .to_string(),
+                "[bitcoin.chain_tip_polling_interval] Zero is not allowed".to_string(),
             ));
         }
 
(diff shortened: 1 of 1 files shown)


signer/src/config/mod.rs
```diff
@@ -167,12 +167,9 @@ impl Validatable for BitcoinConfig {
             }
         }
 
-        if self.chain_tip_polling_interval.as_secs()
-            > MAX_BITCOIN_CHAIN_TIP_POLLING_INTERVAL_SECONDS
-        {
+        if self.chain_tip_polling_interval == std::time::Duration::ZERO {
```
MCJOHN974, line 170: Are we sure we want 0 as minimum? If minimum is zero why we need this constraint at all? Like, since minimum is 0 I can put it to 1 nanosecond and it will make no difference compared to 0. Also, since we parse this interval from number of seconds in config file, tecnical minimum is 1 second.
```suggestion
        if self.chain_tip_polling_interval <= std::time::Duration::from_secs(1) {
```

signer/src/config/mod.rs
```diff
@@ -167,12 +167,9 @@ impl Validatable for BitcoinConfig {
...
             return Err(ConfigError::Message(
-                "[bitcoin.chain_tip_polling_interval] Maximum polling interval exceeded"
-                    .to_string(),
+                "[bitcoin.chain_tip_polling_interval] Zero is not allowed".to_string(),
```
MCJOHN974, line 172:
```suggestion
                "[bitcoin.chain_tip_polling_interval] smaller then one second is not allowed".to_string(),
```

Adjusting comment according to nit above

signer/src/config/mod.rs
```diff
@@ -819,12 +816,11 @@ mod tests {
...
         clear_env();
 
-        let interval_secs = (MAX_BITCOIN_CHAIN_TIP_POLLING_INTERVAL_SECONDS + 1).to_string();
-        set_var("SIGNER_BITCOIN__CHAIN_TIP_POLLING_INTERVAL", interval_secs);
+        set_var("SIGNER_BITCOIN__CHAIN_TIP_POLLING_INTERVAL", "0");
 
         assert_matches!(
             Settings::new_from_default_config(),
-            Err(ConfigError::Message(msg)) if msg.contains("polling interval exceeded")
+            Err(ConfigError::Message(msg)) if msg.contains("Zero is not allowed")
```
MCJOHN974, line 823:
```suggestion
            Err(ConfigError::Message(msg)) if msg.contains("smaller then one second is not allowed")
```

signer/src/config/mod.rs
MCJOHN974: I mean like, 0 don't feel natural bound for me, and it feels confusing, like "we we need this at all, does it mean that 0 and 1 nanosecond make difference here?"

djordon: We only have seconds resolution for that duration, https://github.com/stacks-sbtc/sbtc/blob/ecc0402102fcb2fc3b377547bb195a74a72672b0/signer/src/config/mod.rs#L139-L142

Issue #1797, opened 2025-07-30
[Feature]: Remove `dkg_target_rounds` from DKG


matteojug:
# Feature - Remove `dkg_target_rounds` from DKG

## 1. Description

Now that we can trigger DKG when the signer set changes, the current `dkg_target_rounds` based DKG can be a bit messy since we may have different targets for different signers.

What we could do instead is triggering DKG if we don't have a DKG round (verified|unverified) started after the configured `dkg_min_bitcoin_block_height`.

Since that should make the logic lighter, we should also try to unify the coordinator and signer logic in a single place. The nuisance of it is that the signer should be more aggressive in the logging wrt the coordinator.

PR #1819, opened 2025-08-08
DKG Consolidation Update

setzeus:
## Description

Closes: #1797 

## Changes
Updated the logic in 'should_coordinate_dkg' to be based on the 'min_height' as opposed to 'dkg_rounds.' Additionally, updated 'assert_dkg_begin' to call into the updated function in an attempt to consolidate logic & minimize state divergence.

## Testing Information
Currently this is a draft to receive feedback on the logic so the current tests have not been updated & are therefore *broken.* They will be fixed once initial feedback is received.

## Checklist

### DIFF
--- a/signer/tests/integration/transaction_coordinator.rs
+++ b/signer/tests/integration/transaction_coordinator.rs
@@ -1910,10 +1910,6 @@ async fn pseudo_random_dkg() {
     //   aggregate key and new secret shares.
     // =========================================================================
 
-    // Let's run DKG a third time, where this time we expect new secret
-    // shares to be generated.
-    let new_chain_tip: BitcoinBlockHash = faucet.generate_block().into();
-
     // Update the started_at_height of the second DKG shares to allow the third DKG run
     for (_, db, _, _) in signers.iter() {
         let latest_shares = db.get_latest_encrypted_dkg_shares().await.unwrap().unwrap();


signer/src/transaction_coordinator.rs
```diff
@@ -2505,8 +2505,8 @@ pub fn coordinator_public_key(
         .copied()
 }
 
-/// Determine, according to the current state of the signer and configuration,
-/// whether or not a new DKG round should be coordinated.
+/// Unified DKG decision logic that can be used by both coordinator & signer.
+/// Returns true if DKG should be allowed to proceed.
 pub async fn should_coordinate_dkg(
```
djordon: We'll probably want to rename this to, `should_run_dkg` or something.

signer/src/transaction_coordinator.rs
```diff
@@ -2505,8 +2505,8 @@ pub fn coordinator_public_key(
         .copied()
 }
 
-/// Determine, according to the current state of the signer and configuration,
-/// whether or not a new DKG round should be coordinated.
+/// Unified DKG decision logic that can be used by both coordinator & signer.
+/// Returns true if DKG should be allowed to proceed.
```
djordon:
```suggestion
/// Contains the logic used to decide if we should run DKG.
/// Returns true if DKG should be allowed to proceed.
```

signer/src/transaction_coordinator.rs
```diff
@@ -2551,33 +2551,47 @@ pub async fn should_coordinate_dkg(
...
-    ) {
-        (current, target, Some(dkg_min_height))
-            if current < target.get() && bitcoin_chain_tip.block_height >= dkg_min_height =>
-        {
-            tracing::info!(
+    
+    match dkg_min_bitcoin_block_height {
+        Some(dkg_min_height) => {
+            // Check if we've passed the minimum height and no DKG has been started since then
+            if bitcoin_chain_tip.block_height >= dkg_min_height {
```
djordon:
```suggestion
        // Check if we've passed the minimum height and no DKG has been
        // started since then
        Some(dkg_min_height) if bitcoin_chain_tip.block_height >= dkg_min_height => {
```

djordon (commented):
👍🏽 I think the direction looks good. Left some nits.

signer/src/transaction_coordinator.rs
setzeus: Addressed in [b800e5c](https://github.com/stacks-sbtc/sbtc/pull/1819/commits/b800e5cbabdd49bf35d7f88a400104be0280e451)

signer/src/transaction_coordinator.rs
setzeus: Addressed in [b800e5c](https://github.com/stacks-sbtc/sbtc/pull/1819/commits/b800e5cbabdd49bf35d7f88a400104be0280e451)

signer/src/transaction_coordinator.rs
setzeus: Addressed in [b800e5c](https://github.com/stacks-sbtc/sbtc/pull/1819/commits/b800e5cbabdd49bf35d7f88a400104be0280e451)

PR #1822, opened 2025-08-08
feat: move clippy lints to `Cargo.toml` and enable `unused_imports` and `unused_trait_names`

cylewitruk:
## Description

Just a little clippy improvements

## Changes

- Centralizes clippy configuration for the workspace in `Cargo.toml`
- Enables (disallows) `unused_imports` and `unused_trait_names`

## Checklist


Cargo.toml
```diff
@@ -112,3 +112,14 @@ split-debuginfo = "unpacked"
...
+[workspace.lints.rust]
+missing_docs = "deny"
+unused_imports = "deny"
+
+[workspace.lints.clippy]
+unused_trait_names = "warn"
+unwrap_in_result = "deny"
+#unwrap_used = "warn"
+#expect_used = "warn"
+empty_loop = "deny"
```
djordon: nit: new line here please. And at the end of `sbtc/Cargo.toml` as well.

djordon (dismissed):
Looks good, had a nit.

Cargo.toml
cylewitruk: [739f6a1](https://github.com/stacks-sbtc/sbtc/pull/1822/commits/739f6a1993d3e5973bab6526eeaca4943fad0e79) - ran `cargo sort` on them too

signer/Cargo.toml
```diff
@@ -67,23 +74,23 @@ mockall = { workspace = true, optional = true }
 tonic-build.workspace = true
 
 [dev-dependencies]
+# Local crates
+sbtc = { workspace = true, features = ["testing"] }
+signer = { path = ".", features = ["testing"] }
```
matteojug, line 82: nit: should we keep the workaround reference? ie, https://github.com/rust-lang/cargo/issues/2911#issuecomment-749580481

matteojug (dismissed):

signer/Cargo.toml
cylewitruk: It's not a workaround, it's very common. Or, maybe it's a workaround, but it's pretty much the accepted way to enable local features during dev

djordon: I would still keep the reference around. Even if it's the common solution to the problem, it's not obvious. Plus it doesn't hurt for those who do not know.

Cargo.toml
```diff
@@ -68,40 +78,41 @@ sqlx = { version = "0.8.3", default-features = false, features = [ "postgres", "
...
 
+[workspace.lints.rust]
+missing_docs = "deny"
+unused_imports = "deny"
+
+[workspace.lints.clippy]
+unused_trait_names = "warn"
+unwrap_in_result = "deny"
+#unwrap_used = "warn"
+#expect_used = "warn"
```
MCJOHN974, line 115: Do we need this comments?

Cargo.toml
```diff
@@ -68,40 +78,41 @@ sqlx = { version = "0.8.3", default-features = false, features = [ "postgres", "
...
 toml_edit = { version = "0.22.22", default-features = false, features = ["display"] }
-tempfile = { version = "3.15.0", default-features = false }
 tower = { version = "0.5.2", default-features = false, features = ["util"] }
 
+[workspace.lints.rust]
+missing_docs = "deny"
+unused_imports = "deny"
+
+[workspace.lints.clippy]
+unused_trait_names = "warn"
```
MCJOHN974: Why not "deny" here?

MCJOHN974:
And also do we want to put info to this `Cargo.toml` that we deny clippy warnings and remove `-Dwarnigns` flag from `Makefile`?

Cargo.toml
cylewitruk: Yeah, why not.. [069e3fa](https://github.com/stacks-sbtc/sbtc/pull/1822/commits/069e3fa37a2ff555e2f4468d100756b6e90e4102)

Cargo.toml
cylewitruk: The hope is to actually enable them at some point and this just serves as a reminder. But added a comment [7ed72f9](https://github.com/stacks-sbtc/sbtc/pull/1822/commits/7ed72f9acf2abc620df032265a89ec1f224e7944)

signer/Cargo.toml
cylewitruk: I don't think the referenced issue is very helpful for a newcomer who doesn't know about this, rather confusing. But I added a more informative comment in [10aa130](https://github.com/stacks-sbtc/sbtc/pull/1822/commits/10aa13033d166d0759e08b428fc1aebe538c6dfb).

Cargo.toml
cylewitruk: I'm actually on the fence if we shouldn't use `warn` instead for all of our custom lints, with the comment that we use `-Dwarnings`.

Reason being to allow a developer distinguish between actual code errors and lint errors (but then `-Dwarnings` makes the `make lint|build` etc. fail).

MCJOHN974: yeah, make sense, then I don't mind to use warns if we use them everywhere

MCJOHN974 (approved):

Issue #1815, opened 2025-08-05
[Feature]: Upgrade to the latest version of WSTS


djordon:
# Feature - Upgrade to the latest version of WSTS

## 1. Description

The latest version of WSTS includes two protocol breaking changes:
1. A change in the signing protocol.
2. A change in the network wire format of one of the messages sent during DKG.

Both changes are problematic for the sbtc signer codebase, since the signing protocol changes require a consensus majority of signers to update at the same time in order to avoid downtime. We do not control the sBTC signers so we need to engineer a solution that allows signers to update their systems when they so choose.

## 2. Technical Details

The WSTS signing changes modify what is hashed during the construction of signature shares. This is a breaking change, so all signers need to ensure that they use the same procedure when signing something using WSTS. Two ways that we could do this are:

1. Have the signers use one protocol for signing before a canonical bitcoin block with a known height is observed and another after.
2. Use one protocol for all old shares, and another after any new shares.
3. Update WSTS to allow configuring which version of the signing protocol to use, and making that configurable in the signer.

@djordon was in favor of doing (1), but his proposed solution would have required having enumerations for every WSTS object that sBTC uses.  So instead we are implementing (3).  We could still have a block height at which we swap the configuration, or just wait until all signers upgrade, then have them update their configs.  The current solution is configured via `SIGNER_SIGNER__EXPANSION_TYPE`.

Handling the change in network wire format is relative straightforward. We handled this in our earlier effort to update the signers to a later version of WSTS (https://github.com/stacks-sbtc/sbtc/pull/1789).  The current PR brings that code back into sBTC, with the addition of the signer config which controls the protocol change.

### 2.1 Acceptance Criteria

- [ ] The sBTC signers eventually use the latest version of WSTS for signing.
- [ ] The sBTC signers can upgrade their signer binaries at a leisurely pace without risk of causing a conflict.

## 3. Related Issues and Pull Requests (optional)

Our earlier PR upgrading WSTS https://github.com/stacks-sbtc/sbtc/pull/1789, the reverting PR https://github.com/stacks-sbtc/sbtc/pull/1808, and the proposed solution is in WSTS https://github.com/stacks-sbtc/wsts/pull/201 and sBTC https://github.com/stacks-sbtc/sbtc/pull/1824

PR #1824, opened 2025-08-19
Upgrade wsts with expansion type

xoloki:
## Description
This PR upgrades WSTS as per the previous PR, but with the new `ExpansionType` code.

Closes: #1815 

## Changes
All of the recent `WSTS` security updates, including configuration for the `ExpansionType`.

## Testing Information
This PR adds a unit test: `default_config_toml_loads_xmd_min_bitcoin_block_height`, which tests that the config responds correctly to env variables.  It also adds a parameter to the existing `sign_bitcoin_transaction_multiple_locking_keys` integration test, which if set will have the test use XMD expansion after the second DKG is run.   That test is the run both with and without the flag being set.

## Checklist

### DIFF
--- a/signer/src/config/mod.rs
+++ b/signer/src/config/mod.rs
@@ -567,14 +567,13 @@ impl Settings {
             DEFAULT_MAX_DEPOSITS_PER_BITCOIN_TX,
         )?;
         cfg_builder = cfg_builder.set_default("signer.dkg_target_rounds", 1)?;
-        cfg_builder = cfg_builder.set_default("signer.dkg_target_rounds", 1)?;
         cfg_builder = cfg_builder.set_default("emily.pagination_timeout", 10)?;
         cfg_builder = cfg_builder.set_default("signer.dkg_verification_window", 10)?;
         cfg_builder = cfg_builder.set_default("signer.stacks_fees_max_ustx", 1_500_000)?;
         cfg_builder = cfg_builder.set_default("bitcoin.chain_tip_polling_interval", 5)?;
         cfg_builder = cfg_builder.set_default(
             "signer.xmd_min_bitcoin_block_height",
-            *BitcoinBlockHeight::from(DEFAULT_SIGNER_XMD_MIN_BITCOIN_BLOCK_HEIGHT),
+            DEFAULT_SIGNER_XMD_MIN_BITCOIN_BLOCK_HEIGHT,
         )?;
 
         if let Some(path) = config_path {


signer/src/testing/wsts.rs
```diff
@@ -122,8 +133,9 @@ impl Coordinator {
             dkg_end_timeout: None,
             nonce_timeout: None,
             sign_timeout: None,
-            signer_key_ids,
-            signer_public_keys,
+            public_keys,
+            verify_packet_sigs: false,
+            expansion_type: ExpansionType::Default,
```
matteojug: Should we take `expansion_type` as input here, so that the tests using this can check both expansions?

signer/src/config/mod.rs
```diff
@@ -388,6 +388,9 @@ pub struct SignerConfig {
     /// arrives. The default here is controlled by the
     /// [`MAX_DEPOSITS_PER_BITCOIN_TX`] constant
     pub max_deposits_per_bitcoin_tx: NonZeroU16,
+    /// Configures the signer to use WSTS XMD expansion after the specified
+    /// Bitcoin block height
+    pub xmd_min_bitcoin_block_height: Option<BitcoinBlockHeight>,
```
matteojug: maybe nit: we could remove the `Option` and use max bitcoin block height as default for the same effect as `None`, to avoid having to unwrap it

signer/src/dkg/testing.rs
```diff
@@ -45,8 +46,10 @@ impl TestSetup {
             panic!("must have at least 1 parties");
         }
 
-        let (coordinators, signers) =
-            wsts_test::run_dkg::<frost::Coordinator<v2::Aggregator>, v2::Party>(num_parties, 5);
+        let (coordinators, signers) = wsts_test::run_dkg::<
+            frost::Coordinator<v2::Aggregator>,
+            v2::Party,
+        >(num_parties, 5, ExpansionType::Default);
```
matteojug: As with the other occurrence, can we propagate the expansion from the caller and test both?

signer/src/config/mod.rs
```diff
@@ -388,6 +388,9 @@ pub struct SignerConfig {
     /// arrives. The default here is controlled by the
     /// [`MAX_DEPOSITS_PER_BITCOIN_TX`] constant
     pub max_deposits_per_bitcoin_tx: NonZeroU16,
+    /// Configures the signer to use WSTS XMD expansion after the specified
+    /// Bitcoin block height
```
matteojug: IIUC this is not entirely true: when we load a state machine for signing, we compare `xmd_min_bitcoin_block_height` with the encrypted shares creation bitcoin height, so we will start to use XMD only after the first key rotation past `xmd_min_bitcoin_block_height`.

Imho the current approach is fine, can we just update the docstring to better specify this behaviour? And maybe we should add a test for this, to ensure we can still sign fine after we reach `xmd_min_bitcoin_block_height`, and that we can then DKG fine.

signer/src/wsts_state_machine.rs
```diff
@@ -245,25 +276,44 @@ impl WstsCoordinator for FireCoordinator {
...
         let signer_key_ids = (0..num_signers)
             .map(|signer_id| (signer_id, std::iter::once(signer_id + 1).collect()))
             .collect();
+        let public_keys = wsts::state_machine::PublicKeys {
+            signers,
+            key_ids,
+            signer_key_ids,
+        };
+        let expansion_type = if let Some(xmd_block_height) = xmd_block_height {
+            if xmd_block_height <= block_height {
```
matteojug: nit: we can let chain now 🎉

xoloki: Done

signer/src/config/mod.rs
xoloki: Using `Option` seems cleaner for an optional config item.  We do that for all the other optional config items in this file, including `dkg_min_bitcoin_block_height`.

matteojug: I thought this isn't going to be optional (we want to switch), and we will set a default to ensure a smooth transition

signer/src/testing/wsts.rs
xoloki: Taking `expansion_type` as a parameter to `Coordinator::new` means chaining it up to `SignerSet::new`.   Did that, but hardwired it as `Default` in most callsites.  Added it as a parameter to a DKG test so we can check both there.

signer/src/dkg/testing.rs
xoloki: Done

signer/src/config/mod.rs
xoloki: Fixed comment

signer/src/config/mod.rs
xoloki: I thought @djordon wanted this to be optional, but maybe I'm confused about that?

signer/src/config/mod.rs
xoloki: For the test, should I do something like `run_subsequent_dkg`?  Or maybe add signing to that test?

xoloki: From standup: `sign_bitcoin_transaction_multiple_locking_keys`

djordon: I think we should do something similar to the `sign_bitcoin_transaction_multiple_locking_keys` test.

xoloki:
As discussed in standup, the test harness does not allow us to easily pierce the veil of the wsts state machines, so I added `eprintln` statements to verify that we were in fact using `ExpansionType::Xmd` once the block height had been reached and DKG had run.  Here's the test run:

https://github.com/stacks-sbtc/sbtc/actions/runs/17270987010/job/49015405261?pr=1824#logs

Reverting that code now.

signer/src/config/mod.rs
xoloki: As per our convo during standup, I made it required with default the current bitcoin block height so it will change next time we run DKG.

signer/src/config/mod.rs
xoloki: Done.  As per convo, I added a parameter to that test to set the XMD height to the second DKG height, and run the test with and without it set.

signer/src/wsts_state_machine.rs
xoloki: Not chained anymore since its no longer an option.

PR #1834, opened 2025-10-10
chore: update stacks dependencies for 3.2.0.0.2

cylewitruk:
## Description

Updates our stacks-core dependencies to 3.2.0.0.2 to support Stacks epoch 3.3.

## Changes

- A number of types from stacks-core are no longer `Copy`, so our types using them can no longer be copy. This required adding a number of `.clone()` calls.
- Updates to stacks-core crypto dependencies conflicted with `libp2p-identity`. I was able to include this dependency separately (`0.12.0`) instead of bumping the entire libp2p stack.

## Testing Information

- All tests/integration tests pass. I haven't run this in devenv yet.

## Checklist

### DIFF
--- a/supply-chain/config.toml
+++ b/supply-chain/config.toml
@@ -109,10 +109,6 @@ criteria = "safe-to-deploy"
 version = "2.0.0"
 criteria = "safe-to-deploy"
 
-[[exemptions.atomic-waker]]
-version = "1.1.2"
-criteria = "safe-to-deploy"
-
 [[exemptions.attohttpc]]
 version = "0.24.1"
 criteria = "safe-to-deploy"
@@ -1606,15 +1602,7 @@ version = "0.1.15"
 criteria = "safe-to-deploy"
 
 [[exemptions.stacks-common]]
-version = "0.0.1"
-criteria = "safe-to-deploy"
-
-[[exemptions.stacks-common]]
-version = "0.0.1"
-criteria = "safe-to-deploy"
-
-[[exemptions.stacks-common]]
-version = "0.0.1@git:6c769636d5d598d1872960ca2744ad1b64fa6385"
+version = "0.0.1@git:bd9ee6310516b31ef4ecce07e42e73ed0f774ada"
 criteria = "safe-to-deploy"
 
 [[exemptions.stringprep]]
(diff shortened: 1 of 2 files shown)


signer/src/testing/context.rs
```diff
@@ -482,11 +482,11 @@ impl StacksInteract for WrappedMockStacksInteract {
...
     }
 
-    async fn get_block(&self, block_id: StacksBlockId) -> Result<NakamotoBlock, Error> {
+    async fn get_block(&self, block_id: &StacksBlockId) -> Result<NakamotoBlock, Error> {
         self.inner.lock().await.get_block(block_id).await
     }
 
-    async fn get_tenure(&self, block_id: StacksBlockId) -> Result<TenureBlocks, Error> {
+    async fn get_tenure(&self, block_id: &StacksBlockId) -> Result<TenureBlocks, Error> {
         self.inner.lock().await.get_tenure(block_id).await
```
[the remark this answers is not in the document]
cylewitruk: Not sure what this is referring to, because the impl takes a reference.

Cargo.toml
```diff
@@ -56,8 +56,9 @@ http = { version = "1.2.0", default-features = false }
 include_dir = { version = "0.7.4", default-features = false }
 libp2p = { version = "0.55.0", default-features = false, features = [
     "macros", "kad", "noise", "ping", "tcp", "tokio", "yamux", "mdns", "quic", 
-    "gossipsub", "identify", "tls", "dns", "autonat", "secp256k1"
+    "gossipsub", "identify", "tls", "dns", "autonat"
 ] }
+libp2p-identity = { version = "0.2.12", default-features = false, features = ["secp256k1"] }
```
matteojug, line 61: Can we double check in devenv that this work and we are not missing any required `feature`? I'm asking since in the PR it doesn't mention a devenv run

signer/src/storage/model.rs
```diff
@@ -954,9 +961,21 @@ impl std::fmt::Display for StacksBlockHash {
 }
 
 /// Stacks transaction ID
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct StacksTxId(blockstack_lib::burnchains::Txid);
 
+impl StacksTxId {
+    /// Return the inner bytes for the block hash.
```
matteojug, line 968: nit: not a block hash, not a merge blocker either :)

djordon: Yeah, I'm fixing this in a subsequent PR

Cargo.toml
djordon: I ran devenv and deposits and withdrawals worked just fine. No issues. I believe @cylewitruk did the same and had the same experience.

Issue #1816, opened 2025-08-05
[Feature]: Looser handling of /v2/pox responses


djordon:
# Feature - Looser handling of /v2/pox responses

## 1. Description

We currently use the `RPCPoxInfoData` response type from `stackslib` for API calls to `/v2/pox` which we pull from GitHub at a specific commit hash. This response type includes a field with the `StacksEpochId` enum type from the `clarity` crate.

If the sBTC signer's companion Stacks node is upgraded to a version which includes a new variant of the `StacksEpochId` enum, this will result in a hard deserialization error, rendering the sBTC signer effectively inoperable. This is an event which is expected to occur from time to time (i.e. hard forks).

We need to make this (and any other response types we use that can fail deserialization in this way) resilient to such events.

### 1.1 Context & Purpose

This is remedial task to the incident on 2025-07-25 where this was encountered for the first time on mainnet.

### 2.1 Acceptance Criteria

- [ ] Criterion 1
- [ ] Criterion 2
- [ ] Criterion 3

PR #1836, opened 2025-10-14
feat: relaxed handling of v2 pox responses (specifically epochs)

cylewitruk:
## Description

Closes: #1816 

## Changes

- Adds a new `StacksInteract::get_epoch_status()` method which uses a minimal struct for deserialization of `/v2/pox` responses, treating `epoch_id` as a string instead of `StacksEpochId`. This method returns a new `StacksEpochStatus` enum variant which covers the two cases that we're interested in: "are we in 3.0+, or not?", and associated metadata which is used for logging/logic elsewhere (nakamoto start height and current height).
- Replaced the usage of `get_pox_info()` with the new method everywhere (including tests).
- Added additional tests specifically testing the epoch 3.0 boundary, including new/garbage epochs which we're not interested in (at the moment).

## Testing Information

All tests pass and devenv seems to work fine.

## Checklist

### DIFF
--- a/signer/src/stacks/api.rs
+++ b/signer/src/stacks/api.rs
@@ -288,7 +288,7 @@ pub trait StacksInteract: Send + Sync {
     /// Attempt to get information from a Stacks node about whether or not it is
     /// in a pre- or post-Nakamoto epoch (3.0).
     ///
-    /// Returns a [`StacksEpochInfo`] variant if successful. If the Stacks node
+    /// Returns a [`StacksEpochStatus`] variant if successful. If the Stacks node
     /// does not report an entry for Epoch 3.0, then an
     /// [`Error::MissingNakamotoStartHeight`] error is returned.
     fn get_epoch_status(&self) -> impl Future<Output = Result<StacksEpochStatus, Error>> + Send;


signer/src/stacks/api.rs
```diff
@@ -1649,6 +1723,10 @@ impl StacksInteract for StacksClient {
         self.get_pox_info().await
     }
 
+    async fn get_epoch_info(&self) -> Result<StacksEpochInfo, Error> {
+        self.get_epoch_info().await
+    }
```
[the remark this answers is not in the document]
cylewitruk: Not true due to the way methods are resolved in trait impls.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -168,9 +169,6 @@ use crate::utxo_construction::make_deposit_request;
 
 type IntegrationTestContext<Stacks> = TestContext<PgStore, BitcoinCoreClient, Stacks, EmilyClient>;
 
-pub const GET_POX_INFO_JSON: &str =
-    include_str!("../../tests/fixtures/stacksapi-get-pox-info-test-data.json");
-
 async fn run_dkg<Rng, C>(
```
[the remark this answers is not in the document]
cylewitruk: I did.

signer/src/stacks/api.rs
```diff
@@ -285,8 +285,13 @@ pub trait StacksInteract: Send + Sync {
...
 
-    /// Get information about the current PoX state.
-    fn get_pox_info(&self) -> impl Future<Output = Result<RPCPoxInfoData, Error>> + Send;
+    /// Attempt to get information from a Stacks node about whether or not it is
+    /// in a pre- or post-Nakamoto epoch (3.0).
+    ///
+    /// Returns a [`StacksEpochInfo`] variant if successful. If the Stacks node
+    /// does not report an entry for Epoch 3.0, then an
+    /// [`Error::MissingNakamotoStartHeight`] error is returned.
+    fn get_epoch_status(&self) -> impl Future<Output = Result<StacksEpochStatus, Error>> + Send;
```
djordon, line 294: I think that if we want to add a method that is more tailored to our needs, we should just add a `get_nakamoto_start_height` method and have it return a `BitcoinBlockHeight`. We don't really care about the `StacksEpochStatus::PreNakamoto` variant, even in devenv.

signer/src/stacks/api.rs
```diff
@@ -646,6 +631,68 @@ impl TryFrom<AccountEntryResponse> for AccountInfo {
...
+struct PoxEpoch {
+    epoch_id: String,
+    start_height: BitcoinBlockHeight,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+struct PoxResponse {
+    current_burnchain_block_height: BitcoinBlockHeight,
+    epochs: Vec<PoxEpoch>,
+}
```
djordon, line 659: We should always add comments, even for private fields or structs.

signer/src/stacks/api.rs
cylewitruk: That was my original thought, but isn't compatible with the existing `transaction_coordinator::is_epoch3()` method as it does need to know if we are in epoch >= 3.0 or not. We could instead expose an e.g. `StacksInteract::is_epoch3_or_greater` method, but then we lose logging resolution in the coordinator, and the get ancestors fn + block observer wants the start height. Returning a tuple was an option, but I prefer structured input/output types, so using an enum cleanly covered all of the current use cases while providing the metadata they need/want (without refactoring the _logic_ of callers).

djordon: Wait, where would we change the logic of the callers? `transaction_coordinator::is_epoch3()` needs the current bitcoin block height and we'd pass in a `BitcoinBlockRef`, leaving everything else unchanged. And `fetch_unknown_ancestors` would be almost completely unchanged as well.

cylewitruk: The bitcoin block height in the coordinator is from the *Bitcoin* node; to properly/reliably determine the status of the *Stacks* node we need to compare the burn height from the Stacks node itself (since if the Stacks node is catching up, even if it's using the same Bitcoin backing node, its burn height will be at the last processed tenure boundary).

signer/src/stacks/api.rs
cylewitruk: [607c9c1](https://github.com/stacks-sbtc/sbtc/pull/1836/commits/607c9c1089e415708032f46915ee22a3abcfe3db)

signer/src/stacks/api.rs
djordon: Ahhh I see, makes sense.

signer/src/stacks/api.rs
```diff
@@ -646,6 +631,82 @@ impl TryFrom<AccountEntryResponse> for AccountInfo {
...
+
+/// Information regarding whether or not we are in pre- or post-Nakamoto
+/// era.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum StacksEpochStatus {
+    /// We are in the pre-Nakamoto era.
+    PreNakamoto {
+        /// The current bitcoin block height.
+        current_bitcoin_height: BitcoinBlockHeight,
```
djordon: We might want to rename this to something like `stacks_node_bitcoin_height` (or whatever) to reflect that it is the current processed height of the stacks node. Right now it reads as though it is the current block height of the canonical bitcoin blockchain. We should also update the comment to be similar to the `StacksEpochStatus.current_burnchain_block_height` field.

signer/src/transaction_coordinator.rs
```diff
@@ -267,19 +267,24 @@ where
...
-        let is_epoch3 = pox_info.current_burnchain_block_height > *nakamoto_start_height;
-        if is_epoch3 {
-            self.is_epoch3 = is_epoch3;
-            tracing::debug!("we are in epoch 3 or later; time to do work");
+        match epoch_status {
+            StacksEpochStatus::PreNakamoto {
+                current_bitcoin_height,
+                nakamoto_start_height,
+            } => {
+                tracing::debug!(%current_bitcoin_height, %nakamoto_start_height, "the stacks node has not reached epoch 3.0; skipping this round");
```
djordon: nit (sorry):
```suggestion
                tracing::debug!(
                    %current_bitcoin_height,
                    %nakamoto_start_height,
                    "the stacks node has not reached epoch 3.0; skipping this round"
                );
```

signer/src/stacks/api.rs
cylewitruk: [004ff91](https://github.com/stacks-sbtc/sbtc/pull/1836/commits/004ff9160911c70cda109b27af9129a8e8ffd9ab)

Went with `reported_bitcoin_height` (just because `stacks_node_bitcoin_height` felt pretty long-winded), and updated the comment to match.

signer/src/transaction_coordinator.rs
cylewitruk: [64a20ea](https://github.com/stacks-sbtc/sbtc/pull/1836/commits/64a20ea3e820ed645afb41e180a7cdce1f5ae7a4)

signer/src/stacks/api.rs
```diff
@@ -285,8 +285,13 @@ pub trait StacksInteract: Send + Sync {
     where
         T: AsTxPayload + Send + Sync;
 
-    /// Get information about the current PoX state.
-    fn get_pox_info(&self) -> impl Future<Output = Result<RPCPoxInfoData, Error>> + Send;
+    /// Attempt to get information from a Stacks node about whether or not it is
+    /// in a pre- or post-Nakamoto epoch (3.0).
+    ///
+    /// Returns a [`StacksEpochInfo`] variant if successful. If the Stacks node
```
matteojug: nit: `StacksEpochInfo` was renamed in `StacksEpochStatus`

cylewitruk: Gah, wish Copilot would highlight these things earlier 😅 [99ddb80](https://github.com/stacks-sbtc/sbtc/pull/1836/commits/99ddb80cc97909759be8554e3e81187db2da528a)

djordon (approved):

PR #1838, opened 2025-10-16
chore: fix openness local docker postgresql for devenv and ci

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1831

## Changes

* Change the port configurations for all local test docker services in so that they are only reachable from our local host machines.
* Change the integration tests to point to `127.0.0.1` instead of localhost. For some reason, on macOS, localhost doesn't seem to work when connecting to bitcoin-core but works fine with postgres.

I tried using the images published at https://hub.docker.com/r/bitcoin/bitcoin for local testing, but that image seemed slower than the one we use. Tests took an additional 30 seconds, going from 6m 30s to 7m, so I switched back. 

## Testing Information

I ran devenv locally and everything seemed fine. I also ran integration tests without any issues.

## Checklist

### DIFF
--- a/docker/docker-compose.ci.yml
+++ b/docker/docker-compose.ci.yml
@@ -13,7 +13,6 @@ services:
     stop_grace_period: 10s
     ports:
       - "127.0.0.1:18443:18443"
-      - "127.0.0.1:28332:28332"
     healthcheck:
       test: ["CMD-SHELL", "bitcoin-cli getblockcount"]
       interval: 3s
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -14,7 +14,6 @@ services:
     stop_grace_period: 10s
     ports:
       - "127.0.0.1:18443:18443"
-      - "127.0.0.1:28332:28332"
     healthcheck:
       test: [ "CMD-SHELL", "bitcoin-cli getblockcount" ]
       interval: 3s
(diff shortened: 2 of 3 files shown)


docker/docker-compose.ci.yml
```diff
@@ -12,12 +12,8 @@ services:
...
     stop_grace_period: 10s
     ports:
-      - 8333:8333
-      - 8332:8332
-      - 18443:18443
-      - 18444:18444
-      - 28332:28332
-      - 28333:28333
+      - "127.0.0.1:18443:18443"
+      - "127.0.0.1:28332:28332"
```
matteojug: nit: do we still need `28332`? Since we removed the zmq dependency; also in test compose

matteojug (dismissed):

cylewitruk:
> I tried using the images published at https://hub.docker.com/r/bitcoin/bitcoin for local testing, but that image seemed slower than the one we use. Tests took an additional 30 seconds, going from 6m 30s to 7m, so I switched back.

That's weird -- I used `bitcoin/bitcoin` in the current [testcontainers PR ](https://github.com/stacks-sbtc/sbtc/pull/1572/files#diff-0fb3012a2575019d6b9b68b5090de4eb32e5af283d525a357571c222da21b649R62)and it worked like a charm 🤔

cylewitruk (dismissed):
LGTM, but yeah we can remove the `28332`s

docker/docker-compose.ci.yml
djordon: Oh no we don't. I'll remove it

matteojug (approved):

PR #1840, opened 2025-10-20
Update Emily Workflow Triggers

BowTiedDevOps:
I've updated the emily deployment workflow to run on manual trigger on the main branch, as well as on pull requests that contain one of the allowed environment labels:
- `environment/sbtc-devnet`
- `environment/sbtc-testnet`
- `environment/sbtc-mainnet`
- `environment/sbtc-staging`
- `environment/sbtc-private-mainnet`

 The `stacks-sbtc/actions` versions used were also updated to the latest version.

### DIFF
--- a/.github/workflows/deploy-emily.yaml
+++ b/.github/workflows/deploy-emily.yaml
@@ -47,6 +47,7 @@ jobs:
         id: set_env
         run: |
           ALLOWED_ENVS=("environment/sbtc-devnet" "environment/sbtc-testnet" "environment/sbtc-mainnet" "environment/sbtc-staging" "environment/sbtc-private-mainnet")
+          SHOULD_RUN=true
 
           if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
             if [ "${GITHUB_REF_NAME}" != "main" ]; then
@@ -54,7 +55,6 @@ jobs:
               exit 1
             fi
 
-            SHOULD_RUN=true
             ENVIRONMENT="${{ github.event.inputs.environment }}"
 
             VALID=false


fdefelici (approved):
LGTM!

BowTiedDevOps:
@fdefelici The `on pull` trigger will not work with PRs coming from a fork, but for that case we've enabled `workflow_dispatch` on the main branch so the changes can be deployed after the PR is merged

djordon (commented):
I think this works for PRs where we want to deploy Emily and have it go through the correct review and deployment flow. However, I'm sure what happens for regular PRs that have no intent of deploying anything. I might be missing something, but doesn't this workflow run for all commits on all PRs against main and fail early if there is no `environment/sbtc-*` label selected? If so, then I think this would lead to a red looking CI for most PRs. We require all green before merging into main, so we may need to exit gracefully during the "Resolve Environment" step.

.github/workflows/deploy-emily.yaml
```diff
@@ -16,12 +29,82 @@ env:
...
+
+            if [ "${#MATCHED_ENVS[@]}" -eq 0 ]; then
+              echo "### No valid environment label found on PR. Please add one of: ${ALLOWED_ENVS[*]}" >> $GITHUB_STEP_SUMMARY
+              SHOULD_RUN=false
+            elif [ "${#MATCHED_ENVS[@]}" -gt 1 ]; then
+              echo "### Multiple environment labels found: ${MATCHED_ENVS[*]}. Please use only one." >> $GITHUB_STEP_SUMMARY
+              exit 1
+            fi
+
+            ENVIRONMENT="${MATCHED_ENVS[0]}"
```
matteojug, line 93: Isn't this branch (the pull request one) missing setting `SHOULD_RUN` to true? We should set it before the `if [ "${#MATCHED_ENVS[@]}" -eq 0 ];`, or in a new `else` branch of that if.

BowTiedDevOps: It is set on L57, at the beginning of the script

matteojug: Isn't that in the `if [ "${{ github.event_name }}" = "workflow_dispatch" ];` branch?

BowTiedDevOps: Yeah, you're right - I've moved it above the `if/else` branch, not sure how it worked on the `infrastructure` repo like this. Will move it there as well, thank you for catching this!

matteojug (approved):

.github/workflows/deploy-emily.yaml
```diff
@@ -16,12 +29,82 @@ env:
...
+            for label in $LABELS; do
+              for env in "${ALLOWED_ENVS[@]}"; do
+                if [ "$label" = "$env" ]; then
+                  MATCHED_ENVS+=("$label")
+                fi
+              done
+            done
+
+            if [ "${#MATCHED_ENVS[@]}" -eq 0 ]; then
+              echo "### No valid environment label found on PR. Please add one of: ${ALLOWED_ENVS[*]}" >> $GITHUB_STEP_SUMMARY
```
djordon, line 86: This statement assumes that we should have a label for the PR, which usually isn't the case. Maybe instead it should say something like, "No environment label found for the PR" and just call it a day?

djordon (approved):
Looks good, left a minor comment about one of the logs.

PR #1841, opened 2025-10-31
chore: update stacks core to release 3.3

djordon:
## Description

We will probably need another PR when they cut the final release but this is okay for now.

## Changes

* Bump the version of stacks core used in devenv
* Update stacks-core's configs so that we actually enter Epoch 3.3 in devenv

## Testing Information

Tested this on devenv without any issues. I ran commit 5883e75a40f4e7fca69ed4a48e1c922cd31bab68.

## Checklist

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,4 +1,4 @@
-# Variables.
+# Variables
 # ------------------------------------------------------------------------------
 x-common-vars:
   - &BTC_ADDR miEJtNKa3ASpA19v5ZhvbKTEieYjLpzCYT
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -31,7 +32,7 @@ x-common-vars:
 # Templates.
 # ------------------------------------------------------------------------------
 x-stacks-blockchain: &stacks-blockchain
-  image: blockstack/stacks-core:3.2.0.0.2@sha256:67c7574782c3b7f95a592d46d881d261524a4bc1c5cf4e66d54d39f330e4bf69
+  image: blockstack/stacks-core:release-3.3.0.0.0-rc2@sha256:1b895681b073bc791f89182e9f985cc46d30552049ba8e9f7f16a19aec9c6a86
```
matteojug: Should we also update `docker/mainnet/nodes/docker-compose.chains.yml`? Asking since this is a not the real release, but also we usually keep those two aligned

matteojug (dismissed):
LGTM, do we want to merge this or wait for the actual release?

docker/docker-compose.yml
cylewitruk: Agree in keeping them in sync (or otherwise just making sure we have one version everywhere)

djordon: Good call. I updated to the latest stacks core release and updated that `.chains` docker compose file in https://github.com/stacks-sbtc/sbtc/pull/1841/commits/fe43aba90830b9ea2ae77a3594fe5ad3049a7365.

matteojug (dismissed):

MCJOHN974 (approved):

MCJOHN974 (approved):

PR #1843, opened 2025-10-31
Add Emily Sidecar And Cron Pipelines

BowTiedDevOps:
I've added emily sidecar and cron pipelines in order to deploy to ecr, using the same method of running the CI as for the emily deployment. The core logic here is built by @fabergat in https://github.com/stacks-sbtc/sbtc/pull/1830, on top of which I've brought the modifications to the way in which the workflow is run made in https://github.com/stacks-sbtc/sbtc/pull/1840.

### DIFF
--- a/.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
+++ b/.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
@@ -1,4 +1,4 @@
-FROM rust:1.90.0-slim-bookworm AS build
+FROM rust:1.91.0-slim-bookworm AS build
 
 ARG CARGO_BUILD_ARGS="--release --locked"
 
--- a/.github/actions/dockerfiles/Dockerfile.signer.debian
+++ b/.github/actions/dockerfiles/Dockerfile.signer.debian
@@ -1,4 +1,4 @@
-FROM rust:1.90.0-slim-bookworm AS build
+FROM rust:1.91.0-slim-bookworm AS build
 
 ARG GIT_COMMIT
 RUN test -n "$GIT_COMMIT" || (echo "GIT_COMMIT not set" && false)
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Below is the output on a machine that is able to build and run all the sources a
 
 ```text
 --- sBTC tool versions ---
-cargo 1.90.0 (840b83a10 2025-07-30)
+cargo 1.91.0 (ea2d97820 2025-10-10)
 cargo-lambda 1.6.2 (2025-01-17Z)
 pnpm 9.1.0
 GNU Make 3.81
--- a/docker/mainnet/sbtc-signer/signer-config.toml.in
+++ b/docker/mainnet/sbtc-signer/signer-config.toml.in
@@ -148,14 +148,6 @@ prometheus_exporter_endpoint = "0.0.0.0:9184"
 # Environment: SIGNER_SIGNER__DKG_MIN_BITCOIN_BLOCK_HEIGHT
 dkg_min_bitcoin_block_height = 880088
 
-# When defined, the signer will attempt/allow multiple rounds of DKG until the
-# specified number of rounds have been completed. Please only use this parameter
-# when instructed to by the sBTC team.
-#
-# Required: false
-# Environment: SIGNER_SIGNER__DKG_TARGET_ROUNDS
-dkg_target_rounds = 2
-
 # !! ==============================================================================
 # !! Stacks Event Observer Configuration
 # !!
--- a/docker/sbtc/Dockerfile
+++ b/docker/sbtc/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.90.0-slim-bookworm AS builder
+FROM rust:1.91.0-slim-bookworm AS builder
 
 # Install dependencies.
 RUN apt-get update
--- a/docker/sbtc/signer/signer-config.toml
+++ b/docker/sbtc/signer/signer-config.toml
@@ -207,14 +207,6 @@ bootstrap_signatures_required = 2
 # Environment: SIGNER_SIGNER__DKG_MIN_BITCOIN_BLOCK_HEIGHT
 # dkg_min_bitcoin_block_height = 1234
 
-# When defined, the signer will attempt/allow multiple rounds of DKG until the
-# specified number of rounds have been completed. Please only use this parameter
-# when instructed to by the sBTC team.
-#
-# Required: false
-# Environment: SIGNER_SIGNER__DKG_TARGET_ROUNDS
-# dkg_target_rounds = 1
-
 # The number of bitcoin blocks after a DKG start where we attempt to verify the
 # shares. After this many blocks, we mark the shares as failed. Please only use
 # this parameter when instructed to by the sBTC team.
--- a/emily/handler/tests/integration/limits.rs
+++ b/emily/handler/tests/integration/limits.rs
@@ -40,7 +40,7 @@ async fn adding_and_then_updating_single_accout_limit_works() {
 
     // Arrange.
     // --------
-    let limits_to_set = vec![
+    let limits_to_set = [
         (
             "test_account",
             AccountLimits {
@@ -100,7 +100,7 @@ async fn adding_and_then_updating_single_accout_limit_works() {
 
     // Set the expected account caps at the end to be the most recently
     // applied limits.
-    let expected_account_caps: HashMap<String, AccountLimits> = vec![
+    let expected_account_caps: HashMap<String, AccountLimits> = [
         (
             "test_account_2",
             AccountLimits {
@@ -181,7 +181,7 @@ async fn test_updating_account_limits_via_global_limit_works() {
 
     // Arrange.
     // --------
-    let account_limits_to_set_individually = vec![
+    let account_limits_to_set_individually = [
         (
             "test_account_1",
             AccountLimits {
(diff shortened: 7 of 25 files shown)


.github/workflows/emily-cron-ecr-build.yaml
```diff
@@ -0,0 +1,153 @@
...
+    inputs:
+      environment:
+        description: "Environment name"
+        required: true
+        default: environment/sbtc-devnet
+        type: choice
+        options:
+          - environment/sbtc-devnet
+          - environment/sbtc-testnet
+          - environment/sbtc-mainnet
```
fabrizio-stacks: so technically we have ECR only on private-mainnet, I would enable just private-mainnet and maybe devnet for the future

BowTiedDevOps: Done, updated both ECR actions

.github/workflows/emily-cron-ecr-build.yaml
```diff
@@ -0,0 +1,150 @@
...
+      environment: ${{ steps.set_env.outputs.environment }}
+      should_run: ${{ steps.set_env.outputs.should_run }}
+    env:
+      GH_TOKEN: ${{ github.token }}
+    steps:
+      - name: Resolve Environment
+        id: set_env
+        run: |
+          ALLOWED_ENVS=("environment/sbtc-devnet" "environment/sbtc-private-mainnet")
+          SHOULD_RUN=true
```
matteojug, line 47: nit: we could move this to a reusable workflow or bash script, to avoid the duplication

.github/workflows/emily-cron-ecr-build.yaml
```diff
@@ -0,0 +1,150 @@
...
+  image:
+    name: Build Image
+    runs-on: ubuntu-latest
+    needs: resolve-env
+    if: needs.resolve-env.outputs.should_run == 'true'
+    environment: ${{ needs.resolve-env.outputs.environment }}
+    steps:
+      - name: Configure AWS Credentials via OIDC
+        id: configure_aws
+        uses: stacks-sbtc/actions/aws/configure-aws-credentials@942832cc7f86ead03c05b7035cd4b9573dd6c617
```
matteojug: nit: the pinned action is not the latest commit (`a1754bb`), is it expected?

.github/workflows/emily-cron-ecr-build.yaml
```diff
@@ -0,0 +1,150 @@
...
+      - name: Docker Metadata
+        id: docker_metadata
+        uses: stacks-sbtc/actions/docker/metadata-action@942832cc7f86ead03c05b7035cd4b9573dd6c617
+        with:
+          images: |
+            ${{ steps.ecr_login.outputs.registry }}/${{ vars['AWS_ECR'] }}
+          tags: |
+            type=raw,value=sbtc-emily-cron-${{ github.ref_name }}
+            type=raw,value=sbtc-emily-cron-${{ github.sha }}
+            type=raw,value=sbtc-emily-cron-latest,enable=${{ github.ref_name == 'environment/mainnet' }}
```
matteojug: nit: this last tag is never enabled as we don't have mainnet as possible env

.github/workflows/emily-cron-ecr-build.yaml
```diff
@@ -0,0 +1,150 @@
...
+
+      - name: Build and Push
+        id: docker_build
+        uses: stacks-sbtc/actions/docker/build-push-action@942832cc7f86ead03c05b7035cd4b9573dd6c617
+        with:
+          file: ./docker/sbtc/emily-cron/Dockerfile
+          context: ./docker/sbtc
```
matteojug: This context seems wrong, it should be the project root (as we are copying the `emily_cron` dir from the root) , unless it works differently with the action wrt to the usual `docker compose`

.github/workflows/emily-cron-ecr-build.yaml
```diff
@@ -0,0 +1,150 @@
...
+  resolve-env:
+    name: Resolve Environment
+    runs-on: ubuntu-latest
+    outputs:
+      environment: ${{ steps.set_env.outputs.environment }}
+      should_run: ${{ steps.set_env.outputs.should_run }}
+    env:
+      GH_TOKEN: ${{ github.token }}
+    steps:
+      - name: Resolve Environment
```
matteojug, line 43: sanity check: if we use the labels to trigger a workflow, using a single tag it will run both the build images and the Emily deployment. It's true that we could then just reject the workflow run when it ask for an approval, but I wanted to double check that this _run everything related when we specify (eg) `sbtc-devnet`_ is what we want

.github/workflows/emily-cron-ecr-build.yaml
```diff
@@ -0,0 +1,150 @@
...
+            ENVIRONMENT="${MATCHED_ENVS[0]}"
+          else
+            echo "### Unsupported trigger: ${{ github.event_name }}" >> $GITHUB_STEP_SUMMARY
+            exit 1
+          fi
+
+          echo "environment=$ENVIRONMENT" >> $GITHUB_OUTPUT
+          echo "should_run=$SHOULD_RUN" >> $GITHUB_OUTPUT
+
+  image:
```
matteojug, line 99: we could use a matrix for the `image` job to build both the cron and the sidecar, since they seems identical other than for the target docker file and tags

.github/workflows/emily-cron-ecr-build.yaml
fabrizio-stacks: that's true, probably in the future we will have it but let's remove it for now

.github/workflows/emily-cron-ecr-build.yaml
fabrizio-stacks: mmm what if we don't update them at the same time?

.github/workflows/emily-cron-ecr-build.yaml
fabrizio-stacks: this is true

.github/workflows/emily-cron-ecr-build.yaml
matteojug: oh right, I was just thinking about the PR label case. For the dispatch, we could add some checkboxes, but maybe it's not really worth it then. Leaving it to you, feel free to resolve this convo if no action is required.

.github/workflows/emily-cron-ecr-build.yaml
BowTiedDevOps: That's correct, I was just waiting for the org transition for both the sbtc and actions repositories because we'll still have to update the commit hashes/actions used, otherwise I'll have to open 3 more PRs both here and on the infra repo now and after the transition.

If you believe we should do it now, I'll move this part and edit this PR/open another PRs for the other actions that follow this format

.github/workflows/emily-cron-ecr-build.yaml
BowTiedDevOps: This is not an issue since the latest commit (`a1754bb`) only has the setup terraform action extra compared to the one used here (`942832c`), so it doesn't affect these jobs. Same mention as above, they will have to be updated either way with the org transition, so we could either update them now or then. It's up to you, please let me know how I should proceed

.github/workflows/emily-cron-ecr-build.yaml
BowTiedDevOps: That's correct, they can just be declined/cancelled and we only run what we want. Basically all 3 actions will run, the emily deploy, cron and sidecar builds, but we can choose which to allow and which to decline.

.github/workflows/emily-cron-ecr-build.yaml
matteojug: nah just a nit in case we wanted to do something about it now. If we are planning to do something about it in the future it sounds good to me

.github/workflows/emily-cron-ecr-build.yaml
BowTiedDevOps: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1843/commits/263b78b26d26066afec2e6e8b458f7770c4337b0

.github/workflows/emily-cron-ecr-build.yaml
BowTiedDevOps: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1843/commits/263b78b26d26066afec2e6e8b458f7770c4337b0

.github/workflows/emily-cron-ecr-build.yaml
BowTiedDevOps: Updated them in https://github.com/stacks-sbtc/sbtc/pull/1843/commits/263b78b26d26066afec2e6e8b458f7770c4337b0 and https://github.com/stacks-sbtc/sbtc/pull/1843/commits/1c4488494c4f6d93419fb8691e0991444067861c

.github/workflows/emily-cron-ecr-build.yaml
matteojug: Also for the sidecar

.github/workflows/emily-cron-ecr-build.yaml
BowTiedDevOps: We could do either way, but I think merging them in a matrix and having an input adds complexity to the dispatch flow, while for the PR flow you can't really run only one of them with the approval/cancel flow - so it's either that both or none run. I'm opting to keep them separate for the possibility of only running one of them in the PR and not adding complexity to the workflows, but will do as you decide here @matteojug @fabergat

.github/workflows/emily-cron-ecr-build.yaml
BowTiedDevOps: Sidecar: https://github.com/stacks-sbtc/sbtc/pull/1843/commits/094154c2e177fc1309be30198d00abe160a1de34

.github/workflows/emily-cron-ecr-build.yaml
BowTiedDevOps: Sidecar fix: https://github.com/stacks-sbtc/sbtc/pull/1843/commits/094154c2e177fc1309be30198d00abe160a1de34

djordon (approved):
Looks good

PR #1850, opened 2025-11-03
fix: handle testnet stacks blocks backfill

matteojug:
## Description

Closes: https://github.com/stacks-sbtc/sbtc/issues/1749
Supersedes: https://github.com/stacks-sbtc/sbtc/pull/1803

## Changes

During Stacks blocks backfill, if we encounter a 404 when fetching the tenure for a block we also check if that block is a pre-Nakamoto one and, if so, we stop the backfilling process.

## Testing Information

Added a green ignored integration test targeting testnet (`fetch_unknown_ancestors_works_in_testnet`); devenv backfill worked as expected.

## Checklist

### DIFF
--- a/signer/src/stacks/api.rs
+++ b/signer/src/stacks/api.rs
@@ -250,7 +250,7 @@ pub trait StacksInteract: Send + Sync {
     ) -> impl Future<Output = Result<NakamotoBlock, Error>> + Send;
 
     /// Returns `Ok` if the given block ID is a pre-Nakamoto block, otherwise
-    /// (the block doesn't exists or is a Nakamoto one) `Err` is returned.
+    /// (the block doesn't exist or is a Nakamoto one) `Err` is returned.
     fn check_pre_nakamoto_block(
         &self,
         block_id: &StacksBlockId,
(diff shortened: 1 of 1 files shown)


signer/src/stacks/api.rs
```diff
@@ -1986,6 +2054,66 @@ mod tests {
...
+            StacksClient::new(Url::parse("https://api.testnet.hiro.so/").unwrap()).unwrap();
+
+        // Testnet currently has the following structure:
+        //
+        // BTC 1865 <- Stacks 319
+        // BTC 1900 -- nakamoto_start_height
+        // BTC 1901 <- Stacks 320, ..., 744
+        // BTC 1998 <- Stacks 745, ..., 750, ...
+
+        // This is the block id for block 319 (pre-Nakamoto) on testnet
```
djordon, line 2072: Nice, very helpful

MCJOHN974:
I wonder why at all we want to have pre-Nakamoto blocks in the DB? I think these blocks does not have any sBTC related transactions, and why we need them?

matteojug:
> I wonder why at all we want to have pre-Nakamoto blocks in the DB? I think these blocks does not have any sBTC related transactions, and why we need them?

You're right: we don't want them and we are not storing them. The issue is that if we have gaps around the Nakamoto start height we may ask for the tenure for a pre-Nakamoto block, causing an err and failing the backfilling. With this we are still not storing them, we just check if the error we got is because of a pre-Nakamoto block, and if so we did reach the Nakamoto start height so we can stop the process.

matteojug:
> LGTM! Also, don't we want to backfill not all Nakamoto blocks, but just some blocks (for example, a few months ago from now), just by hardcoding block hash&height in each release? This will save some space, and I don't think we need all these old blocks.

It may be possible to ignore some initial blocks (maybe even after the contract deployment if we are being careful) but if misused it may led to not backfilling enough important blocks. Also usually storage is not the bottleneck so it may not be a priority right now.

PR #1851, opened 2025-11-03
feat: reduce memory usage when fetching stacks blocks headers for a tenure

djordon:
## Description

This PR introduces more local types for stacks core types, continuing on https://github.com/stacks-sbtc/sbtc/issues/1835. It also reduces the amount of memory that can be used when fetching blocks from stacks core, implementing part of the fix for https://github.com/stacks-sbtc/sbtc/issues/1817 where we make sure to reduce the amount of memory that can be used when fetching stacks block for any given tenure.

## Changes

* Use a local type for the response of `GET /v3/tenures/info` RPC to stacks core.
* Only keep the necessary parts of the headers from nakamoto blocks returned from `GET /v3/tenures/<block-id>` RPCs to stacks core.
* Manually implement serialize for `StacksBlockHash`, the same commit for https://github.com/stacks-sbtc/sbtc/pull/1842.

## Testing Information

I just started the build for testing this out on devenv, will report back.

## Checklist

### DIFF
--- a/.github/workflows/emily-cron-ecr-build.yaml
+++ b/.github/workflows/emily-cron-ecr-build.yaml
@@ -1,0 +1,149 @@
+name: Build Docker Image for Emily Cron
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Environment name"
+        required: true
+        default: environment/sbtc-devnet
+        type: choice
+        options:
+          - environment/sbtc-devnet
+          - environment/sbtc-private-mainnet
+  pull_request:
+    branches:
+      - main
+
+concurrency:
+  group: docker-build-emily-cron-${{ github.ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  id-token: write
+  contents: read
+  attestations: write
+  packages: write
+  issues: read
+  pull-requests: read
+
+env:
+  docker_platforms: "linux/amd64"
+
+jobs:
+  resolve-env:
+    name: Resolve Environment
+    runs-on: ubuntu-latest
+    outputs:
+      environment: ${{ steps.set_env.outputs.environment }}
+      should_run: ${{ steps.set_env.outputs.should_run }}
+    env:
+      GH_TOKEN: ${{ github.token }}
+    steps:
+      - name: Resolve Environment
+        id: set_env
+        run: |
+          ALLOWED_ENVS=("environment/sbtc-devnet" "environment/sbtc-private-mainnet")
+          SHOULD_RUN=true
+
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            if [ "${GITHUB_REF_NAME}" != "main" ]; then
+              echo "### The workflow_dispatch can only be run on main branch." >> $GITHUB_STEP_SUMMARY
+              exit 1
+            fi
+
+            ENVIRONMENT="${{ github.event.inputs.environment }}"
+
+            VALID=false
+            for env in "${ALLOWED_ENVS[@]}"; do
+              if [ "$ENVIRONMENT" = "$env" ]; then
+                VALID=true
+                break
+              fi
+            done
+
+            if [ "$VALID" != "true" ]; then
+              echo "### Invalid workflow_dispatch environment: $ENVIRONMENT" >> $GITHUB_STEP_SUMMARY
+              echo "### Allowed environments: ${ALLOWED_ENVS[*]}" >> $GITHUB_STEP_SUMMARY
+              exit 1
+            fi
+          elif [ "${{ github.event_name }}" = "pull_request" ]; then
+            LABELS=$(gh api repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/labels --jq '.[].name')
+
+            MATCHED_ENVS=()
+            for label in $LABELS; do
+              for env in "${ALLOWED_ENVS[@]}"; do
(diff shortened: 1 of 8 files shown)


signer/src/storage/model.rs
```diff
@@ -987,6 +1005,103 @@ impl std::fmt::Debug for StacksBlockHash {
...
+/// This type is displayed as a lowercase hex string, and mirrors what
+/// stacks-core does for the
+/// [`stacks_common::types::chainstate::ConsensusHash`] type.
+///
+/// The stacks-core Display implementation can be found in [1-2].
+///
+/// [1]: <https://github.com/stacks-network/stacks-core/blob/bd9ee6310516b31ef4ecce07e42e73ed0f774ada/stacks-common/src/util/macros.rs#L499-L511>
+/// [2]: <https://github.com/stacks-network/stacks-core/blob/bd9ee6310516b31ef4ecce07e42e73ed0f774ada/stacks-common/src/types/chainstate.rs#L382-L386>
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ConsensusHash([u8; 20]);
```
MCJOHN974, line 1037: I don't really like copying of stacks-core struct here. I agree that `ConsensusHash` sounds very unlikely to be changed, but if it is changed, it will be annoying... So, if we want to keep this local type, I'd suggest adding some unit tests checking that this local consensus hash copies the behavior of stacks-common consensus hash. These tests should fail if a newer version of stacks-common changes `ConsensusHash` behavior.

djordon: We spoke about this out of band and we're all kinda aligned on having our own types for stacks-core related stuff. As for the tests, I definitely agree. I have some for the `Display`, `Debug`, `serde::Serialize`, and `serde::Deserialize` traits.

signer/src/stacks/api.rs
```diff
@@ -1138,14 +1112,14 @@ impl StacksClient {
     /// If the given block ID does not exist or is an ID for a non-Nakamoto
     /// block then a Result::Err is returned.
     #[tracing::instrument(skip(self))]
-    async fn get_tenure(&self, block_id: &StacksBlockId) -> Result<TenureBlocks, Error> {
+    async fn get_tenure(&self, block_id: &StacksBlockHash) -> Result<TenureBlockHeaders, Error> {
```
cylewitruk: I'd consider renaming this to `get_tenure_headers` since that's what it's doing now.

signer/src/stacks/api.rs
```diff
@@ -1202,7 +1174,10 @@ impl StacksClient {
     /// * If the given block ID does not exist or is an ID for a
     ///   non-Nakamoto block then a Result::Err is returned.
     #[tracing::instrument(skip(self))]
-    async fn get_tenure_raw(&self, block_id: &StacksBlockId) -> Result<Vec<NakamotoBlock>, Error> {
+    async fn get_tenure_raw(
```
cylewitruk: Same here, I'd probably try to find a better naming since we're not really getting a "raw tenure" here -- more like `get_block_headers_for_tenure` or something?

signer/src/storage/model.rs
```diff
@@ -987,6 +1005,103 @@ impl std::fmt::Debug for StacksBlockHash {
...
+
```
cylewitruk, line 1038: Any reason to not put these kinds of types in the `sbtc` crate? It's long nagged at me that `storage/model.rs` tends to accumulate "types", particularly those which are not specifically storage models. For the stacks-core/common interop types specifically, they may be useful for the event layer that was moved to `sbtc`.

djordon: > Any reason to not put these kinds of types in the `sbtc` crate?

Yeah, only the signer crate needs this type, the `sbtc` crate and all of our other crates don't need it.

> It's long nagged at me that `storage/model.rs` tends to accumulate "types", particularly those which are not specifically storage models.

Yeah, which module to put these types is a good point. I figure we can address that in some other PR that is specific to that kind of refactor so I put this new type in one of the "usual" places until we settle on this other home. I could put it in stacks API module, I only put it here because it shares the same implementations as `StacksBlockHash` and `StacksTxId` and it was easier to copy pasta in the same file.

> For the stacks-core/common interop types specifically, they may be useful for the event layer that was moved to `sbtc`.

Well, some of these, like `StacksBlockHash`, we want to have in the signer crate because we want to implement sqlx traits on them. But maybe there are others that are more useful in the `sbtc`? It's certainly possible, I haven't checked. Oh maybe the `StacksTxId` type, but we need a local version here too.

signer/src/stacks/api.rs
djordon: I went with `get_tenure_headers_raw`, although that name isn't the best either.

signer/src/stacks/api.rs
djordon: Changed in https://github.com/stacks-sbtc/sbtc/pull/1851/commits/aad0d6ced8fe80875517cc1d43b098f8d47c79e1

signer/src/stacks/api.rs
djordon: Changed in https://github.com/stacks-sbtc/sbtc/pull/1851/commits/aad0d6ced8fe80875517cc1d43b098f8d47c79e1

cylewitruk: Better than implying full tenure data anyway :)

PR #1852, opened 2025-11-04
[chore]: add machete

MCJOHN974:
## Description

Closes: no issue

## Changes

Adds `cargo-machete`, a tool for detecting unused dependencies in `Cargo.toml`.

Note: this tool does not give a 100% guarantee that it will find _all_ unused dependencies, so we still should look if the dependency is necessary during a review.
Note 2: There is a similar tool named `cargo-udeps`, so maybe we should consider using it/both.

## Testing Information
no testing needed
## Checklist

### DIFF
--- a/.github/workflows/on-push.yaml
+++ b/.github/workflows/on-push.yaml
@@ -80,7 +80,7 @@ jobs:
 
       - name: Install Machete
         id: install_machete
-        run: cargo install cargo-machete@0.9.1
+        run: cargo install --git https://github.com/bnjbvr/cargo-machete --rev 7959c845782fed02ee69303126d4a12d64f1db18 cargo-machete
 
       - name: Install Package Dependencies
         id: install_dependencies


MCJOHN974:
As you see [50ce7e8](https://github.com/stacks-sbtc/sbtc/pull/1852/commits/50ce7e8cac49cd1b156b898a0eb3df7b6cf67ce5) we have an unused dependency currently

.github/workflows/on-push.yaml
```diff
@@ -78,6 +78,10 @@ jobs:
           components: clippy, rustfmt
           cache-key: "rust-tests"
 
+      - name: Install Machete
+        id: install_machete
+        run: cargo install cargo-machete
+
```
MCJOHN974, line 84: It seems we want to cache this installation... However, integration tests take longer anyway

MCJOHN974:
Hmm, it seems my changes wasn't a reason for red CI in [3accda6](https://github.com/stacks-sbtc/sbtc/pull/1852/commits/3accda6e7f013a543c5e701d09e6c68815aafc54), so I guess we found a flaky test

BowTiedDevOps (approved):

djordon (commented):
Honestly, I could do without having to install another thing just to remove unused dependencies. But I guess it's not so bad either. Maybe a decent middle ground is to have it run in CI but leave it out of `make lint`?

To be clear, I'm not too thrilled about my suggestion, I just don't want to add another thing unless it gives us some material benefit and I'm not really seeing it here with `cargo machete`.

.github/workflows/on-push.yaml
BowTiedDevOps: My go-to for installations like this is if it takes less than 1 minute to say, better to keep it separate, otherwise you'd give yourself headaches with the caching. This seems to be cached on the runner anyways, since the latest job took 0s on the installation, as opposed to the first run's 51s.
image

MCJOHN974:
@djordon moved machete to CI only: [207696f](https://github.com/stacks-sbtc/sbtc/pull/1852/commits/207696fe1ca6acce1f7e300cfb7c29334b05e8e8)

.github/workflows/on-push.yaml
```diff
@@ -78,10 +78,18 @@ jobs:
           components: clippy, rustfmt
           cache-key: "rust-tests"
 
+      - name: Install Machete
+        id: install_machete
+        run: cargo install cargo-machete
```
matteojug: Can we pin the version of `cargo-machete`?

MCJOHN974: [70c8ede](https://github.com/stacks-sbtc/sbtc/pull/1852/commits/70c8ede1817e2b59fc66e6fa8203fcdcde1daa73)

djordon (approved):
Looks good

PR #1857, opened 2025-11-11
feat: improve stacks blocks backfilling

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1817.

## Changes

1. Adds a new table for writing stacks blocks to "temporarily".
2. Adds new functions for writing to the new temporary table, truncating it, and copy its contents to the `stacks_blocks` table.
3. Modifies the `fetch_unknown_ancestors` function to write blocks to the database one tenure's worth of blocks at a time. Also renames the function so that it better aligns with what it does now.
4. Marked some trait methods as for testing only, since that's where they are currently used.
5. I changed the `cargo build` command used in `make build` to use `--bins --lib` instead of `--all-targets`. Using `--all-targets` meant that test binaries were built -- according to https://doc.rust-lang.org/cargo/commands/cargo-build.html -- which meant that the `testing` feature in the signer crate was enabled in both the testing binary and the main binary due to how cargo unifies features when building targets. However, we do not want this so I included only the necessary targets when compiling a debug and release builds.

I added the change in (5) here because I used it to make sure that things were built correctly for my devenv tests, but I can see it warrenting it's own PR. So if anyone has any strong opinions I can PR that commit too.

## Testing Information

I tested everything up to commit 4bed294a82c402a7213e8e57b739c730a43e86de on devenv. Everything worked fine. I also made sure that the `--all-targets`/`--bins --lib` switch worked as expected. I also build a release image to make sure that everything still at least build correctly and that succeeded.

## Checklist

### DIFF
--- a/signer/src/storage/memory/write.rs
+++ b/signer/src/storage/memory/write.rs
@@ -174,20 +174,17 @@ impl DbWrite for SharedStore {
     }
 
     async fn write_stacks_block_headers(&self, headers: &TenureBlockHeaders) -> Result<(), Error> {
-        let block_iter = headers.clone().into_iter();
         let mut store = self.lock().await;
         store.version += 1;
 
-        headers.headers().iter().for_each(|header| {
+        headers.clone().into_iter().for_each(|header| {
             store
                 .bitcoin_anchor_to_stacks_blocks
                 .entry(headers.anchor_block_hash)
                 .or_default()
-                .push(header.block_id);
+                .push(header.block_hash);
+            store.stacks_blocks.insert(header.block_hash, header);
         });
-        store
-            .stacks_blocks
-            .extend(block_iter.map(|header| (header.block_hash, header)));
 
         Ok(())
     }


signer/src/storage/postgres/write.rs
```diff
@@ -1002,13 +1005,128 @@ impl PgWrite {
...
+        let mut bitcoin_anchors = Vec::with_capacity(headers.len());
+
+        for header in headers {
+            block_ids.push(header.block_id);
+            parent_block_ids.push(header.parent_block_id);
+            let block_height =
+                i64::try_from(header.block_height).map_err(Error::ConversionDatabaseInt)?;
+            chain_lengths.push(block_height);
+            bitcoin_anchors.push(tenure_headers.anchor_block_hash);
+        }
```
MCJOHN974: Maybe better something like
```rust
let (block_ids, parent_block_ids, chain_lengths, bitcoin_anchors) = headers.iter().try_fold(
    Vec::with_capacity(headers.len()),
    Vec::with_capacity(headers.len()),
    Vec::with_capacity(headers.len()),
    Vec::with_capacity(headers.len()),
    |mut block_ids, mut parent_block_ids, mut chain_lengths, mut bitcoin_anchors, header| {
        block_ids.push(header.block_id);
            parent_block_ids.push(header.parent_block_id);
            let block_height =
                i64::try_from(header.block_height).map_err(Error::ConversionDatabaseInt)?;
            chain_lengths.push(block_height);
            bitcoin_anchors.push(tenure_headers.anchor_block_hash);
            (block_ids, parent_block_ids, chain_lengths, bitcoin_anchors)
    })
```

signer/src/stacks/api.rs
```diff
@@ -1378,24 +1398,36 @@ impl StacksClient {
...
 {
-    let starting_tenure = stacks.get_tenure_headers(block_id).await?;
-    let mut headers: Vec<TenureBlockHeaders> = vec![starting_tenure];
+    let mut tenure = stacks.get_tenure_headers(block_id).await?;
+    let end_height = tenure.end_height();
     let nakamoto_start_height = stacks.get_epoch_status().await?.nakamoto_start_height();
 
-    while let Some(tenure) = headers.last() {
+    // Truncate the `stacks_blocks_temp` table to ensure that we start with
+    // a clean slate.
```
MCJOHN974:
```suggestion
    // a clean state.
```

signer/src/stacks/api.rs
```diff
@@ -1378,24 +1398,36 @@ impl StacksClient {
...
-pub async fn fetch_unknown_ancestors<S, D>(
+/// This function fetches all unknown nakamoto blocks that are on the
+/// canonical chain identified by the given StacksBlockHash chain tip that
+/// not already stored in the database. It fetches these blocks one tenure
+/// at a time, and then writes them to the `stacks_blocks_temp` table.
+/// After all such blocks have been fetched, the function copies the blocks
+/// from the `stacks_blocks_temp` table to the `stacks_blocks` table.
+/// Things are done this way to ensure that updates to the `stacks_blocks`
+/// table are completed atomically.
+pub async fn update_db_with_unknown_ancestors<S, D>(
```
MCJOHN974, line 1413: Maybe add some words on why we can't achieve this atomicity via some SQL transactions or so

signer/src/stacks/api.rs
```diff
@@ -389,6 +392,23 @@ impl TenureBlockHeaders {
...
+    pub fn min_height(&self) -> StacksBlockHeight {
+        // SAFETY: It is okay to unwrap here because we know that the
+        // tenure is non-empty. The struct upholds this invariant upon
+        // creation.
+        self.headers.iter().map(|h| h.block_height).min().unwrap()
+    }
+
+    /// Get the height of the block with the greatest height of all blocks
+    /// held within this struct.
+    pub fn end_height(&self) -> StacksBlockHeight {
```
MCJOHN974, line 408: nit: maybe better both `min/max_height` or both `start/end_height`?

signer/src/stacks/api.rs
djordon: Oh, yeah. A "clean slate" is a phrase that means basically the same thing as "clean state". It didn't dawn on me that this wasn't a universally known phrase.

signer/src/stacks/api.rs
djordon: Oh we can achieve this atomicity with a SQL transaction by using SQLx's transaction object thingy. There are some underlying differences on how the database does the writes and what not, but it would be roughly equivalent to what's here.

signer/src/stacks/api.rs
djordon: Yeah, `start/end` height.

signer/src/stacks/api.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1857/commits/736a52cf3076ea1d326b402c49ffd765cf378f5d

signer/src/stacks/api.rs
djordon: Changed in https://github.com/stacks-sbtc/sbtc/pull/1857/commits/3ec10cb4ae407d4f1f55c53869f0d675fb2a92f9

signer/src/storage/postgres/write.rs
djordon: Although I usually prefer these adapters, I find the adapter approach slightly harder to read and understand in this case.

signer/src/stacks/api.rs
MCJOHN974: Whoaa, thanks for english lesson, never heard about it, thought it is a typo!

signer/src/storage/postgres/write.rs
MCJOHN974: I don't have strong opinion on this so second reviewer feel free to write to this thread

signer/src/stacks/api.rs
MCJOHN974: Hmmm... For me adding temporary sql table is kinda strange (even if alternative doing same under the hood), so, I'd add a few lines of comments explaining such design. Not because I don't want to see such design in the codebase, but to make it easy to read for somebody who didn't review this PR, or somebody who will join project when we will have 30th migration

signer/src/stacks/api.rs
```diff
@@ -2080,23 +2110,26 @@ mod tests {
...
-        assert_eq!(blocks, expected);
+        let (min_block_height, max_block_height) = sqlx::query_as::<_, (i64, i64)>(
+            r#"SELECT 
+                 MIN(block_height) as min_block_height
+               , MAX(block_height) as max_block_height
+             FROM sbtc_signer.stacks_blocks"#,
+        )
+        .fetch_one(db.pool())
+        .await
+        .unwrap();
```
matteojug: nit: can we check we have all the blocks in that interval instead of just the boundary? It is an ignored test, but also it's one we can easily run to check the fetching works as expected; checking only the boundary may be too lax.

signer/src/stacks/api.rs
```diff
@@ -1442,11 +1474,16 @@ where
...
-        headers.push(tenure_headers);
+        tenure = tenure_headers;
     }
 
-    headers.reverse();
-    Ok(headers)
+    let start_height = tenure.start_height();
+
+    db.copy_from_stacks_blocks_temp_table().await?;
+    db.truncate_stacks_blocks_temp_table().await?;
```
matteojug: A difference wrt the previous approach is that now there exist an unlucky (but currently impossible, unless running two instances of the sbtc signer against the same db) interleaving of operations that may create gaps in the db view of the chain if we have two concurrent run of `update_db_with_unknown_ancestors`:
1. A does all the work up until the table copy
2. B truncates and start to write some headers in the freshly truncated temp
3. A copies temp
4. B add some more headers
5. A truncate temp
6. B continue to fill, then copy and truncate

The headers at point 4 are lost, but having more recent headers in the db we will not fill that gap in successive runs.
Now, as mentioned above it shouldn't be possible to have concurrent run of `update_db_with_unknown_ancestors` unless running the binary twice, but I don't recall we had to explicitly rely on a singleton running binary for any other operation.

signer/src/stacks/api.rs
djordon: Right, good idea.

signer/src/stacks/api.rs
djordon: Ohhh, well this might be easy to address. Let me see here.

djordon: I wrapped the entire thing in a transaction in https://github.com/stacks-sbtc/sbtc/pull/1857/commits/526a51ea41ecf014645efbf458323b6a19434db0, which should address this.

signer/src/stacks/api.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1857/commits/5391fe994947a71d528eddd7b98b17851dd8d909. I counted distinct block hashes, which I think is enough with the min and max block heights.

signer/src/stacks/api.rs
matteojug: I'm not entirely sure because of the truncate (https://www.postgresql.org/docs/current/mvcc-caveats.html). But then, if we are using a transaction, do we need a temp table at all? vs just writing to the stacks block table (in the same transaction).

signer/src/stacks/api.rs
djordon: I ended up wrapping everything in SQLx's transaction object thingy in https://github.com/stacks-sbtc/sbtc/commit/526a51ea41ecf014645efbf458323b6a19434db0. As for the explanation, I assume you want a few more lines explaining why we're doing it this way right? Something more than "ensuring that things are done atomically"?

signer/src/stacks/api.rs
djordon: I see, a few notes:
* Yeah, I remembered that `TRUNCATE` was special and so I looked into it before hand, and I don't think that it will be a problem.
* I don't think we need a temp table at all, so yeah we can simplify.
* I was, and still am, a bit nervous of using a long running transaction for the backfilling here. I was bitten by a SQLx bug in the past, and so I'm hesitant about these things today. For this change, I was going to ask that we test this out a bit on testnet and private-mainnet before cutting a proper release. Basically truncate the stacks blocks table and see how it backfills happens. We should do this regardless of our implementation here but this came to mind as I changed to a transaction based flow.

On net I agree that where we're at now with a sql transaction and a temp table is strange. We should fully commit one way or the other, let's go with a transaction.

djordon: Okay, I simplified everything now. I like it more, but I'm still nervous.

signer/src/storage/memory/write.rs
```diff
@@ -172,21 +173,14 @@ impl DbWrite for SharedStore {
...
-    ) -> Result<(), Error> {
+    async fn write_stacks_block_headers(&self, headers: &TenureBlockHeaders) -> Result<(), Error> {
+        let block_iter = headers.clone().into_iter();
         let mut store = self.lock().await;
         store.version += 1;
 
-        blocks.iter().for_each(|block| {
-            store.stacks_blocks.insert(block.block_hash, block.clone());
-            store
-                .bitcoin_anchor_to_stacks_blocks
```
matteojug, line 185: We use `bitcoin_anchor_to_stacks_blocks` in `get_stacks_chain_tip`, I would expect some test to fail since we are not populating it in here anymore 🤔

matteojug (commented):
LGTM, just a sanity check

signer/src/storage/memory/write.rs
djordon: Yeah good point. I'll add it back for sure, but yeah it's strange. I also expect that things would fail, since we appear to need it for getting the stacks chain tip, which is used in quite a few places.

djordon: Oh maybe our tests don't use `write_stacks_block_headers` but use `write_stacks_block` instead. The only tests that would use `write_stacks_block_headers` must go through the block observer and I don't think we have tests that use functions where missing this info would lead to a failure.

signer/src/stacks/api.rs
matteojug: We tested this on testnet and it backfilled fine, resolving the convo

PR #1858, opened 2025-11-12
feat: add testcontainers compose scaffolding

matteojug:
## Description

Partial work for https://github.com/stacks-sbtc/sbtc/issues/1854

DRAFT: looking for some initial feedbacks

## Changes

This PR introduces some scaffolding to use `testcontainers` (via compose) for test isolation, and integrate it in a couple of tests as an example: tests can bring up the relevant stack, and then get clients for the required services; on drop the compose stack is removed.

Bringing up the stack does add some time (in the order of ~2s for up+down), so we should probably keep some tests serial reusing a common stack and parallelize only heavy tests (or the ones requiring isolation, like (hopefully) some future stacks-running ones).

Other noteworthy things:
 - `testcontainers` support is not released yet (thus the github crate)
 - Updated `tokio` because of `testcontainers` (indirect) dependency requirement
 - Fixed a flaky test in `signer/src/block_observer.rs` after the tokio update
 - Removed some unused files

For the non-compose approach see:
 - https://github.com/stacks-sbtc/sbtc/pull/1572
 - https://github.com/stacks-sbtc/sbtc/compare/main...cyle/testcontainers-nov-25

<details>
  <summary>Dependency error</summary>

```
error: failed to select a version for `windows-targets`.
    ... required by package `windows-sys v0.59.0`
    ... which satisfies dependency `windows-sys = "^0.59"` of package `etcetera v0.10.0`
    ... which satisfies dependency `etcetera = "^0.10.0"` of package `testcontainers v0.25.2 (https://github.com/testcontainers/testcontainers-rs?rev=9aae21e7330f3ad22a7742c82dd5eea465928510#9aae21e7)`
    ... which satisfies git dependency `testcontainers` of package `sbtc v0.1.0 ([...]/sbtc/sbtc)`
    ... which satisfies path dependency `sbtc` (locked to 0.1.0) of package `emily-handler v0.1.0 ([...]/sbtc/emily/handler)`
    ... which satisfies path dependency `emily-handler` (locked to 0.1.0) of package `emily-openapi-spec v0.1.0 ([...]/sbtc/emily/openapi-gen)`
versions that meet the requirements `^0.52.6` are: 0.52.6

all possible versions conflict with previously selected packages.

  previously selected package `windows-targets v0.52.5`
    ... which satisfies dependency `windows-targets = "^0.52.0"` (locked to 0.52.5) of package `windows-sys v0.52.0`
    ... which satisfies dependency `windows-sys = "^0.52"` (locked to 0.52.0) of package `tokio v1.43.0`
    ... which satisfies dependency `tokio = "^1.43.0"` (locked to 1.43.0) of package `blocklist-client v0.0.1 ([...]/sbtc/blocklist-client)`
    ... which satisfies path dependency `blocklist-client` (locked to 0.0.1) of package `blocklist-openapi-gen v0.1.0 ([...]/sbtc/blocklist-openapi-gen)`

failed to select a version for `windows-targets` which could resolve this conflict
```
</details>

## Checklist


.github/workflows/on-push.yaml
```diff
@@ -305,7 +305,7 @@ jobs:
 
       - name: Run Integration Tests (Rust)
         id: run_rust_tests
-        run: cargo-nextest nextest --color always run --no-fail-fast --test-threads 1 --partition hash:${{ matrix.partition }}/2 --archive-file nextest-archive-serial.tar.zst
+        run: cargo-nextest nextest --color always run --no-fail-fast --test-threads 1 --partition hash:${{ matrix.partition }}/2 --archive-file nextest-archive-serial.tar.zst --workspace-remap .
```
matteojug, line 308: We need the compose files (and we cannot add them to the nextest archives, unless we move them to `target`), so we just tell it to remap the source code -- it's probably unnecessary since the GH worker should use the same checkout path between the different jobs, but better to be explicit about it

signer/tests/integration/postgres.rs
```diff
@@ -129,7 +133,6 @@ async fn should_be_able_to_query_bitcoin_blocks() {
             .expect("failed_to_execute_query");
         assert!(result.is_none());
     }
-    signer::testing::storage::drop_db(store).await;
```
matteojug, line 132: This call got stuck for some reason, maybe I was forgetting some open connection somewhere.

docker/tests/devnet-liveness.sh
djordon, line 1: Is this all incorrect now?

docker/sbtc/emily-aws-setup/Dockerfile.ci
djordon, line 1: Oh we build a binary, so this is unused?

signer/tests/integration/containers.rs
```diff
@@ -0,0 +1,305 @@
...
+const SERVICE_DATABASE: &str = "postgres";
+const SERVICE_DATABASE_PORT: u16 = 5432;
+const SERVICE_DATABASE_USER: &str = "postgres";
+const SERVICE_DATABASE_PASSWORD: &str = "postgres";
+
+fn compose_path(file_name: &str) -> String {
+    // Evaluate `CARGO_MANIFEST_DIR` at runtime for nextest relocation
+    format!(
+        "{}/../docker/tests/{file_name}",
+        std::env::var("CARGO_MANIFEST_DIR").unwrap()
```
djordon, line 33: This is fine, but I believe we can just use `std::env!("CARGO_MANIFEST_DIR")` since cargo always sets that environment variable.

signer/tests/integration/containers.rs
```diff
@@ -0,0 +1,305 @@
...
+impl BitcoinContainer {
+    pub async fn from_containers(containers: &TestContainers) -> Self {
+        let auth = bitcoincore_rpc::Auth::UserPass(
+            BITCOIN_CORE_RPC_USERNAME.to_string(),
+            BITCOIN_CORE_RPC_PASSWORD.to_string(),
+        );
+        let url = containers
+            .get_bitcoin_url()
+            .await
+            .expect("cannot get bitcoin url");
```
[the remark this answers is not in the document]
djordon: 👎🏽 I think `expect` also prints the error, so their suggestion is roughly equivalent to what is already there.

signer/tests/integration/containers.rs
```diff
@@ -0,0 +1,305 @@
...
+}
+
+pub trait TestContainersBuilderExt {
+    async fn start(self) -> TestContainers;
+
+    fn with_bitcoin(self) -> Self;
+    fn with_database(self) -> Self;
+    fn with_emily(self) -> Self;
+
+    async fn start_bitcoin() -> TestContainers;
```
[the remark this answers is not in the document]
djordon: Doesn't clippy catch this? If it doesn't perhaps we should change it so that it does.

signer/tests/integration/postgres.rs
djordon: Does the database automatically disappear when the test ends only if the test succeeds?

signer/tests/integration/containers.rs
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    pub fn get_client(&self) -> BitcoinCoreClient {
+        self.url().try_into().expect("cannot create bitcoin client")
+    }
+}
+
+pub struct DatabaseContainer {
+    url: String,
+}
```
djordon, line 189: Do the database containers, or any of these containers, stick around after the test finishes? One thing that I currently do when debugging or developing is make my an integration test fail so that I can examine the database at the end of the test. It doesn't look like it sticks around, which may be a bummer. If that's the case, then I'd prefer our current approach over using test containers for postgres.

djordon: Hmmm, if we can keep the container around by default and spin them down at the end of a test, then I suppose we can get a setup that is similar to our current setup by printing out the port when a container is created.

djordon (commented):
Largely looks good.

My main question now is around when a container gets spun down. It's quite nice to be able to interact with our containers after a test has run, and I hope they have that capability.

signer/tests/integration/containers.rs
```diff
@@ -0,0 +1,305 @@
...
+    pub fn get_faucet(&self) -> Faucet<'_> {
+        Faucet::new(FAUCET_SECRET_KEY, AddressType::P2wpkh, &self.rpc_client)
+    }
+
+    pub fn get_client(&self) -> BitcoinCoreClient {
+        self.url().try_into().expect("cannot create bitcoin client")
+    }
+}
+
+pub struct DatabaseContainer {
```
cylewitruk, line 187: Btw in my implementation, I opted to not containerize postgres per-test because our current solution with db-per-test-spawning is thread-safe and more efficient than spawning containers. I of course see the benefits of going all-out containers-per-test for consistency, but if the goal is speed then db-per-test will always win (with the downside of still needing an `integration-env-up`-like thingy.

docker/sbtc/emily-aws-setup/Dockerfile.ci
matteojug: Yes: in `make integration-env-up-ci` we run the script directly with `python3`, while for devenv we build `emily-aws-setup` as a target in `docker/sbtc/Dockerfile`

docker/tests/devnet-liveness.sh
matteojug: I'm not sure: we are not using this anywhere, and I'm not aware of anyone using it. This removal was more a _does anybody use this?_ type of thing (as the composes ended up in this same directory). Happy to restore it if we still have a use for it.

signer/tests/integration/containers.rs
matteojug: While I agree that the compile time eval would probably work at the moment, it shouldn't be always the case given how we run the integration tests in CI (with archived tests); eg, if the GH runner changes the home path for different runs, when we copy the archived tests over we may refer to the wrong assets location (mentioned in https://github.com/stacks-sbtc/sbtc/pull/1858#discussion_r2518377120).

signer/tests/integration/postgres.rs
matteojug: If a test concludes "naturally" (succeeding or failing, even with panics) the stack is downed. If you cancel a test (eg,  SIGINT) then the stack is not downed and will be still up after the run finishes.

About the database: the compose client thingy we use by default uses `--volumes` when `docker compose down`, so even if we were to use volumes for the db it should disappear when the stack is downed; but given we are not even using volumes for the test stack, it should just disappear once the stack is removed.

docker/tests/devnet-liveness.sh
djordon: Na, if it's unused let's delete it. And if we want it back some time in the future we can easily restore it.

PR #1866, opened 2025-11-24
feat: use `raw_value` field for stacks core events

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1865

## Changes

* Update the `SmartContractEvent` struct to deserialize from the `raw_value` field instead of the value field.

## Testing Information

I tested this out on devenv and everything worked just fine.

## Checklist


signer/tests/fixtures/rotate-keys-and-invalid-event.json
```diff
@@ -25,7 +25,7 @@
             "committed": true,
             "contract_event": {
                 "contract_identifier": "SN3R84XZYA63QS28932XQF3G1J8R9PC3W76P9CSQS.sbtc-registry",
-                "raw_value": "0x",
+                "raw_value": "0x010000000000000000000000000000002a",
```
djordon, line 28: The is a hex encoded consensus serialized `clarity::vm::Value::Uint(42)`. If it doesn't deserialize at all then we have problems, and the previous value doesn't deserialize.

PR #1869, opened 2025-11-28
feat: prepare emily for parallel testing

matteojug:
## Description

Work for https://github.com/stacks-sbtc/sbtc/issues/1854

## Changes

This PR uses Emily dynamic context to prepare for Emily parallelized integration tests. After the compose changes (more on this later):
```
$ cargo nextest run --package emily-handler --test integration --test-threads 1
[...]
     Summary [  13.732s] 107 tests run: 107 passed, 0 skipped

cargo nextest run --package emily-handler --test integration
[...]
     Summary [   4.330s] 107 tests run: 107 passed, 0 skipped
```
Note that CI (and the various make commands) still run tests sequentially for now, this is just scaffolding.

Emily integration tests now create a new set of tables for each test (on the same dynamodb instance). The penalty for doing this while still running tests in a serial fashion shouldn't be more than some tens of ms per test.

There are also some changes in the testing compose:
 - Removed `--dbname=signer` from the healthcheck to avoid the "no signer db" spam in the logs
 - Moved to a `inMemory` storage for dynamodb (in integration tests): this reduces the table creation from ~100ms to ~10ms (and probably also improve running times for other operations, e.g. before that change locally a serial run of Emily tests took ~17s, vs the ~14 after).

## Testing Information

Add a test for the emily setup, and tested parallel run locally.

## Checklist

### DIFF
--- a/sbtc/src/testing/emily.rs
+++ b/sbtc/src/testing/emily.rs
@@ -127,19 +127,14 @@ async fn create_table(client: &Client, table_name: String, props: &serde_json::V
         let name = attr["AttributeName"].as_str().unwrap();
         let type_str = attr["AttributeType"].as_str().unwrap();
 
-        let scalar_type = match type_str {
-            "S" => ScalarAttributeType::S,
-            "N" => ScalarAttributeType::N,
-            "B" => ScalarAttributeType::B,
-            _ => panic!("unknown attribute type: {type_str}"),
-        };
+        let scalar_type = ScalarAttributeType::try_parse(type_str).expect("unknown attribute type");
 
         attributes.push(
             AttributeDefinition::builder()
                 .attribute_name(name)
                 .attribute_type(scalar_type)
                 .build()
-                .unwrap(),
+                .expect("failed to build attribute definition"),
         );
     }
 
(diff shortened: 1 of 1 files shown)


sbtc/src/testing/emily.rs
```diff
@@ -0,0 +1,258 @@
...
+        "xxxxxxxxxxxx",
+        "xxxxxxxxxxxx",
+        None,
+        None,
+        "test",
+    );
+
+    // We don't need https in tests (and we avoid wasting time in loading certs)
+    let tcp_connector = hyper_014::client::HttpConnector::new();
+    let http_client = HyperClientBuilder::new().build(tcp_connector);
```
matteojug, line 233: Using the default client takes ~150ms, mostly to load the certificates (that we don't use for local tests), and since nextest uses a process for each test we cannot just rely on cached clients :(

sbtc/src/testing/emily.rs
```diff
@@ -0,0 +1,258 @@
...
+    for attr in attrs {
+        let name = attr["AttributeName"].as_str().unwrap();
+        let type_str = attr["AttributeType"].as_str().unwrap();
+
+        let scalar_type = match type_str {
+            "S" => ScalarAttributeType::S,
+            "N" => ScalarAttributeType::N,
+            "B" => ScalarAttributeType::B,
+            _ => panic!("unknown attribute type: {type_str}"),
+        };
```
djordon: The `ScalarAttributeType` implements `From<&str>`. There is also a `ScalarAttributeType::try_parse` method for doing this, which is even better. Same thing for the `ProjectionType` and the `KeyType`.

djordon (commented):
Largely looks good, just had one comment on how to simplify some of the conversions from strings into rust types.

Issue #1860, opened 2025-11-13
[Feature]: Expose more withdrawal information in Emily


djordon:
# Feature - Expose more withdrawal information in Emily

## 1. Description

There are a few pieces of information that builders would like to be able to surface to their users related to withdrawals of sBTC to BTC. The purpose of this information is to build a more pleasant user experience, since there isn’t much information for users exposed in our API for withdrawals.

The two pieces of information that would be most useful are:

1. The number of blocks remaining before the signers will attempt to sweep out the funds.
2. The transaction ID of the sweep transaction fulfilling the withdrawal request.

This information could be exposed in a new endpoint, or in some of our existing endpoints. We just need to be careful about communicating the meaning of these two pieces of information, since they are educated guesses. The transaction ID confirming the withdrawal request can materialize and then change or entirely disappear.

One additional piece of information not requested, but probably useful is whether the withdrawal request’s `max_fee` is sufficiently high given the current prevailing fee rate. Emily could figure this out on her own and surface it to the caller.

## 2. Technical Details

For the number of blocks remaining, this would be pretty straightforward. Emily would need to store the bitcoin block height that anchors the stacks block that confirms the transaction that created the withdrawal request. We would then be able to provide an estimate of how many blocks are remaining given the current chain tip block height and the anchor height.

The transaction ID part is a little trickier. One approach would be for Emily’s sidecar to fetch this information from the mempool and update Emily with whatever data it finds. The sidecar flow would go like this:

1. Emily’s sidecar fetches all accepted withdrawal requests from Emily.
2. If there are any accepted withdrawals, the sidecar looks for all sweep transactions in the mempool. The sidecar will likely need to reach out to stacks-core to get the current aggregate key and use that to derive the address string.
3. The sidecar updates the withdrawals in Emily that are being fulfilled by the latest sweep.

### 2.1 Acceptance Criteria

- [ ]  Emily exposes how many more bitcoin blocks are likely needed before the sBTC signers will consider the withdrawal request.
- [ ]  Emily exposes the transaction ID for any withdrawal requests being fulfilled by the signers that are in the mempool.

## 4. Addendum

We could have the signers send the transaction ID to Emily when the coordinator notifies Emily that the request has been accepted. It’s seems easier to have the signer update Emily during the existing update flow. The biggest downside here is that this information would be unverified. A sole signer could present this information to Emily, and she wouldn’t know whether it was true or not.

[...]

PR #1870, opened 2025-12-01
feat: expose withdrawal information on Emily

MCJOHN974:
## Description

Closes: #1860 

## Changes

Added expected fulfillment info to withdrawals, showing estimate for bitcoin height and txid. Height is calculated by emily while txid is provided by signers (any signer can overwrite txid with any other string, though).

## Testing Information

Added integration test
## Checklist

### DIFF
--- a/.generated-sources/emily/client/rust/private/docs/ExpectedFulfillmentInfo.md
+++ b/.generated-sources/emily/client/rust/private/docs/ExpectedFulfillmentInfo.md
@@ -4,8 +4,8 @@
 
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
-**bitcoin_block_height** | Option<**u64**> | Expected bitcoin block height into which sweep transaction will be included into the block. Calculated by Emily once, no ajustments are made if something goes wrong. | [optional]
-**bitcoin_txid** | Option<**String**> | Expected txid of the sweep transaction. This field is populated once, and will not be changed if there is and rbf (and will show incorrect expectation) | [optional]
+**bitcoin_block_height** | Option<**u64**> | The estimated bitcoin block height for the bitcoin block confirming the transaction fulling the withdrawal request. This value is estimated by Emily once when the withdrawal request is initially received. | [optional]
+**bitcoin_txid** | Option<**String**> | The expected txid of the sweep transaction fulfilling the withdrawal request. This field is populated once, it is not updated if there is an RBF transaction that also fulfills the request. | [optional]
 
 [[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
 
--- a/.generated-sources/emily/client/rust/private/src/models/expected_fulfillment_info.rs
+++ b/.generated-sources/emily/client/rust/private/src/models/expected_fulfillment_info.rs
@@ -14,7 +14,7 @@ use serde::{Deserialize, Serialize};
 /// ExpectedFulfillmentInfo : Expected information about future fulfillment of the withdrawal. This struct represents _estimate_ of how fulfillment for the withdrawal will look like. This estimation is for reference only, and does not provide any guarantees.
 #[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
 pub struct ExpectedFulfillmentInfo {
-    /// Expected bitcoin block height into which sweep transaction will be included into the block. Calculated by Emily once, no ajustments are made if something goes wrong.
+    /// The estimated bitcoin block height for the bitcoin block confirming the transaction fulling the withdrawal request. This value is estimated by Emily once when the withdrawal request is initially received.
     #[serde(
         rename = "bitcoinBlockHeight",
         default,
@@ -22,7 +22,7 @@ pub struct ExpectedFulfillmentInfo {
         skip_serializing_if = "Option::is_none"
     )]
     pub bitcoin_block_height: Option<Option<u64>>,
-    /// Expected txid of the sweep transaction. This field is populated once, and will not be changed if there is and rbf (and will show incorrect expectation)
+    /// The expected txid of the sweep transaction fulfilling the withdrawal request. This field is populated once, it is not updated if there is an RBF transaction that also fulfills the request.
     #[serde(
         rename = "bitcoinTxid",
         default,
--- a/.generated-sources/emily/client/rust/public/docs/ExpectedFulfillmentInfo.md
+++ b/.generated-sources/emily/client/rust/public/docs/ExpectedFulfillmentInfo.md
@@ -4,8 +4,8 @@
 
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
-**bitcoin_block_height** | Option<**u64**> | Expected bitcoin block height into which sweep transaction will be included into the block. Calculated by Emily once, no ajustments are made if something goes wrong. | [optional]
-**bitcoin_txid** | Option<**String**> | Expected txid of the sweep transaction. This field is populated once, and will not be changed if there is and rbf (and will show incorrect expectation) | [optional]
+**bitcoin_block_height** | Option<**u64**> | The estimated bitcoin block height for the bitcoin block confirming the transaction fulling the withdrawal request. This value is estimated by Emily once when the withdrawal request is initially received. | [optional]
+**bitcoin_txid** | Option<**String**> | The expected txid of the sweep transaction fulfilling the withdrawal request. This field is populated once, it is not updated if there is an RBF transaction that also fulfills the request. | [optional]
 
 [[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
 
--- a/.generated-sources/emily/client/rust/public/src/models/expected_fulfillment_info.rs
+++ b/.generated-sources/emily/client/rust/public/src/models/expected_fulfillment_info.rs
@@ -14,7 +14,7 @@ use serde::{Deserialize, Serialize};
 /// ExpectedFulfillmentInfo : Expected information about future fulfillment of the withdrawal. This struct represents _estimate_ of how fulfillment for the withdrawal will look like. This estimation is for reference only, and does not provide any guarantees.
 #[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
 pub struct ExpectedFulfillmentInfo {
-    /// Expected bitcoin block height into which sweep transaction will be included into the block. Calculated by Emily once, no ajustments are made if something goes wrong.
+    /// The estimated bitcoin block height for the bitcoin block confirming the transaction fulling the withdrawal request. This value is estimated by Emily once when the withdrawal request is initially received.
     #[serde(
         rename = "bitcoinBlockHeight",
         default,
@@ -22,7 +22,7 @@ pub struct ExpectedFulfillmentInfo {
         skip_serializing_if = "Option::is_none"
     )]
     pub bitcoin_block_height: Option<Option<u64>>,
-    /// Expected txid of the sweep transaction. This field is populated once, and will not be changed if there is and rbf (and will show incorrect expectation)
+    /// The expected txid of the sweep transaction fulfilling the withdrawal request. This field is populated once, it is not updated if there is an RBF transaction that also fulfills the request.
     #[serde(
         rename = "bitcoinTxid",
         default,
--- a/.generated-sources/emily/client/rust/testing/docs/ExpectedFulfillmentInfo.md
+++ b/.generated-sources/emily/client/rust/testing/docs/ExpectedFulfillmentInfo.md
@@ -4,8 +4,8 @@
 
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
-**bitcoin_block_height** | Option<**u64**> | Expected bitcoin block height into which sweep transaction will be included into the block. Calculated by Emily once, no ajustments are made if something goes wrong. | [optional]
-**bitcoin_txid** | Option<**String**> | Expected txid of the sweep transaction. This field is populated once, and will not be changed if there is and rbf (and will show incorrect expectation) | [optional]
+**bitcoin_block_height** | Option<**u64**> | The estimated bitcoin block height for the bitcoin block confirming the transaction fulling the withdrawal request. This value is estimated by Emily once when the withdrawal request is initially received. | [optional]
+**bitcoin_txid** | Option<**String**> | The expected txid of the sweep transaction fulfilling the withdrawal request. This field is populated once, it is not updated if there is an RBF transaction that also fulfills the request. | [optional]
 
 [[Back to Model list]](../README.md#documentation-for-models) [[Back to API list]](../README.md#documentation-for-api-endpoints) [[Back to README]](../README.md)
 
--- a/.generated-sources/emily/client/rust/testing/src/models/expected_fulfillment_info.rs
+++ b/.generated-sources/emily/client/rust/testing/src/models/expected_fulfillment_info.rs
@@ -14,7 +14,7 @@ use serde::{Deserialize, Serialize};
 /// ExpectedFulfillmentInfo : Expected information about future fulfillment of the withdrawal. This struct represents _estimate_ of how fulfillment for the withdrawal will look like. This estimation is for reference only, and does not provide any guarantees.
 #[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
 pub struct ExpectedFulfillmentInfo {
-    /// Expected bitcoin block height into which sweep transaction will be included into the block. Calculated by Emily once, no ajustments are made if something goes wrong.
+    /// The estimated bitcoin block height for the bitcoin block confirming the transaction fulling the withdrawal request. This value is estimated by Emily once when the withdrawal request is initially received.
     #[serde(
         rename = "bitcoinBlockHeight",
         default,
@@ -22,7 +22,7 @@ pub struct ExpectedFulfillmentInfo {
         skip_serializing_if = "Option::is_none"
     )]
     pub bitcoin_block_height: Option<Option<u64>>,
-    /// Expected txid of the sweep transaction. This field is populated once, and will not be changed if there is and rbf (and will show incorrect expectation)
+    /// The expected txid of the sweep transaction fulfilling the withdrawal request. This field is populated once, it is not updated if there is an RBF transaction that also fulfills the request.
     #[serde(
         rename = "bitcoinTxid",
         default,


signer/src/transaction_coordinator.rs
```diff
@@ -698,6 +698,8 @@ where
             self.sign_and_broadcast(bitcoin_chain_tip.as_ref(), &mut transaction)
                 .await?;
 
+            let expected_height = bitcoin_chain_tip.block_height + WITHDRAWAL_MIN_CONFIRMATIONS;
```
djordon: Hmmm, the expected height should be the bitcoin block height where the withdrawal request is expected to be fulfilled. So it is something that can computed by Emily when the withdrawal request is created, we just need to store the bitcoin block height of the bitcoin block that anchors the stacks block that confirms the withdrawal request. The signers do not need to send this information, they only need to _maybe_ send the transaction ID, and I'm conflicted as to whether the signers should send that at all.

emily/handler/src/api/models/withdrawal/mod.rs
```diff
@@ -56,10 +56,39 @@ pub struct Withdrawal {
...
+)]
+#[serde(rename_all = "camelCase")]
+pub struct PreFulfillment {
+    /// Expected bitcoin block height on which signers will _try_ to fullfill
+    /// withdrawal request.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub maybe_expected_height: Option<u64>,
+    /// Expected txid of the transaction fullfilling withdrawal request
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub maybe_expected_txid: Option<String>,
```
djordon: nit: I have a preference for these field names to follow the same pattern as other fields, and we haven't been prefixing our other optional field names with `maybe_`.

emily/handler/src/api/models/withdrawal/mod.rs
```diff
@@ -56,10 +56,39 @@ pub struct Withdrawal {
...
+    PartialOrd,
+    Ord,
+    Hash,
+    Serialize,
+    Deserialize,
+    ToSchema,
+    ToResponse,
+)]
+#[serde(rename_all = "camelCase")]
+pub struct PreFulfillment {
```
djordon: Can we chose a different name from pre-fulfillment. Maybe `ExpectedFulfillmentInfo`? I don't love that name either, but I think I like it more than `PreFulfillment`.

signer/src/transaction_coordinator.rs
MCJOHN974: We spoke about this online, but what was the consensus about tx_id? Should it be sent from the signer or not? @matteojug @cylewitruk maybe you have thoughts on this?

MCJOHN974: [58ba230](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/58ba2301bbef94ba700733105c1e8993c35f9418)

emily/handler/src/api/models/withdrawal/mod.rs
MCJOHN974: [055a030](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/055a030b8ec3708e8b8aa1e7b59f2de5a28dc101)

emily/handler/src/api/models/withdrawal/mod.rs
MCJOHN974: [055a030](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/055a030b8ec3708e8b8aa1e7b59f2de5a28dc101)

emily_sidecar/test/test_integration.py
```diff
@@ -72,6 +72,7 @@ class IntegrationTests(unittest.TestCase):
     def setUp(self):
         # Wipe the state before each test
         requests.post(WIPE_URL, headers=headers)
+        self.maxDiff = None
```
matteojug, line 76: nit: is this still used?

signer/src/transaction_coordinator.rs
matteojug: mmm shouldn't `expected_height` be `bitcoin_chain_tip.block_height + 1`? We just submitted the sweep tx, so optimistically it will be included in the next bitcoin block

emily/handler/src/api/handlers/new_block.rs
```diff
@@ -305,6 +306,10 @@ fn handle_withdrawal_accept(event: WithdrawalAcceptEvent) -> WithdrawalUpdate {
             btc_fee: event.fee,
             stacks_txid: hex::encode(event.txid.0),
         }),
+        expected_fulfillment_info: ExpectedFulfillmentInfo {
+            expected_height: None,
+            expected_txid: None,
```
matteojug: nit: why not `sweep_block_height` and `event.outpoint.txid`?

emily/handler/src/api/handlers/withdrawal.rs
```diff
@@ -263,6 +263,15 @@ pub async fn create_withdrawal(
...
         let status = WithdrawalStatus::Pending;
 
+        let chainstate =
+            accessors::get_chainstate_entry_at_height(&context, &stacks_block_height).await;
+
+        let expected_height = chainstate
+            .map(|chainstate| chainstate.bitcoin_height)
+            .ok()
+            .flatten()
+            .map(|height| height + sbtc::WITHDRAWAL_MIN_CONFIRMATIONS);
```
matteojug: This is correct if `expected_height` is the height at which the signers will try to do something (as per docstring), but maybe it would be less confusing for it to be the height at which ideally the withdrawal will be fulfilled (as in, the height of the sweep tx)?

If so, we need a `+1` here (and an updated docstring).

emily/handler/src/database/entries/withdrawal.rs
```diff
@@ -686,11 +694,26 @@ impl WithdrawalUpdatePackage {
...
+        let expected_height = update_expected_fulfillment
+            .expected_height
+            .or(latest_event_expected_fulfillment.expected_height);
+        let expected_txid = update_expected_fulfillment
+            .expected_txid
+            .clone()
+            .or(latest_event_expected_fulfillment.expected_txid.clone());
+        let mut new_event = update.event;
+        new_event.expected_fulfillment_info =
+            ExpectedFulfillmentInfo { expected_height, expected_txid };
```
matteojug: With this behaviour we would still have a `expected_fulfillment_info` even after a withdrawal is rejected (in `handle_withdrawal_reject`). Do we want it?

emily/handler/src/api/handlers/new_block.rs
MCJOHN974-STACKS: Right now we fill information about height on withdrawal creation and about txid on signer's update. We can add information update here but for simplicity we also can keep less moving parts.

emily/handler/src/database/entries/withdrawal.rs
MCJOHN974-STACKS: I think why not (aka "if withdrawal would not be rejected it would have the following txid and height: {}", sounds like a valid info).

I have no idea who might need this info, but I don't think it is invalid, and I'd make it up to final frontend to show that withdrawal was rejected and don't show this.

signer/src/transaction_coordinator.rs
MCJOHN974-STACKS: I think it have pros and cons, like "expected block in which tx will be included" sounds more natural for users to see, however "minimal block on which signers will take an action on this withdrawal" is more precise information -- transaction can be not included in expected block for a number of valid reasons, while signers will try to do their job at described block almost always.

Probably it is a good q to speak on call about.

MCJOHN974-STACKS: We disccussed that we want to do +1 .

emily/handler/src/api/models/withdrawal/mod.rs
```diff
@@ -56,10 +56,39 @@ pub struct Withdrawal {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     pub fulfillment: Option<Fulfillment>,
+    /// Details about withdrawal which was approved and is processed. Note that this details are not final, and may change,
+    /// as well as any value of this parameter does not guarantee that the withdrawal will be processed at all.
+    pub expected_fulfillment_info: ExpectedFulfillmentInfo,
     /// The hex encoded txid of the stacks transaction that generated this event.
     pub txid: String,
 }
 
+/// Withdrawal parameters.
```
djordon: We should update this docstring to something like:
```rust
/// A struct containing expected information about the fulfillment of a
/// withdrawal request.
/// 
/// # Notes
///
/// The information in this struct is expected to be correct and accurate,
/// but it is not guaranteed to be. The `expected_height` is computed
/// locally in Emily, but can end up being wrong since the withdrawal
/// request could be fulfilled much later, or not at all. The
/// `expected_txid` is taken from a signer when they update the status of
/// withdrawal requests. The signer that does this is usually the
/// coordinator, but it could be any signer and it could happen at any
/// time. Moreover, it could be wrong, and the withdrawal request could end
/// up being unfulfilled.
```
Maybe we should tone down this a little though?

emily/handler/src/api/handlers/withdrawal.rs
MCJOHN974-STACKS: Added +1: [d274df6](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/d274df6e25547c1b0b211e0ad0ac8d2249e926cb)

emily_sidecar/test/test_integration.py
MCJOHN974-STACKS: This parameters limits diffs the failed test will show, where `None` is "No limit". Aka, if we do `assert a == b`, if diff between a and b is more then value of this parameter diff will not be shown in full. Imo it is more convinient to see no limit.

signer/src/transaction_coordinator.rs
```diff
@@ -698,6 +698,8 @@ where
             self.sign_and_broadcast(bitcoin_chain_tip.as_ref(), &mut transaction)
                 .await?;
 
+            let expected_height = bitcoin_chain_tip.block_height + WITHDRAWAL_MIN_CONFIRMATIONS + 1;
```
djordon: This should be this now, given the new meaning.
```suggestion
            let expected_height = bitcoin_chain_tip.block_height + 1;
```
You could also have the signers not send this information at all since Emily already computes it.

emily/handler/src/api/models/withdrawal/requests.rs
```diff
@@ -63,6 +63,8 @@ pub struct WithdrawalUpdate {
     /// Details about the on chain artifacts that fulfilled the withdrawal.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub fulfillment: Option<Fulfillment>,
+    /// Details of the process of fulfilling the withdrawal
+    pub expected_fulfillment_info: ExpectedFulfillmentInfo,
```
djordon, line 67: Are we sure that we want the signer to update the expected confirmation height?

emily/handler/src/api/models/withdrawal/mod.rs
djordon: This is an old comment before the updates. I forgot to delete it before submitting my review.

emily_sidecar/test/test_integration.py
matteojug: Ah, I see. Can you add a short comment about it? eg "Show all diff on fail" or similar, to make it clear it's not something we use in the test but a config

emily/handler/src/api/handlers/new_block.rs
matteojug: Yeah I guess given the meaning of `expected_fulfillment_info` it's fine not to change it, you can check the actual fulfilment at that point.

emily/handler/src/database/entries/withdrawal.rs
matteojug: I still find it a bit strange but I'm not entirely sure doing something different would actual be better, happy to keep the current behaviour

emily/handler/src/api/models/withdrawal/requests.rs
MCJOHN974: Emily ignores this update

signer/src/transaction_coordinator.rs
MCJOHN974: [35e67cf](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/35e67cfdaddc62983a2d8424e6e9ac0275f36998)

emily_sidecar/test/test_integration.py
MCJOHN974: [40435ee](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/40435ee9d7bf3b1213f8a925befc4dc1168ed136)

emily/handler/src/api/handlers/withdrawal.rs
matteojug: mmm maybe it got lost after other changes?

emily/handler/src/api/models/withdrawal/requests.rs
matteojug: Looking at the code it seems nothing prevents a signer from sending the first update (going from pending to accepted) with some bogus expected height, overwriting the one computed by emily in the beginning. Emily does reject further updates (going from accepted to accepted) from the signers, but the first one seems still susceptible to this. Am I missing something?

emily/handler/src/api/handlers/withdrawal.rs
MCJOHN974: 924b38fd9df30017c4d23c0b1db5b899285104b3

emily/handler/src/api/models/withdrawal/requests.rs
MCJOHN974: [228a4c4](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/228a4c42f5c76752e8eb1b01e945ae652c1967b9) prevents signers from overwriting it. However I don't think overwriting by signer is a big problem

emily/handler/src/database/entries/withdrawal.rs
```diff
@@ -686,11 +694,24 @@ impl WithdrawalUpdatePackage {
         entry
             .latest_event()?
             .ensure_following_event_is_valid(&update.event)?;
+
+        // keep old data for expected_fulfillment details if None was provided.
```
djordon: Wait why do we want to keep the old transaction ID here?

MCJOHN974: If signer A told us that he expects the txid to be X, and then, later, signer B tells us that he don't know what txid it would be, I'd prefer to show users signer A opinion. (however I can imagine it be possible only if B didn't update the binary yet)

djordon: I just had a misunderstanding, we will actually take signer B's txid with the logic here. However, we will reject the second update anyway because we ignore `Accepted` to `Accepted` status changes.

matteojug: nit: can we update this comment with the new logic? ie, we also ignore updates to `expected_height` for now

emily/handler/tests/integration/withdrawal.rs
```diff
@@ -1310,3 +1363,117 @@ async fn only_confirmed_withdrawals_can_have_fulfillment(
...
+        stacks_block_height: chainstate.stacks_block_height,
+        txid: "test_txid".to_string(),
+    };
+
+    // Create the withdrawal with the privileged configuration.
+    apis::withdrawal_api::create_withdrawal(&testing_configuration, request.clone())
+        .await
+        .expect("Received an error after making a valid create withdrawal request api call.");
+
+    // Right after creation txid should be none and height should be chainstate.bitcoin_block_height + 6
```
matteojug: nit
```suggestion
    // Right after creation txid should be none and height should be chainstate.bitcoin_block_height + 6 + 1
```

emily/handler/tests/integration/withdrawal.rs
```diff
@@ -1310,3 +1363,117 @@ async fn only_confirmed_withdrawals_can_have_fulfillment(
...
+        .expected_height
+        .unwrap()
+        .unwrap();
+    assert!(withdrawal.expected_fulfillment_info.expected_txid.is_none());
+    assert_eq!(8, received_expected_height);
+
+    // Now signer updates withdrawal with expected txid
+    let expected_expected_txid = "txid".to_string();
+    let expected_fulfillment_info = ExpectedFulfillmentInfo {
+        expected_height: None,
```
matteojug: nit: can we also send a non empty `expected_height` to test it's ignored during the update?

MCJOHN974: [3099c73](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/3099c73cfed21c896754d3b8094dc7e532213f56)

emily/handler/src/database/entries/withdrawal.rs
MCJOHN974: [da09d00](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/da09d005af0a11ac1d19e7a60bdab7914c95e110)

emily/handler/src/api/models/withdrawal/mod.rs
```diff
@@ -56,10 +56,40 @@ pub struct Withdrawal {
...
+    Serialize,
+    Deserialize,
+    ToSchema,
+    ToResponse,
+)]
+#[serde(rename_all = "camelCase")]
+pub struct ExpectedFulfillmentInfo {
+    /// Expected bitcoin block height into which sweep transaction will be included into the block.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expected_height: Option<u64>,
```
djordon: We all spoke about this offline, and we should change the name of this field to `bitcoin_block_height`.

emily/handler/src/api/models/withdrawal/mod.rs
```diff
@@ -56,10 +56,40 @@ pub struct Withdrawal {
...
+    /// Expected txid of the sweep transaction.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expected_txid: Option<String>,
```
djordon: We also spoke about this one and we agreed to change this name to `bitcoin_txid`.

emily/handler/src/api/models/withdrawal/mod.rs
djordon: Can we update the description here to mention that this information is set once when the request is received and not updated later.

emily/handler/src/api/models/withdrawal/mod.rs
```diff
@@ -56,10 +56,40 @@ pub struct Withdrawal {
...
+    Deserialize,
+    ToSchema,
+    ToResponse,
+)]
+#[serde(rename_all = "camelCase")]
+pub struct ExpectedFulfillmentInfo {
+    /// Expected bitcoin block height into which sweep transaction will be included into the block.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expected_height: Option<u64>,
+    /// Expected txid of the sweep transaction.
```
djordon: We should also mention that this information is set when the signers broadcast the first transaction fulfilling the withdrawal request and not updated if there is an RBF.

MCJOHN974: [7b8754e](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/7b8754e9bed55fabf2e8e2a10381a6c24aed7b7e)

emily/handler/src/api/models/withdrawal/mod.rs
MCJOHN974: [7b8754e](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/7b8754e9bed55fabf2e8e2a10381a6c24aed7b7e)

emily/handler/src/api/models/withdrawal/mod.rs
MCJOHN974: [7b8754e](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/7b8754e9bed55fabf2e8e2a10381a6c24aed7b7e)

emily/handler/src/api/models/withdrawal/mod.rs
MCJOHN974: [7b8754e](https://github.com/stacks-sbtc/sbtc/pull/1870/commits/7b8754e9bed55fabf2e8e2a10381a6c24aed7b7e)

emily/handler/src/api/models/withdrawal/mod.rs
```diff
@@ -83,11 +83,13 @@ pub struct Withdrawal {
 #[serde(rename_all = "camelCase")]
 pub struct ExpectedFulfillmentInfo {
     /// Expected bitcoin block height into which sweep transaction will be included into the block.
+    /// Calculated by Emily once, no ajustments are made if something goes wrong.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub expected_height: Option<u64>,
-    /// Expected txid of the sweep transaction.
+    pub bitcoin_block_height: Option<u64>,
+    /// Expected txid of the sweep transaction. This field is populated once, and will not be changed if there
+    /// is and rbf (and will show incorrect expectation)
```
matteojug: nit: `is an RBF`

emily/handler/src/api/models/withdrawal/mod.rs
```diff
@@ -56,10 +56,42 @@ pub struct Withdrawal {
...
+    Hash,
+    Serialize,
+    Deserialize,
+    ToSchema,
+    ToResponse,
+)]
+#[serde(rename_all = "camelCase")]
+pub struct ExpectedFulfillmentInfo {
+    /// Expected bitcoin block height into which sweep transaction will be included into the block.
+    /// Calculated by Emily once, no ajustments are made if something goes wrong.
```
djordon: I'd go with something like this:
```suggestion
    /// The estimated bitcoin block height for the bitcoin block confirming
    /// the transaction fulling the withdrawal request. This value is
    /// estimated by Emily once when the withdrawal request is initially
    /// received.
```

emily/handler/src/api/models/withdrawal/mod.rs
```diff
@@ -56,10 +56,42 @@ pub struct Withdrawal {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub bitcoin_block_height: Option<u64>,
+    /// Expected txid of the sweep transaction. This field is populated once, and will not be changed if there
+    /// is and rbf (and will show incorrect expectation)
```
djordon:
```suggestion
    /// The expected txid of the sweep transaction fulfilling the withdrawal
    /// request. This field is populated once, it is not updated if there is
    /// an RBF transaction that also fulfills the request.
```

emily/handler/src/database/entries/withdrawal.rs
```diff
@@ -686,11 +694,28 @@ impl WithdrawalUpdatePackage {
         entry
             .latest_event()?
             .ensure_following_event_is_valid(&update.event)?;
+
+        // Keeping the old data for expected_fulfillment txid if None was provided.
+        // For the bitcoin_block_height we intentionally ignore updates for now, relying on
+        // the value computed by Emily.
```
djordon, line 700: Just want to make sure that I understand what happens if we just took the expected fulfillment info as it is. If we did that then the "expected" bitcoin block height would not be returned from the any of the withdrawals endpoints following an update by the signers, correct?

MCJOHN974: I'm not sure I understand your question.

djordon (approved):
Looks good

emily/handler/src/database/entries/withdrawal.rs
djordon: We spoke about this and my understanding is correct 😄.

matteojug (approved):

PR #1872, opened 2025-12-02
feat: use more local types for deserializing stacks core webhook events

djordon:
## Description

This is part 2 or 3 of https://github.com/stacks-sbtc/sbtc/issues/1835, dealing with stacks-core types in the sbtc crate

## Changes

* Remove the `transactions` field from the `NewBlockEvent` struct, delete the types used in the struct associated with that field.
* Remove some of the unnecessary types from the `TransactionEvent`, and remove use of the stacks-core trait.
* Remove the stackslib dependency from the sbtc crate

## Checklist

### DIFF
--- a/sbtc/src/error.rs
+++ b/sbtc/src/error.rs
@@ -73,10 +73,11 @@ pub enum Error {
     #[cfg(any(test, feature = "webhooks"))]
     #[error("could not decode the hex string into bytes: {0}")]
     DecodeHexBytes(#[source] hex::FromHexError),
-    /// This is thrown when failing to parse a hex string into bytes.
+    /// This is thrown when failing to parse a hex string into bytes when
+    /// deserializing a string into a Stacks type.
     #[cfg(any(test, feature = "webhooks"))]
-    #[error("could not parse hex txid: {0}")]
-    DecodeHexTxid(#[source] bitcoin::hex::HexToArrayError),
+    #[error("could not parse hex to Stacks type {1}: {0}")]
+    DecodeHexArray(#[source] bitcoin::hex::HexToArrayError, &'static str),
     /// An error when attempting to generically decode bytes using the
     /// trait implementation.
     #[cfg(any(test, feature = "webhooks"))]
--- a/sbtc/src/webhooks.rs
+++ b/sbtc/src/webhooks.rs
@@ -42,7 +42,7 @@ impl DeserializeHex for StacksTxid {
     fn try_from_hex(hex: &str) -> Result<Self, Error> {
         <[u8; 32]>::from_hex(hex)
             .map(Self)
-            .map_err(Error::DecodeHexTxid)
+            .map_err(|error| Error::DecodeHexArray(error, "StacksTxid"))
     }
 }
 
@@ -50,7 +50,7 @@ impl DeserializeHex for BlockHeaderHash {
     fn try_from_hex(hex: &str) -> Result<Self, Error> {
         <[u8; 32]>::from_hex(hex)
             .map(Self)
-            .map_err(Error::DecodeHexTxid)
+            .map_err(|error| Error::DecodeHexArray(error, "BlockHeaderHash"))
     }
 }
 
@@ -58,7 +58,7 @@ impl DeserializeHex for BurnchainHeaderHash {
     fn try_from_hex(hex: &str) -> Result<Self, Error> {
         <[u8; 32]>::from_hex(hex)
             .map(Self)
-            .map_err(Error::DecodeHexTxid)
+            .map_err(|error| Error::DecodeHexArray(error, "BurnchainHeaderHash"))
     }
 }
 
@@ -66,7 +66,7 @@ impl DeserializeHex for StacksBlockId {
     fn try_from_hex(hex: &str) -> Result<Self, Error> {
         <[u8; 32]>::from_hex(hex)
             .map(Self)
-            .map_err(Error::DecodeHexTxid)
+            .map_err(|error| Error::DecodeHexArray(error, "StacksBlockId"))
     }
 }
 
@@ -221,7 +221,11 @@ where
 
 #[cfg(test)]
 mod tests {
+    use std::marker::PhantomData;
+
     use super::*;
+
+    use test_case::test_case;
 
     /// This was captured using a slightly modified version of the above
     /// function against a stacks-node running Nakamoto on commit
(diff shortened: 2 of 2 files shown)


sbtc/src/error.rs
```diff
@@ -70,12 +70,16 @@ pub enum Error {
     },
 
     /// This is thrown when failing to parse a hex string into bytes.
-    #[cfg(feature = "webhooks")]
+    #[cfg(any(test, feature = "webhooks"))]
     #[error("could not decode the hex string into bytes: {0}")]
     DecodeHexBytes(#[source] hex::FromHexError),
+    /// This is thrown when failing to parse a hex string into bytes.
```
MCJOHN974: Nit: I think it is not nice to have two identical comments for two different error variants

sbtc/src/webhooks.rs
```diff
@@ -20,18 +20,55 @@
...
+    }
+}
+
+impl DeserializeHex for StacksBlockId {
+    fn try_from_hex(hex: &str) -> Result<Self, Error> {
+        <[u8; 32]>::from_hex(hex)
+            .map(Self)
+            .map_err(Error::DecodeHexTxid)
+    }
+}
```
MCJOHN974, line 71: I know you hate macros, but maybe it is a case when even you find them more good than bad?
```suggestion
macro_rules! impl_deserialize_hex {
    ($type:ty) => {
        impl DeserializeHex for $type {
            fn try_from_hex(hex: &str) -> Result<Self, Error> {
                <[u8; 32]>::from_hex(hex)
                    .map(Self)
                    .map_err(Error::DecodeHexTxid)
            }
        }
    };
}

impl_deserialize_hex!(StacksTxid);
impl_deserialize_hex!(BlockHeaderHash);
impl_deserialize_hex!(BurnchainHeaderHash);
impl_deserialize_hex!(StacksBlockId);
```

MCJOHN974 (approved):
Also, I'd think it would be nice to add more tests for this but probably they won't add as much here

sbtc/Cargo.toml
```diff
@@ -25,9 +25,9 @@ thiserror.workspace = true
 bitcoincore-rpc = { workspace = true, optional = true }
 bitcoincore-rpc-json = { workspace = true, optional = true }
 hex = { workspace = true, optional = true }
-stackslib = { workspace = true, optional = true }
 
 [dev-dependencies]
+hex = { workspace = true }
```
matteojug: nit
```suggestion
hex.workspace = true
```

sbtc/src/webhooks.rs
```diff
@@ -20,18 +20,55 @@
...
+
+impl DeserializeHex for BlockHeaderHash {
+    fn try_from_hex(hex: &str) -> Result<Self, Error> {
+        <[u8; 32]>::from_hex(hex)
```
matteojug, line 51: `BlockHeaderHash`, `BurnchainHeaderHash` and `StacksBlockId` have a `from_hex` (via `impl_byte_array_newtype`), we could reuse those (or not, it's not going to look too different either way)

sbtc/src/webhooks.rs
```diff
@@ -83,56 +118,13 @@ pub struct NewBlockEvent {
...
-    NftBurnEvent,
-    /// A fungible-token transfer event
-    FtTransferEvent,
-    /// A fungible-token mint event
-    FtMintEvent,
-    /// A fungible-token burn event
-    FtBurnEvent,
+impl StacksTxid {
+    /// Create a StacksTxid from a hex string.
+    pub fn from_hex(hex: &str) -> Result<Self, Error> {
```
matteojug: nit: this is not used around, is it expected?

sbtc/src/webhooks.rs
djordon: I don't really hate them, I just think our codebase is better off without them. Easier for newcomers, easier for our IDEs, and often easier for us too. But yeah, this crossed my mind.

sbtc/src/error.rs
djordon: 👍🏽 good call.

sbtc/src/webhooks.rs
djordon: Oh oops, missed this on self-review. I used it at one point and forgot that I didn't anymore. Will remove.

sbtc/src/webhooks.rs
djordon: Oh nice, yeah I suppose we could use that. I looked through the source and going through the array seems slightly more efficient (although I'm not sure). No strong preferences here, happy to change if you do.

sbtc/src/webhooks.rs
djordon: Removed in https://github.com/stacks-sbtc/sbtc/pull/1872/commits/82db55cd45abeb082c226bd35b7d69c2c7ac0ebd

sbtc/Cargo.toml
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1872/commits/bd7da47bddf32070864010c32e528f6e1967380b

sbtc/src/error.rs
djordon: Updated the error in https://github.com/stacks-sbtc/sbtc/pull/1872/commits/aed42aa00f50c340e1d0f8db521cdf2a8401768b. I also added some tests

sbtc/src/webhooks.rs
matteojug: Nah, I'm fine with the current version

sbtc/src/webhooks.rs
```diff
@@ -349,14 +308,56 @@ mod tests {
...
+            Self::from_hex(hex)
+        }
+    }
+
+    /// Let's make sure that we decode hex the same way that stacks core
+    /// does.
+    #[test_case(PhantomData::<StacksBlockId>; "StacksBlockId")]
+    #[test_case(PhantomData::<BlockHeaderHash>; "BlockHeaderHash")]
+    #[test_case(PhantomData::<BurnchainHeaderHash>; "BurnchainHeaderHash")]
+    fn stacks_type_display_debug_impl<T>(_: PhantomData<T>)
```
MCJOHN974, line 344:
```suggestion
    fn type_hint<T>() -> T {
        unreachable!()
    }

    /// Let's make sure that we decode hex the same way that stacks core
    /// does.
    #[test_case(type_hint::<StacksBlockId>; "StacksBlockId")]
    #[test_case(type_hint::<BlockHeaderHash>; "BlockHeaderHash")]
    #[test_case(type_hint::<BurnchainHeaderHash>; "BurnchainHeaderHash")]
    fn stacks_type_display_debug_impl<T>(_: fn() -> T)
```

Just in case you would like to not use `PhantomData` something like this would work

MCJOHN974 (approved):

sbtc/src/webhooks.rs
MCJOHN974: In general, I don't like either this nor PhantomData, and maybe  just leaving a comment on why we need this at all is the best way here

MCJOHN974: Okay I think `PhantomData` is less ugly here

djordon: I need to use something like `PhantomData` in order to parameterize tests around _types_, instead of the usual case where we parameterize around the different _values_. I couldn't think of another way to do it.

MCJOHN974: Yeah, I see, basically what I suggested is more or less reinventing my own PhantomData, so I'd just keep it as is. Fyi you can also use some other testing macroses (something like generic test) but they also look not nice if used only in one test and it is +1 dep, So I'm ok with current impl

Issue #1845, opened 2025-10-31
[Feature]: Use the signer state when getting the stacks chain tip


djordon:
# Feature - Use the signer state when getting the stacks chain tip

## 1. Description

Querying the database for the stacks chain tip is one of our slower queries, taking ~2 seconds each time on mainnet. Although, it's not too bad, we do not need to spend the time on the query since we could just store the stacks node's response when we fetch the chain tip from the node.

## 2. Technical Details

Add the stacks chain tip block hash to the signer state and use that instead of querying the database.

### 2.1 Acceptance Criteria

- [ ] We do not query the database for the stacks chain tip

## 3. Related Issues and Pull Requests (optional)

https://github.com/stacks-sbtc/sbtc/issues/1770

PR #1885, opened 2026-01-09
feat: use the signer state when getting the stacks chain tip

djordon:
## Description

Closes: #1845 

## Changes

* Add the stacks chain tip to the signer state, use it in our event loops.
* Do not query for the stacks chain tip in our queries, expect it as an input. The changed queries are:
  * `get_pending_rejected_withdrawal_requests`
  * `get_pending_withdrawal_requests`
  * `get_swept_deposit_requests`
  * `get_swept_withdrawal_requests`
  * `key_rotation_exists`
* Update our tests, which took 95% of the effort.

## Testing Information

~~Still needs some devenv testing, hence the draft status, but it's probably good to go~~. I tested this on devenv and things seemed okay. There were some errors during the initial boot phase, right after nakamoto, but I couldn't figure out what the issue was. Everything resolved in a few blocks and deposits and withdrawals went through without any issues afterwards.

## Checklist

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -647,7 +647,7 @@ where
         aggregate_key: &PublicKey,
         signer_public_keys: &BTreeSet<PublicKey>,
     ) -> Result<(), Error> {
-        // Fetch the stacks chain tip from the database.
+        // Fetch the stacks chain tip from the signer state.
         let stacks_chain_tip = self
             .context
             .state()


signer/src/stacks/api.rs
djordon, line 1: For some reason, my IDE was complaining about the `ClarityName` type in this file, so I made this change to disambiguate things. Not sure how necessary they are.

MCJOHN974: Hmm, for me it is okish without these changes

signer/src/api/info.rs
```diff
@@ -434,6 +424,9 @@ mod tests {
             ..Faker.fake()
         };
         storage.write_stacks_block(&stacks_block).await.unwrap();
```
matteojug: nit: we don't need to store the stacks (and bitcoin) block anymore

signer/src/block_observer.rs
```diff
@@ -536,6 +550,9 @@ impl<C: Context, B> BlockObserver<C, B> {
         tracing::info!("updating the signer state with the current signer set");
         self.set_signer_set_info().await?;
 
+        tracing::info!("updating the signer state with the current stacks chain tip");
+        self.set_stacks_chain_tip(chain_tip).await?;
```
matteojug, line 554: qq: is there a specific reason to do this before and not after `set_bitcoin_chain_tip`?
I can see arguments for both cases: on the one hand, setting first the bitcoin and then stacks ensure we never have a stacks chain tip that is not anchored (from a db pov); otoh, setting the bitcoin last ensure we will not start acting on new chain tip messages until we also set the stacks chain tip.

signer/src/storage/model.rs
```diff
@@ -196,6 +196,26 @@ impl StacksBlock {
     }
 }
 
+/// A struct that references a specific stacks block is identifier and its
+/// position in the blockchain.
```
matteojug: nit?
```suggestion
/// A struct that references a specific stacks block ID and its
/// position in the blockchain.
```

signer/src/transaction_coordinator.rs
```diff
@@ -647,12 +647,11 @@ where
         aggregate_key: &PublicKey,
         signer_public_keys: &BTreeSet<PublicKey>,
     ) -> Result<(), Error> {
-        let storage = self.context.get_storage();
-
         // Fetch the stacks chain tip from the database.
```
matteojug: nit: no longer fetching from the db

signer/tests/integration/postgres.rs
```diff
@@ -5785,7 +5871,7 @@ async fn pending_rejected_withdrawal_already_accepted(setup_tables: SetupTables)
 
     // With a confirmed withdrawal output, we should no longer get the request
     let pending_rejected = db
-        .get_pending_rejected_withdrawal_requests(&bitcoin_chain_tip, 1000)
+        .get_pending_rejected_withdrawal_requests(&bitcoin_chain_tip, &stacks_chain_tip, 1000)
```
matteojug, line 5874: don't we need to recompute it, given the `write_bitcoin_transaction` above?

signer/src/block_observer.rs
djordon: > qq: is there a specific reason to do this before and not after `set_bitcoin_chain_tip`?

Not that I know of. I do not have strong preference for whether we should set stacks chain tip before the bitcoin chain tip or vice versa. So yeah, I can be persuaded either way.

signer/src/storage/model.rs
djordon: Hmmm, I thought I wrote the stacks block ID.

signer/tests/integration/postgres.rs
djordon: I don't think so, since it's only a transaction and not a block. But maybe I missed something?

signer/src/transaction_coordinator.rs
djordon: Updated the comment in https://github.com/stacks-sbtc/sbtc/pull/1885/commits/d2b69f869e9b93e83c3d407303fdd56a60a61e43

signer/src/storage/model.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1885/commits/a994081b81946aa367d430cac866b1fc5078019c

signer/src/api/info.rs
djordon: Oh right, updated in https://github.com/stacks-sbtc/sbtc/pull/1885/commits/79d3c8eb2518bcc3524791557f5ec9148b2cdf56

signer/tests/integration/postgres.rs
matteojug: ops, you're right 😄

signer/src/block_observer.rs
matteojug: After the full review I'm inclined to think having the bitcoin last seems safer at the possible cost of liveness in some unrealistic races, so I'm happy with its current form

PR #1886, opened 2026-01-12
chore: remove retries from integration tests

matteojug:
## Description

Closes: #?

## Changes

Removes `--retries 2` from integration tests.

As we don't get any alert from the GH action, even if it detects flaky tests we would not notice unless we manually inspect the (green) job output. Removing it we will surface such tests by failing; if it ends up we have unfixable flaky tests or too much noise we can introduce it back.

## Refs
 - https://nexte.st/docs/features/retries/#retries-and-flaky-tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,8 +87,8 @@ nextest-archive: emily-cdk-synth
 
 # Runs nextest archives
 nextest-archive-run:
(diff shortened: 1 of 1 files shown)


djordon (approved):
Yeah agreed. This will also prompt us to kill/fix any flaky tests, which is a good thing.

PR #1891, opened 2026-01-13
feat: note the canonical bitcoin blocks in the database

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1889

## Changes

1. Add a new column to the bitcoin blocks table that notes whether a given bitcoin block is canonical.
2. Add a new query that updates the database with a given chain tip to set whether the blocks in the database are canonical with respect to the given chain tip block hash.
3. Update the block observer to note whether the blocks in the database are canonical or not.

## Testing Information

I added a few integration tests, let me know if you can think of some scenarios that you want added.

## Checklist

### DIFF
--- a/signer/src/block_observer.rs
+++ b/signer/src/block_observer.rs
@@ -366,6 +366,14 @@ impl<C: Context, B> BlockObserver<C, B> {
 
         tracing::info!("updating canonical bitcoin blockchain to chain tip");
         let chain_tip: model::BitcoinBlockHash = chain_tip.into();
+
+        // This is a safeguard to prevent us from updating the is_canonical
+        // column of all blocks in the database if the given chain tip has
+        // not been written to the database.
+        if !db.is_known_bitcoin_block_hash(&chain_tip).await? {
+            tracing::error!("chain tip not found in database, not setting is_canonical column");
+            return Err(Error::UnknownBitcoinChainTip(chain_tip));
+        }
         db.set_canonical_bitcoin_blockchain(&chain_tip).await
     }
 
--- a/signer/src/error.rs
+++ b/signer/src/error.rs
@@ -681,6 +681,11 @@ pub enum Error {
     #[error("the given block hash could not be found in the database: {0}")]
     UnknownBitcoinBlock(bitcoin::BlockHash),
 
+    /// The given block hash could not be found in the database before
+    /// doing a DbRead::set_canonical_bitcoin_blockchain call.
+    #[error("the given chain tip block hash could not be found in the database: {0}")]
+    UnknownBitcoinChainTip(BitcoinBlockHash),
+
     /// No stacks chain tip found.
     #[error("no stacks chain tip")]
     NoStacksChainTip,


signer/migrations/0022__add_is_canonical_to_bitcoin_blocks.sql
```diff
@@ -0,0 +1,9 @@
+-- Add is_canonical column to bitcoin_blocks table to track which blocks
+-- are on the canonical Bitcoin blockchain.
+
+ALTER TABLE sbtc_signer.bitcoin_blocks
+ADD COLUMN is_canonical BOOLEAN;
```
MCJOHN974, line 5: Do we want it to be `NOT NULL`?

djordon: Not really. If the column had a `NULL` constraint, we'd need a default value for the existing blocks in the database. The sensible default would be `FALSE`, however, that feels wrong since we do not know whether the block is non-canonical or not. Instead, we can use `NULL` to mean that the value hasn't been set yet, which feels more correct here.

signer/src/block_observer.rs
```diff
@@ -363,13 +364,23 @@ impl<C: Context, B> BlockObserver<C, B> {
...
         for block_header in block_headers {
             self.process_bitcoin_block(block_header).await?;
         }
 
+        if let Some(chain_tip) = chain_tip.map(|h| h.hash.into()) {
+            tracing::info!("updating block canonical status");
+
+            let db = self.context.get_storage_mut();
+            db.set_canonical_bitcoin_blockchain(&chain_tip).await?;
+        }
```
djordon: I just realized that there is a bug here, and this code should node be in `process_bitcoin_blocks_until`. I'll update it in the morning.

signer/migrations/0022__add_is_canonical_to_bitcoin_blocks.sql
MCJOHN974: Yes, and I think we should set this value for all blocks during the migration. In my understanding we want to use this column like this `SELECT WHERE is_canonical == TRUE ORDER BY height` easily retrieveng whole canonical blockchain without recursive requests. 

However, if `is_canonical` is not `NOT NULL`, strictly speaking, there is no guarantees that `SELECT WHERE is_canonical == TRUE ORDER BY height` will return _whole_ canonical blockchain

MCJOHN974: We discussed it in person, now I agree that there is three states a block can be in `TRUE`, `FALSE` and `UNKNOWN`, and in current implementation `NULL` means `UNKNOWN`.  I think a enum with explicit `UNKNOWN` will be better but I'm ok to merge it as is

signer/src/block_observer.rs
djordon: This was fixed in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/2b5494a979f7d0954072b02d9ebc24690a7711e1, so we should be good now.

signer/src/block_observer.rs
```diff
@@ -347,6 +348,28 @@ impl<C: Context, B> BlockObserver<C, B> {
...
+    /// This function must only be called with the bitcoin chain tip since
+    /// it updates all blocks that are reachable from the given block hash
+    /// as canonical and may update blocks not reachable as non-canonical.
+    #[tracing::instrument(skip_all, fields(%chain_tip))]
+    async fn process_bitcoin_chain_tip(&self, chain_tip: BlockHash) -> Result<(), Error> {
+        self.process_bitcoin_blocks_until(chain_tip).await?;
+
+        let db = self.context.get_storage_mut();
+
+        tracing::info!("updating canonical bitcoin blockchain to chain tip");
```
matteojug, line 367: nit: can we add `chain_tip` in the log?

signer/src/block_observer.rs
```diff
@@ -347,6 +348,28 @@ impl<C: Context, B> BlockObserver<C, B> {
...
+        let chain_tip: model::BitcoinBlockHash = chain_tip.into();
+        db.set_canonical_bitcoin_blockchain(&chain_tip).await?;
+        Ok(())
```
matteojug: nit?
```suggestion
        db.set_canonical_bitcoin_blockchain(&chain_tip).await?
```

signer/tests/integration/postgres.rs
```diff
@@ -7871,3 +7871,353 @@ mod sqlx_transactions {
...
+    }
+
+    /// In this test, we check that it is not important whether the forked
+    /// blocks are already stored in the database or not when
+    /// DbWrite::set_canonical_bitcoin_blockchain() is called for the first
+    /// time, it still correctly mark the blocks as canonical or
+    /// non-canonical if they are part of the chain identified by the given
+    /// chain tip.
+    #[tokio::test]
+    async fn test_set_canonical_bitcoin_blockchain_with_fork_known_later() {
```
MCJOHN974, line 8408: nit: maybe use something like `TestData` + `get_rng()` or so in testing to generate various blockchains (and for example just test that in-memory and postgres canonical-related functions behave similarly)

signer/src/storage/memory/write.rs
```diff
@@ -385,6 +385,40 @@ impl DbWrite for SharedStore {
...
     }
+
+    async fn set_canonical_bitcoin_blockchain(
+        &self,
+        chain_tip: &model::BitcoinBlockHash,
+    ) -> Result<(), Error> {
+        let mut store = self.lock().await;
+        store.version += 1;
+
+        // Then, recursively mark all blocks reachable from the chain tip as canonical
```
matteojug, line 397: I guess the part clearing `canonical_bitcoin_blocks` before this got lost somehow? The comment here suggest we should do something first

signer/src/testing/blocks.rs
```diff
@@ -63,32 +67,67 @@ impl BitcoinChain {
...
+            .range(..height.into())
+            .map(|(height, block)| (*height, block.clone()))
+            .collect();
+
+        let mut fork_chain = Self(fork);
+
+        for _ in 0..num_blocks {
+            let new_block = fork_chain.chain_tip().new_child();
+            fork_chain.0.insert(new_block.block_height, new_block);
+        }
```
matteojug: nit? more wasteful but nicer to read
```suggestion
        fork_chain.generate_blocks(num_blocks);
```

signer/tests/integration/block_observer.rs
```diff
@@ -1874,3 +1874,210 @@ fn make_coinbase_deposit_request(
     };
     (deposit_tx, req, info)
 }
+
+/// This test checks that the block observer marks the canonical status of
+/// bitcoin blocks in the database whenever a new block is observed.
+#[tokio::test]
+async fn block_observer_marks_bitcoin_blocks_as_canonical() {
+    let (rpc, faucet) = regtest::initialize_blockchain();
```
matteojug: now that the parallel stuff is on main (🥳 ) consider using it here or mark the test as serial (after syncing main), otherwise bad things may happen; also for the other one

signer/tests/integration/block_observer.rs
```diff
@@ -1874,3 +1874,210 @@ fn make_coinbase_deposit_request(
...
+    assert_eq!(chain_tip_before_invalidation_status, Some(true));
+
+    // Now invalidate the chain tip
+    rpc.invalidate_block(&chain_tip_before_invalidation)
+        .unwrap();
+
+    // Sometimes bitcoin-core will generate an invalid block (maybe the
+    // same block?), after invalidating the chain tip. One wait to combat
+    // this is to generate new blocks to a new address, ensuring the new
+    // block is distinct from the previous ones.
```
matteojug: I thought the issue was that the new block would (always?) result in the same block hash as the invalidated one. I usually just `faucet.send_to(1001, &faucet.address);` before the new `faucet.generate_block()` to ensure a different block hash, just in case you would also prefer it to `generate_to_address`. Feel free to resolve this

signer/tests/integration/block_observer.rs
```diff
@@ -1874,3 +1874,210 @@ fn make_coinbase_deposit_request(
     };
     (deposit_tx, req, info)
 }
+
+/// This test checks that the block observer marks the canonical status of
+/// bitcoin blocks in the database whenever a new block is observed.
+#[tokio::test]
+async fn block_observer_marks_bitcoin_blocks_as_canonical() {
```
matteojug: nit: can we parametrize this and run it also with a fork consisting of a single block (sibling of the invalidated one)?

signer/tests/integration/postgres.rs
```diff
@@ -7871,3 +7871,353 @@ mod sqlx_transactions {
...
+        let fork_chain = canonical_chain.fork_at_height(2u64, 2);
+
+        for block in &fork_chain {
+            db.write_bitcoin_block(block).await.unwrap();
+        }
+
+        let fork_block_1 = fork_chain.nth_block(2u64.into());
+        let fork_block_2 = fork_chain.nth_block(3u64.into());
+
+        // Verify that fork blocks are marked as non-canonical
```
matteojug: nit?
```suggestion
        // Verify that fork blocks are not marked
```

signer/tests/integration/postgres.rs
```diff
@@ -7871,3 +7871,353 @@ mod sqlx_transactions {
...
+            .is_block_canonical(&fork_block_2.block_hash)
+            .await
+            .unwrap();
+        assert_eq!(fork_2_is_canonical, None);
+
+        // These new blocks aren't part of the canonical chain, so they
+        // should be marked as non-canonical. In order to make sure that
+        // they get marked as such, we need to clear the is_canonical
+        // column from some of the blocks in the table.
+        clear_is_canonical_bitcoin_blocks(&db).await;
```
matteojug, line 8161: nit: can we also call `set_canonical_bitcoin_blockchain` before the clear and verify that we don't touch the null ones if not necessary?

signer/src/block_observer.rs
djordon: It should be there already from the tracing instrument.

signer/src/storage/memory/write.rs
djordon: Yeah maybe? I don't recall

signer/tests/integration/block_observer.rs
djordon: Done in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/a2d57b97fdb62d0fa12f84edbe09f5645b23b32b 🥳

signer/tests/integration/block_observer.rs
djordon: Done in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/a98677393b12a1ce6a08e2b5c27bba8b77cbf5bc

signer/tests/integration/postgres.rs
djordon: Well, I did this in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/5f1803571ad19c7a7526d375477c15d9a729bb36

signer/tests/integration/postgres.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/43feedd9d5a14b82fe1fc250643a8a46a0f4e590

signer/tests/integration/block_observer.rs
djordon: Yeah sending a transaction leads to a shorter diff, so let's go with that: https://github.com/stacks-sbtc/sbtc/pull/1891/commits/69b569809f34b9dbb0963951554687e22f68f222

signer/src/testing/blocks.rs
djordon: Done in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/7185893b1c911c6cbca2a5e68ef68afcbac70a84

signer/src/storage/memory/write.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/726d0eaf9c710d0812e995201eb87ff818527ef7

signer/src/block_observer.rs
djordon: Done in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/2662d9df2985a6fc529af284ac6fc752857d3553

signer/tests/integration/postgres.rs
djordon: Done in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/c8d789644811d30941aceb1349e78787eacaef8e

signer/src/block_observer.rs
```diff
@@ -347,6 +348,27 @@ impl<C: Context, B> BlockObserver<C, B> {
...
+    /// as canonical and may update blocks not reachable as non-canonical.
+    #[tracing::instrument(skip_all, fields(%chain_tip))]
+    async fn process_bitcoin_chain_tip(&self, chain_tip: BlockHash) -> Result<(), Error> {
+        self.process_bitcoin_blocks_until(chain_tip).await?;
+
+        let db = self.context.get_storage_mut();
+
+        tracing::info!("updating canonical bitcoin blockchain to chain tip");
+        let chain_tip: model::BitcoinBlockHash = chain_tip.into();
+        db.set_canonical_bitcoin_blockchain(&chain_tip).await
```
MCJOHN974, line 377: Would it be better to set canonical blocks on each write of bitcoin block in the db? I think it would be not so much extra work, but it feels safer. Like, what if `process_bitcoin_blocks_until` would later rely on `is_canonical` and also insert new blocks in the db. And in general I think it is nice to have an invariant that "is_canonical is always correct" while it costs us more or less zero

djordon: The problem with calling this within `process_bitcoin_blocks_until` or `process_bitcoin_block` is that those functions can be called with bitcoin blocks that are non-canonical and we only want to call `set_canonical_bitcoin_blockchain` with the chain tip. Specifically, those functions are also called from the request decider, and the input blocks need not be canonical.

MCJOHN974: Hmm, then I'd probably add `maybe_chaintip` parameter into `write_bitcoin_block` function itself. And `write_bitcoin_block` calls `set_canonical_bitcoin_blockchain` with provided chaintip or chaintip extracted from db (if `maybe_chaintip.is_none()`.

If we merge sql from `set_canonical_bitcoin_blockchain` with sql from `write_bitcoin_block` I think it will be very little overhead, while we would have a guarantee that `is_canonical` field is always properly set.

djordon: I don't think that it works so nicely with forks and the logic that we have when updating the canonical chain.

Here are my thoughts:
1. A lot of these suggestions were considered, and they appeared to me to be trickier than they seemed.
2. A lot of these suggestions are "better", in some sense.
3. Since I agree that there are kinda better, I don't want to discourage them, but I think it would be better to actually work through them and see whether they fail in a separate stream of work.

So I encourage you to properly work through your suggestions in a separate branch. If you could get it to work, then great, we can PR it.

signer/tests/integration/postgres.rs
```diff
@@ -7991,6 +8004,103 @@ mod canonical_bitcoin_blockchain {
...
+            num_bitcoin_blocks: 128,
+            num_stacks_blocks_per_bitcoin_block: 0,
+            num_deposit_requests_per_block: 0,
+            num_withdraw_requests_per_block: 0,
+            num_signers_per_request: 0,
+            consecutive_blocks: false,
+        };
+        let blockchains = TestData::generate(&mut rng, &signer_keys, &params);
+
+        // Insert all blocks from into the database
```
matteojug: nit: from?

signer/src/block_observer.rs
djordon: > I don't think that it works so nicely with forks and the logic that we have when updating the canonical chain.

Just wanted to quickly clarify that there are implementations that work nicely with what we have here, they are just more involved changes. They may be better, in some sense, so it's probably worth exploring. Hey, even I may explore them, but probably on a separate branch.

signer/tests/integration/block_observer.rs
```diff
@@ -1889,3 +1889,198 @@ fn make_coinbase_deposit_request(
...
+    .unwrap();
+
+    // Verify the new blocks are in the database
+    for new_block in new_blocks {
+        let db_new_block = db.get_bitcoin_block(&new_block).await.unwrap();
+        assert!(db_new_block.is_some());
+        assert_eq!(db.is_block_canonical(&new_block).await.unwrap(), Some(true));
+    }
+
+    // Check that the new blocks have is_canonical = true
```
matteojug: nit: old comment, should probably be moved up in the `for`

signer/tests/integration/postgres.rs
```diff
@@ -7865,3 +7865,469 @@ mod sqlx_transactions {
...
+    /// non-canonical if they are part of the chain identified by the given
+    /// chain tip.
+    #[tokio::test]
+    async fn test_set_canonical_bitcoin_blockchain_with_fork_known_later() {
+        let db = testing::storage::new_test_database().await;
+
+        // Create two blockchains, one of length 10 and another of length
+        // 20, where one forks the other at height 3 (the block with height
+        // 3 differs from one to the other but their earlier blocks are the
+        // same)
```
matteojug, line 8414: nit: now fork chain is 17 blocks long, fork block is at height 2

signer/src/block_observer.rs
MCJOHN974: Okay I don't mind to do it in separate pr

signer/tests/integration/postgres.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/d5e4f5e35462ae3b76bd50304c66bb994f7c7cdc.

signer/tests/integration/block_observer.rs
djordon: Right, moved in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/d5e4f5e35462ae3b76bd50304c66bb994f7c7cdc

signer/tests/integration/postgres.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/d5e4f5e35462ae3b76bd50304c66bb994f7c7cdc

signer/tests/integration/block_observer.rs
```diff
@@ -2009,3 +2009,197 @@ async fn block_observer_handles_deposits_with_high_max_fee() {
...
+
+    ctx.wait_for_signal(Duration::from_secs(8), |signal| {
+        matches!(
+            signal,
+            SignerSignal::Event(SignerEvent::BitcoinBlockObserved(block_ref))
+                if block_ref.block_hash == chain_tip_before_invalidation
+        )
+    })
+    .await
+    .unwrap();
```
matteojug: nit: now we have `wait_for_block_observed`, also below

signer/tests/integration/postgres.rs
```diff
@@ -8049,3 +8049,469 @@ mod sqlx_transactions {
...
+            assert_eq!(is_canonical, Some(true));
+        }
+
+        // Check that original main chain blocks with height 0 to 1 are canonical
+        for (_, block) in main_chain.range(stable_block_heights) {
+            let is_canonical = db.is_block_canonical(&block.block_hash).await.unwrap();
+            assert_eq!(is_canonical, Some(true));
+        }
+
+        // Check that original main chain blocks with height 1 to 9 are
```
matteojug: nit
```suggestion
        // Check that original main chain blocks with height 2 to 9 are
```

signer/src/storage/postgres/write.rs
```diff
@@ -1001,6 +1002,73 @@ impl PgWrite {
...
+            ),
+            min_height AS (
+                SELECT COALESCE(MIN(block_height), 0) AS height
+                FROM canonical_chain
+            )
+            UPDATE sbtc_signer.bitcoin_blocks
+            SET is_canonical = (block_hash IN (
+                SELECT block_hash FROM canonical_chain
+            ))
+            WHERE block_height >= (SELECT height FROM min_height)
```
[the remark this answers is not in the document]
djordon: Oh interesting. I suppose it's not a bad idea as a safeguard.

signer/migrations/0022__add_is_canonical_to_bitcoin_blocks.sql
```diff
@@ -0,0 +1,9 @@
...
+
+-- Create an index on is_canonical for efficient queries filtering
+-- canonical vs non-canonical blocks.
+CREATE INDEX ix_bitcoin_blocks_is_canonical ON sbtc_signer.bitcoin_blocks(is_canonical);
```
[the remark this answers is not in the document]
djordon: Hmm, I thought about doing this and decided against it. The partial index doesn't buy is much in this case, since most rows will have `is_canonical = TRUE` anyway. I would expect that 99%+ of rows will have `is_canonical` set to `TRUE` so a partial index doesn't save much. If it turns out that adding a partial index is the way to do, we can add it then and remove this one or something.

Also, I don't feel too strongly here, so if someone else does speak up, I don't mind changing it.

signer/src/storage/memory/write.rs
```diff
@@ -385,6 +385,41 @@ impl DbWrite for SharedStore {
...
+        chain_tip: &model::BitcoinBlockHash,
+    ) -> Result<(), Error> {
+        let mut store = self.lock().await;
+        store.version += 1;
+
+        store.canonical_bitcoin_blocks.clear();
+        // Then, recursively mark all blocks reachable from the chain tip as canonical
+        if let Some(block) = store.bitcoin_blocks.get(chain_tip).cloned() {
+            store.canonical_bitcoin_blocks.insert(*chain_tip, block);
+        }
```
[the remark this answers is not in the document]
djordon: Ummm, maybe I'm missing something, but this matches the behavior of the query and matches what is expected in the docs of the trait method. So until we change that, this should stay as is.

signer/tests/integration/block_observer.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/548025d0ce5e0f758483cc43d811296d0975cd18

signer/tests/integration/postgres.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/a9b09db1909057be12ef4d45de2dcc453512c1d6

signer/tests/integration/postgres.rs
```diff
@@ -8049,3 +8049,469 @@ mod sqlx_transactions {
...
+
+        let fork_2_is_canonical = db
+            .is_block_canonical(&fork_block_2.block_hash)
+            .await
+            .unwrap();
+        assert_eq!(fork_2_is_canonical, None);
+
+        // Let's set the canonical chain again and verity that the new
+        // blocks are still not marked as canonical. Note that the
+        // guarantees of the set_canonical_bitcoin_blockchaion function is
```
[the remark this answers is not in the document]
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1891/changes/14ad78158f39ae2601a644ceabbb9813cfa47b24

signer/tests/integration/postgres.rs
```diff
@@ -8049,3 +8049,469 @@ mod sqlx_transactions {
...
+            .unwrap();
+        assert_eq!(fork_1_is_canonical, None);
+
+        let fork_2_is_canonical = db
+            .is_block_canonical(&fork_block_2.block_hash)
+            .await
+            .unwrap();
+        assert_eq!(fork_2_is_canonical, None);
+
+        // Let's set the canonical chain again and verity that the new
```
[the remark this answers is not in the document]
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1891/changes/14ad78158f39ae2601a644ceabbb9813cfa47b24

signer/src/testing/blocks.rs
```diff
@@ -13,14 +16,15 @@ use crate::storage::model::StacksBlockHeight;
 
 /// Represents a naive, sequential chain of bitcoin blocks and provides basic
 /// functionality for manipulation. Does not handle forks/branches.
```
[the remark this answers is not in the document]
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1891/changes/14ad78158f39ae2601a644ceabbb9813cfa47b24

signer/src/storage/postgres/write.rs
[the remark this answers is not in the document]
matteojug: We are trusting the poller (and Bitcoin rpc), but if we were to get a chain tip at height lower than `get_sbtc_bitcoin_start_height()` we will not write the block in `process_bitcoin_blocks_until`, and thus we will clear all the canonical flags. This does seem correct -- if that's the chain tip, everything we have in db is not canonical -- but maybe explicitly ensuring we have written the block and failing if not may be a reasonable safeguard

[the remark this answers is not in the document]
djordon: I added a safeguard check in https://github.com/stacks-sbtc/sbtc/pull/1891/commits/b0f93465c1f68b785acbe7baf6970da3153f2c0d.
1. The check is outside the function call itself. I find it easier to reason about by having two simple functions doing simple things.
2. There isn't an integration test. I tried adding one, but ran into a catch-22. When we do not write the chain-tip to the database, we cannot update the signer state and do not send the block observed signal. If we write the block to the database, then we cannot test the condition that we want to test. So either way the test ends up feeling goofy.

PR #1893, opened 2026-01-14
refactor: add small emily client creation function for tests

djordon:
## Description

This closes nothing. Nothing!

## Changes

* Add a small helper function for creating the Emily client used in our tests.
* Use this new test function where is makes sense.

## Checklist

### DIFF
--- a/signer/tests/integration/emily.rs
+++ b/signer/tests/integration/emily.rs
@@ -571,12 +571,12 @@ async fn get_deposit_request_works() {
     assert!(request.is_none());
 }
 
-#[test_case(3, 10, Some(2), 3; "handles paging")]
-#[test_case(3, 0, Some(2), 2; "handles timeout")]
+#[test_case(3, Duration::from_secs(10), Some(2), 3; "handles paging")]
+#[test_case(3, Duration::from_secs(0), Some(2), 2; "handles timeout")]
 #[tokio::test]
 async fn test_get_deposits_with_status_request_paging(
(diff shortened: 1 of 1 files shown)


signer/src/emily_client.rs
```diff
@@ -197,6 +197,15 @@ impl EmilyClient {
     }
 }
 
+#[cfg(any(test, feature = "testing"))]
+impl EmilyClient {
+    /// Create a new Emily Client for use in our integration tests.
+    pub fn new_test_client() -> Result<Self, Error> {
+        let url = Url::parse("http://testApiKey@localhost:3031").map_err(Error::InvalidUrl)?;
+        EmilyClient::try_new(&url, Duration::from_secs(1), None)
+    }
```
MCJOHN974, line 206: nit: Since we unwrap result of this function everywhere we can just unwrap inside the function.

djordon: I'm okay with either, but unwrapping in the function has the downside of hiding the original call site if such an error occurs. In this case, you can track it down easily since you know the failing test and the client is created once per test. So maybe it's not so bad here.

MCJOHN974: Yeas, and also if this unwrap fails in 99% of cases it probably means that you forgot to do `make integration-env-up` (and it fails in all tests)

djordon: 😂

PR #1894, opened 2026-01-14
chore: slightly prettier Emily logs

djordon:
## Description

Our Emily logs are not so friendly on the eyes. This change makes them a nicer. Also, we should consider whether we want the full body in the log.

## Changes

* Use the `Display` impl for writing values in our logs where possible.

## Checklist

### DIFF
--- a/.github/workflows/deploy-emily.yaml
+++ b/.github/workflows/deploy-emily.yaml
@@ -264,5 +264,5 @@ jobs:
           echo "Done bootstrapping CDK."
 
           echo "Deploying the stack..."
-          npx aws-cdk deploy
+          npx aws-cdk deploy --require-approval never
           echo "Deployed Emily Successfully!" >> $GITHUB_STEP_SUMMARY


emily/handler/src/api/routes/mod.rs
```diff
@@ -34,10 +34,10 @@ where
 {
     let as_response = reply.into_response();
     tracing::debug!(
-        event = ?"response",
-        status = ?as_response.status().as_u16(),
-        body = ?format!("{:?}", as_response.body()),
-        headers = ?format!("{:?}", as_response.headers()),
+        event = "response",
+        status = as_response.status().as_u16(),
+        body = ?as_response.body(),
```
djordon, line 42: Should we remove the response body from our logs?

djordon: We decided to move it to a trace log, which was done in https://github.com/stacks-sbtc/sbtc/pull/1894/commits/0134899861411d28b287cc04d759ff862a305011

PR #1902, opened 2026-01-16
feat: parallel tests - transaction coordinator

matteojug:
## Description

Partial work for https://github.com/stacks-sbtc/sbtc/issues/1854
Breaking up https://github.com/stacks-sbtc/sbtc/pull/1875

Consider hiding whitespaces for a sane review experience.

## Changes

Use parallel testing setup for transaction signer tests. One test that was already pretty fast was marked as serial.

## Testing Information

CI runs parallel tests.

## Checklist


sbtc/src/testing/regtest.rs
```diff
@@ -383,6 +383,17 @@ impl<'a> Faucet<'a> {
         self.rpc.send_raw_transaction(&tx).unwrap();
         OutPoint::new(tx.compute_txid(), 0)
     }
+
+    /// Generate some transactions to ensure bitcoincore has enough data to
+    /// estimate fees
+    pub fn generate_fee_data(&self) {
+        for _ in 0..10 {
```
djordon, line 390: Can this work with some value less than 10, like 4? I suppose it okay either way.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -4706,23 +4699,22 @@ async fn test_conservative_initial_sbtc_limits() {
...
 
+    let (emily_client, emily_tables) = new_emily_setup().await;
     let emily_config = emily_client.config().as_testing();
 
-    testing_api::wipe_databases(&emily_config).await.unwrap();
-
     let network = WanNetwork::default();
 
+    // Ensure we can estimate fees
+    faucet.generate_fee_data();
```
djordon, line 4716: Seems like this is in the `sign_bitcoin_*` tests and maybe some others but not all. I suppose that is correct, just wondering if this was some other reason.

djordon (commented):
Looks good, just had some questions.

signer/tests/integration/transaction_coordinator.rs
matteojug: I've only added it to tests that would otherwise fail. We could add it to all the tests (directly in `BitcoinContainer::from_containers`) since it only takes ~200ms, it shouldn't impact that much.

sbtc/src/testing/regtest.rs
matteojug: I've added a test for `generate_fee_data`, running it multiple times it seems is not deterministic how many blocks/txs are required to make bitcoind happy. Fixing 3 transactions per block, less than 10 blocks are not always (in 100 runs) enough.

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -1564,19 +1562,13 @@ async fn run_subsequent_dkg() {
...
-    let emily_client = EmilyClient::try_new(
-        &Url::parse("http://testApiKey@localhost:3031").unwrap(),
-        Duration::from_secs(1),
-        None,
-    )
-    .unwrap();
+    let stack = TestContainersBuilder::start_bitcoin().await;
+    let bitcoin = stack.bitcoin().await;
+    let rpc = bitcoin.rpc();
+    let faucet = &bitcoin.get_faucet();
```
MCJOHN974, line 1569: nit: maybe unite this to single helper function

signer/tests/integration/transaction_coordinator.rs
djordon: That makes sense but I'm okay with doing it either way. Your call.

djordon (dismissed):
Looks good

signer/tests/integration/transaction_coordinator.rs
matteojug: We cannot merge the stack and bitcoin since it wouldn't work for tests using Stacks. We could do that for rpc and faucet, but I opted for keeping those separated as we are not always using both, and wrt to `initialize_blockchain` we are constructing a faucet and not returning a reference to a common one.

So I would vote for keeping those as is and leave refactoring the tests to remove the redundancies (in general, not just for that) another time, WDYT?

MCJOHN974 (dismissed):

djordon (approved):

MCJOHN974 (approved):

Issue #1884, opened 2026-01-05
[Feature]: Add a timeout to all requests


djordon:
# Feature - Add a timeout to Emily requests

## 1. Description

The sbtc signer has timeouts on some requests, but not others. Requests to the signer's stacks node have a timeout, while requests to Emily or their Bitcoin node do not. We should always have a timeout on our requests so that we do not stall one of the signer's long lived processes with a request that takes an unreasonable amount of time.

## 2. Technical Details

The Emily and Bitcoin clients have interfaces where we can set a timeout on various aspects of a request. We should set them.

- [ ] Requests to bitcoin core and Emily have a maximum amount of time that they could take.

PR #1896, opened 2026-01-16
feat: add bitcoin timeout

MCJOHN974:
## Description

Closes: #1884

## Changes

Note, that jsonrpc is not a new dependency, bitcoincore_rpc already imported it.

Adds timeout for bitcoin requests

## Checklist

- [ ] I have performed a self-review of my code
- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1903, opened 2026-01-26
feat: add emily timeout

MCJOHN974:
## Description

Closes: #1884 (together with #1896)

## Checklist

### DIFF
--- a/signer/src/emily_client.rs
+++ b/signer/src/emily_client.rs
@@ -507,7 +507,7 @@ impl TryFrom<&EmilyClientConfig> for ApiFallbackClient<EmilyClient> {
         let clients = config
             .endpoints
             .iter()
-            .map(|url| EmilyClient::try_new(url, config.pagination_timeout, config.timeout, None))
+            .map(|url| EmilyClient::try_new(url, config.timeout, config.pagination_timeout, None))
             .collect::<Result<Vec<_>, _>>()?;
 
         Self::new(clients).map_err(Into::into)


signer/src/error.rs
```diff
@@ -710,6 +710,10 @@ pub enum Error {
     #[error("could not parse hex txid: {0}")]
     DecodeHexTxid(#[source] bitcoin::hex::HexToArrayError),
 
+    /// Could not create reqwest client
+    #[error("could not create reqwest client {0}")]
+    ReqwestClientCreation(#[source] reqwest_012::Error),
```
djordon: We can make this variant more specific to the case at hand since it is only used when creating the Emily client. So we should change the error message to be more specific about what went wrong, and also change the variant name to be more specific. For the message, maybe something like: "we received an error when creating the Emily's reqwest client".

signer/tests/integration/emily.rs
```diff
@@ -698,3 +699,34 @@ async fn test_get_deposits_returns_pending_and_accepted() {
...
+
+#[tokio::test]
+async fn emily_timeout_works() {
+    // Client with 1 sec timeout works normally
+    let client = EmilyClient::try_new(
+        &url::Url::parse("http://testApiKey@localhost:3031").unwrap(),
+        Duration::from_secs(1),
+        Duration::from_secs(1),
+        None,
+    )
```
djordon: You might want to merge in main, since we should have some new Emily helper functions.

signer/src/emily_client.rs
```diff
@@ -173,6 +175,13 @@ impl EmilyClient {
         config.base_path = url.to_string().trim_end_matches("/").to_string();
         config.api_key = api_key;
 
+        let c = reqwest_012::Client::builder()
```
djordon: Let's call this `client` instead of `c`.

signer/src/config/mod.rs
```diff
@@ -286,6 +286,9 @@ pub struct EmilyClientConfig {
     /// Pagination timeout in seconds.
     #[serde(deserialize_with = "duration_seconds_deserializer")]
     pub pagination_timeout: std::time::Duration,
+    /// Timeout in seconds.
```
djordon:
```suggestion
    /// The max time waiting for an HTTP response from Emily.
```

signer/src/config/mod.rs
```diff
@@ -286,6 +286,9 @@ pub struct EmilyClientConfig {
...
+    #[serde(deserialize_with = "duration_seconds_deserializer")]
+    pub timeout: std::time::Duration,
```
djordon, line 291: We should add validation that this timeout is greater than zero.

signer/src/emily_client.rs
MCJOHN974: [0a5231d](https://github.com/stacks-sbtc/sbtc/pull/1903/commits/0a5231d3c1e78db1b5af8a417e1290793f7e2c48)

signer/src/error.rs
MCJOHN974: [0a5231d](https://github.com/stacks-sbtc/sbtc/pull/1903/commits/0a5231d3c1e78db1b5af8a417e1290793f7e2c48)

signer/src/config/mod.rs
MCJOHN974: [ad80600](https://github.com/stacks-sbtc/sbtc/pull/1903/commits/ad806002e7297147451d07ae1cb85354e44b9d0a)

signer/tests/integration/emily.rs
```diff
@@ -698,3 +699,34 @@ async fn test_get_deposits_returns_pending_and_accepted() {
 
     clean_emily_setup(emily_tables).await;
 }
+
+#[tokio::test]
+async fn emily_timeout_works() {
```
matteojug: Can we use the parallel setup for this test?

signer/src/config/default.toml
```diff
@@ -43,6 +43,11 @@ endpoints = [
 # Environment: SIGNER_EMILY__PAGINATION_TIMEOUT
 # pagination_timeout = 10
 
+# The timeout, in seconds, used to fetch deposits requests from Emily.
+# Required: false
```
matteojug: nit: isn't this used for all Emily requests?

signer/src/config/mod.rs
```diff
@@ -296,6 +299,20 @@ impl Validatable for EmilyClientConfig {
                 "[emily_client] At least one Emily API endpoint must be provided".to_string(),
             ));
         }
+
+        // All durations should be non-zero
+        let zero = std::time::Duration::ZERO;
+        if self.timeout == zero {
+            return Err(ConfigError::Message(
+                SignerConfigError::ZeroDurationForbidden("emily_timeout").to_string(),
```
matteojug: nit: something like `emily::timeout`? To ensure people don't try to look for `emily_timeout` in the config; also for `pagination_timeout`

signer/src/emily_client.rs
```diff
@@ -140,6 +141,7 @@ impl EmilyClient {
     pub fn try_new(
         url: &Url,
         pagination_timeout: Duration,
+        timeout: Duration,
```
matteojug, line 143: nit: can we move `timeout` above `pagination_timeout`? I find `pagination_timeout` and `page_size` better close together, and `timeout` being more "important" than `pagination_timeout`. Or not, feel free to ignore this

signer/src/config/default.toml
MCJOHN974: [30437d5](https://github.com/stacks-sbtc/sbtc/pull/1903/commits/30437d53ab238237be1f6f675d90767bbc127684)

signer/src/config/mod.rs
MCJOHN974: [f1db1e3](https://github.com/stacks-sbtc/sbtc/pull/1903/commits/f1db1e3d492643b3ff2bb8f6b822189da63c2555)

signer/tests/integration/emily.rs
MCJOHN974: Hmm, I think point of this test is mostly to verify that `EmilyClient::try_new` indeed creates client with given timeout, so I don't think we can comfortably reuse `new_emily_setup` here. We can do what `new_emily_setup` does here, but probably it don't worth it??

matteojug: We are calling an endpoint that uses a table (`get_limits`), doing that concurrently with other tests may result in some shenanigans.

But yeah, we cannot use `new_emily_setup` easily. Can you mark the test as serial by putting it in a `mod serial`?

signer/src/config/mod.rs
djordon: I think you'd want to use `emily.timeout` but I suppose `emily::timeout` is acceptable too.

signer/src/config/default.toml
```diff
@@ -43,6 +43,11 @@ endpoints = [
 # Environment: SIGNER_EMILY__PAGINATION_TIMEOUT
 # pagination_timeout = 10
 
+# The timeout, in seconds, used to fetch requests from Emily.
```
djordon:
```suggestion
# The timeout, in seconds, used when making requests to Emily.
```

djordon (commented):
One small nit

signer/src/emily_client.rs
MCJOHN974: Whatever order you prefer =)

[07eee0b](https://github.com/stacks-sbtc/sbtc/pull/1903/commits/07eee0b9237e24beeb413845b23035c62b9c87cd)

signer/tests/integration/emily.rs
MCJOHN974: [1bc7d64](https://github.com/stacks-sbtc/sbtc/pull/1903/commits/1bc7d6419da1dc37a7f147bd113d554deafed1b9)

signer/src/config/mod.rs
```diff
@@ -956,6 +975,21 @@ mod tests {
         assert_eq!(height, 12345u64.into());
     }
 
+    #[test]
+    fn emily_with_environment() {
```
djordon, line 979: We should probably have a test that having a zero duration for the timeout leads to an error.

signer/src/emily_client.rs
```diff
@@ -498,7 +507,7 @@ impl TryFrom<&EmilyClientConfig> for ApiFallbackClient<EmilyClient> {
         let clients = config
             .endpoints
             .iter()
-            .map(|url| EmilyClient::try_new(url, config.pagination_timeout, None))
+            .map(|url| EmilyClient::try_new(url, config.pagination_timeout, config.timeout, None))
```
djordon: This order is now incorrect.

signer/src/config/mod.rs
MCJOHN974: [c166881](https://github.com/stacks-sbtc/sbtc/pull/1903/commits/c16688109d615b81a308d69ca7d2dfda0045c713)

signer/src/emily_client.rs
MCJOHN974: [bfc5ad7](https://github.com/stacks-sbtc/sbtc/pull/1903/commits/bfc5ad7d3d85a0c10be079dc11a890ad20cba744)

Issue #1863, opened 2025-11-19
[Tech Debt]: Add AWS lambda invocation ID to our logs.


MCJOHN974:
## Description

Currently, Emily is an AWS Lambda, and for each invocation of Lambda, AWS assigns a unique ID. We should add these IDs to our logs generated from Rust. Ideally, CloudWatch have a special column called requestId, but at least add it to the log message.

PR #1904, opened 2026-01-27
feat: add lambda invocation id to logs

MCJOHN974:
## Description

Closes: #1863

## Changes

Added aws lambda invocation ID, and x-ray trace ID to emily logs.

## Testing Information

Manual looking at logs

## Checklist


emily/handler/src/bin/emily-lambda.rs
```diff
@@ -27,15 +25,34 @@ async fn main() {
...
         .build();
 
-    // Setup service filters.
-    let service_filter = api::routes::routes_with_stage_prefix(context)
-        .recover(api::handlers::handle_rejection)
+    let inject_lambda_id = warp::any()
+        .and(warp::ext::optional::<Context>())
+        .map(|ctx: Option<Context>| {
+            if let Some(c) = ctx {
+                tracing::Span::current().record("request_id", c.request_id.as_str());
```
djordon, line 32: What sets the `request_id`, AWS? And how does it differ from the `x-amzn-trace-id`?

MCJOHN974: This `request_id` is an actual id of lambda invocation, which is chosen by some lambda environment at the moment it is called. You can see a lot of such ids in our logs like `STARTING request with id xxxxxx`, `FINISHED request with id xxxxxx. Duration yyyyy, billed duration zzzzz`.

The `x-amzn-trace-id` is chosen by API gateway, and it should be persistent across multiple aws services for same request. For example if you request Emily about something, then, Emily, while processing your request requests DynamoDB, DynamoDB should see same trace-id. So, given the fact that we don't use this trace id anywhere else and emily don't really do a lot of requests to other services during its work we probably should delete it.

djordon: Gotcha. The `request_id` is is set from AWS Lambda and the `x-amzn-trace-id` header is from API Gateway. And we could propagates this `x-amzn-trace-id` header in requests to DynamoDB, but we don't.

emily/handler/src/bin/emily-lambda.rs
```diff
@@ -27,15 +25,34 @@ async fn main() {
...
+
+    let service_filter = inject_lambda_id
+        .and(service_routes)
+        .with(warp::trace(|info| {
+            let trace_id = info
+                .request_headers()
+                .get("x-amzn-trace-id")
+                .and_then(|v| v.to_str().ok())
+                .unwrap_or("unknown");
+            info_span!("request", request_id = tracing::field::Empty, trace_id = %trace_id)
```
djordon, line 48: We probably want this to be a `debug_span`

djordon (approved):
Largely, looks good, we should probably change the `info_span` to be a `debug_span`.

emily/handler/src/bin/emily-lambda.rs
MCJOHN974: Hmm, what the difference would be in our use case? 

From the point of generating this logs, I think we don't plan to use any level of logging on Emily rather then DEBUG in prod.

From the point of searching through the logs I don't think that any particular log level on this span would make searching easier.

From the point "what log level sounds right?" for me info sounds siutable here because it is some information indicating interaction with outer world (user request) and not an internal detail

djordon: Oh right, I had thought that the span's INFO level affected whether the attributes were attached to the DEBUG log event, but does not seem to be the case. So yeah, I don't think this matters for our case.

PR #1912, opened 2026-02-03
feat: use the signer set in the registry for picking the coordinator

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1911

## Changes

* Use the registry as the source of truth for the signer set

Note that these changes only matter if there is a signer set change that is about to happen.

## Testing Information

I tested this branch on the usual flow on devenv, and it worked fine. I did not do any devenv testing that puts this new logic to the test though.

## Checklist

### DIFF
--- a/signer/src/transaction_coordinator.rs
+++ b/signer/src/transaction_coordinator.rs
@@ -1731,6 +1731,11 @@ where
         S: Stream<Item = Signed<SignerMessage>>,
         Coordinator: WstsCoordinator,
     {
+        // We only allow the coordinator to send us certain kinds of
+        // messages. So later, we need to check whether the sender is the
+        // coordinator, and for that we need to use the "coordinator signer
+        // set", which may differ from the set of signers sending us
+        // messages right now.
         let signer_set = self.context.coordinator_signer_set();
         tokio::pin!(signal_stream);
 
@@ -2858,7 +2863,7 @@ mod tests {
             .unwrap();
     }
 
-    /// Check that we is_coordinator uses the registry for figuring out who
+    /// Check that is_coordinator uses the registry for figuring out who
     /// the cooridnator is and falls back to the config if the registry has
     /// no signer set info.
     #[tokio::test]


matteojug (commented):
Don't we need to change also `TxSignerEventLoop::inspect_msg_chain_tip`?

signer/src/transaction_coordinator.rs
```diff
@@ -1860,15 +1860,20 @@ where
     /// Determine if this signer is the signer set's coordinator for the
     /// specified bitcoin block hash.
     ///
-    /// The coordinator is decided using the hash of the bitcoin chain tip.
-    /// We don't use the chain tip directly because it typically starts
-    /// with a lot of leading zeros.
+    /// The coordinator is decided using the hash of the bitcoin chain tip
+    /// and signer set info from the registry if present. We don't use the
+    /// chain tip directly because it typically starts with a lot of
+    /// leading zeros.
```
MCJOHN974, line 1871: It seems this comment should be around `coordinator_public_key` and not here

signer/src/transaction_coordinator.rs
```diff
@@ -1860,15 +1860,20 @@ where
...
     pub fn is_coordinator(&self, bitcoin_chain_tip: &model::BitcoinBlockHash) -> bool {
-        given_key_is_coordinator(
-            self.signer_public_key(),
-            bitcoin_chain_tip,
-            &self.context.config().signer.bootstrap_signing_set,
-        )
+        let default_signer_set = || self.context.config().signer.bootstrap_signing_set.clone();
```
MCJOHN974: nit: this `clone` is not 100% necessary but from the top of my mind I don't see a non-ugly way to remove it

djordon: Yeah, there is a not-so-ugly way to remove it. We can assign the object returned from `registry_signer_set_info` to a variable and then using `as_ref` on the variable to be able to take a reference to the `signer_set` field.

signer/src/transaction_coordinator.rs
djordon: What do you mean?

MCJOHN974: My comment is outdated

signer/src/transaction_coordinator.rs
```diff
@@ -2858,6 +2858,66 @@ mod tests {
             .unwrap();
     }
 
+    /// Check that we is_coordinator uses the registry for figuring out who
```
matteojug: nit
```suggestion
    /// Check that is_coordinator uses the registry for figuring out who
```

signer/src/transaction_coordinator.rs
```diff
@@ -1731,7 +1731,7 @@ where
         S: Stream<Item = Signed<SignerMessage>>,
         Coordinator: WstsCoordinator,
     {
-        let signer_set = self.context.config().signer.bootstrap_signing_set.clone();
+        let signer_set = self.context.coordinator_signer_set();
```
matteojug, line 1739: nit: can we call this `coordinator_signer_set` or similar (or add a comment)? To make it more clear is not related to the wsts signer set

signer/src/transaction_signer.rs
```diff
@@ -2003,4 +2009,198 @@ mod tests {
...
+            .iter()
+            .map(|kp| kp.public_key().into())
+            .collect();
+
+        // Create registry signer set: `overlap` keys from the bootstrap
+        // key pairs, adding + (3 - overlap) new keys
+        let num_new_keys = bootstrap_keypairs.len() - overlap;
+        let registry_keypairs = std::iter::repeat_n(Keypair::new_global(&mut rng), num_new_keys)
+            .chain(bootstrap_keypairs.iter().take(overlap).cloned())
+            .collect::<Vec<Keypair>>();
```
matteojug: Isn't this going to add the same key `num_new_keys` times? If so, `registry_signer_set` below is going to be less than 3 keys for `overlap` 0 and 1.

matteojug (dismissed):
LGTM

signer/src/transaction_signer.rs
djordon: Oh right, yeah it will

MCJOHN974 (dismissed):

signer/src/transaction_coordinator.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1912/commits/42f6e96e8a820e96d1163fea4d691db072a600d9

signer/src/transaction_coordinator.rs
djordon: I chose to add a comment, and did so in https://github.com/stacks-sbtc/sbtc/pull/1912/commits/42f6e96e8a820e96d1163fea4d691db072a600d9

signer/src/transaction_signer.rs
djordon: Fixed this in https://github.com/stacks-sbtc/sbtc/pull/1912/commits/84ea9b67ebba494b55f0bd8be5498ef8e1ee55b9

PR #1913, opened 2026-02-05
tests: add signer set changes test v2

### LINKED ISSUE #1700: [Chore]: Integration test for increasing signer set from `x` to `x+1`, without threshold change
# Chore - Integration test for increasing signer set from `x` to `x+1`, without threshold change

## 1. Description

We need a new integration test which simulates growing the signer set from `x` to `x+1` to ensure that this process works correctly. This test does not need to handle threshold changes, which is currently not supported.

## 2. Why It's Necessary

This is to help ensure the process of growing the signer set in preparation for adding a new signer.

djordon:
## Description

Closes: https://github.com/stacks-sbtc/sbtc/issues/1700

This basically the same stuff in https://github.com/stacks-sbtc/sbtc/pull/1765.

## Changes

* Send the chain tip block ref upon tenure completion. This is to make testing a little better.
* Add integration tests that tests the behavior whenever we change the size of the signer set.
* Add integration tests that tests the behavior whenever we change the size of the signature threshold.

## Checklist


signer/tests/integration/transaction_coordinator.rs
```diff
@@ -669,6 +731,263 @@ async fn mock_stacks_core<D, B, E>(
     .await;
 }
 
+async fn mock_stacks_core2<D, B, E>(
```
djordon, line 723: Not the best name, open to suggestions.

MCJOHN974: Maybe for better readability you can refactor it a bit such that you will use it as:
```
let ctx = mock_stacks_core(ctx)
    .with_anchor_block(chain_tip_info)
    .with_signer_set(signer_set_data)
    .with_sender(broadcast_stacks_tx)
```
and probably you don't really need to have two functions

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -669,6 +733,269 @@ async fn mock_stacks_core<D, B, E>(
...
+    });
+    handles.push(block_observer_handle.abort_handle());
+
+    // We wait to make sure that all spawned tasks have started.
+    main_barrier.wait().await;
+
+    handles
+}
+
+/// What for key rotation contract call transactions, extract the function
```
MCJOHN974:
```suggestion
/// Wait for key rotation contract call transactions, extract the function
```

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -669,6 +733,269 @@ async fn mock_stacks_core<D, B, E>(
...
+/// arguments, and update the signer set data with the details.
+///
+/// # Notes
+///
+/// This function can be used to emulate how signer set info is propogated
```
MCJOHN974:
```suggestion
/// This function can be used to emulate how signer set info is propagated
```

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -669,6 +733,269 @@ async fn mock_stacks_core<D, B, E>(
...
+/// throughout the stacks network. This is necessary so that new signers
+/// have access to the signer set info that the their stacks node will know
+/// about.
```
MCJOHN974:
```suggestion
/// have access to the signer set info known by their Stacks node
```

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -2801,6 +3128,1085 @@ async fn sign_bitcoin_transaction_multiple_locking_keys() {
...
+///
+/// To start the test environment do:
+/// ```bash
+/// make integration-env-up-ci
+/// ```
+///
+/// then, once everything is up and running, run the test.
+#[test_case(TestThresholds {
+    first: NonZeroU16::new(2).unwrap(),
+    second: NonZeroU16::new(3).unwrap(),
```
MCJOHN974: tbh don't really like our `NonZeroU*` usage because of all this casts, however nothing to do about it in our PR

signer/tests/integration/transaction_coordinator.rs
```diff
@@ -2801,6 +3128,1085 @@ async fn sign_bitcoin_transaction_multiple_locking_keys() {
...
+            .with_emily_client(emily_client.clone())
+            .with_mocked_stacks_client()
+            .modify_settings(|settings| {
+                settings.signer.private_key = ctx_old.config().signer.private_key;
+                settings.signer.bootstrap_signatures_required = thresholds.second.get();
+                settings.signer.bitcoin_processing_delay = Duration::from_millis(200);
+            })
+            .build();
+
+        backfill_bitcoin_blocks_unchecked(&db, rpc, &chain_tip_info.hash).await;
```
MCJOHN974, line 3357: Why you need unchecked here and above?

djordon: I want to test what happens when the new signer's database is only kinda updated, it doesn't necessarily have all bitcoin blocks in it's database compared to the existing signer set. So if we were to use the checked variant, then it would panic for the existing signers, who have more blocks in their database. In this case, `chain_tip_info` is not for the actual chain tip, it's kinda stale, so the check would fail for existing signers. 

Also, when we don't need the unchecked variant in the beginning of the test, either variant works there.

signer/tests/integration/transaction_coordinator.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1913/commits/27baf16e4dc82c3381acf441698f33dca81001d9

signer/tests/integration/transaction_coordinator.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1913/commits/27baf16e4dc82c3381acf441698f33dca81001d9

signer/tests/integration/transaction_coordinator.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1913/commits/27baf16e4dc82c3381acf441698f33dca81001d9

signer/tests/integration/transaction_coordinator.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1913/commits/27baf16e4dc82c3381acf441698f33dca81001d9

djordon:
Well, `sign_bitcoin_transaction_withdrawals` failed. I'll wait for https://github.com/stacks-sbtc/sbtc/pull/1942 to land, update this branch with those changes and that should sort things out.

MCJOHN974 (approved):

MCJOHN974 (commented):
whoops, approve above was misclick

djordon-stacks:
Closing since we haven't seen much action pushing it through, and we are unlikely to see that anytime soon. The new tests are flaky and need work. Plus, we can always reopen it later.

Issue #1917, opened 2026-02-06
[Feature]: Enable quick Emily sBTC slowdown 


MCJOHN974:
# Feature - Enable quick Emily sBTC slowdown

## 1. Description

We want to add an ability for builders to reduce sBTC withdrawal limits in case of emergency, without directly contacting sBTC team, capping amount of sBTC available to withdraw.

### 1.1 Context & Purpose

This measure may be helpful if some of DeFi protocols on Stacks will be attacked, by preventing attackers immediately withdrawing everything to bitcoin

## 2. Technical Details

Builders get an emergency key from us. With this key, they can activate sBTC "slow mode" by posting their key to a new Emily endpoint, which sets Emily limits to a pre-defined (by us) settings. Once activated, it can be deactivated by us, overriding slow mode with new limits update.

### 2.1 Acceptance Criteria

We plan to do following for this:

- [ ] Add new DynamoDB table for builders keys (`builder_name - partition_key`, `hash_of_secret`, `is_active`)
- [ ] Add private endpoints for adding and (de)activating keys.
- [ ] Add public endpoint for slow mode activation (with builders secret as parameter)
- [ ] Make sure we have alarm on slow mode activation
- [ ] Finalize our decision on slow mode limits values
- [ ] Update `Limits` structure to include information about if slow mode is active and who activated it
- [ ]

PR #1918, opened 2026-02-09
feat: Enable quick Emily sBTC slowdown.

MCJOHN974:
## Description

Closes: #1917 

## Changes

- adds new table to DynamoDB
- adds private endpoints to manage throttle keys
- adds public endpoint to start throttle mode

## Testing Information

- adds some tests for new endpoints in the `emily/handler/tests/integration/throttle.rs`

## Checklist

### DIFF
--- a/emily/handler/src/database/accessors.rs
+++ b/emily/handler/src/database/accessors.rs
@@ -976,7 +976,7 @@ pub async fn get_throttle_keys_list(
     todo!()
 }
 
-/// Get throttle key by key name.
+/// Get throttle key by key hash.
 pub async fn get_throttle_key(
     context: &EmilyContext,
     hash: &String,


emily/cdk/lib/emily-stack.ts
```diff
@@ -77,12 +77,22 @@ export class EmilyStack extends cdk.Stack {
             pointInTimeRecovery,
         );
 
+        const slowdownTableId: string = 'SlowdownTable';
```
djordon: Can we change the name from "slowdown*" to "throttle*" or "throttler*"? They mean roughly the same thing, jut one feels more correct.

emily/cdk/lib/emily-stack.ts
```diff
@@ -359,6 +370,37 @@ export class EmilyStack extends cdk.Stack {
...
+        const table =  new dynamodb.Table(this, tableId, {
+            tableName: tableName,
+            partitionKey: {
+                name: 'KeyName',
+                type: dynamodb.AttributeType.STRING,
+            },
+            sortKey: {
+                name: 'Zero',
+                type: dynamodb.AttributeType.NUMBER,
+            },
```
djordon, line 396: Can we make this sort key a `createdAt` attribute

emily/cdk/lib/emily-stack.ts
```diff
@@ -359,6 +370,37 @@ export class EmilyStack extends cdk.Stack {
...
+        tableName: string,
+        removalPolicy: cdk.RemovalPolicy,
+        pointInTimeRecovery: undefined | boolean,
+    ): dynamodb.Table {
+        // Create DynamoDB table to store the messages. Encrypted by default.
+        const table =  new dynamodb.Table(this, tableId, {
+            tableName: tableName,
+            partitionKey: {
+                name: 'KeyName',
+                type: dynamodb.AttributeType.STRING,
```
djordon: How about using the secret hash as the partition key. That's basically the primary key here, and it should simplify things slightly.

emily/handler/src/api/handlers/slowdown.rs
```diff
@@ -0,0 +1,266 @@
...
+                tracing::warn!(
+                    key_name = %request.name,
+                    "Attempt to start slow mode failed key verification",
+                );
+                Err(Error::Unauthorized)
+            }
+            KeyVerificationResult::Eligible => {
+                // It is not actually an error, but we want to take maximum attention if it happens.
+                // TODO: we need an alarm on this error.
+                tracing::error!(
```
djordon: This should be an INFO log

emily/handler/src/api/handlers/slowdown.rs
```diff
@@ -0,0 +1,266 @@
...
+                tracing::info!(?new_limits, "Calculated limits to use in slow mode",);
+                let res = crate::api::handlers::limits::set_limits(new_limits.clone(), context)
+                    .await
+                    .into_response();
+                let is_success = res.status().is_success();
+                if is_success {
+                    tracing::info!("successfully started slow mode.");
+                    return Ok(with_status(json(&new_limits), StatusCode::OK));
+                }
+                tracing::error!(?res, "Error setting slow mode limits");
```
djordon: This should also be an `INFO`

emily/handler/src/api/handlers/slowdown.rs
```diff
@@ -0,0 +1,266 @@
...
+    name: String,
+    context: EmilyContext,
+) -> impl warp::reply::Reply {
+    // Internal handler so `?` can be used correctly while still returning a reply.
+    async fn handler(
+        context: EmilyContext,
+        name: String,
+    ) -> Result<impl warp::reply::Reply, Error> {
+        accessors::deactivate_slowdown_key(&context, name).await?;
+        Ok(with_status(json(&()), StatusCode::OK))
```
djordon, line 245: Not for now, but we should return the row in the database of the deactivated key.

emily/handler/src/api/handlers/slowdown.rs
```diff
@@ -0,0 +1,266 @@
...
+)]
+#[instrument(skip(context))]
+pub async fn activate_slowdown_key(name: String, context: EmilyContext) -> impl warp::reply::Reply {
+    // Internal handler so `?` can be used correctly while still returning a reply.
+    async fn handler(
+        context: EmilyContext,
+        name: String,
+    ) -> Result<impl warp::reply::Reply, Error> {
+        accessors::activate_slowdown_key(&context, name).await?;
+        Ok(with_status(json(&()), StatusCode::OK))
```
djordon, line 277: Also here, we should return the row after activation. But yeah, not for this PR

djordon (commented):
We spoke about a lot of these things already, but yeah to summarize:
1. Name change to `throttle` or `throttler`, something like that.
2. Use the hash as the partition key
3. Change the sort key to a created at unix timestamp.

And some other small things. Overall, looks good.

emily/handler/src/api/handlers/slowdown.rs
MCJOHN974: Are you 100% sure? We print this log line if `set_limits` failed, and from the top of my mind I can't imagine very casual situation in which `set_limits` fails, so I'd put maybe not error but warn here. 

Also it would make it more readable to make `if !is_success` not `if is_success`

djordon: Oh I was wrong. I missed that this was actually the failure path.

emily/handler/src/api/handlers/slowdown.rs
MCJOHN974: [32e1e80](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/32e1e807caa3d58aab3e2d0b9348dd980c669de6)

emily/cdk/lib/emily-stack.ts
MCJOHN974: [e49660f](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/e49660f188c83f2a9ef7571c14ab22692b41881b) and [c5908f2](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/c5908f228c68390cbd56dc6ea799ee7fb32c7c4c)

emily/cdk/lib/emily-stack.ts
MCJOHN974: [e49660f](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/e49660f188c83f2a9ef7571c14ab22692b41881b) and [c5908f2](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/c5908f228c68390cbd56dc6ea799ee7fb32c7c4c)

emily/cdk/lib/emily-stack.ts
MCJOHN974: [d513b6e](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/d513b6e29f3ca6cd7e310460f30bfe3d40e2e494) and two next commits do this

emily/handler/src/api/models/throttle.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// The hash of the secret associated with this key.
+    pub hash: String,
+}
+
+/// Represents the throttle reqwest
+#[derive(Clone, Default, Debug, Eq, PartialEq, Serialize, Deserialize, ToSchema, ToResponse)]
+#[serde(rename_all = "camelCase")]
+pub struct ThrottleReqwest {
+    /// The name of the key.
+    pub hash: String,
```
djordon: nit: the comment should be updated.

emily/openapi-gen/generators/private.rs
```diff
@@ -39,6 +39,12 @@ use super::CorsSupport;
         api::handlers::limits::set_limits,
         api::handlers::limits::get_limits_for_account,
         api::handlers::limits::set_limits_for_account,
+        // Throttle endpoints
+        api::handlers::throttle::add_throttle_key,
+        api::handlers::throttle::get_throttle_key,
+        api::handlers::throttle::start_throttle,
```
djordon, line 45: Shouldn't this one be in public?

emily/handler/src/context.rs
```diff
@@ -93,6 +96,7 @@ impl Settings {
             withdrawal_table_name: env::var("WITHDRAWAL_TABLE_NAME")?,
             chainstate_table_name: env::var("CHAINSTATE_TABLE_NAME")?,
             limit_table_name: env::var("LIMIT_TABLE_NAME")?,
+            throttle_table_name: env::var("THROTTLEDOWN_TABLE_NAME")?,
```
djordon, line 99: Note to self, we'll need to make sure that we have this new environment variable during our production deployment.

emily/handler/src/api/models/throttle.rs
```diff
@@ -0,0 +1,24 @@
+//! Request structures for limits api calls.
```
matteojug: nit
```suggestion
//! Request structures for throttle api calls.
```

emily/handler/src/database/accessors.rs
```diff
@@ -946,6 +953,149 @@ pub async fn set_limit_for_account(
...
+    secret: &String,
+) -> Result<KeyVerificationResult, Error> {
+    let key = get_throttle_key(context, hash).await?;
+    if &key.key.hash != hash {
+        return Err(Error::NotFound);
+    }
+    if !key.is_active {
+        return Ok(KeyVerificationResult::Revoked);
+    }
+    let parsed_hash = PasswordHash::new(&key.key.hash).expect("Invalid hash format");
```
djordon: Why expect here? Let's bubble up the error

djordon: Just spoke about this out of band and we should probably return a 500 in this case.

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,270 @@
...
+            .per_withdrawal_cap
+            .map_or(THROTTLE_MODE_PER_WITHDRAWAL_CAP, |curr| {
+                curr.min(THROTTLE_MODE_PER_WITHDRAWAL_CAP)
+            }),
+    );
+    limits.rolling_withdrawal_blocks = Some(
+        limits
+            .rolling_withdrawal_blocks
+            .map_or(THROTTLE_MODE_ROLLING_WINDOW, |curr| {
+                curr.max(THROTTLE_MODE_ROLLING_WINDOW)
```
matteojug, line 79: While the `max` is correct here to keep the tightest limit, I'm not sure if we want to do it, as this will almost always be won by the "normal" limit (eg, currently is 150). Maybe for this one we can just use the throttle one?

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,270 @@
...
+    Ok(limits)
+}
+
+/// Try to turn on throttle mode
+#[utoipa::path(
+    post,
+    operation_id = "startThrottle",
+    path = "/start_throttle",
+    tag = "throttle",
+    request_body = ThrottleReqwest,
```
matteojug: nit
```suggestion
    request_body = ThrottleRequest,
```

emily/handler/src/database/accessors.rs
MCJOHN974: [7f075ed](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/7f075ed6ecee5751e9bbfa1e3a2bc94f5fc1697a)

emily/openapi-gen/generators/private.rs
MCJOHN974: [adf33b7](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/adf33b75db97ba11f81789c0332930ddabd3aab7)

emily/handler/src/api/models/throttle.rs
MCJOHN974: [44487a7](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/44487a7d2276668f465f40918e5ee7f0b8fbdc94)

emily/handler/src/api/models/throttle.rs
```diff
@@ -0,0 +1,24 @@
...
+}
+
+/// Represents the throttle reqwest
+#[derive(Clone, Default, Debug, Eq, PartialEq, Serialize, Deserialize, ToSchema, ToResponse)]
+#[serde(rename_all = "camelCase")]
+pub struct ThrottleReqwest {
+    /// The hash of the key.
+    pub hash: String,
+    /// The secret associated with this key.
+    pub secret: String,
```
matteojug: As UX it's a bit strange that you need to provide both the secret and the hash, as we are using the hash as a username and the secret as password; while `secret` could be something that looks like an API key, the `hash` is something internal. It's also a bit strange that we have to supply the hash when creating a `ThrottleKey`, as we need to match the internal format.

What about:
 - When creating a new key we ask for the secret on key creation and hash it server side
 - When activating the throttle we ask for name and secret, instead of hash?
So that both requests have basically the same body: `name` and `secret`.

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,270 @@
...
+                // TODO: we need an alarm on this error.
+                tracing::info!(
+                    key_hash = %request.hash,
+                    "Successfull request to start throttle mode. Starting throttle mode.",
+                );
+                let new_limits = calculate_throttle_mode_limits(&context, initiator).await?;
+                tracing::info!(?new_limits, "Calculated limits to use in throttle mode",);
+                let res = crate::api::handlers::limits::set_limits(new_limits.clone(), context)
+                    .await
+                    .into_response();
```
matteojug, line 141: Overwriting the limits make it more annoying for us to restore everything, as well as making not possible for the initiator themselves to lift the throttle.
What about we only store a flag -- throttling or not -- for each key and, on the limits read path, we return the throttled limits if we anyone is throttling, or the usual ones if not?

emily/handler/src/context.rs
matteojug: Isn't that done by https://github.com/stacks-sbtc/sbtc/pull/1918/changes#diff-ce74baab94e8367239344f3d267396c13a79b37ca9df5ee2cbbe5e57cd0e5f9eR472 for prod?

emily/handler/tests/integration/throttle.rs
```diff
@@ -0,0 +1,643 @@
...
+    );
+    assert_eq!(
+        new_limits_after_throttle
+            .rolling_withdrawal_cap
+            .unwrap()
+            .unwrap(),
+        emily_handler::api::handlers::throttle::THROTTLE_MODE_ROLLING_CAP
+    );
+
+    // Now lets restore limits back to normal
```
djordon, line 119: Can we set that deposits are unaffected?

djordon (approved):
Looks good, one small nit in the tests.

emily/handler/src/context.rs
djordon: Oh yes, I think so. Nevermind 😄

emily/handler/src/api/handlers/throttle.rs
djordon: Hmmm, why? What do you have in mind?

emily/handler/src/api/models/throttle.rs
djordon: Yeah good point. I think I asked for this in the call too.

emily/handler/tests/integration/throttle.rs
MCJOHN974: You mean add assert that all unrelated limits was untouched??

MCJOHN974: If yes then [ce9ed38](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/ce9ed38043ea3d463dfe36ab513d7e5b07fcfbe4)

emily/handler/src/api/handlers/throttle.rs
MCJOHN974: I'd agree that it would make things better but I'd outscope it from this pr since deadline is today

emily/handler/src/api/models/throttle.rs
djordon: > Yeah good point. I think I asked for this in the call too.

Sorry for the half baked comment. The issue was that argon2 adds a unique and random salt to the secret, so the secret alone does not map to the hash. One alternative approach is to force the salt to be some value, and use a different hashing strategy.

emily/handler/tests/integration/throttle.rs
djordon: That's exactly what I had in mind, 🙏🏽 thank you

emily/handler/src/api/handlers/throttle.rs
matteojug: In normal conditions we have the rolling window at 150; when we enable throttling with its rolling window set at 18, with the current impl we will end up setting the rolling window to 150 (because of the max), meaning that we will have a rolling window of 2BTC per day, instead of per 3 hour window.

This seems "correct" as we are enforcing the strongest limit, but effectively we will not use the throttled time window, that's why I was wondering if here we should just use `THROTTLE_MODE_ROLLING_WINDOW` regardless of the previous time window.

MCJOHN974: I'd stick with using `.max()` here, and just remember this while we do decision making on what actual throttle limits we want. I suspect we may end up having throttle limits very small and thus it dont really matter that much

emily/handler/src/api/models/throttle.rs
matteojug: Yeah, the strange thing is that we are treating the hash as username. If we ask for the `name` on the `start_throttling` path, we can fetch the key using that and then verify the secret.

MCJOHN974: I think it may mean that we need to have `name` back as partition key. Otherwise I don't really see, how we can extract entry from the DB based only on `secret` and `name`.

emily/handler/src/api/handlers/throttle.rs
MCJOHN974: [169871e](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/169871eb017637570411e51057fe6d949e1bd022)

signer/src/error.rs
```diff
@@ -429,9 +429,9 @@ pub enum Error {
     #[error("could not decode the hex string into bytes: {0}")]
     DecodeHexBytes(#[source] hex::FromHexError),
 
-    /// Reqwest error
+    /// Request error
     #[error(transparent)]
-    Reqwest(#[from] reqwest::Error),
+    Request(#[from] reqwest::Error),
```
matteojug: nit: this should be reverted

signer/src/error.rs
```diff
@@ -505,7 +505,7 @@ pub enum Error {
     #[error("coordinator Stacks txn with fee too high: {0}. Highest acceptable fee: {1}")]
     StacksFeeLimitExceeded(u64, u64),
 
-    /// Reqwest error
+    /// Request error
```
matteojug: nit: this should be reverted

signer/src/error.rs
```diff
@@ -712,7 +712,7 @@ pub enum Error {
 
     /// Could not create reqwest client
     #[error("we received an error when creating the Emily's reqwest client: {0}")]
-    EmilyReqwestClientCreation(#[source] reqwest_012::Error),
+    EmilyRequestClientCreation(#[source] reqwest_012::Error),
```
matteojug: nit: this should be reverted

signers.sh
```diff
@@ -21,7 +21,7 @@ LOG_SETTINGS="debug" # Default log level
 LOG_SETTINGS="$LOG_SETTINGS,signer::stacks::api=info" # Stacks API
 LOG_SETTINGS="$LOG_SETTINGS,hyper=info" # Hyper
 LOG_SETTINGS="$LOG_SETTINGS,sqlx=info" # SQLx
-LOG_SETTINGS="$LOG_SETTINGS,reqwest=info" # Reqwest
+LOG_SETTINGS="$LOG_SETTINGS,reqwest=info" # Request
```
matteojug: nit: this should be reverted

signer/src/emily_client.rs
```diff
@@ -179,7 +179,7 @@ impl EmilyClient {
         let client = reqwest_012::Client::builder()
             .timeout(timeout)
             .build()
-            .map_err(Error::EmilyReqwestClientCreation)?;
+            .map_err(Error::EmilyRequestClientCreation)?;
```
matteojug: nit: this should be reverted

emily/handler/src/api/models/throttle.rs
MCJOHN974: [03a0ead](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/03a0eadf5c0392ae07c00561e2afa4d1eb7b7d0e)

emily/handler/src/api/handlers/throttle.rs
matteojug: We can leave this as is and reconsider later

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,277 @@
+//! Handlers for limits endpoints.
```
matteojug, line 1: 🙈 
```suggestion
//! Handlers for throttle endpoints.
```

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,277 @@
...
+    responses(
+        (status = 200, description = "Throttle key retrieved successfully", body = GetThrottleKeyResponse),
+        (status = 404, description = "Throttle key not found", body = ErrorResponse),
+        (status = 405, description = "Method not allowed", body = ErrorResponse),
+        (status = 500, description = "Internal server error", body = ErrorResponse)
+    ),
+    security(("ApiGatewayKey" = []))
+)]
+#[instrument(skip(context))]
+pub async fn get_throttle_key(hash: String, context: EmilyContext) -> impl warp::reply::Reply {
```
matteojug, line 34: We can address this later, but this should probably fetch the key by name, not by hash -- in general, the hash shouldn't need to be exposed outside since it's an internal thingy

emily/handler/src/api/models/throttle.rs
```diff
@@ -0,0 +1,36 @@
...
+#[derive(Clone, Default, Debug, Eq, PartialEq, Serialize, Deserialize, ToSchema, ToResponse)]
+#[serde(rename_all = "camelCase")]
+pub struct ThrottleKey {
+    /// The name of the key.
+    pub name: String,
+    /// The secret of the secret associated with this key.
+    pub secret: String,
+}
+
+/// Represents the throttle reqwest
```
matteojug, line 16:
```suggestion
/// Represents the throttle request
```

emily/handler/src/api/models/throttle.rs
```diff
@@ -0,0 +1,36 @@
...
+use serde::{Deserialize, Serialize};
+use utoipa::{ToResponse, ToSchema};
+
+/// Represents the throttle key
+#[derive(Clone, Default, Debug, Eq, PartialEq, Serialize, Deserialize, ToSchema, ToResponse)]
+#[serde(rename_all = "camelCase")]
+pub struct ThrottleKey {
+    /// The name of the key.
+    pub name: String,
+    /// The secret of the secret associated with this key.
```
matteojug, line 12:
```suggestion
    /// The secret associated with this key.
```

emily/handler/src/api/models/throttle.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/// Response to get_throttle_key endpoint, with information about the key
+#[derive(Clone, Default, Debug, Eq, PartialEq, Serialize, Deserialize, ToSchema, ToResponse)]
+#[serde(rename_all = "camelCase")]
+pub struct GetThrottleKeyResponse {
+    /// The name of the key.
+    pub name: String,
+    /// The hash of the key.
+    pub hash: String,
```
matteojug, line 33: For later, we shouldn't need to expose the `hash`

signer/src/emily_client.rs
MCJOHN974: [94bd0f4](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/94bd0f48c4015dfa9c4c5059a59d281972d53909)

signers.sh
MCJOHN974: [94bd0f4](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/94bd0f48c4015dfa9c4c5059a59d281972d53909)

signer/src/error.rs
MCJOHN974: [94bd0f4](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/94bd0f48c4015dfa9c4c5059a59d281972d53909)

signer/src/error.rs
MCJOHN974: [94bd0f4](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/94bd0f48c4015dfa9c4c5059a59d281972d53909)

signer/src/error.rs
MCJOHN974: [94bd0f4](https://github.com/stacks-sbtc/sbtc/pull/1918/commits/94bd0f48c4015dfa9c4c5059a59d281972d53909)

emily/handler/src/api/routes/throttle.rs
```diff
@@ -0,0 +1,98 @@
...
+
+/// Endpoint to activate existing key.
+fn activate_throttle_key<F>(
+    context: F,
+) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone
+where
+    F: Filter<Extract = (EmilyContext,), Error = std::convert::Infallible> + Clone + Send,
+{
+    warp::path!("throttle" / "activate")
+        .and(warp::patch())
```
matteojug, line 80: maybe nit: use `post` here and below?
We don't use `patch` around; I'm not saying we shouldn't, just we are not using, feel free to ignore this

djordon (approved):

emily/handler/src/database/entries/throttle.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub hash: String,
+    /// The timestamp of key creation, in seconds from UNIX epoch
+    pub created_at: u64,
+}
+
+/// Limit table entry key. This is the primary index key.
+#[derive(Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct ThrottleKeyEntry {
+    /// Limit entry key.
```
matteojug, line 23:
```suggestion
    /// Throttle entry key.
```

emily/handler/src/database/entries/throttle.rs
```diff
@@ -0,0 +1,64 @@
+//! Entries into the throttle keys table.
+
+use std::hash::Hash;
+
+use serde::{Deserialize, Serialize};
+
+use super::{EntryTrait, KeyTrait, PrimaryIndex, PrimaryIndexTrait};
+
+/// Limit table entry key. This is the primary index key.
```
matteojug, line 9:
```suggestion
/// Throttle table entry key. This is the primary index key.
```

emily/handler/src/database/entries/throttle.rs
```diff
@@ -0,0 +1,64 @@
...
+#[derive(Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct ThrottleKeyEntryKey {
+    /// Hash of the secret
+    pub hash: String,
+    /// The timestamp of key creation, in seconds from UNIX epoch
+    pub created_at: u64,
+}
+
+/// Limit table entry key. This is the primary index key.
```
matteojug, line 19:
```suggestion
/// Throttle table entry.
```

emily/handler/src/database/entries/throttle.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Limit entry key.
+    #[serde(flatten)]
+    pub key: ThrottleKeyEntryKey,
+    /// If the key is eligible to start throttle mode.
+    pub is_active: bool,
+    /// Name of the key.
+    pub name: String,
+}
+
+/// Implements the key trait for the deposit entry key.
```
matteojug, line 32:
```suggestion
/// Implements the key trait for the throttle entry key.
```

emily/handler/src/database/entries/throttle.rs
```diff
@@ -0,0 +1,64 @@
...
+    type PartitionKey = String;
+    /// the type of the sort key.
+    type SortKey = u64;
+    /// The table field name of the partition key.
+    const PARTITION_KEY_NAME: &'static str = "Hash";
+    /// The table field name of the sort key.
+    const SORT_KEY_NAME: &'static str = "CreatedAt";
+}
+
+/// Implements the entry trait for the deposit entry.
```
matteojug, line 44:
```suggestion
/// Implements the entry trait for the throttle entry.
```

emily/handler/src/database/entries/throttle.rs
```diff
@@ -0,0 +1,64 @@
...
+impl EntryTrait for ThrottleKeyEntry {
+    /// The type of the key for this entry type.
+    type Key = ThrottleKeyEntryKey;
+    /// Extract the key from the deposit entry.
```
matteojug, line 48:
```suggestion
    /// Extract the key from the throttle entry.
```

emily/handler/src/database/entries/throttle.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn key(&self) -> Self::Key {
+        self.key.clone()
+    }
+}
+
+/// Primary index struct.
+pub struct ThrottleTablePrimaryIndexInner;
+/// Withdrawal table primary index type.
```
matteojug, line 56:
```suggestion
/// Throttle table primary index type.
```

emily/handler/src/database/accessors.rs
```diff
@@ -45,6 +44,26 @@ use super::entries::{
...
+
+use sha2::{Digest as _, Sha256};
+
+/// Converts a string to an argon2 salt.
+pub fn name_to_salt(name: &str) -> Result<SaltString, Error> {
+    let mut hasher = Sha256::new();
+    hasher.update(name);
+    let result = hasher.finalize();
+    let hash_string = format!("{:x}", result);
+    SaltString::from_b64(&hash_string)
```
matteojug, line 63: This works because hex is a subset (charset wise) of base64, but we should do the right thing here for our sanity (in another PR)

emily/handler/src/database/accessors.rs
```diff
@@ -946,6 +967,128 @@ pub async fn set_limit_for_account(
...
+// Throttle keys ---------------------------------------------------------------
+
+/// Get list of all throttle keys.
+pub async fn get_throttle_keys_list(
+    _context: &EmilyContext,
+) -> Result<Vec<ThrottleKeyEntry>, Error> {
+    todo!()
+}
+
+/// Get throttle key by key name.
```
matteojug:
```suggestion
/// Get throttle key by key hash.
```

emily/handler/src/database/accessors.rs
```diff
@@ -946,6 +967,128 @@ pub async fn set_limit_for_account(
...
+    name: &str,
+    secret: &String,
+) -> Result<KeyVerificationResult, Error> {
+    // We use name as salt. It is fine, because we choose names and
+    // we have a guarantee that they are unique, so we won't have a collision on the salt.
+    let argon2 = Argon2::default();
+    let salt = name_to_salt(name)?;
+    let hash = argon2
+        .hash_password(secret.as_bytes(), &salt)
+        .unwrap()
```
matteojug: can we avoid panicking here?

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,283 @@
...
+        let hash = argon2
+            .hash_password(key.secret.as_bytes(), &salt)
+            .inspect_err(|error| {
+                tracing::error!(
+                    ?error,
+                    name = %key.name,
+                    "Error hashing the secret. Usually happens due to failed conversion of name into salt",
+                );
+            })
+            .map_err(|_| Error::Deserialization("Error hashing the secret. Usually happens due to failed conversion of name into salt".to_string()))?
```
djordon, line 195: nit: we should us a different error variant.

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,283 @@
...
+        key: ThrottleKey,
+    ) -> Result<impl warp::reply::Reply, Error> {
+        let argon2 = Argon2::default();
+        let salt = accessors::name_to_salt(&key.name)?;
+        let hash = argon2
+            .hash_password(key.secret.as_bytes(), &salt)
+            .inspect_err(|error| {
+                tracing::error!(
+                    ?error,
+                    name = %key.name,
```
djordon, line 191: nit: you don't need to put the name in this log, it's already there.

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,283 @@
...
+        context: EmilyContext,
+        key: ThrottleKey,
+    ) -> Result<impl warp::reply::Reply, Error> {
+        let argon2 = Argon2::default();
+        let salt = accessors::name_to_salt(&key.name)?;
+        let hash = argon2
+            .hash_password(key.secret.as_bytes(), &salt)
+            .inspect_err(|error| {
+                tracing::error!(
+                    ?error,
```
djordon, line 190: nit: we should use the display impl here instead of the debug one.

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,283 @@
...
+
+/// Add throttle key handler.
+#[utoipa::path(
+    post,
+    operation_id = "addThrottleKey",
+    path = "/throttle",
+    tag = "throttle",
+    request_body = ThrottleKey,
+    responses(
+        (status = 200, description = "Throttle key added successfully", body = ()),
```
djordon, line 166: nit: we return a 201, not a 200.

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,283 @@
...
+pub async fn add_throttle_key(key: ThrottleKey, context: EmilyContext) -> impl warp::reply::Reply {
+    // Internal handler so `?` can be used correctly while still returning a reply.
+    async fn handler(
+        context: EmilyContext,
+        key: ThrottleKey,
+    ) -> Result<impl warp::reply::Reply, Error> {
+        let argon2 = Argon2::default();
+        let salt = accessors::name_to_salt(&key.name)?;
+        let hash = argon2
+            .hash_password(key.secret.as_bytes(), &salt)
```
matteojug, line 187: For later, we should factor out this to be sure we are using the same here and in `verify_throttle_key`

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,283 @@
...
+
+/// Deactivate existing throttle key
+#[utoipa::path(
+    patch,
+    operation_id = "deactivateThrottleKey",
+    path = "/throttle/deactivate",
+    tag = "throttle",
+    request_body = String,
+    responses(
+        (status = 201, description = "Throttle key deactivated successfully", body = ()),
```
djordon, line 226: This function returns a 200

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -0,0 +1,283 @@
...
+
+/// Activate existing (previously deactivated) throttle key
+#[utoipa::path(
+    patch,
+    operation_id = "activateThrottleKey",
+    path = "/throttle/activate",
+    tag = "throttle",
+    request_body = String,
+    responses(
+        (status = 201, description = "Throttle key activated successfully", body = ()),
```
djordon, line 261: This will also return a 200

matteojug (dismissed):
A bunch of nits/do it later

PR #1920, opened 2026-02-13
feat: simple turn off for throttle mode

MCJOHN974:
## Description

This PR allows just send an empty request to new endpoint to turn of throttle mode and return to previous limits instead of sending full `POST limits` with all limits provided

Closes: #1943

## Changes
- adds new private endpoint `stop throttle` which don't take any arguments and restores limits back to pre-throttle values
- don't stores throttle limits in db anymore

## Testing Information
few new integration tests added

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1750,6 +1750,7 @@ dependencies = [
  "aws-sdk-dynamodb",
  "base64 0.22.1",
  "bitcoin",
+ "chrono",
  "clap",
  "clarity",
  "config",
--- a/emily/handler/Cargo.toml
+++ b/emily/handler/Cargo.toml
@@ -26,6 +26,7 @@ aws-config.workspace = true
 aws-sdk-dynamodb.workspace = true
 base64.workspace = true
 bitcoin.workspace = true
+chrono.workspace = true
 clap.workspace = true
 clarity.workspace = true
 config.workspace = true
--- a/emily/handler/src/context.rs
+++ b/emily/handler/src/context.rs
@@ -18,7 +18,6 @@ use serde::Serialize;
 use crate::database::entries::limits::LimitEntry;
 
 use crate::common::error::Error;
-use crate::database::entries::limits::LimitEntryKey;
 
 /// Emily lambda settings.
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
@@ -100,7 +99,7 @@ impl Settings {
             limit_table_name: env::var("LIMIT_TABLE_NAME")?,
             throttle_table_name: env::var("THROTTLEDOWN_TABLE_NAME")?,
             default_limits: LimitEntry {
-                key: LimitEntryKey::default(),
+                key: Default::default(),
                 peg_cap: env::var("DEFAULT_PEG_CAP")
                     .ok()
                     .map(|v| v.parse())
--- a/emily/handler/src/database/entries/limits.rs
+++ b/emily/handler/src/database/entries/limits.rs
@@ -33,7 +33,7 @@ impl std::fmt::Display for LimitEntryType {
 }
 
 /// Limit table entry key. This is the primary index key.
-#[derive(Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
 #[serde(rename_all = "PascalCase")]
 pub struct LimitEntryKey {
     /// Type of the entry, representing if this entry is throttle mode entry or not.
@@ -42,8 +42,23 @@ pub struct LimitEntryKey {
     pub timestamp: u64,
 }
 
+
+
+
+impl Default for LimitEntryKey {
+    fn default() -> Self {
+        Self {
+            entry_type: LimitEntryType::Standard,
+            // The only requirements for this value is to be greater then zero and
+            // smaller then today.
+            // This magic number is timestamp of sBTC mainnet launch.
+            timestamp: 1_734_146_659,
+        }
+    }
+}
+
 /// Limit table entry key. This is the primary index key.
-#[derive(Clone, Default, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[derive(Clone, Debug, Default, Eq, PartialEq, Hash, Serialize, Deserialize)]
 #[serde(rename_all = "PascalCase")]
 pub struct LimitEntry {
     /// Limit entry key.


emily/handler/src/api/handlers/limits.rs
```diff
@@ -97,7 +97,7 @@ pub async fn set_limits(limits: Limits, context: EmilyContext) -> impl warp::rep
             accessors::set_limit_for_account(&context, &entry).await?;
         }
         // Get the limits from the database confirming that the updates were done.
-        let global_limits = accessors::get_limits(&context).await?;
+        let global_limits = accessors::get_limits(&context, true).await?;
```
matteojug: Should this return both non adjusted and the adjusted limits? I can see arguments for both returning the adjusted and non adjusted ones

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -134,8 +94,8 @@ pub async fn start_throttle(
                     key_name = %request.name,
                     "Successfull request to start throttle mode. Starting throttle mode.",
                 );
-                let new_limits = calculate_throttle_mode_limits(&context, initiator).await?;
-                tracing::info!(?new_limits, "Calculated limits to use in throttle mode",);
+                let mut new_limits = accessors::get_limits(&context, true).await?;
+                new_limits.throttle_mode_initiator = Some(initiator.clone());
                 let res = crate::api::handlers::limits::set_limits(new_limits.clone(), context)
                     .await
                     .into_response();
```
matteojug: If we do throttle adjustment in this `get_limits`, and `start_throttle` gets called twice, isn't the `set_limits` going to overwrite the normal limits with the throttled ones on the second call?

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -281,3 +241,40 @@ pub async fn activate_throttle_key(hash: String, context: EmilyContext) -> impl
         .await
         .map_or_else(Reply::into_response, Reply::into_response)
 }
+
+/// Get the throttle key details.
```
matteojug: nit: old comment

emily/handler/src/database/accessors.rs
```diff
@@ -850,13 +850,58 @@ async fn calculate_sbtc_left_for_withdrawals(
...
+pub const THROTTLE_MODE_ROLLING_WINDOW: u64 = 18;
+/// Rolling cap for throttle mode.
+pub const THROTTLE_MODE_ROLLING_CAP: u64 = 200_000_000; // 2 BTC.
+/// Per withdrawal cap for throttle mode.
+pub const THROTTLE_MODE_PER_WITHDRAWAL_CAP: u64 = 150_000_000; // 1.5 BTC
+
+/// Calculates throttle mode limits. It keeps most limits as they are now,
+/// while overwriting some of them.
+pub fn calculate_throttle_mode_limits(limit_entry: LimitEntry) -> LimitEntry {
+    tracing::info!(?limit_entry, "Calculating throttle mode limits");
```
matteojug: nit: debug log?

emily/handler/tests/integration/throttle.rs
```diff
@@ -519,3 +518,174 @@ async fn throttle_mode_initiator_correctly_shown_at_limits() {
...
+        .unwrap();
+
+    // Verify that the throttle_mode_initiator field is set correctly
+    let new_limits = apis::limits_api::get_limits(&configuration).await.unwrap();
+    assert_eq!(
+        new_limits.per_withdrawal_cap,
+        Some(Some(
+            emily_handler::database::accessors::THROTTLE_MODE_PER_WITHDRAWAL_CAP
+        ))
+    );
```
matteojug, line 575: nit: can we do this check also before calling `start_throttle`, and also check who is the initiator (in all the two/three checks)?

emily/handler/src/api/handlers/limits.rs
MCJOHN974: Hmm, why you think it may be useful? For me it feels like requiting to add extra logic to signer/bridge/whoever else gets limits from Emily while I don't really see why canonical signer/bridge needs both options

matteojug: Sorry I was referring to the return value of `set_limits`:
 - If we return the adjusted ones and there's throttling active, you may not know if your operation was successful or not
 - If we return the non adjusted ones you may get the impression your update will be reflected in the used limits, but because of the active throttling the enforced limits will actually be something else

But looking back at this, since you must specify the initiator in the request then maybe it's fine to return just the adjusted ones.

emily/handler/src/api/handlers/throttle.rs
MCJOHN974: [b6476d7](https://github.com/stacks-sbtc/sbtc/pull/1920/commits/b6476d73e18d5b7f1365520a8a8ce204f0f03ca3)[b6476d7](https://github.com/stacks-sbtc/sbtc/pull/1920/commits/b6476d73e18d5b7f1365520a8a8ce204f0f03ca3)

emily/handler/src/api/handlers/limits.rs
MCJOHN974: Hmm, am I correct that you describing situation where:

- slow mode is on

- I want to change original limits

- While getting limits I don't see my changes because they was overwritten by slow mode limits



?

emily/handler/src/api/handlers/throttle.rs
MCJOHN974: [9ce6f68](https://github.com/stacks-sbtc/sbtc/pull/1920/commits/9ce6f682472a84d9259855a4aaa04d38147a1371)[9ce6f68](https://github.com/stacks-sbtc/sbtc/pull/1920/commits/9ce6f682472a84d9259855a4aaa04d38147a1371)

emily/handler/src/database/accessors.rs
MCJOHN974: [3543c9f](https://github.com/stacks-sbtc/sbtc/pull/1920/commits/3543c9f8f2eed24c01c628589ed2c5f111d05e21)[3543c9f](https://github.com/stacks-sbtc/sbtc/pull/1920/commits/3543c9f8f2eed24c01c628589ed2c5f111d05e21)

emily/handler/tests/integration/throttle.rs
MCJOHN974: I think most of this is covered by the other tests but why not: [0eeb1da](https://github.com/stacks-sbtc/sbtc/pull/1920/commits/0eeb1da02bcb6de09b6c21d6431e23ce4bf71ea8)[0eeb1da](https://github.com/stacks-sbtc/sbtc/pull/1920/commits/0eeb1da02bcb6de09b6c21d6431e23ce4bf71ea8)

emily/handler/src/api/handlers/limits.rs
matteojug: Kinda:
 - Slow mode is on
 - I change the limits
 - The return value from the `set_limits` API call returns the (possibly new) current adjusted limits

I was under the impression that this could be confusing because the the return value could have been unexpected if slow mode was on, but since you specify the initiator when setting the limit you know what to expect (slow mode or not).

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -134,8 +94,17 @@ pub async fn start_throttle(
...
-                let new_limits = calculate_throttle_mode_limits(&context, initiator).await?;
-                tracing::info!(?new_limits, "Calculated limits to use in throttle mode",);
+                let mut new_limits = accessors::get_limits(&context, true).await?;
+                if new_limits.throttle_mode_initiator.is_some() {
+                    tracing::info!(
+                        // Safety: unwrap() here is ok because of .is_some() check above.
+                        existing_initiator = %new_limits.throttle_mode_initiator.clone().unwrap(),
+                        new_initiator = %initiator,
+                        "Throttle mode is already on. Ignoring request to start throttle mode.",
+                    );
```
matteojug, line 105: nit: we can remove the unwrap using if let (eg, `if let Some(ref existing_initiator) = new_limits.throttle_mode_initiator { ... }`)

emily/handler/src/api/handlers/throttle.rs
matteojug: nit: we can remove the TODO

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -134,8 +94,17 @@ pub async fn start_throttle(
                     key_name = %request.name,
                     "Successfull request to start throttle mode. Starting throttle mode.",
                 );
```
matteojug, line 95: nit: should this log be moved after the already on check?

emily/handler/src/api/handlers/throttle.rs
matteojug, line 115: Returning `new_limits` we are not returning the throttled limits if throttle mode was just activated (only the initiator would be updated)

emily/handler/tests/integration/throttle.rs
```diff
@@ -519,3 +518,256 @@ async fn throttle_mode_initiator_correctly_shown_at_limits() {
...
+        .unwrap();
+
+    // Register a throttle key
+    let key_name = "aaaaaaaaaaaaaaaa".to_string();
+    let secret = "very secret string".to_string();
+    let throttle_key = ThrottleKey {
+        name: key_name.clone(),
+        secret: secret.clone(),
+    };
+    let throttle_reqwest = ThrottleRequest {
```
matteojug, line 547: nit, multiple occurrences
```suggestion
    let throttle_request = ThrottleRequest {
```

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -281,3 +250,40 @@ pub async fn activate_throttle_key(hash: String, context: EmilyContext) -> impl
...
+        (status = 500, description = "Internal server error", body = ErrorResponse)
+    ),
+    security(("ApiGatewayKey" = []))
+)]
+#[instrument(skip(context))]
+pub async fn stop_throttle(context: EmilyContext) -> impl warp::reply::Reply {
+    // Internal handler so `?` can be used correctly while still returning a reply.
+    async fn handler(context: EmilyContext) -> Result<impl warp::reply::Reply, Error> {
+        let mut limits = accessors::get_limits(&context, false).await?;
+        limits.throttle_mode_initiator = None;
```
matteojug: nit: should this return an error if throttle mode is not activate?

emily/handler/tests/integration/throttle.rs
```diff
@@ -519,3 +518,256 @@ async fn throttle_mode_initiator_correctly_shown_at_limits() {
...
+    apis::throttle_api::add_throttle_key(&configuration, throttle_key)
+        .await
+        .unwrap();
+
+    // Trigger throttle mode multiple times
+    for _ in 0..3 {
+        apis::throttle_api::start_throttle(&configuration, throttle_reqwest.clone())
+            .await
+            .unwrap();
+    }
```
matteojug, line 741: Can we do the successive calls using a different initiator? To check that we persist the first one

emily/handler/tests/integration/throttle.rs
```diff
@@ -519,3 +518,256 @@ async fn throttle_mode_initiator_correctly_shown_at_limits() {
...
+
+    // Verify that the throttle_mode_initiator field is set correctly
```
matteojug, line 743: Can we add the `throttle_mode_initiator` check? Currently we only check the `per_withdrawal_cap`

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -281,3 +250,40 @@ pub async fn activate_throttle_key(hash: String, context: EmilyContext) -> impl
...
 }
+
+/// Stop throttle.
+#[utoipa::path(
+    post,
+    operation_id = "stopThrottle",
+    path = "/throttle/stop",
+    tag = "throttle",
+    responses(
+        (status = 200, description = "Throttle mode turned off. Limits restored", body = ()),
```
djordon, line 262: You want to have this endpoint not return a response, you're supposed to leave out the `body` here, similar to what we do in  `add_throttle_key`.

emily/handler/src/api/handlers/throttle.rs
```diff
@@ -281,3 +250,40 @@ pub async fn activate_throttle_key(hash: String, context: EmilyContext) -> impl
...
+            .await
+            .into_response()
+            .status()
+            .is_success();
+
+        if !is_success {
+            tracing::error!("Error restoring limits while stopping throttle mode");
+            return Err(Error::InternalServer);
+        }
+        Ok(with_status(json(&()), StatusCode::OK))
```
djordon, line 273: You probably want this to be the following:
```suggestion
        Ok(with_status(warp::reply(), StatusCode::OK))
```

emily/handler/src/database/accessors.rs
```diff
@@ -910,6 +954,13 @@ pub async fn get_limits(context: &EmilyContext) -> Result<Limits, Error> {
         .map(|(account, limit_entry)| (account, AccountLimits::from(limit_entry)))
         .collect();
 
+    // Adjust for throttle mode if needed.
+    let global_cap = if do_throttle_adjustment {
+        calculate_throttle_mode_limits(global_cap)
+    } else {
+        global_cap
+    };
```
djordon: Why do this adjustment here? It appears that this function currently "just" fetches the limits. After this change it will fetch the limits and then conditionally changes them to something else. I think it would be better to modify the limits outside of this function, and keep this one focused on getting the current limits.

emily/handler/src/api/handlers/throttle.rs
djordon: I'm okay with either behavior, with a slight preference for always setting the initiator to `None`.

emily/handler/src/api/handlers/limits.rs
djordon: Also, isn't this supposed to be:
```rust
let global_limits = accessors::get_limits(&context, false).await?;
```
or probably, something like:
```rust
let mut global_limits = accessors::get_limits(&context, false).await?;
if global_limits.throttle_mode_initiator.is_some() {
    global_limits = accessors::get_limits(&context, true).await?;
}
```

emily/handler/src/api/handlers/limits.rs
```diff
@@ -30,7 +30,7 @@ use warp::reply::{Reply, json, with_status};
 pub async fn get_limits(context: EmilyContext) -> impl warp::reply::Reply {
     // Internal handler so `?` can be used correctly while still returning a reply.
     async fn handler(context: EmilyContext) -> Result<impl warp::reply::Reply, Error> {
-        let global_limits = accessors::get_limits(&context).await?;
+        let global_limits = accessors::get_limits(&context, true).await?;
```
djordon: Wait, doesn't this always return the throttle limits? I don't think we want this. Instead we want to fetch the limits and then if throttle mode is activated, return the throttle limits.

emily/handler/src/database/accessors.rs
djordon: Ohh, you put it here so that the available to withdraw calculation happens correctly. I see. Doing otherwise requires us to change how we calculate the `available_to_withdraw` value...

PR #1927, opened 2026-02-17
fix: handle any max fee that can fit in a `u64`

djordon:
## Description

Closes: https://github.com/stacks-sbtc/sbtc/issues/1925

I prefer these changes over the ones in https://github.com/stacks-sbtc/sbtc/pull/1926.

## Changes

* Have Emily only return deposit requests where the max-fee was less than `i64::MAX`. This is here until the next release where we have shipped the signer's changes.
* Update the signer to be able to process any max-fee by changing the column that stores the max-fee.

## Checklist

### DIFF
--- a/signer/tests/integration/postgres.rs
+++ b/signer/tests/integration/postgres.rs
@@ -4914,6 +4914,7 @@ async fn deposit_requests_max_fee_migration() {
               AND output_index = $2
             "#,
         )
+        .persistent(false)
         .bind(txid)
         .bind(i32::try_from(output_index).map_err(Error::ConversionDatabaseInt)?)
         .fetch_optional(db.pool())


signer/tests/integration/block_observer.rs
```diff
@@ -1901,31 +1901,26 @@ fn make_coinbase_deposit_request(
     (deposit_tx, req, info)
 }
 
-/// This test checks that the block observer ignores deposits with a max
-/// fee that is too high.
+/// This test checks that the block observer writes deposits with a a high
+/// max–fee to the database.
 #[test_log::test(tokio::test)]
-async fn block_observer_ignores_deposits_with_invalid_max_fee() {
+async fn block_observer_ignores_deposits_with_high_max_fee() {
```
matteojug, line 1907: nit: no longer ignores in test name

djordon: I'll also fix this in a separate PR.

PR #1931, opened 2026-02-24
feat: add rate limiting for signer messaging

MCJOHN974:
## Description

Closes: #?

## Checklist


signer/src/network/libp2p/swarm.rs
```diff
@@ -42,6 +43,61 @@ const MAX_SUBSTREAMS_PER_CONNECTION: usize = 20;
 /// timeout is applied to both inbound and outbound connections.
 const NEGOTIATION_TIMEOUT_SECS: u64 = 10;
 
+use libp2p::gossipsub::{PeerScoreParams, PeerScoreThresholds, TopicHash, TopicScoreParams};
+
+/// The maximum number of messages a peer can send per second before being ignored.
+pub const MESSAGES_PER_SECOND: u32 = 100;
```
MCJOHN974: This number should be lower (such that we can process all messages) or we need some more robust system where we limit not by messages per second but by messages in channel (aka signer drops messages from peer if channel already contains X messages from that peer)

signer/src/context/signer_state.rs
```diff
@@ -480,6 +486,12 @@ impl SignerSet {
             .write()
             .expect("BUG: Failed to acquire write lock");
 
+        #[allow(clippy::expect_used)]
+        self.rate_limits
+            .lock()
+            .expect("BUG: Failed to acquire write lock")
+            .clear();
```
djordon-stacks: Shouldn't only clear the signers who are no longer in the old signer set?

djordon: Shouldn't only clear the signers who are no longer in the old signer set?

signer/src/context/signer_state.rs
```diff
@@ -564,6 +582,33 @@ impl SignerSet {
             .find(|signer| signer.peer_id() == peer_id)
             .map(|signer| *signer.public_key())
     }
+
+    /// Checks if a peer has exceeded the allowed messages per second limit.
+    /// Increments their counter and returns `true` if they are allowed,
+    /// or `false` if they are spamming.
```
djordon:
```suggestion
    /// or `false` if they have exceeded the limit.
```

signer/src/network/libp2p/event_loop.rs
```diff
@@ -19,7 +20,7 @@ use super::TOPIC;
 use super::swarm::{SignerBehavior, SignerBehaviorEvent};
 
 #[tracing::instrument(skip_all, name = "swarm")]
-pub async fn run(ctx: &impl Context, swarm: Arc<Mutex<Swarm<SignerBehavior>>>) {
+pub async fn run(ctx: &impl Context, swarm: Arc<Mutex<Swarm<SignerBehavior>>>, rate_limit: u32) {
```
djordon, line 23: Hmmm, it feels like the rate limit should be part of our `SignerBehavior`.

signer/src/network/libp2p/swarm.rs
```diff
@@ -42,6 +42,9 @@ const MAX_SUBSTREAMS_PER_CONNECTION: usize = 20;
 /// timeout is applied to both inbound and outbound connections.
 const NEGOTIATION_TIMEOUT_SECS: u64 = 10;
 
+/// The maximum number of messages a peer can send per second before being ignored.
+pub const MESSAGES_PER_SECOND: u32 = 10;
```
djordon, line 46: How did you arrive at this value? Seems a little low.

signer/src/network/libp2p/event_loop.rs
MCJOHN974: Yeas, I also wanted to put it to `SignerBehavior` in the beginning. However, `SignerBehavior` does `#[derive(NetworkBehaviour)]` which don't allow to put `rate_limit` there in easy way

signer/src/network/libp2p/swarm.rs
MCJOHN974: For now it not fixed in stone number but more like placeholder.

 About too low, I also initially expected it to be bigger, however now I think that after measurements it will be maybe around 100 but not around 1000

signer/src/context/signer_state.rs
MCJOHN974: Since we doing `.entry().or_insert()` during reading from this hashmap it is fine, but I switched to rwlock to be more consistent with the rest of the code: [01c4c97](https://github.com/stacks-sbtc/sbtc/pull/1931/commits/01c4c97aeebb32b48644c9a0562862e021e0d56c)

signer/src/network/libp2p/swarm.rs
MCJOHN974: p2p communication measurements:

metric observed: messages from same peer within same second
method: run devenv with a lot of deposits/withdrawals

case1: normal demo: 1 dkg + 1 deposit. 10 eps max, during dkg.
case2: blasting 1000 withdrawals + 1000 deposits in a loop without sleeping: around 400 eps max
case3: same as case 2, but mining interval set to 1 minute: 100-200 eps, with ~50 bitcoin tx per block and ~70 stacks tx per bitcoin block (some transactions was failing) due to my setup `NoAvailableUtxos`)

signer/src/network/libp2p/event_loop.rs
djordon: I think we can just add something like:
```rust
    /// Maximum number of requests allowed from a peer within the
    /// configured rate-limiting window.
    #[behaviour(ignore)]
    pub rate_limit: u32,
```
to the `SignerBehavior`. I'm not advocating that we use only a `u32` here, just that we could add whatever we want to our behavior.

djordon: Also, we may want to specify the rate_limit as a number of messages over some duration, so maybe we'd have a struct here with two parameters.

djordon-stacks:
Closing since we haven't seen much action pushing it through, and we are unlikely to see that anytime soon. Plus, when things change, we can always reopen later.

Issue #1941, opened 2026-03-02
[Bug]: Flaky `sign_bitcoin_transaction_withdrawals` test


MCJOHN974:
Test run link: https://github.com/stacks-sbtc/sbtc/actions/runs/22572296229/job/65382917549?pr=1896
Test name: `sign_bitcoin_transaction_withdrawals`
Seed: `412573342467952767u64`
Commit: 6c8a646ebe8d1d0b05ce772270247d787acb9b4e

I tried to run this test locally on this commit with this seed, however, I wasn't able to reproduce a failure

PR #1942, opened 2026-03-03
tests: fix flaky test runs

djordon:
## Description

Closes: #1941 

## Changes

* Send the chain tip block ref upon tenure completion. This is to make testing a little better.
* Use the block observed signal in more of our block observer tests.

## Testing Information

Although we are sending a few more bytes in our channel with each bitcoin block, these changes should only really affect tests.

## Checklist

### DIFF
--- a/signer/tests/integration/block_observer.rs
+++ b/signer/tests/integration/block_observer.rs
@@ -86,7 +86,7 @@ use crate::utxo_construction::make_deposit_request;
 /// it concurrently).
 async fn wait_for_block_observed<S, B, E>(
     ctx: &TestContext<PgStore, B, S, E>,
-    chain_tip: BitcoinBlockHash,
+    block_hash: BitcoinBlockHash,
 ) where
     B: BitcoinInteract + Clone + Send + Sync + 'static,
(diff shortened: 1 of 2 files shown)


signer/tests/integration/block_observer.rs
```diff
@@ -78,6 +81,30 @@ use crate::setup::new_emily_setup;
 use crate::transaction_coordinator::mock_reqwests_status_code_error;
 use crate::utxo_construction::make_deposit_request;
 
+/// Wait for all signers to finish their coordinator duties and do this
+/// concurrently so that we don't miss anything (not sure if we need to do
+/// it concurrently).
+async fn wait_for_block_observed<S, B, E>(
```
matteojug, line 87: nit: old comment

matteojug (approved):
Thank you!

MCJOHN974 (approved):

signer/tests/integration/block_observer.rs
djordon: Oh, nuts. I'll fix this in another PR, since I want to get the code merged so that it doesn't bite anyone else.

PR #1958, opened 2026-03-08
chore: remove unnecessary dependencies from wsts

djordon:
## Description

Closes: https://github.com/stacks-sbtc/sbtc/issues/1957

## Changes

* Replace use of `hashbrown` types with the ones in the standard library
* Replace the use of the `bs58` crate with the hex crate.

## Checklist


MCJOHN974:
> Replace use of hashbrown types with the ones in the standard library

Why we want to do so? We had it in signer as well.

p.s. I don't mind moving everything to std hashset, just interested in your motivation

wsts/src/state_machine/coordinator/fire.rs
```diff
@@ -703,7 +703,7 @@ impl Coordinator {
                 }
             }
             if dkg_failures.is_empty() {
-                warn!("no dkg failures");
+                info!("no dkg failures");
```
matteojug, line 706: ❤️

matteojug (approved):

PR #1965, opened 2026-03-11
feat: remove the signer trait

djordon:
## Description

This is part of https://github.com/stacks-sbtc/sbtc/issues/1964.

## Changes

* Delete the signer trait. There is only one implementing object since https://github.com/stacks-sbtc/sbtc/pull/1956.
* Remove `v2::Party::validate_party_id` replace the simple check.

## Checklist

### DIFF
--- a/wsts/src/traits.rs
+++ b/wsts/src/traits.rs
@@ -52,7 +52,7 @@ pub mod test_helpers {
     ) -> Result<HashMap<u32, PolyCommitment>, HashMap<u32, DkgError>> {
         let public_shares: HashMap<u32, PolyCommitment> = signers
             .iter()
-            .flat_map(|s| s.get_poly_commitment(rng))
+            .filter_map(|s| s.get_poly_commitment(rng))
             .map(|comm| (comm.id.id.get_u32(), comm))
             .collect();
         let mut private_shares = HashMap::new();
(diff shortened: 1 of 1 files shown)


wsts/src/taproot.rs
```diff
@@ -68,31 +68,32 @@ impl From<[u8; 64]> for SchnorrProof {
...
-    pub fn dkg<RNG: RngCore + CryptoRng, Signer: traits::Signer>(
-        signers: &mut [Signer],
+    pub fn dkg<RNG: RngCore + CryptoRng>(
+        signers: &mut [v2::Party],
         rng: &mut RNG,
     ) -> Result<HashMap<u32, PolyCommitment>, HashMap<u32, DkgError>> {
         let polys: HashMap<u32, PolyCommitment> = signers
             .iter()
-            .flat_map(|s| s.get_poly_commitments(rng))
+            .filter_map(|s| s.get_poly_commitment(rng))
```
matteojug, line 90: nit/qq: this was changed to a `filter_map`, while the other two in `traits.rs` were kept with `flat_map`; just curios if there was a specific reason

djordon: No, I don't think I had a reason. But I should change them to all be consistent, I'll make them all use `filter_map`.

djordon: I changed the cases in `traits.rs` in https://github.com/stacks-sbtc/sbtc/pull/1965/commits/0a386c7001ef28c0d50aec8dbfda83f3f75f3578 to use `filter_map`. No strong argument for either one.

PR #1976, opened 2026-03-19
feat: remove necessity of `txindex=1` on bitcoin core

### LINKED ISSUE #1973: [Feature]: Remove necessity of `txindex=1` on bitcoin-core
# Feature - Remove necessity of `txindex=1` on bitcoin-core

## 1. Description

We would like to reduce the resources required to run an sBTC signer. The biggest resource requirements to run a signers are the following dependencies, in order of greatest to least amount of resources:
1. stacks core
2. bitcoin core
3. postgres
4. the sBTC signer binary
5. the blocklist client

We cannot do anything to reduce the resource requirements of stacks core, but we can make things easier on the bitcoin core side of things. Right now, we require signer operators to run bitcoin core with `txindex=1`, which means that operators cannot run pruned nodes. Allowing operators to run a pruned node would greatly reduce the storage requirements to running bitcoin core. 

 

## 2. Technical Details

The sBTC signer requires `txindex=1` because we need to find out the block hash that confirms deposit request transactions and bitcoin core provides only a few ways to do so. The main way is the `getrawtransaction` RPC [[1]], which is what we current use, and that RPC will not work the way that we need if `txindex=0`. The only alternative that I could find is the `gettxout` RPC [[2]], which returns the number of confirmations for the output relative to the chain tip. Fortunately, that is enough information for us to figure out the block hash confirming the transaction.

Replace the use of the `getrawtransaction` without the block hash, with one that uses `gettxout` followed by either a query to get the block hash confirming the transaction or another RPC to bitcoin core for the relevant block hash.

We could also simplify the implementation of `get_transaction_fee`, since we only need to get fee info for transactions in the mempool.

[1]: https://bitcoincore.org/en/doc/25.0.0/rpc/rawtransactions/getrawtransaction/

[2]: https://bitcoincore.org/en/doc/25.0.0/rpc/blockchain/gettxout/

### 2.1 Acceptance Criteria

- [ ] We can remove the requirement of running bitcoin-core with `txindex=1`.

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1973. 

This change will allow signer operators to run a bitcoin core node with `prune=1` and `txindex=0` in their config. But  things will not necessarily work well if they set the `prune` value to be something else. So we do not really allow running a pruned node, at least not yet.

## Changes

* Add a new `get_utxo_info` function that returns the necessary information about unspent outputs without the need for `txindex=1`.
* Made `BitcoinInteract::get_tx` a test only function, since it is no longer used production code.
* Updated devenv to work on a bitcoin core node with `txindex=0`.

## Testing Information

Also, I tested this code on devenv with `txindex=0`, and was able to sweep funds just fine.

## Checklist


docker/docker-compose.yml
```diff
@@ -286,7 +286,7 @@ services:
     ports:
       - "127.0.0.1:18443:18443"
     volumes:
-      - ./bitcoin/bitcoin.conf:/root/.bitcoin/bitcoin.conf
+      - ./bitcoin/bitcoin-pruned.conf:/root/.bitcoin/bitcoin.conf
```
matteojug, line 289: Should we use the pruned config also for tests?

Currently we have 4 tests that fails without txindex:
```
bitcoin_forks::serial::getrawtransaction_simple_fork
bitcoin_forks::serial::getrawtransaction_single_tx_regorged
bitcoin_rpc::serial::btc_client_getstransaction
transaction_coordinator::serial::test_get_btc_state_with_available_sweep_transactions_and_rbf
```
the first 3 are probably no longer relevant, while the last can be fixed by supplying the block hash (or get the tx while still in mempool)

signer/src/bitcoin/rpc.rs
```diff
@@ -422,11 +436,13 @@ impl BitcoinCoreClient {
     /// By default, this call only returns a transaction if it is in the
     /// mempool. If -txindex is enabled on bitcoin-core and no blockhash
     /// argument is passed, it will return the transaction if it is in the
-    /// mempool or any block. We require -txindex to be enabled (same with
-    /// stacks-core[^1]) so this should work with transactions in either
+    /// mempool or any block. We do not require -txindex to be enabled (same with
+    /// stacks-core[2]) so this should work with transactions in either
```
matteojug: In the page we link it's not explicitly stated but in the sample config (in that same page) they use `txindex=1`; maybe we should ask to update that, unless it's there for some specific reason

signer/src/bitcoin/rpc.rs
```diff
@@ -444,6 +460,63 @@ impl BitcoinCoreClient {
...
+    ///   canonical bitcoin blockchain.
+    /// * the height of the associated block confirming the transaction
+    ///   that created the given output is not found in bitcoin core using
+    ///   the `getblockhash` RPC. This condition should never happen.
+    ///
+    /// The documentation of the `getblockhash` RPC can be found at:
+    /// <https://bitcoincore.org/en/doc/25.0.0/rpc/blockchain/getblockhash/>
+    pub fn get_utxo_info(&self, outpoint: &OutPoint) -> Result<Option<OutPointSummary>, Error> {
+        // This will return Some(_) result if the transaction is confirmed
+        // in a canonical block.
```
matteojug: nit
```suggestion
        // This will return Some(_) result if the transaction is confirmed
        // in a canonical block and unspent.
```

signer/src/bitcoin/rpc.rs
```diff
@@ -444,6 +460,63 @@ impl BitcoinCoreClient {
...
+        // that we need to add 1; the confirmations value is one more than
+        // the "offset" value that we want. For example, a transaction that
+        // is confirmed at the chain tip has 1 confirmation, but the offset
+        // from the chain tip is 0.
+        let confirmation_height = chain_tip
+            .height
+            .saturating_sub(out.confirmations as u64)
+            .saturating_add(1u64);
+
+        match self.inner.get_block_hash(*confirmation_height) {
```
matteojug, line 511: There could be race here:
 - the tx was confirmed in block N, so we get the tx out (let's say 1 confirmation) and and the chain tip block header (for block N)
 - now another M become the canonical tip, and it doesn't confirm the tx
 - when fetching `get_block_hash`, we fetch M, and return Ok(M)

We would return the wrong `OutPointSummary`, but it should be fine since we will catch this later when fetching `get_tx_info` in `impl DepositRequestValidator for CreateDepositRequest` `validate` as we will not find the tx in M, but maybe we should state this out (both here and update the comment in the validate (currently `The get_tx_info call here should not return None`)

signer/src/bitcoin/rpc.rs
```diff
@@ -444,6 +460,63 @@ impl BitcoinCoreClient {
...
+            Ok(block_hash) => Ok(Some(OutPointSummary {
+                block_hash,
+                is_coinbase: out.coinbase,
+            })),
+            Err(BtcRpcError::JsonRpc(JsonRpcError::Rpc(RpcError { code: -5, .. }))) => Ok(None),
```
matteojug: I couldn't find any reference for `code: -5` error for `getblockhash`, where does it come from?

docker/docker-compose.yml
djordon: Yeah, we can do so here if you like. I considered making those changes here but decided against it to keep this diff down. The plan was to make the change in a separate PR, but it's your call.

signer/src/bitcoin/rpc.rs
djordon: Yeah, it's worth noting this. Also, I think this race is also true of our current code. But this is not an argument against spelling it out though.

signer/src/bitcoin/rpc.rs
djordon: Yeah, it's not code -5, it's code -8. Whoops

signer/src/bitcoin/rpc.rs
djordon: Oh good catch. I need to update the link to point to the right docs. Will do.

signer/src/bitcoin/rpc.rs
djordon: I removed this in https://github.com/stacks-sbtc/sbtc/pull/1976/commits/44ef366d8a04e79530a101bff2db0c47eeaf0506

signer/src/bitcoin/rpc.rs
djordon: I added some comments in 44ef366d8a04e79530a101bff2db0c47eeaf0506

signer/src/bitcoin/rpc.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1976/commits/b21bf2cea7ba7fa105867e9bc7f475880873520d

signer/src/bitcoin/rpc.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1976/commits/b21bf2cea7ba7fa105867e9bc7f475880873520d

signer/tests/integration/block_observer.rs
```diff
@@ -132,18 +134,27 @@ async fn load_latest_deposit_requests_persists_requests_from_past(blocks_ago: u6
         .build();
     ctx.state().update_current_limits(SbtcLimits::unlimited());
 
-    // We're going to create two confirmed deposits. This also generates
-    // sweep transactions, but this information is not in our database, so
-    // it doesn't matter for this test.
-    let setup0 = TestSweepSetup::new_setup(bitcoin.get_client(), faucet, 100_000, &mut rng);
-    let setup1 = TestSweepSetup::new_setup(bitcoin.get_client(), faucet, 200_000, &mut rng);
+    // We're going to create two confirmed deposits. These deposits
+    // outputshave not been spent, so the block observer should pick them
```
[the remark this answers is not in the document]
djordon: Fixed in 7f0b2958abd9a56db9b83d3c0e3526d4874cfa85

signer/src/bitcoin/rpc.rs
```diff
@@ -422,11 +436,13 @@ impl BitcoinCoreClient {
     /// By default, this call only returns a transaction if it is in the
     /// mempool. If -txindex is enabled on bitcoin-core and no blockhash
     /// argument is passed, it will return the transaction if it is in the
-    /// mempool or any block. We require -txindex to be enabled (same with
-    /// stacks-core[^1]) so this should work with transactions in either
+    /// mempool or any block. We do not require -txindex to be enabled (same with
+    /// stacks-core[2]) so this should work with transactions in either
     /// the mempool and a bitcoin block.
```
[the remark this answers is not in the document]
djordon: I think I fixed this in f8915e51d97ca004364e4b71f66b4cec7983b3cf

signer/src/testing/block_observer.rs
```diff
@@ -217,6 +219,21 @@ impl TryFrom<TestHarness> for ApiFallbackClient<TestHarness> {
...
+
+        let Some(block_hash) = tx_response.block_hash else {
+            return Ok(None);
+        };
+
+        Ok(Some(OutPointSummary {
+            block_hash,
+            is_coinbase: tx_response.tx.is_coinbase(),
+        }))
+    }
```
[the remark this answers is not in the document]
djordon: Yeah true. Not worth fixing, since deposits are indexed off of transaction ID anyway.

matteojug (approved):

cylewitruk (approved):
lgtm 👍

PR #1979, opened 2026-03-20
feat: add withdrawal e2e test

### LINKED ISSUE #1949: [Chore]: Add end to end withdrawal tests using stacks-core
# Chore - Add end to end withdrawal tests using stacks-core

## 1. Description

In https://github.com/stacks-sbtc/sbtc/pull/1906 we added stacks-core in CI and an end to end test for a deposit. We should add a similar test for withdrawals as well. Note that since we don't have a working event observer (yet?), the test will need to inject the events to the signers.

matteojug:
## Description

Closes: https://github.com/stacks-sbtc/sbtc/issues/1949

## Changes

Add a withdrawal to the end to end tests using a stacks node.

Since we don't have an event observer wired for the tests, the test spawns a task polling the chain and replaying blocks to process the events, similarly to the new block handler, but copying the logic since the data returned by replaying a block is not the same as the data received by the block observer.
We could better factor the new block logic to reuse exactly the same in the test, maybe in another PR.

## Testing Information

It's a test, so if it passes everything should be fine.

## Checklist


signer/tests/integration/e2e.rs
```diff
@@ -225,11 +293,63 @@ async fn get_sbtc_balance(
...
+/// Simulates `new_block_handler` in the sBTC signer API. We can't use that
+/// directly because we don't get exactly the same data from the block replay.
+async fn new_block_handler(ctx: &IntegrationTestContext<StacksClient>, block: &BlockReplay) {
+    let PrincipalData::Standard(deployer) = ctx.config().signer.deployer.to_account_principal()
+    else {
+        panic!("unexpected deployer")
+    };
+
+    let registry_address =
+        QualifiedContractIdentifier::new(deployer, ContractName::from(SBTC_REGISTRY_CONTRACT_NAME));
```
djordon: This is fine, but we could also use `SmartContract::SbtcRegistry.contract_name()` here too. And now that I think of it, we might want to replace `SBTC_REGISTRY_CONTRACT_NAME` with that sometime in the future as well.

signer/tests/integration/e2e.rs
```diff
@@ -225,11 +293,63 @@ async fn get_sbtc_balance(
...
+
+/// End to end test for deposit and withdrawal:
+///  - the sBTC signers bootstrap
+///  - a deposit is created on Emily
+///  - the signers fulfill it (with a controlled chain progression) and mint sBTC
+///  - the recipient creates a withdrawal request
+///  - the signers fulfill it (with a controlled chain progression) and pegout BTC
 #[test_log::test(tokio::test)]
-async fn deposit() {
+async fn deposit_and_withdrawal() {
```
djordon, line 354: I suppose this is fine, but I think we should have a test for deposits, one withdrawals, and perhaps a test that does both. That way if one part of the flow breaks, we get a clearer signal which part is broken.

signer/tests/integration/e2e.rs
```diff
@@ -346,12 +474,175 @@ async fn deposit() {
...
+        assert_eq!(accepted_withdrawals.len(), 1);
+        let withdrawal = accepted_withdrawals.first().unwrap();
+
+        assert_eq!(withdrawal.amount, withdrawal_amount);
+        assert_eq!(withdrawal.max_fee, withdrawal_max_fee);
+        if withdrawal_id != 0 {
+            assert_eq!(withdrawal.request_id, withdrawal_id);
+        } else {
+            withdrawal_id = withdrawal.request_id;
+        }
```
djordon, line 588: You should probably also check that `withdrawal.request_id != 0`. It shouldn't be, but it's good to have the check.

djordon (commented):
Nice, looks good. The only question I have is whether we have more than one end-to-end test here. I don't feel to strongly about it, but my default is to have two or three; one for deposits, another for withdrawals, and the another for both, which is the one you have here. All things considered, I'm okay with keeping things as is.

signer/tests/integration/e2e.rs
matteojug: I went for a single one because we require some setup for the withdrawal test: either we do that after a deposit (like here), or we need to do something like a donation big enough to cover the withdrawal, and a manual sBTC mint (issuing directly the mint contract call).
I didn't really like the latter approach since it was kinda fishy, but then having two tests with one being a prefix of the other was strange as well, so I merged the two.

I agree that having separate tests would be the best, but I'm not sure if we can have a nicely divided withdrawal test.

signer/tests/integration/e2e.rs
matteojug: Done in [c9cabc8](https://github.com/stacks-sbtc/sbtc/pull/1979/commits/c9cabc87cd56b709f8009ca84d4eab3c78b33fd3)

signer/tests/integration/e2e.rs
matteojug: Done in [c9cabc8](https://github.com/stacks-sbtc/sbtc/pull/1979/commits/c9cabc87cd56b709f8009ca84d4eab3c78b33fd3)

signer/tests/integration/e2e.rs
```diff
@@ -225,11 +293,65 @@ async fn get_sbtc_balance(
     }
 }
 
-/// End to end test for deposits: after the sBTC bootstrap a deposit is created
-/// on Emily, the signers do their magic (with a controlled chain progression)
-/// and we get sBTC minted.
+/// Simulates `new_block_handler` in the sBTC signer API. We can't use that
+/// directly because we don't get exactly the same data from the block replay.
+async fn new_block_handler(ctx: &IntegrationTestContext<StacksClient>, block: &BlockReplay) {
```
djordon: I would make this take ownership of the `BlockReplay` since it can, and it would mean that we don't need to clone things, or implement `Clone`. But it's your call.

djordon (dismissed):
Looks good, had one suggestion

signer/tests/integration/e2e.rs
djordon: Yeah, that makes sense, I can see that perspective too. Plus we have tests that should cleanly catch breaks in the deposit flow, so we should be able to isolate which paths are broken easily enough.

signer/tests/integration/e2e.rs
matteojug: Yeah it makes sense given we are using each the block replay only once, done in [4b3dc9d](https://github.com/stacks-sbtc/sbtc/pull/1979/commits/4b3dc9d4e0d72dd91dec49f3ba6bd776a0059410)

djordon (dismissed):
Looks good

djordon (approved):

djordon-stacks (approved):

PR #1985, opened 2026-03-25
chore: update dependencies dependabot alerts

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1991

There are a few vulnerabilities in our dependencies, and this PR updates to the patched versions of those dependencies, building on the latest rust since that affects things too. One thing that isn't updated is the `cargo vet` stuff, which is a lot. We can chat about what we want to do here.

## Changes

* This updates to the patched versions of our dependencies.

## Checklist

### DIFF
--- a/supply-chain/imports.lock
+++ b/supply-chain/imports.lock
@@ -1382,8 +1382,8 @@ who = "Lukasz Anforowicz <lukasza@chromium.org>"
 criteria = "safe-to-deploy"
 version = "1.0.35"
 notes = """
-Grepped for "unsafe", "crypt", "cipher", "fs", "net" - there were no hits
-(except for benign "net" hit in tests and "fs" hit in README.md)
+Grepped for \"unsafe\", \"crypt\", \"cipher\", \"fs\", \"net\" - there were no hits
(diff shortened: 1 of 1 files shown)


blocklist-client/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2024"
 [dependencies]
 clap.workspace = true
 config.workspace = true
-reqwest.workspace = true
+reqwest = { version = "0.11.27", default-features = false, features = ["json", "rustls-tls"] }
```
cylewitruk, line 10: Why not use a workspace alias here (and in `emily/handler`) like for `reqwest_012`? e.g.

```toml
// workspace
reqwest_0_11 = { package = "reqwest", version = "0.11.27", ... }
```
```toml
// crate
reqwest_0_11.workspace = true
```

djordon: Yeah, I agree. I think I changed this in a separate branch and then ended up removing all these reqwest versions entirely. In https://github.com/stacks-sbtc/sbtc/pull/1986, the whole workspace uses the same version of reqwest.

PR #1988, opened 2026-03-25
feat: retry constructing tx packages with lower fees

### LINKED ISSUE #1995: [Feature]: Change bitcoin fee estimation mode to economical
# Feature - Change Bitcoin fee estimation mode to economical

## 1. Description

For Bitcoin fee estimation we are currently using `estimatesmartfee` RPC with _conservative_ estimation mode.

Since we noticed that we are usually overpaying fees on mainnet, we should switch to `economical` instead.
We are generally fine with a higher risk of confirmation delay in favor of generally lower fees.

### 1.1 Context & Purpose

The original motivation for _conservative_ was:
> We use the "conservative" estimate because it is more likely to be sufficient for the desired target, but is not as responsive to short term drops in the prevailing fee market when compared to the "economical" fee rate.

matteojug:
## Description

Closes: #1995 #1996 

## Changes

If the Bitcoin proposed fee is too high to construct a sweep transaction, instead of wasting the tenure try to use lower fees.

Also change the estimation mode to economical.

Currently if we have no request with a max fee high enough for the current Bitcoin proposed fee, we fail to construct a sweep transaction. But we could try to use a lower fee -- worst case it will sit in the mempool until either is confirmed or we RBF it. In particular, we try either a fallback fee if configured, or the 3 blocks target estimated fee.

This can be useful for auto-bridging PoX rewards, as they may be low amounts (or purposely use low max fees).

## Testing Information

Adds integration tests for the new logic

## Checklist


djordon (commented):
The one thing that came to mind with this approach is the case where a user's deposit request has a max-fee that is too low and we try to spend it at a rate that we can tell won't be confirmed until after their reclaim timelock expires. It would be best to avoid that case, since the user will not be able to spend their funds unless they spend more than us. We can kinda end up in a similar situation today. But right now we try, to the best of our ability, to avoid a situation where we spend their funds with a fee rate where the sweep transaction won't confirm before their time lock expires.

signer/src/config/default.toml
```diff
@@ -80,6 +80,16 @@ rpc_endpoints = [
 # Environment: SIGNER_BITCOIN__CHAIN_TIP_POLLING_INTERVAL
 # chain_tip_polling_interval = 5
 
+# An optional fallback fee rate in sats/vbyte to use when the initial fee rate
+# is too high to construct any transaction package. When set, this value is used
+# directly as the retry fee rate. When unset, the signer estimates a lower fee
+# rate by targeting a longer confirmation window.
```
djordon, line 86: I would add something here that states that this should be unset on mainnet or actively ignore it on mainnet, so that no one gets the wrong impression.

signer/src/transaction_coordinator.rs
```diff
@@ -683,7 +684,49 @@ where
...
+                    // will not be in the mempool when the requests expire.
+                    // If that happens, users will need to either pay more fees
+                    // to RBF our sweep (for deposits reclaim) or wait for a new
+                    // sweep from us. Note that this can happen regardless of
+                    // what fee we use to construct the sweep.
+                    //
+                    // `estimatesmartfee` RPC expects the confirmation target to
+                    // be within [1, 1008]
+                    let target_blocks = DEPOSIT_LOCKTIME_BLOCK_BUFFER
+                        .min(WITHDRAWAL_EXPIRY_BUFFER.try_into().unwrap_or(u16::MAX))
```
djordon: I would just make this a constant.

signer/src/config/mod.rs
```diff
@@ -145,6 +145,13 @@ pub struct BitcoinConfig {
     /// field.
     #[serde(deserialize_with = "duration_seconds_deserializer")]
     pub timeout: std::time::Duration,
+
+    /// An optional fallback fee rate in sats/vbyte to use when the initial
+    /// fee rate is too high to construct any transaction package. When set,
+    /// this value is used directly as the retry fee rate. When `None`, the
+    /// signer estimates a lower fee rate by targeting a longer confirmation
+    /// window.
+    pub fallback_fee: Option<NonZeroU64>,
```
djordon: Since this is a test-only parameter, I can see an argument to allowing us to set whatever float value that we want. And if we want to forbid some values, we still can by validating the fallback fee.

djordon (commented):
Largely looks good, just had some comments about constants and docs.

signer/src/config/mod.rs
matteojug: Done in [5b62286](https://github.com/stacks-sbtc/sbtc/pull/1988/commits/5b6228657171e9601a30750829e788fc0afd9e5b)

signer/src/transaction_coordinator.rs
matteojug: Done in [5b62286](https://github.com/stacks-sbtc/sbtc/pull/1988/commits/5b6228657171e9601a30750829e788fc0afd9e5b) -- unfortunately we don't have stable const min/clamp, so here we go

signer/src/config/default.toml
matteojug: Done in [5b62286](https://github.com/stacks-sbtc/sbtc/pull/1988/commits/5b6228657171e9601a30750829e788fc0afd9e5b) -- added comments and validation to prevent using it in mainnet

signer/src/transaction_coordinator.rs
djordon: Ouch. That's ugly.

signer/src/transaction_coordinator.rs
```diff
@@ -83,6 +84,37 @@ use wsts::state_machine::coordinator::State as WstsCoordinatorState;
...
+/// (for deposits reclaim) or wait for a new sweep from us. Note that this can
+/// happen regardless of what fee we use to construct the sweep.
+///
+/// This const-computes min(WITHDRAWAL_EXPIRY_BUFFER, DEPOSIT_LOCKTIME_BLOCK_BUFFER).clamp(1, 1008)
+const FEE_RETRY_TARGET_BLOCKS: u16 = {
+    let expiry_buffer = if WITHDRAWAL_EXPIRY_BUFFER > u16::MAX as u64 {
+        u16::MAX
+    } else {
+        WITHDRAWAL_EXPIRY_BUFFER as u16
+    };
```
djordon: You don't need this check, since you are taking the minimum with `DEPOSIT_LOCKTIME_BLOCK_BUFFER` and `DEPOSIT_LOCKTIME_BLOCK_BUFFER` is a `u16`. And if you create an intermediate const value of type `NonZeroU8` from the `DEPOSIT_LOCKTIME_BLOCK_BUFFER`, then you wouldn't need to do the clamp either.

djordon (dismissed):
Looks good. Just had a suggestion to simplify the `FEE_RETRY_TARGET_BLOCKS` definition logic.

signer/src/transaction_coordinator.rs
matteojug: > You don't need this check, since you are taking the minimum with DEPOSIT_LOCKTIME_BLOCK_BUFFER and DEPOSIT_LOCKTIME_BLOCK_BUFFER is a u16

This is a defensive check against wrap arounds, eg if were to have `WITHDRAWAL_EXPIRY_BUFFER: u64 = (u16::MAX as u64)+1;` the resulting FEE_RETRY_TARGET_BLOCKS would be `1` (instead of 3). I'm not saying it's too realistic to have such high values, but I wanted this to be a safe forget-about-it const.

> And if you create an intermediate const value of type NonZeroU8 from the DEPOSIT_LOCKTIME_BLOCK_BUFFER, then you wouldn't need to do the clamp either.

It would not be enough on the lower bound because of the above, but even if it were I'm not sure having a `NonZeroU8` would be better from a readability pov (_where does the u8 come from?_)

djordon: I was thinking about doing something like this:
```rust
/// Intermediate value
pub const DEPOSIT_BUFFER_BLOCKS: std::num::NonZeroU8 = std::num::NonZeroU8::new(3).expect("3 is non zero");

/// The value used elsewhere.
pub const DEPOSIT_LOCKTIME_BLOCK_BUFFER: u16 = DEPOSIT_BUFFER_BLOCKS.get() as u16;

/// The retry blockks
pub const FEE_RETRY_TARGET_BLOCKS: u16 = if WITHDRAWAL_EXPIRY_BUFFER < DEPOSIT_BUFFER_BLOCKS.get() as u64 {
    // This is safe becauuse WITHDRAWAL_EXPIRY_BUFFER is less than u8::MAX
    WITHDRAWAL_EXPIRY_BUFFER as u16
} else {
    DEPOSIT_BUFFER_BLOCKS.get() as u16
};
```
Which has the compile time checks that we want, and does an implicit clamp. But yeah maybe this isn't actually more readable in general. When I wrote my first comment it seemed simpler in my mind.

matteojug: Reworked as suggested offline in [54a4591](https://github.com/stacks-sbtc/sbtc/pull/1988/commits/54a4591b0c9bf4e9511189ee02080e077c4d5fc2)

djordon (dismissed):

signer/src/transaction_coordinator.rs
```diff
@@ -683,7 +684,49 @@ where
...
+                    // be within [1, 1008]
+                    let target_blocks = DEPOSIT_LOCKTIME_BLOCK_BUFFER
+                        .min(WITHDRAWAL_EXPIRY_BUFFER.try_into().unwrap_or(u16::MAX))
+                        .clamp(1, 1008);
+
+                    self.context
+                        .get_bitcoin_client()
+                        .estimate_fee_rate(target_blocks)
+                        .await?
+                }
```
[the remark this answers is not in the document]
matteojug: It will not cause the whole tenure to error

signer/src/config/mod.rs
```diff
@@ -189,6 +197,20 @@ impl Validatable for BitcoinConfig {
             ));
         }
 
+        if let Some(fee) = self.fallback_fee {
+            if cfg.signer.network.is_mainnet() {
+                return Err(ConfigError::Message(
+                    "[bitcoin.fallback_fee] Cannot set in mainnet".to_string(),
```
adriano-stacks: nit: cannot _be_ set?

signer/src/config/mod.rs
```diff
@@ -145,10 +145,18 @@ pub struct BitcoinConfig {
...
     #[serde(deserialize_with = "duration_seconds_deserializer")]
     pub timeout: std::time::Duration,
+
+    /// A test-only optional fallback fee rate in sats/vbyte to use when the
+    /// initial fee rate is too high to construct any transaction package.
+    /// When set, this value is used directly as the retry fee rate.
+    /// When `None`, the signer estimates a lower fee rate by targeting a longer
+    /// confirmation window.
+    /// This is for tests only, it fails validation in mainnet.
+    pub fallback_fee: Option<f64>,
```
adriano-stacks, line 155: I am sure there's a reason, but why not:

1. unsigned
2. integer

?

signer/src/transaction_coordinator.rs
```diff
@@ -83,6 +84,29 @@ use wsts::state_machine::coordinator::State as WstsCoordinatorState;
...
+    let taget_block = if WITHDRAWAL_EXPIRY_BUFFER < DEPOSIT_LOCKTIME_BLOCK_BUFFER as u64 {
+        // This is safe, as DEPOSIT_LOCKTIME_BLOCK_BUFFER is a u16
+        WITHDRAWAL_EXPIRY_BUFFER as u16
+    } else {
+        DEPOSIT_LOCKTIME_BLOCK_BUFFER
+    };
+
+    // `estimatesmartfee` RPC expects the confirmation target to be within [1, 1008].
+    assert!(taget_block >= 1);
+    assert!(taget_block <= 1008);
```
adriano-stacks, line 105: This would result in a `panic` even in release mode, right? Why not having the asserts, but also clamping the value before to the required range to add defense in depth against an availability issue? 

This is probably a nit since a failed condition would likely prevent the signer from starting, I think.

signer/src/config/mod.rs
matteojug: We have validation on this checking it's positive, and we use a float because this is the fee rate in sat/vByte and can be less than 1.
We could use u64 scaling the fee to sat/vKByte, but since it's only a testing config we went with the simple float.

signer/src/transaction_coordinator.rs
matteojug: This is evaluated a compile time, so if the consts are such that we get a bad value in here the build will fail; no panics at runtime from this.

Initially we had a clamp, but it was pretty ugly (since `clamp` is not stable in const expressions yet), and instead we landed on having a safe computation of `FEE_RETRY_TARGET_BLOCKS` that fails for values out of range. Also note that this is computed from consts we define, so we don't risk unexpected failures from this.

signer/src/config/mod.rs
matteojug: Changed in [41e0503](https://github.com/stacks-sbtc/sbtc/pull/1988/commits/41e05033d2bb0f0419e27ab82afba4e5e8a7e088)

signer/src/transaction_coordinator.rs
adriano-stacks: Right it's compile time! I think that's fine then, thanks.

adriano-stacks (approved):

cylewitruk (approved):
LGTM ✅

PR #1990, opened 2026-03-27
feat: prepare for distroless images

djordon:
## Description

This is the meat of the coding changes necessary to do https://github.com/stacks-sbtc/sbtc/issues/1989

## Changes

* Update the `Dockerfile`s to produce binaries that could easily run on distroless images.
* Remove our custom attestation files.
* Update our mainnet, testnet, and devenv docker compose files so that signer logs are shipped using docker's fluentd log driver https://docs.docker.com/engine/logging/drivers/fluentd/. Update the docker compose so that vector expects fluentd logs.
* Turn off ansi coloring in logs if there isn't a terminal.

## Testing Information

Both the mainnet images and the devenv images build fine. Also the devenv seems to work just fine, although I only tried deposits.

## Checklist


docker/sbtc/Dockerfile
```diff
@@ -1,43 +1,56 @@
...
+    ARCH="$(uname -m)" && \
+    export "CC_${ARCH}_unknown_linux_musl=${ARCH}-linux-musl-gcc" && \
+    export "CXX_${ARCH}_unknown_linux_musl=${ARCH}-linux-musl-g++" && \
+    export "AR_${ARCH}_unknown_linux_musl=${ARCH}-linux-musl-ar" && \
+    export "CARGO_TARGET_$(echo "$MUSL_TARGET" | tr '[:lower:]-' '[:upper:]_')_LINKER=${ARCH}-linux-musl-gcc" && \
+    cargo build --target "$MUSL_TARGET" --bin signer --bin blocklist-client --bin emily-server && \
+    mkdir -p /output && \
+    cp "target/$MUSL_TARGET/debug/signer" /output/signer && \
+    cp "target/$MUSL_TARGET/debug/blocklist-client" /output/blocklist-client && \
+    cp "target/$MUSL_TARGET/debug/emily-server" /output/emily-server
```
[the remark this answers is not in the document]
cylewitruk: Or do something more like the following to make it configurable:

```dockerfile
ARG PROFILE=release
...

FROM ...
ARG PROFILE
...
RUN ...
    OUT_DIR=$([ "$PROFILE" = "dev" ] && echo "debug" || echo "$PROFILE") && \
    cargo build --profile "$PROFILE" --locked ...

```

docker/mainnet/docker-compose.yml
```diff
@@ -48,44 +48,31 @@ services:
...
-    command:
-      - /bin/bash
-      - -c
-      - |
-        envsubst < /signer-config.toml.in > /signer-config.toml
-        /usr/local/bin/signer -c /signer-config.toml --migrate-db -o json | tee /logs/sbtc_signer.logs
+    entrypoint:
+      - "/usr/local/bin/signer"
+      - "--config"
+      - "/signer-config.toml"
```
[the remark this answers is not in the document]
djordon: Nuts, I thought I caught that.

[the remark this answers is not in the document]
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/786169e2e18cd419d29c2ea95991d6ca06b1a3ee

docker/sbtc/Dockerfile
```diff
@@ -1,43 +1,56 @@
...
+    ln -s "/opt/${ARCH}-linux-musl-cross/bin/${ARCH}-linux-musl-g++" /usr/local/bin/ && \
+    ln -s "/opt/${ARCH}-linux-musl-cross/bin/${ARCH}-linux-musl-ar" /usr/local/bin/
+
+# Add the musl Rust target.
+RUN rustup target add "$(uname -m)-unknown-linux-musl"
+
 # Download the sbtc code base so that the initial build is done with
 # an appropriate version of the code and the result can be cached.
 WORKDIR /code
 RUN git clone https://github.com/stacks-sbtc/sbtc.git
```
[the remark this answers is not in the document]
djordon: Right, I removed the git clone in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/b2cf85155040a6d1e11df3496c75e9654e0eead5

blocklist-client/src/logging.rs
```diff
@@ -38,10 +39,11 @@ fn setup_logging_json() {
 }
 
 fn setup_logging_pretty() {
-    let dirs = "info,signer=debug";
+    let dirs = "info,blocklist-client=debug";
```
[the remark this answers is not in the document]
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/e6655d912979ab6b930b56cf7502db384108ee9b

docker/observability/vector.toml
```diff
@@ -29,35 +29,34 @@ address = "0.0.0.0:9598"
...
+.docker_source_type = del(.source_type)
 
-item.docker_container_name = .container_name
-item.docker_host = .host
-item.docker_image = .image
-. = item
+parsed, err = parse_json(.log)
+if err == null && is_object(parsed) {
+  . = merge(., object!(parsed))
+  del(.log)
```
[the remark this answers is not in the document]
djordon: I updated the logging comment in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/d23b9607e4db891c34339eddb19ca21d3dd28c5c, it's supposed to be the `log` field, as described in https://vector.dev/docs/reference/configuration/sources/fluent/. Also, you can see from the logs themselves.

docker/sbtc/Dockerfile
```diff
@@ -1,43 +1,56 @@
...
+    wget \
+    ca-certificates \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
 RUN npm install -g pnpm@9
 
+# Install the musl cross-toolchain from musl.cc.
+# This provides a complete musl-native gcc/g++/ar/libstdc++ for the target,
+# avoiding the glibc libstdc++ mismatch that occurs with Debian's g++.
+RUN ARCH="$(uname -m)" && \
+    wget -qO- "https://musl.cc/${ARCH}-linux-musl-cross.tgz" | tar xz -C /opt && \
```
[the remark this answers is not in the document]
djordon: I did the checksum thingy in this file as well in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/29280b9e14aa8d785744ceb22aa3ce4d55b8c1db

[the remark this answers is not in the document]
djordon: My approach is to just hardcode it, but I want to have a better overall approach. Either a trusted binary source for the toolchain, a trusted checksum provider, or just link dynamically to libc like we've always done.

docker/sbtc/Dockerfile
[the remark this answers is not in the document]
djordon: Do people make release builds for use on devenv? I've been assuming no, but maybe I'm mistaken?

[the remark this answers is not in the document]
cylewitruk: I couldn't remember if this was used at all in any release pipelines, but if it's only devenv then probably not worth it other than the ability to verify everything when built in release 🤷

.github/actions/dockerfiles/Dockerfile.signer.debian
```diff
@@ -7,89 +7,67 @@ RUN test -n "$GIT_COMMIT" || (echo "GIT_COMMIT not set" && false)
...
+    export "CC_${ARCH}_unknown_linux_musl=${ARCH}-linux-musl-gcc" && \
+    export "CXX_${ARCH}_unknown_linux_musl=${ARCH}-linux-musl-g++" && \
+    export "AR_${ARCH}_unknown_linux_musl=${ARCH}-linux-musl-ar" && \
+    export "CARGO_TARGET_$(echo "$MUSL_TARGET" | tr '[:lower:]-' '[:upper:]_')_LINKER=${ARCH}-linux-musl-gcc" && \
+    cargo build --target "$MUSL_TARGET" --bin signer ${CARGO_BUILD_ARGS} && \
+    mkdir -p /output && \
+    cp "target/$MUSL_TARGET/release/signer" /output/signer
 
+# distroless/static includes ca-certificates at /etc/ssl/certs/ca-certificates.crt.
+FROM gcr.io/distroless/static-debian13 AS signer
```
cylewitruk: Maybe we should digest-pin the base images (`@sha256:xxx`)?

cylewitruk (commented):
Looks largely good to me, just had the Q on digest-pinning base images

.github/actions/dockerfiles/Dockerfile.signer.debian
djordon: I think we should keep this as is for now, and potentially change it to pinned images if we set up a process for making sure that the hash stays up-to-date. I'm thinking that we set up a dependabot or something that updates the digest whenever there is a release upstream. We could still forget to update the digest, but it would be harder.

cylewitruk (dismissed):
lgtm

cylewitruk (dismissed):

signer/Cargo.toml
```diff
@@ -32,7 +32,7 @@ cfg-if.workspace = true
 clap.workspace = true
 clarity.workspace = true
 config.workspace = true
-emily-client.workspace = true
+emily-client = { workspace = true, features = ["rustls"] }
```
matteojug: qq: do we need these changes?

docker/docker-compose.yml
```diff
@@ -181,6 +187,12 @@ services:
   sbtc-signer-1:
     <<: *sbtc-signer # Inherit all from the "sbtc-signer" service
     container_name: sbtc-signer-1
+    logging:
+      driver: fluentd
+      options:
+        fluentd-address: "localhost:24224"
+        fluentd-async: "true"
+        tag: "sbtc-signer-1"
```
matteojug: We can move this under the common `sbtc-sigener` using `tag: "{{.Name}}"` (https://docs.docker.com/engine/logging/log_tags/)

docker/mainnet/docker-compose.yml
```diff
@@ -48,44 +48,31 @@ services:
...
     image: ghcr.io/stacks-sbtc/sbtc:signer-v1.0.2@sha256:60832405b3f56e8c8ae9ac9429866ce5173f86ff79652bad7f38481f220b2332
-    entrypoint: "/entrypoint.sh"
-    command:
-      - /bin/bash
-      - -c
-      - |
-        envsubst < /signer-config.toml.in > /signer-config.toml
-        /usr/local/bin/signer -c /signer-config.toml --migrate-db -o json | tee /logs/sbtc_signer.logs
+    entrypoint:
+      - "/usr/local/bin/signer"
```
matteojug, line 46: I think this (and the change in the vector config) qualifies as a breaking local change, as the current compose people are using right now will no longer work. If so, can you add the label to the PR?

docker/mainnet/docker-compose.yml
```diff
@@ -48,44 +48,31 @@ services:
...
+      - "--config"
+      - "/signer-config.toml"
+      - "--migrate-db"
+      - "--output-format"
+      - "json"
     depends_on:
       - postgres
+    logging:
+      driver: fluentd
```
matteojug, line 55: Do we know if everybody is running the signer with docker compose (or the raw binary), or someone is using k8s or other runtimes?

docker/mainnet/docker-compose.yml
```diff
@@ -48,44 +48,31 @@ services:
...
-      TRUSTED_ROOT_PATH: trusted_root.jsonl
+      SIGNER_BITCOIN__RPC_ENDPOINTS: "http://${BITCOIN_RPC_USERNAME}:${BITCOIN_RPC_PASSWORD}@${BITCOIN_RPC_HOST}:${BITCOIN_RPC_PORT}"
+      SIGNER_STACKS__ENDPOINTS: "http://${STACKS_RPC_HOST}:${STACKS_RPC_PORT}"
+      SIGNER_SIGNER__PRIVATE_KEY: "${SIGNER_SIGNER__PRIVATE_KEY}"
+      SIGNER_EMILY__ENDPOINTS: "https://${EMILY_API_KEY}@sbtc-emily.com"
+      SIGNER_SIGNER__DB_ENDPOINT: "postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}"
+      TAG: *SIGNER_TAG
     volumes:
-      - ./sbtc-signer/signer-config.toml.in:/signer-config.toml.in:ro
-      - ./logs:/logs
```
matteojug, line 85: We should double check if people are using the logs for their own monitoring, and if so give them an heads-up to move to fluentd

docker/mainnet/docker-compose.yml
```diff
@@ -48,44 +48,31 @@ services:
...
-      POSTGRES_PORT: *POSTGRES_PORT
-      TAG: *SIGNER_TAG  
-      BUNDLE_PATH: signer_bundle.jsonl
-      TRUSTED_ROOT_PATH: trusted_root.jsonl
+      SIGNER_BITCOIN__RPC_ENDPOINTS: "http://${BITCOIN_RPC_USERNAME}:${BITCOIN_RPC_PASSWORD}@${BITCOIN_RPC_HOST}:${BITCOIN_RPC_PORT}"
+      SIGNER_STACKS__ENDPOINTS: "http://${STACKS_RPC_HOST}:${STACKS_RPC_PORT}"
+      SIGNER_SIGNER__PRIVATE_KEY: "${SIGNER_SIGNER__PRIVATE_KEY}"
+      SIGNER_EMILY__ENDPOINTS: "https://${EMILY_API_KEY}@sbtc-emily.com"
+      SIGNER_SIGNER__DB_ENDPOINT: "postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}"
+      TAG: *SIGNER_TAG
```
matteojug: qq: are we still using `TAG` with the removal of the attestation stuff? Also for blocklist client, and also in testnet compose

docker/mainnet/sbtc-signer/signer-config.toml
matteojug, line 1: Should we remove the `.in` suffix from this, now that is a "proper" config?

.github/actions/dockerfiles/Dockerfile.signer.debian
```diff
@@ -7,89 +7,67 @@ RUN test -n "$GIT_COMMIT" || (echo "GIT_COMMIT not set" && false)
...
+    wget \
+    ca-certificates \
+    findutils \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
-RUN npm install -g pnpm@9 && \
-    npm install -g @openapitools/openapi-generator-cli
+# Install the musl cross-toolchain from musl.cc for the target architecture.
+# This provides a complete musl-native gcc/g++/ar/libstdc++ for the target,
+# avoiding the glibc libstdc++ mismatch that occurs with Debian's g++.
```
matteojug: nit: there are some comments in this one that are not in the blocklist client dockerfile. Can you copy them in the other one, so they matches?

matteojug (commented):
Generally LGTM, but I only did a test run with devenv.

We should do a testnet run with the images before merging. We should be able to use https://github.com/stacks-sbtc/sbtc/actions/workflows/nightly-image-build.yaml from this branch already to get a nightly and test in on testnet (after OIDC is fixed)

signer/Cargo.toml
djordon: Nope, will update

docker/docker-compose.yml
djordon: Oh, nice. I didn't see that this was available, will do.

docker/mainnet/sbtc-signer/signer-config.toml
djordon: Yeah, I think so.

docker/mainnet/docker-compose.yml
djordon: Yes, definitely. We'll need to reach out to let people know that this release has some changes that they'll need to pay attention to.

docker/mainnet/docker-compose.yml
djordon: Oh nice, looks like it can be removed. Will update

.github/actions/dockerfiles/Dockerfile.signer.debian
djordon: sure thing

djordon:
We might need to rethink the plan here, pulling the musl toolchain fails for me in CI: https://github.com/djordon/sbtc/actions/runs/24016372614.

docker/mainnet/docker-compose.yml
djordon: Not everyone is running docker compose, some build from source, and some do other things. Why?

signer/Cargo.toml
djordon: Oh this was from earlier changes that got reverted in the other branch. Fixed in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/0b39a54e4a786097259152131227bb207ff35ded and https://github.com/stacks-sbtc/sbtc/pull/1990/commits/2753039fc05e320483f43119694fa64bb7ead388.

docker/docker-compose.yml
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/97890ce60265a73ff6b57e7b73cfbf93a131dd4a

docker/mainnet/docker-compose.yml
djordon: I tagged the PR

docker/mainnet/docker-compose.yml
djordon: Removed them in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/95986d27c6da7f7599fd5d1ab62bfefe79fd1edb and https://github.com/stacks-sbtc/sbtc/pull/1990/commits/736c3c2c92aca0410bb21a17bfe3d1f9c3c76f11

docker/mainnet/sbtc-signer/signer-config.toml
djordon: Renamed in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/953a3ba2dec33b0f974432500e4603adbf47bca3

.github/actions/dockerfiles/Dockerfile.signer.debian
djordon: I switched back to linking to glibc and made tried to make sure that the two dockerfiles reasonably match.

docker/mainnet/docker-compose.yml
matteojug: I was wondering because of fluentd/lack of file logging. Building from source should be fine -- they can keep doing whatever they were doing -- but for other container runtimes fluentd may not be available out of the box.

djordon: Oh right. Yeah, the hope is that we catch this as part of our due diligence process. I don't think this change will matter for people who aren't deploying through docker-compose but you never know.

docker/sbtc/Dockerfile
```diff
@@ -47,20 +32,16 @@ RUN make emily-cdk-synth
...
 # ---------------------------------------
-FROM debian:bookworm-slim AS signer
+# distroless/cc includes glibc, libgcc, libstdc++, ca-certificates.
+FROM gcr.io/distroless/cc-debian13 AS signer
 COPY --from=builder /code/sbtc/target/debug/signer /usr/local/bin/signer
-RUN apt-get update \
-    && apt-get install -y ca-certificates --no-install-recommends \
-    && apt-get clean && rm -rf /var/lib/apt/lists/*
-ENTRYPOINT ["/usr/local/bin/signer --config /signer-config.toml --migrate-db"]
+ENTRYPOINT ["/usr/local/bin/signer", "--config", "/signer-config.toml", "--migrate-db"]
```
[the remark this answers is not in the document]
djordon: I want debug builds here

.dockerignore
```diff
@@ -5,3 +5,4 @@ docker
 .gitmodules
 .gitworktrees
 node_modules
+.cargo/config.toml
```
matteojug: nit: no longer needed

docker/sbtc/Dockerfile
```diff
@@ -1,43 +1,28 @@
...
-RUN rm -rf sbtc signer emily emily_sidecar emily_cron blocklist-client blocklist-openapi-gen .generated-sources protobufs
+WORKDIR /code/sbtc
 COPY . /code/sbtc/
-RUN make install-pnpm && make build
 
 # Build Emily Server.
 # --------------------
 ARG EMILY_SERVER_BUILD_ARGS=""
-RUN cargo build --bin emily-server $EMILY_SERVER_BUILD_ARGS
+RUN cargo build --bin signer --bin blocklist-client --bin emily-server
```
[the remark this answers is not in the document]
djordon: Yeah, I noticed that and didn't get around to updating it. Will do

blocklist-client/src/logging.rs
```diff
@@ -38,10 +39,11 @@ fn setup_logging_json() {
 }
 
 fn setup_logging_pretty() {
-    let dirs = "info,signer=debug";
+    let dirs = "info,blocklist_client=debug";
 
     let main_layer = tracing_subscriber::fmt::layer()
         .pretty()
+        .with_ansi(std::io::stdout().is_terminal())
```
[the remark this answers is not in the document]
djordon: Hmmm, I don't think it matters but will check.

signer/src/logging.rs
```diff
@@ -41,7 +42,9 @@ fn setup_logging_json(directives: &str) {
 }
 
 fn setup_logging_pretty(directives: &str) {
-    let main_layer = tracing_subscriber::fmt::layer().with_timer(UtcTime::rfc_3339());
+    let main_layer = tracing_subscriber::fmt::layer()
+        .with_ansi(std::io::stdout().is_terminal())
```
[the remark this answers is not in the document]
djordon: I don't think it matters but will check.

docker/sbtc/Dockerfile
```diff
@@ -1,43 +1,28 @@
...
     git \
     pkg-config \
     libssl-dev \
     make \
     protobuf-compiler \
     npm \
-    default-jre
+    default-jre \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
 RUN npm install -g pnpm@9
```
matteojug, line 10: Now that we don't build anything outside cargo build, do we still need npm/pnpm/jre/protobuf-compiler/git/make?

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
```diff
@@ -1,74 +1,54 @@
...
-
-RUN \
-    export LIBCLANG_PATH="$(find /usr/lib/llvm-* -name lib -type d | head -n 1)" && \
+# Build the blocklist-client binary, cross-compiling for the target architecture.
+RUN export LIBCLANG_PATH="$(find /usr/lib/llvm-* -name lib -type d | head -n 1)" && \
     if [ "$TARGETARCH" = "arm64" ]; then \
-        echo "Building for ARM64..."; \
-        export PKG_CONFIG_ALLOW_CROSS=1 && \
-        rustup target add aarch64-unknown-linux-gnu; \
+        TARGET="aarch64-unknown-linux-gnu" && \
```
matteojug: `TARGET` is not exported (also below and in the other file)

.dockerignore
djordon: Removed in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/50e5b6d72da398ff657b52ffe27b23ffde58d0f3

docker/sbtc/Dockerfile
djordon: I think this still does `make emily-cdk-synth`, which does a `pnpm --filter run synth`. I didn't check if that means that we can remove jre, but we can remove protobuf and probably git. I will try it out now.

djordon: BTW, I will remove the `make emily-cdk-synth` here so that we can remove as much as possible in the Dockerfile.

docker/sbtc/Dockerfile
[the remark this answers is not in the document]
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/d229129826d48b2b4fe0f41fabf455750809af6d

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
djordon: I don't think any of these variables need to be, since the `&&` caries them in the same process all the way through to the cargo build command. I suppose I should remove all of the exports, or make sure they are all exported, for consistency. Also, if I was wrong the build wouldn't succeed since copying the binary would fail.

docker/sbtc/Dockerfile
```diff
@@ -1,43 +1,28 @@
...
-RUN rm -rf sbtc signer emily emily_sidecar emily_cron blocklist-client blocklist-openapi-gen .generated-sources protobufs
+WORKDIR /code/sbtc
 COPY . /code/sbtc/
-RUN make install-pnpm && make build
 
 # Build Emily Server.
 # --------------------
 ARG EMILY_SERVER_BUILD_ARGS=""
-RUN cargo build --bin emily-server $EMILY_SERVER_BUILD_ARGS
+RUN cargo build --bin signer --bin blocklist-client --bin emily-server
```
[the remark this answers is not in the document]
djordon: I added in the `--locked` in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/8d4e8f342a151464eb88820695a58830f8279874

docker/sbtc/Dockerfile
djordon: I removed a bunch of unnecessary dependencies in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/fedaf5cd22ba97650bab3be45d5f6415569c7028 and https://github.com/stacks-sbtc/sbtc/pull/1990/commits/fedaf5cd22ba97650bab3be45d5f6415569c7028. Tested it out, and things built fine. I'm waiting on https://github.com/djordon/sbtc/actions/runs/24094489619/job/70289317135 to test this out on devenv.

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
djordon: So, it looks like I was wrong or something, since I tried to make a release build and it failed (https://github.com/djordon/sbtc/actions/runs/24094489619), then tried another where everything was exported and it succeeded (https://github.com/djordon/sbtc/actions/runs/24095031136).

Anyway, I tested the latest image on local devenv and deposits and withdrawals worked without issues.

djordon: The commit was https://github.com/stacks-sbtc/sbtc/pull/1990/commits/2c6cdfd2ee6561cd040e8388e2fe128f90919acc

signer/src/logging.rs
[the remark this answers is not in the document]
djordon: No, this appears to be incorrect, it doesn't write to stderr by default, it writes to stdout. I checked https://github.com/tokio-rs/tracing/blob/tracing-subscriber-0.3.19/tracing-subscriber/src/fmt/mod.rs#L333-L336 and https://github.com/tokio-rs/tracing/blob/tracing-subscriber-0.3.19/tracing-subscriber/src/fmt/fmt_layer.rs#L713-L729. Also, I don't think it matters in our case.

blocklist-client/src/logging.rs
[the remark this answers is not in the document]
djordon: Same response as in https://github.com/stacks-sbtc/sbtc/pull/1990#discussion_r3047006730

docker/sbtc/Dockerfile
```diff
@@ -1,66 +1,42 @@
-FROM rust:1.94.0-slim-bookworm AS builder
+FROM rust:1.94.1-slim-trixie AS builder
```
cylewitruk, line 1: Maybe consider specifying the builder/signer/etc. images as `ARG`s with defaults so they can be overridden in e.g. CI if we needed/wanted to? (general comment for all `Dockerfile`s, feel free to resolve if you don't want to do it)

docker/testnet/docker-compose.testnet.yml
```diff
@@ -19,28 +11,31 @@ services:
 
   sbtc-signer:
     image: blockstack/sbtc:signer-0.0.9-rc6@sha256:44d2c828c6452f75ea2fabe5160ba95a9f944c0fcaeeaea5523dff81da8bc342
-    entrypoint: "/entrypoint.sh"
-    command: "/bin/bash -c '/usr/local/bin/signer -c /signer-config.toml --migrate-db -o json | tee /logs/sbtc_signer.logs'"
+    entrypoint:
+      - "/usr/local/bin/signer"
```
cylewitruk, line 15: Did we have a reason for not just specifying this in the `Dockerfile` target's `ENTRYPOINT`? 🤔

signer/src/logging.rs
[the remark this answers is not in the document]
cylewitruk: iirc there was some logic to output warn/error to stderr (didn't look at the source, and I could be wrong) -- but yeah, I don't think this matters in any case since prod workloads run with `--json`.

docker/mainnet/docker-compose.yml
cylewitruk: Ideally some clear instructions/guidance as well, i.e. some updates to the docs page. Maybe @fabergat has input?

fabrizio-stacks: Mmm are you sure is it a good solution to enforce a log system? Organizations can have different solution and maybe not super happy to adapt

docker/sbtc/Dockerfile
cylewitruk: @djordon I noticed the build failed -- is that something you're still looking at/into?

docker/mainnet/docker-compose.yml
djordon: Yeah, we're not enforcing anything, but I take @cylewitruk's comment to be that we should make sure to update our official docs if it's applicable. I don't think there is anything to update there so I think we are good. The docs are at https://docs.stacks.co/operate/run-a-sbtc-signer/best-practices-for-running-an-sbtc-signer

docker/testnet/docker-compose.testnet.yml
djordon: Specifying what exactly? The only thing that isn't in the Dockerfiles's entrypoint is the directive for specifying the log output format.

docker/sbtc/Dockerfile
djordon: I don't really want to do it. As is, there is less indirection and in this case I'm okay with updating these files.

docker/sbtc/Dockerfile
djordon: Yeah, I fixed it up and tried it again. I forgot to update this thread but did in https://github.com/stacks-sbtc/sbtc/pull/1990#discussion_r3046895144

docker/testnet/docker-compose.testnet.yml
cylewitruk: I was just wondering why we even need to specify the `entrypoint` in the docker compose?

docker/mainnet/docker-compose.yml
cylewitruk: > We'll need to reach out to let people know that this release has some changes that they'll need to pay attention to.

👆 This is primarily what I was reacting to; or are we saying that it isn't going to *require* changes for current sbtc operators?

docker/testnet/docker-compose.testnet.yml
djordon: Ohhh, well kinda, in order to specify the log output format.

cylewitruk: Resolving, this was mostly a question

docker/mainnet/docker-compose.yml
cylewitruk: Got clarification offline that the idea here is to ensure that this is included in release notes, so my concern regarding docs is resolved; should this receive a "breaking" label?  @matteojug this comment is yours from the beginning so leaving to you to resolve; I'm satisfied, anyway.

cylewitruk (dismissed):
Looks good to me code-wise, but disclaimer that I have not tested these locally or in any other environment

docker/mainnet/docker-compose.yml
matteojug: I'm generally okay with requiring fluentd/container logs instead of file based ones, I just wanted to be sure we communicate this change thoroughly because operators may need to adjust behind the simple update docker-compose/vector config.

How will this work in case of machine/container crash? Especially for operators that are not sending us the logs via vector

djordon: > How will this work in case of machine/container crash? Especially for operators that are not sending us the logs via vector

What do you mean, how will what work?

Right now I only see the general concern that we need to make sure that operators know that our previous recommendation on how to handle logging when run in docker compose won't work.

matteojug: Yeah sorry, I was wondering if not having file logging will bite us in the future, in particular for operators not sending us the logs via vector. But maybe it would be enough to flag it, so that operators can persist logs in some other way.

matteojug (commented):
LGTM, I'm going to build and test this on testnet

djordon:
I updated this branch to produce Debian images, so upgrades for signer operators will remain the same as they always have. I cut a new release based on the latest commit (https://github.com/stacks-sbtc/sbtc/pull/1990/commits/a89059b144fa9e9aff5dcb550520babfef7a8295), producing https://github.com/djordon/sbtc/pkgs/container/sbtc/794633910?tag=signer-v1.3.0-rc9. I tested this image locally on devenv, and deposits and withdrawals went smoothly.

docker/mainnet/sbtc-signer/signer-config.toml
```diff
@@ -81,7 +81,7 @@ db_endpoint = "postgresql://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST:$PO
 # Required: true
 # Environment: SIGNER_SIGNER__PRIVATE_KEY
 # NOTE: Set through Environment: SIGNER_SIGNER__PRIVATE_KEY
```
matteojug-stacks: nit: we could remove the `NOTE` here

docker/mainnet/docker-compose.yml
```diff
@@ -48,44 +42,30 @@ services:
     restart: on-failure
     # this version can be outdated, check the last image on: https://github.com/stacks-sbtc/sbtc/pkgs/container/sbtc
     image: ghcr.io/stacks-sbtc/sbtc:signer-v1.0.2@sha256:60832405b3f56e8c8ae9ac9429866ce5173f86ff79652bad7f38481f220b2332
```
matteojug-stacks: nit: we should update this to the latest release, in case someone notices the change and (blindly) update the compose; also for the blocklist-client image.

docker/sbtc/Dockerfile
```diff
@@ -1,66 +1,42 @@
...
-RUN rm -rf sbtc signer emily emily_sidecar emily_cron blocklist-client blocklist-openapi-gen .generated-sources protobufs
+WORKDIR /code/sbtc
 COPY . /code/sbtc/
-RUN make install-pnpm && make build
 
 # Build Emily Server.
 # --------------------
 ARG EMILY_SERVER_BUILD_ARGS=""
-RUN cargo build --bin emily-server $EMILY_SERVER_BUILD_ARGS
+RUN cargo build --locked --bin signer --bin blocklist-client --bin emily-server $EMILY_SERVER_BUILD_ARGS
```
matteojug-stacks: nit: this is no longer building "only" the emily server (in the comment).
Also, `EMILY_SERVER_BUILD_ARGS` may affect all the binaries; it's not an issue since we currently use that flag only in the testing compose (and in there we only start Emily), but maybe we should split the build in two commands so that we can apply the arg only to emily

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
```diff
@@ -1,74 +1,55 @@
...
+        export CC_x86_64_unknown_linux_gnu=x86_64-linux-gnu-gcc && \
+        export CXX_x86_64_unknown_linux_gnu=x86_64-linux-gnu-g++ && \
+        export AR_x86_64_unknown_linux_gnu=x86_64-linux-gnu-ar && \
+        export CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER=x86_64-linux-gnu-gcc; \
+    fi && \
+    cargo build --target "$TARGET" --bin blocklist-client --release --locked && \
+    mkdir -p /output && \
+    cp "target/$TARGET/release/blocklist-client" /output/blocklist-client
+
+# distroless/cc includes glibc, libgcc, libstdc++, ca-certificates.
```
matteojug-stacks: nit: we are not using distroless/cc (yet); also in the signer dockerfile

matteojug-stacks (dismissed):
LGTM, and I'm okay with merging this and waiting to nightly to build it to test it.

Sanity check: we don't want/need to run the codegen and check it's not dirty before building, as we are already doing that in the usual on-push job so we can assume it's clean, right?

docker/sbtc/Dockerfile
djordon-stacks: Right, I'll move to two commands. I assumed the args applied only to Emily, but I realized that might not be correct and didn't revisit it.

.github/actions/dockerfiles/Dockerfile.blocklist-client.debian
djordon-stacks: oops forgot to update, will do

djordon-stacks: Removed in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/d0e2401bc5b6049bee6189cd1ba54e9d321bc17d

docker/sbtc/Dockerfile
djordon-stacks: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/26d38b370502589a64489f36b81dd950f1b173de

docker/mainnet/docker-compose.yml
djordon-stacks: Updated in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/a876fd59be0ed6917f7039660b742ed90957e896

docker/mainnet/sbtc-signer/signer-config.toml
djordon-stacks: Removed in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/bf7ed78f1a9b3cd75a870742dd14409f1f5b119a

djordon-stacks:
> LGTM, and I'm okay with merging this and waiting to nightly to build it to test it.
> 
> Sanity check: we don't want/need to run the codegen and check it's not dirty before building, as we are already doing that in the usual on-push job so we can assume it's clean, right?

Yes, I just double checked, and we build the generated sources on pushes to our PRs and to `main`, and run diff checks for them. The merge queue gates merging PRs onto main after a successful run of our checks on a newly generated merge commit. So building against `main` should be fine, since we have all the checks in place (unless we disable them).

matteojug-stacks (dismissed):

docker/testnet/vector/vector.yaml
```diff
@@ -1,15 +1,35 @@
 sources:
-  sbtc_signer_logs:
-    type: file
-    include:
-      - /logs/sbtc_signer.logs
+  sbtc_signer_fluentd:
+    type: fluent
+    address: 0.0.0.0:24224
```
cylewitruk-stacks, line 4: Do we want to use 0.0.0.0 instead of an explicit bind?

docker/testnet/docker-compose.testnet.yml
```diff
@@ -19,28 +11,31 @@ services:
...
+      - "/signer-config.toml"
+      - "--migrate-db"
+      - "--output-format"
+      - "json"
     depends_on:
       - postgres
+    logging:
+      driver: fluentd
+      options:
+        fluentd-address: "localhost:24224"
```
cylewitruk-stacks: We use `127.0.0.1` in some other places, should we prefer that over `localhost` (here and elsewhere)?

cylewitruk-stacks (dismissed):
Approving, but had a couple of questions regarding consistency and our previous decision to avoid global binds.

docker/testnet/vector/vector.yaml
djordon-stacks: What do you mean?

cylewitruk-stacks: 0.0.0.0 binds to all addresses (if a bind), or is otherwise invalid as a target address (not sure which this actually means -- I assume `target` now looking at the removed line). That is ofc unless vector has some special handling of 0.0.0.0 as a target address.

docker/testnet/docker-compose.testnet.yml
djordon-stacks: Oh yeah, we should probably use 127.0.0.1. I'll make the update.

djordon-stacks: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1990/commits/7e78c57ae0e2a187633d8b7e74970e513af99e04

cylewitruk-stacks (approved):

docker/testnet/vector/vector.yaml
djordon-stacks: Yeah, I think it's a bind address, and we want it to be at `0.0.0.0` because this binding happens inside the container, and I don't think `127.0.0.1` will work.

matteojug-stacks (approved):

PR #1994, opened 2026-04-01
fix: `BitcoinPreSignRequest` can be too large to broadcast

### LINKED ISSUE #1992: [Bug]: `BitcoinPreSignRequest` can be too large to broadcast
# Bug - `BitcoinPreSignRequest` can be too large to broadcast

## 1. Description

The bitcoin packager is responsible for taking outstanding deposit and withdrawals requests and organizing them in such a way where the transaction adheres to bitcoin's standardness rules, as well as internal heuristics that make it likely that the transaction package will be signed before the arrival of the next bitcoin block. The packager follows those rules, but neglects the signers networking rules that prevent broadcast of messages that are too large. If there are too many pending deposit and withdrawal requests, the packager will not filter enough requests, leading to a `BitcoinPreSignRequest` that is larger than libp2p's `max_transmit_size`. When that happens libp2p will not broadcast the message.

I believe `BitcoinPreSignRequest` messages are the only ones that can be too large given our `max_transmit_size`, but if there are others, we need to make sure that they are always small enough as well.

## 2. Technical Details

When there are too many requests leading to the signer constructing a `BitcoinPreSignRequest` message that exceeds libp2p's `max_transmit_size`, then that message will not be broadcast to the network. This leads to a temporary outage, where no new transactions will be confirmed on the network until enough deposit and withdrawal requests "expire".

The fix is to have the packager consider the sBTC signer network size of each included request when packaging things up and makes sure that the final network message that envelopes the `BitcoinPreSignRequest` is small enough to transmit.

### 2.1 Acceptance Criteria

- [ ] Constructed `BitcoinPreSignRequest` messages are always small enough to be broadcast to the signer network

djordon:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/1992

I think I want to make some updates to the comments to make things a little clearer, specifically around what gets counted toward libp2p's limits, but this is good enough for review.

## Changes

* Set the `max_transmit_size` in libp2p-gossipsub. This value limits the size that we broadcast and receive. The value chosen is the current default, and is set explicitly to make it more obvious what is value is used.
* Make sure that the packager takes into consideration how much space a deposit or withdrawal request will take in the `BitcoinPreSignRequest`.

It's worth noting that the `max_transmit_size` affects what the signer will publish but also receive. I haven't pieced it all together yet, but will do 

## Testing Information

I still need to test this in devenv.

## Checklist


signer/src/bitcoin/packaging.rs
```diff
@@ -3,12 +3,51 @@
...
+const SIGNED_MESSAGE_OVERHEAD: usize = 512;
+
+/// Maximum serialized size of a `BitcoinPreSignRequest` message, assuming
+/// 0 as the `fee_rate` and no `last_fees`.
+///
+/// The gossipsub [`GOSSIPSUB_MAX_TRANSMIT_SIZE`] limits the total
+/// encoded `Signed<SignerMessage>` size. After subtracting the
+/// [`SIGNED_MESSAGE_OVERHEAD`] for the `Signed` and `SignerMessage`
+/// protobuf wrapper, this is the remaining budget for the
+/// `BitcoinPreSignRequest` payload.
```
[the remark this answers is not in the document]
djordon: Ooops, forgot to update that.

signer/src/bitcoin/packaging.rs
```diff
@@ -3,12 +3,51 @@
...
 
+/// Protobuf overhead per bag (per `TxRequestIds` element) in the
+/// `BitcoinPreSignRequest.request_package` repeated field.
+///
+/// Each `TxRequestIds` embedded in the `request_package` repeated field
+/// incurs a 1-byte field tag and a length-delimiting varint. For small
+/// bags (encoded inner size < 128 bytes) the length varint is 1 byte,
+/// giving 2 bytes total. For larger bags (≥ 128 bytes) the length varint
+/// grows to 2 bytes, giving 3 bytes total. We use the worst case.
+const BAG_OVERHEAD: usize = 3;
```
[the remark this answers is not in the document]
djordon: Oh right, and my test masked this from happening.

signer/src/bitcoin/packaging.rs
```diff
@@ -3,12 +3,93 @@
...
+///
+///    Wrapped into an `RPC` frame (one `Message` per `RPC`):
+///    <https://github.com/libp2p/rust-libp2p/blob/84153a559bdbcb92a48413dd2a31035800cb882d/protocols/gossipsub/src/types.rs#L345-L356>
+///
+/// The overhead varies slightly with the inner payload size because
+/// protobuf length varints grow with the encoded length. A
+/// Signed<SignerMessage> with an empty `BitcoinPreSignRequest` measures
+/// 162 bytes of overhead; with a near-maximum-size payload. The libp2p
+/// gossipsub implementation adds an additional 255 bytes of overhead.
+const SIGNED_MESSAGE_OVERHEAD: usize = 512;
```
matteojug: qq: is `512` an arbitrary upper bound to the components mentioned in the comment?

signer/src/bitcoin/packaging.rs
```diff
@@ -3,12 +3,93 @@
...
+///    Converted to `proto::Message`:
+///    <https://github.com/libp2p/rust-libp2p/blob/84153a559bdbcb92a48413dd2a31035800cb882d/protocols/gossipsub/src/types.rs#L205-L216>
+///
+///    Wrapped into an `RPC` frame (one `Message` per `RPC`):
+///    <https://github.com/libp2p/rust-libp2p/blob/84153a559bdbcb92a48413dd2a31035800cb882d/protocols/gossipsub/src/types.rs#L345-L356>
+///
+/// The overhead varies slightly with the inner payload size because
+/// protobuf length varints grow with the encoded length. A
+/// Signed<SignerMessage> with an empty `BitcoinPreSignRequest` measures
+/// 162 bytes of overhead; with a near-maximum-size payload. The libp2p
```
matteojug: _with a near-maximum-size payload_ is missing the overhead?

signer/src/bitcoin/packaging.rs
```diff
@@ -3,12 +3,93 @@
...
+/// gossipsub implementation adds an additional 255 bytes of overhead.
+const SIGNED_MESSAGE_OVERHEAD: usize = 512;
+
+/// Maximum serialized size of a `BitcoinPreSignRequest` message, assuming
+/// 0 as the `fee_rate` and no `last_fees`.
```
matteojug: Shouldn't we consider the max presign size assuming max (len wise) fee_rate and last_fee?
It could already be the case and just the comment being outdated

signer/src/bitcoin/packaging.rs
```diff
@@ -1183,4 +1344,341 @@ mod tests {
...
+    /// BAG_OVERHEAD`.
+    #[test]
+    fn test_insert_item_respects_presign_size_limit() {
+        let config = PackagerConfig::new(5, 10000);
+        let mut packager = BestFitPackager::<RequestItem>::new(config);
+
+        // Each deposit has presign weight DEPOSIT_PRESIGN_WEIGHT (48 bytes).
+        // The first item also creates a bag, which charges BAG_OVERHEAD.
+        // Fill up to just under the limit.
+        let max_deposits = (MAX_PRESIGN_REQUEST_SIZE - BAG_OVERHEAD) / DEPOSIT_PRESIGN_WEIGHT;
```
matteojug, line 1367: nit: can we also add a sanity check that max_deposits is less than 10000? To be sure we are not being limited by max signatures when adding the extra one

signer/src/bitcoin/packaging.rs
```diff
@@ -1183,4 +1344,341 @@ mod tests {
...
+                max_fee: 10_000,
+                signer_bitmap: BitArray::ZERO,
+                amount: 100_000,
+                deposit_script: bitcoin::ScriptBuf::new(),
+                reclaim_script_hash: TaprootScriptHash::zeros(),
+                signers_public_key,
+            };
+
+            let proto_outpoint = proto::OutPoint::from(deposit.outpoint);
+            let encoded_length = proto_outpoint.encode_to_vec().len();
```
matteojug, line 1413: nit? also for the other test
```suggestion
            let encoded_length = proto_outpoint.encoded_len();
```

signer/src/bitcoin/packaging.rs
```diff
@@ -1183,4 +1344,341 @@ mod tests {
...
+                        txid: bitcoin::Txid::from_byte_array([0xff; 32]),
+                        vout: u32::MAX,
+                    },
+                    deposits_needed,
+                )
+                .collect(),
+                withdrawals: Vec::new(),
+            }],
+            fee_rate: 25.0,
+            last_fees: Some(Fees { total: 1000000, rate: 25.0 }),
```
matteojug: nit: can we use `u64::MAX` for `total`?

signer/src/bitcoin/packaging.rs
```diff
@@ -1183,4 +1344,341 @@ mod tests {
...
+                let mut deposits = Vec::new();
+                let mut withdrawals = Vec::new();
+                for item in bag {
+                    match item {
+                        RequestRef::Deposit(d) => deposits.push(d.outpoint),
+                        RequestRef::Withdrawal(w) => withdrawals.push(w.qualified_id()),
+                    }
+                }
+                TxRequestIds { deposits, withdrawals }
+            })
```
matteojug: nit
```suggestion
            .map(|bag| (&crate::bitcoin::utxo::Requests::new(bag)).into())
```

signer/src/bitcoin/packaging.rs
```diff
@@ -1183,4 +1344,341 @@ mod tests {
...
+            .collect();
+
+        // In local runs the bag count has been between 18 and 20.
+        let num_bags = dbg!(request_package.len());
+        // The packager tries to respect the pre-sign request serialization
+        // limits only when `fee_rate` is 0.0 and there are no last fees,
```
matteojug: Can we rephrase this comment? Maybe something like _`total_presign_size` doesn't account for fee rate or last fees_. Otherwise it seems the packager doesn't enforce anything if we have fee rate or last fees.

matteojug (commented):
Generally LGTM

Can we add a test to check that the gossipsub full wire frame (+ signed message & co) overhead is indeed less than `SIGNED_MESSAGE_OVERHEAD`? Otherwise, IIUC, we are only covering the sending side with the current tests

signer/src/bitcoin/packaging.rs
djordon: Yes. Also this comment needs updating. I think the actual upper bound is like 370 (comes from 168 from our `Signed<SignedMessage>` plus 200 from libp2p). I'm going to change this to 1024 to give us more breathing room just in case I'm wrong somewhere but not badly wrong.

signer/src/bitcoin/packaging.rs
djordon: That was a sentence fragment. It should say "with a near-maximum-size `BitcoinPreSignRequest` payload, the Signed<SignerMessage> measures 168 bytes of overhead.

signer/src/bitcoin/packaging.rs
djordon: Well, that was what I originally wanted to do but I thought it made the code a little harder to follow and reason about. I changed it so that now the packager accounts for the parts of the `BitcoinPreSignRequest` that it can control, and moved the fee rate and last fee accounting into the overhead.

signer/src/bitcoin/packaging.rs
djordon: this is kind of a check that `encoded_len` works the way that we think. So I think we want to use the actual encoded vector length instead of `encoded_len` since we use `encoded_len` to compute the `presign_weight`.

signer/src/bitcoin/packaging.rs
djordon: Oh, didn't remember that we had that 😅.

signer/src/bitcoin/packaging.rs
djordon: Ahh, yeah will do.

djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1994/commits/b7a01cc139dcd6c0f7c21afd8891078931c4643c

signer/src/bitcoin/packaging.rs
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1994/commits/6c08ecd562f5f2324d0bb3815b99c38fcc7f29c0

signer/src/bitcoin/packaging.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1994/commits/3e75b44a4acf2b7f271b78cf40e18641c9d477f4

signer/src/bitcoin/packaging.rs
djordon: Updated in https://github.com/stacks-sbtc/sbtc/pull/1994/commits/5ad7c9de89458e53da03fbf142ab657b9953614e

signer/src/bitcoin/packaging.rs
matteojug: Yeah that makes sense.

The comment seems to suggest that `MAX_PRESIGN_REQUEST_SIZE` is not the worst case bound though, as in if we have last fees and fee rate then we would need a lower max size.
IIUC this is not the case -- the `MAX_PRESIGN_REQUEST_SIZE` is already a bound also for the worst case -- but maybe we should clarify this comment (eg, removing the _assuming_ part).

djordon: Oh right, sorry, this is the same problematic phrasing that we had in the test. Updated in https://github.com/stacks-sbtc/sbtc/pull/1994/commits/c765879a16e0120a7616467aad0d1d3a46816ed1.

signer/src/network/libp2p/swarm.rs
```diff
@@ -890,4 +920,203 @@ mod tests {
...
+            },
+            signature,
+            signer_public_key,
+        };
+
+        let data: Vec<u8> = proto::Signed::from(signed).encode_to_vec();
+        let data_len = data.len();
+
+        // The packager should make sure that the serialized size of
+        // BitcoinPreSignRequest.request_package is than or equal to
```
matteojug: nit
```suggestion
        // BitcoinPreSignRequest.request_package is less than or equal to
```

signer/src/bitcoin/packaging.rs
```diff
@@ -1183,4 +1348,337 @@ mod tests {
...
+        let bags = packager.finalize();
+
+        // Build a BitcoinPreSignRequest from all bags, mimicking the code in
+        // SbtcRequests::construct_transactions.
+        let request_package: Vec<TxRequestIds> = bags
+            .map(|bag| TxRequestIds::from(&crate::bitcoin::utxo::Requests::new(bag)))
+            .collect();
+
+        // In local runs the bag count has been between 18 and 20.
+        let num_bags = dbg!(request_package.len());
```
matteojug: nit: remove `dbg`?

matteojug (dismissed):
LGTM

signer/src/network/libp2p/swarm.rs
djordon: oops, fixed in https://github.com/stacks-sbtc/sbtc/pull/1994/commits/d3eb66526914e9edb2929d46f5663395f012dacc

signer/src/bitcoin/packaging.rs
djordon: Removed in https://github.com/stacks-sbtc/sbtc/pull/1994/commits/d3eb66526914e9edb2929d46f5663395f012dacc

matteojug (approved):

signer/src/bitcoin/packaging.rs
```diff
@@ -1183,4 +1302,332 @@ mod tests {
...
+        more_asserts::assert_le!(estimated_size, MAX_PRESIGN_REQUEST_SIZE);
+
+        // `BAG_OVERHEAD` assumes the worst case for each nested
+        // `TxRequestIds` in `request_package`: a 1-byte field tag plus a
+        // 2-byte length varint (3 bytes total). If a bag's encoded
+        // `TxRequestIds` is under 128 bytes, protobuf uses a 1-byte length
+        // varint instead, so the real overhead is only 2 bytes and our
+        // running total can be high by one byte for that bag. Once the
+        // nested message is large enough for a 2-byte length varint, the
+        // estimate matches. In this test we usually see `estimated_size ==
```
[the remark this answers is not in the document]
djordon: I updated this in https://github.com/stacks-sbtc/sbtc/pull/1994/commits/f9dd451e976b177e6fbbb47a905d5e54d6d18a1d

signer/src/bitcoin/packaging.rs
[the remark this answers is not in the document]
djordon: Fixed in https://github.com/stacks-sbtc/sbtc/pull/1994/commits/f9dd451e976b177e6fbbb47a905d5e54d6d18a1d and the test was updated too, I just forget which commit.

signer/src/bitcoin/packaging.rs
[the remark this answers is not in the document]
djordon: This was updated several times now.

cylewitruk (approved):
The implementation looks OK, I think my only real gripe would be that it feels like we're bleeding protobuf + messaging concerns into packaging; I may have considered trying to find a way to avoid that (e.g. via providing a "size probe" callback or similar), but nothing that I would consider blocking 👍

djordon:
> The implementation looks OK, I think my only real gripe would be that it feels like we're bleeding protobuf + messaging concerns into packaging; I may have considered trying to find a way to avoid that (e.g. via providing a "size probe" callback or similar), but nothing that I would consider blocking 👍

Yeah, I'm curious of what you have in mind, it's not clear to me how to do it differently without hardcoding max sizes or indirectly limiting the message size by adding some other limit.

PR #2005, opened 2026-04-10
fix(emily): Security hardening — CORS restriction and CDK stage classification fix

Valisthea:
## Summary

This PR addresses three security concerns in the Emily API:

### 1. CORS `allow_any_origin()` replaced with explicit allowlist

The Emily API (both Lambda and server binaries) previously used `allow_any_origin()` in the Warp CORS configuration. This allowed any website to make cross-origin GET/POST requests to Emily endpoints — including unauthenticated ones like `GET /limits`, `POST /deposit`, and `POST /start_throttle`.

**Fix:** Replaced with a configurable `ALLOWED_ORIGINS` environment variable (comma-separated), defaulting to production domains (`sbtc-emily.com`, `beta.sbtc-emily.com`, `bridge.sbtc.io`). The local server binary defaults to `localhost` origins.

### 2. `private-mainnet` removed from `isDevelopmentStack()`

The `isDevelopmentStack()` function in `emily-stack-utils.ts` included `PRIVATE_MAINNET_STAGE_NAME` in the development stage list. This caused the **testing API** — including the `/testing/wipe` endpoint that **wipes all DynamoDB tables** — to be deployed on `private-mainnet` infrastructure.

**Fix:** Removed `PRIVATE_MAINNET_STAGE_NAME` from the development check. Only actual development stages (`dev`, `local`, `unit-test`, `temp`) now qualify.

### 3. Security documentation added to README

Added a "Security Considerations" section to `emily/README.md` covering:
- CORS configuration and the `ALLOWED_ORIGINS` env var
- Development vs. production stage behavior
- Application-level authentication recommendations for operators

## Files Changed

- `emily/handler/src/bin/emily-lambda.rs` — CORS fix (production)
- `emily/handler/src/bin/emily-server.rs` — CORS fix (local dev)
- `emily/cdk/lib/emily-stack-utils.ts` — Remove private-mainnet from dev check
- `emily/README.md` — Security documentation

## Impact

- No breaking changes for existing deployments using default configuration
- Deployments with custom frontends should set `ALLOWED_ORIGINS` env var
- `private-mainnet` stage will no longer deploy testing endpoints


djordon-stacks:
Hi @Valisthea, thanks for taking an interest in sBTC! I just gave this a quick pass over and thought that I would share my initial reactions:

1. The "CORS `allow_any_origin()` replaced with explicit allowlist" change doesn't gain us anything. Emily is mainly read-only, with most write operations gated by an API key. The one write endpoint that is not gated is the `POST /deposit` endpoint. A malicious site that calls "create deposit" on behalf of a user doesn't gain anything that the attacker couldn't do themselves.
2. The "`private-mainnet` removed from `isDevelopmentStack()`" change is also unnecessary. The binaries deployed to private-mainnet do not include any test endpoints. Our test endpoints are feature-gated by the `testing` feature; see https://github.com/stacks-sbtc/sbtc/blob/e52920859b41809e931a33737791dcb8d1cd4fd2/emily/handler/src/api/handlers/mod.rs#L22-L23
and https://github.com/stacks-sbtc/sbtc/blob/e52920859b41809e931a33737791dcb8d1cd4fd2/emily/handler/Cargo.toml#L15-L17 and we do not compile Emily with those features when creating a release https://github.com/stacks-sbtc/sbtc/blob/e52920859b41809e931a33737791dcb8d1cd4fd2/.github/workflows/deploy-emily.yaml#L153-L158 But if we did compile our test endpoints in private-mainnet, it would be fine since private-mainnet is a testing environment.
4. I like parts of the "Application-Level Authentication" section in the "Security documentation" change; it seems good.

We don't want to proceed with the CORS changes. It adds operational friction without any real security benefit. I don't think that we should remove private-mainnet from `isDevelopmentStack`. I cannot see why we should do that. The updates to the README seem like a win, though.

Valisthea:
Thanks for the thorough review — this is exactly the kind of technical engagement I was hoping for.

**You're right on both points:**

The `#[cfg(feature = "testing")]` compile-time gate is solid — I missed that the testing module is excluded at the binary level, not just at the CDK level. The dual `routes()` implementation with `#[cfg(not(feature = "testing"))]` makes this airtight. My assumption that `isDevelopmentStack()` controlled endpoint exposure was incorrect.

On CORS — fair point that the practical attack surface is limited. Since `PUT`/`PATCH` aren't in the CORS allow list, the API-key-gated write endpoints can't be reached cross-origin anyway. I overestimated the risk.

**Two minor observations from the deeper dive, for what they're worth:**

- `POST /deposit` validates transaction structure (txid, scripts, format) but doesn't verify on-chain existence — meaning structurally valid but nonexistent transactions could pollute the deposit table. Signers presumably handle this gracefully downstream, but it's worth noting.
- `GET /limits` and other read endpoints expose operational parameters (peg cap, withdrawal caps, rolling limits) cross-origin to any site. Low severity, but not zero — worth considering if those values are ever treated as non-public.

Neither of these changes the conclusion. Just flagging them.

**Happy to update the PR to keep only the README security documentation** since that's the part we agree adds value. I'll strip the CORS and CDK changes.

And genuinely — thanks for taking the time to review this properly. This finding was originally submitted through Immunefi's bug bounty program and was closed as "Spam / UI-UX Issue" with a generic template response — costing the 100 USDC submission deposit in the process. That's what led me to push the fix directly. Your response here is what a proper security review looks like, and I appreciate it.

Cheers.

djordon-stacks:
Hi @Valisthea, thanks for flagging these issues. We're always happy to have an additional set of eyes on these things. Here are my thoughts on them.

>  `POST /deposit` validates transaction structure (txid, scripts, format) but doesn't verify on-chain existence — meaning structurally valid but nonexistent transactions could pollute the deposit table. Signers presumably handle this gracefully downstream, but it's worth noting.

The fact that Emily only validates the structure is a long-standing “issue” with the current flow. But we have a process that automatically updates Emily for deposit requests that don't materialize on-chain. Moreover, the sBTC signers check for the existence of the deposit transaction on-chain before acting. The setup is not ideal, but it appears to work reasonably well. It's a good call out, thank you.

>  `GET /limits` and other read endpoints expose operational parameters (peg cap, withdrawal caps, rolling limits) cross-origin to any site. Low severity, but not zero — worth considering if those values are ever treated as non-public.

They are supposed to be public; everyone should be able to see these values. As for the limits endpoint specifically, we want it to be the case that anyone can build any kind of bridge that they want. But they would still need to know Emily's limits to do so effectively, which is part of the reason why they are public. The other public endpoints expose information already available on either the Stacks or the Bitcoin blockchain.

> And genuinely — thanks for taking the time to review this properly. This finding was originally submitted through Immunefi's bug bounty program and was closed as "Spam / UI-UX Issue" with a generic template response — costing the 100 USDC submission deposit in the process. That's what led me to push the fix directly. Your response here is what a proper security review looks like, and I appreciate it.

I'm sorry to hear about your experience with Immunefi. I've heard that the new process has some rough edges. We need to hear what people have experienced to better calibrate the process, so I'm glad you shared this.

PR #2008, opened 2026-04-13
fix: make sure devenv bitcoin mempool works

djordon:
## Description

Fixes an issue that arose after https://github.com/stacks-sbtc/sbtc/pull/1976 was merged.

The local mempool.space containers needs access to a bitcoin core node with the `txindex` enabled. We added another bitcoin core node that had the `txindex` enabled, but that doesn't appear to work reliably, and this PR fixes it.

## Changes

* Make the main node a non-pruned node on devenv. Point electrs at the main bitcoin node.
* Point the sBTC signers use the pruned node.
* Allow the pruned node to fetch more data from the non-pruned node on devenv. This is to allow the pruned node to have the same mempool data as the main node. 

## Checklist

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -183,6 +183,7 @@ services:
     container_name: sbtc-signer-1
     depends_on:
       - postgres-1
+      - bitcoin-pruned
     environment:
       <<: *sbtc-signer-environment
       SIGNER_SIGNER__DB_ENDPOINT: postgresql://postgres:postgres@postgres-1:5432/signer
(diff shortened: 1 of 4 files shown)


cylewitruk (dismissed):
I haven't tried this locally, but nothing stands out to me

docker/docker-compose.yml
```diff
@@ -286,7 +289,7 @@ services:
     ports:
       - "127.0.0.1:18443:18443"
```
matteojug: nit: IIRC we expose the RPC port locally to run the signers as binary. If so, we should expose the pruned one instead of this, for completeness

matteojug (dismissed):

docker/docker-compose.yml
djordon: Moved it over in https://github.com/stacks-sbtc/sbtc/pull/2008/commits/8c7aac282c3ac8f81233ca5beeaa77819e85f7aa

matteojug-stacks (approved):

djordon-stacks (approved):

PR #2011, opened 2026-04-15
feat: add reclaim to demo-cli

matteojug-stacks:
## Description

Closes: #?

## Changes

Add reclaim command to demo-cli (used to test spox).

Also some random fixes:
 - Fix `make devenev-sbtc-*` commands
 - Fix `demo-cli` to use the non-follower bitcoin node
 - Small nit on a log

## Testing Information

Reclaim can be tested with:
```bash
# After first DKG
cargo run -p signer --bin demo-cli deposit --amount 3000000 --lock-time 1
cargo run -p signer --bin demo-cli reclaim --txid <tx from above>
```

## Checklist


docker/docker-compose.yml
```diff
@@ -568,7 +572,7 @@ services:
           --cookie "devnet:devnet" \
           --http-addr="0.0.0.0:3002" \
           --electrum-rpc-addr="0.0.0.0:60401" \
-          --daemon-rpc-addr="bitcoin-mempool:$${BITCOIN_RPC_PORT}" \
+          --daemon-rpc-addr="bitcoin:$${BITCOIN_RPC_PORT}" \
```
djordon-stacks, line 578: oops, things appear to work for me, but apparently they were broken?

djordon-stacks (commented):
Looks good, I just need to test it.

docker/docker-compose.yml
matteojug-stacks: It was a strange one for me: bitcoin was progressing fine, but the Stacks node was stuck at the first block for some reason. But it may have been just a fluke 🤷

signer/src/bin/demo_cli.rs
```diff
@@ -45,11 +48,15 @@ use stacks_common::address::{
 };
 
 // Demo defaults
+// Public key: 026c44dfe47941b0271c642c549d9a763afce7c6b0495c72f1a32c2f09898ea3df
 const DEMO_PRIVATE_KEY: &str = "2be0a71cb3a27d7f71a790ebe96cd106dd6d9c811b402178d1666ec3034dd64c";
 const DEMO_STACKS_ADDR: &str = "ST3497E9JFQ7KB9VEHAZRWYKF3296WQZEXBPXG193";
 const DEMO_BITCOIN_ADDR: &str = "bcrt1qezfmjvnaeu66wm52h7885mccjfh9lmh2v4kf8n";
 const DEMO_DEPLOYER: &str = "SN3R84XZYA63QS28932XQF3G1J8R9PC3W76P9CSQS";
 
+// We can't use the config since we need access to the main node, not the follower.
```
djordon-stacks, line 57: I'm curious, why do we need access to the main node?

signer/src/bin/demo_cli.rs
```diff
@@ -171,10 +179,26 @@ struct WithdrawArgs {
...
+struct ReclaimArgs {
+    /// The transaction ID of the deposit to reclaim.
+    #[clap(long)]
+    txid: String,
+    /// The vout of the deposit to reclaim.
+    #[clap(long, default_value = "0")]
+    vout: u64,
+    /// The key to use to sign for the reclaim, if a signature is required.
+    #[clap(long = "private-key")]
+    private_key: Option<String>,
```
djordon-stacks: With this PR, we will always set a p2pk reclaim script when creating a deposit request, where the public key used in the script corresponds to the hardcoded private key in the demo-cli. But here, we allow ourselves to reclaim funds in which the reclaim script is effectively empty. It's a little more flexible than the alternative, but you'll need to modify the code to create a deposit where you can reclaim without a private key.

I recommend using a hardcoded private key by default in the CLI, as it is simpler, though I'm not strongly attached to this. Curious, do you have another use for allowing the caller to specify the private key?

djordon-stacks (dismissed):
Looks good, and everything works*. Had a few questions but nothing blocking.

* Well, technically, I had some caching issues or something with the Docker build, but after fixing that, everything worked.

docker/docker-compose.yml
djordon-stacks: Yeah, that is strange. I didn't have any issues with the stacks node, even before the devenv fixes.

signer/src/bin/demo_cli.rs
matteojug-stacks: Because for faucet operations (funding accounts or creating deposits) we sign with a node wallet, and the wallet is created on the main node by the miner script (acting on the main node).

We could export the wallet there and `restorewallet` it on the follower node (or create the wallet in a deterministic way on both nodes), but I went with the lazy way instead.

signer/src/bin/demo_cli.rs
matteojug-stacks: I had this as optional as initially spox was creating empty reclaiming scripts in the demo and I wanted to test spending those, but yeah it makes sense to default to what the demo cli produces.

I changed this to a defaulted string (so now `cargo run -p signer --bin demo-cli reclaim --txid ...` is enough), and if empty (`cargo run -p signer --bin demo-cli reclaim --txid ... --private-key ''`) no signature is used.

signer/src/bin/demo_cli.rs
djordon-stacks: Okay, let's keep it as is. Still, I feel like I'm missing something, because I think this should work on the other node.

signer/src/stacks/api.rs
```diff
@@ -740,7 +740,7 @@ impl StacksClient {
             %contract_principal,
             %contract_name,
             %fn_name,
-            "Fetching contract data variable"
+            "Calling read-only function"
```
djordon-stacks, line 743: 🙏

djordon-stacks (dismissed):

signer/src/bin/demo_cli.rs
```diff
@@ -45,11 +48,17 @@ use stacks_common::address::{
...
+// Public key: 026c44dfe47941b0271c642c549d9a763afce7c6b0495c72f1a32c2f09898ea3df
 const DEMO_PRIVATE_KEY: &str = "2be0a71cb3a27d7f71a790ebe96cd106dd6d9c811b402178d1666ec3034dd64c";
 const DEMO_STACKS_ADDR: &str = "ST3497E9JFQ7KB9VEHAZRWYKF3296WQZEXBPXG193";
 const DEMO_BITCOIN_ADDR: &str = "bcrt1qezfmjvnaeu66wm52h7885mccjfh9lmh2v4kf8n";
 const DEMO_DEPLOYER: &str = "SN3R84XZYA63QS28932XQF3G1J8R9PC3W76P9CSQS";
 
+// We can't use the config since we need access to the main node, not the follower.
+// That's because for some operations (funding, deposit creation) we sign transactions
+// with the node wallet that's created by the devenv miner script on the main node.
+const BITCOIN_RCP_URL: &str = "http://127.0.0.1:18442/wallet/depositor";
```
cylewitruk-stacks: RPC?

matteojug-stacks: Ops, fixed in [2f88d9a](https://github.com/stacks-sbtc/sbtc/pull/2011/commits/2f88d9ab812f537a5ac67f3bc4e15e1a4a13cd6c)

cylewitruk-stacks (approved):
I haven't tried this personally, but I don't see anything else that stands out to me.

djordon-stacks (approved):

PR #2021, opened 2026-04-22
feat: send over the total vsize in the fees object

### LINKED ISSUE #2020: [Feature]: Send over the total vsize in the `Fees` object
# Feature - Send over the total vsize in the `Fees` object

## 1. Description

When sending over the last fee amounts spent for a transaction currently in the mempool, the signers send the total amount and the fee rate. They do not send the total vsize of the transaction package currently in the mempool, and they should do so, as it would simplify things.

## 2. Technical Details

The `Fees` information that the signers send, the total fees, and the fee rate, is basically equivalent to sending the total fees and the total vsize. However, the fee rate is a `double` computed by dividing total fees by the total vsize, both of which are 8-byte integers, and there is a mild concern that those values may differ when running the signer compiled for amd64 and aarch64. This concern is probably not warranted, since those architectures are mainstream, but we don't know for sure, and the change is easy.

### 2.1 Acceptance Criteria

- [ ] Signers send over the total vsize in the `Fees`  object

djordon-stacks:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/2020

## Changes

* Add a new `vsize` field to the `Fees` object.
* Add a `Fees::new` function for constructing a `Fees` object that rejects obviously invalid total fees.
* Update the protobufs in a backward-compatible way, and update the conversion functions to tolerate signers that haven't updated, but not tolerate signers that send bogus total fees data.

## Testing Information

Added some unit tests, which should cover what we want. We'll run proper network testing when we test this on devnet.

## Checklist

### DIFF
--- a/protobufs/stacks/signer/v1/common.proto
+++ b/protobufs/stacks/signer/v1/common.proto
@@ -19,10 +19,12 @@ message QualifiedRequestId {
   stacks.StacksBlockId block_hash = 3;
 }
 
-// Describes the fees for a transaction.
+// Describes the fees for a transaction package.
 message Fees {
-  // The total fee paid in sats for the transaction.
+  // The total fee paid in sats for the transaction package.
   uint64 total = 1;
   // The fee rate paid in sats per virtual byte.
   double rate = 2;
+  // The size of the transaction package in virtual bytes. 
+  uint64 vsize = 3;
 }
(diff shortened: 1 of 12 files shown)


djordon-stacks:
This is all wrong. We shouldn't even send the last fees over.

PR #2023, opened 2026-04-23
fix: reject bogus bitcoin fee rates

### LINKED ISSUE #2022: [Bug]: Reject bogus bitcoin fee rates
# Bug - Reject bogus bitcoin fee rates

## 1. Description

We currently allow the coordinator to propose any fee rate they like for the Bitcoin sweep transaction, so long as the requests fulfilled by the transaction have their max-fee respected. This is okay, but if the coordinator broadcasts a transaction with a fee rate that's too high, the signer will panic due to an integer overflow. Let's not do that and ensure the signer performs minimal validation of the fee rate.

## 2. Technical Details

The pre-validation step when receiving a pre-sign request is the right place to do this check. It is fast and static. We should also ensure that the estimated fee rate we use, the one returned by bitcoin-core, is reasonable.

### 2.1 Acceptance Criteria

- [ ] We reject bad fee rates from bitcoin-core.

djordon-stacks:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/2022

## Changes

* We no longer require signers to send the last fees in the pre-sign request.
* Fetch the last fees from the mempool during pre-sign validation.
* Do a simple check that the provided fee rate is within some broadly acceptable bounds.
* Do some simple checks that the vsize and total fee amount from the existing transaction package are within reasonable bounds.

## Testing Information

I tested this implicitly in devenv with spox just now. Well, I tested up to 6500089780229985540a70fa826e4f617d57cde9, which includes all substantial changes.

## Checklist


signer/src/transaction_coordinator.rs
```diff
@@ -1940,7 +1947,15 @@ where
     ) -> Result<utxo::SignerBtcState, Error> {
         let bitcoin_client = self.context.get_bitcoin_client();
         // Target next block confirmation
-        let fee_rate = bitcoin_client.estimate_fee_rate(1).await?;
+        let mut fee_rate = bitcoin_client.estimate_fee_rate(1).await?;
+
+        // Defensive check against bitcoin-core returning a bogus fee rate.
+        // This doesn't guard against f64::NAN, we'll need to manually
+        // intervene if it happens or hope that it resolves itself.
+        if !BITCOIN_FEE_RATE_RANGE.contains(&fee_rate) {
```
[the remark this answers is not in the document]
djordon-stacks: Yeah, I explicitly thought about NaNs and chose not to handle them here, as the comment shows. That might be the wrong decision; maybe we should set it to `1.0`? I'm curious about what others think here.

signer/src/transaction_coordinator.rs
```diff
@@ -725,6 +727,11 @@ where
                 }
             };
 
+            if !BITCOIN_FEE_RATE_RANGE.contains(&retry_fee_rate) {
```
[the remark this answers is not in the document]
djordon-stacks: Same as in https://github.com/stacks-sbtc/sbtc/pull/2023#discussion_r3132143773, curious about what people think.

signer/src/error.rs
```diff
@@ -764,7 +775,7 @@ pub enum Error {
 
     /// Indicates that the BitcoinPreSignRequest object contains a fee rate
     /// that is less than or equal to zero.
```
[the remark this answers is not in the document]
djordon-stacks: Updated in https://github.com/stacks-sbtc/sbtc/pull/2023/commits/952d63dc54765836bbbf1f27ab4a3524b8a1c6de

signer/src/lib.rs
```diff
@@ -69,6 +69,26 @@ pub const SIGNER_CHANNEL_CAPACITY: usize = 1024;
...
+pub const MAX_BITCOIN_FEE_RATE: f64 = 1000.0;
+
+/// The minimum fee rate, in sats per vbyte, for a bitcoin transaction.
+/// Transactions with a fee rate that exceeds this value will be rejected.
+///
+/// This value is less than the default fee rate in bitcoin core for the
+/// minrelaytxfee setting as of bitcoin-core v30, which is 0.1.
+///
+/// <https://bitcoincore.org/en/releases/30.0/>
+pub const MIN_BITCOIN_FEE_RATE: f64 = 0.001;
```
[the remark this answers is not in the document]
djordon-stacks: Fixed in https://github.com/stacks-sbtc/sbtc/pull/2023/commits/151d0162a84613c1cf829b0c8ebdd4ab8d775894

signer/src/bitcoin/utxo.rs
```diff
@@ -121,18 +123,48 @@ static DUMMY_SIGNATURE: LazyLock<Signature> = LazyLock::new(|| Signature {
...
-    pub const ZERO: Self = Self { total: 0, rate: 0.0 };
+    /// Create a new [`Fees`] instance.
+    ///
+    /// This function will return an error if the total fees are greater
+    /// than the maximum amount of bitcoin, or if the vsize is greater than
+    /// the maximum bitcoin block vsize. We strongly suspect that something
+    /// has gone wrong if these limits are exceeded.
+    pub fn new(total: u64, vsize: u64) -> Result<Self, Error> {
+        if total > Amount::MAX_MONEY.to_sat() {
+            return Err(Error::InvalidTotalFees(total));
```
matteojug-stacks: maybe nit: is this important enough require its own error? vs using `InvalidLastFee`

signer/src/bitcoin/utxo.rs
```diff
@@ -121,18 +123,48 @@ static DUMMY_SIGNATURE: LazyLock<Signature> = LazyLock::new(|| Signature {
...
+            return Err(Error::InvalidLastFee { total, vsize: vsize.get() });
+        }
+
+        Ok(Self { total, vsize })
+    }
+
+    /// Create a new [`Fees`] instance.
+    #[cfg(any(test, feature = "testing"))]
+    pub fn new_unchecked(total: u64, vsize: NonZeroU64) -> Self {
+        Self { total, vsize }
```
matteojug-stacks, line 162: maybe nit: we could accept a u64 and `NonZeroU64::new(vsize).unwrap()` in here to simplify the callers

signer/src/error.rs
```diff
@@ -808,6 +820,17 @@ pub enum Error {
     #[cfg(any(test, feature = "testing"))]
     #[error("Test utility error: {0}")]
     TestUtility(crate::testing::TestUtilityError),
+
+    /// Getting this error means a programming error or an error in
+    /// bitcoin-core.
```
matteojug-stacks: nit: we can get this error because of a malicious peer as well

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
matteojug-stacks: I would prefer to handle the NaN case, but I'm not sure if setting it to 1 or bailing entirely. Setting it to 1 (and logging about it) seems safer, worst case the node will reject our transaction if too low and we will still guarantee the requests max fee as usual if it's "too high".

signer/src/transaction_coordinator.rs
matteojug-stacks: Should the clamping only apply if we don't have a `fallback_fee`? It seems strange to clamp the explicitly set fallback fee

matteojug-stacks (commented):
Generally LGTM, minor comments

signer/src/bitcoin/utxo.rs
djordon-stacks: Oh, I created `InvalidLastFee` after creating the `InvalidTotalFees` variant and forgot to update this. It's probably better to remove it.

signer/src/bitcoin/utxo.rs
djordon-stacks: Yeah, that's a better API.

signer/src/error.rs
djordon-stacks: Oh, I forgot to update this comment. We cannot get this error at all, since it isn't compiled in production.

signer/src/transaction_coordinator.rs
djordon-stacks: Hmmm, it does. However, `BITCOIN_FEE_RATE_RANGE` is the sBTC consensus fee range, so it may not be appropriate to allow it. Hmmm, after some more thought, we should only clamp bitcoin-core's values because fallback fees aren't allowed on mainnet. Plus, it should be safe to send even if the fallback fee was NaN or out of range.

signer/src/bitcoin/utxo.rs
djordon-stacks: Updated to use the one error variant in https://github.com/stacks-sbtc/sbtc/pull/2023/commits/4660836c629c431ad07400e127f49f128488f524

signer/src/bitcoin/utxo.rs
djordon-stacks: Updated the function signature in https://github.com/stacks-sbtc/sbtc/pull/2023/commits/2c551a8ea28b24caa3ae26a218daa031392eb813

signer/src/error.rs
djordon-stacks: I updated the comment in https://github.com/stacks-sbtc/sbtc/pull/2023/commits/c3e5b780d6d2d8febf84817896c0c949218bd0fc

signer/src/transaction_coordinator.rs
djordon-stacks: I fixed this up in https://github.com/stacks-sbtc/sbtc/pull/2023/commits/68c8c6cf899d815d31a42aec148120055ea8eece

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
djordon-stacks: I fixed this up in https://github.com/stacks-sbtc/sbtc/pull/2023/commits/68c8c6cf899d815d31a42aec148120055ea8eece

signer/src/transaction_coordinator.rs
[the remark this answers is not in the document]
djordon-stacks: Updated in https://github.com/stacks-sbtc/sbtc/pull/2023/commits/68c8c6cf899d815d31a42aec148120055ea8eece

matteojug-stacks (dismissed):

signer/src/lib.rs
```diff
@@ -128,6 +151,9 @@ pub const WITHDRAWAL_EXPIRY_BUFFER: u64 = 6;
 /// <https://github.com/bitcoin/bitcoin/blob/v25.0/src/policy/policy.h#L60-L61>
 pub const MAX_MEMPOOL_PACKAGE_SIZE: u64 = 101000;
 
+/// This is the maximum size of a bitcoin block in virtual bytes.
+pub const MAX_BITCOIN_BLOCK_VSIZE: u64 = 4_000_000;
```
cylewitruk-stacks: QQ: Is this statement correct? I want to recall that the 4M limit was weight units and 1 vByte was 4 weight units, with a discount on witness data?

cylewitruk-stacks (commented):
Still looking through but had one Q I wanted to double-check in the meantime

signer/src/lib.rs
djordon-stacks: Yeah, good catch. I think the maximum bitcoin block size is 1_000_000 in vbytes and 4_000_000 in weight units.

djordon-stacks: Fixed in https://github.com/stacks-sbtc/sbtc/pull/2023/commits/c2c394d260895e965cebb19086e0b2ae45f2dc3f

signer/src/bitcoin/utxo.rs
```diff
@@ -2619,14 +2653,10 @@ mod tests {
 
             more_asserts::assert_gt!(input_amounts, output_amounts);
             let fee_total = input_amounts - output_amounts;
-            let fee_rate = fee_total as f64 / utx.tx.vsize() as f64;
-            (fee_total, fee_rate)
+            (fee_total, utx.tx.vsize() as u64)
```
cylewitruk-stacks, line 2656: nit: looks like `vsize()` returns a `u64` already?

cylewitruk-stacks (approved):
Lgtm ✅

matteojug-stacks (approved):

signer/src/bitcoin/utxo.rs
djordon-stacks: Oh, I have it returning a `usize` but it could return a `u64` because the `Transaction::vsize` implementation is:
```rust
self.weight().to_vbytes_ceil() as usize
```
where `to_vbytes_ceil` returns a `u64`: [`vsize`](https://github.com/rust-bitcoin/rust-bitcoin/blob/bitcoin-0.32.5/bitcoin/src/blockdata/transaction.rs#L891-L894), [`to_vbytes_ceil`](https://github.com/rust-bitcoin/rust-bitcoin/blob/bitcoin-0.32.5/units/src/weight.rs#L102-L104).

cylewitruk-stacks: Ah was probably the crappy github.dev editor's intellisense that mislead me 🤣

djordon-stacks:
I just tested this branch again on devenv, from https://github.com/stacks-sbtc/sbtc/pull/2023/commits/cbdbce19e9784f7017288782841a501375a5ac51, and everything worked fine! So merging.

PR #2051, opened 2026-05-08
feat: add sanction file config to blocklist client

matteojug-stacks:
## Description

Partial work for https://github.com/stacks-sbtc/sbtc/issues/2050

Breaking local change: renamed `BLOCKLIST_CLIENT_ASSESSMENT__ASSESSMENT_METHOD` to `BLOCKLIST_CLIENT_RISK_ANALYSIS__ASSESSMENT_METHOD`. The field is defaulted to `sanctions`, so the only breaking case is if someone set it to `risk_analysis` 

## Changes

Add new config options for the sanctions file: the URL to fetch, optional auth header, polling interval and local path to preload the set.

I've kept the current risk API, in case someone implemented their own, but I consolidated the assessment under the `RiskAnalysisConfig`.

Once everything is ready (ie, #2052, #2053 or similar, and https://github.com/stacks-sbtc/sbtc/issues/2054 is addressed) the canonical config will become:
```
BLOCKLIST_CLIENT_SANCTIONS__URL='<emily address/blocklist'
BLOCKLIST_CLIENT_SANCTIONS__HEADER__KEY='x-api-key'
BLOCKLIST_CLIENT_SANCTIONS__HEADER__VALUE='<emily API key>'
BLOCKLIST_CLIENT_SANCTIONS__LOCAL_PATH='<local sanction list>'
```

## Testing Information

Add config tests.

## Checklist

- [X] I have had an AI agent review my code


blocklist-client/src/config/default.toml
```diff
@@ -1,6 +1,3 @@
 [server]
 host = "127.0.0.1"
 port = 3030
-
-[assessment]
-assessment_method = "sanctions"
```
djordon-stacks, line 6: Why not replace this with a `risk_analysis` section? In general, I believe the "default" config file should be well-functioning, even though it seems we haven't been following that principle here 😞.

djordon-stacks: Shouldn't also add in the new configuration values here, probably commented out, similar to what we do in the signer?

blocklist-client/src/config.rs
```diff
@@ -56,13 +50,47 @@ pub enum AssessmentMethod {
...
+
+/// Sanction file config
+#[derive(Deserialize, Clone, Debug)]
+#[allow(dead_code)] // TODO: remove once it's ready
+pub struct SanctionFileConfig {
+    /// The URL to fetch the file from
+    #[serde(deserialize_with = "url_deserializer_single")]
+    pub url: Url,
+    /// Optional header for authorization
+    pub header: Option<SanctionFileConfigHeader>,
```
djordon-stacks, line 69: Is there a world where we want to send more than one header? I don't see it, but I'm wondering whether we want to make this a `Vec<_>` instead of an `Option<_>`?

djordon-stacks (commented):
Looks good. I have a few questions and suggestions about the edits to the config file.

blocklist-client/src/api/handlers.rs
```diff
@@ -34,13 +34,21 @@ pub async fn check_address_handler(
...
+                AssessmentMethod::RiskAnalysis => {
+                    risk_client::check_address(&client, risk_analysis, &address).await
+                }
             }
+        } else if config.sanctions.is_some() {
+            // TODO: implement this
+            Err(Error::InternalServer)
+        } else {
+            // This shouldn't happen since we validate the config
+            Err(Error::InternalServer)
```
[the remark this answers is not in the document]
matteojug-stacks: I'll merge this only when the handler is ready

blocklist-client/src/config.rs
```diff
@@ -123,6 +151,158 @@ impl Settings {
...
+    use assert_matches::assert_matches;
+
+    use super::*;
+
+    fn clear_env() {
+        for var in std::env::vars() {
+            if var.0.starts_with("BLOCKLIST_CLIENT_") {
+                unsafe {
+                    std::env::remove_var(var.0);
+                }
```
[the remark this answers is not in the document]
matteojug-stacks: We do isolated tests

blocklist-client/src/config.rs
```diff
@@ -123,6 +151,158 @@ impl Settings {
...
+            }
+        }
+    }
+    fn set_var<K: AsRef<std::ffi::OsStr>, V: AsRef<std::ffi::OsStr>>(key: K, value: V) {
+        unsafe {
+            std::env::set_var(key, value);
+        }
```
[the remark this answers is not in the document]
matteojug-stacks: We do isolated tests

blocklist-client/src/config/default.toml
matteojug-stacks: I'll add the comments and a basic config, I guess we didn't do it before as we don't use a file in the suggested deployment, we just pass the envs to the container.

blocklist-client/src/config.rs
matteojug-stacks: Initially I had multiple headers, but there wasn't a nice way to handle them via envs:
 - If we use something like `map<string,string>` we have the header name in the env name and in general a header name is not a valid suffix for an env variable name
 - If we use a "simple" vec we cannot set it via envs (or I couldn't find it)
 - we could use a map of dummy integer keys (to get something like `...HEADERS__0__KEY`), but it was a bit ugly on the config file side
 - if we use a serialized json vec (eg, `...HEADERS='{"key-1":"val-1",...}'`) then we need to parse it into a list of headers, which could definitely be done but could have some rough edges

So given the above, and the fact that I couldn't see a use case for multiple headers, I went with the simpler option.

djordon-stacks: Yeah, that makes sense. I want to use the simpler option, so I'm happy. But yeah, just curious if we're sure if we only need the one. We can revisit if we need more, but I don't expect it.

blocklist-client/src/config/default.toml
matteojug-stacks: Added comments in [1f20b11](https://github.com/stacks-sbtc/sbtc/pull/2051/commits/1f20b1104e90ec4d39f20b08661e89089b1b38e7), but I'm not sure about uncommenting `risk_analysis` or `sanctions` stanzas, as they would contain dummy values that would not really be well-functioning.

djordon-stacks (commented):
Looks good, but is this safe to merge right now? If we cut a release tomorrow after this is merged, do operators need to update their environment variables or config? I think the answer is "yes", which makes me think that we should hold off on merging it or update the docs so that the required changes are clear.

djordon-stacks:
> which makes me think that we should hold off on merging it or update the docs so that the required changes are clear.

We don't have many docs here, and we've already updated the mainnet `docker-compose.yml` reference in this PR. I think we should update the PR description for sure and tag this as breaking-local. Is there anything else that we need to do?

blocklist-client/README.md
```diff
@@ -19,9 +19,10 @@ You can run the blocklist client by providing the required environment variables
 
 - BLOCKLIST_CLIENT_RISK_ANALYSIS__API_URL=`<provider-url>`
 - BLOCKLIST_CLIENT_RISK_ANALYSIS__API_KEY=`<your_api_key>`
+- BLOCKLIST_CLIENT_RISK_ANALYSIS__ASSESSMENT_METHOD=`<sanctions|risk_analysis>`
```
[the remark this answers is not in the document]
matteojug-stacks: Updated in [32c428f](https://github.com/stacks-sbtc/sbtc/pull/2051/commits/32c428f4ca8f18d6094360575efa1add37c5f7a6)

docker/mainnet/docker-compose.yml
```diff
@@ -80,7 +80,7 @@ services:
     environment:
       BLOCKLIST_CLIENT_RISK_ANALYSIS__API_KEY: *BLOCKLIST_CLIENT_RISK_ANALYSIS__API_KEY
       BLOCKLIST_CLIENT_RISK_ANALYSIS__API_URL: https://public.chainalysis.com
-      BLOCKLIST_CLIENT_ASSESSMENT__ASSESSMENT_METHOD: sanctions
+      BLOCKLIST_CLIENT_RISK_ANALYSIS__ASSESSMENT_METHOD: sanctions
       BLOCKLIST_CLIENT_SERVER__HOST: "0.0.0.0"
```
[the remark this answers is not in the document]
matteojug-stacks: Made it default again in [32c428f](https://github.com/stacks-sbtc/sbtc/pull/2051/commits/32c428f4ca8f18d6094360575efa1add37c5f7a6)

matteojug-stacks:
I've added some more docs and restored the default `sanctions` for the field, so for operators using our configuration (i.e., not explicitly setting the assessment method to risk analysis) this shouldn't be a breaking change.

> Looks good, but is this safe to merge right now? If we cut a release tomorrow after this is merged, do operators need to update their environment variables or config? I think the answer is "yes", which makes me think that we should hold off on merging it or update the docs so that the required changes are clear.

Now the answer should be "no" and this be safe to merge (with the caveat above), but I would prefer to merge this together with the other PR (#2052) to avoid confusion.

> We don't have many docs here, and we've already updated the mainnet docker-compose.yml reference in this PR. I think we should update the PR description for sure and tag this as breaking-local. Is there anything else that we need to do?

Tagged and added more docs, please let me know if you were thinking of updating some other ref as well.

blocklist-client/src/config.rs
```diff
@@ -56,13 +50,52 @@ pub enum AssessmentMethod {
     RiskAnalysis,
 }
 
+fn default_sanction_polling_interval() -> Duration {
+    Duration::from_secs(3600)
```
cylewitruk-stacks, line 54: nit: I think `Duration::from_secs()` is `const fn`, so could use a constant? `default_assessment_method()` below could be too? But maybe that doesn't work with `serde`'s `default`? Feel free to close if I'm speaking nonsense.

djordon-stacks:
> We don't have many docs here, and we've already updated the mainnet docker-compose.yml reference in this PR

To clarify, we don't have many docs on main, so my earlier post about updating the docs was off base, since there isn't much to update 😄.

> I've added some more docs and restored the default `sanctions` for the field, so for operators using our configuration (i.e., not explicitly setting the assessment method to risk analysis) this shouldn't be a breaking change.

Oh, okay, cool, I will take a look.

> Tagged and added more docs, please let me know if you were thinking of updating some other ref as well.

Will do!

cylewitruk-stacks (approved):

blocklist-client/src/config.rs
cylewitruk-stacks: nvm

djordon-stacks (approved):
Looks good

PR #2053, opened 2026-05-11
feat: emily sanctions passthru

matteojug-stacks:
## Description

More work for https://github.com/stacks-sbtc/sbtc/issues/2036

## Changes

Modify Emily CDK for the file-based blocklist-client:
 - Add an S3 bucket for the file
 - Add an endpoint to the API Gateway to serve the above file (API-key gated)

## Testing Information

Manually deployed (without the `autoDeleteObjects`) to devnet and verified with
```bash
curl --header "x-api-key: ..." --silent 'https://<devnet>/sanctions' | head
```

## Checklist

- [X] I have had an AI agent review my code

### DIFF
--- a/emily/cdk/lib/emily-stack.ts
+++ b/emily/cdk/lib/emily-stack.ts
@@ -135,7 +135,7 @@ export class EmilyStack extends cdk.Stack {
         removalPolicy: cdk.RemovalPolicy,
         props: EmilyStackProps,
     ): s3.Bucket {
-        const bucketId: string = 'SanctionsBucket';
+        const bucketId: string = 'SanctionsLists';
         return new s3.Bucket(this, bucketId, {
             bucketName: EmilyStackUtils.getResourceName(bucketId, props).toLowerCase(),
             blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
--- a/emily/cdk/lib/emily-stack.ts
+++ b/emily/cdk/lib/emily-stack.ts
@@ -519,11 +519,11 @@ export class EmilyStack extends cdk.Stack {
 
     /**
      * Creates or updates the API Gateway to connect with the Lambda function.
-     * @param {lambda.Function} operationLambda The Lambda function to connect to the API.
+     * @param {lambda.Alias} operationLambda The Lambda function to connect to the API.
      * @param {s3.IBucket} sanctionsBucket The sanctions S3 bucket.
      * @param {EmilyStackProps} props The stack properties.
-     * @returns {apig.SpecRestApi} The created or updated API Gateway.
-     * @post An API Gateway with execute permissions linked to the Lambda function is returned.
+     * @returns {apig.SpecRestApi[]} The created or updated API Gateways.
+     * @post API Gateways with execute permissions linked to the Lambda function are returned.
      */
     createOrUpdateApi(
         operationLambda: lambda.Alias,
@@ -561,7 +561,7 @@ export class EmilyStack extends cdk.Stack {
 
     /**
      * Creates or updates a specific API Gateway to connect with the Lambda function.
-     * @param {lambda.Function} operationLambda The Lambda function to connect to the API.
+     * @param {lambda.Alias} operationLambda The Lambda function to connect to the API.
      * @param {number} numApiKeys The number of API keys to create for the API.
      * @param {string} apiPurpose The purpose of the API.
      * @param {s3.IBucket} sanctionsBucket The sanctions S3 bucket.
--- a/emily/cdk/lib/emily-stack.ts
+++ b/emily/cdk/lib/emily-stack.ts
@@ -142,8 +142,7 @@ export class EmilyStack extends cdk.Stack {
             encryption: s3.BucketEncryption.S3_MANAGED,
             enforceSSL: true,
             removalPolicy,
-            // TODO: uncomment this before merging
-            // autoDeleteObjects: removalPolicy === cdk.RemovalPolicy.DESTROY,
+            autoDeleteObjects: removalPolicy === cdk.RemovalPolicy.DESTROY,
         });
     }
 
--- a/emily/cdk/lib/emily-stack.ts
+++ b/emily/cdk/lib/emily-stack.ts
@@ -634,10 +634,8 @@ export class EmilyStack extends cdk.Stack {
             sourceArn: api.arnForExecuteApi(),
         });
 
-        // The sanctions list is only served on the public API.
-        if (apiPurpose === "public") {
-            this.addSanctionsRoute(api, sanctionsBucket, apiPurposeResourceIdSuffix);
-        }
+        // Add the sanctions list to the API.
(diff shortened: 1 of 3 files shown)


emily/cdk/lib/emily-stack.ts
```diff
@@ -110,13 +111,41 @@ export class EmilyStack extends cdk.Stack {
...
+        const bucketId: string = 'SanctionsBucket';
+        return new s3.Bucket(this, bucketId, {
+            bucketName: EmilyStackUtils.getResourceName(bucketId, props).toLowerCase(),
+            blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
+            encryption: s3.BucketEncryption.S3_MANAGED,
+            enforceSSL: true,
+            removalPolicy,
+            autoDeleteObjects: removalPolicy === cdk.RemovalPolicy.DESTROY,
+        });
+    }
```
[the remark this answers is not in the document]
matteojug-stacks: Yeah done in [2f50cd7](https://github.com/stacks-sbtc/sbtc/pull/2053/commits/2f50cd7750f978ec74fe47a719cf184cf40ba1a8)

djordon-stacks (commented):
Nice. Well, I'm still curious about what other people think here. API Gateway feels a little "off" because we don't actually care about gating access to the file; we care about limiting our costs from potential bad actors. And we still pay when someone tries to hit our endpoints, but it's less than it would be if we had the file out in the open. Anyway, I have a feeling that a different setup would be a better fit for us.

fabrizio-stacks:
> Nice. Well, I'm still curious about what other people think here. API Gateway feels a little "off" because we don't actually care about gating access to the file; we care about limiting our costs from potential bad actors. And we still pay when someone tries to hit our endpoints, but it's less than it would be if we had the file out in the open. Anyway, I have a feeling that a different setup would be a better fit for us.

In my opinion the solution is ok, otherwise we can expose the s3 bucket via a CDN like cloudfront and give the address

djordon-stacks:
> In my opinion the solution is ok, otherwise we can expose the s3 bucket via a CDN like cloudfront and give the address

Yeah, that was the alternative approach that I was thinking about. We can go with this for now and change it later if needed.

emily/cdk/lib/emily-stack.ts
```diff
@@ -600,6 +634,11 @@ export class EmilyStack extends cdk.Stack {
             sourceArn: api.arnForExecuteApi(),
         });
 
+        // The sanctions list is only served on the public API.
+        if (apiPurpose === "public") {
+            this.addSanctionsRoute(api, sanctionsBucket, apiPurposeResourceIdSuffix);
+        }
```
djordon-stacks: Curious, why is it only served on the public API? Hmmm, I suppose it doesn't matter; we only need the one.

djordon-stacks (commented):
Looks good, but we should probably add a test here to verify that this resource is created. No idea how easy that is.

emily/cdk/lib/emily-stack.ts
matteojug-stacks: Yeah nothing wrong in exposing it on all the APIs, I limited it to the public one since that's probably the only place in which we will use this, but maybe having it on all of them is cleaner.

matteojug-stacks: Removed the if guard (and added tests) in [45b8468](https://github.com/stacks-sbtc/sbtc/pull/2053/commits/45b84686624fc0fba830a7c693dda25d51d5123a)

emily/cdk/lib/emily-stack.ts
```diff
@@ -600,6 +634,11 @@ export class EmilyStack extends cdk.Stack {
...
+
```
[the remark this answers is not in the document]
matteojug-stacks: Done in [45b8468](https://github.com/stacks-sbtc/sbtc/pull/2053/commits/45b84686624fc0fba830a7c693dda25d51d5123a)

emily/cdk/lib/emily-stack.ts
```diff
@@ -540,6 +573,7 @@ export class EmilyStack extends cdk.Stack {
         operationLambda: lambda.Alias,
         numApiKeys: number,
         apiPurpose: "public" | "private" | "testing",
+        sanctionsBucket: s3.IBucket,
         props: EmilyStackProps,
     ): apig.SpecRestApi {
```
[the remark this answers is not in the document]
matteojug-stacks: Done in [335e1e7](https://github.com/stacks-sbtc/sbtc/pull/2053/commits/335e1e73738370b56184f8e0e21a70319f15af5e)

emily/cdk/lib/emily-stack.ts
```diff
@@ -491,12 +520,14 @@ export class EmilyStack extends cdk.Stack {
...
+     * @param {s3.IBucket} sanctionsBucket The sanctions S3 bucket.
      * @param {EmilyStackProps} props The stack properties.
      * @returns {apig.SpecRestApi} The created or updated API Gateway.
      * @post An API Gateway with execute permissions linked to the Lambda function is returned.
      */
     createOrUpdateApi(
         operationLambda: lambda.Alias,
+        sanctionsBucket: s3.IBucket,
         props: EmilyStackProps
     ): apig.SpecRestApi[] {
```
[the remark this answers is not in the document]
matteojug-stacks: Done in [335e1e7](https://github.com/stacks-sbtc/sbtc/pull/2053/commits/335e1e73738370b56184f8e0e21a70319f15af5e)

emily/cdk/lib/emily-stack.ts
```diff
@@ -110,13 +111,41 @@ export class EmilyStack extends cdk.Stack {
...
+     * `GET /sanctions` on the public API.
+     * @param {cdk.RemovalPolicy} removalPolicy The removal policy for the bucket.
+     * @param {EmilyStackProps} props The stack properties.
+     * @returns {s3.Bucket} The created or updated S3 bucket.
+     */
+    createSanctionsBucket(
+        removalPolicy: cdk.RemovalPolicy,
+        props: EmilyStackProps,
+    ): s3.Bucket {
+        const bucketId: string = 'SanctionsBucket';
```
djordon-stacks: nit: I would rename this to `sanctions-lists` or something similar. I wouldn't have the word "bucket" in the name. My first thought was `ofac-sanctions`, since his bucket only has OFAC sanctions at the moment, but if we wanted to expand it to include other sanction lists, then the name would be incorrect. Also, before I forget, we should probably move `sanctions.txt` to `ofac/sanctions.txt` but we can make that change later.

djordon-stacks (dismissed):
Looks good, small nit for the bucket name.

emily/cdk/lib/emily-stack.ts
matteojug-stacks: Updated in [3d34bea](https://github.com/stacks-sbtc/sbtc/pull/2053/commits/3d34bea9c058dfc38be10763efc99a6127102222)

djordon-stacks (approved):

adriano-stacks (approved):

PR #2111, opened 2026-09-08
feat: add a command that computes the current sbtc signers' address

djordon-stacks:
## Description

Closes #?

I want this CLI command so we can more easily automate getting the sBTC signer's current address without duplicating any code.

## Changes

* Add a command for computing the current sBTC signers' address.

## Testing Information

I tried this locally, and it worked.

## Checklist

- [x] I have had an AI agent review my code

### DIFF
--- a/sbtc/README.md
+++ b/sbtc/README.md
@@ -4,7 +4,9 @@ A library for creating Bitcoin deposit transactions that can be handled by the s
 
 ## CLI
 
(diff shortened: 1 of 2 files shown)


djordon-stacks:
Actually, I don't need this code anymore, my original use for this, in the docs repo, is better served by duplicating the code.

PR #2113, opened 2026-09-08
feat: static webpage deposit addresses

### LINKED ISSUE #2110: [Feature]: Static website for constructing deposit addresses
# Feature - Static website for constructing deposit addresses

## 1. Description

It would be helpful to provide a static webpage where users can independently construct sBTC deposit addresses. Once added, this would give users another way to independently verify the deposit address provided by the bridge, alongside the two methods outlined in https://docs.stacks.co/more-guides/sbtc/bridging-bitcoin/btc-to-sbtc and the CLI at https://github.com/stacks-sbtc/sbtc/issues/2108. We will continue to encourage users to go through the bridge so Emily is properly notified of the deposit request; this site is for independent verification using code from a separate source.

## 2. Technical Details

This website would live in this repository and be deployed as a static GitHub Pages site.

### 2.1 Acceptance Criteria

- [ ] We have a website that users can use to independently verify their deposit address.

## 3. Related Issues and Pull Requests (optional)
 
https://github.com/stacks-sbtc/sbtc/issues/2108

djordon-stacks:
## Description

Closes https://github.com/stacks-sbtc/sbtc/issues/2110

The code in this branch is being demoed at https://djordon-stacks.github.io/sbtc3.

## Changes

* Add a static site for constructing deposit addresses.

## Testing Information

I tested this manually, and it constructed the same address as our bridge does.

## Checklist

- [x] I have had an AI agent review my code


web/deposit-address-verifier/src/main.ts
```diff
@@ -0,0 +1,338 @@
...
+  return `
+    <header class="site-header">
+      <a class="brand" href="#/" aria-label="sBTC address constructor home">
+        <img src="./icons/sbtc.svg" alt="" aria-hidden="true" />
+        <span>sBTC address constructor</span>
+      </a>
+      <nav aria-label="Primary navigation">
+        <a href="#/" data-nav="verify">Construct</a>
+        <a href="#/about" data-nav="about">About</a>
+        <a href="${REPOSITORY_URL}" target="_blank" rel="noreferrer">GitHub ↗</a>
```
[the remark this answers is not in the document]
djordon-stacks: Fixed in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/b12d4dd916e1a5a49bbce481f64b13c69cfc11e6

web/deposit-address-verifier/README.md
```diff
@@ -0,0 +1,27 @@
...
+```bash
+pnpm --filter @stacks-sbtc/deposit-address-verifier test
+pnpm --filter @stacks-sbtc/deposit-address-verifier build
+```
+
+The production files are written to `web/deposit-address-verifier/dist`.
+
+## Deployment
+
+The `Deploy sBTC address constructor` workflow publishes the static build to GitHub Pages after relevant changes reach `main`. The repository's Pages source must be set to **GitHub Actions** under **Settings → Pages** before the first deployment.
```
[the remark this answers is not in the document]
djordon-stacks: Updated in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/b12d4dd916e1a5a49bbce481f64b13c69cfc11e6

web/deposit-address-verifier/src/main.ts
```diff
@@ -0,0 +1,338 @@
...
+    }
+  })
+
+  document.querySelectorAll<HTMLButtonElement>('[data-copy]').forEach(button => {
+    button.addEventListener('click', async () => {
+      const target = document.querySelector(`#${button.dataset.copy}`)?.textContent ?? ''
+      await navigator.clipboard.writeText(target)
+      button.textContent = 'Copied'
+      window.setTimeout(() => (button.textContent = 'Copy'), 1400)
+    })
```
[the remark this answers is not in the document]
djordon-stacks: Fixed in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/b12d4dd916e1a5a49bbce481f64b13c69cfc11e6

web/deposit-address-verifier/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+  "compilerOptions": {
+    "target": "ES2022",
+    "useDefineForClassFields": true,
+    "module": "ESNext",
+    "lib": ["ES2022", "DOM", "DOM.Iterable"],
+    "skipLibCheck": true,
+    "moduleResolution": "Bundler",
+    "paths": {
+      "sbtc": ["./node_modules/sbtc/dist/index.d.ts"]
+    },
```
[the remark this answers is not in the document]
djordon-stacks: From my other machine reviewer:
> Partly correct:
>
> - **Missing `baseUrl`: incorrect.** TypeScript 4.1+ supports `paths` without `baseUrl`; TypeScript explicitly discourages adding it solely for path mappings. The current `tsc --noEmit` also passes. [[TypeScript docs](https://www.typescriptlang.org/docs/handbook/modules/reference#baseurl)](https://www.typescriptlang.org/docs/handbook/modules/reference#baseurl)
> - **Panel height not cleared: correct.** Expanding advanced options leaves the result panel's previously assigned fixed height in place.
> - **`lockTime` required but ignored with `reclaimScript`: correct design criticism**, though not currently a runtime bug. A discriminated input type would model the two modes better.
>
> So two worthwhile improvements, but the review's headline TypeScript claim is wrong. No changes made.

I think I'll fix the lock-time thing when the user selects to use a raw reclaim-script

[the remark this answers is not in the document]
djordon-stacks: Oh, the lock-time "issue" is an internal API thing, not a UI issue. I still might fix it.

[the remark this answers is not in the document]
djordon-stacks: I fixed the internal API issue in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/45e06d8a74a6782edd3f0bce66e147972c342a6b.

hugo-stacks (commented):
Can you share some more context about the goal of this page / project?

I'm afraid that today and in the future, the HTML template strings in TS and the unformatted CSS will be too hard to maintain.

Should we rethink the approach?

web/deposit-address-verifier/src/main.ts
```diff
@@ -0,0 +1,350 @@
...
+        </div>
+      </aside>
+    </section>
+  `)
+
+  document.querySelector('[data-nav="verify"]')?.setAttribute('aria-current', 'page')
+  bindVerifier()
+}
+
+function renderAbout(): void {
```
hugo-stacks: I would make `about.html` an actual page.
Avoid the old `/#/about` SPA trick. Would just be a simpler set up and give better SEO for free

web/deposit-address-verifier/src/main.ts
```diff
@@ -0,0 +1,350 @@
...
+      </nav>
+    </header>
+    <main>${content}</main>
+    <footer>
+      <span>Open-source, client-side verification.</span>
+      <span><a href="${REPOSITORY_URL}" target="_blank" rel="noopener noreferrer">Source code</a> · <a href="${BRIDGE_URL}" target="_blank" rel="noopener noreferrer">Official sBTC Bridge</a></span>
+    </footer>`
+}
+
+function renderVerifier(): void {
```
hugo-stacks: I would make most of it just static HTML in the index.html, and only interact with the dynamic bits from TS

web/deposit-address-verifier/src/style.css
```diff
@@ -0,0 +1,156 @@
+@import url('https://fonts.googleapis.com/css2?family=DM+Mono:wght@400;500&family=Manrope:wght@400;500;600;700&display=swap');
```
hugo-stacks: Users may want to avoid calling google server on this server.
Embed the fonts or use system fonts https://github.com/system-fonts/modern-font-stacks

web/deposit-address-verifier/src/style.css
```diff
@@ -0,0 +1,156 @@
...
+
+:root {
```
hugo-stacks, line 1: use oxfmt on the CSS, and add a check in CI

web/deposit-address-verifier/src/main.ts
```diff
@@ -0,0 +1,350 @@
+import {
```
hugo-stacks, line 1: Use a formatter (preferably oxfmt) on TS files, and add a check in CI

web/deposit-address-verifier/src/main.ts
```diff
@@ -0,0 +1,350 @@
...
+  })
+
+  const syncCollapsedPanelHeight = () => {
+    if (window.matchMedia('(max-width: 860px)').matches) {
+      resultPanel.style.removeProperty('height')
+    } else if (advancedFields.hidden) {
+      resultPanel.style.height = `${form.offsetHeight}px`
+    }
+  }
+  const formObserver = new ResizeObserver(syncCollapsedPanelHeight)
```
hugo-stacks: I think it could be fully implemented in CSS (discarding the copilot review comment above and simplifying the code)

hugo-stacks (commented):
Left suggestions to improve code hygiene and readability

I would also add a strict CSP meta tag (once the google font is remove)

hugo-stacks:
UI nit: I would make sure that:
- the footer is at the bottom of the page instead of floating in the middle on bigger screen
- both pages have the same container width

djordon-stacks:
> Left suggestions to improve code hygiene and readability
> 
> I would also add a strict CSP meta tag (once the google font is remove)

Great suggestions, thank you 😄.

web/deposit-address-verifier/src/style.css
djordon-stacks: Done in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/6880d8c08bf26220405066ae9edab689bfcd69ab

web/deposit-address-verifier/src/main.ts
djordon-stacks: Done in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/52d5f4bcffeba5d901a7e431980b6054f5e9a64f, well the page was made in that commit and wired in another commit.

web/deposit-address-verifier/src/main.ts
djordon-stacks: Added the formater wiring in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/adc10327dc7e98adb59da87f3fd725b101a0737a. The actual format change was done in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/c617b1e954748901cb07cad661b219ba4e484f09

web/deposit-address-verifier/src/main.ts
djordon-stacks: Done in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/c617b1e954748901cb07cad661b219ba4e484f09, I think 😅.

web/deposit-address-verifier/src/style.css
djordon-stacks: Done in https://github.com/stacks-sbtc/sbtc/commit/adc10327dc7e98adb59da87f3fd725b101a0737a

web/deposit-address-verifier/src/main.ts
djordon-stacks: I think this was done in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/c617b1e954748901cb07cad661b219ba4e484f09

.github/workflows/deploy-deposit-address-verifier.yaml
```diff
@@ -0,0 +1,74 @@
...
+        uses: stacks-sbtc/actions/setup-pnpm@505c42b153ba2ac261c276e96bc5dc7b68d61039
+
+      - name: Setup Node
+        uses: stacks-sbtc/actions/setup-node@505c42b153ba2ac261c276e96bc5dc7b68d61039
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: "pnpm"
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
```
hugo-stacks:
```suggestion
        run: pnpm install --frozen-lockfile --filter @stacks-sbtc/deposit-address-verifier...
```

Avoid installing the whole workspace dependencies. Should speed it up.

The trailing `...` also pulls in any workspace dependencies the package might gain later; today it has none

web/deposit-address-verifier/src/main.ts
```diff
@@ -0,0 +1,190 @@
...
+      connectButton.textContent = "Connect wallet";
+    } finally {
+      connectButton.disabled = false;
+    }
+  });
+
+  form.addEventListener("submit", async (event) => {
+    event.preventDefault();
+    invalidateResult();
+    if (!form.reportValidity()) return;
```
hugo-stacks: If there is an error in the `advanced` section, the user won't see it.

Suggestions;

Make the toggle collable;
```ts
const advancedToggle = document.querySelector<HTMLButtonElement>("#advanced-toggle")!;

const setAdvancedExpanded = (expanded: boolean): void => {
  advancedToggle.setAttribute("aria-expanded", String(expanded));
  advancedToggle.lastElementChild!.textContent = expanded ? "−" : "＋";
  advancedFields.hidden = !expanded;
};

advancedToggle.addEventListener("click", () => setAdvancedExpanded(advancedFields.hidden));
```

Replace the bare `reportValidity()` call:
```ts
function reportFormValidity(): boolean {
  if (form.checkValidity()) return true;

  const invalid = form.querySelector<HTMLInputElement | HTMLSelectElement | HTMLTextAreaElement>(
    "input:invalid, select:invalid, textarea:invalid",
  );
  if (invalid?.offsetParent === null) setAdvancedExpanded(true);

  if (invalid && invalid.offsetParent === null) {
    showError(error, new Error(invalid.validationMessage));
  } else {
    form.reportValidity();
  }
  return false;
}
```

At the call site
```ts
    invalidateResult();
    if (!reportFormValidity()) return;
```

web/deposit-address-verifier/src/deposit.ts
```diff
@@ -0,0 +1,226 @@
...
+    if (!output.address) throw new Error("The Taproot deposit address could not be constructed.");
+    return {
+      address: output.address,
+      depositScript: bytesToHex(depositScript),
+      reclaimScript: bytesToHex(reclaimScript),
+    };
+  }
+
+  if (!input.reclaimPublicKey) throw new Error("Enter a reclaim public key.");
+  if (!Number.isInteger(input.lockTime) || input.lockTime < 0 || input.lockTime > 65_535) {
```
hugo-stacks, line 174: A lock time this low produces an address the signers will never sweep.

`signer/src/bitcoin/validation.rs` computes `max_age = height.saturating_sub(DEPOSIT_LOCKTIME_BLOCK_BUFFER)` (the buffer is `3`, from `signer/src/lib.rs`) and returns `LockTimeExpiry` when `deposit_age >= max_age`. For a lock time of 0 through 3 that is `deposit_age >= 0`, which is always true — the deposit is rejected on every attempt.

I checked this against the built site: lock time `0` returns `bc1pppgvmh67ku2hh0mjhyt803prkpnjl9v6xwul5jesctqwdth0f62syyd3ds` with no warning at all. A user who funds that address gets no sBTC and has to fall back to the reclaim path.

```suggestion
  if (!Number.isInteger(input.lockTime) || input.lockTime > 65_535) {
    throw new Error("Lock time must be a whole number of at most 65,535 blocks.");
  }
  // signer/src/bitcoin/validation.rs treats a deposit as expired once
  // deposit_age >= lock_time - DEPOSIT_LOCKTIME_BLOCK_BUFFER, so anything at or
  // below the buffer can never be swept.
  if (input.lockTime <= DEPOSIT_LOCKTIME_BLOCK_BUFFER) {
    throw new Error(
      `A lock time of ${input.lockTime} is always treated as expired. Use more than ${DEPOSIT_LOCKTIME_BLOCK_BUFFER} blocks; the official bridge uses 950.`,
    );
  }
```

with `const DEPOSIT_LOCKTIME_BLOCK_BUFFER = 3;` next to `SBTC_DEPLOYERS`, and `min="0"` changed to `min="4"` on `#lock-time` (`index.html:191`).

One nuance worth deciding: 4 is the *hard* floor derived from the buffer, but a deposit at 4 only sweeps if it lands in the same block it confirmed, so it is still unusable in practice. Rejecting `<= 3` is the factual protocol rule; whether to also warn below some practical threshold (100? 500?) is a judgement call I would leave to you rather than hard-code here.

.github/workflows/deploy-deposit-address-verifier.yaml
```diff
@@ -0,0 +1,74 @@
+name: Deploy sBTC address constructor
+
+on:
+  workflow_dispatch:
+
+permissions: read-all
+
+concurrency:
+  group: pages-deposit-address-verifier
+  cancel-in-progress: true
```
hugo-stacks: Cancelling an in-flight run can abort a Pages deployment partway through.

If the cancellation lands between `upload-pages-artifact` and `deploy-pages`, the deployment is left incomplete and the site can serve a stale or half-updated build until the next successful deploy.

GitHub's own Pages starter workflow sets this to `false` for exactly this reason, with the comment: *"do NOT cancel in-progress runs as we want to allow these production deployments to complete."* Queued runs will wait instead of killing a live deployment.

```suggestion
  cancel-in-progress: false
```

web/deposit-address-verifier/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@stacks-sbtc/deposit-address-verifier",
+  "version": "0.1.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc --noEmit && vite build",
```
hugo-stacks: `tsc` never runs on pull requests today.

`.github/workflows/on-push.yaml:106` runs `pnpm --recursive typecheck`, but this package exposes `tsc --noEmit` only inside `build`, so the recursive typecheck skips it. A type error passes CI and merges, then only surfaces when someone manually triggers the deploy workflow — after the approval gate.

`contracts/package.json` already defines a `typecheck` script, so this is aligning with the existing repo convention rather than adding a new one.

```suggestion
    "build": "pnpm run typecheck && vite build",
    "typecheck": "tsc --noEmit",
```

(Formatting is already covered — `make lint` runs `pnpm --recursive run lint`, so `oxfmt --check` does gate PRs.)

web/deposit-address-verifier/src/deposit.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  computeDepositAddress,
+  findWalletValues,
+  normalizeXOnlyPublicKey,
+  scriptToAsm,
+} from "./deposit";
+
+const KEY = "033920f589c2b367400732d2dd61d11b300ad95b2b1bbf008eabcf8cddfee0c12c";
+const MAINNET_RECIPIENT = "SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4";
+
+describe("deposit address computation", () => {
```
hugo-stacks, line 13: All 16 tests here cover `deposit.ts`. `main.ts` has no coverage at all, and that is where the ~190-line state machine lives that toggles `required`, `disabled` and `hidden` across the two reclaim modes and the collapsible advanced section.

That untested interaction is exactly where the silent-submit bug I commented on in `main.ts` comes from, which is probably why it went unnoticed.

`vitest` is already a dependency, so this only needs `environment: "jsdom"` plus `jsdom` as a devDependency. The regression test is short — and note this path needs no field editing at all, just three interactions:

```ts
it("surfaces validation errors from the collapsed advanced section", async () => {
  document.body.innerHTML = await readFile("index.html", "utf8");
  await import("./main");

  const script = document.querySelector<HTMLInputElement>('input[name="reclaimMode"][value="script"]')!;
  script.checked = true;
  script.dispatchEvent(new Event("change", { bubbles: true }));
  document.querySelector<HTMLButtonElement>("#advanced-toggle")!.click();

  document.querySelector<HTMLFormElement>("#verifier-form")!.requestSubmit();

  expect(document.querySelector<HTMLElement>("#form-error")!.hidden).toBe(false);
});
```

Worth pinning the mode toggles generally while you are in here — they are the fiddliest part of the page.

web/deposit-address-verifier/index.html
```diff
@@ -0,0 +1,348 @@
...
+                    data-tooltip="The number of Bitcoin blocks that must be mined after the deposit is confirmed before the standard reclaim path can be spent. The official bridge currently uses 950 blocks."
+                    >?</span
+                  ></span
+                >
+                <input
+                  id="lock-time"
+                  name="lockTime"
+                  type="number"
+                  value="950"
+                  min="0"
```
hugo-stacks, line 191: Pairs with the lock-time floor in `deposit.ts` — keeps the browser's own constraint in step with the protocol rule.

```suggestion
                  min="4"
```

hugo-stacks (commented):
Moving the markup into real static pages and adding oxfmt to CI made this much
easier to review, thanks.

A few comments below. Two one-liners, and the lock-time floor is a small fix.
The invisible error is something my agent caught twice

.github/workflows/deploy-deposit-address-verifier.yaml
djordon-stacks: Updated in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/ba16ac4cbf3b68216618faebff9f31dee7689957

web/deposit-address-verifier/package.json
djordon-stacks: Fixed in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/cd708b959dfa61ce71ebff8a61912743e13a7f98

.github/workflows/deploy-deposit-address-verifier.yaml
djordon-stacks: Updated in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/ba16ac4cbf3b68216618faebff9f31dee7689957

web/deposit-address-verifier/src/main.ts
djordon-stacks: Fixed in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/a418d44817bd394ea62a3c379951049ec4979aa3

web/deposit-address-verifier/src/deposit.test.ts
djordon-stacks: Tests added in https://github.com/stacks-sbtc/sbtc/pull/2113/commits/8d0a23a3f9cfbe911e96480c52cf7ce4f2e80f6b

web/deposit-address-verifier/src/deposit.ts
djordon-stacks: This change isn't a bad idea, but it shifts the site's purpose from an address construction/verification tool to checking for sBTC signer acceptance. For example, in advanced mode, the site doesn't validate the reclaim script, including its length or the presence of invalid opcodes. I don't want to expand the site's scope yet, so let's add this functionality later, where we tell the user whether the protocol would reject an address and why.

web/deposit-address-verifier/index.html
djordon-stacks: I kept this as-is, but we can revisit it later. Details in https://github.com/stacks-sbtc/sbtc/pull/2113#discussion_r3980271929.

hugo-stacks (commented):
lgtm

adriano-stacks (approved):

cylewitruk-stacks (approved):

